Issue #75, opened 2023-04-06
[Language Specification] Changes to qspan


amccaskey:
# Background 

The `qspan` type has represented thus far a sub-view of an existing container of qubits. This type has non-owning semantics and is typically used as an argument to a pure device kernel, as well as the return type on existing qreg / qvector slicing and sub-register extraction. The qspan is meant to model the standard span. This type may be a bit unknown to the current quantum computing community, and the span is meant to provide a mutable view (non-owning) of existing containers, which is not what we want here with qubits. 

# Proposal 
We propose to change the qspan name to `qview` to better model it as a non-owning view on an existing vector or array of qubits.

bettinaheim:
What exactly are we hoping to achieve with this? I am a bit doubtful about the value of a name change given the amount of docs and samples that people are already familiar with.

amccaskey:
The reasons are laid out pretty well above in the original comment. This was discussed extensively over 2 language specification meetings and consensus was reached that `qspan` -> `qview` is a change that should be made. Same with `qreg` -> `qvector` and `qarray` #38. The main reason is the semantics conveyed by `span`, which implies mutability of the underlying data. The true semantics we want here is a non-owning view, hence `qview`. The changes have already been made in the ASTBridge and the runtime library. I think all that remains is marking the old types as deprecated. 

The only real questions left are regarding `qvector` and Python (e.g. in that language embedding, perhaps it should be a `qlist`)

bettinaheim:
Alright, let's move forward with this, with the appropriate deprecations. Let's aim to keep naming stable within each major release come 1.0 and later.

PR #1016, opened 2023-12-08
[NFC] Disable warnings from the kernel_builder.

schweitzpgi:
The kernel builder still uses qreg, a deprecated type, which is generating a lot of warnings. This change silences the warnings. When qreg is finally removed, this change should be reverted.

See #75.


khalatepradnya:
The changes look good. Can you help me understand why this warning should be suppressed - as in, isn't it helpful to show deprecation message so users know it's going away and start using alternative? Or is the problem with 'too many warnings' - in that case can we show it once per program?

1tnguyen (approved):

PR #966, opened 2023-11-22
MPI support as plugins

1tnguyen:
### Description

Implementing MPI support as plugins as proposed in this [RFC](https://github.com/NVIDIA/cuda-quantum/issues/802). 

This helps improve compatibility in some deployment scenarios:

- Docker images: `libcudaq.so` is no longer directly linked against OpenMPI in the docker image; hence allowing MPI lib swap when using the image.

- Python wheels: with dynamical plugins, we don't need to [disable MPI support during build](https://github.com/NVIDIA/cuda-quantum/issues/920). MPI support is deferred to runtime with support for mpi4py-based plugin or manually activated native MPI plugin.

### Testing and CI

CI tests cover:

- All 3 plugin deployment scenarios: built-in, mpi4py-based, and manually activated (`test_in_devenv.yml`)

- OpenMPI and MPICH. To keep CI matrix size manageable, MPICH is only included in the matrix for the gcc toolchain.

- Test MPI support in wheels (`Run Python MPI tests` in `python_wheels.yml`): activated for Ubuntu only.

PR #832, opened 2023-10-27
Fix workflows for new GHA runners

jjacobelli:
### Description
The new GHA runners are running a rootless `dockerd` which requires the following changes:
- Fix file permissions before using `docker cp`
- Use local action to run a command inside a container instead of using the not maintained `addnab/docker-run-action` action

Working pipeline on new runners: https://github.com/NVIDIA/cuda-quantum/actions/runs/6706737406/job/18223818783?pr=832

PR #1018, opened 2023-12-10
[DO NOT MERGE] Investigate CI (suspicious of run-in-docker issues when mixed with MPI)

bmhowe23:
Some of our CI tests are skipping tests. I think I know the problem (and if so, I may have a fix), but first I'm investigating how pervasive the problem is in order to confirm my suspicions.

### DIFF
--- a/.github/actions/run-in-docker/action.yml
+++ b/.github/actions/run-in-docker/action.yml
@@ -15,4 +15,5 @@ runs:
       run: |
         docker run -i ${{ inputs.image }} ${{ inputs.shell }} -s << 'EOF'
         ${{ inputs.run }}
+        echo "BMH MADE IT TO THE END (inside run-in-docker)"
         EOF
--- a/.github/workflows/python_wheels.yml
+++ b/.github/workflows/python_wheels.yml
@@ -215,6 +215,7 @@ jobs:
                 exit 1
               fi
             done
+            echo "BMH MADE IT TO THE END"
 
       - name: Run Python MPI tests
         if: matrix.os_image == 'redhat/ubi9:9.2'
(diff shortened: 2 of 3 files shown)


bmhowe23:
I added "got here" statements to the end of all run-in-docker actions, and as suspected, we are not reaching the end of all of them.

✅ means that we saw the "got here" statements and the test scripts are working as intended.
❌ means that we did _not_ see the "got here" statements, so our test scripts are not working as intended.

### Results

[Build and test (all matrix elements) / Dev environment (Debug)](https://github.com/NVIDIA/cuda-quantum/actions/runs/7158947684/job/19492306659?pr=1018)
❌  Test CUDA Quantum
❌  Test CUDA Quantum MPI4Py Plugin
❌  Test CUDA Quantum MPI4 Plugin Activation

[Build and test (all matrix elements) / Dev environment (Python)](https://github.com/NVIDIA/cuda-quantum/actions/runs/7158947684/job/19492306726?pr=1018)
✅  Build and test CUDA Quantum (Python)

[Create Python wheels (all matrix elements) / Validate wheel (redhat/ubi9:9.2, --user)](https://github.com/NVIDIA/cuda-quantum/actions/runs/7158947684/job/19492455845?pr=1018)
✅  Run Python tests
❌  Run Python MPI tests

I suspect the reason for the change in behavior is a combination of #966 with #832, where the `mpirun` command is causing our `run-in-docker` commands to terminate early. When this happens, it means that
a) our subsequent Bash commands that check the status of tests are not working, and
b) any tests that follow the `mpirun` in the action don't actually get executed.

bmhowe23:
Closing. The fix for the issue captured here is in #1019 .

PR #1024, opened 2023-12-11
[API] Remove qreg from builder.

schweitzpgi:


schweitzpgi:
As feared, this interface is actually called. Closing this.

PR #1029, opened 2023-12-12
Fix qreg/qspan deprecation warnings in build

### LINKED ISSUE #1025: Remove qreg use from kernel_builder.{h,cpp}
See https://github.com/NVIDIA/cuda-quantum/pull/1024/files for the changes.

This requires removing the calls to this interface as well.

### LINKED ISSUE #1026: Remove qreg use from qubit_qis.h
qreg is used in a concept in this header. Can the concept be deleted?

amccaskey:
Fix #1025 #1026. 

Also update `cudaq::range()` to support stop and step values #1028

### DIFF
--- a/runtime/cudaq/builder/kernel_builder.h
+++ b/runtime/cudaq/builder/kernel_builder.h
@@ -10,7 +10,6 @@
 
 #include "cudaq/builder/QuakeValue.h"
 #include "cudaq/qis/modifiers.h"
-#include "cudaq/qis/qreg.h"
 #include "cudaq/qis/qvector.h"
 #include "cudaq/utils/cudaq_utils.h"
 #include "host_config.h"
@@ -74,12 +73,11 @@ concept KernelBuilderArgTypeIsValid =
 // If you want to add to the list of valid kernel argument types first add it
 // here, then add `details::mapArgToType()` function
 #define CUDAQ_VALID_BUILDER_ARGS_FOLD()                                        \
-  requires(                                                                    \
-      KernelBuilderArgTypeIsValid<                                             \
-          Args, float, double, std::size_t, int, std::vector<int>,             \
-          std::vector<float>, std::vector<std::size_t>, std::vector<double>,   \
-          cudaq::qubit, cudaq::qreg<>, cudaq::qvector<>> &&                    \
-      ...)
+  requires(KernelBuilderArgTypeIsValid<                                        \
+               Args, float, double, std::size_t, int, std::vector<int>,        \
+               std::vector<float>, std::vector<std::size_t>,                   \
+               std::vector<double>, cudaq::qubit, cudaq::qvector<>> &&         \
+           ...)
 #else
 // Not C++ 2020: stub these out.
 #define QuakeValueOrNumericType typename
@@ -134,9 +132,6 @@ KernelBuilderType mapArgToType(std::vector<double> &e);
 
 /// Map a `qubit` to a `KernelBuilderType`
 KernelBuilderType mapArgToType(cudaq::qubit &e);
-
-/// @brief  Map a `qreg` to a `KernelBuilderType`
-KernelBuilderType mapArgToType(cudaq::qreg<> &e);
 
 /// @brief  Map a `qvector` to a `KernelBuilderType`
 KernelBuilderType mapArgToType(cudaq::qvector<> &e);
--- a/runtime/cudaq/builder/kernel_builder.h
+++ b/runtime/cudaq/builder/kernel_builder.h
@@ -23,15 +23,6 @@
 #include <variant>
 #include <vector>
 
-#if defined(__clang__)
-#pragma clang diagnostic push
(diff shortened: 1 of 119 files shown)


runtime/cudaq/builder/kernel_builder.h
```diff
@@ -10,7 +10,6 @@
 
 #include "cudaq/builder/QuakeValue.h"
 #include "cudaq/qis/modifiers.h"
-#include "cudaq/qis/qreg.h"
 #include "cudaq/qis/qvector.h"
 #include "cudaq/utils/cudaq_utils.h"
 #include "host_config.h"
```
schweitzpgi, line 15: Can also removed lines 26-33.

amccaskey: Ah yes, thanks.

runtime/cudaq/utils/cudaq_utils.h
```diff
@@ -240,12 +240,30 @@ std::vector<double> random_vector(const double l_range, const double r_range,
...
-inline std::vector<std::size_t> range(int N) {
-  std::vector<std::size_t> vec(N);
-  std::iota(vec.begin(), vec.end(), 0);
+/// @brief Return a vector of integers. The first element is the
+/// user-specified `start` value. The remaining values are all values
+/// incremented by `step` (defaults to 1) until the `stop` value is reached
+/// (exclusive).
+template <typename ElementType>
+inline std::vector<ElementType> range(ElementType start, ElementType stop,
+                                      ElementType step = 1) {
```
schweitzpgi, line 249: Did you update the bridge for this? When I implemented this in the bridge I remember wondering about the default step value.

test/AST-Quake/passQregNToQSpan.cpp
```diff
@@ -29,7 +29,7 @@ struct entry {
 // CHECK-LABEL:   func.func @__nvqpp__mlirgen__entry() attributes {
 // CHECK:           %[[VAL_1:.*]] = quake.alloca !quake.veq<3>
 // CHECK:           %[[VAL_2:.*]] = quake.relax_size %[[VAL_1]] : (!quake.veq<3>) -> !quake.veq<?>
-// CHECK:           call @__nvqpp__mlirgen__function_mcx._Z3mcxN5cudaq5qspan{{.*}}(%[[VAL_2]]) : (!quake.veq<?>) -> ()
+// CHECK:           call @__nvqpp__mlirgen__function_mcx._Z3mcxN5cudaq5{{.*}}(%[[VAL_2]]) : (!quake.veq<?>) -> ()
```
schweitzpgi, line 32:
```suggestion
// CHECK:           call @__nvqpp__mlirgen__function_mcx._Z3mcxN5cudaq5qview{{.*}}(%[[VAL_2]]) : (!quake.veq<?>) -> ()
```
?

schweitzpgi (approved):
LGTM. Thanks, Alex, for helping drive the stake in this one.

runtime/cudaq/utils/cudaq_utils.h
amccaskey: I did not. I thought I would leave that for a follow up. I'll add the ticket.

test/AST-Quake/passQregNToQSpan.cpp
amccaskey: I had that, but then I figured why not just slurp the qview up too. Ultimately I thought that keeping qview was just a test that Clang can mangle the function signature so I decided we didn't need to test that here.

schweitzpgi: That's fine. I thought this might be trying to disambiguate the call from something else.

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -11,7 +11,6 @@
 #include "common/MeasureCounts.h"
 #include "cudaq/qis/modifiers.h"
 #include "cudaq/qis/qarray.h"
-#include "cudaq/qis/qreg.h"
```
bmhowe23, line 14: I know this is already merged, but ...
While this particular change eliminates the warnings in our build, it will also break any user code that is using `qreg` (as evidenced by #1039), unless they specifically add this `#include` into their source code. So, as written, I think this is a breaking change. We may need to add additional documentation in the release notes that specifies that additional `#include`s are needed if the user wants to use `qreg` with our next release.

bmhowe23: > I know this is already merged, but ... While this particular change eliminates the warnings in our build, it will also break any user code that is using `qreg` (as evidenced by #1039), unless they specifically add this `#include` into their source code. So, as written, I think this is a breaking change. We may need to add additional documentation in the release notes that specifies that additional `#include`s are needed if the user wants to use `qreg` with our next release. Another option would be to add this line back in before our 0.6 release.

#1042 contains a possible alternative to try to keep the best of both worlds. Let me know what you think.

Issue #1030, opened 2023-12-12
Incorrect QIR generated for some conditional measurements (qir_cond_for_loop-6.cpp)


bmhowe23:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

This issue was captured in #783, but #783 was closed because the local emulation was no longer failing. However, the underlying invalid QIR generation issue still exists.

Specifically, For certain unrolled-loop test cases ([like the one in the title](https://github.com/NVIDIA/cuda-quantum/blob/fb4e42e8df0226d030fbc6b4b1c2baf6036cc5f8/test/NVQPP/qir_cond_for_loop-6.cpp#L17)), the compiler is injecting `__quantum__rt__result_record_output` calls even if `if` checks prevent measurements from occurring.

The compiler should not inject those calls for cases like this.

### Steps to reproduce the bug

See [this gist](https://gist.github.com/bmhowe23/cedfe072c2640f8e5c1546b14899472e) for the resulting QIR coming from the compiler at the time of this issue creation (freshly regenerated at the time of the creation of this new issue).

The QIR can also be generated like this in the dev image like this:
```bash
$ nvq++ --target quantinuum --emulate test/NVQPP/qir_cond_for_loop-6.cpp
$ CUDAQ_DUMP_JIT_IR=1 ./a.out
```

You can also run it on the provider hardware in order to see a test actually fail:
```bash
$ nvq++ --target quantinuum --quantinuum-machine H1-1E test/NVQPP/qir_cond_for_loop-6.cpp
$ ./a.out
{ 
  q0result%03 : { 1:8 0:92 }
   q0result%04 : { 1:3 0:97 }
   q0result%10 : { 0:100 }
   q0result%02 : { 1:11 0:89 }
   q0result%07 : { 1:1 0:99 }
   q0result%14 : { 0:100 }
   q0result%08 : { 0:100 }
   q0result%01 : { 1:25 0:75 }
   q0result%12 : { 0:100 }
   q0result%13 : { 0:100 }
   q0result%09 : { 0:100 }
   q0result%11 : { 0:100 }
[...]

PR #1031, opened 2023-12-12
Move QIR result recording calls right after measurement for Adaptive Profile

bmhowe23:
The Base Profile requires output recording calls to come at the end of the program, but the Adaptive Profile does not require that, and depending on the program, it may be logically inconsistent to do so. This changes the `LowertoQIRProfile.cpp` code to make output recording calls come right after the corresponding measurements for Adaptive Profile programs.

Note that this fixes the QIR generation issue in #1030, but the Quantinuum backend still generates results for measurements that should not have occurred.

Note: I have also tested these changes by submitting all of our integration tests to Quantinuum's `H1-1SC` and a few select tests to `H1-1E`. They all still pass.


schweitzpgi (approved):
Ship it.

PR #1033, opened 2023-12-12
[DON'T MERGE] density matrix simulator bit ordering

anthony-santana:


unittests/backends/QPPTester.cpp
```diff
@@ -1241,3 +1280,75 @@ CUDAQ_TEST(QPPTester, checkReset) {
...
+    auto q0 = qppBackend.allocateQubit();
+    // Passes.
+    EXPECT_EQ(0, qppBackend.mz(q0));
+    
+    // Rotate to |1>
+    qppBackend.x(q0);
+    // Passes.
+    EXPECT_EQ(1, qppBackend.mz(q0));
+
+    // Add another qubit. Individually, should be |0>.
```
anthony-santana, line 1342: If you return the density matrix here, you see that the matrix itself is still what we'd expect.

unittests/backends/QPPTester.cpp
```diff
@@ -1241,3 +1280,75 @@ CUDAQ_TEST(QPPTester, checkReset) {
...
+    auto q1 = qppBackend.allocateQubit();
+    // Fails. Qubits have flipped. Even when directly addressing
+    // the variable for q1.
```
anthony-santana, line 1345: It actually seems that the variable has entirely flipped as well. If you do the following instead of measure
```
    qppBackend.x(q1);
```
the 2 tests above suggest we should now be in the |11> state. Instead, we return to |00>, as if `q1` is now addressing the qubit that was previously stored in `q0`.

runtime/nvqir/qpp/QppDMCircuitSimulator.cpp
```diff
@@ -69,11 +69,11 @@ class QppNoiseCircuitSimulator : public nvqir::QppCircuitSimulator<qpp::cmat> {
...
-    state.conservativeResize(stateDimension, stateDimension);
-    for (std::size_t i = previousStateDimension; i < stateDimension; i++) {
-      state.col(i).setZero();
-      state.row(i).setZero();
-    }
+    // Additional qubit is added in the |0><0| state.
+    qpp::cmat zero_state = qpp::cmat::Zero(2, 2);
+    zero_state(0, 0) = 1.0;
+    state = qpp::kron(state, zero_state);
+    return;
```
bmhowe23, line 76: Nit: this `return` and the one on line 93 are not necessary and could be removed.

bmhowe23 (approved):
Nice 👍

runtime/nvqir/qpp/QppDMCircuitSimulator.cpp
schweitzpgi, line 76: nit: `return` is not needed here

runtime/nvqir/qpp/QppDMCircuitSimulator.cpp
```diff
@@ -87,12 +87,9 @@ class QppNoiseCircuitSimulator : public nvqir::QppCircuitSimulator<qpp::cmat> {
...
-    state.conservativeResize(stateDimension, stateDimension);
-    for (std::size_t i = previousStateDimension; i < stateDimension; i++) {
-      state.col(i).setZero();
-      state.row(i).setZero();
-    }
-
+    qpp::cmat zero_state = qpp::cmat::Zero(1 << count, 1 << count);
+    zero_state(0, 0) = 1.0;
+    state = qpp::kron(state, zero_state);
     return;
```
schweitzpgi, line 93: nit: not needed

unittests/backends/QPPTester.cpp
```diff
@@ -26,6 +26,20 @@ void print_state(const qpp::ket &stateVector) {
   std::cout << "]\n";
 }
 
+void print_state(const qpp::cmat &densityMatrix) {
+  std::cout << "state = [";
+  auto rows = densityMatrix.rows();
+  auto cols = densityMatrix.cols();
+  for (auto rdx = 0; rdx < rows; rdx++) {
```
schweitzpgi, line 33: Looks old school. Does this not support standard iterators?

schweitzpgi (commented):
LGTM. Thanks for the fixes.

unittests/backends/QPPTester.cpp
anthony-santana: There may be a cleaner way to do this with Eigen::IOFormat , I just wrote this up quickly as a way to format the output as a nice, square matrix

1tnguyen (approved):
Looks great👍

PR #1035, opened 2023-12-13
Add vector<vector<...>> support to CC.

schweitzpgi:
This allows the IR to express vectors of vectors of ... recursively. Add a test for quake and for lowering to QIR. Bonus: Fix a bug in GKE.

### DIFF
--- a/test/Quake/vector.qke
+++ b/test/Quake/vector.qke
@@ -1,0 +1,138 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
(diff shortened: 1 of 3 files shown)


test/Quake/vector.qke
```diff
@@ -0,0 +1,138 @@
...
+    cc.continue %arg0 : i64
+  } step {
+   ^bb0(%arg0 : i64):
+    %next = arith.addi %arg0, %one : i64
+    cc.continue %next : i64
+  }
+  return
+}
+
+// CHECK-LABEL:   func.func @vector_vector(
```
amccaskey, line 59: Is there any way for us to test that this LLVM code works when executed? Maybe that can come later when we wire this through the bridge.

amccaskey (approved):
LGTM

PR #1041, opened 2023-12-16
Make nvqpp docs tests fail if compilation fails

bmhowe23:
Without this change, if a compilation failure occurs and an old executable already existed in the developer's workspace, the old executable would be run and assessed for pass/fail criteria. This change ensures that if the compilation fails, the test will be marked as a failure. (This may be how the `nvqpp_SampleAsync` test may have escaped detection as part of #1029).


amccaskey (approved):
+1

schweitzpgi (approved):

docs/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ function(add_nvqpp_test TEST_NAME SOURCE_LOCATION)
   NAME
     nvqpp_${TEST_NAME}
   COMMAND
-    bash -c "${CMAKE_BINARY_DIR}/bin/nvq++ ${NVQPP_COMPILE_ARGS} ${PARSED_ARGS_SOURCE_DIR}/${SOURCE_LOCATION} -o ${TEST_NAME} ;\
+    bash -c "${CMAKE_BINARY_DIR}/bin/nvq++ ${NVQPP_COMPILE_ARGS} ${PARSED_ARGS_SOURCE_DIR}/${SOURCE_LOCATION} -o ${TEST_NAME} && \
```
schweitzpgi, line 30: Might want to `rm -f ${TEST_NAME}` just before this as well.

bmhowe23: I think the `&&` should make any additional cleanup (like `rm`) unnecessary. I did multiple tests with it to make sure it fails when compilation fails. (Plus, the `rm` command would emit an error if the file wasn't found, which could be confusing to people looking at the output when running `ctest ... -V`.

schweitzpgi: The `-f` won't fail or issue an error.

the point of deleting the file would be to prevent the false positive scenario of the compiler being a NOP==`true` and running an old, stale file. Far-fetched perhaps, but we're talking about computers here. :)

bmhowe23: OK, updated in c9c58b17

bmhowe23: FWIW, many of our llvm-lit tests in tests/NVQPP would be equally vulnerable to these "far-fetched" scenarios, too.

PR #1042, opened 2023-12-16
Gracefully deprecate qreg rather than fully removing from qubit_qis.h

bmhowe23:
`cudaq::qreg` and `cudaq::qspan` were deprecated in #983, and #1029 cleaned up all the warnings emitted from our cuda-quantum build, but that change made it harder for users to continue to utilize the deprecated elements because they would have to know that they need to manually add `#include "cudaq/qis/qreg.h"` to their programs. This change alleviates the users of that (likely hard to learn) requirement. There may be other solutions to this problem, so I'm open to other ideas, too.

~~This change creates a new `#define` called `CUDAQ_EXCLUDE_QREG_HEADERS`, and enables that `#define` for our project via CMake updates. This enables us to retain all the benefits of #1029 (i.e., a clean cuda-quantum build with no warnings) while still allowing external users (who won't inherit this new `#define` because they aren't using our CMake files) to `#include <cudaq.h>` like they've always done, and that will properly define `qreg` and `qspan` for them.~~

~~Once support for `qreg` and `qspan` is officially dropped, all the references to `CUDAQ_EXCLUDE_QREG_HEADERS` would need to be removed.~~

Later, it was discovered that d33caacb32a018c70b3779a4ee0f866e8967b4d0 will achieve the same goal that was previously achieved with `CUDAQ_EXCLUDE_QREG_HEADERS`, and it is cleaner, so the aforementioned `#define` portions of this PR were reverted.

### DIFF
--- a/runtime/cudaq/qis/qspan.h
+++ b/runtime/cudaq/qis/qspan.h
@@ -15,6 +15,15 @@
 #include <span>
 #endif
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
 namespace cudaq {
 #if CUDAQ_USE_STD20
 inline constexpr auto dyn = std::dynamic_extent;
@@ -113,3 +122,10 @@ public:
 #endif
 };
 } // namespace cudaq
+
+#if defined(__clang__)
+#pragma clang diagnostic pop
+#endif
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,6 @@ set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 if(NOT DEFINED CMAKE_COMPILE_WARNING_AS_ERROR)
   set(CMAKE_COMPILE_WARNING_AS_ERROR ON)
 endif()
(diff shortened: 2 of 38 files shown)


bettinaheim (approved):
Looks good to me. Thank you for making this non-breaking.

schweitzpgi:
I think the
```
#include "qreg.h"
```
line just needs to be added back. It should not have been removed since it drops the deprecated feature completely, which is not what was intended.

On the other hand, we should fully expect warnings from deprecated features. This is just "the way the world works" as far as deprecating features. This is in diametric opposition to setting `-Werror` on everything in the build, and that is understood. Still, we should be emitting the warnings for deprecated features that are referenced anywhere but from within themselves.

schweitzpgi:
The other alternative here, if we're really not ok with warnings and thus are not ok with deprecation, is to remove the obsolete features entirely.

I don't like half measures and having hidden preprocessor flags that build things different ways. That's a bad precedent since it can spiral into a lot of maintenance headaches when overused.

So let's decide if we're ok with warning/deprecation or we're not going to allow any warnings and are disallowing deprecation. If the former, restore the header files and -Wno-error flag, or if the latter, delete these header files.

runtime/cudaq/qis/qspan.h
```diff
@@ -15,6 +15,15 @@
...
 #endif
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
schweitzpgi, line 25: +1

schweitzpgi (approved):
Thanks, Ben.

PR #1043, opened 2023-12-18
Fix direct include of fmt (suppress warnings)

amccaskey:
Fix error in nightly CI where warning was emitted from direct inclusion of a `fmtlib` header.


schweitzpgi (commented):
Looks good to me

schweitzpgi (approved):

PR #1044, opened 2023-12-18
Fix library mode qubit ordering bug.

amccaskey:
Fix library mode bit ordering issue discussed in #978


bmhowe23 (approved):
Thanks, Alex. Additional clarifying documentation will be coming in #1027.

Issue #1046, opened 2023-12-19
Upgrade pybind11 to the latest version


khalatepradnya:
* Using pybind11 version 2.11.0+ results in three test failures in 'eager' mode under 'experimental/python' branch. These tests work as expected with pybind11 version 2.10.4.
* When the behavior is modified either on pybind11 or on CUDA Quantum, upgrade pybind11 to latest version.
* Origin:

> This issue was introduced with the update to pybind11 2.11 (appears to be a pybind bug). pybind version 2.10 and below work. Easiest fix would be to lower to pybind 2.10 until a fix appears in a future version. pybind 2.11 (appears to be a pybind bug). pybind version 2.10 and below work. Easiest fix would be to lower to pybind 2.10 until a fix appears in a future version.

_Originally posted by @amccaskey in https://github.com/NVIDIA/cuda-quantum/pull/1001#discussion_r1426640287_

```[tasklist]
### Tasks
- [x] Identify the changes / commit in pybind11 that introduced change in behavior.
- [ ] Check behavior with latest version of pybind11
- [ ] Triage if fix possible on CUDA Quantum
```

khalatepradnya:
Looking at changelog, pybind11 v2.11.0 introduced https://github.com/pybind/pybind11/pull/4631 and https://github.com/pybind/pybind11/pull/4623. These changes recursively check if a type is movable.
Previously pybind11 relied on `std::is_move_constructible`, now, it has own implementation.
Refer test case: https://github.com/pybind/pybind11/blob/master/tests/test_copy_move.cpp#L383
`qview` is copyable but, `qvector` and `qudits` are non-copyable and non-movable. Also, not sure if related, but, `std::span` isn't supported in default bindings. (https://github.com/pybind/pybind11/discussions/4177).

khalatepradnya:
Attempt to reproduce the behavior stand-alone:
1. C++ composite classes in `bar.hpp`
```#pragma once

#include <ranges>
#include <span>
#include <vector>

class NonCopyableNonMovableInt
{
private:
    int num;

public:
    NonCopyableNonMovableInt() {}
    NonCopyableNonMovableInt(int n) : num(n) {}

    NonCopyableNonMovableInt(NonCopyableNonMovableInt const &) = delete;
    NonCopyableNonMovableInt(NonCopyableNonMovableInt &&) = delete;

    int getValue() { return num; }
    void setValue(int n) { num = n; }
};

[...]

PR #1047, opened 2023-12-19
Downgrading 'pybind11' to version 2.10.4 from version 2.11.1

khalatepradnya:
Reason - the behavior captured in issue https://github.com/NVIDIA/cuda-quantum/issues/1046.

Task:
- [x] Does this diff still work or needs to be adjusted? -  https://github.com/NVIDIA/cuda-quantum/blob/main/tpls/customizations/pybind11/pybind.h.diff : WORKS, but, line numbers have changed, so updated the diff file.


bmhowe23 (commented):
In general this looks good to me, but do you know what caused the CI failure on arm?

khalatepradnya:
> In general this looks good to me, but do you know what caused the CI failure on arm?

It appears to have failed on LLVM checkout `fatal: unable to access 'https://github.com/llvm/llvm-project/': Failed to connect to github.com port 443: Connection timed out`. I am rerunning the step.

bmhowe23 (approved):
👍

PR #1050, opened 2023-12-20
Revert "Update CI to use iqm-client v15.0 (not v15.1) (#1048)"

bmhowe23:
This reverts commit f6f06cd7624fc2adab85cf394ca6408a2c1e2f1f.


bettinaheim (approved):

bettinaheim:
@bmhowe23 I think this would now use 15.2, not 15.0, correct?

bmhowe23:
> @bmhowe23 I think this would now use 15.2, not 15.0, correct?

This change will make the CI use whatever the latest released version is, which is 15.2 right now (https://pypi.org/project/iqm-client/).

PR #1051, opened 2023-12-20
Add end-to-end support for recursive vector arguments.

schweitzpgi:
This is the initial implementation to support `vector<vector<...<T>>` arguments to kernels. It adds code to GenKernelExecution to perform the argument packing/unpacking into the message buffer.

Returning recursive vector data types is *NOT* supported.

 - Implement the recursive unpacking of `vector<vector<...>>` in thunk.
 - Implement packing `vector<vector<...>>` arguments.
 - Refactor and share code with argsCreator.
 - Add test.
 - Fix regression tests.

As a bonus, this patch attempts to repair the `-g` option to `nvq++`.

### DIFF
--- a/include/cudaq/Optimizer/Builder/Factory.h
+++ b/include/cudaq/Optimizer/Builder/Factory.h
@@ -19,6 +19,7 @@
 namespace cudaq {
 namespace cc {
 class LoopOp;
+class PointerType;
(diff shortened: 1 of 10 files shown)


amccaskey (approved):
+1

PR #1053, opened 2023-12-20
Patch density matrix bug in release mode

anthony-santana:
### Description

* Fixes a bug occurring during release builds on the QPPDMTester
* Removes unneeded includes from dm tester


bmhowe23 (approved):
👍

khalatepradnya (approved):

PR #1054, opened 2023-12-20
Break ground on passing/return POD struct values.

schweitzpgi:
There is more to do here. Most notably the bridge doesn't support the ctors and other AST artifacts needed to build struct values, pass them around, etc.

Supports a POD struct if it is created by some external at this point.

### DIFF
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -158,10 +158,10 @@ matrix:
       - open: '(?P<open>`+)'
         close: '(?P=open)'
       # Ignore the first word after \command or @command, which is a code artifact
-      - open: '[\\@]([acep])(?:\s\S+)?'
+      - open: '[\\@]([acep])(?:\s+\S)?'
         close: '(\s|$)'
       # Ignore the doxygen \param command and the code artifact after it.
-      - open: '[\\@]param(\s+\[[a-z]+\])?\s+(?:\S+)'
+      - open: '[\\@]param(\s+\[\w+\])?(?:\s+\S)?'
         close: '(\s|$)'
       # Ignore words that start with a \ or @, e.g. @param used in doc comments
       - open: '[\\@](\w|[\\@&\$#<>%"\.=|]|::|--|---)'
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -157,11 +157,14 @@ matrix:
       # Ignore content between inline backticks
       - open: '(?P<open>`+)'
         close: '(?P=open)'
-      # Ignore the first lowercase word after @param, which is usually the parameter name
-      - open: '@param(?:\s[a-z]+)?'
+      # Ignore the first word after \command or @command, which is a code artifact
+      - open: '[\\@]([abcep]|em)(?:\s\S+)?'
         close: '(\s|$)'
-      # Ignore words that start with an @, e.g. @param used in doc comments
-      - open: '@\S'
+      # Ignore the doxygen \param command and the code artifact after it.
+      - open: '[\\@]param(\s+\[[a-z]+\])?\s+(?:\S+)'
+        close: '(\s|$)'
+      # Ignore words that start with a \ or @, e.g. @param used in doc comments
+      - open: '[\\@](\S+|[n\\@&\$#<>%"\.=|]|::|--|---)'
         close: '(\s|$)'
       # Ignore words before and after double columns
       - open: '([a-zA-Z0-9_])*::'
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -166,7 +166,7 @@ matrix:
       # Ignore words that start with a \ or @, e.g. @param used in doc comments
       - open: '[\\@](\w|[\\@&\$#<>%"\.=|]|::|--|---)'
         close: '(\s|$)'
-      # Ignore words before and after double columns
+      # Ignore words before and after double colons
       - open: '([a-zA-Z0-9_])*::'
         close: '([^a-zA-Z0-9_:]|$)'
       # Ignore words that contain any underscores, numbers, or uppercase letters
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -164,7 +164,7 @@ matrix:
       - open: '[\\@]param(\s+\[[a-z]+\])?\s+(?:\S+)'
         close: '(\s|$)'
       # Ignore words that start with a \ or @, e.g. @param used in doc comments
-      - open: '[\\@](\S+|[n\\@&\$#<>%"\.=|]|::|--|---)'
+      - open: '[\\@](\w|[\\@&\$#<>%"\.=|]|::|--|---)'
         close: '(\s|$)'
       # Ignore words before and after double columns
       - open: '([a-zA-Z0-9_])*::'
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -158,7 +158,7 @@ matrix:
       - open: '(?P<open>`+)'
         close: '(?P=open)'
       # Ignore the first word after \command or @command, which is a code artifact
-      - open: '[\\@]([abcep]|em)(?:\s\S+)?'
+      - open: '[\\@]([acep])(?:\s\S+)?'
         close: '(\s|$)'
       # Ignore the doxygen \param command and the code artifact after it.
       - open: '[\\@]param(\s+\[[a-z]+\])?\s+(?:\S+)'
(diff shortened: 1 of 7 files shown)


include/cudaq/Frontend/nvqpp/ASTBridge.h
```diff
@@ -580,7 +582,7 @@ class QuakeBridgeVisitor
 /// drives the process of walking the Clang AST and translate pertinent nodes to
 /// an MLIR Op tree containing Quake, CC, and other MLIR dialect operations.
 /// This Action will generate this MLIR Module and rewrite the input source code
-/// (using the Clang Rewriter system) to define quantum kernels as extern.
+/// (using the Clang `Rewriter` system) to define quantum kernels as `extern`.
```
amccaskey, line 585: This must be an old comment from way back when. We don't use the Rewriter in the bridge right?

schweitzpgi: I think we replaced all that with the fixup linkage postprocessing.

include/cudaq/Frontend/nvqpp/ASTBridge.h
```diff
@@ -131,10 +131,11 @@ bool ignoredClass(clang::RecordDecl *x);
 /// The general design is to walk the tree in a post-order traversal and
 /// assemble the IR from the leaves back down the tree. Traversals over types
 /// should push Type values to the type stack. Traversals over expressions
-/// should create IR in the ModuleOp as well as push subexpressions on the stack
-/// for parent nodes. A parent node always knows how many children it needs to
-/// be constructed correctly. The types of expressions are carried along with
-/// the expressions in the IR and need not be duplicated on the type stack.
+/// should create IR in the ModuleOp as well as push `subexpressions` on the
```
bmhowe23, line 134: It seems like we should add this to `.github/workflows/config/spelling_allowlist.txt` rather than annotating this as code.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1083,18 +1083,36 @@ bool QuakeBridgeVisitor::TraverseLambdaExpr(clang::LambdaExpr *x,
...
+  if (auto *field = dyn_cast<clang::FieldDecl>(x->getMemberDecl())) {
+    auto loc = toLocation(x->getSourceRange());
+    auto object = popValue(); // DeclRefExpr
+    std::int32_t offset = field->getFieldIndex();
+    auto ty = popType();
+    return pushValue(builder.create<cc::ComputePtrOp>(
+        loc, cc::PointerType::get(ty), object,
+        SmallVector<cc::ComputePtrArg>{0, offset}));
+  }
+  return true;
```
bmhowe23, line 1113: Do we want to throw an error if it's not a `FieldDecl`? I.e. if it's a `VarDecl` (for static data members), a `CXXMethodDecl`, or an `EnumConstantDecl`?

test/AST-Quake/struct.cpp
```diff
@@ -0,0 +1,92 @@
...
+// CHECK:           %[[VAL_3:.*]] = cc.load %[[VAL_2]] : !cc.ptr<i32>
+// CHECK:           %[[VAL_4:.*]] = cc.compute_ptr %[[VAL_1]][0, 1] : (!cc.ptr<!cc.struct<"ArgumentThing" {i32, i32, f64}>>) -> !cc.ptr<i32>
+// CHECK:           %[[VAL_5:.*]] = cc.load %[[VAL_4]] : !cc.ptr<i32>
+// CHECK:           %[[VAL_6:.*]] = cc.compute_ptr %[[VAL_1]][0, 2] : (!cc.ptr<!cc.struct<"ArgumentThing" {i32, i32, f64}>>) -> !cc.ptr<f64>
+// CHECK:           %[[VAL_7:.*]] = cc.load %[[VAL_6]] : !cc.ptr<f64>
+// CHECK:           call @_Z15debug_the_thingiid(%[[VAL_3]], %[[VAL_5]], %[[VAL_7]]) : (i32, i32, f64) -> ()
+// CHECK:           %[[VAL_8:.*]] = call @_Z16help_me_help_youv() : () -> !cc.struct<"ResultThing" {i1, i1, i1, i32}>
+// CHECK:           return %[[VAL_8]] : !cc.struct<"ResultThing" {i1, i1, i1, i32}>
+
+#if 0
```
bmhowe23, line 53: I assume these `#if 0` sections are intended to denote future work but not currently supported yet?

.github/workflows/config/spellcheck_config.yml
```diff
@@ -157,11 +157,14 @@ matrix:
       # Ignore content between inline backticks
       - open: '(?P<open>`+)'
         close: '(?P=open)'
-      # Ignore the first lowercase word after @param, which is usually the parameter name
-      - open: '@param(?:\s[a-z]+)?'
+      # Ignore the first word after \command or @command, which is a code artifact
+      - open: '[\\@]([abcep]|em)(?:\s\S+)?'
```
bmhowe23, line 161: I believe `\a`, `\e`, and `em` are for italics. Why don't we want to spellcheck the word in italics? (I don't believe they will always be code artifacts.)

I believe `\b` is for bold. Why don't we want to spellcheck the word in bold? (Similarly, I don't believe they will always be code artifacts.)

.github/workflows/config/spellcheck_config.yml
```diff
@@ -157,11 +157,14 @@ matrix:
...
         close: '(\s|$)'
-      # Ignore words that start with an @, e.g. @param used in doc comments
-      - open: '@\S'
+      # Ignore the doxygen \param command and the code artifact after it.
+      - open: '[\\@]param(\s+\[[a-z]+\])?\s+(?:\S+)'
+        close: '(\s|$)'
+      # Ignore words that start with a \ or @, e.g. @param used in doc comments
+      - open: '[\\@](\S+|[n\\@&\$#<>%"\.=|]|::|--|---)'
```
bmhowe23, line 167: I'm not sure I follow what `[n\\@&\$#<>%"\.=|]|::|--|---` is trying to catch, but I believe the following complete changes would work for you.

From:
```
      # Ignore the first lowercase word after @param, which is usually the parameter name
      - open: '@param(?:\s[a-z]+)?'
```
To:
```
      # Ignore the first word after @param, which is usually the parameter name
      - open: '@param(?:\s\w+)?'
        close: '(\s|$)'
      # Ignore the first word after \p, which is usually the parameter name
      - open: '\\p(?:\s\w+)?'
        close: '(\s|$)'
```

test/AST-Quake/struct.cpp
schweitzpgi: Yes.

.github/workflows/config/spellcheck_config.yml
schweitzpgi: The problem is that they can be mathematical expressions and not in CamelCase. Simply put, the spellchecking is too restrictive.

.github/workflows/config/spellcheck_config.yml
schweitzpgi: The regular expression is not trying to merely allow `\p` and `\param`. It is allowing any of the multitude of valid doxygen commands as can be found [in the documentation](https://www.doxygen.nl/manual/commands.html). The spellchecking script is too restrictive and rejects the use of valid doxygen which discourages writing documentation.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: Maybe. I can try it. There are still cases where we may traverse C++ code that we don't really process intentionally. Consider that we're visiting post-order, so it is possible that we visit subtrees that then are elided when the parent is visited and recognized as a something we want to translate to Quake or CC more directly.

.github/workflows/config/spellcheck_config.yml
bmhowe23: > The problem is that they can be mathematical expressions and not in CamelCase. Simply put, the spellchecking is too restrictive.

Are you saying that `\a`, `\e`, `\em`, `\b` are being used for mathematical expressions? That would be a bit surprising to me, but maybe I'm missing something.

FWIW, it might be worth taking a look at #1060, as that new PR makes all of current files with similar (but slightly different) modifications than are in this PR. It seems like eliminating all checks for things that we put in bold are italics might be too permissive and lead to missed spelling errors.

schweitzpgi: `\a` is short for argument, so I think it better allow any legal C++ identifier word after it. Spell checking that is overly conservative.

I would use `\e` to write simple references to variables in formula if I could. Since it doesn't work, I don't write that documentation.

I think the small pinhole that this makes in the spell checking is slightly better than enforcing a Draconian spellcheck that doesn't let me write doxygen documentation in a natural way.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: It doesn't look like error messages is the way to go here.

amccaskey (approved):
LGTM

PR #1056, opened 2023-12-21
Example prototyping the Quantum-enhanced MCMC algorithm

poojarao8:
This example implements the QMCMC algorithm from this paper: https://arxiv.org/pdf/2203.12497.pdf
I have also updated the documentation.


docs/sphinx/examples/python/qmcmc.py
```diff
@@ -0,0 +1,184 @@
...
+    # Specify couplings J's and fields h's.
+    J = np.ones((nqubits,nqubits))
+    h = np.ones(nqubits)
+
+    # Generate a random initial bitstring instead of a spin config
+    # Convert it to its equivalent spin config later when needed
+    s = generate_random_bitstring(nqubits)
+
+    # Create a circuit and allocate qubits
+    kernel = cudaq.make_kernel()
```
amccaskey, line 149: Is the kernel supposed to grow / build on itself each iteration? I notice in the appendix of the paper you are just interested in computing `|psi> = exp(-iHt)|s>` each iteration, so should this `make_kernel()` and subsequent `qalloc` call be in the loop body right before `initial_state`?

docs/sphinx/examples/python/qmcmc.py
```diff
@@ -0,0 +1,184 @@
...
+        for i in range(term.get_qubit_count(), nqubits):
+            adjTerm = term * spin.i(i)
+        ordered_H.append(adjTerm)
+
+    return ordered_H
+
+# This is the first-order Trotter gate decomposition
+def trotter_circuit(kernel, qreg, hk, dt, n_qubits):
+    for term in hk:
+        qbitSupport = []
```
amccaskey, line 87: With a simple additional binding in `py_kernel_builder.cpp` we could replace this loop body with 
```python
for term in hk:
    kernel.exp_pauli(2. * dt * term.get_coefficient().real, qreg, term)
return kernel
```
(You could also just use the existing binding with your string split `pauliString` line) 
```cpp
.def(
          "exp_pauli",
          [](kernel_builder<> &self, py::object theta, const QuakeValue &qubits,
             spin_op &pauli) {
            if (pauli.num_terms() != 1)
              throw std::runtime_error(
                  "exp_pauli expects a SpinOperator with a single term.");

            if (py::isinstance<py::float_>(theta))
              self.exp_pauli(theta.cast<double>(), qubits,
                             pauli.to_string(/*printCoeff*/ false));
            else if (py::isinstance<QuakeValue>(theta))
              self.exp_pauli(theta.cast<QuakeValue &>(), qubits,
                             pauli.to_string(/*printCoeff*/ false));
            else
              throw std::runtime_error(
                  "Invalid `theta` argument type. Must be a "
                  "`float` or a `QuakeValue`.");
          },
          "Apply a general Pauli tensor product rotation, `exp(i theta P)`, on "
          "the specified qubit register. The Pauli tensor product is provided "
          "as a `SpinOperator` with a single term, e.g. `XXYX` for a 4-qubit "
          "term. The angle parameter "
          "can be provided as a concrete float or a `QuakeValue`. The "
          "coefficient of the `SpinOperator` is ignored (user must multiply "
          "the coefficient with the `theta` value explicitly).")
```

docs/sphinx/examples/python/qmcmc.py
poojarao8: Yes, that can be done. The way I am doing things now is by resetting the kernel at the end of each iteration. I am not sure which way is more performant.

amccaskey: That reset method on kernel_builder is for resetting the qubit's state.

docs/sphinx/examples/python/qmcmc.py
poojarao8: Okay, sounds good - thanks!

poojarao8:
Spell check is failing due to `QMCMC` and some variable names.  Is this something I can override or can this spell check failing be ignored?

bmhowe23:
> Spell check is failing due to `QMCMC` and some variable names. Is this something I can override or can this spell check failing be ignored?

See https://gist.github.com/bmhowe23/f855bdb06c083dfe07f901593c4a26b2 for suggested fixes for the spell check issues. If you want to accept those changes, you can apply the suggested patch with the command provided in the comment for the gist. If you don't want to accept those changes, you can simply use them as a reference point.

docs/sphinx/examples/python/qmcmc.py
```diff
@@ -0,0 +1,174 @@
...
+                        help='Temperature')
+    parser.add_argument('--shots_count',
+                        type=int,
+                        default=10,
+                        help='Number of shots')
+    args = parser.parse_args()
+
+    num_iterations = args.num_iterations
+    nqubits = args.nqubits
+    T = args.temperature
```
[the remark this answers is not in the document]
poojarao8: Sure thing, I can update those names. Thanks.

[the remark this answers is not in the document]
poojarao8: @sacpis Do you want me to change the variable names?

PR #1062, opened 2023-12-29
Enable vector<const char *> as kernel input argument

amccaskey:
Enable one to inject a vector of strings for use in `exp_pauli`
```cpp
__qpu__ void test(double theta, std::vector<const char *> paulis) {
  cudaq::qvector q(2);
  for (auto &p : paulis)
    exp_pauli(theta, q, p);
}
```
or 
```cpp
auto [kernel, paulis] = cudaq::make_kernel<std::vector<const char *>>();
auto qubits = kernel.qalloc(2);
kernel.for_loop(0, paulis.size(), [&](auto idx) {
   kernel.exp_pauli(M_PI_2, qubits, paulis[idx]);
});
```
or 
```python
kernel, thetas, paulisArg = cudaq.make_kernel(list[float], list[str])
qubits = kernel.qalloc(2)
kernel.x(qubits[0])
kernel.for_loop(
     0, paulisArg.size(),
     lambda idx: kernel.exp_pauli(thetas[idx], qubits, paulisArg[idx]))
```


schweitzpgi:
Hmm. Wouldn't we prefer to use `std:;string` here (since pointers between different address spaces don't have semantics)?

amccaskey:
> Hmm. Wouldn't we prefer to use `std::string` here (since pointers between different address spaces don't have semantics)?

Probably. I had originally introduced the `std::string` parsing in the bridge with the original `exp_pauli` PR, but we pulled back and just used `ptr<i8>` since `string` is a much heavier lift to support in the bridge. That's why I just fell back to `ptr<i8>` here. I imagine in the future we could have runtime support that does a deep copy of the pointed-to data for transfer to a separate memory space. But I'm open to other ideas and maybe I'm thinking about it wrong.

schweitzpgi:
> > Hmm. Wouldn't we prefer to use `std::string` here (since pointers between different address spaces don't have semantics)?
> 
> Probably. I had originally introduced the `std::string` parsing in the bridge with the original `exp_pauli` PR, but we pulled back and just used `ptr<i8>` since `string` is a much heavier lift to support in the bridge. That's why I just fell back to `ptr<i8>` here. I imagine in the future we could have runtime support that does a deep copy of the pointed-to data for transfer to a separate memory space. But I'm open to other ideas and maybe I'm thinking about it wrong.

Yes, that's what I was remembering too. Adding `string` support will be some additional work, but I think that using `string` at the CUDA Quantum level should be the ultimate goal. We can incrementally get there, of course.

amccaskey:
@bettinaheim maybe instead of `const char *` it would be better to introduce a new type, `cudaq::pauli_word`. This would keep the input limited instead of just falling back on all `char` pointers. What do you think?

PR #1069, opened 2024-01-04
Fix issue 1064: bug in bridge when lambda argument contains a loop.

schweitzpgi:
This is a workaround added to the `for` loop handling. A more general solution in the bridge should be implemented.


1tnguyen (approved):
LGTM 👍

PR #1012, opened 2023-12-07
REST Server-Client platform  

1tnguyen:
### Description

Platform (mqpu capable) based on server-client REST communication:

- Serialize/deserialize the IR and `ExecutionContext`.

- Multiple server instances can be created (multi-QPU).

- Server instance is MPI compatible (e.g., running a MPI simulator backend).

- Auto-launch server instances on randomly selected TCP/IP ports.

PR #1070, opened 2024-01-04
Supporting changes for PR#1012

1tnguyen:
### Description

- Const-correctness in MeasureCounts: in case we need to pass a `const MeasureCounts&` (e.g., Json conversion)

- Add string broadcast to MPI interface.

- Minor tidy up some somments/include.

- nvq++ in library mode should also include `OBJS_TO_MERGE` when doing the linking (compiled `backendConfig.cpp`)

### DIFF
--- a/runtime/common/MeasureCounts.cpp
+++ b/runtime/common/MeasureCounts.cpp
@@ -319,10 +319,9 @@ double sample_result::probability(std::string_view bitStr,
     return 0.0;
 
   const auto countIter = iter->second.counts.find(bitStr.data());
-  if (countIter == iter->second.counts.end())
-    return 0.0;
-  else
-    return (double)countIter->second / totalShots;
+  return (countIter == iter->second.counts.end())
+             ? 0.0
+             : (double)countIter->second / totalShots;
 }
 
 std::size_t sample_result::count(std::string_view bitStr,
--- a/runtime/common/MeasureCounts.cpp
+++ b/runtime/common/MeasureCounts.cpp
@@ -43,7 +43,7 @@ ExecutionResult::ExecutionResult(const ExecutionResult &other)
     : counts(other.counts), expectationValue(other.expectationValue),
       registerName(other.registerName), sequentialData(other.sequentialData) {}
 
-ExecutionResult &ExecutionResult::operator=(ExecutionResult &other) {
+ExecutionResult &ExecutionResult::operator=(const ExecutionResult &other) {
   counts = other.counts;
   expectationValue = other.expectationValue;
   registerName = other.registerName;
@@ -69,7 +69,7 @@ bool ExecutionResult::operator==(const ExecutionResult &result) const {
 // represent register name, number of bitstrings M, then for each bit string
 // {l, bs.length, count}
 /// @return
-std::vector<std::size_t> ExecutionResult::serialize() {
+std::vector<std::size_t> ExecutionResult::serialize() const {
   std::vector<std::size_t> retData;
 
   // Encode the classical register name
@@ -116,7 +116,7 @@ void ExecutionResult::deserialize(std::vector<std::size_t> &data) {
   }
 }
 
-std::vector<std::size_t> sample_result::serialize() {
+std::vector<std::size_t> sample_result::serialize() const {
   std::vector<std::size_t> retData;
   for (auto &result : sampleResults) {
     auto serialized = result.second.serialize();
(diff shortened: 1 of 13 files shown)


1tnguyen:
Note: I'm breaking up #1012 into smaller chunks to ease the review process.

runtime/cudaq/utils/registry.h
```diff
@@ -7,8 +7,7 @@
  ******************************************************************************/
 
 #pragma once
-#define LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING 1
-#include "llvm/Support/Registry.h"
```
schweitzpgi, line 11: Are you sure we should remove this? LLVM will crash if there is an ABI version mismatch.

schweitzpgi (commented):
+1

runtime/cudaq/utils/registry.h
1tnguyen: We don't use `llvm/Support/Registry.h` in this header. The other place (`runtime/common/Registry.h`) where we actually use `llvm::Registry` already has that `#define`.

We can compile without this `#include` here, so I guess it should be safe.

runtime/cudaq/platform/QuantumExecutionQueue.h
```diff
@@ -38,8 +38,11 @@ class QuantumExecutionQueue {
   /// Enqueue a Sampling task.
   void enqueue(QuantumTask &task);
 
+  /// Get id of the thread this queue executes on.
+  std::thread::id getExecutionThreadId() const;
+
 protected:
-  /// The mutex, used for locking when adding to the queue
+  /// The `mutex`, used for locking when adding to the queue
```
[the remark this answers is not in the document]
1tnguyen: Apparently, `//!` didn't work. I've reverted the change and added "mutex" to the dictionary.

runtime/common/MeasureCounts.cpp
```diff
@@ -312,12 +313,16 @@ std::size_t sample_result::size(const std::string_view registerName) noexcept {
...
+                                  const std::string_view registerName) const {
   auto iter = sampleResults.find(registerName.data());
   if (iter == sampleResults.end())
     return 0.0;
 
-  return (double)iter->second.counts[bitStr.data()] / totalShots;
+  const auto countIter = iter->second.counts.find(bitStr.data());
+  if (countIter == iter->second.counts.end())
+    return 0.0;
+  else
```
schweitzpgi, line 324: nit: else after return

runtime/cudaq/cudaq.cpp
```diff
@@ -185,6 +185,11 @@ void broadcast(std::vector<double> &data, int rootRank) {
   commPlugin->broadcast(data, rootRank);
 }
 
+void broadcast(std::string &data, int rootRank) {
```
schweitzpgi, line 188:
```suggestion
void broadcast(const std::string &data, int rootRank) {
```
?

runtime/cudaq/utils/registry.h
amccaskey: This is old - from when the registry code was in utils and not in common. Safe to remove

runtime/cudaq/distributed/mpi_plugin.cpp
```diff
@@ -92,6 +92,17 @@ void MPIPlugin::broadcast(std::vector<double> &data, int rootRank) {
       m_comm, data.data(), data.size(), FLOAT_64, rootRank));
 }
 
+void MPIPlugin::broadcast(std::string &data, int rootRank) {
+  std::int32_t strLen = data.size();
+  HANDLE_MPI_ERROR(
+      m_distributedInterface->Bcast(m_comm, &strLen, 1, INT_32, rootRank));
+
+  if (rank() != rootRank)
+    data.resize(strLen);
```
schweitzpgi, line 101: This seems redundant, but maybe the call to `Bcast` is changing the size of the string object... :shrug:

amccaskey (approved):
nice fixes / clean up. +1

runtime/common/MeasureCounts.cpp
1tnguyen: Good catch. I've updated the code to use a ternary operator instead.

runtime/cudaq/cudaq.cpp
1tnguyen: The semantics of these `broadcast` APIs is that other ranks (not `rootRank`) will have its `data` updated; hence cannot be `const`.

runtime/cudaq/distributed/mpi_plugin.cpp
1tnguyen: Indeed, we need to allocate these string buffers before using MPI API to populate them.
Generally, only `rootRank` knows the size of the string, hence we need to do a two-pass `Bcast` as above.

runtime/cudaq/cudaq.cpp
schweitzpgi: +1

PR #1071, opened 2024-01-05
Enable use of single-term spin_op with kernel_builder::exp_pauli

amccaskey:
Currently, the Pauli word argument for `kernel_builder::exp_pauli` must be a string. Allow one to pass a single-term `spin_op`. 

Useful for use cases like in #1056

### DIFF
--- a/runtime/cudaq/builder/kernel_builder.h
+++ b/runtime/cudaq/builder/kernel_builder.h
@@ -373,7 +373,7 @@ private:
   /// @brief The CUDA Quantum Quake function arguments stored as `QuakeValue`s.
   std::vector<QuakeValue> arguments;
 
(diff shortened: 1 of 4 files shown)


runtime/cudaq/builder/kernel_builder.h
```diff
@@ -373,6 +373,19 @@ class kernel_builder : public details::kernel_builder_base {
   /// @brief The CUDA Quantum Quake function arguments stored as `QuakeValue`s.
   std::vector<QuakeValue> arguments;
 
+  /// @brief Return a string representation of a the given spin operator.
```
1tnguyen, line 376:
```suggestion
  /// @brief Return a string representation of the given spin operator.
```

1tnguyen (approved):
LGTM 👍

PR #1072, opened 2024-01-05
Bot cleanup.

schweitzpgi:
Remove some redundant/unneeded includes.

Cleanup doxygen comments to standardize, format, correct punctuation, etc.

Make the function with `__nvqpp__` prefix an extern "C" function. Again to standardize as per the rest of the code.


amccaskey (approved):
LGTM

PR #1077, opened 2024-01-09
Unify python stack

khalatepradnya:
### Description

This PR attempts to get a working docker image and Python wheels file for the experimental Python features.

- Some tests are skipped and annotated with `SKIP_TEST` comment; fixes for which will be handled in separate PR.
- New feature of remote simulation is not (yet) available in Python
- Python wheel skipped for arm64 platform


runtime/cudaq/platform/default/rest/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@
 set(INTERFACE_POSITION_INDEPENDENT_CODE ON)
 
 message(STATUS "Building REST QPU.")
-add_library(cudaq-rest-qpu SHARED RemoteRESTQPU.cpp 
+add_library(cudaq-rest-qpu SHARED PyRemoteRESTQPU.cpp
```
amccaskey, line 12: Might be that I'm missing some new addition, but as I understand it we do need both the `PyRemoteRESTQPU` and the `RemoteRESTQPU` (with common code extracted out into the `common/BaseRemoteRESTQPU`). The reason for the distinction is to avoid static initializers from LLVM being called twice (Python bindings now link statically to LLVM/MLIR, while `libcudaq-rest-qpu` links to `libcudaq-runtime-mlir` which links statically to LLVM/MLIR) . The `RemoteRESTQPU` remains for our existing runtime `kernel_builder` and AST Bridge workflows, but the new `PyRemoteRESTQPU` is needed for any workflows that target remote REST targets

python/tests/unittests/test_kernel_builder.py
```diff
@@ -712,6 +719,7 @@ def test_cry_control_list():
     assert result["1110"] == 1000
 
 
+@pytest.mark.skip(reason="AssertionError on 'assert isinstance(arg, _cext.ir.Value)' from 'kernel.crz(value, [q1, q2, q3], target)'")
```
amccaskey, line 722: Seems like there is an issue in `kernel_builder.py` `__singleTargetSingleParameterControlOperation`

runtime/cudaq/platform/default/rest/CMakeLists.txt
khalatepradnya: Thanks for the early feedback. Some restructuring and clean up still remains, but, I was planning to have the functionality in single place and remove the 'old/legacy' kernel builder workflow, replace it with the new builder workflow - so for the end users, the interface remains the same (but, the plumbing underneath would be different). Does that seem reasonable? Or am I missing something?

python/tests/unittests/test_kernel_builder.py
khalatepradnya: Yeah, once this can run in CI, I will proceed to debug all the tests which have marked as 'skipped'.

runtime/cudaq/platform/default/rest/CMakeLists.txt
amccaskey: The problem I see here is that `libcudaq-rest-qpu` links directly to LLVM/MLIR static libraries. So if you rely on the `RemoteRESTQPU` functionality as part of this library in the Python language bindings, then you will need to link to it (previously we've dynamically loaded it in `LinkedLibraryHolder`). Unfortuntately though, the Python bindings as a library already link statically to LLVM/MLIR, so when this dynamic load happens you will get an error with LLVM static initializers being called more than once. So for any `RemoteRESTQPU` functionality to be used by Python, it has to be decoupled in some way from directly linking to LLVM/MLIR and instead rely on the symbols already loaded in the Python module library (from its static linking to LLVM/MLIR). 

You do need to retain the runtime library's `RemoteRESTQPU` functionality for C++ MLIR-related workflows. The existing solution for this decoupling is the creation of `common/RuntimeMLIRCommonImpl.h` and `common/BaseRemoteRESTQPU.h`. 

I may be wrong and there is some mechanism to enable this unification, but as I see it, we will have to retain `PyRemoteRESTQPU` as separate from the core C++ runtime's `RemoteRESTQPU.cpp`.

Happy to hop on a call to talk in more depth than possible on the chat boxes here

khalatepradnya:
Unable to get CI to work due to `429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit.`.

khalatepradnya:
To be merged after https://github.com/NVIDIA/cuda-quantum/pull/1100

python/CMakeLists.txt
```diff
@@ -1,27 +1,163 @@
...
+    )
+
+################################################################################
+# Common CAPI
+################################################################################
+
+add_mlir_python_common_capi_library(CUDAQuantumPythonCAPI
+  INSTALL_COMPONENT CUDAQuantumPythonModules
+  INSTALL_DESTINATION mlir_cudaq/_mlir_libs
+  OUTPUT_DIRECTORY "${MLIR_BINARY_DIR}/mlir_cudaq/_mlir_libs"
```
amccaskey, line 110: I notice that the python bindings here are left in the top-level of the developer's `build/` directory. One thing I've found useful as a developer is to have the python bindings dumped in a stand-alone folder within the build directory, like `build/python/` instead of `build/`. There are few reasons for this - it better emulates the python bindings as a stand-alone package (easy to point PYTHONPATH too), but also for developers switching between this branch and `main` care has to be taken to delete the python bindings in the top-level of `build` before using those in`build/python/`, otherwise you get conflicts. If we stick with the `build/python` for both, then the bindings from one branch are replaced with the bindings for the other branch when you rebuild.

khalatepradnya: will do

docs/sphinx/api/languages/python_api.rst
```diff
@@ -69,19 +72,20 @@ Backend Configuration
...
 .. automethod:: cudaq::num_available_gpus
 .. automethod:: cudaq::set_random_seed
 
 Data Types
 =============================
 
-.. autoclass:: cudaq::Target
-    :members:
+.. [SKIP_TEST]: Reason - AttributeError: module 'cudaq' has no attribute 'Target'
+.. .. autoclass:: cudaq::Target
```
amccaskey, line 83: Will need to add `Target = cudaq_runtime.Target` to `python/cudaq/__init__.py`

docs/sphinx/api/languages/python_api.rst
```diff
@@ -119,24 +123,28 @@ Data Types
 .. autofunction:: cudaq::spin.y
 .. autofunction:: cudaq::spin.z
 
-.. autoclass:: cudaq::SampleResult
-    :members:
-    :special-members: __getitem__, __len__, __iter__
+.. [SKIP_TEST]: Reason - AttributeError: module 'cudaq' has no attribute 'SampleResult'. Did you mean: 'AsyncSampleResult'?
```
amccaskey, line 126: `SampleResult = cudaq_runtime.SampleResult` required in`__init__.py`

docs/sphinx/api/languages/python_api.rst
```diff
@@ -119,24 +123,28 @@ Data Types
...
+.. .. autoclass:: cudaq::SampleResult
+..     :members:
+..     :special-members: __getitem__, __len__, __iter__
 
 .. autoclass:: cudaq::AsyncSampleResult
     :members:
 
-.. autoclass:: cudaq::ObserveResult
-    :members:
+.. [SKIP_TEST]: Reason - AttributeError: module 'cudaq' has no attribute 'ObserveResult'. Did you mean: 'AsyncObserveResult'?
```
amccaskey, line 134: `ObserveResult = cudaq_runtime.ObserveResult` in `__init__.py`

python/CMakeLists.txt
```diff
@@ -1,27 +1,163 @@
...
+set(_origin_prefix "\$ORIGIN")
+if(APPLE)
+  set(_origin_prefix "@loader_path")
+endif()
+set_property(TARGET CUDAQuantumPythonModules.extension._quakeDialects.dso 
+             APPEND PROPERTY INSTALL_RPATH "${_origin_prefix}:${_origin_prefix}/../../lib/")
+
+
+## TODO: Clean-up following from 'old' CMakeLists.txt
+## FIXME: Delete '_pycudaq' target?
```
amccaskey, line 151: Yea I think this can all go away. Also there are some old c++ files that aren't going to be used any more, like py_sample.cpp

python/tests/domains/test_qnn.py
```diff
@@ -54,6 +54,8 @@ def test_simpleObserveN_QNN():
     assert np.isclose(data[-1], 0.33752631)
 
 
+## [SKIP_TEST]
+@pytest.mark.skip(reason="RuntimeError: Could not pack argument...")
```
amccaskey, line 58: Issue here is that `py_observe_async.cpp` is not running something like `args = validateInputArguments(args)` like the current bindings do. This function will find any `np.ndarray` with 1D shape and call `tolist()` on it to make it compatible with our MLIR code. Here's the diff for the fix 
```diff
diff --git a/python/runtime/cudaq/algorithms/py_observe_async.cpp b/python/runtime/cudaq/algorithms/py_observe_async.cpp
index 8b211eac2..f81bf24ff 100644
--- a/python/runtime/cudaq/algorithms/py_observe_async.cpp
+++ b/python/runtime/cudaq/algorithms/py_observe_async.cpp
@@ -34,7 +34,8 @@ void bindObserveAsync(py::module &mod) {
 
         auto &platform = cudaq::get_platform();
         auto *argData = new cudaq::OpaqueArguments();
-        cudaq::packArgs(*argData, args);
+        args = simplifiedValidateInputArguments(args);
+        cudaq::packArgs(*argData, args, [](OpaqueArguments&, py::object&) {return false;});
         auto kernelName = kernel.attr("name").cast<std::string>();
         auto kernelMod = kernel.attr("module").cast<MlirModule>();
 
diff --git a/python/utils/OpaqueArguments.h b/python/utils/OpaqueArguments.h
index e1d5e7ac8..f8af6fade 100644
--- a/python/utils/OpaqueArguments.h
+++ b/python/utils/OpaqueArguments.h
@@ -123,6 +123,38 @@ inline py::args validateInputArguments(kernel_builder<> &kernel,
   return processed;
 }
 
+/// @brief FIXME. This is a simple version of the above function, 
+// it will only process numpy arrays of 1D shape and make them compatible 
+// with our MLIR code. Future work should make this function perform more checks, 
+// we probably want to take the Kernel MLIR argument Types as input and use that 
+// to validate that the passed arguments are good to go. 
+inline py::args simplifiedValidateInputArguments(py::args &args) {
+  py::args processed = py::tuple(args.size());
+  for (std::size_t i = 0; i < args.size(); ++i) {
+    auto arg = args[i];
+    // Check if it has tolist, so it might be a 1d buffer (array / numpy
+    // ndarray)
+    if (py::hasattr(args[i], "tolist")) {
+      // This is a valid ndarray if it has tolist and shape
+      if (!py::hasattr(args[i], "shape"))
+        throw std::runtime_error(
+            "Invalid input argument type, could not get shape of array.");
+
+      // This is an ndarray with tolist() and shape attributes
+      // get the shape and check its size
+      auto shape = args[i].attr("shape").cast<py::tuple>();
+      if (shape.size() != 1)
+        throw std::runtime_error("Cannot pass ndarray with shape != (N,).");
+
+      arg = args[i].attr("tolist")();
+    }
+
+    processed[i] = arg;
+  }
+
+  return processed;
+}
+
 /// @brief For general function broadcasting over many argument
 /// sets, this function will create those argument sets from
 /// the input `args`.
```

python/tests/unittests/test_observe.py
```diff
@@ -282,9 +282,10 @@ def test_observe_multi_param(angle_0, angle_1, angles, want_state,
     with pytest.raises(RuntimeError) as error:
         # Too few.
         cudaq.observe(kernel, hamiltonian, np.pi, np.pi)
-    with pytest.raises(RuntimeError) as error:
-        # Too many list elements.
-        cudaq.observe(kernel, hamiltonian, np.pi, np.pi, [np.pi, np.pi, np.pi])
+    ## [SKIP_TEST]: Following check fails with "Failed: DID NOT RAISE <class 'RuntimeError'>"
```
amccaskey, line 285: This may be a bad test - it is difficult to know from just looking at the MLIR exactly how many unique vector extractions there are, and then correlated that back to the number of elements the user provided.

python/tests/domains/test_qnn.py
khalatepradnya: Fixed in 6225c74

python/tests/unittests/test_kernel_builder.py
```diff
@@ -916,17 +934,20 @@ def test_exp_pauli():
...
     assert np.isclose(want_exp, -1.13, atol=1e-2)
 
-    kernel, theta = cudaq.make_kernel(float)
-    qubits = kernel.qalloc(4)
-    kernel.x(qubits[0])
-    kernel.x(qubits[1])
-    kernel.exp_pauli(theta, qubits, cudaq.SpinOperator.from_word('XXXY'))
-    want_exp = cudaq.observe(kernel, h, .11).expectation()
-    assert np.isclose(want_exp, -1.13, atol=1e-2)
+    ## [SKIP_TEST]: Reason - AssertionError on 'exp_pauli'
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in fd0a8ff

python/tests/unittests/test_observe.py
khalatepradnya: Removed in 99d9bee

python/tests/unittests/test_observe.py
```diff
@@ -685,7 +682,8 @@ def test_observe_list():
     want_expectation_value = -1.7487948611472093
     assert assert_close(want_expectation_value, sum, tolerance=1e-2)
 
-
+## [SKIP_TEST]
+@pytest.mark.skip(reason="TypeError: setSpinOperator(): incompatible function arguments on 'results = cudaq.observe(circuit, hamiltonianList, angles)'")
```
amccaskey, line 686: Fix 
```diff
diff --git a/python/cudaq/runtime/observe.py b/python/cudaq/runtime/observe.py
index 70e5e4ae5..67249a928 100644
--- a/python/cudaq/runtime/observe.py
+++ b/python/cudaq/runtime/observe.py
@@ -67,22 +67,30 @@ Returns:
     if noise_model != None:
         cudaq_runtime.set_noise(noise_model)
 
+    # Process spin_operator if its a list
+    localOp = spin_operator
+    localOp = cudaq_runtime.SpinOperator()
+    if isinstance(spin_operator, list):
+        for o in spin_operator:
+            localOp += o
+        localOp -= cudaq_runtime.SpinOperator()
+    else:
+        localOp = spin_operator
+
     results = None
     if __isBroadcast(kernel, *args):
         results = __broadcastObserve(kernel,
-                                     spin_operator,
+                                     localOp,
                                      *args,
                                      shots_count=shots_count)
-    else:
-        localOp = spin_operator
-        localOp = cudaq_runtime.SpinOperator()
-        if isinstance(spin_operator, list):
-            for o in spin_operator:
-                localOp += o
-            localOp -= cudaq_runtime.SpinOperator()
-        else:
-            localOp = spin_operator
 
+        if isinstance(spin_operator, list):
+            results = [[
+                cudaq_runtime.ObserveResult(p.expectation(o), o, p.counts(o))
+                for o in spin_operator
+            ]
+                       for p in results]
+    else:
         ctx = cudaq_runtime.ExecutionContext('observe', shots_count)
         ctx.setSpinOperator(localOp)
         cudaq_runtime.setExecutionContext(ctx)
```

python/tests/unittests/test_quake_value.py
```diff
@@ -80,6 +80,8 @@ def test_QuakeValueLifetimeAndPrint():
     assert s.count('quake.x') == 2
 
 
+## [SKIP_TEST]
+@pytest.mark.skip(reason="TypeError: unsupported operand type(s) for /: 'float' and 'QuakeValue' on 'kernel3.rx(np.pi / factor[0], qubit3[0])'")
```
amccaskey, line 84: Fix 
```diff
diff --git a/python/cudaq/kernel/quake_value.py b/python/cudaq/kernel/quake_value.py
index 85c6a7cb6..7b5418108 100644
--- a/python/cudaq/kernel/quake_value.py
+++ b/python/cudaq/kernel/quake_value.py
@@ -138,6 +138,16 @@ class QuakeValue(object):
             return QuakeValue(
                 getattr(arith, opStr)(thisVal, otherVal).result, self.pyKernel)
 
+    def __rtruediv__(self, other):
+        with self.ctx, Location.unknown(), self.pyKernel.insertPoint:
+            thisVal, otherVal, opStr = self.__checkTypesAndCreateQuakeValue(
+                other, 'Div')
+            if opStr == 'DivIOp':
+                opStr = 'DivSIOp'
+
+            return QuakeValue(
+                getattr(arith, opStr)(otherVal, thisVal).result, self.pyKernel)
+    
     def __add__(self, other):
         with self.ctx, Location.unknown(), self.pyKernel.insertPoint:
             thisVal, otherVal, opStr = self.__checkTypesAndCreateQuakeValue(
```

python/tests/unittests/test_observe.py
khalatepradnya: Patched in [d369a0f](https://github.com/NVIDIA/cuda-quantum/pull/1077/commits/d369a0f1b9ffe8635c714f96a151c4695f7a4e44)

python/tests/unittests/test_quake_value.py
khalatepradnya: Patched in [3deecbb](https://github.com/NVIDIA/cuda-quantum/pull/1077/commits/3deecbb637615bfbef6e30b60e182f920b49ad20)

python/CMakeLists.txt
khalatepradnya: Added to https://github.com/NVIDIA/cuda-quantum/issues/1157

amccaskey:
We have FileCheck tests in both the `compiler` and `mlir` folder. We should probably merge these two together under one ctest target and in one folder.

khalatepradnya:
> We have FileCheck tests in both the `compiler` and `mlir` folder. We should probably merge these two together under one ctest target and in one folder.

Sure, I had them merged into 'mlir' but, reverted to keep existing functionality v/s new functionality tests - perhaps not necessary. I will club into 'compiler' directory.

PR #1078, opened 2024-01-09
Improve thread-safety of CUDAQ static variables

1tnguyen:
### Description

There are a couple of static variables in `cudaq.cpp` that might be accessed concurrently.

For example:

- Thread 1 calls `deviceCodeHolderAdd` (`.thunk` function) => `vector::emplace_back`

- Thread 2 calls `get_quake_by_name` => iterate over the same `vector` => BANG.

This PR adds a `mutex` to guard access to those vars.


amccaskey (approved):
+1

bmhowe23 (approved):
Nice find👍

PR #1080, opened 2024-01-10
Part 1: Start adding support for arguments that are compositions

schweitzpgi:
of vectors and structures.

Add test for struct of struct members, struct with vector members and vector of struct type arguments.


schweitzpgi:
ping for review

PR #1081, opened 2024-01-10
Follow-up to #1059 - make space handling more robust

bmhowe23:


schweitzpgi (commented):
Trying all these change together now...

schweitzpgi (approved):
Works for me.

Issue #1003, opened 2023-12-05
[RFC] Internal qubit ordering


boschmitt:
### Background and motivation

The field of quantum computing is more and less split in half when it comes to the choice of qubit ordering convention. These conventions are: 

1. $|q_0q_1\dots q_n\rangle$
2. $|q_n\dots q_1q_0\rangle$

where  $n$ is the number of qubits. In both conventions $q_0$ is considered the least significant qubit, i.e.: the state $|001\rangle$ is interpreted as four by (1) and as one by (2).

In CUDA Quantum, we are currently using (1) both internally and externally (i.e., user facing API where this matters expects users to use convention (1)). However, I propose we internally use (2) to keep consistency with `cuQuantum` default convention.

### Internal implications

Internally the convention will be (2). This will affect:

- Matrix definition of multi-qubit operations.
- Synthesis algorithms, e.g., `cudaq::from_state`.
- Interface with simulators.

### External implications

In user facing API we should allow the user to define which convention will be used.

schweitzpgi:
On the internal aspect at the IR level, qubits are grouped as vectors. The notation is consistent from C++ source code through to the IR. Specifically a vector of qubits, `v` is indexed such that `v[0]` refers to the first qubit in the vector. As in the C++ language, this means `v[0]` is identical to `v.front()`. Redefining this such that `v[0]` become `v.back()` would, I fear, be very confusing to the user.

This extends to any operations in the IR that construct or allow mixing of qubits. Again, consistent with the C++ surface language concatenation of qubits creates a left-to-right mapping of indices. For example, a vector, `v`, concatenated with another vector, `w`, as in `u = v + w` logically implies `u[0]` is `v.front()` and `u.back()` is `w.back()`. Again, reversing this ordering would be counterintuitive and violate the principle of least surprise.

 Hopefully, this RFC does not mean to change these semantics.

PR #1082, opened 2024-01-10
Change internal qubit ordering

boschmitt:
### Description

This PR implements #1003


bmhowe23:
Are these changes consistent with `runtime/nvqir/qpp/README.md`, or will that file need to be updated?

boschmitt:
> Are these changes consistent with `runtime/nvqir/qpp/README.md`, or will that file need to be updated?

Ultimately, I think we should remove this file. The idea of this PR is to not break current users, so from their perspective what is described in it is still valid. Internally, however, it we will be using the alternative ordering described in the file.

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -25,62 +25,40 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
   StateType state;
 
   /// Convert from little endian to big endian.
-  std::size_t bigEndian(const int n_qubits, const int bit) {
-    return n_qubits - bit - 1;
+  std::size_t bigEndian(const int bit) {
+    return std::log2(stateDimension) - bit - 1;
```
bmhowe23, line 29: While working on #668, I encountered the fact that `std::log2(0)` produces different results on Mac vs Linux. It would be best to specifically check for that case and return a defined value in that case (even if we don't expect it to occur right now) as it may help save debugging in the future.

runtime/cudaq/algorithms/state.cpp
```diff
@@ -39,7 +39,12 @@ std::complex<double> state::operator[](std::size_t idx) {
   if (shape.size() != 1)
     throw std::runtime_error("Cannot request 1-d index into density matrix. "
                              "Must be a state vector.");
-  return stateData[idx];
+  std::size_t numQubits = std::log2(stateData.size());
+  unsigned int newIdx = 0;
+  for (auto i = 0u; i < numQubits; ++i)
+    if (idx & (1 << i))
+      newIdx |= (1u << ((numQubits - 1) - i));
```
bmhowe23, line 46: I believe `newIdx` and `i` should be declared as `std::size_t` in order to support >32 qubits.

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -232,45 +207,52 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
 
   /// @brief Reset the qubit
   /// @param qubitIdx
-  void resetQubit(const std::size_t qubitIdx) override {
+  void resetQubit(const std::size_t index) override {
```
bmhowe23, line 210: Recommend updating `@param` with new parameter name and providing a valid description of the parameter (i.e. `@param index 0-based index of qubit to reset`).

bmhowe23 (commented):
I know this is still draft but I took an early peek.

runtime/nvqir/qpp/QppCircuitSimulator.cpp
boschmitt: As `std::log2` comes from C, `std::log2(0)` should set a floating-point exception flag and a `errno` flag. They are both annoying to check, and, in this case, it wound't make sense to return any value if the dimension is `0` (unless we add the notion of invalid qubit index). I will add an assertion for the final version of this PR.

Note: In reality, I didn't want to retrieve the number of qubits in this way. I thought that `nQubitsAllocated` should do the trick, but it doesn't (this variable has a misleading name, as it doesn't represent the number of allocated qubits, but the number of "live"/"in use" qubits).

bmhowe23:
> > Are these changes consistent with `runtime/nvqir/qpp/README.md`, or will that file need to be updated?
> 
> Ultimately, I think we should remove this file. The idea of this PR is to not break current users, so from their perspective what is described in it is still valid. Internally, however, it we will be using the alternative ordering described in the file.

I would recommend either updating that file to be correct (preferred) or removing it as you suggested as part of this PR.

boschmitt:
> > > Are these changes consistent with `runtime/nvqir/qpp/README.md`, or will that file need to be updated?
> > 
> > 
> > Ultimately, I think we should remove this file. The idea of this PR is to not break current users, so from their perspective what is described in it is still valid. Internally, however, it we will be using the alternative ordering described in the file.
> 
> I would recommend either updating that file to be correct (preferred) or removing it as you suggested as part of this PR.

Thanks for the reminder! I deleted the file as I added the necessary bits a comment on the function that convert qubit indices.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -552,10 +551,8 @@ public:
     // Use custatevec accessor to retrieve the view
     custatevecAccessorDescriptor_t accessor;
     const uint32_t nIndexBits = std::log2(stateDimension);
-    // Note: we use MSB bit ordering when reporting the state vector
-    // hence, bit ordering vector = [N-1, N-2, ..., 0]
     std::vector<int32_t> bitOrdering(nIndexBits);
-    std::iota(std::rbegin(bitOrdering), std::rend(bitOrdering), 0);
+    std::iota(std::begin(bitOrdering), std::end(bitOrdering), 0);
```
amccaskey, line 555: I think Thien mentioned that this implementation used the accessor functions from custatevec because he had issues getting the bit ordering right. Do you think you could try replacing these calls with a simple call to `cudaMemcpy` with type `cudaMemcpyDeviceToHost` and see if you get the correct ordering your are looking for? I think this memcpy would ultimately be optimal.

runtime/cudaq/algorithms/state.cpp
```diff
@@ -39,7 +39,12 @@ std::complex<double> state::operator[](std::size_t idx) {
   if (shape.size() != 1)
     throw std::runtime_error("Cannot request 1-d index into density matrix. "
                              "Must be a state vector.");
-  return stateData[idx];
+  std::size_t numQubits = std::log2(stateData.size());
```
amccaskey, line 42: Would you mind providing an you explain the updates here? Why the need to transform the index?

boschmitt: **Short explanation**: Otherwise this would be a breaking change from the users perspective. This PR changes the qubit ordering for the internal representation of the state. There is a need for further clarification about the `cudaq::state` API.

**Long explanation**: There is an ambiguity on what `operator[](std::size_t i)` of `cudaq::state` should return. To illustrate the problem, suppose we define a two qubit state vector:
```
cudaq::state s = {a, b, c, d}
```
The question here is: What is the value of `s[1]`?

Well, it depends of what `1` represents. If `1` is a index to the underlying representation of the state, say a `std::vector` or `std::array`, then we would return `s[1] == b`. This interpretation, however, leaks internal details about the representation of the state. What if we eventually decide to some other underlying representation, e.g., allow `cudaq::state` to also represent sparse states?

Okay, so interpreting `1` as the index to the underlying representation is not a good idea. So we decide to say that `1` is the quantum state `|1>`. Now qubit ordering comes in to play: the state `|1>` can be represented as either `|10>` or `|01>`, depending on the chosen order. Currently, users seem to expect `|1> == |10>`, meaning that the binary string that represents `|1>` does not correspond to `1` in binary (and thus it also does correspond to the index `1` on the underlying array that hold the amplitudes): 
```
array: [ a,  b,  c,  d ]
         0   1   2   3  -> indices as decimals
        00  01  10  11  -> indices as binary
        |0> |2> |1> |3> -> state as interpreted by the users
```
So, when the users does `s[1]`,  they are asking for `s["|10>"]`, and thus we need to translate `|1>` to the corresponding index into the vector of amplitudes. The high-level understanding of this translations is: we take `1`, translate it to `10`, and interpret that as a binary number, i.e., `2`. Then `s[1] == c`.

amccaskey: The original intent was that the method would act like `s[1] = b`, give the user a view on the vector data. Granted, you are right that this may not apply to other state representations like those from cuTensorNet. If we move to this second interpretation, I would think we should put forward a better method to extract amplitudes (e.g. see discussion on #1107)

amccaskey: I would prefer we keep it as is - there are methods for direct vector and matrix extraction. We should provide the correct updates as part of #1123 to extract amplitudes more generally as discussed in #1107

amccaskey: Talked offline. We are picking an amplitude ordering (following cuQuantum's) which is a good thing - enforces backend simulators to be compliant, helps future compiler passes. But here we are retaining the current amplitude ordering for the user-facing get_state function. It would be preferable for the get_state function to return a state of the same bit ordering as the internal system (cuQuantum's after this PR) but this is a "breaking change". To get this PR across the finish line, we'll retain this status quo, but future work should allow users to specify the bit ordering when calling get_state, and ultimately the default should be the same as the simulation infrastructure.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -549,39 +548,9 @@ public:
...
-                                       /*end*/
-                                       stateDimension));
-    // destroy descriptor
-    HANDLE_ERROR(custatevecAccessorDestroy(accessor));
-    // free extra workspace if allocated
-    if (extraWorkspaceSizeInBytes > 0)
-      HANDLE_CUDA_ERROR(cudaFree(extraWorkspace));
+    cudaMemcpy(tmp.data(), deviceStateVector,
+               stateDimension * sizeof(std::complex<ScalarType>),
+               cudaMemcpyDeviceToHost);
```
1tnguyen, line 553: This change looks good. 

Only a minor suggestion to wrap the CUDA call in `HANDLE_CUDA_ERROR` for better error handling.

```suggestion
    HANDLE_CUDA_ERROR(cudaMemcpy(tmp.data(), deviceStateVector,
               stateDimension * sizeof(std::complex<ScalarType>),
               cudaMemcpyDeviceToHost));
```

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -24,63 +24,51 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
...
-        }
-      }
-      return (count % 2) == 0;
+  double calculateExpectationValue(const std::vector<std::size_t> &qubits) {
+    const auto hasEvenParity = [&qubits](std::size_t x) -> bool {
+      bool evenParity = true;
+      for (auto q : qubits)
+        if (x & (1ULL << q))
+          evenParity = !evenParity;
+      return evenParity;
```
1tnguyen, line 49: Not related to this PR but at some point, we probably need to optimize this: e.g., construct a bitmask based on the `qubits` list and use population count intrinsic, like `evenParity = std::popcount(x & bitmask) % 2 == 0`.

boschmitt: Good point. The problem is that the bit manipulation functions are c++20, and now we also need to support c++17. Thus, for latter we would need to use compiler intrinsics (or this), and I'm not sure if we have a guideline about using those.

amccaskey: We need to support C++17 in the public user-facing API, but the rest of the framework is meant to be developed with C++20. As long as this is internal code, no need to worry.

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -24,63 +25,50 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
...
-        }
-      }
-      return (count % 2) == 0;
+  double calculateExpectationValue(const std::vector<std::size_t> &qubits) {
+    const auto hasEvenParity = [&qubits](std::size_t x) -> bool {
+      std::size_t bitmask = 0;
+      for (auto q : qubits)
+        bitmask |= (1ULL << q);
+      return std::popcount(x & bitmask) % 2 == 0;
     };
```
1tnguyen, line 50: Thanks @boschmitt 👍 for addressing this in this PR. 

We should also move the `bitmask` construction out of the lambda: 
  
```suggestion
    std::size_t bitmask = 0;
    for (auto q : qubits)
      bitmask |= (1ULL << q);
    const auto hasEvenParity = [&bitmask](std::size_t x) -> bool {
      return std::popcount(x & bitmask) % 2 == 0;
    };
```

1tnguyen (approved):
LGTM 👍

PR #1092, opened 2024-01-15
Workflow to have commonly used images on GHCR

bettinaheim:
This PR is an attempt at avoiding the request rate limit for pulling docker images.
Workflow to update images once a day: https://github.com/NVIDIA/cuda-quantum/actions/runs/7534002904


.github/workflows/python_wheels.yml
```diff
@@ -173,11 +182,20 @@ jobs:
...
+          if ${{ inputs.environment != '' }}; then
+            # Validate against the public images
+            base_image=$os_image
+          else
+            # If we are not staging, we instead validate the images
+            # pushed to GHCR to limit our pull rate on public registries.
+            repo_owner=${{ github.repository_owner }}
+            image_tag=`echo $os_image | sed -e 's@.*\.io/\(\)@\1@'`
+            base_image=ghcr.io/${repo_owner,,}/${image_tag#${repo_owner,,}}
+          fi
```
bmhowe23, line 194: Just thinking out loud ... it might be nice to have some sort of manual override to allow us to fall back on the public registries in case ours get deleted and/or out of date for some reason. If that happens, though, we could presumably just manually run the gh_registry workflow to repopulate, right? I'll leave that up to you.

.github/workflows/python_wheels.yml
```diff
@@ -225,7 +243,7 @@ jobs:
           run: |
             # Install openmpi and mpi4py by conda
             # Install conda
-            dnf install -y --nobest --setopt=install_weak_deps=False wget openssh-clients.$(uname -m)
+            dnf install -y --nobest --setopt=install_weak_deps=False wget openssh-clients
```
bmhowe23, line 246: Just curious ... is this fixing some pre-existing bug?

bmhowe23 (approved):
Overall, LGTM. Just a few comments/questions....

.github/workflows/gh_registry.yml
```diff
@@ -1,21 +1,50 @@
...
-        required: true
-        description: The full name of the image, including tag.
-      platforms:
-        type: string
-        required: false
-        description: The platform(s) for which the image should be available. 
-        default: linux/amd64,linux/arm64
-  
-name: Push to GHCR
+  schedule:
```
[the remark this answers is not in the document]
bettinaheim: 1) I didn't hardcode the repo owner anywhere, so packages should actually be pushed to your own org as part of this workflow.
2) You can always disable a specific workflow on your repo. Overall, it hence seems more friendly to allow this workflow to be run on forks depending on whether that repo enabled it, rather than having to have an additional commit on forks to enable the workflow if desired.

.github/workflows/gh_registry.yml
```diff
@@ -24,6 +53,43 @@ jobs:
...
+        image: ${{ fromJson(needs.ghcr_config.outputs.external_images).images }}
+      fail-fast: false
+
+    steps:
+      - name: Login to GitHub CR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
```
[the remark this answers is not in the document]
bettinaheim: The actor may change, but that's fine/good. The only thing that is relevant are the permissions of the github token. Any use of PATs at this point I think is mostly superseded by the gh token and should be avoided when possible. The github token here is in fact the only token that would be permitted to push the images, since I personally can't push images without using a GitHub workflow (nor can anyone else).

.github/workflows/python_wheels.yml
bettinaheim: Yes, exactly, the idea is to manually run the gh_registry workflow. The override would require dragging a configuration through all edited workflows, so I don't want to do that pre-emptively. If we see we need it, I'll add it.

.github/workflows/python_wheels.yml
bettinaheim: No, I don't think so. Just seems nicer to let the package manager do it's job and find the right package for us.

.github/workflows/gh_registry.yml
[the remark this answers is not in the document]
bmhowe23: > 1. I didn't hardcode the repo owner anywhere, so packages should actually be pushed to your own org as part of this workflow.

Does that mean they will be pushed to ghcr.io/bmhowe23 for me? Or would it auto-detect that I am part of the NVIDIA org and push it to NVIDIA's folder?

> 2. You can always disable a specific workflow on your repo.

True, but I suspect most of the people who fork our repo don't take the time to do that. So I think we'd end up with ~92 forks all doing the same thing every night and making separate copies of the same data. Is there a way to make it default to off for everybody else and just have the NVIDIA repo enable it?

Issue #1000, opened 2023-12-04
Documentation for MPI plugin


1tnguyen:
We need to add documentation about the new Cuda Quantum MPI, e.g., how to activate custom plugin for local MPI installation, plugin fallback mechanism, etc.

PR #1095, opened 2024-01-16
MPI plugin docs

1tnguyen:
### Description

Adding documentation for the MPI plugin:  

- types of plugins and high-level descriptions of the fallback mechanism.

- how to activate custom plugin for local MPI installation.

Resolved: https://github.com/NVIDIA/cuda-quantum/issues/1000


docs/sphinx/install.rst
```diff
@@ -586,6 +586,93 @@ inside the container. Note that most Python packages that use GPU-acceleration,
...
+  .. note::
+
+    Manually-activated MPI plugin takes precedence over builtin plugin. 
+    For instance, the user may choose to reinstall a different MPI library inside the Docker container and 
+    activate a new CUDA Quantum MPI plugin. The newly-activated plugin will always be used even though a builtin one 
+    was shipped with the Docker image. 
+
+.. tab:: Manual Activation
+
+  **Use case**: data centers with custom MPI installation, CUDA Quantum binary distributions without MPI support (including Python wheels).
```
efratshabtai, line 620: Does it include also users using the docker image but have a local MPI implementation that is not open MPI?

docs/sphinx/install.rst
```diff
@@ -586,6 +586,93 @@ inside the container. Note that most Python packages that use GPU-acceleration,
...
+
+  **Instructions:**
+
+  (1) Locate the CUDA Quantum `distributed_interfaces` sub-directory at the top-level of the CUDA Quantum installation directory.
+
+  (2) Execute the `activate_custom_mpi.sh` bash script.
```
efratshabtai, line 626: What is there are more than one local MPI implementation available? (is that possible?)
Is there a way to specify which one we want to use?

docs/sphinx/install.rst
efratshabtai, line 620: Curious to why we don't have MPI support in binary distributions and Python wheels

docs/sphinx/install.rst
```diff
@@ -586,6 +586,93 @@ inside the container. Note that most Python packages that use GPU-acceleration,
...
+    CUDA Quantum requires users to set an environment variable `$CUDAQ_MPI_COMM_LIB` containing the path to that shared library.
+    This is done automatically when executing that activation script. Users may wish to persist that environment variable 
+    between bash sessions, e.g., via the `.bashrc` file.
+
+
+.. tab:: `mpi4py <https://mpi4py.readthedocs.io/>`__ Wrapper
+
+  **Use case**: Python users
+
+  Although we recommend manual MPI activation for optimized performance and compatibility, users may wish
```
efratshabtai, line 660: What is the performance impact of using mpi4py over manual activation? 
Maybe write it differently by removing this part and at the end specifying that it might have a perf hit and why.

docs/sphinx/install.rst
1tnguyen: Yes, it does. The manually-activated MPI plugin (targeting a local MPI implementation) will override the built-in one in the docker image (based on OpenMPI).

1tnguyen: 'CUDA Quantum binary distributions' here refers to general distributions that may or may not have MPI bundled. Python wheel distribution is one example where we don't currently have MPI in the distributed binary. 
I think the main reason is that the default Python `pip` package manager doesn't support non-Python packages, like MPI, as a package dependency.

docs/sphinx/install.rst
1tnguyen: Good point. I'll rewrite this part.

docs/sphinx/install.rst
bettinaheim: Re Python wheels: wheels have a pretty strict size limit, and it is hence expected that wheels do take dependencies on other wheels rather than include the same content. If you look at the [mpi4py page on PyPI](https://pypi.org/project/mpi4py/) you find that the wheel itself only contains the bindings, but expects an existing MPI installation on the system - like I already mentioned regarding CUDA. It hence makes little sense for us to declare that wheel as a dependency. A better way to get a fully functional MPI installation usable from Python is to follow the instructions we give, which install a *CUDA-aware!* MPI implementation along with the python bindings using conda.

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+CUDA Quantum MPI API for C++ and Python will then be delegated 
+to the plugin implementation selected at runtime.  
+
+The following table summarizes mechanisms whereby a CUDA Quantum MPI plugin can be picked up.
+
+.. tab:: Built-in
+
+  **Use case**: build from source, Docker containers
+
+  When a local MPI installation is detected at build time, CUDA Quantum will build an MPI plugin targeting this implementation when build from source.
```
bettinaheim, line 718:
```suggestion
  When a local MPI installation is detected at build time, CUDA Quantum will build an MPI plugin targeting this implementation.
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+  Similarly, the :ref:`CUDA Quantum Docker image <install-docker-image>` is shipped with a builtin MPI plugin based on its optimized OpenMPI installation inside the image.
```
bettinaheim, line 719:
```suggestion
  Similarly, the :ref:`CUDA Quantum Docker image <install-docker-image>` is shipped with a built-in MPI plugin based on an optimized OpenMPI installation included in the image. 
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+
+  **Use case**: build from source, Docker containers
+
+  When a local MPI installation is detected at build time, CUDA Quantum will build an MPI plugin targeting this implementation when build from source.
+  Similarly, the :ref:`CUDA Quantum Docker image <install-docker-image>` is shipped with a builtin MPI plugin based on its optimized OpenMPI installation inside the image. 
+  No action is required to use this built-in plugin. 
+
+  .. note::
+
+    Manually-activated MPI plugin takes precedence over builtin plugin.
```
bettinaheim, line 724:
```suggestion
    A manually activated MPI plugin takes precedence over a built-in plugin. 
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+  (1) Using the CUDA Quantum Docker image on systems that have a different MPI implementation.
+  For instance, HPC data centers often have a vendor-optimized MPI library pre-installed on
+  their system, which need to be swapped with the MPI library in the Docker image at runtime.
+  With manual activation, a custom CUDA Quantum MPI plugin can be built against the system MPI library, 
+  guaranteeing runtime compatibility when MPI injection into the container occurs. 
+
+  (2) Post-deployment MPI activation for CUDA Quantum binaries that do not have MPI support.
+  Depending on the distribution channel, CUDA Quantum binaries may not come with any pre-built
+  MPI plugin. For example, CUDA Quantum Python wheels from PyPI and :ref:`pre-built binaries <install-prebuilt-binaries>` 
+  do not have a built-in MPI plugin. CUDA Quantum MPI capabilities can be activated post-installation
```
bettinaheim, line 742:
```suggestion
  do not include a built-in MPI plugin. CUDA Quantum MPI capabilities can be activated post-installation
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+    `${MPI_PATH}/lib64` or `${MPI_PATH}/lib` is expected to contain `libmpi.so`.
+
+  .. note::
+
+    After the initial activation, if performed from the CUDA Quantum install directory,the activated will always be used to handle CUDA Quantum MPI API.
+
+  .. note::
+
+    Executing the activation script from the CUDA Quantum installation directory requires *write* permissions to that directory.
+    If that is not feasible, the user might copy the `distributed_interfaces` sub-directory to a local location and execute the
```
bettinaheim, line 770:
```suggestion
    If you do not have the necessary permissions, copy the `distributed_interfaces` sub-directory to a local location and execute the 
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+    CUDA Quantum requires users to set an environment variable `$CUDAQ_MPI_COMM_LIB` containing the path to that shared library.
+    This is done automatically when executing that activation script. Users may wish to persist that environment variable 
+    between bash sessions, e.g., via the `.bashrc` file.
+
+
+.. tab:: mpi4py Wrapper
+
+  **Use case**: Python users 
+
+  If manual activation of CUDA Quantum MPI plugin is not feasible, users may attain MPI support via the
```
bettinaheim, line 783:
```suggestion
  If you have not manually activated a CUDA Quantum MPI plugin can attain MPI support via the 
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,108 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+    between bash sessions, e.g., via the `.bashrc` file.
+
+
+.. tab:: mpi4py Wrapper
+
+  **Use case**: Python users 
+
+  If manual activation of CUDA Quantum MPI plugin is not feasible, users may attain MPI support via the 
+  the `mpi4py <https://mpi4py.readthedocs.io/>` Python package. On a system with no builtin nor manually-activated plugin, 
+  CUDA Quantum will try to locate the `mpi4py` package.
```
bettinaheim, line 785:
```suggestion
  CUDA Quantum will automatically try to locate the `mpi4py` package.
```

bettinaheim (approved):

PR #1099, opened 2024-01-17
UCCSD Bug Fixes #952

### LINKED ISSUE #952: UCCSD code in the latest version (0.4.1) is incorrect
### Required prerequisites

### Describe the bug

The coupled cluster wave function can be written as following:

image

When we convert from fermionin to qubit using jordan wigner we obtain

image

Currently, if you check the UCCSD code [(https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/domains/chemistry/uccsd.h)](url) you will find that you only implemented the  X and Y operator but not thee Z operator. This leads to incorrect result. 

Note that the UCCSD in the previous version 0.4.0 is correct. 

You can use this H2 example in the tutorial.

`cudaq.kernels.uccsd(kernel, qubits, params, nelec, qubits_num)`
`energy_of,data=cudaq.vqe(kernel,molecule,optimizer,tot_params)`

### Expected behavior

 Correcting it is essential to preserve the fermionic properties of the electrons. 

Not a regression

### Environment

- **CUDA Quantum version**: 0.4.1
- **Python version**: 3.11
- **C++ compiler**: N/A
- **Operating system**: linux

### Suggestions

You should add the missing "Z" as in the equations above.

amccaskey:
UCCSD bug fixes, implementation adapted from @marwafar 

Fixes #952

### DIFF
--- a/runtime/cudaq/domains/chemistry/uccsd.h
+++ b/runtime/cudaq/domains/chemistry/uccsd.h
@@ -18,187 +18,608 @@ using DoubleIndices = std::vector<std::size_t>;
 using Excitations =
     std::tuple<std::vector<SingleIndices>, std::vector<DoubleIndices>>;
 
-Excitations generateExcitations(std::size_t nElectrons, std::size_t nOrbitals) {
-  std::vector<double> sz(nOrbitals);
-  for (auto &i : cudaq::range(nOrbitals))
-    sz[i] = (i % 2 == 0) ? 0.5 : -0.5;
+using excitation_list = std::vector<std::vector<std::size_t>>;
 
-  std::vector<SingleIndices> singles;
-  for (auto &r : cudaq::range(nElectrons))
-    for (std::size_t p = nElectrons; p < nOrbitals; p++) {
-      if (sz[p] - sz[r] == 0)
-        singles.emplace_back(SingleIndices{r, p});
-    }
+std::tuple<excitation_list, excitation_list, excitation_list, excitation_list,
+           excitation_list>
+get_uccsd_excitations(std::size_t numElectrons, std::size_t numQubits) {
+  auto numSpatialOrbs = numQubits / 2;
+  // check rounding
+  auto numOccupied = static_cast<std::size_t>(std::ceil(numElectrons / 2));
+  auto numVirtual = numSpatialOrbs - numOccupied;
+  excitation_list singlesAlpha, singlesBeta, doublesMixed, doublesAlpha,
+      doublesBeta;
+  std::vector<std::size_t> occupiedAlpha, virtualAlpha, occupiedBeta,
+      virtualBeta;
 
-  std::vector<DoubleIndices> doubles;
-  for (auto &s : cudaq::range(nElectrons - 1))
-    for (std::size_t r = s + 1; r < nElectrons; r++)
-      for (std::size_t q = nElectrons; q < nOrbitals - 1; q++)
-        for (std::size_t p = q + 1; p < nOrbitals; p++)
-          if (sz[p] + sz[q] - sz[r] - sz[s] == 0)
-            doubles.emplace_back(DoubleIndices{s, r, q, p});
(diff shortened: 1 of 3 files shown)


bmhowe23 (commented):
Do our chemistry interfaces need to remain stable? I.e., should the old function interfaces (`generateExcitations`, `singletExcitation`, `doubletExcitation`, `uccsd_num_parameters`, and `uccsd`) be retained and/or deprecated? We normally try to do that for other interfaces, but I'm not sure about this one.

amccaskey:
> Do our chemistry interfaces need to remain stable? I.e., should the old function interfaces (`generateExcitations`, `singletExcitation`, `doubletExcitation`, `uccsd_num_parameters`, and `uccsd`) be retained and/or deprecated? We normally try to do that for other interfaces, but I'm not sure about this one.

No. These will ultimately be moved to library code that builds atop cuda-quantum outside of this repository. With this PR I'm simply fixing a bug so that users of the current code can still use it while the libraries are being implemented.

1tnguyen (approved):
Looking good. Fixing a tricky bug in UCCSD 👍

unittests/domains/ChemistryTester.cpp
```diff
@@ -14,21 +14,24 @@
...
-  EXPECT_EQ(1, doubles[0][1]);
-  EXPECT_EQ(2, doubles[0][2]);
-  EXPECT_EQ(3, doubles[0][3]);
+  std::size_t numElectrons = 2;
+  std::size_t numQubits = 4;
+
+  auto [singlesAlpha, singlesBeta, doublesMixed, doublesAlpha, doublesBeta] =
+      cudaq::get_uccsd_excitations(numElectrons, numQubits);
+  EXPECT_TRUE(doublesAlpha.empty());
+  EXPECT_TRUE(doublesBeta.empty());
```
1tnguyen, line 23: Very minor: It'd be nice to have a test case for this (maybe with more spin orbitals and electrons).

runtime/cudaq/domains/chemistry/uccsd.h
```diff
@@ -13,192 +13,621 @@
...
+/// @brief An excitation_list is a vector of lists of indices.
+using excitation_list = std::vector<std::vector<std::size_t>>;
+
+/// @brief UCCSD excitations for single and double excitations.
+struct excitations {
+  excitation_list singles_alpha;
+  excitation_list singles_beta;
+  excitation_list doubles_mixed;
+  excitation_list doubles_alpha;
+  excitation_list doubles_beta;
```
1tnguyen, line 26: Question only: is there a reason to split them out explicitly vs. just having `singles` and `doubles` lists?
For example, we need to write 3 identical for loops to handle double excitations.

unittests/domains/ChemistryTester.cpp
marwafar: I have done test for up to 10 electrons and 20 spin molecular orbitals for system like LiH, N2, and H2O and it works. I am happy to test it again.

runtime/cudaq/domains/chemistry/uccsd.h
marwafar: They are not identical loops. But we can concatenate singles_alpha and singles_beta in singles and doubles_mixed, doubles_alpha, and doubles_beta into doubles. The reason I did this splitting because I want to further optimize the UCCSD quantum circuit for the restricted spin case. In restricted spin case,  alpha and beta electrons are identical and therefore singles_alpha equals singles_beta and doubles_alpha equals doubles_beta. However, this is not possible when you have the number of alpha electrons does not equal the number of beta electrons.

unittests/domains/ChemistryTester.cpp
amccaskey: Marwa has given me a 4 electron, 8 qubit case. I will add it here.

bmhowe23 (approved):
👍

PR #1101, opened 2024-01-17
Enable additional tests when installed outside a docker image

bettinaheim:
Currently, a handful of tests is still failing. They are disabled and tracked in https://github.com/NVIDIA/cuda-quantum/issues/1102 and https://github.com/NVIDIA/cuda-quantum/issues/1103.


test/NVQPP/mapping_test-2.cpp
```diff
@@ -6,8 +6,8 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ -v %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t |& FileCheck %s
-// RUN: nvq++ -std=c++20 --enable-mlir %s -o %t
+// RUN: nvq++ %cpp_std -v %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t |& FileCheck %s
+// RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
```
bettinaheim, line 10: I am assuming the second line here was intended to be c++17 rather than fixing it at c++20.

bmhowe23: I believe that came from @schweitzpgi as part of #973. I don't know for sure, but I suspect he threw one `c++20` in there just to make sure that it worked when specifying the newer verison (and not just the older version).

bettinaheim: Let me keep that then, provided C++20 is available.

test/NVQPP/conditional_sample.cpp
```diff
@@ -7,9 +7,11 @@
  ******************************************************************************/
 
 // clang-format off
-// RUN: nvq++ --enable-mlir %s -o %t && %t
-// RUN: nvq++ --target quantinuum --emulate %s -o %t && %t
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t; \
+// RUN: fi
```
bmhowe23, line 13: Why do we need this special `if` check here? I think this could be simplified. (I checked my suggestion locally with `-std=cpp17` and `-std=cpp20`.)
```suggestion
// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t
```
Or if you want to keep the explicit `-std=c++17` in there:
```suggestion
// RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t
```

test/NVQPP/device_code_loaded.cpp
```diff
@@ -6,8 +6,10 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t | FileCheck %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t | FileCheck %s; \
+// RUN: fi
```
bmhowe23, line 12: Same question as in test/NVQPP/conditional_sample.cpp

test/NVQPP/include.cpp
```diff
@@ -6,8 +6,10 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t | FileCheck %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t | FileCheck %s; \
+// RUN: fi
```
bmhowe23, line 12: Same question as in test/NVQPP/conditional_sample.cpp.

test/NVQPP/lambda_introspection.cpp
```diff
@@ -6,8 +6,10 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t | FileCheck %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t | FileCheck %s; \
+// RUN: fi
```
bmhowe23, line 12: Same question as in test/NVQPP/conditional_sample.cpp

test/NVQPP/preprocessor_defines.cpp
```diff
@@ -6,9 +6,11 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t | FileCheck %s
-// RUN: nvq++ --enable-mlir -DCUDAQ_HELLO_WORLD %s -o %t && %t | FileCheck --check-prefixes=DEFINE_ON %s
+// RUN: nvq++ %cpp_std --enable-mlir -DCUDAQ_HELLO_WORLD %s -o %t && %t | FileCheck --check-prefixes=DEFINE_ON %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t | FileCheck %s; \
+// RUN: fi
```
bmhowe23, line 13: Same question as in test/NVQPP/conditional_sample.cpp

test/NVQPP/to_integer.cpp
```diff
@@ -6,8 +6,10 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
+// RUN:   nvq++ --enable-mlir %s -o %t && %t; \
+// RUN: fi
```
bmhowe23, line 12: Same question as in test/NVQPP/conditional_sample.cpp

docker/build/assets.Dockerfile
```diff
@@ -99,7 +99,9 @@ ENV CUDA_QUANTUM_VERSION=$release_version
 
 RUN cd /cuda-quantum && source scripts/configure_build.sh && \
     ## [>CUDAQuantumBuild]
-    CUDAQ_PYTHON_SUPPORT=OFF CUDAQ_WERROR=false \
+    CUDAQ_WERROR=false \
+    CUDAQ_PYTHON_SUPPORT=OFF \
+    CUDAHOSTCXX="$CXX" \
```
bmhowe23, line 104: Doesn't this effectively undo the change in `scripts/build_cudaq.sh`? That is - when building using this Dockerfile, will CUDA be properly detected and supported?

bmhowe23 (approved):

test/NVQPP/conditional_sample.cpp
bettinaheim: My thinking was exactly to keep the explicit c++17 so that we also have some minimal c++17 tests when running in our dev images. And the if-statement was only there to not run the same exact test twice.

bmhowe23: > My thinking was exactly to keep the explicit c++17 so that we also have some minimal c++17 tests when running in our dev images. And the if-statement was only there to not run the same exact test twice.

OK. That's what I was suspecting, but I wasn't sure.

docker/build/assets.Dockerfile
bettinaheim: The need for this here was the motivation for having originally added that to the build_cudaq.sh script. This variable is indeed needed for the build to work within this environment (otherwise that "check cuda compiler works" checks that cmake runs fail - not sure why the same checks do not run in the dev image...). Conversely, setting it here is fine since we are actually using a supported host compiler (gcc11 rather than gcc12 in the dev image).

test/MixedLanguage/cuda-1.cpp
```diff
@@ -8,7 +8,7 @@
 
 // REQUIRES: nvcc
 
-// RUN: (nvcc -c -std=c++17 %p/cuda-1.cu -o %t.o && \
+// RUN: (nvcc -c -Xcompiler -fPIC -std=c++17 %p/cuda-1.cu -o %t.o && \
```
schweitzpgi, line 11: Should this use `%cpp_std` for both compilers? We'll eventually want to use a recent CUDA toolkit, no?

test/NVQPP/IQPE_conditionals.cpp
```diff
@@ -6,8 +6,8 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir %s -o %t && %t | FileCheck %s
-// RUN: nvq++ --target quantinuum --emulate %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t | FileCheck %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
```
schweitzpgi, line 11: Why leave the `-std=c++17` test line?

test/NVQPP/auto_kernel.cpp
```diff
@@ -6,8 +6,10 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ --enable-mlir -v %s -o %t && %t | FileCheck %s
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
```
schweitzpgi, line 10: What is this conditional about?

bmhowe23: > What is this conditional about?

See https://github.com/NVIDIA/cuda-quantum/pull/1101#discussion_r1457449084

schweitzpgi: Thanks, Ben. I saw that after I started commenting.

I may be wrong, but this is running an entire new build so that it can stick `-std=c++17` in on a few more tests use `nvq++` and `cudaq-quake`?

bmhowe23: Yes, that's my interpretation...it's giving more test coverage to the `c++17` code than previously existed.

PR #1104, opened 2024-01-18
Follow-up to #1010: Update REST client to support gzip

bmhowe23:
The new version of curl (introduced in #1010) does not automatically decompress REST data that has `Content-Encoding: gzip`. Add support for that here.


runtime/common/RestClient.cpp
```diff
@@ -9,10 +9,45 @@
...
+std::string decompress_gzip(const std::string &data) {
+  if (data.empty())
+    return data;
+
+  std::string decompressed;
+  z_stream zs{};
+  zs.avail_in = data.size();
+  zs.next_in = reinterpret_cast<Bytef *>(const_cast<char *>(data.data()));
+
+  // Add 16 to indicate gzip decoding
```
bettinaheim, line 27: Is this magic number part of an official standard?

bmhowe23: > Is this magic number part of an official standard?

It is from https://www.zlib.net/manual.html#Advanced, which says:

> windowBits can also be greater than 15 for optional gzip decoding. Add 32 to windowBits to enable zlib and gzip decoding with automatic header detection, or add 16 to decode only the gzip format (the zlib format will return a Z_DATA_ERROR)

It's also in zlib.h (the file that declares `inflateInit2`)

> windowBits can also be greater than 15 for optional gzip decoding.  Add
   32 to windowBits to enable zlib and gzip decoding with automatic header
   detection, or add 16 to decode only the gzip format (the zlib format will
   return a Z_DATA_ERROR).  If a gzip stream is being decoded, strm->adler is a
   CRC-32 instead of an Adler-32.  Unlike the gunzip utility and gzread() (see
   below), inflate() will *not* automatically decode concatenated gzip members.
   inflate() will return Z_STREAM_END at the end of the gzip member.  The state
   would need to be reset to continue decoding a subsequent gzip member.  This
   *must* be done if there is more data after a gzip member, in order for the
   decompression to be compliant with the gzip standard (RFC 1952).

bettinaheim (approved):

PR #1112, opened 2024-01-19
Including uninstall script when installed via installer

bettinaheim:


.github/workflows/prebuilt_binaries.yml
```diff
@@ -323,13 +323,38 @@ jobs:
...
           image: validation:local
           user: cudaq
           shell: bash -l
           run: |
+            # Uninstall CUDA Quantum
+            sudo bash ${CUDA_QUANTUM_PATH}/uninstall.sh -y
+            llvm_dir=`bash -c 'source build_environment.sh && echo $LLVM_INSTALL_PREFIX'`
+            cuquantum_dir=`bash -c 'source build_environment.sh && echo $CUQUANTUM_INSTALL_PREFIX'`
+            cutensor_dir=`bash -c 'source build_environment.sh && echo $CUTENSOR_INSTALL_PREFIX'`
+            if [ -d "$llvm_dir" ] || [ -d "$llvm_dir" ] || [ -d "$llvm_dir" ] ||  [ -d "$CUDA_QUANTUM_PATH" ]; then
```
bmhowe23, line 338: This looks like duplicate checks for `$llvm_dir`. Is it supposed to be this instead?
```suggestion
            if [ -d "$llvm_dir" ] || [ -d "$cuquantum_dir" ] || [ -d "$cutensor_dir" ] ||  [ -d "$CUDA_QUANTUM_PATH" ]; then
```

.github/workflows/prebuilt_binaries.yml
```diff
@@ -323,13 +323,38 @@ jobs:
...
+            env -i bash -l
+            if [ -n "$CUDA_QUANTUM_PATH" ] || [ -n "$(echo $PATH | grep -o cudaq)" ]; then
+              echo "::error::Some environment variables still list CUDA Quantum after uninstalling."
+              exit 101
+            fi
+
+            # Re-install CUDA Quantum
+            MPI_PATH=/usr/local/openmpi \
+            sudo -E bash install_cuda_quantum.* --accept > install_log.txt
+            cat install_log.txt
```
bmhowe23, line 351: Potential simplification?
```suggestion
            sudo -E bash install_cuda_quantum.* --accept | tee install_log.txt
```

docker/release/installer.Dockerfile
```diff
@@ -86,7 +50,7 @@ RUN source /cuda-quantum/scripts/configure_build.sh && \
     ## [<CUDAQuantumAssets]
 
 ## [Self-extracting Archive]
-RUN bash /makeself/makeself.sh --gzip --license /cuda-quantum/LICENSE \
+RUN bash /makeself/makeself.sh --gzip --sha256 --license /cuda-quantum/LICENSE \
```
bmhowe23, line 53: There is a similar `code-block` in `data_center_install.rst`. Does that also need to be updated since you're updating this?

docker/test/installer/linux.Dockerfile
```diff
@@ -72,6 +72,7 @@ RUN . /etc/profile && nvq++ --help
 
 ## [ADD tools for validation]
 ADD scripts/validate_container.sh /home/cudaq/validate.sh
+ADD scripts/configure_build.sh /home/cudaq/build_environment.sh
```
bmhowe23, line 75: I would recommend keeping the filenames the same to make these scripts a bit easier to follow. If you accept this change, you'll probably need to update your new changes in `.github/workflows/prebuilt_binaries.yml`, too.
```suggestion
ADD scripts/configure_build.sh /home/cudaq/configure_build.sh
```

docs/sphinx/install.rst
```diff
@@ -697,6 +697,31 @@ While the above packages are sufficient to use GPU-acceleration within CUDA Quan
...
+--------------------------------
+
+If you installed the CUDA Quantum Python wheels, you can update to the latest release
+by running the command 
+
+.. code-block:: console
+
+    python3 -m pip install --upgrade cuda-quantum
+
+If you installed the CUDA Quantum pre-built binaries, you should first uninstall your
```
bmhowe23, line 710:
```suggestion
If you previously installed the CUDA Quantum pre-built binaries, you should first uninstall your 
```

scripts/migrate_assets.sh
```diff
@@ -51,17 +51,67 @@ while getopts ":c:s:t:" opt; do
...
 function move_artifacts {
     mkdir -p "$2" -m 755 && cd "$1"
     echo "Updating $2 with artifacts in $1:"
+
+    # Prompting for confirmation to remove the copied files in that folder:
+    echo 'echo "Cleaning up '$2':"' >> "$remove_assets"
+    echo "$confirmation_prompt" >> "$remove_assets"
+    echo 'if $continue; then' >> "$remove_assets"
+
     for file in `find . -type f`;
```
bmhowe23, line 105: I know you didn't change this for this review, but I happened to see it and think about it (regarding filenames with spaces). I believe the following change would handle filenames with spaces better (but feel free to save for later if desired).
```suggestion
    find . -type f -print0 | while IFS= read -r -d '' file; do
```

docker/release/installer.Dockerfile
bettinaheim: It's not necessary but an additional check to validate the archive integrity. I added it to the docs as well.

docker/test/installer/linux.Dockerfile
bettinaheim: done

bmhowe23 (approved):
👍

PR #1113, opened 2024-01-19
Workflow configs for feature branches

bettinaheim:
These changes make sure that feature branches get their own build caches.
Given the increasingly larger number of tags we create, I've also added a job once a day that keeps only a certain number of tags for images on GHCR (exact number is defined in the ghcr_config.json file): https://github.com/NVIDIA/cuda-quantum/actions/runs/7585004562


bmhowe23 (approved):
LGTM

PR #1114, opened 2024-01-19
Fix bug 1108 - race condition in sample_async

amccaskey:
Race condition due to non-uniqueness of hash value for storage of opaque arguments in async multi-qpu scenario.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_sample.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample.cpp
@@ -99,7 +99,7 @@ async_sample_result pySampleAsync(kernel_builder<> &builder,
   std::hash<std::string> hasher;
   // Create a unique integer key that combines the kernel name
   // and the validated args.
-  std::size_t uniqueHash = hasher(kernelName) + hasher(py::str(args));
+  std::size_t uniqueHash = hasher(kernelName) + hasher(py::str(args)) + qpu_id;
   // Add the opaque args to the holder and pack the args into it
   asyncArgsHolder.emplace(uniqueHash, std::make_unique<OpaqueArguments>());
   packArgs(*asyncArgsHolder.at(uniqueHash).get(), validatedArgs);
--- a/python/runtime/cudaq/algorithms/py_sample.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample.cpp
@@ -23,11 +23,6 @@
 
 namespace cudaq {
 
-/// @brief Global cache map of OpaqueArguments
-// Using the same cache map provided by py_observe (hence extern).
-extern std::unordered_map<std::size_t, std::unique_ptr<OpaqueArguments>>
-    asyncArgsHolder;
-
 /// @brief Sample the state produced by the provided builder.
 sample_result pySample(kernel_builder<> &builder, py::args args = {},
                        std::size_t shots = 1000,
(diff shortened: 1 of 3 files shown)


python/runtime/cudaq/algorithms/py_sample.cpp
```diff
@@ -99,7 +99,7 @@ async_sample_result pySampleAsync(kernel_builder<> &builder,
   std::hash<std::string> hasher;
   // Create a unique integer key that combines the kernel name
   // and the validated args.
-  std::size_t uniqueHash = hasher(kernelName) + hasher(py::str(args));
+  std::size_t uniqueHash = hasher(kernelName) + hasher(py::str(args)) + qpu_id;
```
bmhowe23, line 102: https://github.com/NVIDIA/cuda-quantum/blob/main/python/runtime/cudaq/algorithms/py_observe.cpp#L80 has the same hash formula. Is it vulnerable to the same problem?

amccaskey: Yes good catch. I'll make the update.

bmhowe23: It may be helpful to gracefully avoid collisions. I tested the following gist which avoids collisions and it appears to work (even without the `+ qpu_id` change), and I believe it would be insensitive to collisions.
https://gist.github.com/bmhowe23/429773e97aea6d2a3c21b8650b0bd178

(The corresponding changes for `observe_async` may need adjustments, though. `observe_async` seems to be failing for me even with the `+ qpu_id` change, so I can't properly test out this change there.)

1tnguyen: Playing devil's advocate here :-) 

What do you think about removing the `asyncArgsHolder` cache map altogether and instead attaching the `OpaqueArguments` to the functor itself?

e.g.,
```
  // Add the opaque args to the holder and pack the args into it
  auto argsHolder = std::make_unique<OpaqueArguments>();
  packArgs(*argsHolder, validatedArgs);
  return details::runSamplingAsync(
      detail::make_copyable_function(
          [&builder, argData = std::move(argsHolder)]() mutable {
            builder.jitAndInvoke(argData->data());
          }),
      platform, kernelName, shots, qpu_id);
```

amccaskey: Thanks to both of you for the suggestions. Thien I like your approach and have updated the code with it for both sample and observe

python/runtime/cudaq/algorithms/py_sample.cpp
```diff
@@ -23,11 +23,6 @@
 
 namespace cudaq {
 
-/// @brief Global cache map of OpaqueArguments
```
1tnguyen, line 26: nit: we could also clean up this var in `py_observe.cpp`

```
/// @brief Global cache map of OpaqueArguments
///
/// For asynchronous execution, we need to construct OpaqueArguments
/// outside of the async lambda invocation. If we don't, then we will be
/// using Python types outside of the current GIL context. Bad things happen
/// then.
std::unordered_map<std::size_t, std::unique_ptr<OpaqueArguments>>
    asyncArgsHolder;
```

1tnguyen (approved):
LGTM 👍

python/runtime/cudaq/algorithms/py_sample.cpp
amccaskey: my bad. good catch

PR #1115, opened 2024-01-19
Add library mode and kernel_builder state allocation with initialization

amccaskey:
This moves the work in PR #1090 into this new feature branch. Work by others will build off of this. 

Things that still need to be done: 

- [ ] Python bindings
- [ ] Testing 
- [ ] Update simulators to set state data (custatevec, cutensornet, mgpu)
- [ ] Hook up the ASTBridge 
- [ ] Review / refactor InitializeStateOp and LowerToQIR.cpp updates (@schweitzpgi) 
- [ ] Verify bit ordering (@boschmitt)

### DIFF
--- a/include/cudaq/Optimizer/CodeGen/Pipelines.h
+++ b/include/cudaq/Optimizer/CodeGen/Pipelines.h
@@ -57,6 +57,7 @@ void addPipelineToQIR(mlir::PassManager &pm,
   if constexpr (QIRProfile) {
     cudaq::opt::addQIRProfilePipeline(pm, convertTo);
   }
+  pm.addPass(mlir::createCanonicalizerPass());
 }
 
 inline void addPipelineToOpenQASM(mlir::PassManager &pm) {
--- a/include/cudaq/Optimizer/CodeGen/Pipelines.h
+++ b/include/cudaq/Optimizer/CodeGen/Pipelines.h
@@ -57,6 +57,8 @@ void addPipelineToQIR(mlir::PassManager &pm,
   if constexpr (QIRProfile) {
     cudaq::opt::addQIRProfilePipeline(pm, convertTo);
   }
+
+  // Make sure we don't have any leftover dead code
   pm.addPass(mlir::createCanonicalizerPass());
 }
 
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -281,6 +281,26 @@ def quake_ExtractRefOp : QuakeOp<"extract_ref", [Pure]> {
   }];
 }
 
+def InitializeStateOp : QuakeOp<"qinit"> {
+  let summary = "Initialize the quantum state to a specific complex vector.";
+  let description = [{
+    Given a !cc.ptr pointing to a complex data array of size 2**N, where N is the number of qubits 
+    in the targets operand, initialize the state of those target qubits to 
+    the provided state vector. This operation returns a new quake.veq instance 
+    and subsequent uses of the original veq should be replaced during code generation. 
+    This is to support the immediate use of this operation after an initial allocation.
+  }];
+
+  let arguments = (ins VeqType:$targets, 
+                       cc_PointerType:$state);
+  let results = (outs VeqType:$initializedState);
+
+  let assemblyFormat = [{
+    `(` $targets `,` $state `)` `:` functional-type(operands, results) attr-dict
+  }];
+
+}
+
 def quake_RelaxSizeOp : QuakeOp<"relax_size", [Pure]> {
   let summary = "Relax the constant size on a !veq to be unknown.";
   let description = [{
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -298,7 +298,6 @@ def InitializeStateOp : QuakeOp<"qinit"> {
   let assemblyFormat = [{
     `(` $targets `,` $state `)` `:` functional-type(operands, results) attr-dict
   }];
-
 }
 
 def quake_RelaxSizeOp : QuakeOp<"relax_size", [Pure]> {
--- a/lib/Optimizer/CodeGen/LowerToQIR.cpp
+++ b/lib/Optimizer/CodeGen/LowerToQIR.cpp
@@ -31,6 +31,7 @@
 #include "mlir/Conversion/LLVMCommon/Pattern.h"
 #include "mlir/Conversion/LLVMCommon/TypeConverter.h"
 #include "mlir/Conversion/MathToLLVM/MathToLLVM.h"
+#include "mlir/Conversion/ReconcileUnrealizedCasts/ReconcileUnrealizedCasts.h"
 #include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
 #include "mlir/Target/LLVMIR/ModuleTranslation.h"
 #include "mlir/Transforms/DialectConversion.h"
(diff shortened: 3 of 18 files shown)


include/cudaq/Optimizer/CodeGen/Pipelines.h
```diff
@@ -57,6 +57,9 @@ void addPipelineToQIR(mlir::PassManager &pm,
   if constexpr (QIRProfile) {
     cudaq::opt::addQIRProfilePipeline(pm, convertTo);
   }
+
+  // Make sure we don't have any leftover dead code
+  pm.addPass(mlir::createCanonicalizerPass());
```
schweitzpgi, line 62: We need to figure out why the unrealized casts are getting inserted.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -281,6 +281,25 @@ def quake_ExtractRefOp : QuakeOp<"extract_ref", [Pure]> {
...
+  let summary = "Initialize the quantum state to a specific complex vector.";
+  let description = [{
+    Given a !cc.ptr pointing to a complex data array of size 2**N, where N is the number of qubits 
+    in the targets operand, initialize the state of those target qubits to 
+    the provided state vector. This operation returns a new quake.veq instance 
+    and subsequent uses of the original veq should be replaced during code generation. 
+    This is to support the immediate use of this operation after an initial allocation.
+  }];
+
+  let arguments = (ins VeqType:$targets,
```
schweitzpgi, line 294: Do we want this to apply to a single qubit? (Ref type)

lib/Optimizer/CodeGen/LowerToQIR.cpp
```diff
@@ -76,6 +76,34 @@ class AllocaOpRewrite : public ConvertOpToLLVMPattern<quake::AllocaOp> {
...
+    {
+      std::size_t count = 0, numDeallocs = 0;
+      quake::InitializeStateOp maybeInit;
+      for (auto user : alloca->getUsers()) {
+        count++;
+        if (auto init = dyn_cast<quake::InitializeStateOp>(user))
+          maybeInit = init;
+
+        if (dyn_cast<quake::DeallocOp>(user))
+          numDeallocs++;
```
schweitzpgi, line 93: Can be simplified if addDealloc pass is corrected as above.

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -45,6 +45,43 @@ namespace cudaq::details {
...
+      fmt::runtime(
+          R"#(llvm.mlir.global external local_unnamed_addr @nvqpp.state.{}() {{addr_space = 0 : i32, alignment = 8 : i64, dso_local}} : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)> {{
+    %0 = llvm.mlir.null : !llvm.ptr<struct<(f64,f64)>>
+    %1 = llvm.mlir.constant({} : i32) : i32
+    %2 = llvm.mlir.undef : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)>
+    %3 = llvm.insertvalue %0, %2[0] : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)> 
+    %4 = llvm.insertvalue %1, %3[1] : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)> 
+    llvm.return %4 : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)>
+  }})#"),
+      hashValue, size);
```
schweitzpgi, line 62: Can't we add this to the intrinsics in Intrinsics.cpp by making the hash and size arguments passed in the calling context?

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -938,9 +1049,9 @@ std::string to_quake(ImplicitLocOpBuilder &builder) {
 
   // Strategy - we want to clone this ModuleOp because we have to
   // add a valid terminator (func.return), but it is not gauranteed that
-  // the programmer is done building up the kernel even though they've asked to
-  // look at the quake code. So we'll clone here, and add the return op (we have
-  // to or the print out string will be invalid (verifier failed)).
+  // the programmer is done building up the kernel even though they've asked
+  // to look at the quake code. So we'll clone here, and add the return op
+  // (we have to or the print out string will be invalid (verifier failed)).
```
schweitzpgi, line 1054: Is this a clang-format change? Are we using multiple versions of clang-format?

unittests/integration/builder_tester.cpp
```diff
@@ -1221,6 +1221,25 @@ CUDAQ_TEST(BuilderTester, checkControlledRotations) {
 
 #ifndef CUDAQ_BACKEND_DM
 
+// FIXME Get other backends updated to support this
```
schweitzpgi, line 1224: Maybe an issue?

runtime/cudaq/builder/kernel_builder.h
```diff
@@ -430,6 +451,15 @@ class kernel_builder : public details::kernel_builder_base {
     return details::qalloc(*opBuilder.get(), size);
   }
 
+  // @brief Return a `QuakeValue` representing the allocated
+  // quantum register, initialized to the given state vector.
+  // Note - input argument os not const here, user has to own the data.
+  QuakeValue qalloc(std::vector<cudaq::complex> &state) {
```
anthony-santana, line 457: std::span ?

runtime/cudaq/builder/kernel_builder.h
```diff
@@ -430,6 +451,15 @@ class kernel_builder : public details::kernel_builder_base {
     return details::qalloc(*opBuilder.get(), size);
   }
 
+  // @brief Return a `QuakeValue` representing the allocated
+  // quantum register, initialized to the given state vector.
+  // Note - input argument os not const here, user has to own the data.
```
anthony-santana, line 456:
```suggestion
  // Note - input argument is not const here, user has to own the data.
```

runtime/cudaq/qis/qudit.h
```diff
@@ -10,7 +10,15 @@
 
 #include "execution_manager.h"
 
+using namespace std::complex_literals;
+
 namespace cudaq {
+using complex = std::complex<double>;
+
+namespace ket {
+inline static const std::vector<complex> zero{1. + 0i, 0. + 0i};
```
anthony-santana, line 19: I think it'd be worth it to fill out this type more from the start. Something simple that allows for a variably sized 0 or 1 state. I think it should be added to the python bindings so I could lump it in with my work there

schweitzpgi (approved):
rubberstamp to get things started

include/cudaq/Optimizer/CodeGen/Pipelines.h
amccaskey: Agreed. Would you mind looking into it? Just remove this canonicalizer and you'll see the error

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
amccaskey: I think so.

runtime/cudaq/builder/kernel_builder.cpp
amccaskey: I tried at first but could not figure out how to fit it in as-is. It would be great to do so though, maybe needs just some small changes.

runtime/cudaq/builder/kernel_builder.h
amccaskey: I wonder about that - since we also need to support C++17 now.

schweitzpgi: Shouldn't this actually be `const` so that the compiler doesn't make a temporary?

include/cudaq/Optimizer/CodeGen/Pipelines.h
schweitzpgi: +1

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -281,6 +281,25 @@ def quake_ExtractRefOp : QuakeOp<"extract_ref", [Pure]> {
...
 }
 
+def InitializeStateOp : QuakeOp<"qinit"> {
+  let summary = "Initialize the quantum state to a specific complex vector.";
+  let description = [{
+    Given a !cc.ptr pointing to a complex data array of size 2**N, where N is the number of qubits 
+    in the targets operand, initialize the state of those target qubits to 
+    the provided state vector. This operation returns a new quake.veq instance 
+    and subsequent uses of the original veq should be replaced during code generation. 
+    This is to support the immediate use of this operation after an initial allocation.
```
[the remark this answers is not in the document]
schweitzpgi: See #1141

runtime/cudaq/builder/kernel_builder.cpp
schweitzpgi: We don't need these if we just create a cc.global, right?

PR #1119, opened 2024-01-22
Fix an oversight in MultiGPU code refactor

1tnguyen:
### Description

It was an oversight not setting the QPU Id there. `setId` was added to the base class to facilitate that.

Thanks @amccaskey for catching this.


amccaskey (approved):
tested locally and the bug if fixed. +1

PR #1124, opened 2024-01-23
Revert "resync features/qalloc"

schweitzpgi:
Reverts NVIDIA/cuda-quantum#1122


schweitzpgi:
Really want to merge these commits independently with a merge commit rather than crush the soda can into a single commit.

amccaskey (approved):

PR #1129, opened 2024-01-23
Additional validation during publishing

bettinaheim:
- Enables nvidia-mgpu tests for the installer
- Adds installer validation on ubuntu:22.04
- Adds a script for QA to validate the python wheels outside our CI
- Adds some comments and clean up for the remote-mqpu examples
- Other miscellaneous fixes and tweaks


.github/workflows/publishing.yml
```diff
@@ -166,6 +166,7 @@ jobs:
 
             echo "Pushing $1 to $staging_branch"
             git add "$staging_folder" && git commit -m "$image_hash"
+            git pull origin/$staging_branch 2> /dev/null || true
```
bmhowe23, line 169: I thought the remote repository had to be specified? Maybe it's inferred from the `origin/`?
```suggestion
            git pull origin origin/$staging_branch 2> /dev/null || true
```

.github/workflows/publishing.yml
```diff
@@ -713,23 +719,50 @@ jobs:
...
+        if: startsWith(matrix.os_image, 'ubuntu')
+        run: |
+          apt-get update && apt-get install -y --no-install-recommends \
+            wget ca-certificates libstdc++-11-dev libopenmpi-dev
+          distro=`echo ${{ matrix.os_image }} | tr -d . | tr -d :`
+          CUDA_DOWNLOAD_URL=https://developer.download.nvidia.com/compute/cuda/repos
+          wget "${CUDA_DOWNLOAD_URL}/$distro/x86_64/cuda-keyring_1.1-1_all.deb"
+          dpkg -i cuda-keyring_1.1-1_all.deb && version_suffix=11-8
+          apt-get update && apt-get install -y --no-install-recommends \
+              cuda-nvtx-${version_suffix} cuda-cudart-${version_suffix} \
```
bmhowe23, line 732: Just a suggestion: I'm not sure that `cuda-nvtx` is really needed. Since it is a big download, it may speed it up to remove it, but I'm ok either way.

scripts/validate_wheel.sh
```diff
@@ -0,0 +1,166 @@
...
+
+python3 "$root_folder/snippets/using/cudaq/platform/sample_async_remote.py" \
+    --backend nvidia-mgpu --servers "$servers"
+if [ ! $? -eq 0 ]; then
+    echo -e "\e[01;31mRemote platform test failed.\e[0m" >&2
+    status_sum=$((status_sum+1))
+fi
+kill %1 && wait %1 2> /dev/null
+if [ -n "$server2_devices" ]; then
+    kill %2 && wait %2 2> /dev/null
```
bmhowe23, line 160: I was seeing segfaults reported with `kill` and adding `2> /dev/null` did not make them go away (since the errors were coming from the original processes). They went away for me with `kill -9`. If you are seeing the same thing, I'd recommend adding the `-9` in order to make the logs not have benign segfaults.

bmhowe23 (approved):
👍

.github/workflows/publishing.yml
bettinaheim: I think you are right and this was originally needed for the tensornet simulator.

scripts/validate_wheel.sh
bettinaheim: I intentionally did not use the -9 because I wanted to see the segfaults on tear down. While they are potentially benign, they still indicate an issue during tear down (which Thien has since fixed).

.github/workflows/publishing.yml
bettinaheim: For the record: cublas is the only cuda runtime dependency for the nvidia-mgpu backend. cublas, cusolver and cudart are needed for the tensornet and tensornet-mps backends. Maybe we should just link these statically in the installer.

PR #1134, opened 2024-01-24
Add support for the new qvector constructor with an initialization list

schweitzpgi:
to the bridge.

### DIFF
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -2189,10 +2189,11 @@ bool QuakeBridgeVisitor::VisitInitListExpr(clang::InitListExpr *x) {
 
   // These initializer expressions are not quantum references. In this case,
   // we allocate some memory for a variable and store the init list elements
-  // there. Add the array size value
+  // there.
   auto structTy = dyn_cast<cc::StructType>(initListTy);
   std::int32_t structMems = structTy ? structTy.getMembers().size() : 0;
   std::int32_t numEles = structMems ? size / structMems : size;
+  // Generate the array size value.
   Value arrSize = builder.create<arith::ConstantIntOp>(loc, numEles, 64);
 
   // Allocate the required memory chunk.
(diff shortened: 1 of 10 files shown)


lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -2201,6 +2202,34 @@ bool QuakeBridgeVisitor::VisitInitListExpr(clang::InitListExpr *x) {
...
+  if (initializerIsGlobal) {
+    static unsigned counter = 0;
+    auto *ctx = builder.getContext();
+    auto globalTy = cc::ArrayType::get(ctx, eleTy, size);
+    SmallVector<Attribute> values;
+    auto f64Ty = builder.getF64Type();
+    for (Value v : last) {
+      auto fp = opt::factory::maybeValueOfFloatConstant(v);
+      assert(fp);
+      values.push_back(FloatAttr::get(f64Ty, *fp));
```
amccaskey, line 2215: We do want these values to be of ComplexType. You could see a simple float, but it should represent `FloatVal + 0.0 * i`.

schweitzpgi: For this PR, I'm made the assumption (probably bogus) that the initializer list is a $2^n$ list of floats and the pairwise complex values and just even and odd offsets in that list.

If we want to allow fancier complex literals, we'll have to extend this.

amccaskey: The state-vector is always a complex-valued entity. But there are likely use cases where the user would like to skirt around having to specify the full complex representation for the elements because the state is real-valued. Hence the `qvector` constructors on both `std::vector<complex<double>>` and `std::vector<double>` (we also will need `std::vector<float>` and `std::vector<complex<float>>`). I imagine bridge handling for complex elements is harder than what we have here, but we will need it. Do you want to do that in a follow up PR or here?

schweitzpgi:
I don't think these failures have anything to do with this PR.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: Let's make it a follow-up. There is a complex builtin type, but the calling conventions for complex values are mysterious. That'll take some extra work.

amccaskey (approved):

PR #1139, opened 2024-01-25
Make mock server startup more robust to delays

### LINKED ISSUE #1138: [test] Intermittent mock server test failures can occur due to insufficient time for server startup
### Required prerequisites

### Describe the bug

This is a test script issue, not a product issue.

The mock server tests spawn a server in the background and wait 3 seconds for it to start. On many platforms, this is enough time, but one some platforms, it is not. In failure cases, the verbose logs will look like this:
```
test 465
    Start 465: quantinuum-tests

465: Test command: /usr/bin/bash "QuantinuumStartServerAndTest.sh"
465: Working Directory: /workspaces/cuda-quantum/build/unittests/backends/quantinuum/
465: Test timeout computed to be: 1500
465: [==========] Running 11 tests from 1 test suite.
465: [----------] Global test environment set-up.
465: [----------] 11 tests from quantinuum_QuantinuumTester
465: [ RUN      ] quantinuum_QuantinuumTester.checkSampleSync
465: unknown file: Failure
465: C++ exception with description "HTTP POST Error - status code 0: Failed to connect to localhost port 62440 after 0 ms: Couldn't connect to server: " thrown in the test body.
465:
465: [  FAILED  ] quantinuum_QuantinuumTester.checkSampleSync (140 ms)
```

This occurs more often when CUDA Quantum is compiled with debug, but I imagine it could happen in Release mode, too, depending on the machine.

### Expected behavior

The tests should allow for more mock server startup time. That is: scripts like this need to be updated:
```bash
# Launch the fake server
PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/mock_qpu/quantinuum/__init__.py &
# we'll need the process id to kill it
pid=$(echo "$!")
sleep 3
# Run the tests
./test_quantinuum
# Did they fail? 
testsPassed=$?
# kill the server
kill -INT $pid
# return success / failure
exit $testsPassed
```

Not a regression

### Environment

- **CUDA Quantum version**: main
- **Python version**: 3.10.12
- **C++ compiler**: N/A

### Suggestions

I have a forthcoming PR to help address the issue.

bmhowe23:
Some servers require more time to start the mock servers. I was able to reproduce the problem on gorby w/ Debug builds, and applying this change fixes it. The fix is to change a hard coded sleep time to a dynamic delay loop (so as to not always wait the maximum amount of time) and increase the max time to 10 seconds.

Resolves #1138.


amccaskey (approved):
+1

PR #1142, opened 2024-01-26
Fix buildbot error.

schweitzpgi:


amccaskey (approved):
thanks for catching this

PR #1143, opened 2024-01-26
Test QPP state allocation

anthony-santana:


unittests/backends/QPPTester.cpp
```diff
@@ -1242,3 +1242,211 @@ CUDAQ_TEST(QPPTester, checkReset) {
...
+    std::vector<std::complex<double>> inputState(pow(2, qubit_count), M_PI_4);
+    auto qubits = qppBackend.allocateQubits(qubit_count, inputState.data());
+
+    qpp::ket got_state = qppBackend.getStateVector();
+    qpp::ket want_state = getZeroState(qubit_count);
+    want_state.fill(M_PI_4);
+    EXPECT_EQ(want_state, got_state);
+  }
+
+  // Broken integration test with sampling and measurements.
```
anthony-santana, line 1299: Excuse the mess in the rest of this PR, you could cut and paste just this test into the head features/qalloc branch and not need to pull down the rest of this code. 

We have a dimension mismatch issue when intermixing conventional qubit allocations with state vector allocations. The dimension error only shows up *after* the state allocation, but prevents *any* further state manipulation. I can reproduce the error with a sample call, measure call, and any gate application (qpp::apply). I'm sure there are more, but it isn't fully clear to me where the dimension issues come from. We are only using free QPP functions, no classes that should be retaining any sort of state dimension -- that should come directly from the state vector itself. And that state vector that I'm printing out looks as I'd expect it to.

runtime/nvqir/cutensornet/simulator_cutensornet.h
```diff
@@ -52,7 +52,7 @@ class SimulatorTensorNetBase : public nvqir::CircuitSimulatorBase<double> {
   virtual cudaq::ExecutionResult observe(const cudaq::spin_op &op) override;
 
   /// @brief Add qubits to the underlying quantum state
-  virtual void addQubitsToState(std::size_t count) override;
```
anthony-santana, line 55: I had a build error from this being marked as override. Not sure if anyone else ran into this

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -140,11 +139,10 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
       zero_state(0) = 1.0;
       state = qpp::kron(state, zero_state);
     } else {
-      qpp::ket initState =
-          qpp::ket::Map(const_cast<std::complex<double> *>(stateData), count);
+      qpp::ket initState = qpp::ket::Map(
+          const_cast<std::complex<double> *>(stateData), (1UL << count));
```
schweitzpgi, line 143: Might want to note in the doxygen comment that `count` is the power-of-2 exponent.

unittests/backends/QPPDMTester.cpp
```diff
@@ -128,4 +128,200 @@ CUDAQ_TEST(QPPTester, checkDensityOrderingBug) {
     EXPECT_EQ(1, qppBackend.mz(q2));
     EXPECT_EQ(0, qppBackend.mz(q3));
   }
-}
\ No newline at end of file
+}
+
+// CUDAQ_TEST(QPPTester, checkSetStateDensity) {
```
schweitzpgi, line 133: Will this be enabled with another PR?

CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ option(CUDAQ_TEST_MOCK_SERVERS "Enable Remote QPU Tests via Mock Servers." OFF)
 option(CUDAQ_DISABLE_RUNTIME "Build without the CUDA Quantum runtime, just the compiler toolchain." OFF)
 option(CUDAQ_SKIP_MPI "Do not build with MPI, even when it is available." OFF)
 option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA Quantum C++ Clang-based Frontend." OFF)
+option(CUDAQ_ENABLE_CC "Enable CUDA Quantum code coverage generation." OFF)
```
schweitzpgi: Can we use something other than `CC` for code coverage?

(CC is already overloaded with it being an environment variable, a dialect, and a C compiler.)

docs/sphinx/examples/cpp/other/builder/builder.cpp
```diff
@@ -114,6 +122,37 @@ int main() {
     counts.dump();
   }
 
+  {
+    // In a simulated environment, it is sometimes useful to be able to
```
schweitzpgi: This looks oddly familiar...

schweitzpgi (commented):
I think I am reviewing a merge ... heh

unittests/backends/QPPDMTester.cpp
schweitzpgi: This refers to the test below now. vv

anthony-santana: Yes, this has been moved to a branch where it will sit until we support state initialization on the density simulator

Issue #1098, opened 2024-01-17
[RFC] Introduce cudaq::pauli_word type to the language


amccaskey:
A powerful mechanism for circuit synthesis would be to accept `cudaq::spin_op` as input to a quantum kernel and use that with a synthesis operation like `exp_pauli`. Fully supporting `spin_op` via the AST Bridge and MLIR is a large task. An easier stepping stone that would enable a wide range of use cases (ones we are already seeing) is to introduce a `cudaq::pauli_word` type. This type would model a single term in a `spin_op`. For example, the `spin_op op = x(0) * y(1) - 2.0 * z(0) z(1)` would contain 2 pauli words, specifically `X0 Y1` and `Z0 Z1` (I propose we ignore the coefficient as part of our `cudaq::pauli_word` type, coefficients can be extracted from a `spin_op` and passed separately as a `std::vector<double>`). So the type would effectively provide an opaque type that map to a `ptr<i8>` in the AST Bridge and be compatible with our existing `ExpPauliOp`. 

```cpp
__qpu__ void test(double theta, std::vector<cudaq::pauli_word> paulis) {
  cudaq::qvector q(2);
  for (auto &p : paulis)
    exp_pauli(theta, q, p);
}

void useTest() {
  spin_op op = ... ;
  auto words = op.to_pauli_words();
  auto results = cudaq::sample(test, M_PI_2, words);
}
```
```python
kernel, thetas, paulisArg = cudaq.make_kernel(list[float], list[cudaq.pauli_word])
qubits = kernel.qalloc(2)
kernel.x(qubits[0])
kernel.for_loop(
     0, paulisArg.size(),
     lambda idx: kernel.exp_pauli(thetas[idx], qubits, paulisArg[idx]))
```

Note this is related to PR #1062. It seems smarter to force this type of input to be a separate type in the language rather than a catch-all `const char *`.

justinlietz:
> For example, the `spin_op op = x(0) * y(1) - 2.0 * z(0) z(1)` would contain 2 pauli words, specifically `X0 Y1` and `Z0 Z1` (I propose we ignore the coefficient as part of our `cudaq::pauli_word` type, coefficients can be extracted from a `spin_op` and passed separately as a `std::vector<double>`). 

I think there is a lot of value in having additional pauli types. To properly model a single in a spin_op, would we support all of the math operations that can be done between two spin_op_terms?
[...]

PR #1146, opened 2024-01-29
Introduce pauli_word in the ASTBridge, kernel_builder, and Python

amccaskey:
#1098

Enable 

```cpp
__qpu__ void test(double theta, std::vector<cudaq::pauli_word> paulis) {
  cudaq::qvector q(2);
  for (auto &p : paulis)
    exp_pauli(theta, q, p);
}

void useTest() {
  spin_op op = ... ;
  auto words = op.to_pauli_words();
  auto results = cudaq::sample(test, M_PI_2, words);
}
```
```python
kernel, thetas, paulisArg = cudaq.make_kernel(list[float], list[cudaq.pauli_word])
qubits = kernel.qalloc(2)
kernel.x(qubits[0])
kernel.for_loop(
     0, paulisArg.size(),
     lambda idx: kernel.exp_pauli(thetas[idx], qubits, paulisArg[idx]))
```

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -833,7 +833,7 @@ def ExpPauliOp : QuakeOp<"exp_pauli"> {
   let arguments = (ins
     AnyFloat:$parameter,
     VeqType:$qubits,
-    cc_PointerType:$pauli
+    AnyTypeOf<[cc_PointerType, PauliWordType]>:$pauli
   );
   let results = (outs);
 
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -1421,6 +1421,8 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
           auto ptrEleTy = computePtr.getType().getElementType();
           Value loaded = builder.create<cudaq::cc::LoadOp>(loc, ptrEleTy, v);
           processedArgs.push_back(loaded);
+        } else if (isa<quake::PauliWordType>(v.getType())) {
+          processedArgs.push_back(v);
         } else {
           reportClangError(x, mangler, "could not determine string argument");
         }
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -1037,7 +1037,8 @@ bool QuakeBridgeVisitor::VisitMaterializeTemporaryExpr(
   // The following cases are λ expressions, quantum data, or a std::vector view.
   // In those cases, there is nothing to materialize, so we can just pass the
   // Value on the top of the stack.
-  if (isa<cc::CallableType, quake::VeqType, quake::RefType, cc::StdvecType>(ty))
+  if (isa<cc::CallableType, quake::VeqType, quake::RefType,
+          quake::PauliWordType, cc::StdvecType>(ty))
     return true;
 
   // If not one of the above special cases, then materialize the value to a
@@ -1411,10 +1412,16 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
         // The C-string argument (char*) may be loaded by an lvalue to rvalue
         // cast. Here, we must pass the pointer and not the first character's
         // value.
+        v.dump();
         if (isCharPointerType(v.getType())) {
           processedArgs.push_back(v);
         } else if (auto load = v.getDefiningOp<cudaq::cc::LoadOp>()) {
           processedArgs.push_back(load.getPtrvalue());
+        } else if (auto computePtr =
+                       v.getDefiningOp<cudaq::cc::ComputePtrOp>()) {
+          auto ptrEleTy = computePtr.getType().getElementType();
+          Value loaded = builder.create<cudaq::cc::LoadOp>(loc, ptrEleTy, v);
+          processedArgs.push_back(loaded);
         } else {
           reportClangError(x, mangler, "could not determine string argument");
         }
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -1412,7 +1412,6 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
         // The C-string argument (char*) may be loaded by an lvalue to rvalue
         // cast. Here, we must pass the pointer and not the first character's
         // value.
-        v.dump();
         if (isCharPointerType(v.getType())) {
           processedArgs.push_back(v);
         } else if (auto load = v.getDefiningOp<cudaq::cc::LoadOp>()) {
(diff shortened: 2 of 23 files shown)


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -833,7 +833,7 @@ def ExpPauliOp : QuakeOp<"exp_pauli"> {
   let arguments = (ins
     AnyFloat:$parameter,
     VeqType:$qubits,
-    cc_PointerType:$pauli
+    AnyTypeOf<[cc_PointerType, PauliWordType]>:$pauli
```
schweitzpgi, line 836: I'm a little confused by this.

In the context of quake, where does a value of `PauliWordType` come from? Is it correct that this is the only operation that using this type?

Below it looks like this new type is just an alias for a `i8*`. Is that true? Are we going to allow other compositions of values to have this type eventually?

schweitzpgi: Should we be phasing out the `i8*` entirely? Maybe add a note to that intention?

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1415,6 +1416,13 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
           processedArgs.push_back(v);
         } else if (auto load = v.getDefiningOp<cudaq::cc::LoadOp>()) {
           processedArgs.push_back(load.getPtrvalue());
+        } else if (auto computePtr =
+                       v.getDefiningOp<cudaq::cc::ComputePtrOp>()) {
+          auto ptrEleTy = computePtr.getType().getElementType();
+          Value loaded = builder.create<cudaq::cc::LoadOp>(loc, ptrEleTy, v);
```
schweitzpgi, line 1422: This load doesn't look right. The other cases are passing the pointer to the Pauli string data. Shouldn't this just forward the pointer computed by `ComputePtr`?

lib/Frontend/nvqpp/ConvertType.cpp
```diff
@@ -75,7 +75,8 @@ static bool isRecursiveArithmeticProductType(Type t);
 static bool isRecursiveArithmeticSequenceType(Type t) {
   if (auto vec = dyn_cast<cudaq::cc::StdvecType>(t)) {
     auto eleTy = vec.getElementType();
-    return isArithmeticType(eleTy) || isRecursiveArithmeticProductType(eleTy) ||
+    return isa<quake::PauliWordType>(eleTy) || isArithmeticType(eleTy) ||
```
schweitzpgi, line 78: I don't see how a type can be both arithmetic (classical) and quantum.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -474,8 +474,11 @@ class GenerateKernelExecution
     auto ptrI8Ty = cudaq::cc::PointerType::get(builder.getI8Type());
     Type eleTy = stdvecTy.getElementType();
     auto innerStdvecTy = dyn_cast<cudaq::cc::StdvecType>(eleTy);
-    std::size_t eleSize =
-        innerStdvecTy ? /*(i64Type/8)*/ 8 : dataLayout->getTypeSize(eleTy);
+    std::size_t eleSize = [&]() -> std::size_t {
+      if (isa<quake::PauliWordType>(eleTy))
+        return 8;
```
schweitzpgi, line 479: Is the intention to pack/unpack this as a dynamic length item? i.e., the length is encoded as a i64 (as this suggests) and the Pauli word value appended in the next dynamic block range? More code than this is needed.

python/utils/OpaqueArguments.h
```diff
@@ -39,7 +39,8 @@ class OpaqueArguments {
   std::vector<OpaqueArgDeleter> deleters;
 
 public:
-  /// @brief Add an opaque argument and its `deleter` to this OpaqueArguments
+  /// @brief Add an opaque argument and its deletion function to this
+  /// `OpaqueArguments`
```
schweitzpgi, line 43:
```suggestion
  /// `OpaqueArguments`.
```

runtime/cudaq/builder/kernel_builder.h
```diff
@@ -305,7 +319,7 @@ struct ArgumentValidator<std::vector<T>> {
     // Validate the input vector<T> if possible
     if (auto nRequiredElements = arg.getRequiredElements();
         arg.canValidateNumElements())
-      if (input.size() != nRequiredElements)
+      if (input.size() < nRequiredElements)
```
schweitzpgi, line 322: Is `input.size()` more than `nRequiredElements` ok then?

runtime/cudaq/spin/spin_op.cpp
```diff
@@ -331,6 +331,29 @@ spin_op spin_op::from_word(const std::string &word) {
...
+
+  internalPauliWordStorage.clear();
+
+  internalPauliWordStorage.resize(num_terms());
+  std::size_t counter = 0;
+  for_each_term([&](spin_op &term) {
+    auto termStr = term.to_string(false);
+    internalPauliWordStorage[counter].term = new char[termStr.length() + 1];
+    std::strncpy(const_cast<char *>(internalPauliWordStorage[counter].term),
+                 termStr.c_str(), termStr.size() + 1);
```
schweitzpgi, line 351: This seems like low-level C code. Would it be possible to leverage STL like `std::string` here?

test/AST-Quake/cudaq-pauli-word.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+__qpu__ void test0(double theta, std::vector<cudaq::pauli_word> paulis) {
+  cudaq::qvector q(2);
+  for (auto &p : paulis)
+    exp_pauli(theta, q, p);
+}
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__function_test0._Z5test0dSt6vectorIN5cudaq10pauli_wordESaIS1_EE(
+// CHECK-SAME:                                                                                                %[[VAL_0:.*]]: f64,
+// CHECK-SAME:                                                                                                %[[VAL_1:.*]]: !cc.stdvec<!quake.pauli_word>) attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
```
schweitzpgi, line 21:
```suggestion
// CHECK-SAME:   %[[VAL_0:.*]]: f64,
// CHECK-SAME:   %[[VAL_1:.*]]: !cc.stdvec<!quake.pauli_word>) attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
```

test/AST-Quake/cudaq-pauli-word.cpp
```diff
@@ -0,0 +1,62 @@
...
+// CHECK:         }
+
+__qpu__ void test1(double theta, cudaq::qview<> q, cudaq::pauli_word term) {
+  exp_pauli(theta, q, term);
+}
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__function_test1._Z5test1dN5cudaq5qviewILm2EEENS_10pauli_wordE(
+// CHECK-SAME:                                                                                              %[[VAL_0:.*]]: f64,
+// CHECK-SAME:                                                                                              %[[VAL_1:.*]]: !quake.veq<?>,
+// CHECK-SAME:                                                                                              %[[VAL_2:.*]]: !quake.pauli_word) attributes {"cudaq-kernel", no_this} {
```
schweitzpgi, line 56: More unneeded whitespace added by the script.

PR #1147, opened 2024-01-29
Fixes for the lower to QIR pass.

schweitzpgi:
- Splits subgraph matching from the conversion pass. The conversion rewriter doesn't really let one do both at the same time given the way it traversing the IR.
  - Add a new codegen dialect. Only used in codegen. Part of the previous bullet. The dialect will be the logical point to add any future combination ops.
  - Removes the extra canonicalizer pass

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -72,7 +72,9 @@ def quake_AllocaOp : QuakeOp<"alloca", [MemoryEffects<[MemAlloc, MemWrite]>]> {
   }];
 
   let arguments = (ins Optional<AnySignlessInteger>:$size);
-  let results = (outs AnyRefType:$ref_or_vec);
+  let results = (outs
+    AnyRefType:$ref_or_vec
+  );
 
   let builders = [
     OpBuilder<(ins ), [{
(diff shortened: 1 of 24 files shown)


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -92,6 +94,14 @@ def quake_AllocaOp : QuakeOp<"alloca", [MemoryEffects<[MemAlloc, MemWrite]>]> {
 
   let hasCanonicalizer = 1;
   let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    bool hasInitializedState() {
+      auto *self = getOperation();
+      return self->hasOneUse() &&
```
amccaskey, line 101: Is it possible for the Deallocation pass to have added an extra user to this veq?

schweitzpgi: Well, it's theoretically possible, but it should not happen.

This updates (almost) all the passes so that they do not naively slap the dealloc in the wrong place. (There is one more case I'll fix in another PR.) It also adds a check to the verifier so that if a pass does do the wrong thing, the compiler raises an internal error.

amccaskey (approved):

PR #1150, opened 2024-01-30
Restore build_info.txt in Docker images

bmhowe23:
I believe PR #1010 accidentally got rid of the build_info.txt file in our Docker images. This PR brings it back (or at least the `source-sha` portion of that file, which is used in the nightly integration tests [[logs to error](https://github.com/NVIDIA/cuda-quantum/actions/runs/7720616344/job/21045804379)]).

### DIFF
--- a/.github/workflows/docker_images.yml
+++ b/.github/workflows/docker_images.yml
@@ -310,6 +310,7 @@ jobs:
           build-args: |
             base_image=${{ steps.prereqs.outputs.base_image }}
             install="CMAKE_BUILD_TYPE=Debug"
+            git_source_sha=${{ github.sha }}
           tags: ${{ steps.metadata.outputs.tags }}
           labels: ${{ steps.metadata.outputs.labels }}
           platforms: ${{ inputs.platforms }}
@@ -511,6 +512,7 @@ jobs:
           build-args: |
             base_image=${{ steps.prereqs.outputs.devdeps_image }}
             install="CMAKE_BUILD_TYPE=Release CUDA_QUANTUM_VERSION=${{ steps.prereqs.outputs.image_tag }}"
+            git_source_sha=${{ github.sha }}
           tags: ${{ steps.cudaqdev_metadata.outputs.tags }}
           labels: ${{ steps.cudaqdev_metadata.outputs.labels }}
           platforms: ${{ inputs.platforms }}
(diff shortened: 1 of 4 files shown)


docker/release/cudaq.Dockerfile
```diff
@@ -88,15 +88,17 @@ RUN echo "$CUDA_QUANTUM_PATH" > /usr/local/lib/python$(python --version | egrep
 
 # Some tools related to shell handling.
 
+ARG git_source_sha=xxxxxxxx
```
bettinaheim, line 91: The source code is coming from the cudaqdev_image (cuda-quantum-dev), and it hence feels odd to pass that separately here. I would instead suggest to add the build_info.txt to the cuda-quantum-dev image (in $CUDAQ_INSTALL_PREFIX) and then it will automatically get copied over here.

Edit: I see that you are already creating this file in the dev image - why would you let it be overwritten here? It can only either become inaccurate or remain the same.

docker/release/cudaq.Dockerfile
```diff
@@ -88,15 +88,17 @@ RUN echo "$CUDA_QUANTUM_PATH" > /usr/local/lib/python$(python --version | egrep
...
 ARG COPYRIGHT_NOTICE="=========================\n\
    NVIDIA CUDA Quantum   \n\
 =========================\n\n\
 Version: ${CUDA_QUANTUM_VERSION}\n\n\
-Copyright (c) 2023 NVIDIA Corporation & Affiliates \n\
+Copyright (c) 2024 NVIDIA Corporation & Affiliates \n\
```
bettinaheim, line 96: Good find. :)

.github/workflows/docker_images.yml
```diff
@@ -564,6 +566,7 @@ jobs:
             cudaqdev_image=${{ steps.prereqs.outputs.dev_image_name }}@${{ steps.release_build.outputs.digest }}
             base_image=${{ steps.prereqs.outputs.base_image }}
             release_version=${{ steps.prereqs.outputs.image_tag }}
+            git_source_sha=${{ github.sha }}
```
bettinaheim, line 569:
```suggestion
```

docker/release/cudaq.Dockerfile
bmhowe23: I wasn't sure if it would automatically get copied over or not. I will remove it from here.

.github/workflows/docker_images.yml
bmhowe23: Updated in d0deb3651

docker/release/cudaq.Dockerfile
bmhowe23: Updated in d0deb3651

bettinaheim (approved):

PR #1151, opened 2024-01-31
Move our new LLVM JIT code into cudaq-mlir-runtime instead of llvm_jit

bmhowe23:
This LLVM JIT code is quite similar to the rest of the code in cudaq-mlir-runtime, so I think it makes sense to group them together. This would help consolidate some of my yet-to-be-published prototyping code with Thien's code for the remote REST server.

The only functional change worth noting is that I had to change some `cudaq::info` log messages to `llvm::dbgs()`. Hopefully that is ok, but I'm open to other ideas if it is not.


runtime/common/CMakeLists.txt
```diff
@@ -105,8 +105,9 @@ install(EXPORT cudaq-common-targets
 get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
 get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 
-add_library(cudaq-mlir-runtime SHARED RuntimeMLIR.cpp)
+add_library(cudaq-mlir-runtime SHARED RuntimeMLIR.cpp JIT.cpp)
```
khalatepradnya, line 108: Not sure, but wondering if this has implications on LLVM static initialization issue since the `jit_util` depended on LLVM libraries.

bmhowe23: I think cudaq-mlir-runtime depends on the same libraries, so hopefully it isn't any worse than it already was? (But I don't know a good way to test it, so I'm open to other ideas, too.)

bmhowe23: @amccaskey can confirm, but I think this should not be any worse than it currently is.

khalatepradnya: Agreed (from the build experiments of past two days).

1tnguyen (approved):
LGTM 👍 Thanks!

Issue #1096, opened 2024-01-17
ImportError: Invalid simulator requested: custatevec_fp32


sirgeorgesawcon:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Since `cuda-quantum` is only available for Linux OS, I tried using it on Google Colab. 
For CPU runtime, I ran the code:

```python
%pip install cuda-quantum
```

It installed perfectly.

Then running the code 
```python
import cudaq

# We begin by defining the `Kernel` that we will construct our
# program with.
kernel = cudaq.make_kernel()

# Next, we can allocate qubits to the kernel via `qalloc(qubit_count)`.
# An empty call to `qalloc` will return a single qubit.
qubit = kernel.qalloc()

# Now we can begin adding instructions to apply to this qubit!
# Here we'll just add every non-parameterized
# single qubit gate that is supported by CUDA Quantum.
kernel.h(qubit)
kernel.x(qubit)
kernel.y(qubit)
kernel.z(qubit)
kernel.t(qubit)
kernel.s(qubit)

# Next, we add a measurement to the kernel so that we can sample
# the measurement results on our simulator!
kernel.mz(qubit)

# Finally, we can execute this kernel on the state vector simulator
# by calling `cudaq.sample`. This will execute the provided kernel
# `shots_count` number of times and return the sampled distribution
# as a `cudaq.SampleResult` dictionary.
result = cudaq.sample(kernel)

# Now let's take a look at the `SampleResult` we've gotten back!
print(result)  # or result.dump()
```

Also gave the desired output.

But when I changed the runtime to GPU, and installed the `cuda-quantum` again. This time running the code

```python
import cudaq
```

gave me this error:
```

ImportError: Invalid simulator requested: custatevec_fp32
```

The full error being

```
ImportError                               Traceback (most recent call last)

[<ipython-input-3-a2e13d3e4800>](https://localhost:8080/#) in <cell line: 1>()
----> 1 import cudaq
      2 
      3 # We begin by defining the `Kernel` that we will construct our
      4 # program with.
      5 kernel = cudaq.make_kernel()

[...]

PR #1152, opened 2024-01-31
Don't throw fatal exception during 'import cudaq' if missing dependencies

bmhowe23:
If the user has GPUs but doesn't have all the right dependencies installed, we should not throw fatal exceptions in that case. Rather, we should fallback on qpp-cpu and log an `info` message that we are not selecting the `nvidia` target because dependencies are missing.

Helps mitigate the impact of missing dependencies as encountered in #1096.

Testing this in CI is difficult so I did standalone testing as follows:

First perform the following to force missing dependencies:
```bash
$ mv /usr/local/cuda/targets/x86_64-linux/lib/libcublas.so.11 /usr/local/cuda/targets/x86_64-linux/lib/libcublas.so.11.ignore
```

Without this change, it behaves as follows:
```bash
$ python3 -c "import cudaq"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/cudaq/cudaq/__init__.py", line 28, in <module>
    from ._pycudaq import *
ImportError: Invalid simulator requested: custatevec_fp32
```

With this change, it behaves as follows:
```bash
$ python3 -c "import cudaq"
<no errors>
```

### DIFF
--- a/python/utils/LinkedLibraryHolder.cpp
+++ b/python/utils/LinkedLibraryHolder.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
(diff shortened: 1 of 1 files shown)


khalatepradnya (approved):
👍

PR #1154, opened 2024-01-31
Triage tests

### LINKED ISSUE #1102: Failing unit tests when run with C++17 outside the Docker image
Some of the llvm-lit tests are currently failing when they are run in almalinux with C++17. 
Concretely, the failing tests are: 
- [x] [AST-Quake/lambda_variable.cpp](https://github.com/NVIDIA/cuda-quantum/pull/1101/files#diff-2832ba15fa7ec44914ad37a192127fb081a70cda15f5d3c6839ef37d179401b1) (tests source code contains C++20 concepts and hence won't/shouldn't compile with C++17 flag)
- [ ] AST-error/negation_error.cpp, AST-error/kernel_invalid_argument-2.cpp (a warning for using a C++20 feature pops up that could be suppressed by default?)
- [x] NVQPP/mapping_test-1.cpp (fixed)
- [x] NVQPP/custom_pass.cpp and NVQPP/custom_pass.qke (won't be supported with a different build configuration than we use for CUDA Quantum itself, documented as such)

The custom pass setup likely needs a more general update, but the other two tests a bug fixes.

schweitzpgi:
Closes #1102.

  - lambda_variable.cpp will never compile as C++17.
  - Testing errors: one cannot expect identical error messages between different C++ standard settings.
  - Syntax in these tests uses C++20 features and will produce warnings.
  - As the runtime is compiled with C++20, a plugin pass compiled with C++17 may not behave correctly.

### DIFF
--- a/test/NVQPP/custom_pass.cpp
+++ b/test/NVQPP/custom_pass.cpp
@@ -5,11 +5,12 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
+
+// REQUIRES: c++20
+
 // clang-format off
 // RUN: nvq++ -std=c++17 --enable-mlir --opt-plugin %cudaq_lib_dir/CustomPassPlugin.so --opt-pass 'func.func(cudaq-custom-pass)' %s -o %t && %t | FileCheck %s
-// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
-// RUN:   nvq++ --enable-mlir --opt-plugin %cudaq_lib_dir/CustomPassPlugin.so --opt-pass 'func.func(cudaq-custom-pass)'  %s -o %t && %t | FileCheck %s; \
-// RUN: fi
+// RUN: nvq++ --enable-mlir --opt-plugin %cudaq_lib_dir/CustomPassPlugin.so --opt-pass 'func.func(cudaq-custom-pass)'  %s -o %t && %t | FileCheck %s
 // clang-format on
 
 #include <cudaq.h>
--- a/test/NVQPP/mapping_test-1.cpp
+++ b/test/NVQPP/mapping_test-1.cpp
@@ -6,11 +6,9 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// REQUIRES: c++20
-
-// RUN: nvq++ %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t 2> %t.code | FileCheck %s && FileCheck --check-prefix=QUAKE %s < %t.code && rm %t.code
-// RUN: nvq++ %s -o %t --target iqm --iqm-machine Adonis --mapping-file "%p/../Supplemental/Adonis Variant.txt" --emulate && %t
-// RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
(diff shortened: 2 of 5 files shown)


test/NVQPP/mapping_test-1.cpp
```diff
@@ -6,9 +6,11 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ %cpp_std -v %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t |& FileCheck %s
+// REQUIRES: c++20
+
+// RUN: nvq++ %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t | FileCheck %s
+// RUN: nvq++ %s -o %t --target iqm --iqm-machine Adonis --mapping-file "%p/../Supplemental/Adonis Variant.txt" --emulate && %t
 // RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
```
bmhowe23, line 13: How does this required c++20 and yet still have a `-std=c++17` version of this test?

schweitzpgi: There isn't a link to the failures in the issue.

This test will compile using the `-std=c++17` option on a correctly configured machine with a C++20 compiler. So that is what is being tested.

Whether the binary will execute on a malconfigured system is a different question.

test/NVQPP/mapping_test-1.cpp
```diff
@@ -6,9 +6,11 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ %cpp_std -v %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t |& FileCheck %s
+// REQUIRES: c++20
```
bettinaheim, line 9: There is clearly an issue with using C++17 here - I believe the error, if I remember correctly - was a failure to lower. 
There might even be the suggestion in the message to file a bug report. This needs to be fixed, not permanently disabled.

test/NVQPP/custom_pass.cpp
```diff
@@ -5,11 +5,12 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
+
+// REQUIRES: c++20
```
bettinaheim, line 9: Here we need to make a judgement call regarding the requirements for defining an using a custom pass plugin.
Does the fact that we compile CUDA Quantum with C++20 mean anyone building a pass plugin also needs to compile with the same C++ standard? Our simulators (that are loaded dynamically), for example are compiled with C++17, so it is not clear to me why this should be a requirement. If it is a requirement, we need to document this. If not, this test needs to be fixed and reenabled.

bettinaheim (changes requested):
See my comments on two of the permanently disabled tests.

test/NVQPP/mapping_test-1.cpp
bettinaheim: Interesting enough, this test now actually passes and no longer fails with C++17 - one of the more recent PRs seems to have fixed it.

schweitzpgi: https://github.com/NVIDIA/cuda-quantum/pull/1154/commits/62a59874eabbc8442a2cdc5c22a21d5096b0356b

bettinaheim (approved):
side note: the two AST-error tests that are marked as requiring C++20 here produce the same output with C++17 up to a warning about using C++20 features.

Issue #1136, opened 2024-01-25
Tensornet backend scratch space allocator is prone to race condition


1tnguyen:
This code block to allocate scratch space for the `tensornet` backend is prone to race condition:

https://github.com/NVIDIA/cuda-quantum/blob/5ece3200511c1347ad3da5e964a26d15f05ada95/runtime/nvqir/cutensornet/tensornet_utils.h#L70-L73

In particular, according to the [docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g376b97f5ab20321ca46f7cfa9511b978) of `cudaMemGetInfo`:

> In a multi-tenet situation, free estimate returned is prone to race condition where a new allocation/free done by a different process or a different thread in the same process between the time when free memory was estimated and reported, will result in deviation in free value reported and actual free memory. 

For example, two `tensornet` simulator instances may call this at an exact same time, hence wrongly estimate the scratch memory to use.

1tnguyen:
Should be closed by [#1158](https://github.com/NVIDIA/cuda-quantum/pull/1158)

PR #1158, opened 2024-02-01
Fixes for the `remote-mqpu` platform

1tnguyen:
### Description

(1) Clean shutdown behavior

- Client side: enforce cleanup order: qpus first then any auto-launch server instances. This prevents some stray errors during shut down. For example, the server process is shut down before the client, which causes the client to report an error (lost connection to the server) momentarily before it being destructed.

- Server app: make sure that we always have a loaded simulator in the runtime to prevent segfault during shutdown when the runtime would try to find one.

(2) Harden the `tensornet` scratch memory allocation against race condition.
Related to https://github.com/NVIDIA/cuda-quantum/issues/1136


bettinaheim (approved):
lgtm

PR #1161, opened 2024-02-02
Enable LLVM's compiler-rt in our devdeps images

bmhowe23:


scripts/build_llvm.sh
```diff
@@ -182,3 +191,12 @@ else
   cp bin/llvm-lit "$LLVM_INSTALL_PREFIX/bin/"
   cd "$working_dir" && echo "Installed llvm build in directory: $LLVM_INSTALL_PREFIX"
 fi
+
+if $compiler_rt; then
+  cd $llvm_source/build && ninja runtimes && ninja install-runtimes
```
bettinaheim, line 196: Out of curiosity: what exactly does the initial `ninja install-distribution-stripped` or `ninja install` build/install, and what only built/installed here?

scripts/build_llvm.sh
```diff
@@ -182,3 +191,12 @@ else
...
+  status=$?
+  if [ "$status" = "" ] || [ ! "$status" -eq "0" ]; then
+    echo "Build failed. Please check the files in the `pwd`/logs directory."
+    cd "$working_dir" && if $is_sourced; then return 1; else exit 1; fi
+  fi
```
bettinaheim, line 201:
```suggestion
  else
    cd "$working_dir" && echo "Successfully added runtime components."
  fi
```
Please cd back to the working dir - nothing more annoying that a script that dumps me in a "random" directory once it is done. ;)

bettinaheim (approved):

scripts/build_llvm.sh
bmhowe23: `ninja runtimes` builds the runtimes specified at configuration time (`-DLLVM_ENABLE_RUNTIMES`) using the just-built compiler that was just built with `ninja build`. And `ninja install-runtimes` simply installs the runtime files into `/opt/llvm`.

[This gist](https://gist.github.com/bmhowe23/67292ab39ac1109bc130b4c70f566d8c) shows the build logs for those steps if you want to see them. (Note they appear in reverse order in the gist.)

The TL;DR of those logs is that new files are added to `/opt/llvm/lib/clang/16/include` and `/opt/llvm/lib/clang/16/lib`. Those new libs are ~57MB according to the `du` command.

Re: `install-distribution` vs `install`, I wasn't too familiar with the distinction, but I did find [this webpage](https://llvm.org/docs/BuildingADistribution.html#difference-between-install-and-install-distribution) explaining the distinction. (I didn't change that portion, so I didn't do a deep dive on that.) That webpage also has a section on `LLVM_ENABLE_RUNTIMES`, too.

The reason I did `$compiler_rt` in its own command-line flag is that I didn't want to build the compiler-rt binaries for our first-pass of the bootstrapped LLVM build (which is only applicable when our `toolchain=llvm`).

bettinaheim: I think we might want to build the compiler rts in the first pass as well when bootstrapping to get rid of all libstdc++ dependencies in favor of libc++. However, that's a change for another time.

PR #1169, opened 2024-02-05
[rt] Remove unused header.

boschmitt:
### Description
~Add the missing `cstdint` header~ and remove the unused `vector` one from `JIT.h`.

Edit: Seems like someone else fixed the missing header issue on a previous PR.


bmhowe23:
The remaining change was incorporated in https://github.com/NVIDIA/cuda-quantum/commit/b56fab0b237a7e5d829471af778e9935533ccbda, so I suspect this PR can be closed now.

PR #1170, opened 2024-02-05
Fix default llvm path in `build_cudaq.sh` script.

boschmitt:
### Description
The `build_llvm.sh` script uses the `$HOME/.llmv` as default LLVM install prefix. Therefore, `build_cudaq.sh` should do the same, otherwise the combination of both scripts won't work by default.

### DIFF
--- a/scripts/build_cudaq.sh
+++ b/scripts/build_cudaq.sh
@@ -34,7 +34,7 @@
 # By default, the CUDA Quantum is done with warnings-as-errors turned on.
 # You can turn this setting off by defining the environment variable CUDAQ_WERROR=OFF.
 
-LLVM_INSTALL_PREFIX=${LLVM_INSTALL_PREFIX:-/opt/llvm}
+LLVM_INSTALL_PREFIX=${LLVM_INSTALL_PREFIX:-$HOME/.llvm}
 CUQUANTUM_INSTALL_PREFIX=${CUQUANTUM_INSTALL_PREFIX:-/opt/nvidia/cuquantum}
 CUDAQ_INSTALL_PREFIX=${CUDAQ_INSTALL_PREFIX:-"$HOME/.cudaq"}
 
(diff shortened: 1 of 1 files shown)


scripts/build_cudaq.sh
```diff
@@ -34,7 +34,7 @@
 # By default, the CUDA Quantum is done with warnings-as-errors turned on.
 # You can turn this setting off by defining the environment variable CUDAQ_WERROR=OFF.
 
-LLVM_INSTALL_PREFIX=${LLVM_INSTALL_PREFIX:-/opt/llvm}
+LLVM_INSTALL_PREFIX=${LLVM_INSTALL_PREFIX:-$HOME/.llvm}
```
schweitzpgi, line 37: I think the install should be to `/opt/llvm`. It seems like that is hardcoded in other places.

There is also a copy of the sources made in ${HOME}/.llvm-project, which bypasses the subproject at tpls/llvm.

boschmitt: I'm unsure if a user will always have the correct privileges to install in `/opt/`. As I said in the comment, without any options `scripts/build_llvm.sh` will install llvm in `$HOME/.llvm`. Then `scripts/build_cudaq.sh` will look for it in `/opt/llvm`. Thus we end up with the situation the following not working:

scripts/build_cudaq.sh
```diff
@@ -118,6 +118,7 @@ fi
 echo "Preparing CUDA Quantum build with LLVM installation in $LLVM_INSTALL_PREFIX..."
 cmake_args="-G Ninja "$repo_root" \
   -DCMAKE_INSTALL_PREFIX="$CUDAQ_INSTALL_PREFIX" \
+  -DLLVM_DIR="$LLVM_INSTALL_PREFIX/lib/cmake/llvm" \
```
bettinaheim, line 121: Handled correctly in the CMake files without this.
```suggestion
```

scripts/build_cudaq.sh
bettinaheim: 1) Invoking the build_cudaq.sh script will invoke build_llvm correctly if you allow it to install prerequisites (-u). You hence should not invoke both scripts manually.
2) If you are choosing to build individual steps manually, you should set the corresponding *_INSTALL_PREFIX variable(s). 
The setup is intentionally configured so that the prebuilt packages our images come with are under /opt, whereas local modifications will by default not overwrite this. I don't mind changing the default paths here, since they don't matter for the overall build. 
3) While there are default values for convenience, we no longer have anything "hardcoded" in the sense that other files rely on specific locations. The only exception is CUDA itself, where the path /usr/local/cuda is fixed (it's not configurable, so it necessarily is hardcoded).

bettinaheim: The build_cudaq script will automatically invoke the build_llvm script if you run it with -u and everything will work. 
The setup with different paths is intended for the recommended way of developing, where you work inside the dev container, but want a side-by-side installation of a different llvm version to update llvm.

PR #1171, opened 2024-02-05
Improve handling of quantum library calls in kernel function bodies

amccaskey:
Uniformly support cases like the following for both eager and mlir mode 
```python
import cudaq
from cudaq.lib import fermionic_swap 
...
def useFSwap():
   ... 
   fermionic_swap(...)
```
and 
```python
import cudaq
... 
def useFSwap():
    ...  
    cudaq.lib.fermionic_swap(...)
```

Clean up and unify the tests. 

Fix observe_async for eager mode kernels


python/tests/eager/test_library_kernels.py
```diff
@@ -0,0 +1,69 @@
+# ============================================================================ #
```
khalatepradnya, line 1: Can you run the whole Python test suite with same named test files? I recall that since we don't have `__init.py__` files in the individual test directories, they need to be uniquely named. So, I've renamed those under 'eager' with suffix `_eager`

khalatepradnya (commented):
Code changes look good to me. 
Couple of Qs on tests.

python/tests/eager/test_library_kernels.py
amccaskey: I didn't know that - I'll update the file name.

python/tests/ast/test_library_kernels.py
```diff
@@ -12,14 +12,9 @@
...
-# [RFC]:
-# FIXME: Why does having this here instead of inside the respective test cause issues?
-# cudaq.enable_jit()
-# from cudaq.lib import fermionic_swap, givens
-
-
 def test_fswap_lib_kernel():
     cudaq.enable_jit()
+
     from cudaq.lib import fermionic_swap
```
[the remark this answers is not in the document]
amccaskey: I would like to keep it here in order to keep `enable_jit()` out of global scope. You could move this import to the top of the file, but you would also have to move the `enable_jit` call there and make sure `disable_jit` got called. 

To me this is ok - this JIT enable / disable functions are mainly developer functions, and in the future we'll default to JIT and the user will not even worry about these things. 

I just tested both ways locally, and what you suggest works the same. Just need to enable JIT before the import.

python/CMakeLists.txt
```diff
@@ -152,72 +152,5 @@ endif()
...
-  RESULT_VARIABLE PYTEST_result)
-execute_process(COMMAND ${Python_EXECUTABLE} -c "import numpy"
-  OUTPUT_VARIABLE NUMPY_output
-  ERROR_VARIABLE NUMPY_error
-  RESULT_VARIABLE NUMPY_result)
-
-if(CUDAQ_BUILD_TESTS)
-  if(${PYTEST_result} EQUAL 0 AND ${NUMPY_result} EQUAL 0)
-    message(STATUS "Pytest and Numpy found, building Python tests.")
-    add_subdirectory(tests)
```
khalatepradnya, line 219: `add_subdirectory(tests)` - Don't we need this for testing?

amccaskey: I was wondering that - it seems like the CI testing always directly calls `python3 -m pytest ....` so there is no need for the CTest targets, maybe I'm wrong?

python/cudaq/__init__.py
```diff
@@ -107,6 +107,7 @@ def disable_jit():
 # Expose chemistry domain functions
 from .domains import chemistry
 from .kernels import uccsd
+import lib
```
khalatepradnya, line 110: Importing as `from .lib import *` got all the ctests to work (including ` qpp_H2MoleculeTester.checkHamiltonian_Sampling`)

python/CMakeLists.txt
khalatepradnya: I see that we use the `CUDAQ_BUILD_TESTS` build argument to skip this target - which is used when Python wheel is built. Additionally, perhaps anyone directly building from source may find it useful (non-CI use case).

amccaskey: Sounds good, I'll put it back

python/tests/builder/test_NoiseModel.py
```diff
@@ -144,6 +145,7 @@ def test_phase_flip_simple():
     noisy_counts = cudaq.sample(kernel, noise_model=noise)
     got_zero_counts = noisy_counts["0"]
     assert got_zero_counts == want_counts
+    cudaq.reset_target
```
khalatepradnya, line 148:
```suggestion
    cudaq.reset_target()
```

python/tests/kernel/test_ir_operations.py
```diff
@@ -13,6 +13,14 @@
...
+@pytest.fixture(autouse=True)
+def do_something():
+    if os.getenv("CUDAQ_PYTEST_EAGER_MODE") == 'OFF':
+        cudaq.enable_jit()
+    yield
+
+    if cudaq.jit_enabled(): cudaq.__clearKernelRegistries()
+    cudaq.disable_jit()
 
 def test_synthesize():
```
khalatepradnya, line 25: Does the comment on line#26 still apply? (code on line#27 may not be needed)
Same with `jit=True` on line#37

python/cudaq/__init__.py
```diff
@@ -104,6 +105,10 @@ def disable_jit():
     PyKernelDecorator.globalJIT = False
 
 
+def jit_enabled():
```
khalatepradnya, line 108: Nit - rename to `is_jit_enabled()`

khalatepradnya (approved):

python/tests/kernel/test_ir_operations.py
amccaskey: I'll restructure this one to only run in MLIR mode.

python/cudaq/__init__.py
amccaskey: fixed

python/tests/kernel/test_library_kernels.py
```diff
@@ -39,26 +41,37 @@ def bar(angle: float):
...
 
-def test_givens_lib_kernel():
-    cudaq.enable_jit()
     from cudaq.lib import givens
 
     angle = 0.2
     c = np.cos(angle)
     si = np.sin(angle)
 
+    # PYTEST Error, baz fails
```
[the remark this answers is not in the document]
amccaskey: This has been fixed.

khalatepradnya (changes requested):
Following required for 'release' build. Should fix 'docs' failure too.

```
diff --git a/python/CMakeLists.txt b/python/CMakeLists.txt
index 1c7b57268..64368d92b 100644
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -155,6 +155,16 @@ set_property(TARGET CUDAQuantumPythonModules.extension._quakeDialects.dso
 add_subdirectory(runtime/cudaq/distributed)
 add_subdirectory(runtime/cudaq/domains/plugins)

+if (NOT SKBUILD)
+  SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:$ORIGIN/../../lib:$ORIGIN/../../lib/plugins")
+else()
+  SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:$ORIGIN/../cuda_quantum.libs")
+endif()
+
+if (NOT SKBUILD)
+  install(DIRECTORY cudaq DESTINATION .)
+endif()
+
 # Do we have pytest and numpy?
 execute_process(COMMAND ${Python_EXECUTABLE} -m pytest --version
   OUTPUT_VARIABLE PYTEST_output
```

khalatepradnya (approved):

bmhowe23 (approved):
I reviewed .github/workflows/test_in_devenv.yml and it LGTM.

PR #1173, opened 2024-02-05
Fix incorrect warning for IonQ command line parameters

bmhowe23:
Base64 encoding was incorrectly triggering a warning. This change fixes that by performing base64 decoding on the appropriate command-line parameters.


1tnguyen (approved):
lgtm

PR #1176, opened 2024-02-06
[rt] Minor clean up of `BasicExecutionManager`

boschmitt:
### Description
Minor clean up to the `BasicExecutionManager`:
* Remove unused headers
* Prefer the use of `std::vector` instead of `std::queue` and `std::stack`. (This simplifies the code and avoid the use of unnecessarily involved data structures.)


amccaskey (approved):
much cleaner. thanks @boschmitt 

+1

PR #1178, opened 2024-02-06
[rt] Add `Trace` class

boschmitt:
### Description
The execution of a kernel in the `tracer` context will produce a `Trace` object: a netlist representation of the executed computation as seen by the execution manager. This object can then be used for analysis, e.g., `cudaq::estimate_resoruces` can compute number of operations or a circuit be drawn (this functionality will be added by a later PR).

With this `Trace` class, we can improve the `cudaq::Resource` one. However, that abstraction is leaky and would require braking changes that are out-of-scope for the present PR.

Depends on #1176

### DIFF
--- a/runtime/common/Trace.h
+++ b/runtime/common/Trace.h
@@ -1,0 +1,50 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include "cudaq/qis/execution_manager.h"
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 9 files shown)


runtime/common/Trace.h
```diff
@@ -0,0 +1,50 @@
...
+/// of instructions on qudits). Since the execution manager cannot "see" control
+/// flow, the trace of a kernel with control flow represents a single execution
+/// path, and thus two calls to the same kernel might produce traces.
+class Trace {
+public:
+  struct Instruction {
+    std::string name;
+    std::vector<double> params;
+    std::vector<std::size_t> controls;
+    std::vector<std::size_t> targets;
```
amccaskey, line 27: do you think we'll want to retain `QuditInfo` here? So the `Instruction` will also know the number of qudit levels?

boschmitt: The two analyses that we do on top of it are agnostic to the number of levels---both only care about unique identifier. I chose this way mainly because it is how `estimate_resources` currently work. It should not be hard to modify to retain it.

amccaskey (approved):
+1

Issue #1157, opened 2024-02-01
[experimental/python] Improvements, gaps and fixes: Part#1


khalatepradnya:
Gaps in features and enhancement:

Must have:
- [x] Build Python wheel on arm64 platform
- [x] Enable remote simulator target (and corresponding tests)
- [x] Enhance tests for remote backend with new Python APIs
- [x] Port `vqe` module
- [x] Port `parallel` module
- [x] Port `chemistry` module
- [x] Fix `State`, `get_state_async`, `AsyncStateResult`
- [x] Enable `OptimizationResult`
- [x] Enable `broadcast` in `mpi` 
- [x] Implement `observe_async` in eager mode
- [x] Bug in AST compilation where `register_name='name'` is ignored
- [x] bug with ordering of kernel definitions and kernel composition
- [x] bug with transitive kernel dependencies
- [x] sample_async eager mode
- [x] check control self.popValue() calls (bug with too many calls to popValue fixed for one scenario, could be others)
- [x] vqe for eager / mlir mode kernels
- [x] drop `qlist`
-  Implement `from_state`: Will be addressed separately 
- [x] `range(q, p, -1), p == 0`
- [x]  Add Python binding for `NvcfSimulatorQPU` similar to `PyRemoteSimulatorQPU`

Nice to have:
- Enable building and usage of old Python bindings and new MLIR bindings side-by-side (fix for LLVM static initialization issue)

khalatepradnya:
Issues related with `State` will be addressed by https://github.com/NVIDIA/cuda-quantum/pull/1123

khalatepradnya:
- `parallel` fixed in commit [d5459b2](https://github.com/NVIDIA/cuda-quantum/pull/1077/commits/d5459b227ca941fe0584d66dfeb1a1e81a43f41d) on PR #1077
- `vqe` fixed in commit [4bf7a4a](https://github.com/NVIDIA/cuda-quantum/pull/1077/commits/4bf7a4a7b65c6425201b3dca4292330cfd836d86) on PR #1077

khalatepradnya:
- `chemistry` fixed in commit [c635d9f](https://github.com/NVIDIA/cuda-quantum/pull/1077/commits/c635d9f2a75f6a4605870a21a6ad75e567bd089c) on PR https://github.com/NVIDIA/cuda-quantum/pull/1077

khalatepradnya:
[...]

PR #1182, opened 2024-02-06
[experimental/python] Added 'remote' tests for new modes in Python

khalatepradnya:
* 'remote' tests in 'ast' (AKA 'mlir') mode

Addresses one issue in https://github.com/NVIDIA/cuda-quantum/issues/1157


python/tests/remote/test_remote_platform.py
```diff
@@ -19,81 +21,120 @@ def startUpMockServer():
...
     numQpus = target.num_qpus()
     assert numQpus == num_qpus
 
 
+def check_sample(entity):
+
+    def check_basic(counts):
+        assert len(counts) == 2
+        assert "00" in counts
+        assert "11" in counts
```
bettinaheim, line 46: assert 01 and 10 are not in counts?

bettinaheim (approved):

python/tests/remote/test_remote_platform.py
amccaskey: I think that is covered by len(counts) == 2

amccaskey (approved):

PR #1184, opened 2024-02-06
[experimental/python] Reenable and fix the arm64 parts of the workflows.

schweitzpgi:
Don't bother reviewing this for now.

### DIFF
--- a/docker/release/cudaq.wheel.Dockerfile
+++ b/docker/release/cudaq.wheel.Dockerfile
@@ -25,7 +25,8 @@ ADD "$workspace" "$destination"
 
 ARG python_version=3.10
 RUN echo "Building MLIR bindings for python${python_version}" \
-    && python${python_version} -m pip install --no-cache-dir numpy \
+    && python${python_version} -m pip uninstall numpy pytest \
+    && python${python_version} -m pip install --no-cache-dir numpy pytest fastapi uvicorn llvmlite \
     && rm -rf "$LLVM_INSTALL_PREFIX/src" "$LLVM_INSTALL_PREFIX/python_packages" \
     && export Python3_EXECUTABLE="$(which python${python_version})" \
     && LLVM_PROJECTS='clang;mlir;python-bindings' \
--- a/docker/test/wheels/debian.Dockerfile
+++ b/docker/test/wheels/debian.Dockerfile
@@ -24,8 +24,8 @@ ENV VIRTUAL_ENV=/opt/venv
 RUN python${python_version} -m venv "$VIRTUAL_ENV"
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN if [ -n "$preinstalled_modules" ]; then \
-        python${python_version} -m pip uninstall $preinstalled_modules \
-        python${python_version} -m pip install $preinstalled_modules \
+        python${python_version} -m pip uninstall $preinstalled_modules ;\
+        python${python_version} -m pip install $preinstalled_modules ;\
     fi
 
 ARG optional_dependencies=
--- a/docker/test/wheels/debian.Dockerfile
+++ b/docker/test/wheels/debian.Dockerfile
@@ -24,7 +24,8 @@ ENV VIRTUAL_ENV=/opt/venv
 RUN python${python_version} -m venv "$VIRTUAL_ENV"
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN if [ -n "$preinstalled_modules" ]; then \
-        echo $preinstalled_modules | xargs python${python_version} -m pip install; \
+        python${python_version} -m pip uninstall $preinstalled_modules \
+        python${python_version} -m pip install $preinstalled_modules \
     fi
 
 ARG optional_dependencies=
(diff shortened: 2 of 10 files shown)


docker/release/cudaq.wheel.Dockerfile
```diff
@@ -25,7 +25,8 @@ ADD "$workspace" "$destination"
 
 ARG python_version=3.10
 RUN echo "Building MLIR bindings for python${python_version}" \
-    && python${python_version} -m pip install --no-cache-dir numpy \
+    && python${python_version} -m pip uninstall numpy pytest \
+    && python${python_version} -m pip install --no-cache-dir numpy pytest fastapi uvicorn llvmlite \
```
bettinaheim, line 29:
```suggestion
    && python${python_version} -m pip install --no-cache-dir numpy \
```
As far as we had it documented, the other package were only used for tests, and hence should not be needed because this file only produces the wheels. Tests are run separately in a clean environment (i.e. not the build environment). It could be, of course, that for some reason one of these fixes a missing dependency on arm that the llvm build does not properly complain about?

Also, the two commands above don't do anything after removing the unneeded packages. I think you possibly meant to uninstall numpy from the default python version instead (any pip packages we install in devdeps.manylinux.Dockerfile are already uninstalled again there).

docker/test/wheels/debian.Dockerfile
```diff
@@ -24,7 +24,8 @@ ENV VIRTUAL_ENV=/opt/venv
 RUN python${python_version} -m venv "$VIRTUAL_ENV"
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN if [ -n "$preinstalled_modules" ]; then \
-        echo $preinstalled_modules | xargs python${python_version} -m pip install; \
+        python${python_version} -m pip uninstall $preinstalled_modules ;\
```
bettinaheim, line 27: These files are used only for validation. Hence, anything in these files is what a user would need to do. 
These lines really don't do anything otherwise that would be a bug with pip.

scripts/build_llvm.sh
```diff
@@ -24,7 +24,7 @@
 LLVM_INSTALL_PREFIX=${LLVM_INSTALL_PREFIX:-$HOME/.llvm}
 LLVM_PROJECTS=${LLVM_PROJECTS:-'clang;lld;mlir;python-bindings'}
 PYBIND11_INSTALL_PREFIX=${PYBIND11_INSTALL_PREFIX:-/usr/local/pybind11}
-Python3_EXECUTABLE=${Python3_EXECUTABLE:-python3}
+Python3_EXECUTABLE=${Python3_EXECUTABLE:-python3.8}
```
bettinaheim, line 27: 3.8 is rare to be installed on any machine. Hence please keep it as the more general command. The only place where it should matter which python version is used is the build of the wheel itself, where this variable is already explicitly set. However, clearly there is some bug so if we need to nail down the python version for building the llvm parts that are - or rather should be! - *unrelated to python*, then we should do that be instead setting this environment variable in the devdeps.manylinux.Dockerfile.

scripts/build_llvm.sh
```diff
@@ -74,16 +74,13 @@ fi
 if [ "$llvm_source" = "" ]; then
   echo "Cloning LLVM submodule..."
   cd "$this_file_dir" && cd $(git rev-parse --show-toplevel)
-  llvm_source=~/.llvm-project
-  llvm_repo="$(git config --file=.gitmodules submodule.tpls/llvm.url)"
-  llvm_commit="$(git submodule | grep tpls/llvm | cut -c2- | cut -d ' ' -f1)"
-  git clone --filter=tree:0 "$llvm_repo" "$llvm_source"
-  cd "$llvm_source" && git checkout $llvm_commit
+  llvm_source=$(pwd)/tpls/llvm
+  git submodule update --init --recommend-shallow --single-branch tpls/llvm
```
bettinaheim, line 78: shallow doesn't work well to cherry-pick commits. `git clone --filter=tree:0` avoids the limitations of a shallow repo clone and is equally as fast.

docker/release/cudaq.wheel.Dockerfile
schweitzpgi: The uninstall followed by an install is effectively a _forced_ reinstall. `pip` doesn't have a reinstall option.

I don't think this is an issue in the script stack. Having these packages installed sideways can happen and was happening when I was experimenting.

scripts/build_llvm.sh
schweitzpgi: In my opinion, this is not the best way to do this. The llvm-project repo is massive and constantly cloning it live in a number of different CI tasks is not a good idea.

First, everything we do should use the llvm-project release/16.x branch (which is the SHA on tpls/llvm). This avoids any issues from using arbitrary llvm versions for disparate parts of our project. This never changes as we're stuck with the limitations of QIR on 16.x.

Second, the subprojects and other prerequisites should be downloaded as part of a standard "development" image build. This image would only be rebuilt when the prerequisites change. It should be a known, published image and not dependent on caches which may be cleared out from underneath the CI.

Finally, instead of cherry-picking SHAs from future branches like this, we should just have a diff file with the patches and apply those changes to the top of the branch. We already do this for pybind11. Applying the patch from a diff file avoids downloading an entire copy of llvm every time which saves bandwidth and time. It's also trivially reversible.

PR #1193, opened 2024-02-08
Installing cuquantum and cutensor as prereqs

bettinaheim:
Simplifying source build by automatically installing cuquantum/cutensor


bmhowe23 (approved):
LGTM (assuming it doesn't significantly change our installer size).

bettinaheim:
> LGTM (assuming it doesn't significantly change our installer size).

This doesn't change the installer content. It only changes whether the cuQuantum and cuTensor dependencies are downloaded automatically as part of building CUDA Quantum, or whether they are expected to be pre-installed on the system, like CUDA (there's not really a good way to download the complete CUDA compiler automatically for different systems).

PR #1196, opened 2024-02-08
[experimental/python] Build kernel call graph and add all dependent kernels

amccaskey:
Fix issue where all dependent kernels were not added to MLIR ModuleOp

```python
    @cudaq.kernel()
    def func0(q : cudaq.qubit):
        x(q)

    @cudaq.kernel()
    def func1(q: cudaq.qubit):
        func0(q)

    @cudaq.kernel
    def func2(q: cudaq.qubit):
        func1(q)

    @cudaq.kernel()
    def callMe():
        q = cudaq.qubit()
        func2(q)
```
@marwafar


khalatepradnya (approved):
👍

PR #1201, opened 2024-02-09
[rt] Unitary circuit drawing

boschmitt:
### Description
This PR adds the capability of drawing certain kernels: It should be restricted to unitary kernels, i.e., those that implement a unitary operator, and thus do not have measurements. This restrictions is not yet implemented because the drawer is based on the tracing capabilities of the execution manager, and the trace still allows measurements---but it shouldn't (a later PR should handle that).

Example:
```cpp
#include <cudaq.h>
#include <cudaq/algorithms/draw.h>
#include <iostream>

__qpu__ void foo() {
  using namespace cudaq;
  qubit q0;
  qubit q1;
  qubit q2;
  h(q0);
  x<ctrl>(q0, q1);
  z(q2);
  x<ctrl>(q0, q2, q1);
  r1(1.2, q0);
  swap(q0, q1);
}

int main() {
  // Get the string and print it:
  auto result = cudaq::draw(foo);
  std::cout << result;

  std::cout << '\n';
  // Print it directly to `std::cout`:
  cudaq::draw(std::cout, foo);

  return 0;
}
``` 

Outputs:
```
#  nvq++ draw_test.cpp
# ./a.out
     ╭───╮          ╭─────────╮   
q0 : ┤ h ├──●────●──┤ r1(1.2) ├─╳─
     ╰───╯╭─┴─╮╭─┴─╮╰─────────╯ │ 
q1 : ─────┤ x ├┤ x ├────────────╳─
     ╭───╮╰───╯╰─┬─╯              
q2 : ┤ z ├───────●────────────────
     ╰───╯                        

     ╭───╮          ╭─────────╮   
q0 : ┤ h ├──●────●──┤ r1(1.2) ├─╳─
     ╰───╯╭─┴─╮╭─┴─╮╰─────────╯ │ 
q1 : ─────┤ x ├┤ x ├────────────╳─
     ╭───╮╰───╯╰─┬─╯              
q2 : ┤ z ├───────●────────────────
     ╰───╯                        
```

TODOs:
- [x] Qubit labels (These are automatically generate as `q<index>`)
- [x] Testing


unittests/integration/draw_tester.cpp
```diff
@@ -0,0 +1,48 @@
...
+     ├───┤╭─┴─╮╰─────────────╯ │ │       │
+q1 : ┤ h ├┤ x ├────────────────┼─┤● swap ├
+     ├───┤├───┤                │ │       │
+q2 : ┤ h ├┤ z ├────────────────╳─┤       ├
+     ╰───╯╰───╯                  ╰───────╯
+)";
+  // clang-format on
+
+  expected_str = expected_str.substr(1);
+  std::string produced_str = cudaq::draw(kernel);
```
bettinaheim, line 46: It would be good to also have an example, an notebook, and documentation.

unittests/integration/draw_tester.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+CUDAQ_TEST(DrawTester, checkOps) {
+
+  auto kernel = []() __qpu__ {
+    cudaq::qvector q(3);
+    h(q);
+    x<cudaq::ctrl>(q[0], q[1]);
+    z(q[2]);
+    r1(3.14159, q[0]);
```
bettinaheim, line 28: What does the plot look like when
a) the rotation angle is a variable
b) the precision of the variable is higher/full double precision?

unittests/integration/draw_tester.cpp
```diff
@@ -0,0 +1,48 @@
...
+CUDAQ_TEST(DrawTester, checkEmpty) {
+
+  auto kernel = []() __qpu__ {};
+
+  std::string expected_str = "<empty trace>";
+  auto produced_str = cudaq::draw(kernel);
+  EXPECT_EQ(expected_str, produced_str);
+}
+
+CUDAQ_TEST(DrawTester, checkOps) {
```
bettinaheim, line 21: We need more testing than that. 
- We need tests to check the error message if the kernel contains measurements (at the end or in the middle)
- We need to test the output if the kernel calls another kernel 
- We need to cover all combinations of adjoint and controlled (does ctrl adj and adj ctrl plot properly?)
- We need to cover all operations in the default set documented [here](https://nvidia.github.io/cuda-quantum/latest/api/default_ops.html)

unittests/integration/draw_tester.cpp
```diff
@@ -0,0 +1,48 @@
...
+    h(q);
+    x<cudaq::ctrl>(q[0], q[1]);
+    z(q[2]);
+    r1(3.14159, q[0]);
+    swap(q[0], q[2]);
+    swap<cudaq::ctrl>(q[1], q[0], q[2]);
+  };
+
+  // clang-format off
+  std::string expected_str = R"(
```
bettinaheim, line 34: Nice! A good first step to help visualize kernels.

runtime/cudaq/algorithms/draw.cpp
```diff
@@ -0,0 +1,521 @@
...
+  std::vector<int> wire_layer(diagram.num_wires(), -1);
+
+  auto convertToIDs = [](const std::vector<QuditInfo> &qudits) {
+    std::vector<int> ids;
+    ids.reserve(qudits.size());
+    std::transform(qudits.cbegin(), qudits.cend(), std::back_inserter(ids),
+                   [](auto &info) { return info.id; });
+    return ids;
+  };
+  // TODO: fix
```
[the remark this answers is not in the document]
boschmitt: Thanks! Leftover note to self.

unittests/integration/draw_tester.cpp
boschmitt: > We need more testing than that.

These tests capture few edge cases that can happen in the drawer. I will add the tests you are requesting, but they won't be testing the drawer, they will be testing the execution manager capability of creating a correct trace. (It turns out it does get lost in some cases, but maybe those tests shouldn't here)

> * We need tests to check the error message if the kernel contains measurements (at the end or in the middle)

As stated in the PR description: This restrictions is not yet implemented because the drawer is based on the tracing capabilities of the execution manager, and the tracer still allows measurements.

A later PR needs implement this restriction but it will be a breaking change, as the current tracer functionality is public facing. (Although, I have doubts of if anyone is using it.)

> * We need to test the output if the kernel calls another kernel

Just keep in mind that the execution manager does not "see" calls to other kernels. All it sees are call to operations. Furthermore, the tracer doesn't even see calls to qubit allocation and deallocation, so it must infer the number of qubits from operations---this will make results a bit unexpected in the drawer, but is a fix that needs to be done in the execution manager.

> * We need to cover all combinations of adjoint and controlled (does ctrl adj and adj ctrl plot properly?)

The current tracer does not discriminate `adj`. It receives only an operation name, which is used as the label, and a list of qudits (the execution manager modifies the name in the case of `adj`). Also, because `cudaq::adj` and `cudaq::ctrl` are types, one cannot combine `op<ctrl, adj>(...)` (a pain point we should address in the future)

bettinaheim: 1) Could you please create a GitHub issue to capture the remaining work, including any test cases that still need to be added (minimally the ones I listed above).
2) I am looking for functionality tests - hence I do not care which piece of code ultimately supports or is changed to support that; the tests should not need to know that or rely on that. In the end, we want a certain behavior for the end-usage, and that's what I would like to see some tests for. It's fine to split that up across multiple PRs, but please capture all work in the same GitHub issue so that it is easier to track.
3) Re not seeing other kernels: Yes, I'm aware and that's fine - we'll make this fancier with and when we revise this more broadly. However, I'd still like to have a test now, since we still want that to work - just broken down into individual gates (kernel is effectively "inlined").
4) Having Adj be reflected in the gate name sounds fine.

boschmitt: The tests are already testing all gates. They also test calling other kernels and a combination of `cudaq::control`, `cudaq::adjoint` and `op<cudaq::adj>(...)`. 

I opened an issues regarding running the tracer in the presence of measurements, and the fact that it does not capture all quantum operations, e.g., qubit allocation.

unittests/integration/draw_tester.cpp
```diff
@@ -0,0 +1,106 @@
...
+     ╰───╯                                                             »
+
+################################################################################
+
+╭───────╮╭───────╮ ╭───────────────────────╮          ╭─────╮          »
+┤>      ├┤>      ├─┤ rx(2.718281828459045) ├──────────┤ sdg ├──────────»
+│       ││       │ ├───────────────────────┤          ╰──┬──╯          »
+┤●      ├┤●      ├─┤ ry(3.141592653589793) ├─────────────●─────────────»
+│  swap ││  swap │╭┴───────────────────────┴╮╭────────────────────────╮»
+│       │┤●      ├┤ rz(-3.1415927410125732) ├┤ rz(3.1415927410125732) ├»
```
bettinaheim, line 83: Hm, when trying it out locally on my machine the default precision was limited to a less digits. I think it would be good to truncate this more and use <iomanip>:
```
std::cout << std::setprecision(4) << value << std::endl
```
instead of
```
std::cout << value << std::endl
```

bettinaheim (approved):
This is fine - only pending item being to deal with measurements. (Left for separate PR.)

bettinaheim:
Superseded by/incorporated into https://github.com/NVIDIA/cuda-quantum/pull/1299.

PR #1206, opened 2024-02-09
Kernel builder JIT cache optimizations

bmhowe23:
This PR introduces 2 optimizations:

1. This changes the JIT cache key calculation from `std::hash<std::string>{}(modulePrintOut)` to one using `llvm::hash_code`. In my timing tests, the latter is ~27-29x faster than the former. (The generation of the `modulePrintOut` was the slow part, not the `std::hash`.)
2. Additionally, this eliminates redundant calls to `kernel_builder::jitCode()` when it is subsequently followed up with calls to `kernel_builder::jitAndInvoke()`. With the JIT cache key calculation change (above), this change is less important, but the redundant calls should not be needed, and are hence removed anyway.

Under some VQE test code, the combination of these two optimizations reduces the overall simulation runtime by ~50%, but YMMV.

### DIFF
--- a/runtime/cudaq/builder/kernel_builder.cpp
+++ b/runtime/cudaq/builder/kernel_builder.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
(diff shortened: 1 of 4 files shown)


runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -760,12 +760,11 @@ jitCode(ImplicitLocOpBuilder &builder, ExecutionEngine *jit,
...
-  {
-    llvm::raw_string_ostream os(modulePrintOut);
-    currentModule.print(os);
-  }
-  auto moduleHash = std::hash<std::string>{}(modulePrintOut);
+  auto hash = llvm::hash_code{0};
+  currentModule.walk([&hash](Operation *op) {
+    hash = llvm::hash_combine(hash, OperationEquivalence::computeHash(op));
+  });
+  auto moduleHash = static_cast<size_t>(hash);
```
boschmitt, line 767: I think this could be:
```suggestion
  auto moduleHash = llvm::hash_value(hash);
```

bmhowe23: Hmm, that suggestion doesn't compile for me, and I was intending to use [this API](https://llvm.org/doxygen/classllvm_1_1hash__code.html#a8db88897f6240357ac4264feccae5d57) to convert it to `size_t`, not `llvm::hash_code` (because that's what our `unordered_map` is assuming).

The way I interpret the docs on `hash_value()` function is that types can implement that function for their type, and then other LLVM infrastructure would know how to hash that type. That is - if you wanted to hash operations in a special way, you would write the `hash_value` function for your type, so `hash_value` [function](https://llvm.org/doxygen/Hashing_8h_source.html#l00096) that you're referring to is really defined as a way to _hash hashes_, which isn't really what I'm trying to do. But maybe that's just my interpretation?

FWIW, it looks like the `static_cast<size_t>`(hash) is how it's done in the [LLVM hash test](https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/llvm/unittests/ADT/HashingTest.cpp#L306).

boschmitt: You are right, my suggestion was based on my vague memory of `llvm::have_value`, sorry for the noise.

amccaskey (approved):
Thanks for the perf improvement!

PR #1210, opened 2024-02-12
Ability to run qpud in QIR-only mode with `CallOp` verification

1tnguyen:
### Description

- `nvcf.config` to disable library mode + sanity check within the client to verify that.

- Add a subclass for server implementation to do server-side filtering for MLIR (lowered to QIR).

- Add a verification for `CallOp`: either QIR functions or NVQIR-known functions as defined in `QIRFunctionNames.h`
 
- Add a missing MLIR pass in the config.

### DIFF
--- a/lib/Optimizer/CodeGen/LowerToQIRProfile.cpp
+++ b/lib/Optimizer/CodeGen/LowerToQIRProfile.cpp
@@ -617,7 +617,7 @@ struct VerifyNVQIRCallOpsPass
         if (!isKnownFunctionName(funcName)) {
           call.emitOpError("unexpected function call in NVQIR");
           passFailed = true;
-          return WalkResult::advance();
+          return WalkResult::interrupt();
         }
         return WalkResult::advance();
       }
--- a/lib/Optimizer/CodeGen/LowerToQIRProfile.cpp
+++ b/lib/Optimizer/CodeGen/LowerToQIRProfile.cpp
@@ -587,6 +587,54 @@ cudaq::opt::verifyQIRProfilePass(llvm::StringRef convertTo) {
   return std::make_unique<VerifyQIRProfilePass>(convertTo);
 }
 
+namespace {
+/// Verify that the QIR doesn't have any "bonus" calls to arbitrary code that is
+/// not possibly defined in the QIR standard or NVQIR runtime.
+struct VerifyNVQIRCallOpsPass
+    : public cudaq::opt::VerifyNVQIRCallOpsBase<VerifyNVQIRCallOpsPass> {
+  explicit VerifyNVQIRCallOpsPass() : VerifyNVQIRCallOpsBase() {}
+
+  void runOnOperation() override {
+    LLVM::LLVMFuncOp func = getOperation();
+    bool passFailed = false;
+    // Check that a function name is either QIR or NVQIR registered.
+    const auto isKnownFunctionName = [](llvm::StringRef functionName) -> bool {
+      if (functionName.startswith("__quantum_"))
+        return true;
+      static const std::vector<llvm::StringRef> NVQIR_FUNCS = {
+          cudaq::opt::NVQIRInvokeWithControlBits,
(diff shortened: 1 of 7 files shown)


runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -396,6 +409,14 @@ class NvcfRuntimeClient : public RemoteRestRuntimeClient {
             kernelName;
       return false;
     }
+
+    if (request.format != cudaq::CodeFormat::MLIR) {
+      // The `.config` file may have been tampered with.
+      std::cerr << "Internal error: unsupported kernel IR detected.\nThis may "
+                   "indicate a corrupted CUDA Quantum installation.";
+      std::abort();
```
bmhowe23, line 417: Just curious: why not use a runtime_exception here?

tools/cudaq-qpud/RestServerMain.cpp
```diff
@@ -31,7 +31,22 @@ int main(int argc, char **argv) {
   llvm::cl::ParseCommandLineOptions(argc, argv, "CUDA Quantum REST server\n");
   if (cudaq::mpi::available())
     cudaq::mpi::initialize();
-  auto restServer = cudaq::registry::get<cudaq::RemoteRuntimeServer>("rest");
+  std::string serverType = "rest";
+  // Check environment variable for any specific server subtype.
+  if (auto serverSubType = std::getenv("CUDAQ_SERVER_TYPE")) {
+    if (cudaq::registry::isRegistered<cudaq::RemoteRuntimeServer>(
+            serverSubType)) {
+      printf("[cudaq-qpud] Using server subtype: %s\n", serverSubType);
```
bmhowe23, line 39: When this is deployed, will we be able to see this message to verify that we have correctly set it up? I'm wondering if it might be more clear if we use regular command-line parsing here vs environment variables. What do you think?

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
1tnguyen: I just wanted to denote a truly abnormal condition, hence not allowing any recover actions down the line, e.g., with try-catch. Happy to revise it to a `runtime_exception` if you think that makes more sense.

bmhowe23: Nah, I don't think it needs to be changed. That makes sense.

tools/cudaq-qpud/RestServerMain.cpp
1tnguyen: Yes, we should be able to see this log in NVCF. I think it makes more sense to use regular command-line parsing. Will push an update shortly.

1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1210/commits/27e5eb8bd44591d4c71fc02c73c41cfa7e4b9fab along with `CallOp` verification. 
We allow calls to QIR functions and a couple of registered NVQIR functions.

lib/Optimizer/CodeGen/LowerToQIRProfile.cpp
```diff
@@ -587,6 +587,54 @@ cudaq::opt::verifyQIRProfilePass(llvm::StringRef convertTo) {
...
+             NVQIR_FUNCS.end();
+    };
+
+    func.walk([&](Operation *op) {
+      if (auto call = dyn_cast<LLVM::CallOp>(op)) {
+        auto funcName = call.getCalleeAttr().getValue();
+        if (!isKnownFunctionName(funcName)) {
+          call.emitOpError("unexpected function call in NVQIR");
+          passFailed = true;
+          return WalkResult::advance();
```
bmhowe23, line 620:
```suggestion
          return WalkResult::interrupt();
```

bmhowe23 (approved):
👍

PR #1212, opened 2024-02-12
re-enable state, get_state, and get_state_async

amccaskey:
Also remove unnecessary source files.


khalatepradnya:
Changes look good. Thanks for the clean up as well!
Docs stage is failing in CI for missing reference to `State`, I think should go in this - `docs/sphinx/api/languages/python_api.rst`. Also it has some `SKIP_TEST` markers that can be removed.

khalatepradnya (approved):
👍🏽

PR #1213, opened 2024-02-12
Add support for CURL_CA_BUNDLE environment variable

bmhowe23:
This change allows users to specify the path to CA bundle they want to use for remote backends, and it is in a similar way to how they would do it if they were using the curl command-line tool. (Our statically linked curl does not read that environment variable, and the standard curl tool does.)

Also, if no environment variable is specified, fall back on CUDAQ `cacert.pem` file in CUDAQ installation directory (if it exists).

See https://curl.se/docs/sslcerts.html for additional details.


runtime/common/RestClient.cpp
```diff
@@ -48,6 +49,38 @@ std::string decompress_gzip(const std::string &data) {
...
+    if (auto *curlCABundleStr = getenv("CURL_CA_BUNDLE")) {
+      if (std::filesystem::exists(curlCABundleStr))
+        return curlCABundleStr;
+      else
+        cudaq::info(
+            "{} does not exist. Will fall back on CUDAQ installed certs",
+            curlCABundleStr);
+    }
+    std::filesystem::path cudaqLibPath{cudaq::getCUDAQLibraryPath()};
+    auto certPath = cudaqLibPath.parent_path().parent_path() / "cacert.pem";
```
bmhowe23, line 63: @bettinaheim please double check me on this path. It would be map to `/opt/nvidia/cudaq/cacerts.pem` once everything is installed.

runtime/common/RestClient.h
```diff
@@ -11,6 +11,11 @@
 #include <map>
 #include <string>
 
+// Forward declarations to avoid including CPR header files
+namespace cpr {
+struct SslOptions;
+};
```
schweitzpgi, line 17:
```suggestion
}
```

schweitzpgi:
>This change allows users to specify the path to CA bundle they want to use for remote backends, and it is in a similar way to how they would do it if they were using the curl command-line tool. (Our statically linked curl does not read that environment variable, and the standard curl tool does.)

This begs the question: should we be upgrading our version of curl here?

bmhowe23:
> > This change allows users to specify the path to CA bundle they want to use for remote backends, and it is in a similar way to how they would do it if they were using the curl command-line tool. (Our statically linked curl does not read that environment variable, and the standard curl tool does.)
> 
> This begs the question: should we be upgrading our version of curl here?

I think an upgraded version of curl would have the same behavior.

runtime/common/RestClient.cpp
bmhowe23: > @bettinaheim please double check me on this path. It would map to `/opt/nvidia/cudaq/cacerts.pem` once everything is installed (or the top level of the Python wheel)

I checked this in the `latest-base` image, which includes the change to include that file, and it is indeed in that location.

1tnguyen (approved):
LGTM 👍

PR #1214, opened 2024-02-12
Building wheels with gcc12

bettinaheim:


schweitzpgi (commented):
Shouldn't this be doing a `scl enable gcc-toolset-12 bash` someplace?

bettinaheim:
> scl enable gcc-toolset-12 bash

Setting CC and CXX to the right gcc version should be enough - gcc should find its own libraries. However, I gave it a quick shot to enable the toolchain (within a docker file, environment variables set by a RUN command do not persist beyond that command - hence we enable that toolchain upon invocation as part of the wrapper script to the used compiler).

PR #1226, opened 2024-02-13
Add a cmake variable to disable building the REST parts of the runtime.

schweitzpgi:
Under some circumstances, it may be expediant to disable the building of the REST and REST Server libraries in the runtime. This adds a cmake variable, CUDAQ_ENABLE_REST, for exactly that purpose.


1tnguyen (approved):
LGTM 👍

PR #1227, opened 2024-02-14
NVCF nightly integration tests

1tnguyen:
### Description

Add NVCF integration tests to our nightly integration testing CI.

- Deploy a test function using the NGC CLI: build and push NVCF deployment container, create and deploy NVCF function version. 

- Run C++ and Python examples and tests using our nightly container targeting the function id + version id of the above function.

- Retrieve the wheel from Publishing artifact + install it in a clean environment, run NVCF Python examples to check that it works.  Since this is mainly a sanity check for the wheel, only Python 3.10 is used. 

- Clean up the function and Docker image when the tests complete.


1tnguyen:
A sample run of this new workflow: https://github.com/1tnguyen/cuda-quantum/actions/runs/7908912750

docker/release/cudaq.nvcf.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+#   DOCKER_BUILDKIT=1 docker build -f docker/release/cudaq.nvcf.Dockerfile . --output out
+
+# Base image is CUDA Quantum image 
+ARG base_image=nvcr.io/nvidia/nightly/cuda-quantum:latest
+FROM $base_image as nvcf_image
+
+USER root
+
+ENV CUDAQ_LOG_LEVEL=info
+ADD ./scripts/launch_qpud.sh /launch_qpud.sh
```
bettinaheim, line 22: Seeing as this is just one command, it seems unnecessary to add a separate script.
Why not just 
```
RUN echo 'mpiexec --allow-run-as-root -np $(nvidia-smi --list-gpus | wc -l) cudaq-qpud --type nvcf' > launch.sh
```
and
```
ENTRYPOINT ["bash", "-l", "launch.sh"]
```

.github/actions/install-ngc-cli/action.yml
```diff
@@ -0,0 +1,20 @@
+name: install-ngc-cli
```
bettinaheim, line 1: Since this is now its own action, it would be nice to update docker_images.yml and publishing.yml to use it.

.github/workflows/integration_tests.yml
```diff
@@ -23,28 +29,166 @@ on:
...
+        uses: ./.github/actions/install-ngc-cli
+        with:
+          version: 3.38.0
+          checksum: 427c67684d792b673b63882a6d0cbb8777815095c0f2f31559c1570a91187388
+
+      - name: Deploy NVCF Function
+        id: deploy
+        env:
+          NGC_CLI_API_KEY: ${{ secrets.NGC_NVCF_CREDENTIALS }}
+          NGC_CLI_ORG: ${{ secrets.NGC_NVCF_ORG }}
```
bettinaheim, line 147: can't you use the org name instead of the org ID here? It's odd that the org is a variable but the team is hardcoded. I would say we either hardcode both or I create a normal variable for the org name and for the team name.

.github/workflows/integration_tests.yml
```diff
@@ -253,3 +397,228 @@ jobs:
...
+            exit 1
+          fi
+        shell: bash
+
+  wheel_integration_test:
+    name: Integration test using Python wheels
+    runs-on: ubuntu-latest
+    needs: [metadata, build_nvcf_image, deploy_nvcf_test_function]
+    environment: backend-validation
+    if: (success() || failure()) && (inputs.target == 'nvcf' || github.event_name == 'schedule' || inputs.target == 'nightly')
```
bettinaheim, line 542: Why run this job when a dependency failed?

.github/workflows/integration_tests.yml
```diff
@@ -23,28 +29,166 @@ on:
...
+            exit 1
+          fi
+
+  integration_test:
+    name: Integration test using Docker
+    runs-on: ubuntu-latest
+    needs: [metadata, build_nvcf_image, deploy_nvcf_test_function]
+    # when none of the needed jobs fail or are cancelled (skipped or successful jobs are ok).
+    # For example, if only one hardware backend test is requested (ionq, iqm, etc.), NVCF setup jobs are skipped.
+    if: ${{ always() && !failure() && !cancelled() }}
```
bettinaheim, line 181: I think it would be cleaner to just create a separate job to validate submission to nvcf. Then you could also make the different tests separate steps to make it easier to find what failed.

.github/workflows/integration_tests.yml
```diff
@@ -23,28 +29,166 @@ on:
...
+            $NVCF_FUNCTION_ID)
+          version_id=$(echo "$create_function_result" | grep 'Version: \S*' |  head -1 | cut -d ':' -f 2 | tr -d ' ')
+          echo "Create version Id: $version_id"
+          echo "nvcf_function_version_id=$version_id" >> $GITHUB_OUTPUT
+          # Deploy it
+          ngc-cli/ngc cloud-function function deploy create --deployment-specification $NGC_NVCF_DEPLOYMENT_SPEC $NVCF_FUNCTION_ID:$version_id
+          function_status=DEPLOYING
+          while [ "$function_status" =  "DEPLOYING" ]; do 
+            echo "Waiting for deploying NVCF function version $version_id ..." 
+            sleep 300
```
bmhowe23, line 166:
```suggestion
            sleep 30 
```

.github/workflows/integration_tests.yml
```diff
@@ -23,28 +29,166 @@ on:
...
+      workflow_id:
+        type: string
+        required: false
+        description: 'Workflow Id to retrieve the Python wheel for testing. Default to the wheels produced by the Publishing workflow associated with the latest nightly CUDA Quantum Docker image if left blank'
+
   schedule:
     - cron: 0 3 * * *
 
+env:
+  NVCF_FUNCTION_ID: 3bfa0342-7d2a-4f1b-8e81-b6608d28ca7d
```
bmhowe23, line 55: Will this be a unique function ID? With our default client behavior of "use the latest version", we don't want these tests to interfere with others' usage. I'm also wondering if this should be moved to a secret.

.github/workflows/integration_tests.yml
```diff
@@ -253,3 +397,228 @@ jobs:
...
+        with:
+          ref: ${{ needs.metadata.outputs.cudaq_commit }}
+          fetch-depth: 1
+
+      - name: Install wheel
+        # Note: we only test Python 3.10 wheel in this workflow to do a quick sanity check for NVCF integration. 
+        run: |
+          workflow_id=${{ inputs.workflow_id }}
+          if [ -z "${workflow_id}" ]; then
+            workflow_id=$(gh run list --commit ${{ needs.metadata.outputs.cudaq_commit }} --workflow Publishing --status completed --event workflow_dispatch --limit 1 --json databaseId | jq '.[0].databaseId')
```
bmhowe23, line 556: How long will the wheels be retained? If we don't publish any new images a week, will the oldest ones still be there?

.github/workflows/integration_tests.yml
```diff
@@ -253,3 +397,228 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Test NVCF 
+        run: |
+          echo "### Submit to NVCF from Python wheels" >> $GITHUB_STEP_SUMMARY
+          export NVCF_API_KEY="${{ secrets.NVCF_API_KEY }}"
+          export NVCF_FUNCTION_ID="$NVCF_FUNCTION_ID"
+          export NVCF_FUNCTION_VERSION_ID="${{ needs.deploy_nvcf_test_function.outputs.nvcf_function_version_id }}"
+          export CUDAQ_LOG_LEVEL="info"
```
bmhowe23, line 572: It looks like you redirecting stdout to /dev/null, so the log level environment variable can be deleted.

.github/actions/install-ngc-cli/action.yml
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1227/commits/2da783c3988a680fef5d7f6aa360da33fe7a33a5

.github/workflows/integration_tests.yml
1tnguyen: We'll reserve/use this function Id for CI and create new functions for other usage, such as development/staging/production in the future. 
Re: repo secret vs. hardcoded value. There is no real reason to hide this UUID and keeping it in the code will allow updates via a regular PR workflow.

docker/release/cudaq.nvcf.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+#
+# Usage:
+# Must be built from the repo root with:
+#   DOCKER_BUILDKIT=1 docker build -f docker/release/cudaq.nvcf.Dockerfile . --output out
+
+# Base image is CUDA Quantum image 
+ARG base_image=nvcr.io/nvidia/nightly/cuda-quantum:latest
+FROM $base_image as nvcf_image
+
+USER root
```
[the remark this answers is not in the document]
1tnguyen: It was a copy-and-paste oversight. I've removed this line.

docker/release/cudaq.nvcf.Dockerfile
1tnguyen: Good point. I've updated the file as suggested (without `--allow-run-as-root`) and removed the `launch_qpud.sh` file.

.github/workflows/integration_tests.yml
1tnguyen: No, I cannot use the org name. I've hardcoded the org Id instead to keep it consistent.

.github/workflows/integration_tests.yml
1tnguyen: It's an oversight. I've removed that condition.

.github/workflows/integration_tests.yml
1tnguyen: Agreed. I've refactored nvcf submission into a separate job.

.github/workflows/integration_tests.yml
1tnguyen: I think 30 secs is too short. Typically, the deployment would take 10+ minutes. I've reduced this value to `120` :)

.github/workflows/integration_tests.yml
1tnguyen: I've revised this logic to allow for the 'expired artifacts' condition and it seems to work when I tested it against this [run](https://github.com/NVIDIA/cuda-quantum/actions/runs/7733520121), whose artifacts are all expired.

.github/workflows/integration_tests.yml
1tnguyen: Thanks. I've removed this line.

.github/workflows/integration_tests.yml
```diff
@@ -253,3 +397,228 @@ jobs:
...
+        run: |
+          workflow_id=${{ inputs.workflow_id }}
+          if [ -z "${workflow_id}" ]; then
+            workflow_id=$(gh run list --commit ${{ needs.metadata.outputs.cudaq_commit }} --workflow Publishing --status completed --event workflow_dispatch --limit 1 --json databaseId | jq '.[0].databaseId')
+          fi
+          echo "Using artifacts from workflow id $workflow_id"
+          gh run download $workflow_id
+          apt-get update && apt-get install -y --no-install-recommends python3.10 python3-pip
+          wheelfile=$(find . -name 'cuda_quantum*cp310*x86_64.whl')
+          python3.10 -m pip install $wheelfile
```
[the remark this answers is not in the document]
1tnguyen: Added python version as a workflow input.

bmhowe23 (approved):
LGTM 👍

PR #1228, opened 2024-02-14
Fix bug in logic for decrementing range

amccaskey:


python/cudaq/kernel/ast_bridge.py
```diff
@@ -718,14 +718,26 @@ def visit_Call(self, node):
...
+                    loadedCounter = cc.LoadOp(counter).result
+                    tmp = arith.MulIOp(loadedCounter, stepVal).result
                     arrElementVal = arith.AddIOp(startVal, tmp).result
                     eleAddr = cc.ComputePtrOp(
-                        cc.PointerType.get(self.ctx, iTy), iterable, [iterVar],
+                        cc.PointerType.get(self.ctx, iTy), iterable,
+                        [loadedCounter],
                         DenseI32ArrayAttr.get([-2147483648], context=self.ctx))
                     cc.StoreOp(arrElementVal, eleAddr)
+                    incrementedCounter = arith.AddIOp(loadedCounter, one).result
```
schweitzpgi, line 739: If this is decrementing is the step "normalized" to be non-positive?

python/tests/mlir/ast_decrementing_range.py
```diff
@@ -0,0 +1,82 @@
...
+        qubits = cudaq.qvector(5) 
+        for k in range(q, p, -1):
+            x(qubits[k])
+
+    print(test)
+    test(2, 0)
+
+# CHECK-LABEL:   func.func @__nvqpp__mlirgen__test(
+# CHECK-SAME:                                      %[[VAL_0:.*]]: i64,
+# CHECK-SAME:                                      %[[VAL_1:.*]]: i64) attributes {"cudaq-entrypoint"} {
```
schweitzpgi, line 32: I think this case is fine, but as a heads-up, the script can get these wrong. For example, if there were 3 arguments of type `i64` we'd get back to back lines
```
# CHECK-SAME:                                      %[[VAL_0:.*]]: i64,
# CHECK-SAME:                                      %[[VAL_1:.*]]: i64,
```
That is a problem because of the greedy match on the first line of the pair.

In my experience, it is always a good idea to clean these up by fusing the CHECK-SAME lines back together. We can also erase tons of gratuitous whitespace that the script adds for no reason.

schweitzpgi (commented):
LGTM

python/cudaq/kernel/ast_bridge.py
amccaskey: They way I have it implemented is that I keep track of an incrementing counter, so this Value here should be always incrementing by one. But the step Value could be a negative or positive integer. So 
```python
tmp = arith.MulIOp(loadedCounter, stepVal).result
arrElementVal = arith.AddIOp(startVal, tmp).result
```
will increment or decrement based on whatever the step value is. I'm trying to create this 
```python
array = [start, start +- step, start +- 2*step, start +- 3*step, ...]
#- or -
array = [start +- counter * step ...] 
```
so step is the only thing that determines that `+-`

schweitzpgi: So, that sounds like a yes.

I find the `+-` notation confusing since, as you say, the counter is monotonically increasing. Maybe a notation like

```
array := [start + (step * 0), start + (step * 1), start + (step * 2) ... ] where step is non-zero
```
would be clearer?

(One can also transform these into an exact count, but beware there are incorrect ways of doing it all over the internet.)

PR #1229, opened 2024-02-14
[tests] Small fix: add unnamed namesapces

boschmitt:
### Description
We need unnamed namespaces here otherwise these functions will leak and might interfere with other tests.


schweitzpgi (approved):
Might consider using `static` to conform to LLVM coding guidelines.

PR #1243, opened 2024-02-15
[testing] Reorganize the test structure.

schweitzpgi:
The various directories should be organized as follows:

test - these are small regression tests that can be run very quickly and test
  small combinations of functionality to prevent regressions. does not use
  gtest framework, but rather FileCheck and friends.

unittests - similar to test, but fixating on one particular "unit". tests the
  unit's boundaries, limits, etc. to make sure that a particular encapsulated
  blob of computation (the unit) works as advertised. uses gtest framework.

targettests - not like the 2 above. these are end-to-end tests that might take
  a long time to run but test a whole 9 yards of functionality. so if you
  wanted to run a game of quantum nethack, it would go here.

  note: eventually these tests should be executable out-of-tree so that
  installations of cuda quantum can be tested sans source code.

python/tests - these should be folded into the above testing
  infrastructure.


bmhowe23:
What is the desired convention for when items would go into `test/NVQPP` instead of `targettests`?

boschmitt:
It would be great to put your PR description on a README for future reference.

schweitzpgi:
> What is the desired convention for when items would go into `test/NVQPP` instead of `targettests`?

Basically, don't put a test that is expected to execute itself anywhere is `test/`. The holdovers were not executing themselves, but printing out the compiler's version string, making sure `nvq++` was functional as a linker, etc.

I was thinking of renaming the `test/NVQPP` directory, but backed off on that in this PR. For one thing, it's not clear what would be a "perfect name" exactly.

schweitzpgi:
FWIW: With this change, the regression tests run quickly and do not freeze or crush a smaller/weaker/malnourished system.

bmhowe23 (approved):
Thanks, Eric.

PR #1244, opened 2024-02-15
[experimental/python] JIT is on by default in Python

khalatepradnya:
### Description
Make 'MLIR' mode the default, i.e. JIT=True

### DIFF
--- a/.github/workflows/test_in_devenv.yml
+++ b/.github/workflows/test_in_devenv.yml
@@ -219,11 +219,11 @@ jobs:
               echo "::error file=test_in_devenv.yml::Python tests failed with status $pytest_status."
               exit 1
             fi
-            CUDAQ_PYTEST_EAGER_MODE=OFF python3 -m pytest -v python/tests/ \
+            CUDAQ_PYTEST_EAGER_MODE=ON python3 -m pytest -v python/tests/ \
               --ignore python/tests/backends
             pytest_status=$?
             if [ ! $pytest_status -eq 0 ]; then
-              echo "::error file=test_in_devenv.yml::Python tests failed with status $pytest_status."
+              echo "::error file=test_in_devenv.yml::Python tests failed with status $pytest_status in eager mode."
               exit 1
             fi
             for backendTest in python/tests/backends/*.py; do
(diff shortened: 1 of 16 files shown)


khalatepradnya:
For C++ https://github.com/NVIDIA/cuda-quantum/pull/1015

schweitzpgi (commented):
+1

.github/workflows/test_in_devenv.yml
```diff
@@ -219,11 +219,11 @@ jobs:
               echo "::error file=test_in_devenv.yml::Python tests failed with status $pytest_status."
               exit 1
             fi
-            CUDAQ_PYTEST_EAGER_MODE=OFF python3 -m pytest -v python/tests/ \
+            CUDAQ_PYTEST_EAGER_MODE=ON python3 -m pytest -v python/tests/ \
```
bmhowe23, line 222: Do some of the tests not support MLIR mode yet?

docs/sphinx/examples/python_runtime/noise_amplitude_damping.py
```diff
@@ -22,7 +22,7 @@
 
 
 # Now we may define our simple kernel function and allocate a qubit.
-@cudaq.kernel(jit=True)
+@cudaq.kernel(jit=False)
```
bmhowe23, line 25: Why go from true to false on this one?

docs/sphinx/examples/python_runtime/qaoa_maxcut.py
```diff
@@ -26,7 +26,7 @@
 parameter_count: int = 2 * layer_count
 
 
-@cudaq.kernel(jit=True)
+@cudaq.kernel(jit=False)
```
bmhowe23, line 29: Why go from true to false on this one?

bmhowe23 (approved):

.github/workflows/test_in_devenv.yml
khalatepradnya: It's the other way round, some tests are only for MLIR mode. 
Prior to this change, eager mode was default, and we ran all tests in both modes. Now that MLIR mode is default, I changed this environment variable to 'ON' to run other mode.

docs/sphinx/examples/python_runtime/noise_amplitude_damping.py
khalatepradnya: This was a typo, all the examples under `python_mlir` had `jit=True` and those under `python_runtime` either relied on default (which was eager mode) or `jit=False`.

docs/sphinx/examples/python_runtime/qaoa_maxcut.py
khalatepradnya: Same as above.

PR #1248, opened 2024-02-16
[unittest] Quick fix: precision issue with f32 backend

boschmitt:
### Description
When using a cuStateVec_fp32 backend for simulation, this test runs into a precision issue when comparing floating point numbers. The problem is that our whole frame work assumes double precision (`double`)---all matrices, the `cudaq::state` object, and etc, use `double` to represent real numbers. This backend, however, works on single precision (`float`), meaning that all those values are silently casted to `float`, the simulation is done, and the result is casted back to `double` and presented to the user, who might be unsuspecting of the change.

This is a quick fix, but we need to think about a holistic solution to how define precision when using CUDA Quantum.


bmhowe23 (approved):
👍

Issue #1245, opened 2024-02-16
[experimental/python] Python wheel on v3.8 not working


khalatepradnya:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Installing `cuda_quantum-0.0.0-cp38-cp38-manylinux_2_28_x86_64.whl` results in `ImportError`

`ImportError while importing test module '/tmp/tests/backends/test_IQM.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
usr/lib64/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tmp/tests/backends/test_IQM.py:14: in <module>
    import cudaq
usr/local/lib64/python3.8/site-packages/cudaq/__init__.py:11: in <module>
    from .kernel.kernel_decorator import kernel, PyKernelDecorator, globalImportedKernels
usr/local/lib64/python3.8/site-packages/cudaq/kernel/kernel_decorator.py:15: in <module>
    from .ast_bridge import compile_to_mlir, PyASTBridge
usr/local/lib64/python3.8/site-packages/cudaq/kernel/ast_bridge.py:9: in <module>
    import graphlib
E   ModuleNotFoundError: No module named 'graphlib'`

### Steps to reproduce the bug

* Install the new wheel in Python 3.8 environment

### Expected behavior

* Able to run examples successfully

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: experimental/python
- **Python version**: 3.8
- **C++ compiler**: N/A
- **Operating system**: Fedora38

khalatepradnya:
Fixed. Ref: https://github.com/NVIDIA/cuda-quantum/actions/runs/8001702940

PR #1252, opened 2024-02-17
[experimental/python] Fixes to support Python v3.8

khalatepradnya:
### Description

Addresses https://github.com/NVIDIA/cuda-quantum/issues/1245


pyproject.toml
```diff
@@ -19,6 +19,7 @@ license = { file="LICENSE" }
 dependencies = [
   'astpretty ~= 3.0',
   'cuquantum-cu11 ~= 23.10',
+  'graphlib-backport >= 1.0',
```
bettinaheim, line 22: What exactly is using graphlib?

amccaskey: There is a topological sort needed for the kernel's dependent kernel call graph. I didn't realize graphlib was a 3.9+ library.

bettinaheim: Ah, I see. Not urgent to address, but I think we should consider just writing the topological sort ourselves and getting rid of this dependency. 
Edit: I just checked eol support; I would say we should drop support for Python versions when the security support ends, which would be October 24 for Python 3.8. If we don't have any issue with the backport library, we can leave it as is until then.

bettinaheim (approved):

amccaskey (approved):
+1 for dropping 3.8 when it goes EOL

PR #1253, opened 2024-02-17
Implement list construction from list comprehension and simple kernel return types

amccaskey:
* list construction from list comprehensions 
```python 
myList = [[i, i+1] for i in range(N)]
myList2 = [k for k in range(10)]
```
* Disallow empty list construction `l = []` and `l.append(...)`
* Kernels with return types (first pass, only `int` and `float` implemented)
```python
@cudaq.kernel
def kernel(a:int, b:int) -> int:
     return a * b
eight = kernel(2, 4)
```
* Throw an exception for failures in `pyAltLaunchKernel` and `ExecutionEngine::create(...)` (was silently seg faulting)


schweitzpgi:
Do the list comprehensions get object inlined into a static vector structure?

python/tests/kernel/test_kernel_features.py
```diff
@@ -344,3 +342,68 @@ def test(q: int, p: int):
...
+
+
+def test_list_creation():
+
+    N = 10 
+    @cudaq.kernel
+    def kernel(N :int, idx: int) -> int:
+        myList = [i+1 for i in range(N-1)]
+        return myList[idx]
+
```
khalatepradnya, line 395: Q: Is this allowed?
`qubits_list = list(range(N))`
where, `N:int` is kernel argument.

amccaskey:
> Do the list comprehensions get object inlined into a static vector structure?

Yes that is correct. I allocate a stdvec of known size and then loop through and fill that the vector.

python/tests/kernel/test_kernel_features.py
amccaskey: `list(...)` is not yet implemented, but I can add that here.

khalatepradnya (approved):
👍🏽

PR #1254, opened 2024-02-18
Optimize logger implementation

bmhowe23:
Bypass some processing if the current log level prevents the message from actually being logged.

For the most part, this means that we check the log level before calling `fmt::format()`. However, note that some log messages form strings outside of the `cudaq::info` structs and then pass the result into the `cudaq::info()` structs. We cannot optimize those out without more invasive code changes, which for the most part, I've left for a future task. The only exception to that was one particularly hot piece of code in `runtime/nvqir/CircuitSimulator.h`, where I've wrapped it in a new `cudaq::details::should_log()` function.

Testing on Gorby with some UCCSD benchmarks from @marwafar reveals these changes **reduce runtime by 17-32%**, but as always, YMMV.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -1,5 +1,5 @@
 /****************************************************************-*- C++ -*-****
- * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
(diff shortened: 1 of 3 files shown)


amccaskey (approved):

runtime/nvqir/CircuitSimulator.h
```diff
@@ -935,7 +935,10 @@ class CircuitSimulatorBase : public CircuitSimulator {
                                const std::vector<std::size_t> &targets) {
     flushAnySamplingTasks();
     QuantumOperation gate;
-    cudaq::info(gateToString(gate.name(), controls, angles, targets));
+    // This is a very hot section of code. Don't form the log string unless
+    // we're actually going to use it.
+    if (cudaq::details::should_log(cudaq::details::LogLevel::info))
```
amccaskey, line 940: big thumbs up 👍

amccaskey (approved):
+1

PR #1258, opened 2024-02-20
Update function call verification for remote targets

1tnguyen:
### Description

(1) Enhance the client side in MLIR: incorporate all quake code of cuda quantum kernels, not just the entry-point kernel. This is to cope with inlining being disabled.

(2) Update the server-side verification helper to allow functions defined in the submitted module to be called.


targettests/Remote-Sim/no_inline.cpp
```diff
@@ -0,0 +1,43 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
bmhowe23, line 2: Nit since this is a new file.
```suggestion
 * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
```

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -136,11 +136,19 @@ class RemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
       // Add entry-point attribute if not already set.
       if (!func->hasAttr(cudaq::entryPointAttrName))
         func->setAttr(cudaq::entryPointAttrName, builder.getUnitAttr());
       auto moduleOp = builder.create<ModuleOp>();
-      moduleOp.push_back(func.clone());
+      for (auto &op : *module) {
+        auto funcOp = dyn_cast<func::FuncOp>(op);
+        // Add quantum kernels defined in the module.
+        if (funcOp && funcOp->hasAttr(cudaq::kernelAttrName))
+          moduleOp.push_back(funcOp.clone());
```
bmhowe23, line 150: Double check me on this, but wouldn't we also want to allow classical calls defined in the kernel, too? (But only if they're truly defined, not just declared and hoped to be linked in later.)

1tnguyen: That's right. But the bridge would only process quantum kernels, thus only those related to them are present in the quake code. This filter is mainly for non-JIT wrappers (required for direct execution) such as `altLaunchKernel`, `thunk`, etc., which we don't need when invoking the JIT'ed function.

A side effect of those wrappers is that they are not compatible with the synthesizer pass since they call the original kernel signature, hence will fail IR verification.

PR #1261, opened 2024-02-20
Optimize LLVM JIT for large circuits

bmhowe23:
This makes a big difference for large UCCSD circuits when using `llvm::CodeGenOpt::None`, which we do in multiple places.

For example, the JIT time for one of @marwafar's benchmark circuits went from ~260 seconds to ~45 seconds. (Most of the remaining time is spent in our Canonicalizer pass.)

Additional info: prior to this PR, we were spending a large amount of time in https://github.com/llvm/llvm-project/blob/llvmorg-16.0.6/llvm/lib/CodeGen/SelectionDAG/SelectionDAGISel.cpp#L1581 for large circuits. This change remedies that by disabling `-fast-isel`. The "fast" instruction selection was supposed to be faster than the alternatives, but for us, it is actually slower for large circuits where it matters.

Also note that I believe https://github.com/llvm/llvm-project/blob/llvmorg-16.0.6/llvm/lib/CodeGen/TargetPassConfig.cpp#L997 is preventing us from doing this in cleaner way where we specify this option via `TargetMachine::setFastISel(false)` and `TargetMachine::setO0WantsFastISel(false)`.


runtime/common/RuntimeMLIR.cpp
```diff
@@ -572,6 +572,13 @@ std::unique_ptr<MLIRContext> initializeMLIR() {
 
 ExecutionEngine *createQIRJITEngine(ModuleOp &moduleOp,
                                     llvm::StringRef convertTo) {
+  // The "fast" instruction selection compilation algorithm is actually very
+  // slow fast for large quantum circuits. Disable that here. Revisit this
```
1tnguyen, line 576: Just checking that "very slow fast" is intended.

1tnguyen (approved):
Nice find 👍

runtime/common/RuntimeMLIR.cpp
bmhowe23: Fixed (and elaborated a bit more) in 28254ef28.

PR #1264, opened 2024-02-21
Fixes for nightly integration workflow

1tnguyen:
### Description

(1) Use the same `NGC_CREDENTIALS` as other NGC publishing jobs.

(2) Set the proper `ghcr-deployment` environment where these secrets live.

(3) Change the default deployment backend. It seems to be more responsive as compared to GFN for deployment.

(4) Make the artifact lookup more resilient: can cope with unknown commit info and/or incomplete Publishing runs (e.g., picking up a newly-pushed docker image while the workflow is still running).

Sample workflow run on this branch: https://github.com/NVIDIA/cuda-quantum/actions/runs/7987784286


bmhowe23 (approved):
LGTM 👍

PR #1273, opened 2024-02-22
Publishing fixes

bettinaheim:
This PR:
- Fixes the installer validation
- Fixes the missing build_info.txt file in the docker image
- Fixes the accidentally skipped validation of the docker images and adds a check to avoid a passing pipeline when no tests are found in the future. 
- Configures the validation of the remote-mqpu target to always use MLIR compilation.

Installer validation on ubuntu is just missing the documented runtime libraries. 
For the rhel8 validation, we need make sure bashrc is invoked, since it apparently is not executed by default.
Publishing workflow: https://github.com/NVIDIA/cuda-quantum/actions/runs/8020426619

Note: We do have a bug in the JIT compilation used for the remote-mqpu backend when MLIR generation is disabled. 
Since the primary scenario for us is to have MLIR generation enabled, I updated the validation script to always explicitly require MLIR compilation.


.github/workflows/publishing.yml
```diff
@@ -768,6 +769,7 @@ jobs:
       - name: Install and run sanity checks
         shell: bash
         run: |
+          if [ -f ~/.bashrc ]; then . ~/.bashrc; fi
```
bmhowe23, line 772: I'm wondering if that would source `.bashrc` twice in some platforms, which might be confusing to debug in the future. The bash man page offers an alternative that might work across a few different platforms better:
```
       When bash is started non-interactively, to run a shell script, for example, it  looks  for  the
       variable  BASH_ENV  in the environment, expands its value if it appears there, and uses the ex‐
       panded value as the name of a file to read and execute.  Bash behaves as if the following  com‐
       mand were executed:
              if [ -n "$BASH_ENV" ]; then . "$BASH_ENV"; fi
       but the value of the PATH variable is not used to search for the filename.
```

```suggestion
        env: |
          BASH_ENV: ~/.bashrc
        run: |
```
I tested this locally on my machine, and it works, but I can't guarantee that I got the yml syntax correct.

bmhowe23 (approved):
Thanks. I offered one comment, but I'll leave it up to you whether you want to take the suggestion or not.

scripts/validate_container.sh
```diff
@@ -170,49 +170,53 @@ do
...
+            fi
 
         else
             echo "Testing on $t target..."
-            if [ "$t" == "default" ]; then 
-                nvq++ $ex && status=$?
-            else
-                nvq++ $ex --target $t && status=$?
-            fi
+            nvq++ $ex $target_flag && status=$?
```
bmhowe23, line 216: Won't `status=$?` get skipped if `nvq++` fails? Should it be a `;` instead of `&&`?

PR #1275, opened 2024-02-22
Update NGC CLI URL in Install Guide

bmhowe23:
The NGC CLI URL has changed. The team will make a redirect for the old URL, but we should probably use the new URL going forward.


bmhowe23:
They restored the original link, and the other link I had must've been temporary because it no longer works. I will close this PR. No further action is needed.

Issue #1251, opened 2024-02-16
[experimental/python] Improvements, gaps and fixes: Part#2


khalatepradnya:
## Feature scope
- [x] Remove `jit=False`
- [x] All tests run and work with `jit=True`
- [x] Fix Pylance errors on quantum operations

## Bugs
- [x] Fix list element access
 https://gist.github.com/schweitzpgi/587e23fd9348202b8013d564bfde1493
- [x] Cannot perform logical operations on measurement results https://gist.github.com/bmhowe23/537a319f410ca2db2179addd8967b25b

## Gaps
- [x] Support previous control operations with same signature `cx` --> `x.ctrl`
- Import processing - retain previous behavior, for e.g. not require `import uccsd` separately: Won't fix now
Discrepancy between how the compiler processes imports and how python processes imports. Details see the draft here: https://github.com/NVIDIA/cuda-quantum/pull/1249
- [x] Spec enforcement
       Issues:
      - `sample` accepts `expectation()`
      - `observe` accepts measurements and executes

## Error handling and messaging
- [x] Better wording for type errors 
(Currently, when there is a type error, tool gives a "printing generic form" followed by a dump of the MLIR representation)
- [x] A simple typo like `x[qubits[0]]` (square brackets) produces quite a long error message (stack trace)
- [x] Misleading error if you pass an `np.ndarray` with an unsupported subtype as a kernel argument 
https://gist.github.com/anthony-santana/a060471777196ece1d93b1839fd16c6b
- [x] Show a comprehensive error message instead of the raw MLIR error when error occurs due to different scoping rules than Python

bettinaheim:
Re error handling: I would say never show the generated MLIR in an error message by default. Any print of MLIR should require setting an flag or environment variable explicitly.

zohimchandani:
Unable to run `sample_async`, code runs fine with `sample`: 

```
import cudaq
from cudaq import spin 
import numpy as np 
```

```

cudaq.set_target('nvidia-mqpu')
```

```
@cudaq.kernel(jit=True)
def kernel(num_qubits: int, params: np.ndarray):
    
    qubits = cudaq.qvector(num_qubits)
    
    param_counter = 0 
    for i in range(0, num_qubits): 
        
        if i+1 == num_qubits: 
            break
        else: 
            
            x.ctrl(qubits[i], qubits[i+1]) 
            
            ry(params[param_counter], qubits[i])
            param_counter += 1
            
[...]

PR #1279, opened 2024-02-22
[experimental/python] Support for legacy control operations in MLIR mode

khalatepradnya:
Addressing one gap in issue https://github.com/NVIDIA/cuda-quantum/issues/1251

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -972,7 +972,7 @@ class PyASTBridge(ast.NodeVisitor):
                         negCtrlBools[i] = c in self.controlNegations
                     negatedControlQubits = DenseBoolArrayAttr.get(negCtrlBools)
                     self.controlNegations.clear()
-                # Map "cx" to "XOp"
+                # Map `cx` to `XOp`...
                 opCtor = getattr(
                     quake, '{}Op'.format(node.func.id.title()[1:].upper()))
                 opCtor([], [],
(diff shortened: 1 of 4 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -960,6 +960,26 @@ def bodyBuilder(iterVar):
                 self.__applyQuantumOperation(node.func.id, [], qubitTargets)
                 return
 
+            if node.func.id in ["ch", "cx", "cy", "cz", "cs", "ct"]:
+                # These are single target controlled quantum operation
+                target = self.popValue()
+                # Should be number of arguments minus one for the controls
+                controls = [self.popValue() for i in range(len(node.args) - 1)]
```
amccaskey, line 967: Here we should check that self.valueStack has 2 elements and be careful about slurping up all values when these operations only support a single control and a single target.

khalatepradnya: Fixed in [dd363e9](https://github.com/NVIDIA/cuda-quantum/pull/1279/commits/dd363e9bd59b3622c5dc053ef61ea3374f6d7a18)

python/tests/kernel/test_control_operations.py
```diff
@@ -0,0 +1,381 @@
...
+        cx(qubits[0], qubits[1])
+        # 2-qubit controlled operation with control in 1-state.
+        x(qubits[2])
+        cx(qubits[2], qubits[3])
+
+        # Overload 2: list of control qubits, one target.
+        # The last qubit in `qubits` is still in 0-state, as is our
+        # `control` register. A controlled gate between them should
+        # have no impact.
+        cx([controls[0], controls[1]], qubits[4])
```
amccaskey, line 36: I guess this is a question we'll have to raise at the language meeting. I see that you stuck here to the 2 operands idea, but this is still allowing a multi-control X operation, which I was trying to avoid (really this is a `ccx` operation). I think we should only allow 2 qref operands for these new legacy `c` operations.

khalatepradnya: Q: The behavior should be consistent across the modes, builder and MLIR, (backward compatibility breaks?), or only the new mode enforces this check (inconsistency?). 
We can discuss in language meeting.

amccaskey: I don't care if it stays for now

amccaskey (approved):

PR #1280, opened 2024-02-23
Add ability to clean up simulation state in the runtime

1tnguyen:
### Description

Since both the `ExecutionManager` and `Simulator` keep track of the simulation state (instruction and allocation/deallocation queues, number of qubits, etc.), an early exit, e.g., exceptions, will leave some remnants. 

Note: this will only be a problem if these exceptions are handled.    

This PR:

- Adds reset functions to support clearing the static `ExecutionManager` and `Simulator` instances.

- Resets these states when QPU daemon handles simulation requests.

- Adds a test.


1tnguyen:
Superseded by https://github.com/NVIDIA/cuda-quantum/pull/1292. Closing this.

PR #1282, opened 2024-02-23
Move AST-to-MLIR compilation to the call operator.

amccaskey:
We really need to change when / where the actual AST-to-MLIR JIT happens. As of right now all library kernels are JIT compiled when they are imported - and this occurs even if they are not necessarily needed. For example, the UCCSD compilation time takes ~.9 seconds on my machine, and this occurs every time `import cudaq` is called. It would be better if the kernels JIT compiled when they were called. 

This PR updates the `PyKernelDecorator` class to do AST-to-MLIR JIT when the kernel is invoked.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -77,6 +77,8 @@ testing = cudaq_runtime.testing
 
 
 def synthesize(kernel, *args):
+    # Compile if necessary, no-op if already compiled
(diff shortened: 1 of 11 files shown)


khalatepradnya (approved):
Looks good to me.

schweitzpgi (commented):
LGTM

PR #1286, opened 2024-02-23
Fix SymbolTable error with print statements

amccaskey:
Also implement `theta *= -1.0` aug-assign node.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -2292,10 +2292,10 @@ class PyASTBridge(ast.NodeVisitor):
                 return
             elif F64Type.isinstance(loaded.type):
                 if IntegerType.isinstance(value.type):
-                    value = arith.SIToFPOp(self.getFloatType(), value).result 
-                res = arith.MulFOp(loaded, value).result 
+                    value = arith.SIToFPOp(self.getFloatType(), value).result
+                res = arith.MulFOp(loaded, value).result
                 cc.StoreOp(res, target)
-                return 
+                return
             else:
                 self.emitFatalError("unhandled AugAssign.Mult types.", node)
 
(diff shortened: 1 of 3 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2227,6 +2227,12 @@ def visit_AugAssign(self, node):
                 res = arith.MulIOp(loaded, value).result
                 cc.StoreOp(res, target)
                 return
+            elif F64Type.isinstance(loaded.type):
+                if IntegerType.isinstance(value.type):
+                    value = arith.SIToFPOp(self.getFloatType(), value).result
+                res = arith.MulFOp(loaded, value).result
+                cc.StoreOp(res, target)
+                return
```
khalatepradnya, line 2235: If I understand correctly, this logic is performing multiply and assignment operation... can we add a test that demonstrates this newly supported op - will be useful when documenting what's allowed and what's not.

khalatepradnya: This patch will test it and also address a `FIXME`
```
diff --git a/python/cudaq/kernels/uccsd.py b/python/cudaq/kernels/uccsd.py
index a6ad363cc..1c1ba66b4 100644
--- a/python/cudaq/kernels/uccsd.py
+++ b/python/cudaq/kernels/uccsd.py
@@ -170,15 +170,14 @@ def double_excitation_opt(qubits: cudaq.qview, p_occ: int, q_occ: int,
         j_occ = q_occ
         a_virt = s_virt
         b_virt = r_virt
-        # theta *= -1.0 FIXME
-        theta = theta * -1.
+        theta *= -1.0
 
     elif (p_occ > q_occ) and (r_virt < s_virt):
         i_occ = q_occ
         j_occ = p_occ
         a_virt = r_virt
         b_virt = s_virt
-        theta = theta * -1.0
+        theta *= -1.0
     #Block I: x_i x_j x_a y_b + x_i x_j y_a x_b + x_i y_i y_a y_b - x_i y_j x_a x_b
     #Block II: - y_i x_j x_a x_b +y_i x_j y_a y_b - y_i x_j x_a x_b - y_i y_j y_a x_b
```

khalatepradnya (approved):

PR #1294, opened 2024-02-26
[experimental/python] Pylance stopgap

khalatepradnya:
Import all supported quantum operations, 
-- Stub / dummy definitions added from previous eager mode 

Potential fix to Pylance error captured in issue https://github.com/NVIDIA/cuda-quantum/issues/1251


docs/sphinx/examples/python_mlir/intro.py
```diff
@@ -1,4 +1,5 @@
 import cudaq
+from cudaq.qis import *
```
khalatepradnya, line 2: image

docs/sphinx/examples/python_mlir/intro.py
```diff
@@ -1,4 +1,5 @@
...
```
khalatepradnya, line 3: image

python/cudaq/qis/qis.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @classmethod
+    def adj(cls, *args):
+        """
+        Invoke the general adjoint version of the quantum operation. 
+        The arguments can contain float parameters (of the correct number according
+        to `get_num_parameters`) and quantum types (`qubit`, `qvector`, `qview`).
+        """
+        opName = cls.get_name()
+        parameters = list(args)[:cls.get_num_parameters()]
```
bettinaheim, line 112: Ah, that's what you use them for. 
Can't we look at the type of the parameter instead? 

More broadly, when are these implementations (ctrl and adj) executed? I.e. isn't there direct support for them in the ast bridge? If we had support before this change, what additional support does this enable or replace (I don't see anything removed)?

Is this implementation here to support the existing `c<op>` (e.g. cx) function name? If that's the purpose, why not simply translate it into <op>.ctrl here and let the compiler handle dealing with it?

python/cudaq/qis/qis.py
```diff
@@ -0,0 +1,271 @@
...
+
+        cudaq_runtime.applyQuantumOperation(opName, parameters, controls,
+                                            targets, False, SpinOperator())
+        for q in quantumArguments:
+            if isinstance(q, qubit) and q.is_negated():
+                x()(q)
+                q.reset_negation()
+
+    @classmethod
+    def adj(cls, *args):
```
bettinaheim, line 105: How does this deal with s and t?

python/cudaq/qis/qis.py
```diff
@@ -0,0 +1,271 @@
...
+        'get_name': staticmethod(lambda: 'rz'),
+        'get_num_parameters': staticmethod(lambda: 1)
+    })
+r1 = type(
+    'r1', (quantum_operation,), {
+        'get_name': staticmethod(lambda: 'r1'),
+        'get_num_parameters': staticmethod(lambda: 1)
+    })
+
+ch = type('ch', (quantum_operation,), {'get_name': staticmethod(lambda: 'ch')})
```
bettinaheim, line 154: I don't see any support for `<op>dg` (e.g. tdg) here, which is what is used for Adjoint(T) in the kernel builder. It would be nice to add that as well.

python/cudaq/qis/qis.py
amccaskey: I can answer these questions: 

None of these functions will ever be executed - they were from the eager mode implementation. They are potentially useful here for the Pylance problems since they define all our operations + docstrings succinctly. 

"Can't we look at the type of the parameter" - you could do that, but it would be messier than the current implementation, since you'll have to insert a number of `if isinstance(...)` calls, here you just have the information on the operation class and ask it for that information.

python/cudaq/qis/qis.py
amccaskey: In the new language, the equivalent is `t.adj(...)`, we can add convenience ops though to provide `tdg(...)` like we did for `x.ctrl(...) -> cx(...)`

bettinaheim: Yes, that's exactly my thinking - it would be good to make it as straightforward as possible to migrate code, and having a 1:1 equivalent here hence seems like a good thing to add.

python/cudaq/qis/qis.py
amccaskey: I think that's a good idea

khalatepradnya:
Created a new PR based on `main`: https://github.com/NVIDIA/cuda-quantum/pull/1317
Closing this one.

PR #1013, opened 2023-12-07
Adding new tutorials

zohimchandani:
### Description

Adding new tutorials and fixing typos in old ones.

PR #1295, opened 2024-02-26
MLIR updates for python examples

anthony-santana:
### Description

* Replaces all python builder code in examples and tutorials with MLIR mode
* Removes python_mlir and python_runtime folders from docs examples
* Adds notebooks from PR #1013


docs/sphinx/examples/python/advanced_vqe.py
```diff
@@ -1,22 +1,23 @@
...
 hamiltonian = 5.907 - 2.1433 * spin.x(0) * spin.x(1) - 2.1433 * spin.y(
     0) * spin.y(1) + .21829 * spin.z(0) - 6.125 * spin.z(1)
 
-kernel, thetas = cudaq.make_kernel(list)
-qubits = kernel.qalloc(2)
-kernel.x(qubits[0])
-kernel.ry(thetas[0], qubits[1])
-kernel.cx(qubits[1], qubits[0])
+
+@cudaq.kernel
```
khalatepradnya, line 14: _[At the risk for sounding like a broken record]_ Shouldn't this go into `python_mlir` directory?
I might have misunderstood the intention...

docs/sphinx/snippets/python/using/cudaq/platform/observe_mqpu.py
```diff
@@ -20,6 +20,15 @@
 kernel.x(qvec[0])
 kernel.ry(theta, qvec[1])
 kernel.cx(qvec[1], qvec[0])
+
+# FIXME:: why does this just hang??
+# @cudaq.kernel
```
khalatepradnya, line 25: Thanks for capturing this in a reproducible test!

docs/sphinx/examples/python/advanced_vqe.py
khalatepradnya: Looking at the latest commits. this seems to be resolved.

docs/sphinx/examples/python/advanced_vqe.py
```diff
@@ -29,10 +30,10 @@
 gradient = cudaq.gradients.CentralDifference()
 
 
-def objective_function(parameter_vector: List[float],
+def objective_function(parameter_vector: list[float],
```
khalatepradnya, line 33: Note that this will work only in Python 3.9+. We support Python 3.8 as well and it needs `List` from `typing` module.
Consider keeping it so that same code works across all supported Python versions.

anthony-santana: Good catch, I wasn't even considering that. Reverting back now (I think it was only in this file and maybe 1 or 2 others)

docs/sphinx/examples/python/providers/nvcf_sample.py
```diff
@@ -0,0 +1,32 @@
...
+    # Place qubits in GHZ state.
+    h(qvector[0])
+    for qubit in range(qubit_count - 1):
+        x.ctrl(qvector[qubit], qvector[qubit + 1])
+    # Measure.
+    mz(qvector)
+
+
+qubit_count = 50
+counts = cudaq.sample(ghz, qubit_count, shots_count=100)
```
khalatepradnya, line 31: Q: Is this expected to work as of today?
Anyone tried it? @1tnguyen

anthony-santana: I don't have nvcf configured locally so I was unable to test on the backend. I only "tested" by running this at ~5 qubits on the nvidia backend

docs/sphinx/examples/python/bernstein_vazirani.py
```diff
@@ -1,113 +1,83 @@
-import argparse
 import cudaq
 import random
 
+from typing import List
+
+cudaq.enable_jit()
```
khalatepradnya, line 6: This is no longer required. I think CI is failing because of that.

khalatepradnya: Applied fix in 1a9e5d6

docs/sphinx/examples/python/tutorials/cost_minimization.ipynb
```diff
@@ -30,20 +29,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "qubit_count = 1\n",
-    "\n",
     "# Initialize a kernel/ ansatz and variational parameters.\n",
-    "kernel, parameters = cudaq.make_kernel(list)\n",
+    "@cudaq.kernel\n",
+    "def kernel(angles: list[float]):\n",
```
khalatepradnya, line 34: list --> List

docs/sphinx/examples/python/bernstein_vazirani.py
```diff
@@ -1,113 +1,81 @@
...
+# If you have a NVIDIA GPU you can use this example to see
+# that the GPU-accelerated backends can easily handle a
+# larger number of qubits compared the CPU-only backend.
+
+# Depending on the available memory on your GPU, you can
+# set the number of qubits to around 30 qubits, and un-comment
+# the `cudaq.set_target(nvidia)` line.
+
+# Note: Without setting the target to the `nvidia` backend,
+# a 30 qubit simulation simply seems to hang; that is
```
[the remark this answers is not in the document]
khalatepradnya: How about "appears stalled"?

docs/sphinx/examples/python/bernstein_vazirani.py
```diff
@@ -1,113 +1,81 @@
...
+
 
 def random_bitstring(length: int):
-    bitstring = ""
-    for bit in range(length):
-        bitstring += str(random.randint(0, 1))
+    bitstring = []
+    for _ in range(length):
+        bitstring.append(random.randint(0, 1))
     return bitstring
```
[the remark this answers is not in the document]
anthony-santana: I agree that it's messy to call this list a bitstring. I didn't do the original translation (@khalatepradnya or @amccaskey  ?), but am I correct in thinking the type was changed because we don't support strings as kernel arguments? My issue is that Bernstein Vazirani deals with finding hidden strings; so "bitstring" is the most technically precise name from a physics POV

khalatepradnya (approved):

docs/sphinx/examples/python/bernstein_vazirani.py
```diff
@@ -1,113 +1,81 @@
...
-        cudaq.set_target(args.target)
-
-    print(f"Running on target {cudaq.get_target().name} ...")
-    kernel, hidden_bitstring = bernstein_vazirani(qubit_count)
-    result = cudaq.sample(kernel)
-
-    print(f"encoded bitstring = {hidden_bitstring}")
-    print(f"measured state = {result.most_probable()}")
-    print(f"Were we successful? {hidden_bitstring == result.most_probable()}")
+print(bernstein_vazirani)
```
bettinaheim, line 74: *Please* do not print kernels - we don't want to confront users with MLIR bitcode as part of examples.

docs/sphinx/examples/python/bernstein_vazirani.py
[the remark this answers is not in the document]
schweitzpgi: That's good. Thanks, Anthony.

docs/sphinx/examples/python/bernstein_vazirani.py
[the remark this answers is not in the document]
schweitzpgi: Adding some points.

- bitstring is also a computer science term
   - that term implies a significance to the order, which a list does not; not clear if that is significant here, but we've had issues before
- string is a computer science term and within many contexts is indicative of the data structure (Python being one of those contexts)
- we should be careful not to mislead or confuse users with terminology

To be clear, I don't have any strong objection. Just want it to be clear as this is an example.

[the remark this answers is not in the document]
anthony-santana: I completely agree with your points. What are thoughts on: `hidden_bits`, `bit_list`, `bits` ? I can't think of anything better that satisfies both CS/Physics, but happy for suggestions 😃

khalatepradnya:
This PR will be merged later into 'main'. Closing...

docs/sphinx/examples/python/bernstein_vazirani.py
bettinaheim: How about instead using the new draw functionality introduced in https://github.com/NVIDIA/cuda-quantum/pull/1299?

docs/sphinx/examples/python/providers/nvcf_sample.py
```diff
@@ -0,0 +1,33 @@
+import cudaq
```
bettinaheim, line 1: I think this file needs to be renamed to nvqc_sample.py

docs/sphinx/examples/python/providers/nvcf_state.py
```diff
@@ -0,0 +1,30 @@
+import cudaq
```
bettinaheim, line 1: file name change

docs/sphinx/examples/python/providers/nvqc_vqe.py
```diff
@@ -30,16 +34,16 @@
 
 def opt_gradient(parameter_vector):
     # Evaluate energy and gradient on different remote QPUs
-    # (i.e., concurrent job submissions to NVQC)
-    energy_future = cudaq.observe_async(kernel,
+    # (i.e., concurrent job submissions to NVCF)
```
bettinaheim, line 37:
```suggestion
    # (i.e., concurrent job submissions to NVQC)
```

docs/sphinx/examples/python/qaoa_maxcut.py
```diff
@@ -26,28 +28,26 @@
 parameter_count: int = 2 * layer_count
 
 
-def kernel_qaoa() -> cudaq.Kernel:
+@cudaq.kernel
+def kernel_qaoa(qubit_count: int, layer_count: int, thetas: List):
```
bettinaheim, line 32: The data type needs to be fully defined - I believe this sample should not longer work as is:
```suggestion
def kernel_qaoa(qubit_count: int, layer_count: int, thetas: List<float>):
```

docs/sphinx/examples/python/qaoa_maxcut.py
```diff
@@ -58,17 +58,21 @@ def kernel_qaoa() -> cudaq.Kernel:
...
-    parameter_count=parameter_count)
+
+# Define the objective, return `<state(params) | H | state(params)>`
+def objective(parameters):
+    return cudaq.observe(kernel_qaoa, hamiltonian, qubit_count, layer_count,
+                         parameters).expectation()
+
+
+# Optimize!
+optimal_expectation, optimal_parameters = optimizer.optimize(
```
bettinaheim, line 69: What was the reason here for no longer using cudaq.vqe?

docs/sphinx/examples/python/bernstein_vazirani.py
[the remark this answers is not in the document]
bettinaheim: either one sounds good to me

docs/sphinx/examples/python/qaoa_maxcut.py
anthony-santana: To be honest, I don't remember updating this file myself but I must have since I don't see it tracked for anyone else on the excel file... So the short answer is I don't know, but I will try to update it now 😃

docs/sphinx/examples/python/bernstein_vazirani.py
anthony-santana: I'm going to add the cudaq.draw to our python notebooks so that we can show it in the docs. Removing the print here.

docs/sphinx/examples/python/qaoa_maxcut.py
anthony-santana: Ok so this isn't as straightforward of an update as I thought it'd be because we now have arguments to our kernel, which need to be mapped by vqe. I can't seem to get that working, so will merge with this and follow-up with a change later

PR #1297, opened 2024-02-26
Update NVCF logging

1tnguyen:
### Description

NVQC logging enhancements:

- Fixed the error message when the number of GPUs is invalid: remove the function Ids, mention `ngpus`.

- Add a metadata struct to capture server execution: timing data and NVIDIA GPU device info.

- Log the info client side:

Example log:
```
===== NVQC Device Info ===== 
GPU Device Name: "NVIDIA A100-SXM4-80GB"
CUDA Driver Version / Runtime Version: 12.0 / 11.8
Total global memory (GB): 79.2
Memory Clock Rate (MHz): 1593.000
GPU Clock Rate (MHz): 1410.000
================================== 
```

```
===== NVQC Execution Timing ===== 
 - Pre-processing: 11 milliseconds 
 - Execution: 68 milliseconds 
================================== 
```


runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -97,6 +97,14 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
   // Server to exit after each job request.
   // Note: this doesn't apply to ping ("/") endpoint.
   bool exitAfterJob = false;
+  // Time-point data
+  std::optional<std::chrono::time_point<std::chrono::system_clock>>
+      requestStart;
+  std::optional<std::chrono::time_point<std::chrono::system_clock>>
+      simulationStart;
+  std::optional<std::chrono::time_point<std::chrono::system_clock>>
+      simulationEnd;
```
bmhowe23, line 106: Nit: I would recommend using `high_resolution_clock` instead of `system_clock` because the system clock can be updated with NTP (or any other system time updates) whereas `high_resolution_clock` is better suited for these "elapsed time" situations. It may not be worth changing here, but I wanted to bring it up for any potential future timing code.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -279,6 +279,12 @@ class RemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
 /// REST client submitting jobs to NVCF-hosted `cudaq-qpud` service.
 class NvcfRuntimeClient : public RemoteRestRuntimeClient {
 private:
+  // None: Don't log; Info: basic info; Trace: Timing data per invocation.
+  enum LogLevel : int { None = 0, Info, Trace };
```
bmhowe23, line 283: Nit: I would recommend use of scoped enumerations.
```suggestion
  enum class LogLevel : int { None = 0, Info, Trace };
```

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -357,6 +363,21 @@ class NvcfRuntimeClient : public RemoteRestRuntimeClient {
 public:
   virtual void setConfig(
       const std::unordered_map<std::string, std::string> &configs) override {
+    {
+      // Check if user set a specific log level (e.g., disable logging)
+      if (auto logConfigEnv = std::getenv("NVQC_LOG_LEVEL")) {
```
bmhowe23, line 368: This seems reasonable to me, but it would be good if @bettinaheim would weigh in here. (Maybe you've already covered this with her offline?)

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -371,7 +392,15 @@ class NvcfRuntimeClient : public RemoteRestRuntimeClient {
...
       const auto funcIdIter = configs.find("function-id");
       if (funcIdIter != configs.end()) {
+        // User overrides a specific function Id.
         m_functionId = funcIdIter->second;
+        if (m_logLevel > LogLevel::None) {
+          // Print out the configuration
+          printf("%s\n",
+                 fmt::format("Submitting jobs to NVQC using function Id {}.",
+                             m_functionId)
+                     .c_str());
```
bmhowe23, line 402: Why not use `fmt::print()` instead? (Same question for other usages.)

bmhowe23 (commented):
This LGTM (pending a few nits), but hopefully others (@bettinaheim, @efratshabtai ) can weigh in to see if the new `NVQC_LOG_LEVEL` satisfies their request?

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
1tnguyen: Agree. Fixed as recommended.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
1tnguyen: I guess because `fmt::print` is not very widely used in our codebase yet (it should be, I think) so my hands just typed `printf` :-)
Converted all `printf` usages in this file to `fmt::print`.

bmhowe23 (approved):

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
bmhowe23: > This seems reasonable to me, but it would be good if @bettinaheim would weigh in here. (Maybe you've already covered this with her offline?)

We'll cover this in a separate PR if anything else is needed. Marking as resolved for now.

PR #1298, opened 2024-02-27
Change test to eliminate strange failure.

schweitzpgi:
It is sufficient to run this test against the C++17 version of the headers once. This is now achieved by having a distinct task run all tests as C++17. There is no added value in performing the C++17 test multiple times in multiple CI tasks.


targettests/execution/auto_kernel.cpp
```diff
@@ -6,10 +6,7 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: nvq++ -std=c++17 --enable-mlir %s -o %t && %t | FileCheck %s
-// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
```
schweitzpgi, line 10: This line aborts when %cpp_std is not defined.

bmhowe23: > This line aborts when %cpp_std is not defined.

Isn't this line in many of our tests? What's special about this one?

bettinaheim: ech, bash is not very helpful - I think the command exit code is 0 if the value is  not an integer.
This should work as intended:
`if [ -n  "$(echo %cpp_std)" ] && [ $(echo %cpp_std | cut -c9- ) -ge 20 ]; then`

bmhowe23 (approved):

PR #1299, opened 2024-02-27
Finish up circuit drawing, expose to python

amccaskey:
Enable circuit drawing apart from library mode. Supersedes #1201 , picked up that work to help out @boschmitt on paternity leave.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_draw.cpp
+++ b/python/runtime/cudaq/algorithms/py_draw.cpp
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include <pybind11/complex.h>
+#include <pybind11/stl.h>
+
+#include "utils/OpaqueArguments.h"
+#include "cudaq/algorithms/draw.h"
+
+namespace cudaq {
+
+/// @brief Run `cudaq::get_state` on the provided kernel and spin operator.
+std::string pyDraw(kernel_builder<> &kernel, py::args args) {
+  // Ensure the user input is correct.
+  auto validatedArgs = validateInputArguments(kernel, args);
+  OpaqueArguments argData;
+  packArgs(argData, validatedArgs);
+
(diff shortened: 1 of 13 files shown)


python/runtime/cudaq/algorithms/py_draw.cpp
```diff
@@ -0,0 +1,34 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include <pybind11/complex.h>
+#include <pybind11/stl.h>
+
+#include "utils/OpaqueArguments.h"
+#include "cudaq/algorithms/draw.h"
```
schweitzpgi: Whenever I see includes in this order, I immediately think it's a bug. Guess I've been doing LLVM for too long.

At any rate, the LLVM argument of always including local sources first is a good one as it prevents one from writing broken .h files that have dependences on other headers that the file doesn't include itself.

python/tests/unittests/test_kernel_builder.py
```diff
@@ -978,6 +978,71 @@ def test_fermionic_swap_op():
...
+┤●      ├┤ ry(3.142) ├───────●──────┤ ry(-3.142) ├
+│  swap │├───────────┴╮╭───────────╮╰────────────╯
+┤●      ├┤ rz(-3.142) ├┤ rz(3.142) ├──────────────
+│       │╰────────────╯╰───────────╯              
+┤>      ├─────────────────────────────────────────
+╰───────╯                                         
+'''
+
+    # FIXME add this later
+    # assert circuit == expected_str
```
bettinaheim, line 1045: Looks like there is an assert missing.

bettinaheim (commented):
In principle, lgtm, but there is a todo in the test to enable it.

python/runtime/cudaq/algorithms/py_draw.cpp
amccaskey: Will fix this

python/tests/unittests/test_kernel_builder.py
amccaskey: Turns out we had a bug in lowering S<adj> and T<adj> to QIR, and that's why the I couldn't get the strings to equal each other.

bettinaheim: Good find!

bettinaheim (approved):
Thanks!

PR #1305, opened 2024-02-28
Bug bash grab bag fixes

amccaskey:
This PR addresses most of the issues brought up in today's bug bash: 

* do not generate mlir for docstrings
* do not push result `i1` value to stack for occurrences of `mz` with no assignment node
* error for erroneous `cudaq.qubit(N)` constructor
* error for unknown `cudaq.*` function or class constructors (e.g. `cudaq.qreg`)
* Support `if i == True`, `True` `ast.Constant` node was not being handled
* Fix `cmpIOp` error for lhs being a float type
* Update error message to not use the word symbol table
* Fix bug where source location for error was not right (`cudaq.QuakeValue` induced error)
* Allow `list[int]` passed at runtime to argument type `list[float]` (automated casting in call operator)
* Disallow generic `list` or `List`
* Fix cmpi error for  something like `if mz(q) == 1` (incompatible integer bit widths)
* Implement aug assign for floats
* Implement returning floats from kernels
* Error message for invalid return value types
* Fix Sample / SampleAsync Seg Fault (TN-1)


khalatepradnya (approved):
Looks great!

PR #1307, opened 2024-02-28
[rt] Refactor

boschmitt:
### Description
This PR aims to disentangle (pun intended) various components of the runtime.
Before talking about some of the changes, it will be helpful to broadly describe
what happens at runtime.

Broadly speaking, we can compile CUDA Quantum programs in two ways:

  1. Using an off-the-shelf C++ compiler, e.g., "clang" or "gcc." We call this
     case library mode, and it allows the user to use the full power of C++ when
     writing kernel. (While this ability is attractive for exploration and
     prototyping, it can be a double-edged sword as the user might end up with
     kernels incompatible with QPU execution).

  2. Compiled (or JIT) mode uses  "nvq++" MLIR pipeline. This pipeline
     translates kernel code into our MLIR dialects and translates them for QPU
     execution. (The subset of supported C++ is still growing, so the users
     might feel over-constrained by this method.)

The execution of a CUDA Quantum program changes depending on the chosen method.
The following diagram provides a bird's eye view of runtime events when 
targeting a QVM, i.e., a simulator.

```
                                      * Library mode:
┌──┬────────────────────────────────┐ When this kernel is called by one of the
│1 │#include <cudaq.h>              │ execution primitives. e.g., cudaq::sample,
│2 │                                │ the quantum instructions are processed by
│3 │using namespace cudaq;          │ the ExecutionManager.
│4 │                                │ 
│5 │__qpu__ void foo() {            │    ┌─────┐
│6 │  qvector q(2);                 │    │     │
│7 │  h(q[0]);                      │    │ QIS │
│8 │  x<cudaq::control>(q[0], q[1]);│    │     │
│9 │}                               │    └──┬──┘
└──┴────────────────────────────────┘    ┌──▽─────────────┐
                                         │ExecutionManager│
* Compiled (JIT) mode:                   └──┬─────────────┘
In this case, the kernel is compiled        │   The ExecutionManager main
by nvq++, and instructions become NVQIR     │   resposibility is to handle
calls:                                      │   regions, the control region and
                                            │   the adjoint region, which are
              ┌───────┐                     │   started by cudaq::control and
              │ NVQIR │                     │   cudaq::ajoint, respectively. It
              └┬──────┘                     │   also does other taks such as
               │                            │   queueing gates and allocations 
               │                            │   for later execution.
               │                            │
               │                            │
              ┌▽────────────────────────────▽┐    
              │       CircuitSimulator       │ The CircuitSimulator further
              └──────────────┬───────────────┘ processes instructions before
                             │                 sending them to the backend
                             │                 simulators.
              ┌──────────────▽───────────────┐              
              │          Simulator           │              
              └──────────────────────────────┘              
```

#### Changes
* We change a couple of API names to hide components' internal behavior. For example, the `qudit` is an RAI-like class that asks the `ExecutionManager` for quantum memory. It does not need to know how indices (unique identifiers) are managed, and thus, we change the name of the API from `getAvailableIndex` to `allocateQudit`. (We do something similar to deallocate.)
* The `CircuitSimulator` layer replicates many functionalities of the `ExecutionManager` e.g., queueing gates and deferring allocations. This is necessary because compiled kernels bypass it. We deduplicate these functionalities, leaving them only at the `CircuitSimulator` layer. (The `ExecutionManager` only needs to handle regions.) 
  * This simplification allows us to remove various indirections that become unnecessary.
* Move the tracer from the `ExecutionManager` to the `CircuitSimulator`. This enable its functionalities (or the ones based one it) to work with compiled (JITed) kernels.
* Move functionality that requires knowing that the `CircuitSimulator` queue gates from the `ExecutionManager`. The different pieces of the runtime should not know about the internal behavior of others.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -188,10 +188,10 @@ public:
   virtual cudaq::observe_result observe(const cudaq::spin_op &term) = 0;
 
   /// @brief Allocate a single qubit, return the qubit as a logical index
-  virtual std::size_t allocateQubit() = 0;
+  virtual std::size_t allocateQudit() = 0;
 
   /// @brief Allocate `count` qubits.
-  virtual std::vector<std::size_t> allocateQubits(const std::size_t count) = 0;
+  virtual std::vector<std::size_t> allocateQudits(const std::size_t count) = 0;
 
   /// @brief Deallocate the qubit with give unique index
   virtual void deallocate(const std::size_t qubitIdx) = 0;
@@ -770,8 +770,8 @@ public:
                              "observe(const cudaq::spin_op &).");
   }
 
-  /// @brief Allocate a single qubit, return the qubit as a logical index
-  std::size_t allocateQubit() override {
+  /// @brief Allocate a qudit, and returns its unique identifier.
+  std::size_t allocateQudit() override {
     // Get a new qubit index
     auto newIdx = tracker.getNextIndex();
 
@@ -805,9 +805,8 @@ public:
     return newIdx;
   }
 
-  /// @brief Allocate `count` qubits.
-  std::vector<std::size_t> allocateQubits(std::size_t count) override {
-    ScopedTraceWithContext("allocateQubits", count);
+  /// @brief Allocate `count` qudits.
+  std::vector<std::size_t> allocateQudits(std::size_t count) override {
     std::vector<std::size_t> qubits;
     for (std::size_t i = 0; i < count; i++)
       qubits.emplace_back(tracker.getNextIndex());
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -193,11 +193,11 @@ public:
   /// @brief Allocate `count` qubits.
   virtual std::vector<std::size_t> allocateQudits(const std::size_t count) = 0;
 
-  /// @brief Deallocate the qubit with give unique index
-  virtual void deallocate(const std::size_t qubitIdx) = 0;
+  /// @brief Deallocate qudit with give unique identifier.
+  virtual void deallocateQudit(const std::size_t uid) = 0;
 
-  /// @brief Deallocate all the provided qubits.
-  virtual void deallocateQubits(const std::vector<std::size_t> &qubits) = 0;
+  /// @brief Deallocate all the provided qudits.
+  virtual void deallocateQudits(const std::vector<std::size_t> &uids) = 0;
 
   /// @brief Reset the current execution context.
   virtual void resetExecutionContext() = 0;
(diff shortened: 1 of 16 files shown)


runtime/nvqir/NVQIR.cpp
```diff
@@ -207,16 +207,16 @@ void __quantum__rt__qubit_release_array(Array *arr) {
...
   auto qubit = std::make_unique<Qubit>(qubitIdx);
   nvqir::allocatedSingleQubits.emplace_back(std::move(qubit));
   return nvqir::allocatedSingleQubits.back().get();
 }
 
 /// @brief Once done, release that qubit
 void __quantum__rt__qubit_release(Qubit *q) {
-  ScopedTraceWithContext("NVQIR::release_qubit");
-  nvqir::getCircuitSimulatorInternal()->deallocate(q->idx);
+  cudaq::ScopedTrace trace("NVQIR::release_qubit");
```
bmhowe23, line 218: I'd suggest using `ScopedTraceWithContext` as that will show the source file and line number.
```suggestion
  ScopedTraceWithContext("NVQIR::release_qubit");
```

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1163,6 +1172,57 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
+    flushGateQueue();
+
+    if (executionContext->canHandleObserve) {
+      auto result = observe(*executionContext->spin.value());
+      return {*result.expectationValue, cudaq::sample_result(result)};
+    }
+
+    assert(op.num_terms() == 1 && "Number of terms is not 1.");
+
+    cudaq::info("Measure {}", op.to_string(false));
```
bmhowe23, line 1186: If this is called in a loop, then it would probably be worth adding this to save throughput.
```suggestion
    if (cudaq::details::should_log(cudaq::details::LogLevel::info))
      cudaq::info("Measure {}", op.to_string(false));
```

runtime/nvqir/CircuitSimulator.h
```diff
@@ -843,20 +838,25 @@ class CircuitSimulatorBase : public CircuitSimulator {
   }
 
   /// @brief Deallocate the qubit with give index
-  void deallocate(const std::size_t qubitIdx) override {
+  void deallocateQudit(const std::size_t uid) override {
```
bmhowe23, line 841: Nit: maybe it's just me, but I think that `qubits` or `qudits` is a bit more descriptive than `uids`. I'm curious what the "u" in `uids` represents.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -188,16 +188,16 @@ class CircuitSimulator {
   virtual cudaq::observe_result observe(const cudaq::spin_op &term) = 0;
 
   /// @brief Allocate a single qubit, return the qubit as a logical index
-  virtual std::size_t allocateQubit() = 0;
+  virtual std::size_t allocateQudit() = 0;
```
amccaskey, line 191: Should we use qudit here? Thus far the CircuitSimulator is tailored to qubits.

boschmitt:
This PR is broken beyond repair. Changes to the runtime require that I rethink the strategy to simplify the runtime.

PR #1311, opened 2024-02-29
Update NVQC Examples

1tnguyen:
### Description

- Only keep the note about using the environment variable to set the API key since it's the recommended method.

- Add an `mqpu` example.


schweitzpgi:
Q: was the config file option removed? planned to be removed? or just not being documented?

1tnguyen:
> Q: was the config file option removed? planned to be removed? or just not being documented?

The last one. In the doc, the environment variable option is the recommended method in the quick start guide. The config file option will only be mentioned much later in the doc in the extended section.

khalatepradnya (approved):
Looks good.

PR #1312, opened 2024-02-29
Enable the Python features from 'experimental/python' in mainline

khalatepradnya:
### Description
Mega-merge of 'experimental/python' into 'main'


.github/workflows/ci.yml
```diff
@@ -206,7 +206,7 @@ jobs:
     strategy:
       matrix:
         platform: [amd64, arm64]
-        python_version: ['3.9', '3.11']
+        python_version: ['3.8', '3.11']
```
bmhowe23, line 209: Any details on this change? I suspect this is the reason the CI says that not all the tests are passing.

bmhowe23: Could we / should we change this back to 3.9, at least for the purposes of the merge?

bmhowe23: Suggest reverting this back to what is on main right now, and we will sort out the rest post-merge.
```suggestion
        python_version: ['3.9', '3.11']
```

.github/workflows/ci.yml
```diff
@@ -206,7 +206,7 @@ jobs:
     strategy:
       matrix:
         platform: [amd64, arm64]
-        python_version: ['3.8', '3.11']
+        python_version: ['3.9', '3.11']
```
khalatepradnya, line 209: While working on the new features, we realized that some tests and functionality didn't work on Python 3.8 (things like `list[float]`, `graphlib`), and we didn't catch it until deployment phase. So, this change was made to ensure we check on the lowest supported version...

bmhowe23: OK, I generated https://github.com/NVIDIA/cuda-quantum/issues/1313 so that we can consider changing the branch protection rules and the CI to go back to `3.8` once this is merged.

bmhowe23 (approved):
+100

PR #1314, opened 2024-02-29
Implement proper Pythonic capture of variables from parent scope

amccaskey:
Capture `int`, `float`, `bool`, or `list[int|float|bool]` from parent scope. Track potential variable changes and update the MLIR code accordingly. 
```python
n = 5
 
@cudaq.kernel
def kernel():
    q = cudaq.qvector(n)
    x(q)
 
counts = cudaq.sample(kernel)
assert '1' * n in counts

# Change the variable, code should reflect the change
n = 7
counts = cudaq.sample(kernel)
assert '1' * n in counts
```


khalatepradnya (approved):
Looks good.
(Same as previously approved PR#1309)

PR #1318, opened 2024-03-01
Refresh remote seeds for executions within a client session

1tnguyen:
### Description

Currently, when sending REST requests to a remote QPUD, the seed value is set to `cudaq::get_random_seed()`, hence all kernel executions will be using the same seed value. This deviates from the local execution model, whereby the static random number generator in the simulator is seeded **once** when `set_random_seed` is called. 

This PR fixed that difference by letting the client to seed the server using a random number generator seeded by `cudaq` seed value.

Note: in general, we don't expect an absolute seed value to have the same effect between local execution and remote execution. Hence, we can randomly generate the server seed using a random sequence seeded by the `cudaq` seed.

### DIFF
--- a/python/tests/remote/test_remote_platform.py
+++ b/python/tests/remote/test_remote_platform.py
@@ -170,6 +170,41 @@ def test_multi_qpus_kernel():
     check_multi_qpus(parameterized_ansatz)
 
 
+# Check randomness and repeatability by setting the seed value
+def test_seed():
+    kernel = cudaq.make_kernel()
+    qubit = kernel.qalloc()
+    kernel.h(qubit)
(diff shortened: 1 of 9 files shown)


python/tests/remote/test_remote_platform.py
```diff
@@ -170,6 +170,41 @@ def parameterized_ansatz(theta: float):
     check_multi_qpus(parameterized_ansatz)
 
 
+# Check randomness and repeatability by setting the seed value
+def test_seed():
```
bmhowe23, line 174: Nice test 👍

bmhowe23 (approved):

PR #1323, opened 2024-03-01
Elevate capture vars to stack slot at function entry block start

amccaskey:
Previously, captured variables were treated as constant MLIR values created in whatever scope they were first used. This means they cannot be used in other function scopes and they can't be set / changed (though they are still captured by value). This PR fixes this by elevating capture variables to stack variables at the start of the kernel entry block.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -42,6 +42,12 @@ class PyScopedSymbolTable(object):
     def popScope(self):
         self.symbolTable.pop()
 
+    def add(self, symbol, value, level=-1):
+        """
+        Add a symbol to the scoped symbol table at any scope level.
+        """
+        self.symbolTable[level][symbol] = value
+
     def __contains__(self, symbol):
         for st in reversed(self.symbolTable):
             if symbol in st:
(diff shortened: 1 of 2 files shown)


schweitzpgi:
Seems like a good update.

This should match the LEGB scoping rules of python better. Since variables are bound lazily in python, we do have the possibility that the user can write code that will access a python variable that is not defined yet in the L scope. Specifically, it is defined along some paths but not all. Python will throw an exception rather than access undefined values. That's beyond the scope of this PR, but we'll probably want to simulate the dictionary lookup somehow. Throwing an exception seems like a poor idea since it's across the device boundary, but maybe we can add error handling on kernels.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -42,6 +42,12 @@ def pushScope(self):
     def popScope(self):
         self.symbolTable.pop()
 
+    def add(self, symbol, value, level=-1):
```
khalatepradnya, line 45: Suggest changing function name to be more descriptive, something like `addSymbol` or `addToSymbolTable`...

khalatepradnya (approved):

PR #1326, opened 2024-03-01
[Core] A rollup of misc bug fixes wrt argument and return values

schweitzpgi:
- Reorganize and add more end-to-end tests for passing arguments and returning result values. The tests include cases that are not yet implemented. Those cases are inactive until implemented.

- Fix bug with sret returns.

- Fix bugs with small struct ABI passing on C++ side. TODO: it would be better to use the clang backend code, but it is not exported.

- Fix bug in vector result return handler.

- Fix GKE to handle argument register pairs (needed for small struct)

- Fix bug with multi-result kernels

- Corrections made to tests for changes.


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1106,6 +1108,37 @@ class GenerateKernelExecution
...
+          if (cudaq::opt::factory::isX86_64(module)) {
+            builder.create<cudaq::cc::StoreOp>(loc, arg, cast);
+            if (cudaq::opt::factory::structUsesTwoArguments(quakeTy)) {
+              auto cast = builder.create<cudaq::cc::CastOp>(
+                  loc, cudaq::cc::PointerType::get(builder.getI8Type()), tmp);
+              auto hiPtr = builder.create<cudaq::cc::ComputePtrOp>(
+                  loc, cudaq::cc::PointerType::get(builder.getI8Type()), cast,
+                  cudaq::cc::ComputePtrArg{8});
+              ++iter;
+              Value nextArg = *iter;
```
bmhowe23, line 1128: I'm curious why `iter` would be incremented but `idx` would not?

targettests/Kernel/signature-0.cpp
```diff
@@ -0,0 +1,121 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
bmhowe23, line 2: Nit: new files should probably have the right copyright date, though we should probably revisit how to handle these as a team.

bmhowe23 (approved):

lib/Optimizer/Transforms/GenKernelExecution.cpp
schweitzpgi: The purpose of that change and the others in this loop is that there are more physical arguments, potentially, than logical arguments. This is the case where a logical argument (idx) is 2 physical arguments (iter).

PR #1328, opened 2024-03-02
[NVQC] Update NCA ID and add ability to filter functions based on version numbers

1tnguyen:
### Description

- Update NCA ID to query/filter function endpoints and add the ability to override it with an environment variable.

- Extend the scope of `RestRequest`'s `version` field to capture generic Client-Server compatibility (not just schema per se). Update error messages and code comments accordingly.

- Based on discussion w/ @bmhowe23: use deployment environment variable as the mechanism to capture and communicate this version number between NVCF functions and clients: (1) this information is visible to clients (listing available functions) and (2) we can perform a sanity check during the deployment to ensure that this information is actually correct.

- Update NVQC function matcher to only match the correct version.

### Merging check list

- Before merge: Update all current deployments to have the environment variable `NVQC_REST_PAYLOAD_VERSION=1`, e.g., with `--container-environment-variable NVQC_REST_PAYLOAD_VERSION:1` (NGC CLI). We won't have any deployment check for this environment variable at this point. It's to make sure new clients (e.g., compile from source right after the merge) will discover these functions.

- After merge: when a new Publishing image is available, re-deploy new functions. At this point, `NVQC_REST_PAYLOAD_VERSION` will be checked at deployment time.


bmhowe23 (approved):

python/runtime/utils/PyRestRemoteClient.cpp
```diff
@@ -79,29 +79,49 @@ class PyNvcfRuntimeClient : public PyRestRemoteClient {
...
       }
-    }
-    return info;
+      return info;
+    };
+
+    const std::string baseCudaqNvcfFuncNamePrefix = "cuda_quantum";
+    const std::string versionedCudaqNvcfFuncNamePrefix =
+        baseCudaqNvcfFuncNamePrefix + "_" +
+        std::to_string(cudaq::RestRequest::REST_PAYLOAD_VERSION);
```
bmhowe23, line 113: It would be best to add an underscore at the end so that when `REST_PAYLOAD_VERSION` = 1 that it does not match `cuda_quantum_11`.

```suggestion
    const std::string versionedCudaqNvcfFuncNamePrefix =
        baseCudaqNvcfFuncNamePrefix + "_" +
        std::to_string(cudaq::RestRequest::REST_PAYLOAD_VERSION) + "_";
```

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
```diff
@@ -76,29 +76,49 @@ class NvcfRuntimeClient : public RemoteRestRuntimeClient {
...
       }
-    }
-    return info;
+      return info;
+    };
+
+    const std::string baseCudaqNvcfFuncNamePrefix = "cuda_quantum";
+    const std::string versionedCudaqNvcfFuncNamePrefix =
+        baseCudaqNvcfFuncNamePrefix + "_" +
+        std::to_string(cudaq::RestRequest::REST_PAYLOAD_VERSION);
```
bmhowe23, line 110: Recommend same change as in python/runtime/utils/PyRestRemoteClient.cpp.

bmhowe23 (approved):

python/runtime/utils/PyRestRemoteClient.cpp
1tnguyen: Good catch. Thanks!

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteClient.cpp
1tnguyen: Fixed as recommended.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -120,6 +120,10 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
       : cudaq::RemoteRuntimeServer(),
         m_simHandle(DEFAULT_NVQIR_SIMULATION_BACKEND,
                     loadNvqirSimLib(DEFAULT_NVQIR_SIMULATION_BACKEND)) {}
+
+  virtual int version() const override {
+    return cudaq::RestRequest::REST_PAYLOAD_VERSION;
```
bmhowe23, line 125: (Hopefully) one last request - can we please allow an environment variable override for this? It might be useful for testing purposes to have the client target different `REST_PAYLOAD_VERSION` values other than what it was built with. I.e. In future integration and/or backwards-compatibility testing, we might have client version `3` run tests with server versions `2` and `4` for specific payloads. This wouldn't help if we actually change the API schema, but it might help for cases when the server is bumping the version to signify support for a new runtime function.

1tnguyen: Please take a look at https://github.com/NVIDIA/cuda-quantum/pull/1328/commits/3c8b12007b4eee99aa92940761ff7eb12bfa6516 and let me know if I missed anything :) 

Basically, both the client and server interfaces have the abstract `version` method. I only added environment variable override for the client side as suggested.

bmhowe23: Looks great...thanks!

PR #1330, opened 2024-03-03
[nvqc] Add watchdog timeout option for remote rest server

bmhowe23:
This PR allows us to set a `WATCHDOG_TIMEOUT_SEC` environment variable with NVQC deployment that allows us to set a maximum quantum execution time per request. If the time is exceeded, the `cudaq-qpud` process aborts and restarts. The client will see an error message like this:

```
  what():  Failed to launch kernel. Error: HTTP GET Error - status code 500: : {"type":"urn:nvcf-worker-service:problem-details:internal-server-error","title":"Internal Server Error","status":500,"detail":"Internal error while making inference request","instance":"/v2/nvcf/exec/status/4f6f4e3d-0a43-41fe-8967-2eb450a7f893","requestId":"4f6f4e3d-0a43-41fe-8967-2eb450a7f893"}
Aborted
```

I set the default timeout to be ridiculously large (i.e. 30 days), but presumably we need to set this to something lower in our actual deployments.


1tnguyen (approved):
LGTM 👍

PR #1331, opened 2024-03-03
Add `cudaq.__version__` attribute to Python

bmhowe23:
This change exposes the `cudaq` Python wheel version similar to how `numpy.__version__` reports a version string from within the Python wheel.

When run in a development environment, it produces something like this, which is the same as `nvq++ --version:
```python
>>> import cudaq
>>> print(cudaq.__version__)
CUDA Quantum Version slush-0.7.0 (https://github.com/NVIDIA/cuda-quantum b9c52cd3fe4425ad0b3780049a101a7284e5aa5f)
```

When run from a fully released version, it *should* look like this:
```python
>>> import cudaq
>>> print(cudaq.__version__)
CUDA Quantum Version 0.7.0 (https://github.com/NVIDIA/cuda-quantum <exact hash of release>)
```


khalatepradnya (approved):
Looks good!

PR #1332, opened 2024-03-04
Enable NVQC checks in `validate_container.sh`

1tnguyen:


schweitzpgi:
Clearing milestone as 0.7.0 has shipped.

PR #1333, opened 2024-03-04
Fix the example to use std::vector<bool>, which is a type that is

schweitzpgi:
named in the CUDA Quantum Specification.

### DIFF
--- a/docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
+++ b/docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
@@ -10,19 +10,19 @@
 // simulated using the multi-GPU backend. The amount of resources required for
 // the simulation doubles with with each additional qubit.
 
-#include <bitset>
 #include <cudaq.h>
 #include <iostream>
 #include <random>
+#include <vector>
 
 #ifndef SIZE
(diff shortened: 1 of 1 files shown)


docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
```diff
@@ -10,19 +10,19 @@
...
 #ifndef SIZE
 #define SIZE 5
 #endif
 
 template <int nrOfBits>
-std::bitset<nrOfBits> random_bits(int seed) {
+std::vector<bool> random_bits(int seed) {
 
-  std::bitset<nrOfBits> randomBits;
+  std::vector<bool> randomBitsi(nrOfBits);
```
bmhowe23, line 25:
```suggestion
  std::vector<bool> randomBits(nrOfBits);
```
Also - I don't think `std::vector<bool>` supports the `.set` method used in line 30?

amccaskey (approved):

khalatepradnya:
I see that this is merged... follow-up Q:
on line#75, `printf("Encoded bitstring:  %s\n", bitvector.to_string().c_str());` , is `to_string()` supported?

PR #1334, opened 2024-03-04
[refactoring] Remove code duplication for nvcf client classes

khalatepradnya:
* Define base class `BaseNvcfRuntimeClient` from which `NvcfRuntimeClient` (C++) and `PyNvcfRuntimeClient` (Python) inherit
so as to remove the duplicate code across files / classes.


schweitzpgi (approved):
Wow. Nice.

PR #1335, opened 2024-03-04
[refactoring] Remove code duplication for nvcf simulator classes

khalatepradnya:
* Define a base class `BaseNvcfSimulatorQPU` to encapsulate QPU subtype that submits to NVCF and serves as parent class for C++ and Python classes.


python/runtime/utils/PyRemoteSimulatorQPU.cpp
```diff
@@ -67,125 +65,58 @@ class PyRemoteSimulatorQPU : public cudaq::BaseRemoteSimulatorQPU {
...
 };
 
-/// NOTE: Following is a duplication of `NvcfSimulatorQPU` from
-/// `runtime/cudaq/platform/mqpu/remote/RemoteSimulatorQPU.cpp`
-/// TODO: Refactor
 /// Implementation of QPU subtype that submits simulation request to NVCF.
-class PyNvcfSimulatorQPU : public PyRemoteSimulatorQPU {
+/// NOTE: This class duplicates the `isEmulated` and `launchKernel` methods from
+/// `PyRemoteSimulatorQPU` class above; tried using multiple inheritance, but,
+/// got errors from the functionality to register type.
```
schweitzpgi, line 71: Not sure I follow what the problem is, but this may be a case to use CRTP instead of inheritance.

khalatepradnya: The error was -
`/opt/llvm/include/llvm/Support/Registry.h:118:71: error: could not convert ‘std::make_unique(_Args&& ...) [with _Tp = {anonymous}::class_name; _Args = {}; __detail::__unique_ptr_t<_Tp> = __detail::__unique_ptr_t<{anonymous}::class_name>]()’ from ‘unique_ptr<{anonymous}::class_name,default_delete<{anonymous}::class_name>>’ to ‘unique_ptr<cudaq::QPU,default_delete<cudaq::QPU>>’`
class_name = {PyNvcfSimulatorQPU, NvcfSimulatorQPU}

1tnguyen: A minor request: is it possible to move `struct ArgsWrapper` declarations here and the ones in `PyRemoteRESTQPU.cpp` and `py_alt_launch_kernel.cpp` into a shared header file (visible to all places)?

1tnguyen (approved):
LGTM. Thanks, Pradnya.

PR #1339, opened 2024-03-05
[clean-up] [docs] Remove dead code

khalatepradnya:
* Removing dead code. 
  * Moving the docstrings from deleted class to their new implementation.


amccaskey (approved):
thanks for the cleanup!

python/cudaq/kernels/hwe.py
```diff
@@ -11,10 +11,25 @@
 
 
 def num_hwe_parameters(numQubits, numLayers):
+    """
+    For the given number of qubits and layers, return the required number of `hwe` parameters.
+    """
```
khalatepradnya, line 16: When I generated docs locally, this did not show up - not sure where it should have got populated...

khalatepradnya: Searched this in the current docs at https://nvidia.github.io/cuda-quantum/latest/ and it didn't yield any results either...

PR #1342, opened 2024-03-05
Runtime error if kernel arguments missing / incorrect when invoking   'draw'

khalatepradnya:
Description:
> We should throw a better error in cudaq.draw() if someone forgets to pass along the kernel arguments. Right now, it just returns a segfault without explanation.


amccaskey (approved):
+1 That works!

PR #1343, opened 2024-03-06
[NVQC] Fix edge case when response data could be corrupted during shutdown

1tnguyen:
### Description

CrowCpp has a [threshold](https://github.com/CrowCpp/Crow/blob/c47540bb8c5adf96add972c240284f7d4e4e30b0/include/crow/http_connection.h#L426) to switch on HTTP streaming (i.e., perform async. write to the TCP buffer). 

This mode is susceptible to corruption when the server is shutting down. In particular, response size > 65k bytes is likely to be written in multiple passes, which might not be complete when the server is stopping. Hence, setting a small `stream_threshold` to use synchronous write.


runtime/cudaq/platform/default/rest_server/helpers/server_impl/RestServer.cpp
```diff
@@ -39,6 +39,10 @@ cudaq::RestServer::RestServer(int port, const std::string &name) {
   m_impl->app.port(port);
   m_impl->app.server_name(name);
   m_impl->app.loglevel(crow::LogLevel::Warning);
+  // Set a small stream threshold to enforce synchronous TCP buffer write.
+  // Asynchronous buffer write (`asio::async_write`) is susceptible to
+  // corruption if the app is shut down right after handling a request.
+  m_impl->app.stream_threshold(16 * 1024);
```
bmhowe23, line 45: Why not 0? If we set it to 16KB, wouldn't there still be an edge case if the `res.body.length()` is < 16KB?

bmhowe23 (approved):
Thanks, Thien!

PR #1344, opened 2024-03-06
[NVQC] Add logging info for timeout 

1tnguyen:
### Description

- Retrieve timeout info from NVQC deployment (container environment variables).

- Log timeout limit alongside the NVQC info line.

- Warn users when timeout is approaching (less than 5 min remained).

### DIFF
--- a/python/runtime/utils/PyRestRemoteClient.cpp
+++ b/python/runtime/utils/PyRestRemoteClient.cpp
@@ -38,8 +38,15 @@ private:
   std::string m_functionId;
   // NVCF version Id of that function to use
   std::string m_functionVersionId;
-  // Available functions: function Id to number of GPUs mapping
-  using DeploymentInfo = std::unordered_map<std::string, std::size_t>;
+  // Information about function deployment from environment variable info.
+  struct FunctionEnvironments {
+    // These configs should be positive numbers.
+    int version;
+    int numGpus;
+    int timeoutSecs;
+  };
+  // Available functions: function Id to info mapping
+  using DeploymentInfo = std::unordered_map<std::string, FunctionEnvironments>;
   DeploymentInfo m_availableFuncs;
   const std::string CUDAQ_NCA_ID = cudaq::getNvqcNcaId();
   // Base URL for NVCF APIs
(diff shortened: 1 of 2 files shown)


python/runtime/utils/PyRestRemoteClient.cpp
```diff
@@ -371,15 +383,29 @@ class PyNvcfRuntimeClient : public PyRestRemoteClient {
...
+            std::chrono::duration_cast<std::chrono::seconds>(
+                std::chrono::system_clock::now() - requestStartTime)
+                .count();
         if (m_logLevel > LogLevel::None)
           cudaq::log("Polling NVQC result data for Request Id {}", reqId);
+        // Warns if the remaining time is less than this threshold.
+        constexpr int TIMEOUT_WARNING_SECS = 5 * 60; // 5 minutes.
+        const int remainingSecs = std::max(
+            m_availableFuncs[m_functionId].timeoutSecs - elapsedTimeSecs, 0);
+        if (remainingSecs < TIMEOUT_WARNING_SECS)
```
1tnguyen, line 404: @bmhowe23  I wonder whether we should log this "Approaching the wall time limit ..." line only once or repeat (like this).
Please let me know if you have any recommendations.

bmhowe23: > @bmhowe23 I wonder whether we should log this "Approaching the wall time limit ..." line only once or repeat (like this). Please let me know if you have any recommendations.

I put some suggestions in 69008e3f3.

bmhowe23 (approved):

python/runtime/utils/PyRestRemoteClient.cpp
```diff
@@ -38,8 +38,15 @@ class PyNvcfRuntimeClient : public PyRestRemoteClient {
   std::string m_functionId;
   // NVCF version Id of that function to use
   std::string m_functionVersionId;
-  // Available functions: function Id to number of GPUs mapping
-  using DeploymentInfo = std::unordered_map<std::string, std::size_t>;
+  // Information about function deployment from environment variable info.
+  struct FunctionEnvironments {
+    // These configs should be positive numbers.
```
schweitzpgi, line 43: Positive or non-negative?

1tnguyen: They should be positive number.

PR #1347, opened 2024-03-06
Add support for vector<int64_t>, vector<int32_t>, and vector<float>

schweitzpgi:
to the synthesizer.
Add test for vector<int64_t>.
Fix some bugs.
Implement folding of size() calls on vector inputs in synthesis. Allow vectors that are longer than what is required.

**Release**
These changes enable code that was failing on nvqc.

### DIFF
--- a/lib/Optimizer/Transforms/QuakeSynthesizer.cpp
+++ b/lib/Optimizer/Transforms/QuakeSynthesizer.cpp
@@ -40,8 +40,8 @@ using namespace mlir;
 /// BlockArgument with it.
 template <typename ConcreteType>
 void synthesizeRuntimeArgument(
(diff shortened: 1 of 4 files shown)


bmhowe23:
Thanks, Eric. Does this support the unsigned types as well?

schweitzpgi:
> Thanks, Eric. Does this support the unsigned types as well?

It doesn't matter in this particular case all that much since integral values are defined as a series of bits of a specific length. The 8 bits `11111110` are both the value `-2` (signed interpretation) and `254` (unsigned interpretation). Here we just get the length and sequence of bits initialized and it is up to the operations to interpret those bits as signed and/or unsigned values.

1tnguyen (approved):
LGTM

PR #1349, opened 2024-03-07
Ability to decode deployment configs from function names

1tnguyen:
### Description

When container environments are not visible, function name is the only mechanism to encode deployment configurations.

Proposed naming convention: `cuda_quantum_v1_t3600_8x` (encode API version, timeout in secs, and the number of GPUs).

Implement a regex match procedure to parse these configs from the name. It will take precedence over container environment variables.

### DIFF
--- a/python/runtime/utils/PyRestRemoteClient.cpp
+++ b/python/runtime/utils/PyRestRemoteClient.cpp
@@ -8,6 +8,8 @@
 
 #include "common/BaseRestRemoteClient.h"
 #include "common/NvqcConfig.h"
+#include <regex>
+
 using namespace mlir;
 
 namespace {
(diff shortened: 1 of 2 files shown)


khalatepradnya (approved):
👍🏽

bmhowe23 (approved):
+1

bettinaheim:
@1tnguyen Could you clarify for me when/in which cases container environments are not visible and when they are?

1tnguyen:
> @1tnguyen Could you clarify for me when/in which cases container environments are not visible and when they are?

When we share a function to another NGC organization, the container environment variables are not visible for users from that organization. Users from the host organization can see the environment variables. This is currently a known limitation of NVCF (by design). There will be function metadata in the near future to store these extra pieces of information.

PR #1356, opened 2024-03-07
[Core] Tweak the synthesizer to apply constant vector values more

schweitzpgi:
generally. These changes allow the inlining of the constants to happen more lazily and before fully flattening the kernel.

(#1347 should be merged first.)


1tnguyen (approved):
LGTM.

Issue #1233, opened 2024-02-15
[experimental/python] Test failure when running with whole suite of Python tests


khalatepradnya:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

* 'remote' tests pass when run individually, but, one test fails if run with whole suite of tests
* Renaming the kernel in the failing test makes it pass
* Potential kernel name clashing / caching issue

### Steps to reproduce the bug

Apply this patch to 'experimental/python' branch
```diff --git a/python/tests/remote/test_remote_platform.py b/python/tests/remote/test_remote_platform.py
index f439b2635..e62dfd482 100644
--- a/python/tests/remote/test_remote_platform.py
+++ b/python/tests/remote/test_remote_platform.py
@@ -169,13 +169,13 @@ def test_multi_qpus():
 def test_multi_qpus_kernel():
 
     @cudaq.kernel
-    def parameterized_ansatz(theta: float):
+    def ansatz(theta: float):
         qreg = cudaq.qvector(2)
         x(qreg[0])
         ry(theta, qreg[1])
         x.ctrl(qreg[1], qreg[0])
 
-    check_multi_qpus(parameterized_ansatz)
+    check_multi_qpus(ansatz)
 
 
 # leave for gdb debugging
```

and run the Python tests:

`$ python3 -m pytest -v python/tests/remote/`
-- Success
`$ python3 -m pytest python/tests/ --ignore python/tests/backends/`
-- Failure
For quick repro, run only `kernel` and `remote` tests
`$ python3 -m pytest -v python/tests/kernel/ python/tests/remote/`
-- Failure

### Expected behavior

* All tests pass when run together

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: experimental/python
- **Python version**: 3.10
- **C++ compiler**: N/A
- **Operating system**: Ubuntu 22.04

khalatepradnya:
Updated patch for `main`

```
diff --git a/python/tests/remote/test_remote_platform.py b/python/tests/remote/test_remote_platform.py
[...]

PR #1357, opened 2024-03-08
[Bug fix] Replace entries in the quake registry

khalatepradnya:
* Allow redefinitions of kernel and replace them in the `quakeRegistry`.

### Description
Addresses issue https://github.com/NVIDIA/cuda-quantum/issues/1233


schweitzpgi (commented):
Seems like a good approach. :)

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -44,10 +44,11 @@ class GenerateDeviceCodeLoader
...
           loc, "deviceCodeHolderAdd",
-          LLVM::LLVMFunctionType::get(
-              cudaq::opt::factory::getVoidType(ctx),
-              {cudaq::opt::factory::getPointerType(ctx),
-               cudaq::opt::factory::getPointerType(ctx)}));
+          LLVM::LLVMFunctionType::get(cudaq::opt::factory::getVoidType(ctx),
+                                      {cudaq::opt::factory::getPointerType(ctx),
+                                       cudaq::opt::factory::getPointerType(ctx),
+                                       // bool --> char type (int 8)?
+                                       cudaq::opt::factory::getCharType(ctx)}));
```
schweitzpgi: Can just use `builder.getI1Type()` here.

schweitzpgi: You have to change the code where the call is made as well, and add a `ConstantIntOp` of `i1 0` as the final argument. Otherwise, the code will just pass garbage from the stack at runtime.

bmhowe23 (approved):
👍

PR #1358, opened 2024-03-08
[GenDeviceCodeLoader] Recursively adding all dependent kernels' Quake code

1tnguyen:
### Description

We have a logic to add all dependent kernel quake code but it only works for first-level calls.

This adds a recursive step to chase the call chain if needed.

A sample use case is Python MLIR [uccsd](https://github.com/NVIDIA/cuda-quantum/blob/main/python/cudaq/kernels/uccsd.py ) where there are multi-level call chains, e.g., `uccsd` -> `uccsd_odd_electrons` -> `single_excitation`.


test/NVQPP/gen_device_code.cpp
bmhowe23, line 1: I think that per the `tests/README.txt` file, this should probably go in the `targettests/execution` directory.

bmhowe23 (approved):
Nice find!

test/NVQPP/gen_device_code.cpp
1tnguyen: I've moved the file to the `targettests/execution` directory

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -91,22 +91,27 @@ class GenerateDeviceCodeLoader
...
-            if (includedNeededFuncs.contains(neededFunc.getName()))
-              return WalkResult::skip();
-            neededFunc.print(strOut, opf);
-            includedNeededFuncs.insert(neededFunc.getName());
-            strOut << '\n';
-            return WalkResult::advance();
-          }
-          return WalkResult::advance();
-        });
+        const std::function<WalkResult(func::CallOp)> emitDependentFunc =
```
schweitzpgi, line 94: This variable isn't really needed.

bmhowe23: I'm curious ... if this variable isn't needed, how would you perform the recursive call in line 108?

schweitzpgi: We're already walking every function in the module. This just turns a linear scan into a recursive dive back through the same functions. Why do that?

Create a map for each function in turn. Then scan that map and create the transitive closure.

1tnguyen: Thanks @schweitzpgi. Now I see your point. I added #1366 to get it fixed.

PR #1363, opened 2024-03-08
Fix spin-op deserialization bug when there is exactly 1 term

bmhowe23:
This fixes a bug in formulas used when deserializing a spin op from a vector. The new test in the PR fails without the code change and passes with the accompanying code change.

I confirmed that this error was only happening when there was exactly 1 term in the spin op.


1tnguyen (approved):
LGTM 👍

PR #1364, opened 2024-03-09
[NFC] Update the copyrights in file banners

schweitzpgi:


khalatepradnya (approved):
👍🏽

khalatepradnya:
I guess this was done with a script, not sure if we want to add this script to source code and just run it next year...

PR #1368, opened 2024-03-11
Use actual json string size to compare against NVCF message size limit

1tnguyen:
### Description

This PR fixes an oversight in checking whether we need to upload the request as an asset: need to use the full JSON string size rather than only the main IR payload.

### DIFF
--- a/python/runtime/utils/PyRestRemoteClient.cpp
+++ b/python/runtime/utils/PyRestRemoteClient.cpp
@@ -405,7 +405,7 @@ public:
     });
 
     // Upload this request as an NVCF asset if needed.
-    if (request.code.size() > MAX_SIZE_BYTES) {
+    if (json(request).dump().size() > MAX_SIZE_BYTES) {
       assetId = uploadRequest(request);
       if (!assetId.has_value()) {
         if (optionalErrorMsg)
--- a/python/runtime/utils/PyRestRemoteClient.cpp
+++ b/python/runtime/utils/PyRestRemoteClient.cpp
@@ -405,7 +405,12 @@ public:
     });
 
     // Upload this request as an NVCF asset if needed.
-    if (json(request).dump().size() > MAX_SIZE_BYTES) {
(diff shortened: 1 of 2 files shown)


khalatepradnya (approved):
Good catch!

python/runtime/utils/PyRestRemoteClient.cpp
```diff
@@ -405,7 +405,7 @@ class PyNvcfRuntimeClient : public PyRestRemoteClient {
     });
 
     // Upload this request as an NVCF asset if needed.
-    if (request.code.size() > MAX_SIZE_BYTES) {
+    if (json(request).dump().size() > MAX_SIZE_BYTES) {
```
bmhowe23, line 408: Is there any way to just allocate a size padding to add to `request.code.size()` rather than construct a new, temporary, full JSON object copy just capture the resulting size?

1tnguyen: We could, but I think it might be a bit fragile. Do you have a recommendation for a reasonable size padding?
There was a discussion about possibly sending on the arguments to the server. In that case, we might need to revise the padding size or this check condition as a whole.

Another potential solution is to make it a 2-layer condition:  ` if (request.code.size() > MAX_SIZE_BYTES || json(request).dump().size() > MAX_SIZE_BYTES)`:  i.e., if the IR payload (the major one) is already large, then no need to check any further, hence saving the temp json allocation.

bmhowe23: I like the idea of the 2-layer condition. It prevents the worst case scenario (constructing large objects only to throw them away later), and I suppose the less-than-worst-case scenario probably isn't that big of a deal.

1tnguyen: I've revised this PR to implement the above strategy.

bmhowe23 (approved):

PR #1369, opened 2024-03-11
Fix issue #1367

### LINKED ISSUE #1367: Synthesizer produces error when mixing certain argument types
### Required prerequisites

### Describe the bug

When running a test program, the synthesizer produces the following error:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted
```

Compile and run the following code using `nvq++ --target nvqc path/to/file.cpp`:
```cpp
#include <ctime>
#include <cudaq.h>
#include <cudaq/algorithm.h>
#include <cudaq/optimizers.h>
#include <random>

struct kernel {
  auto operator()(const int n_qubits, const std::vector<double> &parameters) __qpu__ {
    cudaq::qvector qubits(n_qubits);
    for (size_t i = 0; i < n_qubits; i++)
      rx(parameters[i], qubits[i]);
  }
};

std::vector<double> initial_parameters(int n_parameters, int seed) {
  std::default_random_engine generator(seed);
  std::uniform_real_distribution<float> distribution(0.0, 1.0);
  std::vector<double> parameters(n_parameters);
  for (size_t i = 0; i < n_parameters; i++)
    parameters[i] = distribution(generator);
  return parameters;
}

int main() {

  const int n_qubits = 26;
  const int n_parameters =  n_qubits;
  std::vector<double> parameters = initial_parameters(n_parameters, 13);
  auto h = cudaq::spin::z(0);
  auto exp_val = cudaq::observe(kernel{}, h, n_qubits, parameters);
  printf("Expectation Value: %f \n", exp_val.expectation());

  return 0;
}
```

### Expected behavior

The program should run without errors.

Not a regression

### Environment

- **CUDA Quantum version**: 8dd93ad46
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

I believe the problem has something to do with `offset` not being correct in this line of code: https://github.com/NVIDIA/cuda-quantum/blob/8dd93ad466cd3058350d5b2a0c5518cd86aa1f17/lib/Optimizer/Transforms/QuakeSynthesizer.cpp#L387.

I think that `offset` was incremented by 4 to skip past the 32-bit value, but the `vector` is actually beginning at `offset=8` instead of `offset=4`.

schweitzpgi:
Need to use LLVM's understanding of how the struct of arguments is laid out in memory in order to recover the values that the synthesizer will use to instantiate a circuit from a kernel.

Fix #1367

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -378,6 +378,7 @@ public:
       func->setAttr(cudaq::entryPointAttrName, builder.getUnitAttr());
     auto moduleOp = builder.create<ModuleOp>();
     moduleOp.push_back(func.clone());
+    moduleOp->setAttrs(m_module->getAttrDictionary());
 
     // Lambda to apply a specific pipeline to the given ModuleOp
     auto runPassPipeline = [&](const std::string &pipeline,
--- a/include/cudaq/Optimizer/Builder/Factory.h
+++ b/include/cudaq/Optimizer/Builder/Factory.h
@@ -75,7 +75,7 @@ mlir::Type genArgumentBufferType(mlir::Type ty);
 /// ```c++
 /// i32_t operator() (i16_t, std::vector<double>, double);
 /// ```
-/// will generate the llvm struct
+/// will generate the LLVM struct
 /// ```llvm
 /// { i16, i64, double, i32 }
 /// ```
--- a/include/cudaq/Optimizer/Builder/Factory.h
+++ b/include/cudaq/Optimizer/Builder/Factory.h
@@ -64,6 +64,25 @@ inline mlir::Type getPointerType(mlir::Type ty) {
 
 cudaq::cc::PointerType getIndexedObjectType(mlir::Type eleTy);
 
+mlir::Type genArgumentBufferType(mlir::Type ty);
+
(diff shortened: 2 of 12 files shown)


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -378,6 +378,7 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
       func->setAttr(cudaq::entryPointAttrName, builder.getUnitAttr());
     auto moduleOp = builder.create<ModuleOp>();
     moduleOp.push_back(func.clone());
+    moduleOp->setAttrs(m_module->getAttrDictionary());
```
1tnguyen, line 381: Do we need to add a similar call to `runtime/common/BaseRestRemoteClient.h`: 164? e.g.,

```
moduleOp->setAttrs((*module)->getAttrDictionary());
```

bmhowe23: I suspect so. Without that additional change, I was still getting the same failure described in the issue. With that change, it appears to work for me.

schweitzpgi: Very likely.

bmhowe23 (approved):

PR #1370, opened 2024-03-12
Fix bug - enable list[T] argument specification with kernel builder

amccaskey:
We now require argument annotations with the Python language kernels, and lists must have element type specified (e.g. `list[float]` instead of just `list`). 

Users likely will try to specify list element types with the kernel builder as well, but this currently emits an error 
```python
import cudaq 
kernel, thetas = cudaq.make_kernel(list[float])
print(kernel)
```
```bash
RuntimeError: error: Can not handle conversion of python type list[float] to MLIR type.

Offending code:
  File "/workspaces/cuda-quantum/builds/gcc-12-debug/test.py", line 3, in <module>
    kernel, thetas = cudaq.make_kernel(list[float])
```

This PR fixes this by enabling element type inference for kernel builder list arguments.

### DIFF
--- a/python/cudaq/kernel/kernel_builder.py
+++ b/python/cudaq/kernel/kernel_builder.py
@@ -7,8 +7,11 @@
 # ============================================================================ #
 
 from functools import partialmethod
+from pydoc import locate
 import random
+import re
 import string
+import typing
 from .quake_value import QuakeValue
 from .kernel_decorator import PyKernelDecorator
 from .utils import mlirTypeFromPyType, nvqppPrefix, emitFatalError, mlirTypeToPyType
@@ -215,7 +218,9 @@ class PyKernel(object):
 
         with self.ctx, InsertionPoint(self.module.body), self.loc:
             self.mlirArgTypes = [
-                mlirTypeFromPyType(argType, self.ctx) for argType in argTypeList
+                mlirTypeFromPyType(argType[0], self.ctx, argInstance=argType[1])
+                for argType in
+                [self.__processArgType(ty) for ty in argTypeList]
             ]
 
             self.funcOp = func.FuncOp(self.funcName, (self.mlirArgTypes, []),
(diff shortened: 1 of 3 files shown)


python/cudaq/kernel/kernel_builder.py
```diff
@@ -231,6 +236,23 @@ def __init__(self, argTypeList):
...
+        """
+        Process input argument type. Specifically, try to infer the 
+        element type for a list, e.g. list[float]. 
+        """
+        if ty in [cudaq_runtime.qvector, cudaq_runtime.qubit]:
+            return ty, None
+        if typing.get_origin(ty) == list or isinstance(ty(), list):
+            if '[' in str(ty) and ']' in str(ty):
+                # Infer the slice type
+                result = re.search(r'ist\[(.*)\]', str(ty))
```
bmhowe23, line 249: I was initially concerned that this might not scale to a 2D list, but I think that wouldn't work anyway (for other reasons). From a language specification perspective, do we intend to allow it?

bmhowe23 (commented):
Thanks, Alex. LGTM, but could we add some tests to make sure we don't break it in the future? (I can create some and push to the PR if you prefer ... just let me know.)

amccaskey:
> Thanks, Alex. LGTM, but could we add some tests to make sure we don't break it in the future? (I can create some and push to the PR if you prefer ... just let me know.)

Thanks Ben. I appreciate your offer to help - it would be very helpful if you were able to do so.

python/cudaq/kernel/kernel_builder.py
amccaskey: Yea I played around with that a bit. I fell back to retaining the support we already had for the Python builder - we didn't allow `list[list[...]]` before so I think its fine to throw an error here for that. I think we can vet that decision in the language meetings and then adjust this in the future if necessary.

amccaskey:
> > Thanks, Alex. LGTM, but could we add some tests to make sure we don't break it in the future? (I can create some and push to the PR if you prefer ... just let me know.)
> 
> Thanks Ben. I appreciate your offer to help - it would be very helpful if you were able to do so.

Test added.

bmhowe23 (approved):
👍

PR #1372, opened 2024-03-12
Use `mlir::CallGraph` to manage kernel dependency in `GenerateDeviceCodeLoader`

### LINKED ISSUE #1366: Improve Quake code generation for kernels with nested call chains
The recursive implementation in PR#1358 is sub-optimal since:

We're already walking every function in the module. This just turns a linear scan into a recursive dive back through the same functions. Why do that? [Assume you remove the continue condition.]

Recommeded fix:

Create a map for each function in turn. Then scan that map and create the transitive closure.

_Originally posted by @schweitzpgi in https://github.com/NVIDIA/cuda-quantum/pull/1358#discussion_r1518425832_

1tnguyen:
### Description

- Simplify `GenerateDeviceCodeLoader` using `CallGraph` and `df_interator` utils. This helps eliminate the need to track repeated kernel calls or to make recursive calls. 

- Change the `targettests/execution/gen_device_code.cpp` test to be mote comprehensive.

Fixes #1366


lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -10,16 +10,43 @@
...
+#include "mlir/Analysis/CallGraph.h"
 #include "mlir/IR/Diagnostics.h"
 #include "mlir/Transforms/Passes.h"
 
 #define DEBUG_TYPE "device-code-loader"
 
 using namespace mlir;
 
+namespace llvm {
+// FIXME: `GraphTraits<const mlir::CallGraphNode *>` as defined in
```
schweitzpgi, line 26: Not sure I understand this. Did you look at `Optimizer/CodeGen/TranslateToOpenQASM.cpp`? It is using `mlir::CallGraph`.

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -84,34 +113,27 @@ class GenerateDeviceCodeLoader
...
-              }
-              return WalkResult::advance();
-            };
-
-        funcOp.walk(emitDependentFunc);
+        // Note: the `CallGraphNode` does include 'this' function.
+        mlir::CallGraphNode *node =
+            callGraph.lookupNode(funcOp.getCallableRegion());
+        // Iterate over all dependent kernels starting at this node.
+        for (auto Itr = llvm::df_begin(node), ItrE = llvm::df_end(node);
```
schweitzpgi, line 124: nit: don't use CamelCase, use camelCase for variable names.

targettests/execution/gen_device_code.cpp
```diff
@@ -20,27 +20,41 @@ struct baz {
...
-  cudaq::sample(foo{});
   std::cout << cudaq::get_quake(foo{}) << "\n";
   return 0;
 }
 // CHECK-LABEL:   func.func @__nvqpp__mlirgen__foo
-// CHECK-LABEL:   func.func @__nvqpp__mlirgen__bar
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__bar2
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__bah
 // CHECK-LABEL:   func.func @__nvqpp__mlirgen__bax
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__bar1
```
schweitzpgi, line 59: Are these guaranteed to be ordered in any way by the call graph iterator? Might want to use `CHECK-DAG` to make the test imperious to random iterator traversal changes.

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
1tnguyen: Fixed.

targettests/execution/gen_device_code.cpp
1tnguyen: This should be guaranteed by the 'depth first' traversal iterator that we chose. I think it might be worth checking the deterministic behavior that we expected in this case.

schweitzpgi (approved):

schweitzpgi:
We're going to have another issue, that the call graph won't solve. Specifically, a kernel may reference constant data, which also will need to appear in the excised module.

1tnguyen:
> We're going to have another issue, that the call graph won't solve. Specifically, a kernel may reference constant data, which also will need to appear in the excised module.

Good point. I've added #1432 to make sure it doesn't fall through the cracks.

PR #1376, opened 2024-03-12
[submodule] Add CPR as a submodule

schweitzpgi:
The CPR repository is downloaded and built on-the-fly in the cmake configuration at the moment. ~The next step will be to modify the build to correctly build this package as a prereq.~


.gitmodules
```diff
@@ -35,3 +35,6 @@
 	path = tpls/qpp
 	url = https://github.com/softwareQinc/qpp.git
 	shallow = true
+[submodule "tpls/cpr"]
```
bettinaheim, line 38: It's odd to add this as a submodule if we are not actually using the submodule to build the code - I think this is just confusing, because changing the submodule commit will not change the code.

schweitzpgi: I can add the changes to cmake to this PR. I had thought it made more sense to make it 2 steps.

schweitzpgi:
This restructuring gets past the ci.

schweitzpgi:
Circle the airport until release 0.7 gets finished.

.gitmodules
```diff
@@ -35,3 +35,6 @@
...
+	path = tpls/cpr
+	url = https://github.com/libcpr/cpr.git
```
khalatepradnya, line 40: [Adding comment here about the tag]
Q: Why are we using `v1.9.0` which is ~2 years old? Why not the latest stable version?

runtime/common/CMakeLists.txt
```diff
@@ -48,37 +48,6 @@ endif()
 if(OPENSSL_FOUND)
   message(STATUS "OpenSSL Found, building REST Client.")
 
-  include(FetchContent)
-  FetchContent_Declare(cpr 
-                       GIT_REPOSITORY https://github.com/libcpr/cpr.git
-                       GIT_TAG 871ed52d350214a034f6ef8a3b8f51c5ce1bd400)
```
khalatepradnya, line 54: I see it was the same tag previously...

khalatepradnya (approved):
Looks good to me.

.gitmodules
schweitzpgi: I don't know if that decision has some background to it. @1tnguyen might know?

I was conservative here and preserved the same SHA number.

PR #1377, opened 2024-03-12
[docs] Document 'draw' functionality

khalatepradnya:


python/runtime/cudaq/algorithms/py_draw.cpp
```diff
@@ -39,6 +39,37 @@ std::string pyDraw(py::object &kernel, py::args args) {
...
+
+  # Example
+  import cudaq
+  @cudaq.kernel
+  def bell_pair():
+      q = cudaq.qvector(2)
+      h(q[0])
+      cx(q[0], q[1])
+      mz(q)
+  print(cudaq.draw(bell_pair))
```
amccaskey, line 66: It would be good to have an example here in the docs that requires runtime arguments, so that users know draw requires them.

khalatepradnya: Addressed in 16049a5

amccaskey (approved):
+1

PR #1380, opened 2024-03-12
[submodule] Add Crow and asio as submodules.

schweitzpgi:
Change the cmake files to use the submodules rather than downloading on-the-fly.


.gitmodules
```diff
@@ -35,3 +35,9 @@
 	path = tpls/qpp
 	url = https://github.com/softwareQinc/qpp.git
 	shallow = true
+[submodule "tpls/asio"]
```
bmhowe23, line 38: When I brought this up at a [prior code review](https://github.com/NVIDIA/cuda-quantum/pull/1012#discussion_r1449173665), I think Thien preferred to keep it where it was. Any objections now @1tnguyen ?

schweitzpgi: We should use submodules for these things.

schweitzpgi:
Circle the airport until release-0.7 is finalized.

runtime/cudaq/platform/default/rest_server/helpers/server_impl/CMakeLists.txt
```diff
@@ -6,26 +6,10 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-include(FetchContent)
-FetchContent_Declare(
-    asio
-    GIT_REPOSITORY    https://github.com/chriskohlhoff/asio.git
-    GIT_TAG           asio-1-28-2
```
khalatepradnya, line 13: We may want to track if / when to update the submodules regularly... (outside the scope of this PR)

khalatepradnya (approved):
Looks good to me.

1tnguyen (approved):

PR #1381, opened 2024-03-12
[submodule] Update pybind11

### LINKED ISSUE #1046: Upgrade pybind11 to the latest version

schweitzpgi:
~The local custom mods are no longer required.~ Bump pybind11 up to the latest version on the master branch. It includes some fixes to its cmake files which avoid potential cmake conflicts.

In addition, simplify the use of Python in cmake. Instead of shotgunning find_package(Python) simply add it to the top-level and let that find propagate everywhere. Avoids errors when doing incremental rebuilds.

Fix #1046


khalatepradnya:
Curious if this issue https://github.com/NVIDIA/cuda-quantum/issues/1046 is still relevant...

schweitzpgi:
> Curious if this issue #1046 is still relevant... [Edit]: May be not since the issue was reported for 'eager' mode which has been removed.

Correct. Since we're not going to support library mode, that disappears. Yay!

khalatepradnya:
> The local custom mods are no longer required.

So, the https://github.com/pybind/pybind11/issues/4537 is fixed or no longer relevant?

schweitzpgi:
> > The local custom mods are no longer required.
> 
> So, the [pybind/pybind11#4537](https://github.com/pybind/pybind11/issues/4537) is fixed or no longer relevant?

That does not appear to be fixed in pybind11. Is there some reason we didn't submit a PR to fix it?

khalatepradnya:
> > > The local custom mods are no longer required.
> > 
> > 
> > So, the [pybind/pybind11#4537](https://github.com/pybind/pybind11/issues/4537) is fixed or no longer relevant?
> 
> That does not appear to be fixed in pybind11. Is there some reason we didn't submit a PR to fix it?
> 
> Added it back for this PR.

Q: Does the patch need to be updated? 
(If apply patch succeeds, then may not be required)

khalatepradnya (approved):

schweitzpgi:
Circle the airport until release 0.7 is finalized

PR #1382, opened 2024-03-12
Introduce pauli_word and expose to Python

amccaskey:
Enable expressions like this 
```python
 @cudaq.kernel
 def test(theta : float, paulis: list[cudaq.pauli_word]):
      q = cudaq.qvector(4)
      x(q[0])
      x(q[1])
      for p in paulis:
          exp_pauli(theta, q, p)

expval = cudaq.observe(test, h, .11, ['XXXY']).expectation()
```
or 
```python 
kernel, theta, paulis = cudaq.make_kernel(float, list[cudaq.pauli_word])
q = kernel.qalloc(4)
kernel.x(q[0])
kernel.x(q[1])
kernel.for_loop(0, paulis.size(), lambda idx : kernel.exp_pauli(theta, q, paulis[idx]))

expval = cudaq.observe(kernel, h, .11, ['XXXY']).expectation()
```

Future PR should expose this to C++


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -538,8 +538,11 @@ class GenerateKernelExecution
     auto ptrI8Ty = cudaq::cc::PointerType::get(builder.getI8Type());
     Type eleTy = stdvecTy.getElementType();
     auto innerStdvecTy = dyn_cast<cudaq::cc::StdvecType>(eleTy);
-    std::size_t eleSize =
-        innerStdvecTy ? /*(i64Type/8)*/ 8 : dataLayout->getTypeSize(eleTy);
+    std::size_t eleSize = [&]() -> unsigned int {
+      if (isa<quake::PauliWordType>(eleTy))
+        return 32;
```
schweitzpgi, line 543: Why 32?

schweitzpgi:
This looks a lot simpler now. Nice.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -833,7 +833,8 @@ def ExpPauliOp : QuakeOp<"exp_pauli"> {
   let arguments = (ins
     AnyFloat:$parameter,
     VeqType:$qubits,
-    cc_PointerType:$pauli
+    Optional<AnyTypeOf<[cc_PointerType, cc_StdVectorType]>>:$pauli,
```
schweitzpgi, line 836: Could just use SpanningType after my PR.

lib/Optimizer/CodeGen/LowerToQIR.cpp
```diff
@@ -367,11 +367,78 @@ class ExpPauliRewrite : public ConvertOpToLLVMPattern<quake::ExpPauliOp> {
...
-    // Make sure to drop any length information from the type of the Pauli word.
+    // First need to check the type of the Pauli word. We expect
+    // a pauli_word directly `{i8*,i64}` or a string literal
+    // `ptr<i8>`. If it is a string literal, we need to map it to
+    // a pauli word.
     auto pauliWord = operands.back();
-    operands.pop_back();
+    if (auto ptrTy = dyn_cast<LLVM::LLVMPointerType>(pauliWord.getType())) {
+      // Make sure we have the right types to extract the
+      // length of the string literal
```
schweitzpgi, line 377: I wasn't sure how this was to work either. The char span has the length, so we know what it is. But the QIS function that we're calling doesn't have a length argument. Maybe we ought to add one?

schweitzpgi (approved):
Seems reasonable to me. This one and my std::string one should be independent at the moment. Though I think they ought to be dovetailed together, which shouldn't be too hard.

My question with both of these is though is the Pauli word must end with a `'\0'` or whether we should pass the length as another argument to the qis function. Since the char span will have the size in hand, I'd prefer the extra argument.

Issue #1384, opened 2024-03-13
`QuakeSynthesizer` cannot handle `std::vector<char>` argument type


1tnguyen:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When there is a kernel argument of type `std::vector<char>`, `QuakeSynthesizer` couldn't perform the synthesis. This affects the hardware backends and remote backends (`nvqc`).

### Steps to reproduce the bug

```
#include <cudaq.h>
void f(const std::vector<char>& k) __qpu__ {
  cudaq::qubit q;
  for (int i = 0; i < k.size(); ++i)
    rx(1.0*k[i], q);
}

int main() {
  std::vector<char> x { 1, 2};  
  auto counts = cudaq::sample(f, x);
  counts.dump();
  return 0;
}
```

Run with:
```
nvq++ --target quantinuum --emulate test.cpp
./a.out
```

Error:

```
loc("test.cpp":2:1): error: 'func.func' op synthesis failed for vector<integral-type>.
loc("test.cpp":2:1): error: argument(s) still in use after synthesis.
terminate called after throwing an instance of 'std::runtime_error'
  what():  Could not successfully apply quake-synth.
Aborted (core dumped)
```

### Expected behavior

Successful synthesis.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: main
- **Python version**: 3.10
- **C++ compiler**: gcc-12
- **Operating system**: Ubuntu22.04

schweitzpgi:
FWIW, this isn't really a valid kernel signature. The vector must be pass by value according to the spec.
```c++
void f(const std::vector<char>& k) __qpu__ { // error
```

Issue #1383, opened 2024-03-13
nvq++ bridge compilation issue with `std::vector<bool>` argument


1tnguyen:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Compilation error: `'cc.if' op operand #0 must be 1-bit signless integer, but got '!cc.ptr<i1>'` when dereferencing a `std::vector<bool>` argument in MLIR mode. 

### Steps to reproduce the bug

```
#include <cudaq.h>

void f(const std::vector<bool>& k) __qpu__ {
  cudaq::qubit q;
  for (int i = 0; i < k.size(); ++i)
    if (k[i])
        x(q);
}

int main() {
  std::vector<bool> x { true, false};  
  auto counts = cudaq::sample(f, x);
  counts.dump();
  return 0;
}
```

The above example works in library mode but fails in MLIR mode:

```
error: 'cc.if' op operand #0 must be 1-bit signless integer, but got '!cc.ptr<i1>'
"builtin.module"() ({
  "func.func"() ({
  ^bb0(%arg0: !cc.stdvec<i1>):
    %0 = "quake.alloca"() : () -> !quake.ref
    "cc.scope"() ({
      %1 = "arith.constant"() {value = 0 : i32} : () -> i32
      %2 = "cc.alloca"() {elementType = i32} : () -> !cc.ptr<i32>
      "cc.store"(%1, %2) : (i32, !cc.ptr<i32>) -> ()
      "cc.loop"() ({
        %3 = "cc.load"(%2) : (!cc.ptr<i32>) -> i32
        %4 = "arith.extsi"(%3) : (i32) -> i64
        %5 = "cc.stdvec_size"(%arg0) : (!cc.stdvec<i1>) -> i64
        %6 = "arith.cmpi"(%4, %5) {predicate = 6 : i64} : (i64, i64) -> i1
        "cc.condition"(%6) : (i1) -> ()
      }, {
[...]

PR #1385, opened 2024-03-13
Adding more argument synthesis tests for NVQC

1tnguyen:
### Description

There is a need for a comprehensive set of tests to cover subtle issues, like the one in #1367 

- This PR adds a test suite for different argument types (and combinations of them in different orders). Using a bit of macro programming for test case generation. Note: some tests are currently disabled due to #1383 and #1384. Use `remote-mqpu` in MLIR mode as a mock environment for `nvqc`.

- Also, adding the one in #1367 as a test case, to be run in our nightly-integration test.


docs/sphinx/examples/cpp/providers/nvqc_qml.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <ctime>
+#include <cudaq.h>
+#include <cudaq/algorithm.h>
+#include <cudaq/optimizers.h>
+#include <random>
+
+struct kernel {
+
+  auto operator()(const int n_qubits,
+                  const std::vector<double> &parameters) __qpu__ {
```
bmhowe23, line 18: Passing by reference isn't really allowed
```suggestion
                  const std::vector<double> parameters) __qpu__ {
```

targettests/Remote-Sim/args_synthesis.cpp
```diff
@@ -0,0 +1,559 @@
...
+    mz(q);                                                                     \
+  }                                                                            \
+  const bool added##func_ = []() {                                             \
+    ALL_TEST_FUNCTORS.emplace_back([]() {                                      \
+      std::cout << "Testing: " << #func_ << "\n";                              \
+      auto counts = cudaq::sample(func_ END(INVOKE_LOOP_0 __VA_ARGS__));       \
+      counts.dump();                                                           \
+    });                                                                        \
+    return true;                                                               \
+  }();
```
bmhowe23, line 145: Nit: semicolon here and semicolon during invocations are produce double-semicolons in the resulting code. There's not really any harm in that, but it could be avoided.
```suggestion
  }()
```

targettests/Remote-Sim/args_synthesis.cpp
```diff
@@ -0,0 +1,559 @@
...
+static const std::vector<bool> VEC_BOOL_VAL = {true, false};
+static const std::vector<char> VEC_CHAR_VAL = {'a', 'b'};
+static const std::vector<int> VEC_INT_VAL = {1, 2};
+static const std::vector<std::size_t> VEC_SIZE_T_VAL = {1, 2};
+static const std::vector<double> VEC_DOUBLE_VAL = {1.0, 2.0};
+static const std::vector<float> VEC_FLOAT_VAL = {1.0, 2.0};
+
+// Define tests
+// Note: these are single-qubit tests, so will run very fast.
+DEFINE_TEST_KERNEL(test1, (bool, a, BOOL_VAL));
```
bmhowe23, line 164: When one of these tests fails, will we be able to easily tell which one is failing? My quick test of uncommenting out test that we expect to fail (e.g. `bool`) left me with a stack trace that wasn't immediately obvious to me which was failing.

1tnguyen: For those runtime failures (`vector<char>` in the `QuakeSynthesizer`), we do see a log about the name of the test. 

In the case of compiler errors (`vector<bool>`), unfortunately, I don't have a good solution for the logging. 
BTW, to see the actual pass error of `vector<bool>`, we also need to uncomment the overload for `vector<bool>` (line 87-91)

schweitzpgi: Can we use templates here instead of the macros? (The macros are impressive though.)

1tnguyen: I guess we can. In fact, I've tried to used templates but gave up after an hour or so :) 

The primary reason of using macros is that I can debug the generated code more quickly. Plus, I think the generated code would also mimic the end-users' usage that we want to test (writing kernels with concrete types, not having an extra template instantiation step to be handled by the bridge, etc.)

schweitzpgi: This sounds interesting. We handle templates in other places, so I'm a bit curious about the problems you saw.

bmhowe23: > The primary reason of using macros is that I can debug the generated code more quickly. Plus, I think the generated code would also mimic the end-users' usage that we want to test (writing kernels with concrete types, not having an extra template instantiation step to be handled by the bridge, etc.)

That's true. And being able to compile with `-E` has great benefits here, in my opinion.

> In the case of compiler errors (vector<bool>), unfortunately, I don't have a good solution for the logging.
BTW, to see the actual pass error of vector<bool>, we also need to uncomment the overload for vector<bool> (line 87-91)

Re: identifying specific test cases when failures occur, would it be possible to split the tests into groups to reduce the amount of commenting and uncommenting that would be required to identify the failure?

1tnguyen: Thanks @bmhowe23 for the suggestion. I followed a similar approach, but split the tests into headers instead (https://github.com/NVIDIA/cuda-quantum/pull/1385/commits/fce39625860e3e6ab73d2dda9fe1eef42ed43a19) 

Please let me know if it addresses the failure analysis need. I also added source location info of each test in the runtime logging.

@schweitzpgi I don't have a specific reproducer. It might well be a convoluted failure early on where we had these issues with small types, which have since been fixed.

targettests/Remote-Sim/args_synthesis.cpp
```diff
@@ -0,0 +1,559 @@
...
+      std::cout << "Testing: " << #func_ << "\n";                              \
+      auto counts = cudaq::sample(func_ END(INVOKE_LOOP_0 __VA_ARGS__));       \
+      counts.dump();                                                           \
+    });                                                                        \
+    return true;                                                               \
+  }()
+
+// Define some dummy argument values to use for testing.
+static const bool BOOL_VAL = true;
+static const char CHAR_VAL = 'a';
```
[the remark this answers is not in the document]
1tnguyen: Added tests for `short` and `vector<short>`.

targettests/Remote-Sim/args_synthesis.cpp
```diff
@@ -0,0 +1,32 @@
...
+// platforms. Note: we use the remote-mqpu platform in MLIR mode as a mock
+// environment for NVQC.
+
+// The test cases are defined in these headers.
+// Comment out include lines to exclude tests for debugging.
+#include "args_synthesis_test_cases/1_single_arg_tests.h"
+#include "args_synthesis_test_cases/2_two_pod_args_tests.h"
+#include "args_synthesis_test_cases/3_two_args_pod_container.h"
+#include "args_synthesis_test_cases/4_two_container_args.h"
+#include "args_synthesis_test_cases/5_mixed_args_tests.h"
```
schweitzpgi, line 25: I know you just added this, but it seems more klunky to me to have this test split out into separate include files.

I'd rather have it self-contained. I think you can get a lot of mileage here with some banner comments.

```c++
//===--------------------------------===//
// Section 1: Single Argument Tests
//===--------------------------------===//
```

schweitzpgi (approved):

targettests/Remote-Sim/args_synthesis.cpp
1tnguyen: Cool, will refactor this PR. Seems like an easy code refactor this time around :-)

1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1385/commits/c88c4fb8791a25b33037715c83975f7874b8a26b

bmhowe23 (approved):
👍

targettests/Remote-Sim/args_synthesis.cpp
bmhowe23: I see you later went with the `#define`'s...that looks good to me, so I'm marking this conversation as resolved.

PR #1390, opened 2024-03-13
[Core] Add poison op and fold optimization

schweitzpgi:
We can't use it yet since we're lagging LLVM, but this greases the skids to make folding of invalid constant array accesses to poison values.


1tnguyen (approved):
LGTM.

PR #1391, opened 2024-03-13
[Core] Add a charspan type to CC dialect

schweitzpgi:
Added support for std::string to the frontend as well.

Supports #1382


include/cudaq/Optimizer/Dialect/CC/CCTypes.td
```diff
@@ -226,4 +227,31 @@ def cc_StdVectorType : CCType<"Stdvec", "stdvec"> {
...
 }
 
+def cc_CharSpanType
+    : CCType<"Charspan", "charspan", [], "cudaq::cc::SpanLikeType"> {
+  let summary = "Proxy for a span over a std::string in cc";
+  let description = [{
+    This captures a C++ std::string(of char) type as an abstract CC type. The
+    intention is to allow us to reason about std::string at the CC level before
+    decomposing these into more refined data structures
+                .
```
amccaskey, line 237: should this period be all the way out here?

include/cudaq/Optimizer/Dialect/CC/CCTypes.td
```diff
@@ -226,4 +227,31 @@ def cc_StdVectorType : CCType<"Stdvec", "stdvec"> {
...
+
+  let extraClassDeclaration = [{
+    mlir::IntegerType getElementType() const {
+      return mlir::IntegerType::get(getContext(), 8);
+    }
+  }];
+}
+
+def IsSpanLikePred : CPred<"$_self.isa<cudaq::cc::SpanlikeType>()>">;
+def fir_SpanLikeType : Type<IsSpanLikePred, "dynamic length type">;
```
amccaskey, line 255: old `fir` prefix here

lib/Optimizer/CodeGen/LowerToQIR.cpp
```diff
@@ -1652,6 +1652,10 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
     auto eleTy = typeConverter.convertType(type.getElementType());
     return factory::stdVectorImplType(eleTy);
   });
+  typeConverter.addConversion([&typeConverter](cc::CharspanType type) {
+    auto eleTy = typeConverter.convertType(type.getElementType());
+    return factory::stdStringImplType(eleTy);
```
amccaskey, line 1657: Why use `stdStringImplType` or why does `stdStringImplType` return a `{elePtrTy, i64Ty, i64Ty, i64Ty}`? Shouldn't a span type just be `{elePtrTy, i64Ty}`?

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -107,6 +107,14 @@ inline mlir::LLVM::LLVMStructType stdVectorImplType(mlir::Type eleTy) {
   return mlir::LLVM::LLVMStructType::getLiteral(ctx, eleTys);
 }
 
+inline mlir::LLVM::LLVMStructType stdStringImplType(mlir::Type eleTy) {
+  auto *ctx = eleTy.getContext();
+  auto elePtrTy = cudaq::opt::factory::getPointerType(eleTy);
+  auto i64Ty = mlir::IntegerType::get(ctx, 64);
+  llvm::SmallVector<mlir::Type> eleTys = {elePtrTy, i64Ty, i64Ty, i64Ty};
```
amccaskey, line 114: I thought we were going for span types to be `{T*, i64}`?

schweitzpgi: Yes, we are.

include/cudaq/Optimizer/Dialect/CC/CCTypes.td
schweitzpgi: oops. Old habits die hard.

lib/Optimizer/CodeGen/LowerToQIR.cpp
schweitzpgi: My bad. We should be using the span form here. Doh!

lib/Optimizer/Builder/Factory.cpp
```diff
@@ -225,19 +225,27 @@ bool factory::hasHiddenSRet(FunctionType funcTy) {
...
-                                 .isa<cc::StdvecType, cc::StructType,
+                                 .isa<cc::SpanLikeType, cc::StructType,
                                       cc::ArrayType, cc::CallableType>());
 }
 
+cc::StructType factory::stlStringType(MLIRContext *ctx) {
+  auto i8Ty = IntegerType::get(ctx, 8);
+  auto ptrI8Ty = cc::PointerType::get(i8Ty);
+  auto i64Ty = IntegerType::get(ctx, 64);
+  auto padTy = cc::ArrayType::get(ctx, i8Ty, 16);
```
1tnguyen, line 236: Would it make sense to add a `static_assert` here to validate our layout assumption?
e.g.,
```
static_assert(sizeof(std::string) == sizeof(char*) + sizeof(std::size_t) + 16);
```

1tnguyen (approved):
LGTM 👍

PR #1395, opened 2024-03-14
Fix #1384 - synthesis of vectors with small integers.

### LINKED ISSUE #1384: `QuakeSynthesizer` cannot handle `std::vector<char>` argument type

schweitzpgi:
The constant array op is using at ArrayAttr, which does not naturally support integral sizes less than i32. This patch upcasts the short values to i32 to get around the issue. Another approach would be to have the constant array op take a DenseArrayAttr, which does support the shorter integral types. Might need a custom parser/pretty printer if we go that route?

Fix #1384

### DIFF
--- a/lib/Optimizer/Transforms/QuakeSynthesizer.cpp
+++ b/lib/Optimizer/Transforms/QuakeSynthesizer.cpp
@@ -219,6 +219,40 @@ LogicalResult synthesizeVectorArgument(OpBuilder &builder,
   return success();
 }
 
+template <typename A>
+std::vector<std::int32_t> asI32(const std::vector<A> &v) {
+  std::vector<std::int32_t> result(v.size());
+  for (auto iter : llvm::enumerate(v))
+    result[iter.index()] = static_cast<std::int32_t>(iter.value());
+  return result;
+}
+
+// TODO: consider using DenseArrayAttr here instead. NB: such a change may alter
+// the output of the constant array op.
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
LGTM.

PR #1399, opened 2024-03-14
Path fix in publishing

bettinaheim:


.github/workflows/publishing.yml
```diff
@@ -511,6 +511,9 @@ jobs:
           openmpi_buildcache=`cat $build_info | grep -o 'openmpi-build-cache: \S*' | cut -d ' ' -f 2`
           for file in `ls *zip`; do unzip "$file" && rm "$file"; done && cd -
 
+          echo "/tmp/install:"
+          find /tmp/install -type f
```
bettinaheim, line 515:
```suggestion
          if [ -d /tmp/install ]; then find /tmp/install -type f; fi
```

.github/workflows/publishing.yml
```diff
@@ -905,11 +905,11 @@ jobs:
       # https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
       - name: Create release
         run: |
-          for installer in `ls installers/`; do 
-            mv "installers/$installer" "$installer"
+          for installer in `find installers/ -type f`; do
```
bettinaheim, line 908:
```suggestion
          for installer in `find installers/ -type f -not -name '*.whl'`; do 
```

.github/workflows/publishing.yml
```diff
@@ -905,11 +905,11 @@ jobs:
       # https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
       - name: Create release
         run: |
-          for installer in `ls installers/`; do 
-            mv "installers/$installer" "$installer"
+          for installer in `find installers/ -type f`; do 
+            mv -v "$installer" "$(basename -- "$installer")"
```
bmhowe23, line 909: Question just because I'm curious: is the `--` truly needed? I was thinking it would only be needed if the `$installer` started with a hyphen for some reason, but maybe I'm missing something.

bmhowe23 (approved):

.github/workflows/publishing.yml
bettinaheim: I don't think it is needed, especially with the quotes. In general, it is useful for the case you mention but also to avoid command injection vulnerabilities.

PR #1401, opened 2024-03-14
Quick Start updates

bettinaheim:


docs/sphinx/snippets/cpp/quick_start.cpp
```diff
@@ -10,14 +10,18 @@
...
+    cx(qubits[0], qubits[i]);
+  }
+  mz(qubits);
 }
 
-int main() {
-  auto result = cudaq::sample(kernel);
-  result.dump(); // Example: { 1:500 0:500 }
+int main(int argc, char *argv[]) {
+  auto n_qubits = 1 < argc ? atoi(argv[1]) : 2;
```
anthony-santana, line 23:
```suggestion
  auto qubit_count = 1 < argc ? atoi(argv[1]) : 2;
```

docs/sphinx/snippets/python/quick_start.py
```diff
@@ -1,20 +1,25 @@
...
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
 # [Begin Documentation]
+import sys
 import cudaq
 
-print(f"Simulation Target = {cudaq.get_target().name}")
+print(f"Running on target {cudaq.get_target().name}")
+n_qubits = int(sys.argv[1]) if 1 < len(sys.argv) else 2
```
anthony-santana, line 14:
```suggestion
qubit_count = int(sys.argv[1]) if 1 < len(sys.argv) else 2
```

docs/sphinx/snippets/python/quick_start.py
```diff
@@ -1,20 +1,25 @@
...
 
 kernel = cudaq.make_kernel()
-qubit = kernel.qalloc()
-kernel.h(qubit)
-kernel.mz(qubit)
+qubits = kernel.qalloc(n_qubits)
+kernel.h(qubits[0])
+for i in range(1, nr_qubits):
+    cnot(qubits[0], qubits[i])
```
anthony-santana, line 20: `kernel.cx(qubits[0], qubits[i])` ?

docs/sphinx/using/quick_start.rst
```diff
@@ -1,110 +1,40 @@
...
-   .. code-block:: bash
-
-      python3 first_program.py
-
-   and confirm that `Simulation Target = nvidia` is printed to the console.
-
-   For further information on available targets, see :doc:`Backends <backends/backends>`.
-
-   Now that you have successfully run your first program, you are ready to move on to our :doc:`Basics Section <basics/basics>`.
+   To develop CUDA Quantum applications using Python,
```
anthony-santana, line 26: I thought the point of the Quick Start page was to put everything needed for a basic setup in one place. I think it slightly defeats the purpose to link someone to another page for those commands, when they could just copy and paste them directly where they're reading everything else from

anthony-santana (approved):

anthony-santana:
Approved so that you can merge this once it passes CI (looks like only a CLA issue)

docs/sphinx/using/quick_start.rst
bettinaheim: We need to make sure the instructions here are automatically validated, plus we want the instructions on PyPI to be the same as here. I figured out how to include markdown content and hence directly included the content in the PyPI readme here.

PR #1404, opened 2024-03-14
Adds: the ability to pass vector<bool> values to kernels as input.

schweitzpgi:


1tnguyen (approved):
LGTM. Tested by using this patch in the `remote-mqpu` argument tests.

PR #1409, opened 2024-03-15
Catch up to upstream

anthony-santana:
### Description

* Include upstream commit for updating pybind11
-->


schweitzpgi:
FWIW, I forgot to update the commit. "The local custom mods are no longer required." isn't true. We needed to mods for documentation building.

bmhowe23 (approved):

PR #1403, opened 2024-03-14
Enable CI testing for Python 3.8

bmhowe23:
I suspect we'll see a few minor test failures. Triggering the CI to find out which ones.

Then we need to change the branch protection rules to look for test results from 3.8 vice 3.9.

Resolves #1313.

PR #1410, opened 2024-03-16
Fix Python handling of kernel list parameters

bmhowe23:
While #1370 allowed list types to be specified in kernel signatures, they couldn't actually all be invoked with real calls like `cudaq.sample`. This PR add tests to demonstrate the problem, and provides a fix.

As a temporary test, use Python 3.8 for CI testing. #1403 will make this take effect for all future PRs.


amccaskey (approved):
thanks for the fix!

PR #1411, opened 2024-03-16
Add better support for timing analysis on simulator backends

bmhowe23:
This PR includes a few updates related to simulator timing.

1. It includes tests to measure per-gate timing information for simulated backends.
2. It also extends `ScopedTrace` by introducing "timing tags", which can be enabled and disabled on a per-tag basis, independently of the prior `CUDAQ_LOG_LEVEL=trace` method of extracting this timing data. The way this is enabled is via a new environment variable called `CUDAQ_TIMING_TAGS`, which takes a comma separated list of timing tags (integers). Tags are defined in a new `runtime/common/Timing.h`, which will undoubtedly grow over time.

An example of item 2 is the following instrumentation:
```bash
$ CUDAQ_TIMING_TAGS=1,3 python benchmark_with_observe.py
[2024-03-16 13:46:42.473] [tag=3] QPU::launchKernel executed in 94.194 ms.
[2024-03-16 13:47:37.461] - [tag=1] handleObservation flushGateQueue() executed in 54987.533 ms.
[2024-03-16 13:47:53.141] - [tag=1] QPU::handleObservation (after flush) executed in 15680.524 ms.
[2024-03-16 13:47:53.144] [tag=1] DefaultPlatform::resetExecutionContext executed in 70670.458 ms. (args = {observe})
```

In particular, it is very useful to be able to separate the following two items:
a) `handleObservation flushGateQueue()` time (which when combined with the other changes related to `cudaDeviceSynchronize` in this PR, represents the amount of time that the GPU is working on the base UCCSD circuit)
b) `QPU::handleObservation (after flush)`, which is the time spent doing per-term spin op operations after the base circuit.


runtime/cudaq/platform/qpu.h
```diff
@@ -50,7 +52,17 @@ class QPU : public registry::RegisteredType<QPU> {
...
   void handleObservation(ExecutionContext *localContext) {
-    if (localContext && localContext->name == "observe") {
+    // The reason for the 2 if checks is simply to do a flushGateQueue() before
+    // initiating the trace.
+    bool execute = localContext && localContext->name == "observe";
+    if (execute) {
+      cudaq::ScopedTrace trace(cudaq::TIMING_OBSERVE,
+                               "handleObservation flushGateQueue()");
+      getExecutionManager()->flushGateQueue();
+    }
```
bmhowe23, line 62: @1tnguyen - I think that since I am adding this code here, that the new `ScopedTrace` and `flushGateQueue` you added to `runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu / observe()` is no longer needed. Do you agree?

1tnguyen: Yes, I do. 

I missed the GitHub notification for this comment. Sorry about that!

runtime/common/Logger.cpp
```diff
@@ -7,15 +7,34 @@
...
+// NOTE: the only time that this list should be modified is at startup time.
+static std::set<int> &g_timingList() {
+  static std::set<int> timingList;
+  return timingList;
+}
+
+bool isTimingTagEnabled(int tag) {
+  // Note: this function is called very frequently, so it needs to be fast. It
+  // is assumed that g_timingList() contains a small number of elements
+  // (typically less than 10).
```
1tnguyen, line 34: Reading this comment, do you know whether `unordered_set` is faster/slower in terms of lookup time?
I got mixed results googling this :-)

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -334,6 +342,11 @@ public:
     randomEngine = std::mt19937(randomSeed);
   }
 
+  /// @brief Device synchronization
+  void synchronize() override {
+    cudaDeviceSynchronize();
```
1tnguyen, line 347:
```suggestion
     HANDLE_CUDA_ERROR(cudaDeviceSynchronize());
```

runtime/nvqir/cutensornet/simulator_cutensornet.cpp
```diff
@@ -137,6 +137,9 @@ void SimulatorTensorNetBase::resetQubit(const std::size_t qubitIdx) {
   m_state->applyQubitProjector(m_gateDeviceMemCache[projKey], qubitIdx);
 }
 
+/// @brief Device synchronization
+void SimulatorTensorNetBase::synchronize() { cudaDeviceSynchronize(); }
```
1tnguyen, line 141:
```suggestion
void SimulatorTensorNetBase::synchronize() { HANDLE_CUDA_ERROR(cudaDeviceSynchronize()); }
```

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -278,6 +282,12 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
                         {state.data(), state.data() + state.size()}};
   }
 
+  bool isStateVectorSimulator() const override {
+    if constexpr (std::is_same_v<StateType, qpp::ket>)
+      return true;
+    return false;
```
1tnguyen, line 288: Just a question: we cannot do `return std::is_same_v<StateType, qpp::ket>;` ?

1tnguyen (approved):
This looks great 👍!

runtime/nvqir/qpp/QppCircuitSimulator.cpp
bmhowe23: Yes, thanks for the suggestion. Updated in 7999c2e80.

runtime/common/Logger.cpp
bmhowe23: I did some quick tests, and it depends on how many timing tags you have enabled. For fewer elements (i.e. like 5 elements), it seems like `std::set` is faster, so I'll probably leave it like it is for now. We can always change it later.
(As an aside, it looks like the call to `std::chrono::high_resolution_clock::now()` is about 3-4x larger than the `set` or `unordered_set` calls right now.)

runtime/cudaq/platform/qpu.h
bmhowe23: Updated in 8634b4e2c

Issue #884, opened 2023-11-09
Enable automatic controlled-gate decomposition for `tensornet-mps` backend


1tnguyen:
Currently, we don't handle gates on 3 or more qubits in the `tensornet-mps` backend.
These gates are typically controlled gates. 

To lift this constraint, we can enforce mlir mode when using the backend and use our multi-control decomposition pass to decompose those gates.

PR #1420, opened 2024-03-20
MPS backend to support gates on 3+ qubits

1tnguyen:
### Description

(1) Support gates on 3+ qubits for the MPS backend via control gate decomposition. 
Experimenting with different strategies to support this feature, implementing it at the simulator implementation level seems to be the best option:

- A pure `__qpu__` kernel based solution only works for individual gates, but not for the `cuda::control` wrapper. The later adds control qubits at runtime (library mode).

- Need to support C++ kernels, kernel builder, and Python hence different runtimes (NVQIR and library mode execution manager).

Hence, intercepting these multi-controlled gates at the simulator-level prevent the need to spread this logic in multiple places.

(2) Remove all MPS-specific `#ifdef` in unit tests (bypass tests that have multi-control gates) 

(3) Reduce test time for tensornet backends: skip those tests that are not backend-specific; adjust a few tests to make it run faster for tensornet backends (`test_runtime_tensornet` and `test_runtime_tensornet-mps` about 10-15 mins each)

Fixes: https://github.com/NVIDIA/cuda-quantum/issues/884


runtime/nvqir/cutensornet/tensornet_state.h
```diff
@@ -19,6 +19,8 @@ class TensorNetState {
   std::size_t m_numQubits;
   cutensornetHandle_t m_cutnHandle;
   cutensornetState_t m_quantumState;
+  // Track id of tensors that are applied to the state tensors.
+  int64_t m_tensorId = -1;
```
schweitzpgi, line 23: `-1` magic value. What does it mean?

runtime/nvqir/cutensornet/simulator_mps_register.cpp
```diff
@@ -108,6 +110,245 @@ class SimulatorMPS : public SimulatorTensorNetBase {
...
+  // Swap gate implementation
+  using CircuitSimulator::swap;
+  void swap(const std::vector<std::size_t> &ctrlBits, const std::size_t srcIdx,
+            const std::size_t tgtIdx) override {
+    if (ctrlBits.empty())
+      return SimulatorTensorNetBase::swap(ctrlBits, srcIdx, tgtIdx);
+    // Controlled swap gate: using cnot decomposition of swap gate to perform
+    // decomposition.
+    {
+      std::vector<std::size_t> ctls = ctrlBits;
```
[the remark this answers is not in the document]
1tnguyen: Agreed. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1420/commits/d62007a492ba73558acebef6e37b0eabbfc2593e

unittests/integration/builder_tester.cpp
```diff
@@ -63,14 +63,14 @@ CUDAQ_TEST(BuilderTester, checkSimple) {
...
-    optimizer.max_line_search_trials = 10;
+    optimizer.initial_parameters = {0.359, 0.257};
+    optimizer.max_eval = 4;
+    optimizer.max_line_search_trials = 8;
     auto [opt_val_0, optpp] =
         cudaq::vqe(ansatz, gradient, h3, optimizer, 2, argMapper);
     printf("Opt-params: %lf %lf \n", optpp[0], optpp[1]);
     printf("<H3> = %lf\n", opt_val_0);
-    EXPECT_NEAR(opt_val_0, -2.045375, 1e-3);
+    EXPECT_NEAR(opt_val_0, -2.045375, 1e-2);
```
bmhowe23, line 73: If we made the initial parameters more accurate, why did we have to make the tolerance larger?

runtime/nvqir/cutensornet/tensornet_state.h
1tnguyen: It's intended to represent an invalid index value. When applying gate tensors to the qubit state, `cutensornet` will return the index of the gate tensor. Previously, we just discarded this index, now I use it (stored in this var) as a way to tell if the underlying tensor network has been modified beyond an initial state.

unittests/integration/builder_tester.cpp
1tnguyen: We need these 3 things altogether: reduce the number of optimization runs (line search and iterations), better initial values, and relax the test conditions. If preferred, I'll restore the tolerance and increase the optimizer parameters a bit.

bmhowe23: Was this particular change to reduce the test runtime?

What would you think about updating  the `-2.045375` to a new value and leaving `1e-3`?

bmhowe23: If this is really just to reduce test runtime, and if -2.045375 is truly the "correct" answer, then I suppose I'm fine leaving it the way it is.

1tnguyen: Yes, it's really just to reduce the test runtime. I think I'd go with keeping the true value here for future references.

bmhowe23 (approved):

runtime/nvqir/cutensornet/tensornet_state.h
schweitzpgi: Sounds good.

I'd give the magic value a name so it can be found and understood though. Somewhere at the top of the file perhaps,
```c++
/// This is used to track whether the tensor state is default initialized vs initialized in some other state.
constexpr std::int64_t InvalidTensorIndexValue = -1;
```

1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1420/commits/bdce5e63d490551986f0191b6cc5fa45e943b527

PR #1424, opened 2024-03-20
Fix memory leak in CuStateVecCircuitSimulator.cu

### LINKED ISSUE #1423: GPU memory leak in `nvidia` target
### Required prerequisites

### Describe the bug

When running the example in #1421 on an x86 machine (Gorby) using the cuda-quantum Docker containers, a memory leak is appearing (as can be seen by watching memory usage in `nvidia-smi` grow without bounds). I believe this is unrelated to the ARM-specific issue in #1421, so that's why I'm creating a new issue to track it.

The memory leak appears in both 0.6.0 and 0.7.0.

See the example in #1421.

I believe the leak only occurs when `shots_count` is being overridden as it is being done in that example.

### Expected behavior

No memory leak

Not a regression

### Environment

- **CUDA Quantum version**: 0.6.0 and 0.7.0
- **Python version**: 3.10
- **C++ compiler**: N/A

### Suggestions

_No response_

bmhowe23:
Resolves #1423


1tnguyen (approved):
LGTM 💯

PR #1425, opened 2024-03-20
[docs] Update cutensornet example in CI

anthony-santana:
### Description

Sets the default qubit count to 5 for the cutensornet example to speed up CI


bmhowe23 (approved):
LGTM, but I would recommend updating the PR title to say `cuquantum` rather than `cutensornet` since this isn't really using the cutensornet backend.

PR #1428, opened 2024-03-21
[Core] Add complex support

schweitzpgi:

### DIFF
--- a/runtime/cudaq/qis/qvector.h
+++ b/runtime/cudaq/qis/qvector.h
@@ -56,9 +56,9 @@ public:
     qvector(std::vector<complex>{vector.begin(), vector.end()});
   }
   qvector(const std::initializer_list<double> &list)
-     : qvector(std::vector<complex>{list.begin(), list.end()}) {}
+      : qvector(std::vector<complex>{list.begin(), list.end()}) {}
   qvector(const std::initializer_list<complex> &list)
-     : qvector(std::vector<complex>{list.begin(), list.end()}) {}
+      : qvector(std::vector<complex>{list.begin(), list.end()}) {}
 
   /// @cond
   /// Nullary constructor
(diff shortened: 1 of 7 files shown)


runtime/cudaq/qis/qvector.h
```diff
@@ -52,10 +52,14 @@ class qvector {
     getExecutionManager()->initializeState(targets, vector.data());
   }
 
+  qvector(const std::vector<double> &vector) {
```
amccaskey, line 55: should this be `qvector(const std::vector<double>& vector) : qvector(std::vector<complex>{vector.begin(), vector.end()}) {}` ?

schweitzpgi: Good catch. Thanks.

test/AST-Quake/qalloc_initialization.cpp
```diff
@@ -24,19 +23,305 @@ struct Vanilla {
...
+// CHECK:           cc.store %[[VAL_8]], %[[VAL_12]] : !cc.ptr<complex<f64>>
+// CHECK:           %[[VAL_13:.*]] = cc.compute_ptr %[[VAL_10]][2] : (!cc.ptr<!cc.array<complex<f64> x 3>>) -> !cc.ptr<complex<f64>>
+// CHECK:           cc.store %[[VAL_9]], %[[VAL_13]] : !cc.ptr<complex<f64>>
+// CHECK:           %[[VAL_14:.*]] = quake.alloca !quake.veq<3>
+// CHECK:           %[[VAL_15:.*]] = quake.init_state %[[VAL_14]], %[[VAL_10]] : (!quake.veq<3>, !cc.ptr<!cc.array<complex<f64> x 3>>) -> !quake.veq<?>
+
+struct MooseTracks {
+  std::vector<bool> operator()() __qpu__ {
+    using namespace std::complex_literals;
+    cudaq::qvector v = {std::complex<double>{0.0, 1.0},
```
amccaskey, line 64: would be good to change these to have power of 2 number of elements.

amccaskey (approved):

Issue #1111, opened 2024-01-19
Remote-MQPU to support C++-17 standard


1tnguyen:
Currently, `--target remote-mqpu` does not work with C++-17 standard. 
We need to add C++17 support for this target.

schweitzpgi:
What is the error?

bmhowe23:
I believe you can search our codebase for `FIXME: https://github.com/NVIDIA/cuda-quantum/issues/1111`. Here is one example:
https://github.com/NVIDIA/cuda-quantum/blob/main/test/Remote-Sim/args_parsing.cpp#L11 .

I wonder if it's possible that it's also just missing the `%cpp_std` from those tests?

schweitzpgi:
The linked one has a `REQUIRES: c++20` which empirically avoids the broken configuration.

bettinaheim:
> The linked one [Ed: all of them] has a `REQUIRES: c++20` which empirically avoids the broken configuration.
> 
> Still would be good to know what the failure mode actually is.

I just checked. 3 tests pass when setting the appropriate --std=c++17 flag, the the failures for the others are the following:

These all run into the same issue, which looks related to serialization/deserialization (@1tnguyen I though we didn't have a new dependency there, do we?): 
- CUDAQ :: Remote-Sim/args_parsing.cpp 
- CUDAQ :: Remote-Sim/free_func.cpp
- CUDAQ :: Remote-Sim/observe.cpp
- CUDAQ :: Remote-Sim/sample.cpp
```
JIT session error: Symbols not found: [ _ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEcvSt17basic_string_viewIcS2_EEv ]
Failure value returned from cantFail wrapped call
[...]

Issue #1281, opened 2024-02-23
Misleading error (JIT compilation issue) for remote-mqpu backend when MPI plugin is not activated


bettinaheim:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

In some cases, the execution on the remote-mqpu backend fails with a JIT error along the lines of 
`JIT session error: Symbols not found: [ _Unwind_Resume, _ZNSaIcED2Ev, ...]`

The error is caused by the invokeWrappedKernel logic in /runtime/common/JIT.cpp. Specifically, I think we are running into something like this: https://stackoverflow.com/questions/57612173/llvm-jit-symbols-not-found
The _Unwind_Resume symbol is from the GNU C++ standard library, specifically from libsupc++.a. I double checked that the produced executable itself (a.out) contains that symbol, so I suspect it is indeed something about these lines that is not working as expected:

```
  // Resolve symbols that are statically linked in the current process.
  llvm::orc::JITDylib &mainJD = jit->getMainJITDylib();
  mainJD.addGenerator(llvm::cantFail(
      llvm::orc::DynamicLibrarySearchGenerator::GetForCurrentProcess(
          dataLayout.getGlobalPrefix())));
```

### Steps to reproduce the bug

Minimal repro:
Download the latest version of the CUDA Quantum installer for C++, or build it from source. Then run
```
container=`docker run -itd --rm ubuntu:22.04`
docker cp install_cuda_quantum.$(uname -m) $container:/tmp/
docker cp docs/sphinx/examples/cpp/algorithms/amplitude_estimation.cpp $container:/tmp
docker attach $container
apt-get update && apt-get install -y --no-install-recommends \
            wget ca-certificates libstdc++-11-dev libopenmpi-dev
chmod +x /tmp/install_cuda_quantum.x86_64
/tmp/install_cuda_quantum.x86_64 --accept && . /etc/profile

# Fails with the reported JIT exception during execution:
[...]

PR #1435, opened 2024-03-22
Fix QPUD Library Mode Installer Build

1tnguyen:
### Description

**Root cause**: missing symbols (e.g., those errors reported in #1281) is due to static linking to `libstdc++` (only installer build uses this configuration). Specifically, the incoming LLVM IR may reference symbols that `qpud` and its dependencies never saw.

Note: in #1281, when we activated the MPI plugin, everything worked because the MPI plugin `.so` is linked (dynamic) to `libstdc++.so`. That plugin is then loaded (`dlopen`) at runtime, hence all symbols are available via dynamic lookup.

**Fixes:**    

- Modify CMake config of the `qpud` tool to include the full `libstdc++` archive if static linking to `libstdc++` is required.

- Also, add the missing `export_executable_symbols_for_plugins`. Remove the unnecessary `setAutoClaimResponsibilityForObjectSymbols` (was probably a workaround for the missing `export_executable_symbols_for_plugins`).

- Unify the C++-20 and C++-17 paths in some of the `observe`/`sample` APIs using the existing `make_copyable_function` helper.

- Re-add the `targettests` lit test suite to installer CI check. Remove the bypass clause (`REQUIRES: c++20`) in `remote-mqpu` tests. 

Resolved: #1111 and #1281


runtime/cudaq/algorithms/observe.h
```diff
@@ -493,21 +493,13 @@ auto observe_async(const std::size_t qpu_id, QuantumKernel &&kernel, spin_op &H,
...
   auto kernelName = cudaq::getKernelName(kernel);
 
-#if CUDAQ_USE_STD20
   return details::runObservationAsync(
-      [&kernel, ... args = std::forward<Args>(args)]() mutable {
-        cudaq::invokeKernel(std::forward<QuantumKernel>(kernel),
-                            std::forward<Args>(args)...);
-      },
+      detail::make_copyable_function(
+          [&kernel, ... args = std::forward<Args>(args)]() mutable {
```
schweitzpgi, line 498: How does this syntax compile as C++17? It's using the same perfect forwarding syntax as the C++20 code without the ifdef guard.

schweitzpgi: Are we sure this template is being instantiated in the C++17 tests?

1tnguyen: I'm pretty sure that these functions are used in the `targettests/Remote-Sim/observe_async.cpp` and `targettests/Remote-Sim/sample_async.cpp` tests, which are both tested with `-std=c++17` for the installer CI.

I think the `make_copyable_function` helper does the trick in a way that is compatible with C++17.

targettests/Remote-Sim/observe_async.cpp
```diff
@@ -22,7 +22,16 @@ struct ansatz {
     cudaq::qvector q(2);
     x(q[0]);
     ry(theta, q[1]);
+// Note:
+// (1) The x<cudaq::ctrl> syntax requires C++-20.
```
bettinaheim, line 26: Since we compile the code with our own clang build, and this construct is eliminated during compilation, this syntax should be perfectly usable also when C++17 is requested. In face, nvq++ suppresses the warning about this being a C++20 feature by default. Did you run into any issues using it? 
In any case, the test should use x<cudaq::ctrl> regardless of C++ standard.

1tnguyen: > nvq++ suppresses the warning about this being a C++20 feature by default. 

In fact, this is the thing we want to check. Some of our public headers (e.g., `observe.h` or `sample.h`) have `#ifdef` switches for C++17 vs. C++20 compilation. Due to the default `-Wno-c++20-extensions` flag of `nvq++`, we didn't verify that the C++17 paths in these headers are actually free from C++20 code.  

In these test files, I added a specific `RUN` with `nvq++ -std=c++17 -Wc++20-extensions -Werror ` to verify that these C++17 paths in our public headers only contain C++17 code.

runtime/cudaq/algorithms/observe.h
1tnguyen: The change above has been reverted, i.e., keeping two different paths for C++17 and C++20. The pre-existing compiler error in the C++17 path is fixed by using the `make_copyable_function` helper.

Also, the unit tests (`Remote-Sim/observe_async.cpp` and  `Remote-Sim/sample_async.cpp`) are updated to exercise all code paths (the two overloads of `observe_async` and `sample_async`)

targettests/Remote-Sim/free_func.cpp
```diff
@@ -21,7 +19,11 @@ void ghz(std::size_t N) __qpu__ {
   cudaq::qvector q(N);
   h(q[0]);
   for (int i = 0; i < N - 1; i++) {
+#if CUDAQ_USE_STD20
     x<cudaq::ctrl>(q[i], q[i + 1]);
+#else
+    cx(q[i], q[i + 1]);
+#endif
```
schweitzpgi, line 26: It's better to make this 2 distinct tests. The C++20 variation will only work with C++20 and so on. But the C++17 version must work with C++17, C++20, C++23, ...

targettests/Remote-Sim/free_func.cpp
```diff
@@ -30,7 +32,11 @@ void ansatz(double theta) __qpu__ {
   cudaq::qvector q(2);
   x(q[0]);
   ry(theta, q[1]);
+#if CUDAQ_USE_STD20
```
schweitzpgi, line 35: Same.

targettests/Remote-Sim/observe.cpp
```diff
@@ -26,7 +24,11 @@ struct ansatz {
     cudaq::qvector q(2);
     x(q[0]);
     ry(theta, q[1]);
+#if CUDAQ_USE_STD20
```
schweitzpgi, line 27: Same

targettests/Remote-Sim/observe_async.cpp
```diff
@@ -22,7 +20,7 @@ struct ansatz {
     cudaq::qvector q(2);
     x(q[0]);
     ry(theta, q[1]);
-    x<cudaq::ctrl>(q[1], q[0]);
+    cx(q[1], q[0]);
```
schweitzpgi, line 23: Same and with all the rest below.

It's a little more work to do it correctly, but we'll run the different syntax of the tests in all the proper modes.

targettests/Remote-Sim/free_func.cpp
1tnguyen: Good point. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1435/commits/4ad7ecd84a4f0d7c8dac820095463a5234365331

targettests/Remote-Sim/free_func.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1435/commits/4ad7ecd84a4f0d7c8dac820095463a5234365331

targettests/Remote-Sim/observe.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1435/commits/4ad7ecd84a4f0d7c8dac820095463a5234365331

targettests/Remote-Sim/observe_async.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1435/commits/4ad7ecd84a4f0d7c8dac820095463a5234365331

schweitzpgi (approved):

runtime/common/JIT.cpp
```diff
@@ -109,10 +111,6 @@ void invokeWrappedKernel(std::string_view irString,
           return std::make_unique<llvm::SectionMemoryManager>();
         });
     llvm::Triple targetTriple(llvm::Twine(llvmModule->getTargetTriple()));
-    // IMPORTANT: need to setAutoClaimResponsibilityForObjectSymbols to true to
```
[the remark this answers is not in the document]
1tnguyen: Fixed https://github.com/NVIDIA/cuda-quantum/pull/1533.

PR #1438, opened 2024-03-25
Improved type handling for OpaqueArguments::packArgs

amccaskey:
Inject the kernel `FuncOp` when packing input runtime arguments into the opaque arg used by `altLaunchKernel`. This provides better handling of runtime argument types via the LLVM TypeSwitch utility.

Without this we should experience issues with regards to processing empty lists passed as runtime arguments. Currently we can only know that the argument is a `list` and infer the element type by checking the type of the list's first element, which does not exist if the list is empty. This PR brings the kernel argument signature to this processing step, and so we can check / process the args knowing the required argument type.

### DIFF
--- a/python/utils/OpaqueArguments.h
+++ b/python/utils/OpaqueArguments.h
@@ -9,8 +9,13 @@
 #pragma once
 
 #include "common/FmtCore.h"
+#include "cudaq/Optimizer/Dialect/CC/CCTypes.h"
 #include "cudaq/builder/kernel_builder.h"
 #include "cudaq/qis/pauli_word.h"
+
+#include "mlir/CAPI/IR.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+
 #include <chrono>
 #include <functional>
 #include <future>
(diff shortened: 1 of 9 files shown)


python/utils/OpaqueArguments.h
```diff
@@ -156,212 +163,197 @@ inline py::args simplifiedValidateInputArguments(py::args &args) {
...
-        std::vector<cudaq::pauli_word> *ourAllocatedArg =
-            new std::vector<cudaq::pauli_word>(casted.size());
-        for (std::size_t counter = 0; auto el : casted) {
-          auto pw = el.cast<cudaq::pauli_word>();
-          (*ourAllocatedArg)[counter++] = cudaq::pauli_word(pw.str());
-        }
-        argData.emplace_back(ourAllocatedArg, [](void *ptr) {
-          delete static_cast<std::vector<cudaq::pauli_word> *>(ptr);
+    auto kernelArgTy = kernelFuncOp.getArgument(i).getType();
+    llvm::TypeSwitch<mlir::Type, void>(kernelArgTy)
```
bmhowe23, line 202: I'm curious - does `TypeSwitch` construct all objects regardless of which `Case` you enter?

amccaskey: Eric may be the best to answer this, but I think that TypeSwitch constructs a visitor pattern under the hood and requires construction of only the lambda type overloads that you provide, while all other types default to the `Default()` method lambda you provide.

bmhowe23 (approved):
👍

PR #1440, opened 2024-03-25
Fix issue with `spin_op` result retrieval

### LINKED ISSUE #1396: Batched measurement outputs dont allign to expected results
### Required prerequisites

### Describe the bug

In summary, I have a circuit with 3 qubits and none of them are entangled with each other hence we can treat them as 3, 1 qubit circuits effectively. 

Therefore, whether I measure a qubit or not, it should not change the output of other qubits. 

Let us describe the bug with an example: 

```
import cudaq 
from cudaq import spin 
import numpy as np 

np.random.seed(4)

qubit_count = 3 
num_of_samples = 3 
num_of_params = 3
theta_vals = np.random.rand(num_of_samples, num_of_params)

kernel, thetas = cudaq.make_kernel(list)

qubits = kernel.qalloc(qubit_count)

kernel.ry(thetas[0], qubits[0])
kernel.rx(thetas[1], qubits[1])
kernel.rz(thetas[2], qubits[2])


```

```
#Helper function to extract results 

def print_results(num_of_samples, hamiltonian, results):
    for s in range(num_of_samples): 
        
        for h in range(len(hamiltonian)): 
            
            print(results[s][h].expectation())
```

```
hamiltonian = [spin.z(0)]
results = cudaq.observe(kernel, hamiltonian, theta_vals)

print_results(num_of_samples, hamiltonian, results)

0.5677470285445452 #sample 0, hamiltonian term 0
0.7552138976752758 #sample 1, hamiltonian term 0
0.5601126664678304 #sample 2, hamiltonian term 0

```

```
hamiltonian = [spin.z(0), spin.z(1)]

results = cudaq.observe(kernel, hamiltonian, theta_vals)

print_results(num_of_samples, hamiltonian, results)

0.5677470285445452 #sample 0, hamiltonian term 0
0.8539678547531366 #sample 0, hamiltonian term 1
0.7552138976752758 #sample 1, hamiltonian term 0
0.7663033716380596 #sample 1, hamiltonian term 1
0.5601126664678304 #sample 2, hamiltonian term 0
0.999980520849931 #sample 2, hamiltonian term 1
```

```
hamiltonian = [spin.z(0), spin.z(1), spin.z(2)]

results = cudaq.observe(kernel, hamiltonian, theta_vals)

print_results(num_of_samples, hamiltonian, results)

0.5677470285445452    #sample 0, hamiltonian term 0
0.0              #sample 0, hamiltonian term 1 
0.9999999310821295.   #sample 0, hamiltonian term 2
0.7552138976752758.    #sample 1, hamiltonian term 0
0.0              #sample 1, hamiltonian term 1
1.0000000335276127     #sample 1, hamiltonian term 2
0.5601126664678304     #sample 2, hamiltonian term 0
0.0              #sample 2, hamiltonian term 1
0.9999999288265826    #sample 2, hamiltonian term 2
```

Some of the outputs across the last 2 examples should be the same since the circuit is not entangled and hence measurement outcomes should not affect each other however the outcomes differ. 

NA

### Expected behavior

NA

Not a regression

### Environment

- **CUDA Quantum version**: latest
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

1tnguyen:
### Description

We have the logic to expand the term strings (used as index keys) for result look-up but it was limited to single-qubit cases (e.g., only expanding Z0 but not Z1 to the target number of qubits).

This caused the issue reported in #1396 whereby the second `spin_op` in that list is Z1. It wasn't expanded properly, hence reported as a zero result.

This PR fixed the issue and added a test based on the bug report.

Resolved #1396

### DIFF
--- a/runtime/common/ObserveResult.h
+++ b/runtime/common/ObserveResult.h
@@ -80,7 +80,7 @@ public:
     // Expand the string representation of the term to match the number of
     // qubits of the overall spin_op this result represents by appending
     // identity ops.
-    if (!data.has_expectation(termStr) && termStr.size() < numQubits)
+    if (!data.has_expectation(termStr))
       for (std::size_t i = termStr.size(); i < numQubits; i++)
         termStr += "I";
     return data.expectation(termStr);
--- a/runtime/common/ObserveResult.h
+++ b/runtime/common/ObserveResult.h
@@ -77,8 +77,11 @@ public:
     // on more than 1 qubit can be <ZIII...III>
     auto numQubits = spinOp.num_qubits();
     auto termStr = term.to_string(false);
-    if (!data.has_expectation(termStr) && termStr.size() == 1 && numQubits > 1)
-      for (std::size_t i = 1; i < numQubits; i++)
+    // Expand the string representation of the term to match the number of
+    // qubits of the overall spin_op this result represents by appending
+    // identity ops.
+    if (!data.has_expectation(termStr) && termStr.size() < numQubits)
+      for (std::size_t i = termStr.size(); i < numQubits; i++)
         termStr += "I";
     return data.expectation(termStr);
   }
--- a/python/tests/builder/test_observe.py
+++ b/python/tests/builder/test_observe.py
@@ -738,6 +738,46 @@ def test_combine_sweep():
         assert assert_close(expectedEnergy, sum, tolerance=1e-2)
 
 
+def test_batched_observe_results():
+    # Test bug #1396
+    # Seed for repeatability
+    np.random.seed(4)
+    qubit_count = 3
+    num_of_samples = 3
+    num_of_params = 3
+    theta_vals = np.random.rand(num_of_samples, num_of_params)
+    kernel, thetas = cudaq.make_kernel(list)
+    qubits = kernel.qalloc(qubit_count)
+    kernel.ry(thetas[0], qubits[0])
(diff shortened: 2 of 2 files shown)


runtime/common/ObserveResult.h
```diff
@@ -77,8 +77,11 @@ class observe_result {
     // on more than 1 qubit can be <ZIII...III>
     auto numQubits = spinOp.num_qubits();
     auto termStr = term.to_string(false);
-    if (!data.has_expectation(termStr) && termStr.size() == 1 && numQubits > 1)
-      for (std::size_t i = 1; i < numQubits; i++)
+    // Expand the string representation of the term to match the number of
+    // qubits of the overall spin_op this result represents by appending
+    // identity ops.
+    if (!data.has_expectation(termStr) && termStr.size() < numQubits)
```
bmhowe23, line 83: Nit: I don't think the 2nd expression is required (because it is handled in your for loop below), so it could be simplified if you wanted:
```suggestion
    if (!data.has_expectation(termStr))
```

bmhowe23 (approved):
👍

runtime/common/ObserveResult.h
1tnguyen: Good point. Thank!

PR #1443, opened 2024-03-26
[Core] Remove the use of IndexType

### LINKED ISSUE #925: Remove the use of IndexType in the bridge and builders
There are still a few places, such as in the invariant loop builder, where the code uses `mlir::IndexType` that can be replaced with the `i64` type. Removing the `index` type will help reduce a number of casts that clutter up the IR.

schweitzpgi:
Fix #925

Remove the last few traces of the mlir::IndexType from the compiler and update the tests accordingly.


khalatepradnya (approved):
Looks good to me.

PR #1462, opened 2024-04-01
[feature] Support custom unitary operations in simulation

khalatepradnya:
### Description
First step for supporting Unitary Synthesis (Ref: Issue https://github.com/NVIDIA/cuda-quantum/issues/1475)
Given an arbitrary unitary matrix, allow it to be used as a custom quantum operation.

This PR covers the following work items:
- [ ] Support in simulation for Python -
  - [x]  Kernel mode
  - [x]  Builder mode 
  - [x]  State vector simulators
  - [x]  Tensornet simulators

WIP: Parameterized custom operations in builder mode.

- [ ] Support in simulation for C++-
  - [x]  Library mode
  - [ ]  MLIR mode 

Not covered: Decomposition (synthesis)


python/cudaq/qis/qis.py
```diff
@@ -222,3 +222,12 @@ def compute_action(compute, action):
     Apply the U V U^dag given U and V unitaries.
     """
     raise_error
+
+
+def register_operation(unitary, operation_name=None):
```
khalatepradnya, line 227: Remove this.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+
+  Value mapUnitaryDataToLLVMPointer(Location &loc,
+                                    ConversionPatternRewriter &rewriter,
+                                    ModuleOp parentModule,
+                                    std::int64_t numUnitaryElements,
+                                    std::string &operationName,
+                                    ArrayRef<double> data) const {
+    auto f64Ty = rewriter.getF64Type();
+    SmallVector<std::int64_t> shape{numUnitaryElements};
+    auto tensorTy = VectorType::get(shape, f64Ty);
```
schweitzpgi, line 1095: The name of the variable is `tensor` but it is a `Vector` instead?

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+/// targets)
+class UnitaryOpRewrite : public OpConversionPattern<quake::UnitaryOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+
+  Value mapUnitaryDataToLLVMPointer(Location &loc,
+                                    ConversionPatternRewriter &rewriter,
+                                    ModuleOp parentModule,
+                                    std::int64_t numUnitaryElements,
+                                    std::string &operationName,
```
schweitzpgi, line 1091: Prefer `StringRef`

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+    auto tensorTy = VectorType::get(shape, f64Ty);
+    auto insertPoint = rewriter.saveInsertionPoint();
+    rewriter.setInsertionPointToStart(parentModule.getBody());
+
+    LLVM::GlobalOp realGlobalOp;
+    if (auto existingArr =
+            parentModule.lookupSymbol<LLVM::GlobalOp>(operationName))
+      realGlobalOp = existingArr;
+    else
+      realGlobalOp = rewriter.create<LLVM::GlobalOp>(
```
schweitzpgi, line 1104: I'm thinking this global should be created early and UnitaryOp should just carry a symbol name rather than a copy of the flattened matrix. Also, if it is a symbol the values in the matrix don't even need to be known at compile-time.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+        realGlobalOp.getSymName());
+    Value zero =
+        rewriter.create<LLVM::ConstantOp>(loc, rewriter.getIntegerType(64), 0);
+    return rewriter.create<LLVM::GEPOp>(loc, LLVM::LLVMPointerType::get(f64Ty),
+                                        realPartAddr, ValueRange{zero, zero});
+  }
+
+  Value getQubitArray(Location &loc, ConversionPatternRewriter &rewriter,
+                      ValueRange qubits, FlatSymbolRefAttr &symbolRef,
+                      FlatSymbolRefAttr &getSymbolRef) const {
```
schweitzpgi, line 1122:
```suggestion
  Value getQubitArray(Location loc, ConversionPatternRewriter &rewriter,
                      ValueRange qubits, FlatSymbolRefAttr symbolRef,
                      FlatSymbolRefAttr getSymbolRef) const {
```
In MLIR, there is no need to pass many types by reference. (They effectively are by default.)

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+
+    // Create (qubit...) -> Array* control array
+    auto ctx = rewriter.getContext();
+    Value eight = rewriter.create<arith::ConstantIntOp>(loc, 8, 32);
+    auto qirArrayTy = cudaq::opt::getArrayType(ctx);
+    auto i8PtrTy =
+        cudaq::opt::factory::getPointerType(IntegerType::get(ctx, 8));
```
schweitzpgi, line 1129: I thought there was a factory function for `char*` as well?

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+    Value numQubits =
+        rewriter.create<arith::ConstantIntOp>(loc, qubits.size(), 64);
+
+    auto createQubitsCall = rewriter.create<LLVM::CallOp>(
+        loc, qirArrayTy, symbolRef, ArrayRef<Value>{eight, numQubits});
+    auto controlsArr = createQubitsCall.getResult();
+    for (std::size_t i = 0; auto control : qubits) {
```
schweitzpgi, line 1136:
```suggestion
    for (auto iter : llvm::enumerate(qubits)) {
      auto i = iter.index();
      auto control = iter.value();
```
for consistency with the rest of the code.

Is the variable `qubits` actually required to be controls here?

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+      auto idx =
+          rewriter.create<LLVM::ConstantOp>(loc, rewriter.getI64Type(), i);
+      auto call = rewriter.create<LLVM::CallOp>(
+          loc, i8PtrTy, getSymbolRef, ArrayRef<Value>{controlsArr, idx});
+      Value pointer = rewriter.create<LLVM::BitcastOp>(
+          loc, cudaq::opt::factory::getPointerType(i8PtrTy), call.getResult());
+      auto cast = rewriter.create<LLVM::BitcastOp>(loc, i8PtrTy, control);
+      rewriter.create<LLVM::StoreOp>(loc, cast, pointer);
+      i++;
```
schweitzpgi, line 1145:
```suggestion
```

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
+    auto ctx = op.getContext();
+    auto parentModule = op->getParentOfType<ModuleOp>();
+    auto constantUnitary = op.getConstantUnitary();
+    auto opName = op.getOpName();
+
+    auto qirArrayTy = cudaq::opt::getArrayType(ctx);
+    auto i8PtrTy =
+        cudaq::opt::factory::getPointerType(IntegerType::get(ctx, 8));
```
schweitzpgi, line 1162:
```suggestion
        cudaq::opt::factory::getPointerType(ctx);
```
does the same thing at the moment. It is also better since typed pointers are going away. Hopefully soon.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1077,6 +1077,164 @@ class MyToMz : public OpConversionPattern<quake::MyOp> {
...
+    SmallVector<double> realPart(unitary.size()), imagPart(unitary.size());
+    for (std::size_t i = 0; auto &element : unitary) {
+      auto values = dyn_cast<DenseF32ArrayAttr>(element).asArrayRef();
+      realPart[i] = values[0];
+      imagPart[i] = values[1];
+      i++;
+    }
+
+    std::string opNameReal = llvm::formatv("{0}_real", opName);
+    std::string opNameImag = llvm::formatv("{0}_imag", opName);
```
schweitzpgi, line 1205:
```suggestion
    std::string opNameReal = opName.str() + "_real";
    std::string opNameImag = opName.str() + "_imag";
```
Don't think we need the overhead of fancy formatters to concat these.

schweitzpgi (commented):
Some initial comments.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1118,4 +1118,38 @@ def ZOp : OneTargetOp<"z", [Hermitian]> {
...
+                       OptionalAttr<ArrayAttr>:$constantUnitary,
+                       Optional<StdvecOf<[Complex<AnyFloat>]>>:$unitary,
+                       OptionalAttr<UnitAttr>:$is_adj,
+                       OptionalAttr<DenseBoolArrayAttr>:$negated_qubit_controls
+                       );
+  let results = (outs);
+
+  let assemblyFormat = [{
+    ( `<` `adj` $is_adj^ `>` )? (`[` $controls^ (`neg` $negated_qubit_controls^ )? `]`)?
+    $targets (`(` $unitary^ `)`)? `:` functional-type(operands, results) attr-dict
```
schweitzpgi, line 1149:
```suggestion
    $targets (`,` $unitary^ )? `:` functional-type(operands, results) attr-dict
```
Remove the parens to keep the syntax looking a bit more consistent.

How is the constant unitary supposed to appear here?

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -1808,7 +1966,7 @@ class QuakeToQIRRewrite : public cudaq::opt::QuakeToQIRBase<QuakeToQIRRewrite> {
     populateFuncToLLVMConversionPatterns(typeConverter, patterns);
 
     patterns.insert<GetVeqSizeOpRewrite, RemoveRelaxSizeRewrite, MxToMz, MyToMz,
-                    ReturnBitRewrite>(context);
+                    UnitaryOpRewrite, ReturnBitRewrite>(context);
```
schweitzpgi, line 1969: How does this not need the typeConverter?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -608,6 +608,63 @@ void quake::WrapOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    for (auto target : targets)
+      if (auto veqTy = dyn_cast<quake::VeqType>(target.getType())) {
+        if (!veqTy.hasSpecifiedSize())
+          return op->emitOpError(
+              "quake.unitary cannot have runtime-known veq<?> types for "
+              "constant-sized matrix data.");
+        numQubits += veqTy.getSize();
+      } else // otherwise, must be a quake.ref
+        numQubits++;
+
```
schweitzpgi, line 647:
```suggestion
      } else {
      // otherwise, must be a quake.ref
        numQubits++;
}
```

schweitzpgi:
I don't think we've tried adding many new core features without first prototyping it in the C++ compiler. I'm not even sure off the top of my head how this is meant to work in that context. So in that regard, this feels incomplete to me.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1118,4 +1118,38 @@ def ZOp : OneTargetOp<"z", [Hermitian]> {
   }];
 }
 
+def UnitaryOp : QuakeOp<"unitary", [AttrSizedOperandSegments]> {
+  let summary = "General unitary operation specified by constant unitary matrix data";
+  let description = [{
+    This operation affects a general user-specified unitary operation, where 
+    the unitary can be specified at compile-time or at runtime via its matrix representation.
+
+    The matrix data must be provided as a row-major 1D array. It can be a
```
1tnguyen, line 1127: Should we document the endian convention here as well along with the row-major convention?

runtime/cudaq/qis/managers/BasicExecutionManager.h
```diff
@@ -225,14 +228,16 @@ class BasicExecutionManager : public cudaq::ExecutionManager {
 
     if (!adjointQueueStack.empty()) {
       // Add to the adjoint instruction queue
-      adjointQueueStack.back().emplace_back(
-          mutable_name, mutable_params, mutable_controls, mutable_targets, op);
+      adjointQueueStack.back().emplace_back(mutable_name, mutable_params,
```
1tnguyen, line 231: Did we handle the adjoint of custom operations somewhere for non-self-adjoint matrices?
perhaps we need to add `adjoint(unitary)` to the `adjointQueueStack` if it is not self-adjoint here instead.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
khalatepradnya: Good point! 
@boschmitt: Thoughts on how to document the default qubit ordering?

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1118,4 +1118,38 @@ def ZOp : OneTargetOp<"z", [Hermitian]> {
   }];
 }
 
+def UnitaryOp : QuakeOp<"unitary", [AttrSizedOperandSegments]> {
```
schweitzpgi, line 1121: Adding this for reference.

Generally, adding a new op to either of our dialects will require the following testing:

- [ ] adding it to the roundtrip test
- [ ] adding specific cudaq-opt tests that test the ways it is deployed in passes
- [ ] adding a C++ test to make sure the bridge can/does generate the new op
- [ ] adding a backend test to make sure the Quake/CC is lowered to QIR and/or QIR base profile, etc. as expected

khalatepradnya:
Closing this PR to chunk the feature into smaller units of changes.
For example, PR#1708 and PR#1781

PR #1463, opened 2024-04-02
Fix Publishing validation failures

1tnguyen:
### Description

There are failures in Publishing validation, e.g., https://github.com/NVIDIA/cuda-quantum/actions/runs/8443846828

```
 loc(callsite("bernstein_vazirani.cpp":33:22 at "bernstein_vazirani.cpp":52:3)): error: 'quake.extract_ref' op invalid index [6] because >= size [6]
```

This only showed up in the `remote-mqpu` (probably hardware emulation too) since we performed argument synthesis and loop unrolling => the pass manager can detect indexing issues.

This PR fixes the example.

Tested by: running locally

```
nvq++ --enable-mlir --target remote-mqpu  /home/cuda-quantum/examples/cpp/basics/bernstein_vazirani.cpp && ./a.out 
nvq++ --target quantinuum --emulate /home/cuda-quantum/examples/cpp/basics/bernstein_vazirani.cpp && ./a.out
```

Both produced correct results (`Encoded bitstring` == `Measured bitstring`)

### DIFF
--- a/docs/sphinx/examples/cpp/basics/bernstein_vazirani.cpp
+++ b/docs/sphinx/examples/cpp/basics/bernstein_vazirani.cpp
@@ -16,7 +16,7 @@
 
 std::vector<int> random_bitstring(int qubit_count) {
   std::vector<int> vector_of_bits;
-  for (auto i = 0; i <= qubit_count; i++) {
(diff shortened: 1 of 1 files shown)


bmhowe23 (approved):
👍

PR #1465, opened 2024-04-02
Fix a thread-safety issue in `jitAndCreateArgs`

1tnguyen:
### Description

There is an intermittent crash when running `test_remote_platform.py::test_multi_qpus` first reported in the `features/qalloc` branch's CI.

The backtrace of such failure points to a `llvm::report_fatal_error` in `getOrLoadDialect` (when the `PassManager::run` method is called).
```
#5  0x00007fbdaf48214f in llvm::report_fatal_error(llvm::Twine const&, bool) [clone .cold] () from /usr/local/cudaq/mlir/_mlir_libs/libCUDAQuantumPythonCAPI.so
#6  0x00007fbdb15d1812 in mlir::MLIRContext::getOrLoadDialect(llvm::StringRef, mlir::TypeID, llvm::function_ref<std::unique_ptr<mlir::Dialect, std::default_delete<mlir::Dialect> > ()>) () from /usr/local/cudaq/mlir/_mlir_libs/libCUDAQuantumPythonCAPI.so
```

i.e.,

```
 if (impl.multiThreadedExecutionContext != 0)
      llvm::report_fatal_error(
          "Loading a dialect (" + dialectNamespace +
          ") while in a multi-threaded execution context (maybe "
          "the PassManager): this can indicate a "
          "missing `dependentDialects` in a pass for example.");
```          

Hence, this PR fixes that by adding a mutex guard around the `PassManager` part of `jitAndCreateArgs`.

Note: this fix is not specific to the `features/qalloc`  branch. Perhaps the timing of the async tests on that branch is more susceptible to this failure mode.


amccaskey (approved):
+1

amccaskey:
@schweitzpgi for qalloc viz

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -56,18 +56,22 @@ jitAndCreateArgs(const std::string &name, MlirModule module,
...
-    cudaq::opt::addPipelineToQIR<>(pm);
-    if (failed(pm.run(cloned)))
-      throw std::runtime_error(
-          "cudaq::builder failed to JIT compile the Quake representation.");
-
+    static std::mutex g_passManagerMutex;
+    {
+      // Applying passes is not thread-safe, especially in the same MLIRContext.
+      std::scoped_lock<std::mutex> lock(g_passManagerMutex);
+      PassManager pm(context);
```
bmhowe23, line 63: We have a lot more places where we apply passes. Do we need locks there, too? This is a bit confusing to me. Unless I'm misinterpreting something, I'm afraid this is a band-aid for a deeper problem. Do we need to pre-load the dialects in a single-threaded context and then apply passes after they're loaded?

1tnguyen: Looking into this, I 'think' the root cause is changes in the way Python `kernel_builder` is handled: from being a pybind11 binding into native Python code. Previously, we had these `kernel.jitCode();` calls (on the main thread) before launching these async. threads hence all the `PassManager`'s dependent dialects are loaded into the context. 

Note: this issue is more prevalent on the first time we run the `PassManager` in a context. The new decorator mode has the `compile` method (which does run a `PassManager`) executed beforehand (main Python thread), thus doesn't have this issue.

I think an alternative solution could be to add method to `kernel_builder` similar to `compile` of `kernel_decorator`, which does use the pass manager on the main thread. @amccaskey Interested to hear your thoughts. 

Also, I might be missing something reading MLIR code but I thought the list of dialects to be looked up/loaded is pass dependent. If so, the mutex locking solution could be more generic. We could go a bit further by manually doing the `getDependentDialects` then load them in a mutex-locked scope before running the `PassManager`.

bmhowe23: Thanks, @1tnguyen. I am also interested in @amccaskey's thoughts, but I suspect we should fully track this down and either add comments to the code as to why this `PassManager` invocation is special (and therefore needing a lock) or go for a more generic solution. I know that `initializeMLIR()` calls `registerAllPasses()`; would that prevent this issue if that were called? Interestingly enough, that call is not thread-safe because there is an unprotected static variable in use, but perhaps we should call that function during some earlier step, like when `import cudaq` is called. If that function is already being called in this path, is it possible that we need to add additional logic to `registerAllPasses()`?

schweitzpgi: The PassManager is multithreaded and runs multiple threads within passes.

Loading dialects, etc., is done ad hoc. Typically, a tool loads "everything it needs" in a top-level set of registration calls. Once registered, there shouldn't be a need to re-register or re-load dialects. On the other hand, it is possible to let that be more ad hoc and let the pass manager do it prior to running a pass. And if the code calling the pass manager is itself multithreaded, then it will probably encounter some race conditions within the pass manager itself. That is, I don't think the pass manager is meant to be instantiated and have competing threads racing one another within that pass manager. That's sort of like having employees report to _n_ different managers at the same time, I guess.

schweitzpgi: The `import cudaq` probably ought to be single threaded and must be done before much of anything else can be done. So I'm inclined to agree with Ben here; that seems like the best place to register and load all the MLIR passes, dialects, and whatnot. (And if `import cudaq` could be executed concurrently by different threads, it should be easy to use a mutex to guarantee the registrations parts.)

1tnguyen: This is a bit more complicated than I thought. There is a gotcha here: the `CodeGenDialect` is a pass dependent (e.g., `QuakeToQIR`) yet is internal to the pass system (not exposed in public headers); hence we would have missed it even if `registerAllDialects`/`loadAllAvailableDialects`/etc. are called as usual.

@schweitzpgi Any objections to moving some of those headers from `lib/Optimizer/CodeGen` to `include/cudaq/Optimizer/CodeGen` (so that we can add `CodeGenDialect` to the `DialectRegistry` similar to the `CCDialect` and  `QuakeDialect`)?

amccaskey: I defer to whatever you guys find here. I'm not sure adding `compile` to kernel_builder is the right solution though, since the kernel_builder is built on the fly at runtime, and there is no notion of lazy compilation (which is the case for the decorator, defers compilation to the call site).

1tnguyen:
superseded by https://github.com/NVIDIA/cuda-quantum/pull/1474

Issue #1431, opened 2024-03-22
Endian representation bug in `QppCircuitSimulator::observe`


1tnguyen:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

In PR #1082, we reverted the endianness of `qpp` state vector but didn't adjust the `QppCircuitSimulator::observe` implementation, which uses `spin_op::to_matrix` to construct the full Hamiltonian matrix. This matrix and the state vector need to be in sync. w.r.t. endianness. 

Note: this code path is **not** active by default

 

### Steps to reproduce the bug

With CUDA-Q version 0.7.0,

```
import cudaq
import os

os.environ["CUDAQ_OBSERVE_FROM_SAMPLING"] = "0"
cudaq.set_target("qpp-cpu")
kernel = cudaq.make_kernel()
qreg = kernel.qalloc(2)
kernel.x(qreg[0])
hamiltonian = cudaq.spin.z(0)

print("<Z0> =", cudaq.observe(kernel, hamiltonian, shots_count=-1).expectation())
```

This is expected to print `-1` but will print `1`.
  
Change the `hamiltonian` to `z(1)` will produce -1, indicating an endian issue.

### Expected behavior

Correct calculation for all modes.

### Is this a regression? If it is, put the last known working version (or commit) here.

Since PR 1082

### Environment

- **CUDA Quantum version**: 0.7.0
- **Python version**: 3.10
- **C++ compiler**: gcc12
- **Operating system**: Ubuntu22.04

### Suggestions

There is another implementation of this 'direct' calculation in https://github.com/NVIDIA/cuda-quantum/blob/main/unittests/backends/qpp_observe/QPPObserveBackend.cpp#L22.

Not sure why we have two different implementations. We might want to consolidate them.

schweitzpgi:
C++ generates the expected results.

sacpis:
It is working as expected now. 

```
<Z0> = -1.0
```

[...]

PR #1467, opened 2024-04-02
Improve cudaq::state data handling

amccaskey:
Note: this also fixes https://github.com/NVIDIA/cuda-quantum/issues/1431 by changing the endian of `spin_op::to_matrix()`


runtime/nvqir/cutensornet/mps_simulation_state.cpp
```diff
@@ -0,0 +1,254 @@
...
+    numModes[i] = m_mpsTensors[i].extents.size();
+    numModes[mpsNumTensors + i] = mpsOtherTensors[i].extents.size();
+    tensExtents[i] = m_mpsTensors[i].extents;
+    tensExtents[mpsNumTensors + i] = mpsOtherTensors[i].extents;
+    tensAttr[i] = cutensornetTensorQualifiers_t{0, 0, 0};
+    tensAttr[mpsNumTensors + i] = cutensornetTensorQualifiers_t{0, 0, 0};
+  }
+  std::vector<std::vector<int32_t>> tensModes(numTensors);
+  int32_t umode = 0;
+  for (int i = 0; i < mpsNumTensors; ++i) {
```
1tnguyen, line 66: There could be a one-qubit edge case here: single MPS tensor of rank 1.

runtime/common/SimulationState.h
```diff
@@ -0,0 +1,201 @@
...
+
+public:
+  /// @brief Runtime-known precision for the simulation data
+  enum class precision { fp32, fp64 };
+
+  /// @brief Simulation states are made up of a
+  /// vector of Tensors. Each tensor tracks its data pointer
+  /// and the tensor extents.
+  struct Tensor {
+    void *data = nullptr;
```
1tnguyen, line 94: Should we have a `toHost` method so that users can retrieve the full tensor (e.g., MPS tensor) as host data at once?
(just for convenience since users can also do `cudaMemCpy` in user code)

runtime/common/SimulationState.h
```diff
@@ -0,0 +1,200 @@
...
```
schweitzpgi, line 95: Should we have a layout parameter (to be like numpy arrays)?

taylorpatti:
Are CuPy array to state conversions intended to be supported? The array to state constructor that works in this branch is `cudaq.State.from_data`, but it generates an `Implicit conversion to a NumPy array is not allowed.` error.

1tnguyen:
> Are CuPy array to state conversions intended to be supported? The array to state constructor that works in this branch is `cudaq.State.from_data`, but it generates an (edited)
> 
> `TypeError: from_data(): incompatible function arguments. The following argument types are supported: 1. (arg0: Buffer) -> cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.State`
> 
> error.

Yes, it should be supported. Currently, the Python binding for `from_data` is missing CuPy support. I'll add that shortly.

Side notes: 

(1) Computing overlap between a GPU simulation state (state retrieved from a `custatevec`-based backend) and a CuPy array is already supported.

(2) The feature to load a `State` (e.g., after constructed from a CuPy array) directly to the simulator backend to do simulation will require the new `qalloc` feature under development in the `features/qalloc` branch in case you're also interested in that feature.

taylorpatti:
Thanks, Thien. Yes, I'd like feature (2) - I don't mind if I have to load the data into a `State` instance first or not, but I do need to load in specific, non-product states.

Is there a timeline on this feature? I've been chatting with Justin for a couple weeks about using this CuPy/statevec friendly branch instead of another simulator for a research collaboration, but I do need to start coding the main components pretty immediately now.

1tnguyen:
> Thanks, Thien. Yes, I'd like feature (2) - I don't mind if I have to load the data into a `State` instance first or not, but I do need to load in specific, non-product states.
> 
> Is there a timeline on this feature? I've been chatting with Justin for a couple weeks about using this CuPy/statevec friendly branch instead of another simulator for a research collaboration, but I do need to start coding the main components pretty immediately now.

Both this and the `qalloc` features should be available in the next CUDA-Q release. As (2) is the main use case, `features/qalloc` branch contains under-development APIs to support it. We can sync up offline to see when you can start experimenting with the `features/qalloc` branch.

PR #1468, opened 2024-04-02
[c++17] changes to cudaq headers for pure c++17 environment

schweitzpgi:
The driver and headers were allowing all C++20 syntax to pass, even in C++17 mode. This was happening by leaning into the idea that a C++20 compiler was available, it was just being told to compile as C++17. This meant that in a pure C++17 environment, which didn't know anything about C++20 whatsoever, would choke on the CUDA-Q syntax.

This PR disables the use of the offending syntax in the headers _unless_ a C++20 (or better) compiler is truly in use. This distinction means that CUDA-Q programs written for C++17 _may be required to use a different syntax_ in order to compile.

For example, the following C++20 syntax
```c++
  // C++20 only!
  void qernel(cudaq::qubit &c, cudaq::qubit &q) __qpu__ {
    x<cudaq::ctrl>(c, q);
  }
```
must be written as
```c++
  // C++17 or C++20
  void qernel(cudaq::qubit &c, cudaq::qubit &q) __qpu__ {
    cx(c, q);
  }
```

### DIFF
--- a/runtime/cudaq/qis/qubit_qis.h
+++ b/runtime/cudaq/qis/qubit_qis.h
@@ -43,7 +43,11 @@ ConcreteQubitOp(h) ConcreteQubitOp(x) ConcreteQubitOp(y) ConcreteQubitOp(z)
 inline QuditInfo qubitToQuditInfo(qubit &q) { return {q.n_levels(), q.id()}; }
 inline bool qubitIsNegative(qubit &q) { return q.is_negative(); }
 
-/// @brief This function will apply the specified `QuantumOp`. It will check the
+#if CUDAQ_USE_STD20
+
+/// C++20 variant using templates.
+
+/// This function will apply the specified `QuantumOp`. It will check the
 /// modifier template type and if it is `base`, it will apply the operation to
 /// any qubits provided as input. If `ctrl`, it will take the first `N-1` qubits
 /// as the controls and the last qubit as the target.
@@ -57,13 +61,13 @@ void oneQubitApply(QubitArgs &...args) {
   // Get the number of input qubits
   constexpr std::size_t nArgs = sizeof...(QubitArgs);
 
-  // Map the qubits to their unique ids and pack them into a std::array
+  // Map the qubits to their unique ids and pack them.
   std::vector<QuditInfo> quditInfos{qubitToQuditInfo(args)...};
   std::vector<bool> qubitIsNegated{qubitIsNegative(args)...};
 
-  // If there are more than one qubits and mod == base, then
-  // we just want to apply the gate to all qubits provided
-  if constexpr (nArgs > 1 && std::is_same_v<mod, base>) {
+  // If mod == base, then we just want to apply the gate to all qubits provided.
+  // This is a broadcast application.
+  if constexpr (std::is_same_v<mod, base>) {
     for (auto &qubit : quditInfos)
       getExecutionManager()->apply(gateName, {}, {}, {qubit});
 
@@ -71,13 +75,13 @@ void oneQubitApply(QubitArgs &...args) {
     return;
   }
 
-  // If we are here, then `mod` must be control or adjoint
-  // Extract the controls and the target
+  // If we are here, then `mod` must be control or adjoint. Extract the controls
+  // and the target
   std::vector<QuditInfo> controls(quditInfos.begin(),
                                   quditInfos.begin() + nArgs - 1);
 
-  // If we have controls, check if any of them
-  // are negative controls, and if so apply an x
+  // If we have controls, check if any of them are negative controls, and if so
+  // apply an x.
   if (!controls.empty())
     for (std::size_t i = 0; i < controls.size(); i++)
       if (qubitIsNegated[i])
@@ -87,10 +91,9 @@ void oneQubitApply(QubitArgs &...args) {
   getExecutionManager()->apply(gateName, {}, controls, {quditInfos.back()},
                                std::is_same_v<mod, adj>);
 
-  // If we did apply any X ops for a negative control,
-  // we need to reverse it
-  if (!controls.empty())
-    for (std::size_t i = 0; i < controls.size(); i++)
+  // If we did apply any X ops for a negative control, we need to reverse it.
+  if (!controls.empty()) {
+    for (std::size_t i = 0; i < controls.size(); i++) {
       if (qubitIsNegated[i]) {
         getExecutionManager()->apply("x", {}, {}, {controls[i]});
         // fold expression which will reverse the negation
@@ -101,16 +104,14 @@ void oneQubitApply(QubitArgs &...args) {
             }(),
             ...);
       }
+    }
+  }
 }
 
-/// @brief This function will apply a multi-controlled operation with the given
-/// control register on the single qubit target.
-#if CUDAQ_USE_STD20
+/// This function will apply a multi-controlled operation with the given control
+/// register on the single qubit target.
 template <typename QuantumOp, typename mod = ctrl, typename QubitRange>
   requires(std::ranges::range<QubitRange>)
-#else
-template <typename QuantumOp, typename mod, typename QubitRange>
-#endif
 void oneQubitApplyControlledRange(QubitRange &ctrls, qubit &target) {
   // Get the name of the operation
   auto gateName = QuantumOp::name();
@@ -125,18 +126,6 @@ void oneQubitApplyControlledRange(QubitRange &ctrls, qubit &target) {
                                {cudaq::qubitToQuditInfo(target)});
 }
 
-#if CUDAQ_USE_STD20
-#define TEMPLATE(SORT)                                                         \
-  template <typename mod = SORT, typename QubitRange>                          \
-    requires(std::ranges::range<QubitRange>)
-#else
-#define TEMPLATE(SORT)                                                         \
-  template <typename mod = SORT, typename QubitRange,                          \
-            typename = std::enable_if_t<!std::is_same_v<                       \
-                std::remove_reference_t<std::remove_cv_t<QubitRange>>,         \
-                cudaq::qubit>>>
-#endif
-
 #define CUDAQ_QIS_ONE_TARGET_QUBIT_(NAME)                                      \
   namespace types {                                                            \
   struct NAME {                                                                \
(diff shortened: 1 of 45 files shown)


runtime/cudaq/qis/qubit_qis.h
```diff
@@ -147,31 +136,158 @@ void oneQubitApplyControlledRange(QubitRange &ctrls, qubit &target) {
...
+  struct NAME {                                                                \
+    inline static const std::string name{#NAME};                               \
+  };                                                                           \
+  }                                                                            \
+  template <typename... QubitArgs>                                             \
+  void NAME(QubitArgs &...args) {                                              \
+    oneQubitApply<qubit_op::NAME##Op>(args...);                                \
+  }                                                                            \
+  template <typename... QubitArgs>                                             \
+  void c##NAME(QubitArgs &...args) {                                           \
```
1tnguyen, line 251: Do we need to provide an equivalence for adjoint, e.g., `sdg` and `tdg`?

1tnguyen:
LGTM. 

I also want to ask @bmhowe23's opinion regarding breaking/non-breaking classification of the syntax change from `x<ctrl>(a,b,c,d)` to `cx(a,b,c,d)`, etc. when compiling with `-std=c++17`.

bettinaheim:
@schweitzpgi What's the issue with the reasoning that the nvq++ compiler is necessarily used to compile the code and nvq++ contains C++20 support? Is there a scenario where the code with `nvq++ --std=c++17` would not compile or run when only a C++17 runtime is available? 

If we pull in this change, we should remove the -Wno-c++20-extensions flag [here](https://github.com/NVIDIA/cuda-quantum/blob/7a73033daf550a26420351543387dba34b450407/tools/nvqpp/nvq%2B%2B.in#L586C27-L586C48) and [here](https://github.com/NVIDIA/cuda-quantum/blob/7a73033daf550a26420351543387dba34b450407/tools/cudaq-quake/cudaq-quake.cpp#L346).

amccaskey:
> LGTM.
> 
> I also want to ask @bmhowe23's opinion regarding breaking/non-breaking classification of the syntax change from `x<ctrl>(a,b,c,d)` to `cx(a,b,c,d)`, etc. when compiling with `-std=c++17`.

traditionally, `cOP` has meant the singly-controlled version of `OP`. Then we'd see Toffoli gates (doubly-controlled) expressed as `ccx`. This change departs from this and allows the full `OP<ctrl>(...)` functionality for multi-controlled operations. I do not think we should do this since it departs from a norm. If a C++17 user wants a general multi-control operations on more than 1 control, they should use `cudaq::control(...)` on a kernel calling the `OP`.

runtime/cudaq/qis/qubit_qis.h
amccaskey: This allows `cOP(anycontrols..., target)` which I don't think is what we want.

bmhowe23:
> @schweitzpgi What's the issue with the reasoning that the nvq++ compiler is necessarily used to compile the code and nvq++ contains C++20 support? Is there a scenario where the code with `nvq++ --std=c++17` would not compile or run when only a C++17 runtime is available?

I assume the requirement that Eric is trying to satisfy is that cudaq.h, which includes `runtime/cudaq/qis/qubit_qis.h`, should be includable by any C++17 compiler (e.g. nvcc). However, I think this is already achievable.

For example, this file:
```cpp
#include "cudaq.h"
int test() {
    return 0;
}
```
is compilable with all 3 of these commands.
```bash
$ nvcc -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
$ g++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
$ /opt/llvm/bin/clang++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
```
Is the requirement that our full language is callable from C++17 code, or just that the standard cudaq.h header files cannot interfere with a C++17-only compiler?

schweitzpgi:
> > LGTM.
> > I also want to ask @bmhowe23's opinion regarding breaking/non-breaking classification of the syntax change from `x<ctrl>(a,b,c,d)` to `cx(a,b,c,d)`, etc. when compiling with `-std=c++17`.
> 
> traditionally, `cOP` has meant the singly-controlled version of `OP`. Then we'd see Toffoli gates (doubly-controlled) expressed as `ccx`. This change departs from this and allows the full `OP<ctrl>(...)` functionality for multi-controlled operations. I do not think we should do this since it departs from a norm. If a C++17 user wants a general multi-control operations on more than 1 control, they should use `cudaq::control(...)` on a kernel calling the `OP`.

Yes, this was a bit of a backdoor extension. I can remove it. Thanks for the feedback.

schweitzpgi:
> If we pull in this change, we should remove the -Wno-c++20-extensions flag [here](https://github.com/NVIDIA/cuda-quantum/blob/7a73033daf550a26420351543387dba34b450407/tools/nvqpp/nvq%2B%2B.in#L586C27-L586C48) and [here](https://github.com/NVIDIA/cuda-quantum/blob/7a73033daf550a26420351543387dba34b450407/tools/cudaq-quake/cudaq-quake.cpp#L346).

Taken care of in c16e5ffa9

runtime/cudaq/qis/qubit_qis.h
schweitzpgi: See change c16e5ffa9

schweitzpgi:
> > @schweitzpgi What's the issue with the reasoning that the nvq++ compiler is necessarily used to compile the code and nvq++ contains C++20 support? Is there a scenario where the code with `nvq++ --std=c++17` would not compile or run when only a C++17 runtime is available?
> 
> I assume the requirement that Eric is trying to satisfy is that cudaq.h, which includes `runtime/cudaq/qis/qubit_qis.h`, should be includable by any C++17 compiler (e.g. nvcc). However, I think this is already achievable.
> 
> For example, this file:
> 
> ```c++
> #include "cudaq.h"
> int test() {
>     return 0;
> }
> ```
> 
> is compilable with all 3 of these commands.
> 
> ```shell
> $ nvcc -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> $ g++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> $ /opt/llvm/bin/clang++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> ```
> 
> Is the requirement that our full language is callable from C++17 code, or just that the standard cudaq.h header files cannot interfere with a C++17-only compiler?

Yes, it should be true that the cuda.h file is (and should remain) legal C++17 and should compile with any C++17 compiler. That was the requirement: supporting C++17 only systems. We don't want to silently regress from this point, which is possible because we are disabling our compiler checks currently as a workaround for an omission.

Thanks for verifying.

bmhowe23:
> > > @schweitzpgi What's the issue with the reasoning that the nvq++ compiler is necessarily used to compile the code and nvq++ contains C++20 support? Is there a scenario where the code with `nvq++ --std=c++17` would not compile or run when only a C++17 runtime is available?
> > 
> > 
> > I assume the requirement that Eric is trying to satisfy is that cudaq.h, which includes `runtime/cudaq/qis/qubit_qis.h`, should be includable by any C++17 compiler (e.g. nvcc). However, I think this is already achievable.
> > For example, this file:
> > ```c++
> > #include "cudaq.h"
> > int test() {
> >     return 0;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is compilable with all 3 of these commands.
> > ```shell
> > $ nvcc -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> > $ g++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> > $ /opt/llvm/bin/clang++ -c -I/usr/local/cudaq/include -std=c++17 bentest.cpp -o bentest.o
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is the requirement that our full language is callable from C++17 code, or just that the standard cudaq.h header files cannot interfere with a C++17-only compiler?
> 
> Yes, it should be true that the cuda.h file is (and should remain) legal C++17 and should compile with any C++17 compiler. That was the requirement: supporting C++17 only systems. We don't want to silently regress from this point, which is possible because we are disabling our compiler checks currently as a workaround for an omission.
> 
> Thanks for verifying.

To be clear, the above test file and test compile commands worked for me without any of the changes in this PR. Were you seeing failures before this PR with a specific compiler, or was this issue found by inspection only?

schweitzpgi:
> To be clear, the above test file and test compile commands worked for me without any of the changes in this PR. Were you seeing failures before this PR with a specific compiler, or was this issue found by inspection only?

This was found by inspection. The current setup accepts C++20 programs even when adding the C++17 flag, which is a problem.

targettests/execution/conditional_sample.cpp
```diff
@@ -6,13 +6,8 @@
...
 
-// clang-format off
-// RUN: nvq++ -std=c++17 --enable-mlir %s -o %t
-// RUN: if [ $(echo %cpp_std | cut -c4- ) -ge 20 ]; then \
-// RUN:   nvq++ --enable-mlir %s -o %t && %t; \
-// RUN: fi
-// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t
-// clang-format on
+// REQUIRES: c++20
+// RUN: nvq++ --target quantinuum --emulate %s -o %t && %t
```
bettinaheim, line 10: Why was the test case `nvq++ --enable-mlir %s -o %t && %t` removed?

schweitzpgi: It was wrapped in the odd `if` construct. I'll add it back.

schweitzpgi:
ping: needs the rubberstamp

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -293,15 +525,44 @@ void swap(QubitArgs &...args) {
...
+void swap(Qubit &src, Qubit &target) {
+  static_assert(std::is_same<qubit, Qubit>::value,
+                "Cannot operate on a qudit with Levels != 2");
+  std::vector<QuditInfo> qubitIds{qubitToQuditInfo(src),
+                                  qubitToQuditInfo(target)};
+  getExecutionManager()->apply("swap", {}, {}, qubitIds);
+}
+
+template <typename... QubitArgs>
+void cswap(QubitArgs &...args, qubit &src, qubit &target) {
```
amccaskey, line 550: This is also a multi-control swap when cswap the name implies a singly-controlled swap

schweitzpgi: This one didn't use the template. I'll remove that.

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -252,26 +398,112 @@ void oneQubitSingleParameterControlledRange(ScalarAngle angle,
...
+  struct NAME {                                                                \
+    inline static const std::string name{#NAME};                               \
+  };                                                                           \
+  }                                                                            \
+  template <typename ScalarAngle, typename... QubitArgs>                       \
+  void NAME(ScalarAngle angle, QubitArgs &...args) {                           \
+    oneQubitSingleParameterApply<qubit_op::NAME##Op>(angle, args...);          \
+  }                                                                            \
+  template <typename ScalarAngle, typename... QubitArgs>                       \
+  void c##NAME(ScalarAngle angle, QubitArgs &...args) {                        \
```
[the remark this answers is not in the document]
schweitzpgi: Yep. Missed that one. Good catch.

amccaskey (approved):

PR #1470, opened 2024-04-03
Simplify MLIR code generated for simple Python for loops over range(N)

amccaskey:


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2182,6 +2182,25 @@ def visit_For(self, node):
 
         self.currentNode = node
 
+        # We can simplify `for i in range(N)` MLIR code immensely
+        # by just building a for loop with N as the upper value,
+        # no need to generate an array from the `range` call.
+        if isinstance(node.iter, ast.Call):
+            if node.iter.func.id == 'range' and len(node.iter.args) == 1:
```
bmhowe23, line 2189: Is it possible to handle things like `range(1,10)` or `range(10, 0, -1)` using similar logic?

amccaskey: Technically, I think this is possible. I'll look into it. But I think this is not implemented in the C++ bridge, you might double check me on that statement

bmhowe23 (approved):
👍

PR #1471, opened 2024-04-03
Prevent usage of cmake 3.29.0 in manylinux build images

bmhowe23:
Similar to #1441 but for manylinux this time.

[This](https://github.com/NVIDIA/cuda-quantum/actions/runs/8543730308/job/23408199009) is a sample build that is broken right now.

I chose version 3.26.4 because that's what is currently called out in `scripts/install_prerequisites.sh`.


khalatepradnya (approved):
👍🏽

PR #1472, opened 2024-04-03
[fix] [test] Updates to 'nvidia-mqpu' test - 'testLargeProblem'

khalatepradnya:
### Description
* Update the `nTerms` to match the `spin_op` random generator logic
  - For 4 qubits, `nTerms = 70`


bmhowe23 (commented):
As long as you've got the hood open, I would recommend setting the random seed in the call to `cudaq.SpinOperator.random` (e.g. `cudaq.SpinOperator.random(nQubits, nTerms, seed=13)` so it's repeatable.

bmhowe23 (approved):
👍

PR #1473, opened 2024-04-03
Make sure that we register and load all dialects when using `PassManager` at runtime

1tnguyen:
### Description

According to MLIR: "The process of loading a Dialect in the context is not thread-safe, which forces all involved Dialects to be loaded before the multi-threaded pass manager starts the execution."

This PR makes sure that we load all the dialects when MLIR is needed at runtime and, hence may be used in a multi-threading environment (async. execution).

- In particular, add a method to register the internal `CodeGenDialect`, which is only used by our passes (not the frontend).

- Make sure that we load all Dialects in C++ (CUDAQ `MLIRRuntime`) and Python (builder/decorator).

Tested by: running `python/tests/remote/test_remote_platform.py` with these changes on the `features/qalloc` branch.


1tnguyen:
Wrong target branch.

PR #1476, opened 2024-04-04
Add support for recently introduced cudaq::range(a,b,c) function.

schweitzpgi:
This allows the recently introduced range function to be handled in the C++ bridge correctly.


include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -197,6 +197,19 @@ createInvariantLoop(mlir::OpBuilder &builder, mlir::Location loc,
                                             mlir::Region &, mlir::Block &)>
                         bodyBuilder);
 
+/// Builds a monotonic loop. A monotonic loop is a loop that is guaranteed to
+/// execute the body of the loop from \p start to (but not including) \p stop
+/// stepping by \p step times. Exceptional conditions will cause the loop body
+/// to execute 0 times. Early exits are not allowed. This builder threads the
```
bmhowe23, line 203: > Early exits are not allowed.

Does this mean that `break` statements are not permitted?

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -67,11 +67,77 @@ static constexpr IntrinsicCode intrinsicTable[] = {
...
+
+    // Compute the total number of iterations, which is the value `n`, from a
+    // semi-open triple style notation. The set defined by the triple is `{ i,
+    // i+s, i+2*s, ... i+(n-1)*s }` where `i` is the initial value `%start`, `s`
+    // is the step, `%step`, and the value `i+(n-1)*s` is strictly in the
+    // interval `[start .. stop)` or `(stop .. start]` depending on whether step
+    // is positive or negative. Invalid triples, such as the step being zero or
+    // the lower and upper bounds being transposed will return a value of 0.
+    // Note that all three parameters are assumed to be signed values, which is
+    // required to have a decrementing loop.
```
bmhowe23, line 118: Should we throw an error somewhere upstream if unsigned values here to prevent edge cases where any of the values are >= 2^63? (Not sure where that should be ... perhaps in ConvertExpr.cpp?

include/cudaq/Optimizer/Builder/Factory.h
schweitzpgi: It may behave correctly, but it is highly discouraged.

lib/Optimizer/Builder/Intrinsics.cpp
bmhowe23: It could conceivably still happen with a small number of iterations, i.e. `range(2^63, 2^63+10, 1)`

schweitzpgi: Yes, I realize.

I will change the header to force signed types.

schweitzpgi: See f9ab89a29

test/AST-Quake/ranged_for.cpp
```diff
@@ -307,3 +307,48 @@ struct FreeRange {
 // CHECK:             cc.continue %[[VAL_15]] : i64
 // CHECK:           } {invariant}
 // clang-format on
+
+struct FreeRangeChicken {
+  void operator()(cudaq::qvector<> r, int Z, int N, int S) __qpu__ {
+    for (auto i : cudaq::range(Z, N, S)) {
```
bmhowe23, line 313: Is there already a separate issue covering docs updates for this new feature, or should we create a new one? In any case, we should probably specify any limitations w.r.t. `break` if needed.

bmhowe23 (approved):
LGTM

test/AST-Quake/ranged_for.cpp
schweitzpgi: It's in the task list of #1483

schweitzpgi: I'd add a milestone of 0.8, but I cannot.

PR #1478, opened 2024-04-05
[features/qalloc] Rework kernel_builder.

schweitzpgi:
The code in the kernel builder has been simplified.

### DIFF
--- a/runtime/cudaq/builder/kernel_builder.cpp
+++ b/runtime/cudaq/builder/kernel_builder.cpp
@@ -30,57 +30,11 @@
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Target/LLVMIR/ModuleTranslation.h"
 #include "mlir/Transforms/Passes.h"
-
 #include <numeric>
 
 using namespace mlir;
 
-extern "C" {
-void altLaunchKernel(const char *kernelName, void (*kernelFunc)(void *),
-                     void *kernelArgs, std::uint64_t argsSize);
-}
-
 namespace cudaq::details {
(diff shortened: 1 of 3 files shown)


runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -510,50 +463,45 @@ QuakeValue qalloc(ImplicitLocOpBuilder &builder, QuakeValue &size) {
...
+  // initializer data.
+  auto complexTy = ComplexType::get(builder.getF64Type());
+  auto stateDataTy = cudaq::cc::PointerType::get(
+      cudaq::cc::ArrayType::get(context, complexTy, size));
+  cudaq::cc::GlobalOp globalData;
+  {
+    OpBuilder::InsertionGuard guard(builder);
+    static std::size_t counter = 0;
+    builder.setInsertionPointToEnd(parentModule.getBody());
+    auto backingData = builder.getF64ArrayAttr(ArrayRef<double>{*dataPtr});
```
amccaskey, line 483: If I understand this correctly - the concern I have here is that we would be backing the actual state data into the ModuleOp as part of this GlobalOp. What if we have a 30 qubit problem, this will store all 2**30 complex values in the Module, right?

The current implementation was trying to get around the need to back the values into the module op. The data exists in user-space so the idea was to just keep it there and only set a pointer to the data in the Module / Global when it is needed (when the code is invoked)

schweitzpgi:
After further discussion this approach is "too simplified". Closing this one.

PR #1482, opened 2024-04-05
Update kernel_builder to support vector state input. Implement CuStateVecSimulator kron prod

amccaskey:
Support 
```cpp
auto [kernel, initState] =
        cudaq::make_kernel<std::vector<std::complex<double>>>();
auto qubits = kernel.qalloc(initState);
...
```

Implement on-GPU kronecker product for allocating qubits in a specific state (when a state already exists on GPU). 

Turn builder_tester test on for all this now that QPP and CuStateVec support it.


lib/Optimizer/CodeGen/LowerToQIR.cpp
```diff
@@ -138,6 +138,18 @@ class QmemRAIIOpRewrite
...
+    // know if its f32 or f64
+    Type eleTy = rewriter.getF64Type();
+    if (auto llvmPtrTy = dyn_cast<LLVM::LLVMPointerType>(ccState.getType())) {
+      auto ptrEleTy = llvmPtrTy.getElementType();
+      if (auto llvmStructTy = dyn_cast<LLVM::LLVMStructType>(ptrEleTy))
+        if (llvmStructTy.getBody().size() == 2 &&
+            llvmStructTy.getBody()[0] == llvmStructTy.getBody()[1] &&
+            llvmStructTy.getBody()[0].isF32())
+          eleTy = rewriter.getF32Type();
+    }
```
schweitzpgi, line 151: Shouldn't this always be a `complex<f>*` type?  If so, we should probably add hard checks here. If we don't find that type, then raise an error on this Op as being broken.

amccaskey: I'll update to throw an error

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -194,39 +261,99 @@ protected:
...
   }
 
+  /// @brief Nice utility function to have to print the state vector contents on
+  /// GPU.
+  void printStateFromGPU(const std::string &name, void *ptr, std::size_t size) {
+    std::vector<std::complex<ScalarType>> tmp(size);
+    cudaMemcpy(tmp.data(), ptr, size * sizeof(std::complex<ScalarType>),
+               cudaMemcpyDeviceToHost);
+    for (auto &r : tmp)
+      printf("%s: %.12lf\n", name.c_str(), r.real());
```
1tnguyen, line 271: Should we also print the imaginary part?

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -194,39 +261,99 @@ protected:
...
+    // Allocate new vector to place the kron prod result
+    void *newDeviceStateVector;
+    HANDLE_CUDA_ERROR(cudaMalloc((void **)&newDeviceStateVector,
+                                 stateDimension * sizeof(CudaDataType)));
+
+    // Place the state data on device. Could be that
+    // we just need the zero state, or the user could have provided one
+    void *otherState;
+    HANDLE_CUDA_ERROR(cudaMalloc((void **)&otherState,
+                                 stateDimension * sizeof(CudaDataType)));
```
1tnguyen, line 333: We might be over-allocating this buffer. 
```suggestion
                                 (1UL << count) * sizeof(CudaDataType)));
```

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -194,39 +261,99 @@ protected:
...
-          (stateDimension + threads_per_block - 1) / threads_per_block;
-      setFirstNElements<<<n_blocks, threads_per_block>>>(
-          reinterpret_cast<CudaDataType *>(newDeviceStateVector),
-          reinterpret_cast<CudaDataType *>(deviceStateVector),
-          previousStateDimension);
-      HANDLE_CUDA_ERROR(cudaFree(deviceStateVector));
-      deviceStateVector = newDeviceStateVector;
+
+      // FIXME Handle case where data is already on GPU
+      std::vector<std::complex<ScalarType>> in(state, state + stateDimension);
```
1tnguyen, line 340: I think we might be accessing the input `state` buffer beyond its memory region here.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -194,39 +261,99 @@ protected:
...
+                                   cudaMemcpyHostToDevice));
     }
+
+    // Compute the kronecker product
+    kronprod<CudaDataType><<<n_blocks, threads_per_block>>>(
+        previousStateDimension,
+        reinterpret_cast<CudaDataType *>(deviceStateVector), (1UL << count),
+        reinterpret_cast<CudaDataType *>(otherState),
+        reinterpret_cast<CudaDataType *>(newDeviceStateVector));
+
```
1tnguyen, line 352: Since this is a fairly complex kernel, we might want to add `HANDLE_CUDA_ERROR(cudaGetLastError());` to detect any failures early.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
amccaskey: Yea definitely, will fix

runtime/nvqir/CircuitSimulator.h
```diff
@@ -596,7 +598,7 @@ class CircuitSimulatorBase : public CircuitSimulator {
 
   /// @brief Add the given number of qubits to the state.
   virtual void addQubitsToState(std::size_t count,
-                                const std::complex<double> *state = nullptr) {
+                                const void *state = nullptr) {
```
schweitzpgi, line 601:
```suggestion
                                 void *state = nullptr) {
```
Can we use parametric polymorphism as a type adaptor instead of a bare pointer here? (Something like a `std::variant`.)

include/cudaq/Optimizer/CodeGen/Pipelines.h
```diff
@@ -61,6 +61,9 @@ void addPipelineToQIR(mlir::PassManager &pm,
   if constexpr (QIRProfile) {
     cudaq::opt::addQIRProfilePipeline(pm, convertTo);
   }
+
+  // Make sure we don't have any leftover dead code
+  pm.addPass(mlir::createCanonicalizerPass());
```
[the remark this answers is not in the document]
amccaskey: bad rebase on my part

lib/Optimizer/CodeGen/LowerToQIR.cpp
```diff
@@ -1360,7 +1378,7 @@ class GlobalOpPattern : public ConvertOpToLLVMPattern<cudaq::cc::GlobalOp> {
     bool isReadOnly = global.getConstant();
     Attribute initializer = global.getValue().value_or(Attribute{});
     auto linkage =
-        global.getExternal() ? LLVM::Linkage::External : LLVM::Linkage::Private;
+        global.getExternal() ? LLVM::Linkage::Linkonce : LLVM::Linkage::Private;
```
schweitzpgi, line 1381: Great change! 👍

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -874,10 +911,11 @@ jitCode(ImplicitLocOpBuilder &builder, ExecutionEngine *jit,
   pm.addPass(cudaq::opt::createGenerateDeviceCodeLoader(/*genAsQuake=*/true));
   pm.addPass(cudaq::opt::createGenerateKernelExecution());
   optPM.addPass(cudaq::opt::createLowerToCFGPass());
-  // We want quantum allocations to stay where they are if we are simulating and
-  // have user-provided state vectors. This check could be better / smarter
-  // probably, in tandem with some synth strategy to rewrite initState with
-  // circuit synthesis result
+  // We want quantum allocations to stay where they are if
```
schweitzpgi, line 914: `clang-format` gets dumb about multiple line comments, particularly if it gets confused and starts over indenting code.

To fix these, you can convert the comment to a single line and re-run clang-format.

schweitzpgi (approved):
LGTM

1tnguyen (approved):
Looks great 💯

PR #1484, opened 2024-04-05
Compute the correct number of qubits.

schweitzpgi:
The initializer length must be 2**q where q is the number of qubits.


schweitzpgi:
Looks like more wheel failures. Good to go then.

amccaskey (approved):

PR #1487, opened 2024-04-08
Kernel builder cleanup.

schweitzpgi:
This is basically a NFC. Cleanup some of the kernel builder code. Modify some of the interfaces to use `const T &` arguments to avoid making copies, etc., in particular with `std::string`.


schweitzpgi:
Python wheel dependency tasks continue to fail for whatever reason.

amccaskey (approved):

schweitzpgi:
> > Python wheel dependency tasks continue to fail for whatever reason.
> 
> I believe you need to pickup #1471 and #1485 in your feature branch. A merge from `main` should be sufficient to fix it (I hope).

Thanks, Ben. I'll rebase this branch soon.

PR #1489, opened 2024-04-08
Use cc dialect.

schweitzpgi:
Port the intrinsics to cc dialect, which should have everything required.
Add external/private linkage control.


runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -49,35 +49,22 @@ static std::size_t regCounter = 0;
...
-    %0 = llvm.mlir.null : !llvm.ptr<struct<(f64,f64)>>
-    %1 = llvm.mlir.constant({} : i32) : i32
-    %2 = llvm.mlir.undef : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)>
-    %3 = llvm.insertvalue %0, %2[0] : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)> 
-    %4 = llvm.insertvalue %1, %3[1] : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)> 
-    llvm.return %4 : !llvm.struct<(ptr<struct<(f64,f64)>>, i32)>
-  }})#"),
-      hashValue, size);
+  return fmt::format(fmt::runtime(R"#(
+  cc.global extern @nvqpp.state.{} : !cc.struct<{!cc.ptr<!cc.struct<{f64,f64}>>, i32}>
```
amccaskey, line 53: I like this change a lot. I think though that it may be better to just remove this "intrinsic" and create the extern global programmatically, and then we can parameterize it easier on f32 or f64 (as in 1482) vs string replacement. what do you think?

schweitzpgi: Makes sense to me. Do you want me to do it in this PR?

amccaskey: Yea I think that would be good

amccaskey (approved):

PR #1492, opened 2024-04-09
python kernel builder qalloc support

amccaskey:
Implement `qalloc(state_vector)` support for the Python Kernel builder. 

Support patterns like 
```python
# Unknown state as function input 
kernel, initState = cudaq.make_kernel(list[complex])
qubits = kernel.qalloc(initState)
state = [.70710678, 0., 0., 0.70710678]
counts = cudaq.sample(kernel, state)

# Known state as qalloc input 
state = np.array([.70710678, 0., 0., 0.70710678], dtype=complex)
kernel = cudaq.make_kernel()
qubits = kernel.qalloc(state)
counts = cudaq.sample(kernel)
```

This requires a better handle at the Python level on the floating point precision of simulation state data. 
```python 
cudaq.set_target('nvidia-fp64')
state = np.array([.70710678, 0., 0., 0.70710678], dtype=complex)
kernel = cudaq.make_kernel()
# Works, Simulator is FP64 and so is state.dtype
qubits = kernel.qalloc(state)

# Raises an exception 
state = np.array([.70710678, 0., 0., 0.70710678], dtype=np.complex64)
kernel = cudaq.make_kernel()
# Fails, Simulator is FP64 and but state.dtype is FP32
qubits = kernel.qalloc(state)

# Provide convenience functions for user 
# code to work for any state vector target 
state = np.array([.70710678, 0., 0., 0.70710678],
                     dtype=cudaq.simulation_dtype()) # Note simulation_dtype() function
kernel = cudaq.make_kernel()
qubits = kernel.qalloc(state)
counts = cudaq.sample(kernel)

# Set the dtype automatically from any array data
state = cudaq.create_state([.70710678, 0., 0., 0.70710678])
kernel = cudaq.make_kernel()
qubits = kernel.qalloc(state)
counts = cudaq.sample(kernel)
```

Would be great to get reviewer thoughts on this approach to handling simulation data type mismatches.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -6,7 +6,7 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-import sys, os
+import sys, os, numpy
 from ._packages import *
 from .kernel.kernel_decorator import kernel, PyKernelDecorator
 from .kernel.kernel_builder import make_kernel, QuakeValue, PyKernel
@@ -85,6 +85,26 @@ def synthesize(kernel, *args):
     return PyKernelDecorator(None,
                              module=cudaq_runtime.synthesize(kernel, *args),
                              kernelName=kernel.name)
+
+
+def simulation_dtype():
+    """
+    Return the data type for the current simulation backend, 
+    either `complex128` or `complex64`.
+    """
+    target = get_target()
+    precision = target.get_precision()
+    if precision == cudaq_runtime.SimulationPrecision.fp64:
+        return complex
+    return numpy.complex64
+
+
+def create_state(array_data):
+    """
+    Create a state array with the appropriate data type for the 
+    current simulation backend target. 
+    """
+    return numpy.array(array_data, dtype=simulation_dtype())
 
 
 def __clearKernelRegistries():
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -90,7 +90,7 @@ def synthesize(kernel, *args):
 def simulation_dtype():
     """
     Return the data type for the current simulation backend, 
-    either `complex128` or `complex64`.
+    either `numpy.complex128` or `numpy.complex64`.
     """
     target = get_target()
     precision = target.get_precision()
--- a/python/cudaq/kernel/kernel_builder.py
+++ b/python/cudaq/kernel/kernel_builder.py
@@ -7,6 +7,7 @@
 # ============================================================================ #
 
 from functools import partialmethod
+import hashlib
 import random
 import re
 import string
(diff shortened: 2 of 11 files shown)


python/cudaq/__init__.py
```diff
@@ -87,6 +87,26 @@ def synthesize(kernel, *args):
                              kernelName=kernel.name)
 
 
+def simulation_dtype():
+    """
+    Return the data type for the current simulation backend, 
+    either `complex128` or `complex64`.
```
schweitzpgi, line 93: Are you really supporting quad precision? I suspect this is meant as `complex_fp64` and `complex_fp32` though. (Note that there are also things like complex<int> out there, so using the `fp` won't be a bad idea.

amccaskey: correct me if I'm wrong, but `complex128 == 2 * doubles (FP64)` and `complex64 == 2 floats (FP32)` right? Numpy seems to use this jargon for `complex_on_fp64` and `complex_on_fp32`

python/cudaq/kernel/kernel_builder.py
```diff
@@ -544,22 +554,114 @@ def qalloc(self, initializer=None):
...
+                    cc.GlobalOp(TypeAttr.get(globalTy),
+                                globalName,
+                                external=True)
+                    setStateFunc = func.FuncOp(setStateName,
+                                               FunctionType.get(
+                                                   inputs=[ptrComplex],
+                                                   results=[]),
+                                               loc=self.loc)
+                    entry = setStateFunc.add_entry_block()
+                    kDynamicPtrIndex: int = -2147483648
```
schweitzpgi, line 619: I guess there is no binding generated for this magic number?

python/cudaq/kernel/kernel_builder.py
```diff
@@ -544,22 +554,114 @@ def qalloc(self, initializer=None):
...
+                            cc.PointerType.get(self.ctx, ptrComplex), address,
+                            [zero, zero],
+                            DenseI32ArrayAttr.get(
+                                [kDynamicPtrIndex, kDynamicPtrIndex],
+                                context=self.ctx))
+                        cc.StoreOp(entry.arguments[0], ptr)
+                        func.ReturnOp([])
+
+                zero = self.getConstantInt(0)
+                numQubits = np.log2(size)
```
schweitzpgi, line 635: Do we test that size is a power of 2?

schweitzpgi (approved):

python/cudaq/kernel/kernel_builder.py
amccaskey: No, the MLIR bindings don't generate it

python/tests/builder/test_kernel_builder.py
```diff
@@ -898,44 +898,99 @@ def test_from_state():
...
+    # Now test constant array data, not kernel input
+    state = np.array([.70710678, 0., 0., 0.70710678], dtype=complex)
+    kernel = cudaq.make_kernel()
+    qubits = kernel.qalloc(state)
+    counts = cudaq.sample(kernel)
+    print(counts)
+    assert '11' in counts
+    assert '00' in counts
+
+    state = [.70710678 + 0j, 0., 0., 0.70710678]
```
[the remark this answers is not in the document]
amccaskey: Sure thing

PR #1493, opened 2024-04-10
Fix Python wheel installation issues (multiple subtlety linked items)

bmhowe23:
## Description

This PR contains a variety of fixes that all stemmed from the fact that the instructions provided on https://pypi.org/project/cuda-quantum/0.7.0/ (which originate from `python/README.md`) were no longer working after new versions of transitive dependencies were added to conda-forge, and that meant that the `tensornet` backend could no longer be used. The main point of this PR is to get the instructions in a working state again.

## Details

The full list of changes in this PR are:
* Place conda's library path before system library path in case multiple CUDA versions are installed on the host.
* ~~Pin the cuquantum conda-forge version~~ (this was the original solution for the PR, but there was a desire to remove this conda dependency and simply rely on the Python PIP package dependency that was already pinned).
* Remove cuquantum from the Python installation instructions. (e1e84aee205d2bf079a08c26ab7cfcc0d3de1fd8)
  * Since we're removing this from the install instructions, I had to create `tools/cudaq-qpud/cudaq-qpud.py` as a launch script.
  * Updated docs for the new launch script in 2cd4f29fe7e78c66889030f20a69c1d4faeab5ae.
* Fix a Python library initialization bug that was preventing the pre-existing library detection code from working. (91c9cbf9e370ad5e3cdf376907d7fa37efeb2022)
* Fix a lambda capture bug that was only happening with g++ 11.2.1. (246b1328375139dba81aecaa638d7db60c03fe14 and 7fb1dff4a1c1d680fe57e8a1a5ea7a84bf108a5e). This bug appears to have been due to a compiler bug w/ g++ 11.2.1 that we hadn't previously encountered.
* Fix a bug where the sample_async_remote.py example file was accidentally auto-launching `cudaq-qpud` rather than using the ones provided with the `--server` option (a18113d121576ead860d1473c05805545a87a0d5)
* Add a Publishing test to validate the Python wheel instructions every time we push a new PR into main
  * Link to successful run of new test: https://github.com/NVIDIA/cuda-quantum/actions/runs/8715669430/job/23908299709
    * The failed steps were due to artifacts/images expiring, but they are unrelated to the new test.


python/README.md
```diff
@@ -40,8 +40,8 @@ Quantum with all its dependencies:
 ```console
 conda create -y -n cuda-quantum python=3.10 pip
 conda install -y -n cuda-quantum -c "nvidia/label/cuda-11.8.0" cuda
-conda install -y -n cuda-quantum -c conda-forge mpi4py openmpi cxx-compiler
-conda env config vars set -n cuda-quantum LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$CONDA_PREFIX/envs/cuda-quantum/lib"
+conda install -y -n cuda-quantum -c conda-forge mpi4py openmpi cxx-compiler cuquantum~=23.10
```
amccaskey, line 43: forgive my ignorance because I am not very close to this problem and you have dug into it way more than me - but shouldn't we keep ourselves up to date with the latest from cuquantum?

bmhowe23: Depending on pending discussions with @bettinaheim and @1tnguyen, I think the `cuquantum` piece will eventually come out of this line. But note that our pyproject.toml file has us version locked to 23.10 as well. I think our problems were stemming from the fact that there was a discrepancy between our pyproject.toml file (which locked us to 23.10) and these instructions (which allowed the version to float).

python/README.md
```diff
@@ -40,8 +40,8 @@ Quantum with all its dependencies:
...
+conda env config vars set -n cuda-quantum LD_LIBRARY_PATH="$CONDA_PREFIX/envs/cuda-quantum/lib:$LD_LIBRARY_PATH"
```
bmhowe23, line 44: Note that regardless of the uncertainty about the above change, this change is important because if the user has multiple versions of CUDA installed on the host, this new change will ensure that the CUDA we just installed above (cuda-11.8.0) will take precedence.

python/README.md
bmhowe23: With my latest push to this PR, this change is reverted, so I'm going to close the issue. But we should consider updating to the latest 24.x version in 0.8. (The underlying issue was a bug in __init__.py.)

.github/workflows/publishing.yml
```diff
@@ -791,6 +791,46 @@ jobs:
...
+          apt-get update && apt-get install -y --no-install-recommends ca-certificates vim wget openssh-client
+          cd /tmp
+          # Setup links for validate_wheel.sh script
+          ln -s $GITHUB_WORKSPACE/scripts/validate_wheel.sh .
+          ln -s $GITHUB_WORKSPACE/docs/sphinx/examples/python /tmp/examples
+          ln -s $GITHUB_WORKSPACE/docs/sphinx/snippets/python /tmp/snippets
+          ln -s $GITHUB_WORKSPACE/python/tests /tmp/tests
+          ln -s $GITHUB_WORKSPACE/python/README.md .
+          # Run the script
+          # FIXME - should we pass in "-q" to make it a quick test?
```
bmhowe23, line 831: Need to address this FIXME

python/README.md
```diff
@@ -41,7 +41,7 @@ Quantum with all its dependencies:
 conda create -y -n cuda-quantum python=3.10 pip
 conda install -y -n cuda-quantum -c "nvidia/label/cuda-11.8.0" cuda
 conda install -y -n cuda-quantum -c conda-forge mpi4py openmpi cxx-compiler
-conda env config vars set -n cuda-quantum LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$CONDA_PREFIX/envs/cuda-quantum/lib"
+conda env config vars set -n cuda-quantum LD_LIBRARY_PATH="$CONDA_PREFIX/envs/cuda-quantum/lib:$LD_LIBRARY_PATH"
```
leofang, line 44: This is really nerve-wrecking. Conda has explicitly suggested users to **NOT** include conda paths in `LD_LIBRARY_PATH` (see, e.g., [here](https://docs.conda.io/projects/conda-build/en/stable/resources/use-shared-libraries.html#shared-libraries-in-macos-and-linux)). Given that NVIDIA is fully committed to make conda-forge a first-class citizen (ex: we attempt to do Day 1+N release on conda-forge for new CTKs, see e.g. [here](https://github.com/conda-forge/cuda-feedstock/issues/23)), may I nudge the team to put serious consideration to sticking to a safer practice for conda on your roadmap?

bettinaheim: One thing to point out here is that this only modifies LD_LIBRARY_PATH *within this conda environment*, not generally for the shell or session.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -146,16 +146,16 @@ class CircuitSimulator {
...
+      auto qId = qubitIds[qubitIdx];
       if (type != cudaq::pauli::I)
-        qubitSupport.push_back(qubitIds[qubitIdx]);
+        qubitSupport.push_back(qId);
 
       if (type == cudaq::pauli::Y)
-        basisChange.emplace_back([&, qubitIdx](bool reverse) {
-          rx(!reverse ? M_PI_2 : -M_PI_2, qubitIds[qubitIdx]);
+        basisChange.emplace_back([this, qId](bool reverse) {
+          rx(!reverse ? M_PI_2 : -M_PI_2, qId);
```
bmhowe23, line 155: 2 points to the person that explains to me -- without falling back on a hypothetical compiler bug explanation -- why the Python changes in this PR require this change. And only in gcc11/x86 builds when paired side-by-side with the installer....on some (but not all) operating systems. It worked fine in gcc11/ARM builds (all operating systems) in the "Side-by-side with installer" test, too, so it really was only gc11+x86+side-by-side-with-installer-on-some-operating-systems.

All other standard Docker tests and wheel tests pass with gcc12, clang16, and llvm bootstrapped builds (both before and after this change). 

In any case, if we need to make this change, I think we have similar code in `runtime/nvqir/cutensornet/simulator_mps_register.cpp`, so we should probably put it in that file as well.

I'm open to other ideas if people have them.

schweitzpgi: I didn't confirm this, but it looks a C++ coding bug.

1. There is the `for_each_pauli` iterator that applies a lambda to each element.
1. Each time, a new instance of the lambda is created with an argument `qubitIdx`.
1. When the type is X or Y, a second lambda is instantiated and pushed on the `basisChange` queue.
1. The second lambda captures a reference to `qubitIdx` from the outer lambda.
1. When the `for_each_pauli` ends, all the outer lambdas are no longer live, so the compiler could reclaim them.
1. `basisChange` is still live, still contains the inner lambdas, and the inner lambdas still hold references to the `qubitIdx` arguments that were in the now defunct and possibly reclaimed outer lambdas.

Passing the value of the element by-value to the inner lambda(s) looks like the correct fix then.

schweitzpgi: Trying again. The inner lambdas are causing a reference capture of the reference qubitIds argument. It seems like this should work fine (and it seems to with most of the compilers Ben has tested). But apparently, it is also known to potentially tickle some bugs. Here is a discussion: https://stackoverflow.com/questions/21443023/capturing-a-reference-by-reference-in-a-c11-lambda

scripts/validate_wheel.sh
```diff
@@ -141,6 +162,16 @@ for ex in `find "$root_folder/examples" -name '*.py' -not -path '*/providers/*'`
...
+# First determine the location of the cuquantum libraries (the [^, ] is to catch a suffix like "-cu11")
+CUQUANTUM_PACKAGE=$(python3 -m pip show --files cuda-quantum | grep Requires: | grep -o -P "cuquantum[^, ]*")
+CUQUANTUM_DEPS=$(python3 -m pip show --files $CUQUANTUM_PACKAGE | grep Requires: | cut -d' ' -f2- | tr -d ,)
+CUQUANTUM_LIB_PATH=$(for PACKAGE in $CUQUANTUM_DEPS; do
+  package_loc=$(python3 -m pip show --files $PACKAGE | grep Location: | cut -d' ' -f2)
+  package_so_dir=$(python3 -m pip show --files $PACKAGE | grep "\.so" | sed -re "s/ +//" | xargs dirname | head -1)
+  echo $package_loc/$package_so_dir
+done | sort | uniq | paste -d':')
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${CUQUANTUM_LIB_PATH}
+# Now continue with cudaq-qpud location
```
bmhowe23, line 174: @bettinaheim - I think this (i.e., updating validate_wheel.sh per these new lines) is the only way we can keep the `cuquantum` conda package out of the Python installation instructions. This means that pip package users would have to do something similar if they ever wanted to manually launch `cudaq-qpud` like we're doing in this test. Please let me know if you had something else in mind.

bettinaheim: I think we can clean this up a bit by including a launch script instead that automatically does the same thing as the __init__ file for the cudaq python package. We can make that script a python file, since I believe it would only be used by python users (other users don't need that script since there all necessary paths need to be set anyway).

bmhowe23: Just to confirm ... this Python-script approach won't work in the C++-only environment. Is that ok?

bmhowe23: Updated in 1a59a7251 and 2cd4f29fe.

bmhowe23: > Updated in [1a59a72](https://github.com/NVIDIA/cuda-quantum/commit/1a59a7251193517ab4627e6efc5315558390114c) and [2cd4f29](https://github.com/NVIDIA/cuda-quantum/commit/2cd4f29fe7e78c66889030f20a69c1d4faeab5ae).

Well, it looks like those didn't work. (Failed publishing run is here: https://github.com/NVIDIA/cuda-quantum/actions/runs/8697234380/job/23852549810). So I'll keep investigating.

docs/sphinx/releases.rst
```diff
@@ -19,7 +19,7 @@ Check out our `documentation <https://nvidia.github.io/cuda-quantum/latest/using
 to get started with the new Python syntax support we have added, and `follow our blog <https://developer.nvidia.com/cuda-q>`__
 to learn more about the new setup and its performance benefits.
 
-- `Docker image <https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda-quantum>`__
+- `Docker image <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nightly/containers/cuda-quantum/tags>`__ (use tag 0.7.0)
```
bmhowe23, line 22: @bettinaheim - do you want this change included in this PR, or is a new change forthcoming such that it should be backed out?

bmhowe23: This was backed and and covered in a different PR, so I'm resolving this conversation.

schweitzpgi:
Just checking in here. Is this a draft (per the Conda comments) or are you waiting for a review?

bmhowe23:
> Just checking in here. Is this a draft (per the Conda comments) or are you waiting for a review?

This is no longer just a draft. It is ready for review and is the primary PR required for 0.7.1. It needs code owner approval due to the changes to the .github directory. (@bettinaheim)

bettinaheim (approved):

PR #1495, opened 2024-04-10
[features/qalloc] Simplify the kernel_builder's use of vector data.

schweitzpgi:
Fixes a bug in the implementation where a vector captured as an initial state by the builder may not accurately reflect the content of the vector when the kernel is actually invoked. This captures a reference to the initialization vector instead of the data in the vector at the point of creating the kernel.

Simplifies the state vector cache to be a simple vector instead of a sorted map. Eliminates the use of hashes. Uses generic callbacks, specialized by type, to fetch the length and data pointer from the captured vector. Uses a type-safe variant for the two cases. Because this uses callbacks, the linker should fail if the kernel is not run in an environment that contains the runtime (such as on hardware).


amccaskey:
Rebase should fix your test errors

amccaskey (approved):
much nicer approach

PR #1499, opened 2024-04-11
Fix issue with unknown base on attribute call (throw an error)

### LINKED ISSUE #1496: Compiler doesn't raise an error when using kernel.h(q[0])
### Required prerequisites

### Describe the bug

In the code snippet below, the compiler doesn't raise an error for `kernel.h(q[0])`. Even though no error is raised, this operation doesn't seem to be get executed.

```
import cudaq

cudaq.set_target("nvidia")

@cudaq.kernel
def test_state(N: int):
    q = cudaq.qvector(N)
    kernel.h(q[0])

counts = cudaq.sample(test_state, 10)
counts.dump()
```

Output is: { 0000000000:1000 }

### Expected behavior

It should raise an error.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

amccaskey:
Fixes #1496 - issue with unknown attribute call expression 
```python 
import cudaq
@cudaq.kernel
def test_state(N: int):
    q = cudaq.qvector(N)
    h(q[0])
    kernel.h(q[0]) # user error
test_state.compile()
```

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1631,8 +1631,8 @@ class PyASTBridge(ast.NodeVisitor):
                 if quake.VeqType.isinstance(var.type):
                     if node.func.attr == 'size':
                         # Handled already in the Attribute visit
-                        return 
-                    
+                        return
+
                     # `qreg` or `qview` method call
                     if node.func.attr == 'back':
                         qrSize = quake.VeqSizeOp(self.getIntegerType(),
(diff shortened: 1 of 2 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -1804,6 +1808,9 @@ def bodyBuilder(iterVal):
                     f'Unknown attribute on quantum operation {node.func.value.id} ({node.func.attr}). {maybeProposeOpAttrFix(node.func.value.id, node.func.attr)}'
                 )
 
+            self.emitFatalError(
+                f"Invalid function call - '{node.func.value.id}' is unknown.")
```
bmhowe23, line 1812: Would it be possible to add a test to see this in action?

amccaskey: Sure thing. Test added.

schweitzpgi (commented):
LGTM

bmhowe23 (approved):
👍 thanks, Alex!

PR #1500, opened 2024-04-11
Add range(size_t) for common programmer patterns

amccaskey:


runtime/cudaq/utils/cudaq_utils.h
```diff
@@ -278,6 +278,15 @@ inline std::vector<ElementType> range(ElementType N) {
   return range(ElementType(0), N);
 }
 
+/// @brief Return a vector of unsigned integers. The first element is zero, and
+/// the remaining elements are all values incremented by 1 to the total
+/// size value provided (exclusive).
+inline std::vector<std::size_t> range(std::size_t N) {
+  std::vector<std::size_t> vec(N);
```
schweitzpgi, line 285: We should probably add a constraint that `N <= LLONG_MAX` to be consistently paranoid.

schweitzpgi: Add: the portable spelling being the "obvious" `std::numeric_limits<std::make_signed_t<std::size_t>>::max`

amccaskey: Check the latest and see if that works for you

PR #1502, opened 2024-04-11
Allow measurement of variadic list of qubits

### LINKED ISSUE #1498: Output discrepancy when args of mz are in a list vs when comma separated
### Required prerequisites

### Describe the bug

If the qubits to be measured are specified in a list, that is, mz(  [  q[], ..,q[]  ]  ), the correct number of digits in the bitstrings is output upon sampling. But if you don't specify them as a list but simply as comma separated args, no matter how many args there are, only a single digit bitstring is output. 

```
import cudaq

@cudaq.kernel
def my_kernel():
    q = cudaq.qvector(5)
    x(q[2])
    #mz([q[0], q[2]]) #2-digit bitstring
    mz(q[0], q[1], q[2]) # 1-digit bitstring

result = cudaq.sample(my_kernel)
print(result)
```

### Expected behavior

To address this issue, we can go down either of these routes:

1. Raise an error when qubits are provided in a comma-separated format, prompting the user to input qubits in a list format for consistency.
or 
2. Make modification to ensure that the output format for comma-separated qubits is consistent with the list format.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

amccaskey:
Fixes #1498

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1337,17 +1337,18 @@ class PyASTBridge(ast.NodeVisitor):
                                 "measurement register_name keyword must be a constant string literal.",
                                 node)
                         registerName = userProvidedRegName.value.value
-                qubit = self.popValue()
-                self.checkControlAndTargetTypes([], [qubit])
+                qubits = [self.popValue() for _ in range(len(self.valueStack))]
+                self.checkControlAndTargetTypes([], qubits)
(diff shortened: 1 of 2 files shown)


bmhowe23 (approved):
Thanks, Alex. FWIW, I also double checked that the C++ interface behaves similarly (i.e. `mz(q[0], q[1])` behaves similar to your test).

PR #1503, opened 2024-04-11
[features/qalloc] Documentation.

schweitzpgi:
Add an example using kernel builder's qalloc with a vector initializer.

The "real" change here is to docs/sphinx/examples/cpp/other/builder/builder.cpp.

There is other churn relating to getting the example to compile and work with CI testing cleanly.


docs/sphinx/examples/cpp/other/builder/builder.cpp
```diff
@@ -114,6 +122,37 @@ int main() {
...
 
+  {
+    // In a simulated environment, it is sometimes useful to be able to
+    // specify an initial state vector. The initial state vector is 2 to the
+    // power `n` where `n` is the number of qubits.
+
+    // In this example, we create a kernel template `sim_kernel` that captures
+    // the variable `init_state` by reference.
+    auto sim_builder = cudaq::make_kernel();
+    std::vector<std::complex<double>> init_state;
```
amccaskey, line 133: we should use `std::vector<cudaq::simulation_scalar>>` instead so this code will work for any of the simulators (wrt their floating point precision)

schweitzpgi:
@amccaskey Python tests are failing with unknown target "nvidia-fp64". That is clearly a .config file however.

docs/sphinx/examples/cpp/other/builder/builder.cpp
```diff
@@ -114,6 +122,37 @@ int main() {
...
+    // In this example, we create a kernel template `sim_kernel` that captures
+    // the variable `init_state` by reference.
+    auto sim_builder = cudaq::make_kernel();
+    std::vector<cudaq::simulation_scalar> init_state;
+    auto q = sim_builder.qalloc(init_state);
+    // Build the quantum circuit template here.
+    sim_builder.mz(q);
+
+    // Now we are ready to instantiate the kernel and invoke it. So we can set
+    // the `init_state` vector to a vector with 2 complex values (1 qubit) and
```
anthony-santana, line 139: nit
```suggestion
    // the `init_state` to a vector with 2 complex values (1 qubit) and
```

docs/sphinx/examples/cpp/other/builder/builder.cpp
```diff
@@ -114,6 +122,37 @@ int main() {
...
+  {
+    // In a simulated environment, it is sometimes useful to be able to
+    // specify an initial state vector. The initial state vector is 2 to the
+    // power `n` where `n` is the number of qubits.
+
+    // In this example, we create a kernel template `sim_kernel` that captures
+    // the variable `init_state` by reference.
+    auto sim_builder = cudaq::make_kernel();
+    std::vector<cudaq::simulation_scalar> init_state;
+    auto q = sim_builder.qalloc(init_state);
```
anthony-santana, line 134: nit: I think it would be worth spelling out this variable name to implicitly remind people that the return type here is still qubit/s.

docs/sphinx/examples/cpp/other/builder/builder.cpp
```diff
@@ -17,6 +17,14 @@
 #include <cudaq/gradients.h>
 #include <cudaq/optimizers.h>
 
+static bool results_are_close(const cudaq::sample_result &f1,
```
anthony-santana, line 20: Am I missing something or is this returning true even if f1 =/= f2 ?

anthony-santana (approved):

docs/sphinx/examples/cpp/other/builder/builder.cpp
schweitzpgi: This is completely fake. The kernel has no quantum code and hence does nothing. So this stub doesn't actually compare anything. As a stub, it just rubber stamps the results.

docs/sphinx/examples/cpp/other/builder/builder.cpp
schweitzpgi: This is following the convention used in the tests in this file. I agree that it could be improved. I'll file a PR.

PR #1504, opened 2024-04-11
Expose the U3 operation to frontends

khalatepradnya:
### Description
Partially addresses issue https://github.com/NVIDIA/cuda-quantum/issues/1507

### DIFF
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -1582,6 +1582,13 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
                                          reportNegateError, isAdjoint,
                                          /*control=*/true);
 
+    if (funcName.equals("u3")) {
+      const auto size = args.size();
+      assert(size == 4);
+      SmallVector<Value> params(args.begin(), args.end() - 1);
+      builder.create<quake::U3Op>(loc, params, args.back());
+    }
+
     if (funcName.equals("control")) {
       // Expect the first argument to be an instance of a Callable. Need to
       // construct the name of the operator() call to make here.
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -127,12 +127,13 @@ template <typename A, typename P = void>
 bool buildOp(OpBuilder &builder, Location loc, ValueRange operands,
              SmallVector<Value> &negations,
              llvm::function_ref<void()> reportNegateError,
-             bool isAdjoint = false, bool isControl = false) {
+             bool isAdjoint = false, bool isControl = false,
+             size_t paramCount = 1) {
   if constexpr (std::is_same_v<P, Param>) {
     assert(operands.size() >= 2 && "must be at least 2 operands");
-    auto params = operands.take_front();
-    auto [target, ctrls] =
-        maybeUnpackOperands(builder, loc, operands.drop_front(1), isControl);
+    auto params = operands.take_front(paramCount);
+    auto [target, ctrls] = maybeUnpackOperands(
+        builder, loc, operands.drop_front(paramCount), isControl);
     for (auto v : target)
       if (std::find(negations.begin(), negations.end(), v) != negations.end())
         reportNegateError();
@@ -1582,12 +1583,10 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
                                          reportNegateError, isAdjoint,
                                          /*control=*/true);
 
-    if (funcName.equals("u3")) {
-      const auto size = args.size();
-      assert(size == 4);
-      SmallVector<Value> params(args.begin(), args.end() - 1);
-      builder.create<quake::U3Op>(loc, params, args.back());
-    }
+    if (funcName.equals("u3"))
+      return buildOp<quake::U3Op, Param>(builder, loc, args, negations,
+                                         reportNegateError, isAdjoint,
+                                         isControl, /*paramCount=*/3);
 
     if (funcName.equals("control")) {
       // Expect the first argument to be an instance of a Callable. Need to
--- a/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
+++ b/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
@@ -895,7 +895,8 @@ void quake::U3Op::getOperatorMatrix(Matrix &matrix) {
     phi *= -1;
     lambda *= -1;
   }
-
+  // ASKME: Is this column-major representation? Or is that in
+  // `runtime/nvqir/NVQIR.cpp` - `__quantum__qis__u3` wrong?
   matrix.assign({std::cos(theta / 2.),
                  std::exp(phi * 1i) * std::sin(theta / 2.),
                  -std::exp(lambda * 1i) * std::sin(theta / 2.),
--- a/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
+++ b/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
@@ -895,6 +895,7 @@ void quake::U3Op::getOperatorMatrix(Matrix &matrix) {
     phi *= -1;
     lambda *= -1;
   }
+
   matrix.assign({std::cos(theta / 2.),
                  std::exp(phi * 1i) * std::sin(theta / 2.),
                  -std::exp(lambda * 1i) * std::sin(theta / 2.),
--- a/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
+++ b/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
@@ -895,8 +895,6 @@ void quake::U3Op::getOperatorMatrix(Matrix &matrix) {
     phi *= -1;
     lambda *= -1;
   }
-  // ASKME: Is this column-major representation? Or is that in
-  // `runtime/nvqir/NVQIR.cpp` - `__quantum__qis__u3` wrong?
   matrix.assign({std::cos(theta / 2.),
                  std::exp(phi * 1i) * std::sin(theta / 2.),
                  -std::exp(lambda * 1i) * std::sin(theta / 2.),
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1373,10 +1373,11 @@ class PyASTBridge(ast.NodeVisitor):
                 params.reverse()
                 for idx, val in enumerate(params):
                     if IntegerType.isinstance(val.type):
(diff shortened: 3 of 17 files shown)


lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1592,6 +1592,13 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
                                          reportNegateError, isAdjoint,
                                          /*control=*/true);
 
+    if (funcName.equals("u3")) {
+      const auto size = args.size();
+      assert(size == 4);
```
schweitzpgi, line 1597: Can you add a comment here? Something like `&& "The u3 operator must have 4 arguments"` would be great.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1364,6 +1364,19 @@ def bodyBuilder(iterVal):
                 opCtor([], [], [], [qubitA, qubitB])
                 return
 
+            if node.func.id == 'u3':
+                # Single target, three parameters `u3(θ,φ,λ)`
+                numValues = len(self.valueStack)
+                target = self.popValue()
+                params = [self.popValue() for _ in range(numValues - 1)]
+                params.reverse()
```
schweitzpgi, line 1372: Not a big deal, but we expect exactly 3 angles here so it would be straightforward to create the params without the `reverse()`.

```python
lambd4 = self.popValue()
phi = self.popValue()
theta = self.popValue()
params = [theta, phi, lambd4]
```

I assume there is no way to pop a vector of values in natural order like in the C++ bridge.

```c++
std::vector<Value> args = popValues(3);
```

python/tests/builder/test_kernel_builder.py
```diff
@@ -1243,6 +1244,16 @@ def test_list_subscript():
     cudaq.sample(kernelAndArgs[0], False, [], [], [])
 
 
+def test_u3_op():
+    kernel = cudaq.make_kernel()
+    q = kernel.qalloc(1)
+    # TODO: Allow use of only `q`
```
schweitzpgi, line 1250: Does this mean you want python to accept a list of qubits in the final position? I assume that would be checked such that the list must contain exactly one qubit or it is ill-formed.

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -36,6 +36,7 @@ namespace qubit_op {
 ConcreteQubitOp(h) ConcreteQubitOp(x) ConcreteQubitOp(y) ConcreteQubitOp(z)
     ConcreteQubitOp(s) ConcreteQubitOp(t) ConcreteQubitOp(rx)
         ConcreteQubitOp(ry) ConcreteQubitOp(rz) ConcreteQubitOp(r1)
+            ConcreteQubitOp(u3)
```
schweitzpgi, line 39: Should we be adding `u2` here while we're at it?

runtime/nvqir/Gates.h
```diff
@@ -114,6 +114,7 @@ getGateByName(GateName name, const std::vector<Scalar> angles = {}) {
     auto theta = angles[0];
     auto phi = angles[1];
     auto lambda = angles[2];
+    // TODO: Confirm matrix representation
```
schweitzpgi, line 117: @boschmitt

runtime/nvqir/NVQIR.cpp
```diff
@@ -326,6 +326,16 @@ void __quantum__qis__phased_rx(double theta, double phi, Qubit *q) {
   nvqir::getCircuitSimulatorInternal()->applyCustomOperation(matrix, {}, {qI});
 }
 
+void __quantum__qis__u3(double theta, double phi, double lambda, Qubit *q) {
+  auto qI = qubitToSizeT(q);
+  std::complex<double> i(0, 1.);
+  std::vector<std::complex<double>> matrix{
+      std::cos(theta / 2.), -std::exp(i * lambda) * std::sin(theta / 2.),
+      std::exp(i * phi) * std::sin(theta / 2),
+      std::exp(i * (lambda + phi)) * std::cos(theta / 2.)};
```
schweitzpgi, line 335: We should find a way of specifying this formula in a single place. It keeps repeating over and over.

schweitzpgi (commented):
Looks great. Just a few questions.

runtime/nvqir/NVQIR.cpp
1tnguyen: In `runtime/nvqir/Gates.h`, we have the `getGateByName` support for U3.

lib/Frontend/nvqpp/ConvertExpr.cpp
amccaskey: Also should we not be leveraging `isControl` and/or `negations` and/or `isAdjoint`?

python/tests/kernel/test_kernel_features.py
```diff
@@ -1278,4 +1278,25 @@ def test_state(N: int):
...
     with pytest.raises(RuntimeError) as e:
         test_state.compile()
-    assert "Invalid function call - 'kernel' is unknown." in repr(e)
\ No newline at end of file
+    assert "Invalid function call - 'kernel' is unknown." in repr(e)
+
+
+def test_u3_op():
+    @cudaq.kernel
+    def check_x():
```
amccaskey, line 1286: would be good to add a test that takes the 3 angles as input to the kernel

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -559,6 +559,16 @@ CUDAQ_ONE_QUBIT_PARAM_IMPL(ry, RyOp)
...
 CUDAQ_ONE_QUBIT_PARAM_IMPL(r1, R1Op)
 
+void u3(ImplicitLocOpBuilder &builder, std::vector<QuakeValue> &parameters,
+        QuakeValue &target) {
+  cudaq::info("kernel_builder apply u3");
+  std::vector<Value> parameterValues;
+  std::transform(parameters.begin(), parameters.end(),
+                 std::back_inserter(parameterValues),
+                 [](auto &el) { return el.getValue(); });
+  builder.create<quake::U3Op>(parameterValues, target.getValue());
```
amccaskey, line 569: We'll need the control version too.

schweitzpgi: Do we want/need a `cu3` for C++17?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -895,7 +895,8 @@ void quake::U3Op::getOperatorMatrix(Matrix &matrix) {
     phi *= -1;
     lambda *= -1;
   }
-
+  // ASKME: Is this column-major representation? Or is that in
+  // `runtime/nvqir/NVQIR.cpp` - `__quantum__qis__u3` wrong?
```
boschmitt, line 899: This is indeed a column-major representation, while the one in `runtime/nvqir/NVQIR.cpp` is a row-major one. Matrices must be in row-major! (Hence this one, is wrong.)

runtime/nvqir/Gates.h
boschmitt: It's wrong, this matrix must be in row-major representation.

runtime/nvqir/NVQIR.cpp
boschmitt: +1

boschmitt (commented):
For historical reasons the expected matrix representation changed from column-major to row-major. (It has to do with different backends expecting different forms by default.) In any case, for most operators, this changes nothing but in the case of `u3` the representation is relevant.

As we see in this PR, we have `u3` matrix definition in various places and with different representations! We need to find a way to centralize operators matrix definitions and make sure they are all in row-major form.

runtime/cudaq/qis/qubit_qis.h
khalatepradnya: No, I don't plan to. U2 can be implemented with U3 if required.

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
khalatepradnya: Looking at other operations in this file, it seems to me that they all are column-major, for e.g. https://github.com/NVIDIA/cuda-quantum/blob/main/lib/Optimizer/Dialect/Quake/QuakeOps.cpp#L909

python/tests/builder/test_kernel_builder.py
khalatepradnya: I changed the comment - the current behavior requires `q[0]`, asking if we should / need to support just using `q` and if `q` has multiple qubits, the operation is applied to all

runtime/nvqir/NVQIR.cpp
khalatepradnya: Will address separately...

python/tests/builder/test_kernel_builder.py
khalatepradnya: Fixed

python/tests/kernel/test_kernel_features.py
```diff
@@ -1310,3 +1316,93 @@ def test():
...
+
+
+def test_u3_adj():
+
+    @cudaq.kernel
+    def rotation_adjoint_test():
+        q = cudaq.qubit()
+
+        # # implement Rx gate with U3
+        # u3(1.1, -np.pi / 2, np.pi / 2, q)
```
khalatepradnya, line 1380: @boschmitt : Can you please check the Rx representation?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1365,6 +1365,22 @@ def bodyBuilder(iterVal):
                 opCtor([], [], [], [qubitA, qubitB])
                 return
 
+            if node.func.id == 'u3':
+                # Single target, three parameters `u3(θ,φ,λ)`
```
schweitzpgi, line 1369: Aside: at some point we should decide as a group how to write these angles. Code has gone from greek letters, to a mix of greek letters and English spellings of greek letters, to fully spelling them out in English (e.g. "angle_one"). I don't have a preference, but I don't think we need churn for the sake of churn.

schweitzpgi (approved):
Approving, contingent on Bruno double-checking that the matrices are the correct encodings.

python/tests/kernel/test_kernel_features.py
khalatepradnya: Fixed in [20c9b87](https://github.com/NVIDIA/cuda-quantum/pull/1504/commits/20c9b87086e68f18af7438356ad22a71f70502b7)

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Tried to address in [d013595](https://github.com/NVIDIA/cuda-quantum/pull/1504/commits/d013595eacb390d5e6eb6b8044a1c7a1c325d7c3)
Did not change the comments within the code.

PR #1505, opened 2024-04-12
[docs] Add cudaq::range function.

### LINKED ISSUE #1483: cudaq::range(start,stop,step) added to API
A new function was added to the cudaq API to generate monotonic sequences, either increasing or decreasing. This function generates a sequence of `n` integral values in the set `{ start, start+step, start+2*step, ... start+(n-1)*step }` such that `start+n*step` is equal to or beyond `stop`. That is, the semi-open interval is defined as either `[start, stop)` or `(stop, start]` depending on whether `step` is positive or negative, resp.

The implementation support is in #1476 .

- [ ] Add to documentation.

schweitzpgi:
Closes #1483

### DIFF
--- a/docs/sphinx/api/languages/cpp_api.rst
+++ b/docs/sphinx/api/languages/cpp_api.rst
@@ -61,6 +61,9 @@ Common
 
 .. doxygenclass:: cudaq::Trace
 
+.. doxygenfunction:: cudaq::range(ElementType total)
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):

schweitzpgi:
@1tnguyen Any ideas on why the Remote-Sim test is failing? This change has literally nothing to do with that test.

PR #1508, opened 2024-04-12
[features/qalloc] Not all targets are always available.

schweitzpgi:
The nvidia and nvidia-fp64 targets may not be built or installed depending on the configuration, so they cannot be relied upon to always be present. This adds code to catch the exception throw when looking up these targets fails.


python/tests/builder/test_kernel_builder.py
```diff
@@ -879,120 +879,134 @@ def test_recursive_calls():
 
 def test_from_state():
```
amccaskey, line 880: I think there's a better Pytest way to do this. @khalatepradnya might be able to point you in the right direction on how to create a Pytest decorator that does a conditional check before enabling the test

schweitzpgi: Updated [66eb983](https://github.com/NVIDIA/cuda-quantum/pull/1508/commits/66eb983fe19a0557074f94a5a7d9b35255d464cd)

khalatepradnya (approved):

python/tests/builder/test_kernel_builder.py
```diff
@@ -876,8 +876,20 @@ def test_recursive_calls():
...
 
-
-def test_from_state():
+def can_set_target(name):
+    target_installed = True
+    try:
+        cudaq.set_target(name)
+    except RuntimeError:
+        target_installed = False
+    target_installed
```
amccaskey, line 885: do we need a `return` here?

PR #576, opened 2023-08-23
Enable code coverage pipeline with codecov

actione:
### Description

Target 1: Sync and display code coverage in PRs
This is the result of the tests conducted in my test PR after forking the project:
[ https://github.com/actione/cuda-quantum/pull/2 ](https://github.com/actione/cuda-quantum/pull/2 )
Target 2: Track branch coverage changes
[https://app.codecov.io/github/actione/cuda-quantum](https://app.codecov.io/github/actione/cuda-quantum)

Things to do to enable this feature:
-  Install the Codecov app for the cuda-quantum project. You can find it at: 
   https://github.com/apps/codecov

PR #1510, opened 2024-04-15
Provide a script to generate code coverage

actione:
### Description

We previously implemented diff branch code coverage Per-PR in PR #576, but due to the installation limitations of code-cov, we temporarily abandoned this plan.
A script is provided here to facilitate generating code coverage locally and on automated systems. Later I will enable this to the internal automation system so that code coverage can be generated regularly

### DIFF
--- a/scripts/generate_cc.sh
+++ b/scripts/generate_cc.sh
@@ -11,9 +11,9 @@
 # Usage:
 # bash scripts/generate_cc.sh -c
 # -or-
-# bash scripts/build_cudaq.sh -p
+# bash scripts/generate_cc.sh -p
 # -or-
-# bash scripts/build_cudaq.sh -c -p
+# bash scripts/generate_cc.sh -c -p
 # -c flag generates coverage information for C and C++ codes.
 # -p flag generates coverage information for Python codes.
 # C and C++ coverage reports are generated in the directory 'build/ccoverage'
--- a/scripts/generate_cc.sh
+++ b/scripts/generate_cc.sh
@@ -1,0 +1,145 @@
+#!/bin/bash
+
+# ============================================================================ #
+# Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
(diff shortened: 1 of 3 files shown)


scripts/generate_cc.sh
```diff
@@ -0,0 +1,145 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# Usage:
+# bash scripts/generate_cc.sh -c
+# -or-
+# bash scripts/build_cudaq.sh -p
+# -or-
+# bash scripts/build_cudaq.sh -c -p
```
bmhowe23, line 16: Perhaps a copy/paste error?
```suggestion
# bash scripts/generate_cc.sh -p
# -or-
# bash scripts/generate_cc.sh -c -p
```

bmhowe23 (commented):
This will be nice to have ... thanks @actione.

scripts/generate_cc.sh
actione: Thanks, adding the -c and -p options at the same time means generating both C code coverage and python code coverage.

actione:
Hi, @bmhowe23 , I think the arm64 caching error should be unrelated with the modification of this PR. We only added two llvm-cov components and their sizes are very small. I rebase this branch. Could you help retrigger CICD? thanks

scripts/generate_cc.sh
bmhowe23: > Thanks, adding the -c and -p options at the same time means generating both C code coverage and python code coverage.

I was meaning that I wonder if the references to `build_cudaq.sh` are correct. Shouldn't they be referencing `generate_cc.sh`?

actione: Sorry, I didn't catch your meaning correctly :)

actione:
Hi, @bmhowe23 , If there are no further actions or concerns, could you please help merge this PR so that I can generate code coverage regularly on the internal automation platform ASAP :).

PR #1517, opened 2024-04-15
Creating a nvqc regression test to test nvqc image 0.7.0 against the cuda_quantum client image 0.7.1 or latest

sacpis:
Creating a nvqc regression test to test nvqc image 0.7.0 against the cuda_quantum client image 0.7.1 or latest

Here is the test result for the Nvqc regression test action
https://github.com/sacpis/cuda-quantum/actions/runs/8696750631

### DIFF
--- a/python/cudaq/kernel/analysis.py
+++ b/python/cudaq/kernel/analysis.py
@@ -38,24 +38,45 @@ class MidCircuitMeasurementAnalyzer(ast.NodeVisitor):
         ]:
             self.measureResultsVars.append(target.id)
 
+    # Get the variable name from a variable node.
+    # Returns an empty string if not something we know how to get a variable name from.
+    def getVariableName(self, node):
+        if isinstance(node, ast.Name):
+            return node.id
+        if isinstance(node, ast.Subscript):
+            return self.getVariableName(node.value)
+        return ''
+
     def visit_If(self, node):
         condition = node.test
         # catch `if mz(q)`
         if self.isMeasureCallOp(condition):
             self.hasMidCircuitMeasures = True
             return
-
-        # Catch if val, where `val = mz(q)`
-        if 'id' in condition.__dict__ and condition.id in self.measureResultsVars:
+        # Catch `if val`, where `val = mz(q)` or `if var[k]`, where `var = mz(qvec)`
+        if self.getVariableName(condition) in self.measureResultsVars:
             self.hasMidCircuitMeasures = True
-        # Catch `if UnaryOp mz(q)`
+        elif isinstance(condition, ast.Compare):
+            # Compare: look at left expression.
+            # Handle: `if var == True/False` and `if var[k] == True/False:`
+            if self.getVariableName(condition.left) in self.measureResultsVars:
+                self.hasMidCircuitMeasures = True
+            elif self.isMeasureCallOp(condition.left):
+                # Handle: `if mz(q) == True/False`
+                self.hasMidCircuitMeasures = True
+        # Catch `if UnaryOp mz(q)` or `if UnaryOp var` (`var = mz(q)`)
         elif isinstance(condition, ast.UnaryOp):
-            self.hasMidCircuitMeasures = self.isMeasureCallOp(condition.operand)
-        # Catch `if something BoolOp mz(q)`
+            self.hasMidCircuitMeasures = self.isMeasureCallOp(
+                condition.operand) or (self.getVariableName(condition.operand)
+                                       in self.measureResultsVars)
+        # Catch `if something BoolOp mz(q)` or `something BoolOp var` (`var = mz(q)`)
         elif isinstance(condition,
                         ast.BoolOp) and 'values' in condition.__dict__:
             for node in condition.__dict__['values']:
                 if self.isMeasureCallOp(node):
+                    self.hasMidCircuitMeasures = True
+                    break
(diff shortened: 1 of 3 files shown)


.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+  cancel-in-progress: false
+
+# Run on request and every day at 3 AM UTC
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: 'Target (choose nightly to run like nightly tests)'
+        required: true
+        default: 'nightly'
```
bmhowe23, line 15: I'm wondering if this `target` input parameter applies to this new test. We might be able to do without it.

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+
+  schedule:
+    - cron: 0 3 * * *
+
+env:
+  # NGC nv-quantum organization: pnyjrcojiblh
+  NGC_QUANTUM_ORG: pnyjrcojiblh
+  NGC_QUANTUM_TEAM: cuda-quantum
+  # <Backend>:<GPU Type>:<Instance Type>:<Min Instances>:<Max Instances>
+  NGC_NVQC_DEPLOYMENT_SPEC: NVCFInternal:A100:Standard_ND96amsr_A100_v4_1x:1:1
```
bmhowe23, line 48: I believe all of these lines can be deleted for this new workflow.
```suggestion
```

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+  # If vars below are changed, it is recommended to also update the
+  # workflow_dispatch defaults above so they stay in sync.
+  cudaq_test_image: nvcr.io/nvidia/nightly/cuda-quantum:0.7.1
```
bmhowe23, line 51: This value doesn't seem to be in sync with parameter default value above.
```suggestion
  cudaq_test_image: nvcr.io/nvidia/nightly/cuda-quantum:latest
```

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+  cudaq_nvqc_deploy_image: nvcr.io/nvidia/nightly/cuda-quantum:latest
```
bmhowe23, line 52: I suspect this isn't needed for this new workflow.
```suggestion
```

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+          if [ -n "${{ inputs.commit_sha }}" ]; then
+            echo "sha=${{ inputs.commit_sha }}" >> $GITHUB_OUTPUT
+          else
+            echo "sha=$(cat $CUDA_QUANTUM_PATH/build_info.txt | grep -o 'source-sha: \S*' | cut -d ' ' -f 2)" >> $GITHUB_OUTPUT
+          fi
+
+  nvqc_integration_docker_test:
+    name: NVQC integration test using Docker image
+    runs-on: ubuntu-latest
+    if: (inputs.target == 'nvqc' || github.event_name == 'schedule' || inputs.target == 'nightly')
```
bmhowe23, line 92: I think this line can go away.
```suggestion
```

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+            exit 1
+          fi
+        shell: bash
+
+  nvqc_integration_wheel_test:
+    name: NVQC integration test using Python wheels
+    runs-on: ubuntu-latest
+    needs: [metadata]
+    environment: backend-validation
+    if: inputs.target == 'nvqc' || github.event_name == 'schedule' || inputs.target == 'nightly'
```
bmhowe23, line 245: I suspect this line can be deleted.
```suggestion
```

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -0,0 +1,351 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Test NVQC
+        if: ${{ ! steps.install_wheel.skipped }}
+        run: |
+          echo "### Submit to NVQC from Python wheels" >> $GITHUB_STEP_SUMMARY
+          python_version=${{ inputs.python_version || env.python_version }}
+          export NVQC_API_KEY="${{ secrets.NVQC_SERVICE_KEY }}"
+          export NVQC_NCA_ID="${{ secrets.NVQC_NCA_ID }}"
```
bmhowe23, line 315: I think this may need adjustment for this repo (as opposed to your repo). We should probably discuss this with @bettinaheim. If we are using the production NCA ID, then we don't need to specify the `NVQC_NCA_ID`.
```suggestion
          export NVQC_API_KEY="${{ secrets.NVQC_PROD_SERVICE_KEY }}"
```

bmhowe23 (commented):
Thanks, @sacpis ... looking good.

.github/workflows/nvqc_regression_tests.yml
bettinaheim: Yes, having this workflow submit to the production environment is the intent, so Ben's change looks good. I am not sure what you mean by the "env variables for PROD"?

Issue #1454, opened 2024-03-27
Support complex values passed to numpy functions in python


anthony-santana:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

We should support complex argument values to the appropriate numpy functions. This includes the trig functions:

```
@cudaq.kernel
def kernel():
   # This works:
   value_1 = np.sin(1.0)
   # This doesn't work:
   value_2 = np.sin(1.0j)
```

Issue #1086, opened 2024-01-11
[RFC] [Language] Quantum allocation with state initialization


amccaskey:
I propose we update the language to support quantum allocation with user-provided initial state specification. This should supersede functions like `from_state(...)` on the `kernel_builder`. 

The following snippet demonstrates what this might look like: 
```cpp
__qpu__ auto test0() {
  // Init from state vector
  cudaq::qubit q = {0., 1.};
  return mz(q);
}

__qpu__ auto test1() {
  // Init from predefined state vectors
  cudaq::qubit q = cudaq::ket::one;
  return mz(q);
}

__qpu__ void test2() { 
  // Init from state vector
  cudaq::qubit q = {M_SQRT1_2, M_SQRT1_2}; 
}

__qpu__ void test3() { 
  // Init from state vector 
  cudaq::qvector q = {M_SQRT1_2, 0., 0., M_SQRT1_2}; 
}

__qpu__ void test4(const std::vector<cudaq::complex> &state) {
  // State vector from host 
  cudaq::qvector q = state;
}

void useBuilder() {
  std::vector<cudaq::complex> state{M_SQRT1_2, 0., 0., M_SQRT1_2}; 

  {
    // (deferred) qubit allocation from concrete state vector
    auto kernel = cudaq::make_kernel();
    auto qubitsInitialized = kernel.qalloc(state);
  }
  {
    // kernel parameterized on input state data
    auto [kernel, inState] = cudaq::make_kernel<std::vector<cudaq::complex>>();
    auto qubitsInitialized = kernel.qalloc(inState); 
   
    cudaq::sample(kernel, state).dump();
  }
}
```

The language level allocation initialization can be achieved via the addition of new constructors to `qudit.h` and `qvector.h`
```cpp
qudit(const std::vector<complex>&);
qudit(const std::initializer_list<complex>&);

qvector(const std::vector<complex>&);
qvector(const std::initializer_list<complex>&);
```

boschmitt:
Thanks @amccaskey for proposing this. 

I have a clarification question regarding the semantics of a `cudaq::qvector`'s state. For example, what will be the return value of the following kernel?
```cpp
__qpu__ bool foo() { 
  // Init from state vector 
  cudaq::qvector q = {0., 1., 0., 0.};
  return cudaq::mz(q[0]);
}
```
I see two possibilities:
* It returns `true`.
[...]

PR #1518, opened 2024-04-15
Support using complex numbers and lists of complex numbers inside kernel functions

annagrin:
Make sure we can use complex numbers inside kernel function definitions, for example:

```python
  c = [.70710678 + 0j, 0., 0., 0.70710678]
  
  # Pass a list of complex numbers as a parameter
  @cudaq.kernel
  def test_complex_vec_param(vec : list[complex]):
      c1 = vec

  counts = cudaq.sample(test_complex_vec_param, c)
```

```python
  # Capture a list of complex numbers
  @cudaq.kernel
  def test_complex_vec_capture():
      c1 = c

   counts = cudaq.sample(test_complex_vec_capture)
```

```python
  # Define a list of complex numbers inside a kernel
  @cudaq.kernel
  def test_complex_vec_definition():
      c1 = [1.0 + 0j, 0., 0., 1.]

  counts = cudaq.sample(test_complex_vec_definition)
```

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1454


annagrin:

annagrin:
recheck

python/cudaq/kernel/ast_bridge.py
```diff
@@ -246,11 +246,26 @@ def getFloatAttr(self, type, value):
     def getConstantFloat(self, value):
         """
         Create a constant float operation and return its MLIR result Value.
-        Takes as input the concrete float value. 
+        Takes as input the concrete float value.
         """
         ty = self.getFloatType()
         return arith.ConstantOp(ty, self.getFloatAttr(ty, value)).result
 
+    def getComplexType(self):
```
amccaskey, line 254: It would be good to parameterize this with the precision or bitwidth in MLIR terms (32 or 64 is what we support)

python/cudaq/kernel/ast_bridge.py
```diff
@@ -246,11 +246,26 @@ def getFloatAttr(self, type, value):
...
+        """
+        Return an MLIR complex type (double precision).
+        """
+        return ComplexType.get(self.getFloatType())
+
+    def getConstantComplex(self, value):
```
amccaskey, line 260: Same here (parameterize with 32 or 64)

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2726,6 +2730,22 @@ def visit_Continue(self, node):
         else:
             cc.ContinueOp([])
 
+    def promote_operand_type(self, ty, operand):
```
amccaskey, line 2733: Can we move this up with the other utility methods on this visitor class?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2896,16 +2906,21 @@ def visit_Name(self, node):
 
         if node.id in self.capturedVars:
             # Only support a small subset of types here
+            complexType = type(1j)
```
amccaskey, line 2909: I think you can just use `complex` instead of getting the type of `1j`

amccaskey: We should also probably check for `np.complex64`

python/tests/kernel/test_kernel_float_complex.py
```diff
@@ -0,0 +1,118 @@
...
+    cudaq.set_target('nvidia-fp64')
+
+    f = [0., 1., 1., 0.]
+
+    # Pass a list of float as a parameter
+    @cudaq.kernel
+    def test_float_vec_param(vec : list[float]):
+        f1 = vec
+
+    counts = cudaq.sample(test_float_vec_param, f)
```
amccaskey, line 37: A better test here would be to define a `complex` return type on the kernel, and `assert test_float_vec_param(f) == that value`

something like 
```python
@cudaq.kernel
def test_float_vec_param(vec : list[float], i : int) -> complex: 
    f1 = vec 
    return vec[i]

for i in range(len(f)): 
   assert np.isclose(f[i], test_float_vec_param(f, i), atol=1e-6) 
```

python/tests/kernel/test_kernel_float_complex.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Use a float inside np in a kernel
+    @cudaq.kernel
+    def test_float_np_use():
+        f1 = np.sin(0)
+
+    counts = cudaq.sample(test_float_np_use)
+    assert len(counts) == 0
+
+def test_complex():
```
amccaskey, line 74: I would suggest also adding support for `c = np.array([...], dtype=np.complex64)` and demonstrate with the `nvidia` target. Also it would be good to test that we raise an exception for the use of `complex` with `nvidia` and `np.complex64` with FP64 backends (all the other ones).

python/cudaq/kernel/ast_bridge.py
```diff
@@ -246,11 +246,26 @@ def getFloatAttr(self, type, value):
...
+        Takes as input the concrete float value.
         """
         ty = self.getFloatType()
         return arith.ConstantOp(ty, self.getFloatAttr(ty, value)).result
 
+    def getComplexType(self):
+        """
+        Return an MLIR complex type (double precision).
+        """
+        return ComplexType.get(self.getFloatType())
```
schweitzpgi, line 258: Q: Is there a way to request either F32 or F64 in the python bindings? (This getFloatType() feels more ambiguous.)

python/tests/kernel/test_kernel_float_complex.py
annagrin: I created a separate issue for supporting np arrays: https://github.com/NVIDIA/cuda-quantum/issues/1523, will work in it in the next PR

python/cudaq/kernel/ast_bridge.py
annagrin: > I think you can just use complex instead of getting the type of 1j

We use complex dialect and it seems to override the complex type. Maybe worth decoupling in a separate PR...

> We should also probably check for np.complex64

Done.

annagrin:
Replaced by https://github.com/NVIDIA/cuda-quantum/pull/1550

python/cudaq/kernel/ast_bridge.py
annagrin: Done.

python/cudaq/kernel/ast_bridge.py
annagrin: Done.

python/cudaq/kernel/ast_bridge.py
annagrin: Done.

python/tests/kernel/test_kernel_float_complex.py
annagrin: Changed the tests for now to work around this issue (will update the tests after the issue is fixed in the experimental branch as well)

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2726,6 +2730,22 @@ def visit_Continue(self, node):
...
+                if IntegerType.isinstance(operand.type):
+                    operand = arith.SIToFPOp(self.getFloatType(),
+                                             operand).result
+                operand = complex.CreateOp(ComplexType.get(self.getFloatType()),
+                                           operand,
+                                           self.getConstantFloat(0.0)).result
+
+        if F64Type.isinstance(ty):
+            if IntegerType.isinstance(operand.type):
+                operand = arith.SIToFPOp(ty, operand).result
```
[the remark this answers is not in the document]
annagrin: I'd like to postpone this to the next PR(s) when I have examples that are running with different data types in `np.array` and assigning the state to a `qvector` (to be able to tests my changes properly) - does it is sound reasonable?

PR #1519, opened 2024-04-16
[rt] Rename getAvailableIndex function.

boschmitt:
### Description
The execution manager should not expose details of how qudit allocation is handle internally through its public API. User must only know that, when they ask for a qudit allocation, they get a qudit identifier back.

Users should not write code that rely on a qudit having a particular identifier---something that the current name might invite them to do.


amccaskey (approved):
This looks like a good change to me

PR #1524, opened 2024-04-17
Bump `cuquantum` to 24.03 and `cutensor` to 2.0

1tnguyen:
### Description

Some updates according to the [release notes](https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/release_notes.html#cutensornet-v2-4-0):    

-  `cutensornetStateApplyTensor` -> `cutensornetStateApplyTensorOperator`
       (introducing the latter, deprecating the former)

- Use the new `cutensornetStateApplyControlledTensorOperator`, which supports multi-controlled single-target gates
    
   + Add special handling for cSWAP, which is the only two-target gate defined in NVQIR runtime.

   + Remove the need for manual decomposition for MPS.

- Update version info in Dockerfile/build script/etc. as necessary.

Tested by: running `tensornet` and `tensornet-mps` unit tests.


bmhowe23 (commented):
Is this intended for 0.7.1? In any case, we should probably wait for the additional test in #1493 before merging.
Also, will this affect NVQC compatibility? Based on the changeset alone, it doesn't seem like it to me.

1tnguyen:
> Is this intended for 0.7.1? In any case, we should probably wait for the additional test in #1493 before merging. Also, will this affect NVQC compatibility? Based on the changeset alone, it doesn't seem like it to me.

No, it's intended for 0.8 and should only be merged after #1493 and 0.7.1 release.

runtime/nvqir/cutensornet/simulator_cutensornet.cpp
```diff
@@ -71,29 +71,25 @@ void SimulatorTensorNetBase::applyGate(const GateApplicationTask &task) {
...
   }
-  std::vector<int32_t> qubits;
-  for (const auto &qId : controls)
-    qubits.emplace_back(qId);
-  for (const auto &qId : targets)
-    qubits.emplace_back(qId);
-  m_state->applyGate(qubits, m_gateDeviceMemCache[gateKey]);
+  // Type conversion
+  const std::vector<int32_t> ctrlQubits(controls.begin(), controls.end());
+  const std::vector<int32_t> targetQubits(targets.begin(), targets.end());
```
schweitzpgi, line 91:
```suggestion
  const std::vector<std::int32_t> ctrlQubits(controls.begin(), controls.end());
  const std::vector<std::int32_t> targetQubits(targets.begin(), targets.end());
```
For consistency and safety.

runtime/nvqir/cutensornet/simulator_mps_register.cpp
```diff
@@ -111,243 +96,18 @@ class SimulatorMPS : public SimulatorTensorNetBase {
...
+    if (m_state->getNumQubits() > 64)
       throw std::runtime_error("State vector data is too large.");
     // Handle empty state (e.g., no qubit allocation)
     if (!m_state)
       return cudaq::State{{0}, {}};
-    const uint64_t svDim =
-        (1ull << (m_state->getNumQubits() - m_auxQubitsForGateDecomp.size()));
-    const std::vector<int32_t> projectedModes(m_auxQubitsForGateDecomp.begin(),
-                                              m_auxQubitsForGateDecomp.end());
+    const uint64_t svDim = (1ull << m_state->getNumQubits());
```
schweitzpgi, line 107:
```suggestion
    const std::uint64_t svDim = (1ull << m_state->getNumQubits());
```

schweitzpgi (approved):
Changes look good, but still some residual failures.

PR #1527, opened 2024-04-17
Add comments to the newly added docs example

1tnguyen:
### Description

The example `docs/sphinx/examples/cpp/providers/nvqc_qml.cpp` needs some comments to clarify what it illustrates.

### DIFF
--- a/docs/sphinx/examples/cpp/providers/nvqc_qml.cpp
+++ b/docs/sphinx/examples/cpp/providers/nvqc_qml.cpp
@@ -8,8 +8,9 @@
 // attain NVQC API key.
 
 // This example demonstrates the simulation of large parameterized quantum
-// circuits with NVQC. These circuits can be applied to quantum machine learning
-// to classify data points.
+// circuits with NVQC. These parameterized circuits can be applied to quantum
+// machine learning to classify data points, e.g. see
+// https://arxiv.org/pdf/1906.07682.pdf.
 
 #include <ctime>
 #include <cudaq.h>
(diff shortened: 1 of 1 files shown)


docs/sphinx/examples/cpp/providers/nvqc_qml.cpp
```diff
@@ -6,6 +6,11 @@
 // Assumes a valid NVQC API key has been set in the `NVQC_API_KEY` environment
 // variable. Please refer to the documentations for information about how to
 // attain NVQC API key.
+
+// This example demonstrates the simulation of large parameterized quantum
+// circuits with NVQC. These circuits can be applied to quantum machine learning
+// to classify data points.
```
bmhowe23, line 12: Is there a good article that might provide more context? An outsider might not make the connection between the `observe` calls in this example and machine learning. Perhaps https://arxiv.org/pdf/1906.07682.pdf or some other reference?

bmhowe23 (approved):
LGTM....one minor comment that I'll leave up to you whether or not to address; I'm ok either way.

docs/sphinx/examples/cpp/providers/nvqc_qml.cpp
1tnguyen: > Is there a good article that might provide more context?

I didn't know any at the time. 

Thanks for providing the reference. Added in https://github.com/NVIDIA/cuda-quantum/pull/1527/commits/54d48fa094ee6cd2fb59d78bb7bd628c000cb5f4

PR #1405, opened 2024-03-14
[Tests] [Python] Updates to 'pycudaq-mlir' tests

khalatepradnya:
Run with `cmake --build . --target check-pycudaq-mlir` from the `build` directory.

PR #1532, opened 2024-04-17
[Tests] [Python] Follow-up for 'pycudaq-mlir' tests

khalatepradnya:
Run the Python MLIR tests enabled in [PR#1405](https://github.com/NVIDIA/cuda-quantum/pull/1405) while building from source in the 'assets' dockerfile.


khalatepradnya:
Converting to draft till the CI stage for building assets is fixed...

docker/build/assets.Dockerfile
```diff
@@ -171,7 +172,9 @@ RUN cd /cuda-quantum && source scripts/configure_build.sh && \
     "$LLVM_INSTALL_PREFIX/bin/llvm-lit" -v build/test \
         --param nvqpp_site_config=build/test/lit.site.cfg.py && \
     "$LLVM_INSTALL_PREFIX/bin/llvm-lit" -v build/targettests \
-        --param nvqpp_site_config=build/targettests/lit.site.cfg.py
+        --param nvqpp_site_config=build/targettests/lit.site.cfg.py && \
+    "$LLVM_INSTALL_PREFIX/bin/llvm-lit" -v build/python/tests/mlir \
+        --param nvqpp_site_config=build/python/tests/mlir/lit.site.cfg.py
```
bettinaheim, line 177: Python is built separately in its own stage. You will need to copy over the build folder from that stage. Scikit is used to build the python support; the pyproject.toml defines the build and disables building tests. I haven't checked where they would end up in if they are enabled during the build. Give it a go and let me know if you run into trouble.

khalatepradnya:
Closing this since it needs further work in build configuration.

Setting this flag in the toml file - `CUDAQ_BUILD_TESTS=TRUE`, results in failure to build wheel file since there are dependencies on 'C++ frontend' and 'tools' components. Merely setting both these flags `CUDAQ_DISABLE_CPP_FRONTEND=FALSE` and `CUDAQ_DISABLE_TOOLS=FALSE` still results in build failure. The building of Python-specific `ctest` tests (i.e. `pycudaq-mlir`) isn't isolated and needs further work.

khalatepradnya:
> @khalatepradnya I fixed the issues with building the tests for the wheels in #1603. There is an update needed in the pyproject.toml file to add numpy and pytest as requirements for the build-system, and I updated the CMakefile.txt to only add the top-level test subdirectory if the CUDAQ_DISABLE_CPP_FRONTEND is false. Please double check if that picks up all necessary tests, or if I missed some (I see 50 tests running).

Thank you, I will check.

PR #1535, opened 2024-04-17
Unify NVQC image with regular release image

bmhowe23:
This change makes the NVQC image identical to the regular Docker image. The way to activate the "NVQC" behavior of the image is to set the `NVQC` environment variable to some non-zero length string.


docker/release/cudaq.nvqc.Dockerfile
```diff
@@ -15,29 +15,3 @@
 # Base image is CUDA Quantum image 
 ARG base_image=nvcr.io/nvidia/nightly/cuda-quantum:latest
 FROM $base_image as nvcf_image
-
```
schweitzpgi, line 18: Do we even need this dockerfile any more?

schweitzpgi (commented):
LGTM

Adrian999999999 (approved):

PR #1541, opened 2024-04-18
[rt] Remove ExecutionManager's qudit tracker.

boschmitt:
### Description

This commit achieves two goals:

1. Deduplicate the qudit tracking functionality, which is present in both the `ExecutionManager` and the `CircuitSimulator.

The `ExecutionManager` does not need to keep track of qudit's allocation and deallocation, it just needs to forward requests. Indeed, it is the simulators' job to manage the qudits and the quantum state representation.

Note: We should even consider moving qudit tracking from the `CircuitSimulator` base class to the subtypes since each simulator might want to handle memory differently. For example: eager deallocation of qudits might be desired when this deallocation can be done with a simple resize of the state vector. This deallocation would improve the performance of applying gates but requires specific information from the simulators themselves.

2. Completely moves the `Tracer` to the `CircuitSimulator`. Before this commit, the `Tracer` implementation was divided between the `ExecutionManager` and the `CircuitSimulator`.

Note: This move required adding another level of indirection for qudit reset within the simulators. Before, this indirection was handled by the `ExecutionManager` alone. We probably had an unknown bug since the execution of JITed kernels bypass the `ExecutionManager`, and thus it is quite likely the execution of JITed kernels in the `tracer` execution context would have exploded in the presence of qudit resets.


runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
```diff
@@ -37,60 +30,17 @@ class DefaultExecutionManager : public cudaq::BasicExecutionManager {
...
-    allocateQudits(requestedAllocations);
-    requestedAllocations.clear();
-  }
 
 protected:
-  void allocateQudit(const cudaq::QuditInfo &q) override {
-    requestedAllocations.emplace_back(2, q.id);
-  }
-
-  void allocateQudits(const std::vector<cudaq::QuditInfo> &qudits) override {
```
amccaskey, line 59: It was useful from a performance perspective to observe (in some way) that we were really allocating a qvector, and to queue them up and request the allocation all at once (means we can create the correct GPU pointer once, and not once + N-1 changes to the data on device). Is this type of functionality retained somewhere else? maybe i'll see it further down

boschmitt: It is not present anywhere else. I can implement it on the `CircuitSimulator` layer, so that JITed kernels would also benefit from it.

PR #1542, opened 2024-04-19
[stateHandling][RFC] Skeleton for chaining `get_state` -> `qvector` allocation

1tnguyen:
### Description

- Allowing users to retrieve the state (`get_state`) then use that state to initialize a `qvector` later on. 

- Retaining the underlying memory as much as we can (e.g., unless we need to do a Kronecker product)

### Notes

This has some dependencies on https://github.com/NVIDIA/cuda-quantum/pull/1537: `simulator_tensornet_register.cpp` implementation is incomplete, pending for a rebase after merging 1537.  

MPS is not yet implemented, pending cuquantum v24.03 update.

### DIFF
--- a/runtime/cudaq/qis/qvector.h
+++ b/runtime/cudaq/qis/qvector.h
@@ -10,6 +10,7 @@
 
 #include "cudaq/host_config.h"
 #include "cudaq/qis/qview.h"
+#include "cudaq/qis/state.h"
 
 namespace cudaq {
 
@@ -65,6 +66,19 @@ public:
       : qvector(std::vector<complex>{list.begin(), list.end()}) {}
   qvector(const std::initializer_list<complex> &list)
       : qvector(std::vector<complex>{list.begin(), list.end()}) {}
+
+  /// @brief Construct a `qvector` in a specific state.
+  // TODO: determine if we want to support state copying.
+  // On one hand, there could be a use case whereby the user want to keep the
+  // state intact. On the other hand, it's prone to excessive memory usage if
+  // not carefully managed.
+  qvector(const cudaq::state &initState) = delete;
+  qvector(cudaq::state &&initState) : qudits(initState.get_num_qubits()) {
+    std::vector<QuditInfo> targets;
+    for (auto &q : qudits)
+      targets.emplace_back(QuditInfo{Levels, q.id()});
+    getExecutionManager()->initializeState(targets, initState.release());
+  };
 
   /// @cond
   /// Nullary constructor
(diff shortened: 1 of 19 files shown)


amccaskey:
Very cool start. 

One question I had - This update makes the `state` follow `unique_ptr` semantics, what was the reason for that change?

I was hoping that the `state` could adhere to value semantics via the `shared_ptr` and just track the use-count for the state and deallocate when all uses were gone. One concern is that (as in your test) we require a `std::move` to pass the state to the kernel, which adds a bit of boilerplate. Another concern is that I think it is setup here to delete the unique pointer once its been used in `addQubitsToState`, but what if we wanted to seed subsequent kernels with that same state?

runtime/cudaq/qis/qvector.h
```diff
@@ -66,6 +67,19 @@ class qvector {
   qvector(const std::initializer_list<complex> &list)
       : qvector(std::vector<complex>{list.begin(), list.end()}) {}
 
+  /// @brief Construct a `qvector` in a specific state.
+  // TODO: determine if we want to support state copying.
+  // On one hand, there could be a use case whereby the user want to keep the
+  // state intact. On the other hand, it's prone to excessive memory usage if
+  // not carefully managed.
+  qvector(const cudaq::state &initState) = delete;
```
schweitzpgi, line 75: Are you saying the qvector will necessarily make a copy of `initState`? We can just promote the reference to a pointer here, no? That's not as safe as move/swap operations, but it can be documented that the reference has to remain live.

1tnguyen:
> One question I had - This update makes the state follow unique_ptr semantics, what was the reason for that change?

The main reason was that we'll give the underlying memory back to the simulator (e.g., directly set the `deviceStateVector` for `custatevec`, not doing any `cudaMalloc` the memCopy, etc.) 

In the `qvector` ctor, I purposely disallowed `qvector(const cudaq::state &initState) = delete;` for the same reason.
I don't like the need for `std::move` either. We could certainly make copy the state the default unless the user specially asks for a move.
I'm open to suggestions here...

runtime/cudaq/qis/qvector.h
1tnguyen: I think in general we could but it'd make the simulator implementation a bit trickier since it now needs to track whether the memory is a borrowed one or something it owns. Plus, if user later call a `get_state`, we then need to do a copy since `get_state` follows the move semantics. 

It's this complication that I think a move semantics loop is tidier...

PR #1547, opened 2024-04-19
[fixup-linkage] Open the door wider.

schweitzpgi:
This tool was being very cautious about rewriting the linkages on the C++ code. This change makes things more aggressive. This implies that the tool may fail by changing linkages that don't need to be changed and were not formerly being changed. The motivation for this is to allow linkages when generating shared library code to be fixed up.

### DIFF
--- a/tools/fixup-linkage/fixup-linkage.cpp
+++ b/tools/fixup-linkage/fixup-linkage.cpp
@@ -75,7 +75,7 @@ int main(int argc, char *argv[]) {
       continue;
     }
     if (line.find(" linkonce_odr ") != std::string::npos ||
-	line.find(" weak dso_local ") != std::string::npos) {
+        line.find(" weak dso_local ") != std::string::npos) {
       outFile << line << std::endl;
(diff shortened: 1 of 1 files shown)


schweitzpgi:
This PR is merging to `main` and also faceplants on the "Assets" part of the CI.

bmhowe23:
> This PR is merging to `main` and also faceplants on the "Assets" part of the CI.

I think this PR (https://github.com/scikit-build/scikit-build-core/pull/698) caused the wheels in our Assets build to get renamed from `cuda_quantum-0.0.0-cp311-cp311-manylinux_2_28_x86_64.whl` to `cuda_quantum-0.0.0-cp311-cp311-linux_x86_64.whl`, and the `docker/build/assets.Dockerfile` doesn't support that. I'll make a PR to updated our CI, but we should really investigate pinning all the versions of software used in our CI so that this stuff doesn't keep happening.

amccaskey (approved):

Issue #1548, opened 2024-04-19
Support complex numbers and lists of complex numbers defined inside, passed as parameters, or captured by kernel


annagrin:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

Make sure we can use complex numbers inside kernel function definitions, for example:

```python
  c = [.70710678 + 0j, 0., 0., 0.70710678]
  
  # Pass a list of complex numbers as a parameter
  @cudaq.kernel
  def test_complex_vec_param(vec : list[complex]):
      c1 = vec

  counts = cudaq.sample(test_complex_vec_param, c)
```

```python
  # Capture a list of complex numbers
  @cudaq.kernel
  def test_complex_vec_capture():
      c1 = c

   counts = cudaq.sample(test_complex_vec_capture)
```

```python
  # Define a list of complex numbers inside a kernel
  @cudaq.kernel
  def test_complex_vec_definition():
      c1 = [1.0 + 0j, 0., 0., 1.]

  counts = cudaq.sample(test_complex_vec_definition)
```

annagrin:
Closed by https://github.com/NVIDIA/cuda-quantum/pull/1550

Issue #1549, opened 2024-04-19
Support using complex numbers and lists in kernel functions


annagrin:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

This is a parent issue for supporting using complex numbers and lists in kernel functions.

The work is split up into the following parts:

- [x] Support complex numbers and lists of complex numbers defined inside, passed as parameters, or captured by kernel functions
  - https://github.com/NVIDIA/cuda-quantum/issues/1548
  - https://github.com/NVIDIA/cuda-quantum/pull/1550

- [ ] Support `np.complex` arrays defined inside, passed as parameters, or captured by kernel
  - https://github.com/NVIDIA/cuda-quantum/issues/1523

- [ ] Consider supporting half or quad precision complex and float numbers
  - TODO

- [x] Return complex numbers from kernel functions
  - https://github.com/NVIDIA/cuda-quantum/issues/1608
  - https://github.com/NVIDIA/cuda-quantum/pull/1610

schweitzpgi:
@annagrin This one may be complete now too. Please close if warranted.

bettinaheim:
Closing this as done, since np arrays are tracked separately.

PR #1550, opened 2024-04-19
Support using complex inside kernel functions

annagrin:
### Description

Allow users to do following in kernel functions:

- pass complex numbers and lists as parameters
- capture complex numbers and lists
- define lists of complex numbers
- pass complex numbers to function calls
- allow using `complex`, `np.complex128` and `np.complex64`

- add tests for complex and float numbers and lists

Helps: https://github.com/NVIDIA/cuda-quantum/issues/1086

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1548
Closes: https://github.com/NVIDIA/cuda-quantum/issues/1454

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -231,25 +231,67 @@ class PyASTBridge(ast.NodeVisitor):
         """
         return IntegerAttr.get(type, value)
 
-    def getFloatType(self):
+    def getFloatType(self, width=64):
         """
-        Return an MLIR float type (double precision).
+        Return an MLIR float type (single or double precision).
         """
-        return F64Type.get()
+        return F64Type.get() if width == 64 else F32Type.get()
 
     def getFloatAttr(self, type, value):
         """
-        Return an MLIR float attribute (double precision).
+        Return an MLIR float attribute (single or double precision).
         """
         return FloatAttr.get(type, value)
 
-    def getConstantFloat(self, value):
+    def getConstantFloat(self, value, width=64):
         """
         Create a constant float operation and return its MLIR result Value.
-        Takes as input the concrete float value. 
+        Takes as input the concrete float value.
         """
-        ty = self.getFloatType()
+        ty = self.getFloatType(width=width)
+        return self.getConstantFloatWithType(value, ty)
+
+    def getConstantFloatWithType(self, value, ty):
+        """
+        Create a constant float operation and return its MLIR result Value.
+        Takes as input the concrete float value.
+        """
         return arith.ConstantOp(ty, self.getFloatAttr(ty, value)).result
+
+    def getComplexType(self, width=64):
+        """
+        Return an MLIR complex type (single or double precision).
+        """
+        return self.getComplexTypeWithElementType(
+            self.getFloatType(width=width))
+
+    def getComplexTypeWithElementType(self, eTy):
+        """
+        Return an MLIR complex type (single or double precision).
+        """
+        return ComplexType.get(eTy)
+
+    def getConstantComplex(self, value, width=64):
+        """
+        Create a constant complex operation and return its MLIR result Value.
+        Takes as input the concrete complex value.
+        """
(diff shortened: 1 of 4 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -231,26 +231,68 @@ def getIntegerAttr(self, type, value):
...
-        Return an MLIR float type (double precision).
+        Return an MLIR float type (single or double precision).
         """
-        return F64Type.get()
+        return F64Type.get() if width == 64 else F32Type.get()
 
     def getFloatAttr(self, type, value):
         """
-        Return an MLIR float attribute (double precision).
+        Return an MLIR float attribute (single or double precision).
```
schweitzpgi, line 242: I guess someone should ask: there are no plans for half or quad precision, right?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -231,26 +231,68 @@ def getIntegerAttr(self, type, value):
...
         """
-        ty = self.getFloatType()
         return arith.ConstantOp(ty, self.getFloatAttr(ty, value)).result
 
+    def getComplexType(self, width=64):
+        """
+        Return an MLIR complex type (single or double precision).
+        """
+        return self.getComplexTypeWithElementType(
+            self.getFloatType(width=width))
```
schweitzpgi, line 266: We have to be careful here (document it?) since MLIR's complex<64> is the same as the numpy complex128 as far as I understand it. (I found this below.)

python/tests/kernel/test_kernel_complex.py
```diff
@@ -0,0 +1,311 @@
...
+        return v.real
+    assert is_equal(0., complex_np_use_real())
+
+
+
+
+
+
+
+
```
schweitzpgi, line 311: nit: whitespace

schweitzpgi (commented):
Looking good overall. I assume you're looking at the test failures.

python/cudaq/kernel/ast_bridge.py
annagrin: I think we should add them, but it is better to add it separately when needed - trying to not to grow this PR any bigger...
Will create and issue for that as part of https://github.com/NVIDIA/cuda-quantum/issues/1549

python/cudaq/kernel/ast_bridge.py
annagrin: Added comments, thanks!

python/tests/kernel/test_kernel_complex.py
annagrin: removed, thanks!

schweitzpgi:
Looks good to me. Thanks, Anna.

schweitzpgi (approved):

amccaskey:
I see a `dictionary.dic` file in the diff. Can you remove it?

python/tests/kernel/test_kernel_complex.py
```diff
@@ -0,0 +1,536 @@
...
+    """Test that we can pass complex lists to kernel functions."""
+
+    # Pass a list of complex as a parameter
+    c = [.70710678 + 1j, 2j, 0., 0.70710678]
+
+    @cudaq.kernel
+    def complex_vec_param(vec: list[complex], i: int) -> complex:
+        return vec[i]
+
+    # Returning complex is not supported yet
```
amccaskey, line 36: any reason not to support this here. seems like it'd be a simple update to py_alt_launch_kernel where you already updated for F32

amccaskey (approved):
small comments, +1

python/tests/kernel/test_kernel_complex.py
annagrin: Will add complex return types in a separate PR

schweitzpgi: I think a train of PRs, each one a smaller chunk of work that's easier to review, is a good idea. 👍

PR #1554, opened 2024-04-22
[experimental/StateHandling] add new state type

schweitzpgi:
Add a cc.state type to the CC dialect.
Replace !cc.struct<"cudaq.state"> with !cc.state.
Requires #1534 to be merged first.

### DIFF
--- a/lib/Optimizer/Builder/Intrinsics.cpp
+++ b/lib/Optimizer/Builder/Intrinsics.cpp
@@ -159,10 +159,10 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   })#"},
 
     {cudaq::getNumQubitsFromCudaqState, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.struct<"cudaq.state">>) -> i64
+  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.state>) -> i64
   )#"},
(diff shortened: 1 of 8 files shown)


lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -159,10 +159,10 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   })#"},
 
     {cudaq::getNumQubitsFromCudaqState, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.struct<"cudaq.state">>) -> i64
+  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.state>) -> i64
   )#"},
     {cudaq::getCudaqStateAsVector, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_vectorData(%p : !cc.ptr<!cc.struct<"cudaq.state">>) -> !cc.ptr<f64>
+  func.func private @__nvqpp_cudaq_state_vectorData(%p : !cc.ptr<!cc.state>) -> !cc.ptr<f64>
```
amccaskey, line 165: This function is the only question I have here. Not all simulators will produce a state that can be described by a 1D array of complex values. I'm not sure we will be able to use an intrinsic like this, and am wondering if we can just keep the `cc.state` around until final lowering in LowerToQIR, and lower to a specific function for it like `__quantum__qis__initialize_state_abstract(..., state: ptr)`?

amccaskey (approved):

PR #1556, opened 2024-04-22
[experimental/stateHandling] extend the semantics of quake.init_state

schweitzpgi:
Extend the init_state op to take !cc.state types.

This changes the semantics of init_state to be a family of initializations. It can use a vector of floats directly or a cudaq::state object. Additionally, it may use a runtime library call with move semantics if the initializer is a cudaq::state object.

Requires #1554 to be merged first.

### DIFF
--- a/lib/Optimizer/Builder/Intrinsics.cpp
+++ b/lib/Optimizer/Builder/Intrinsics.cpp
@@ -159,10 +159,10 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   })#"},
 
     {cudaq::getNumQubitsFromCudaqState, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.state>) -> i64
+  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.state) -> i64
   )#"},
     {cudaq::getCudaqStateAsVector, {}, R"#(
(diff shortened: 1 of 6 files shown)


lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -159,10 +159,10 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   })#"},
 
     {cudaq::getNumQubitsFromCudaqState, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!cc.struct<"cudaq.state">>) -> i64
+  func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.state) -> i64
   )#"},
     {cudaq::getCudaqStateAsVector, {}, R"#(
-  func.func private @__nvqpp_cudaq_state_vectorData(%p : !cc.ptr<!cc.struct<"cudaq.state">>) -> !cc.ptr<f64>
+  func.func private @__nvqpp_cudaq_state_vectorData(%p : !cc.state) -> !cc.ptr<f64>
```
amccaskey, line 165: Should we just delete this intrinsic now?

schweitzpgi: Seems reasonable.

amccaskey (approved):

PR #1557, opened 2024-04-23
[tools] Fix segfault in `cudaq-translate` when `emit-llvm=0`

fabianmcg:
This patch fixes a segmentation fault in `cudaq-translate` when `--emit-llvm` is false. The issue occurs because the initialization lambda has non-trivial storage, and `llvm::function_ref` only holds a reference to a function, trashing the lambda after initialization.


fabianmcg:
recheck

fabianmcg:

fabianmcg:
recheck

schweitzpgi (commented):
Thanks for the bug fix, Fabian.

PR #1563, opened 2024-04-24
[experimental/stateHandling] Add rvalue reference to bridge.

schweitzpgi:
This is intended to lower the rvalue reference to a cudaq::state to the newly extended quake.init_state operation, which now supports lowering of move semantics.


schweitzpgi:
ping

anthony-santana (approved):

PR #1564, opened 2024-04-24
Remove mpi4py plugin that is no longer in use 

1tnguyen:
### Description

We've removed this plugin type in https://github.com/NVIDIA/cuda-quantum/pull/1191 from users' docs (0.7.0).
This is the actual code cleanup to remove the unused code.

**Note**: there is an edit to the `Run Python MPI tests` CI job, which checks the install instructions in Python README (i.e., using `conda` then activate the MPI plugin). Since the docker image (`wheel_validation:local`) has a pre-installed cuda-quantum pip package (not via conda), the step `conda run -n cuda-quantum pip install <wheel file>` parsed from README file may result in `conda` picking up that pip package (if it was installed with `--user`) rather than installing the wheel to `$CONDA_PREFIX`. This would make the activation step failed due to wrong file location. Previously, it was masked by the mpi4py-based plugin, which has now been removed.


bmhowe23 (commented):
Does `python/README.md` need to be updated to change anything with respect to the `activate_custom_mpi.sh` script?
In any case, it might be good to push #1493 before merging this PR as #1493 adds additional tests that might be useful for this.

1tnguyen:
> Does `python/README.md` need to be updated to change anything with respect to the `activate_custom_mpi.sh` script?

The question in my mind about changing the README vs. updating the test setup is whether `pip install cuda-quantum` and then `conda` steps are common usage or just an artifact of reusing the wheel validation container image. 

If we expect the first, then we definitely should change the README file instead. 

@bmhowe23 Do you have any suggestions?

> In any case, it might be good to push #1493 before merging this PR as #1493 adds additional tests that might be useful for this.

 Agreed.

bmhowe23:
> > Does `python/README.md` need to be updated to change anything with respect to the `activate_custom_mpi.sh` script?
> 
> The question in my mind about changing the README vs. updating the test setup is whether `pip install cuda-quantum` and then `conda` steps are common usage or just an artifact of reusing the wheel validation container image.
> 
> If we expect the first, then we definitely should change the README file instead.
> 
> @bmhowe23 Do you have any suggestions?
> 
> > In any case, it might be good to push #1493 before merging this PR as #1493 adds additional tests that might be useful for this.
> 
> Agreed.

I think the instructions say to do the `conda` steps first and then the `pip install cuda-quantum`, so it sounds to me like no README.md update is needed.

schweitzpgi (approved):
LGTM

bmhowe23 (approved):

PR #1565, opened 2024-04-24
Split the remove measurements pass into its own source file.

schweitzpgi:
~Apply pass in parallel.~

Remove dead code.


runtime/cudaq/platform/default/rest/helpers/ionq/ionq.config
```diff
@@ -23,7 +23,7 @@ CODEGEN_EMISSION=qir-base
 
 # Additional passes to run after lowering to QIR
 # This is required due to https://github.com/NVIDIA/cuda-quantum/issues/512
-POST_CODEGEN_PASSES="remove-measurements"
+POST_CODEGEN_PASSES="func.func(remove-measurements),llvm.llvmfunc(remove-measurements)"
```
bmhowe23, line 26: What does `llvm.llvmfunc()` do here?

bmhowe23: Does it run `func.func(...)` in parallel with `llvm.llvmfunc(...)`, or does it just run each one sequentially? (And each one can run on the corresponding ops in parallel?)

schweitzpgi: IIRC, the pass manager can run them both in parallel.

schweitzpgi: Alas, making this parallel was causing the async tests to fail with the ionq backend. I'll file an issue.

bmhowe23 (approved):

PR #1567, opened 2024-04-25
Fix Pauli word conversion bug in `observe_async`

### LINKED ISSUE #1566: Cannot pass a list of Pauli words to `observe_async`
### Required prerequisites

### Describe the bug

Passing a list of Pauli words to a `cudaq.observe_async()` function produces the following error:
```
RuntimeError: Unable to cast Python instance of type <class 'str'> to C++ type 'cudaq::pauli_word'
```

It works fine if you pass to `cudaq.observe()`.

You can reproduce the bug by create a file called test.py with the following:
```python
import cudaq
import numpy as np
import random

random.seed(13)
np.random.seed(13)

def generateRandomPauliStrings(numQubits, numPaulis):
    s = ['X', 'Y', 'Z', 'I']
    return [
        ''.join([random.choice(s)
                for i in range(numQubits)])
        for i in range(numPaulis)
    ]

def build_cudaq_obs(hs, paulis):
    observable = cudaq.SpinOperator()
    for h, p in zip(hs, paulis):
        observable += h * cudaq.SpinOperator.from_word(p)
    return observable - cudaq.SpinOperator()


@cudaq.kernel
def gqeCirc(N: int, thetas: list[float], paulis: list[cudaq.pauli_word]):
    q = cudaq.qvector(N)
    for i in range(len(paulis)):
        exp_pauli(thetas[i], q, paulis[i])

numQubits = 4
numPaulis = 8
numberOfTerms = 4

# Generate the observable
obs_ps = generateRandomPauliStrings(numQubits, numberOfTerms)
obs = build_cudaq_obs([1.0]*len(obs_ps), obs_ps)

pauliStings = generateRandomPauliStrings(numQubits, numPaulis)
ts = np.random.rand(len(pauliStings))

exp_val = cudaq.observe_async(gqeCirc, obs, numQubits, list(ts), pauliStings).get().expectation()
print(exp_val)
```
When run, it produces this error:
```bash
$ python3 test2.py 
RuntimeError: Unable to cast Python instance of type <class 'str'> to C++ type 'cudaq::pauli_word'
```

### Expected behavior

It should run without errors

Not a regression

### Environment

- **CUDA Quantum version**: latest (bf579701d)
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

PR forthcoming

bmhowe23:
Fixes #1566


schweitzpgi (commented):
LGTM

python/utils/OpaqueArguments.h
```diff
@@ -155,22 +155,22 @@ inline py::args simplifiedValidateInputArguments(py::args &args) {
         throw std::runtime_error("Cannot pass ndarray with shape != (N,).");
 
       arg = args[i].attr("tolist")();
+    } else if (py::isinstance<py::str>(arg)) {
+      arg = cudaq::pauli_word(py::cast<std::string>(arg));
```
amccaskey, line 159: If you do it this way you may be able to remove the casting behavior from kernel_decorator

bmhowe23: > If you do it this way you may be able to remove the casting behavior from kernel_decorator

Hmm, removing that code seems to cause errors in `python/tests/kernel/test_kernel_features.py::test_pauli_word_input`:
```
RuntimeError: error: Can not handle conversion of python type <class 'str'> to MLIR type.
```

amccaskey (approved):

PR #1573, opened 2024-04-25
[Core] Add support for python's for-else to cc.loop.

schweitzpgi:
Add an else region to cc.loop.
Add builders, verifiers, etc.
Add roundtrip test.
Add lowering of cc.loop to CFG.
Add test for python for-else to cfg conversion.


schweitzpgi:
ping for review

anthony-santana (approved):

PR #1574, opened 2024-04-26
[NFC] Code cleanup

schweitzpgi:
- Add a note that the backendConfig.cpp file and it's content can be done more cleanly by generating the string in the compiler when we deprecate support for library mode and server-side execution.
- Tweak a few definitions to be in "storage-class specifiers type name = init;" order for regularity.


1tnguyen (approved):
LGTM. Thanks @schweitzpgi!

PR #1577, opened 2024-04-26
Refactor pipelines.

schweitzpgi:
NFC adjacent.

- Remove the template code. Not needed. Simple overload is sufficient.
- Move definitions out of the header. Not performance critical. Have to link the lib anyway.
- Update doxygen to reflect changes.
- Use MLIR's "convert" and "translate" distinction in terminology.
- Update all callers.

### DIFF
--- a/include/cudaq/Optimizer/CodeGen/Pipelines.h
+++ b/include/cudaq/Optimizer/CodeGen/Pipelines.h
@@ -37,7 +37,7 @@ inline void addPipelineConvertToQIR(mlir::PassManager &pm) {
 inline void addPipelineConvertToQIR(mlir::PassManager &pm,
                                     mlir::StringRef convertTo) {
   commonPipelineConvertToQIR(pm, convertTo);
-  cudaq::opt::addQIRProfilePipeline(pm, convertTo);
+  addQIRProfilePipeline(pm, convertTo);
 }
 
 void addPipelineTranslateToOpenQASM(mlir::PassManager &pm);
--- a/include/cudaq/Optimizer/CodeGen/Pipelines.h
+++ b/include/cudaq/Optimizer/CodeGen/Pipelines.h
@@ -21,8 +21,8 @@
 
 namespace cudaq::opt {
 
-void commonPipelineConvertToQIR(mlir::PassManager &pm,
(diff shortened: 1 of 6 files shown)


include/cudaq/Optimizer/CodeGen/Pipelines.h
```diff
@@ -21,60 +21,28 @@
...
-  pm.addPass(cudaq::opt::createConvertToQIRPass());
-  if constexpr (QIRProfile) {
-    cudaq::opt::addQIRProfilePipeline(pm, convertTo);
-  }
-}
+/// The common pipeline.
+/// Adds the common pipeline (with or without a profile specifier) but without
+/// the final QIR profile lowering passes.
+void commonPipelineConvertToQIR(
+    mlir::PassManager &pm, const std::optional<mlir::StringRef> &convertTo);
```
bmhowe23, line 28: Nit: any reason to have a reference to an already reference-like object (StringRef)?
```suggestion
    mlir::PassManager &pm, const std::optional<mlir::StringRef> convertTo);
```

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "cudaq/Optimizer/CodeGen/Pipelines.h"
+
+using namespace mlir;
+
+void cudaq::opt::commonPipelineConvertToQIR(
+    PassManager &pm, const std::optional<StringRef> &convertTo) {
```
bmhowe23, line 14: If you accept the above suggestion, you'd need this change, too.
```suggestion
    PassManager &pm, const std::optional<StringRef> convertTo) {
```

bmhowe23 (approved):
👍

include/cudaq/Optimizer/CodeGen/Pipelines.h
schweitzpgi: Prevents making an unneeded copy of the std::optional value.

PR #1582, opened 2024-04-29
[StateHandling] State data with value semantics via `state` acting as a thin pointer wrapper   

1tnguyen:
### Description

This is an alternative approach to https://github.com/NVIDIA/cuda-quantum/pull/1542 whereby the user-facing `state` is a thin wrapper around the underlying `SimulationState` data (i.e., passing by reference or value doesn't really matter); plus sub-state allocation should force an implicit copy, and `qvector` initialization from existing state + subsequent `get_state` should also force a copy.

- Implementations for in-source simulator backends (except MPS) are provided. 

- Adding an NVQIR function prototype (along the existing allocation from vector QIR functions)

- Adding NVQIR tests.

### DIFF
--- a/runtime/cudaq/qis/execution_manager.h
+++ b/runtime/cudaq/qis/execution_manager.h
@@ -122,6 +122,14 @@ public:
   /// simulation state.
   virtual void initializeState(const std::vector<QuditInfo> &targets,
                                const SimulationState *state) = 0;
+
+  /// @brief Initialize the state of the given qudits to the provided
+  /// simulation state.
+  virtual void initializeState(const std::vector<QuditInfo> &targets,
+                               const SimulationState &state) {
+    initializeState(targets, &state);
+  }
+
   /// Apply the quantum instruction with the given name, on the provided target
   /// qudits. Supports input of control qudits and rotational parameters. Can
   /// also optionally take a spin_op as input to affect a general Pauli
--- a/runtime/cudaq/qis/execution_manager.h
+++ b/runtime/cudaq/qis/execution_manager.h
@@ -17,6 +17,7 @@
 
 namespace cudaq {
 class ExecutionContext;
+class SimulationState;
 using SpinMeasureResult = std::pair<double, sample_result>;
 
 /// A QuditInfo is a type encoding the number of \a levels and the \a id of the
@@ -117,7 +118,10 @@ public:
   virtual void initializeState(const std::vector<QuditInfo> &targets,
                                const void *state,
                                simulation_precision precision) = 0;
-
+  /// @brief Initialize the state of the given qudits to the provided
+  /// simulation state.
+  virtual void initializeState(const std::vector<QuditInfo> &targets,
+                               const SimulationState *state) = 0;
   /// Apply the quantum instruction with the given name, on the provided target
   /// qudits. Supports input of control qudits and rotational parameters. Can
   /// also optionally take a spin_op as input to affect a general Pauli
--- a/runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
+++ b/runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
@@ -86,6 +86,25 @@ protected:
     requestedAllocations.clear();
   }
 
+  void initializeState(const std::vector<cudaq::QuditInfo> &targets,
+                       const cudaq::SimulationState *state) override {
+    // Note: a void* ptr doesn't provide enough info to the simulators, hence
+    // need a dedicated code path.
+    // TODO: simplify/combine the two code paths (raw vector and state).
+    if (!requestedAllocations.empty() &&
+        targets.size() != requestedAllocations.size()) {
+      assert(targets.size() < requestedAllocations.size());
+      const auto numDefaultAllocs =
+          requestedAllocations.size() - targets.size();
+      simulator()->allocateQubits(numDefaultAllocs);
+      // The targets will be allocated in a specific state.
+      simulator()->allocateQubits(targets.size(), state);
+    } else {
+      simulator()->allocateQubits(requestedAllocations.size(), state);
+    }
+    requestedAllocations.clear();
+  }
+
   void deallocateQudit(const cudaq::QuditInfo &q) override {
 
     // Before trying to deallocate, make sure the qudit hasn't
--- a/runtime/cudaq/qis/qvector.h
+++ b/runtime/cudaq/qis/qvector.h
@@ -82,28 +82,23 @@ public:
   qvector() : qudits(1) {}
   /// @endcond
 
+  //===--------------------------------------------------------------------===//
+  // qvector with an initial state
+  //===--------------------------------------------------------------------===//
   /// @brief Construct a `qvector` from a pre-existing `state`.
   /// This `state` could be constructed with `state::from_data` or retrieved
   /// from an cudaq::get_state.
-  qvector(state state) : qudits(state.get_num_qubits()) {
+  explicit qvector(const state &state) : qudits(state.get_num_qubits()) {
     std::vector<QuditInfo> targets;
     for (auto &q : qudits)
       targets.emplace_back(QuditInfo{Levels, q.id()});
     // Note: the internal state data will be cloned by the simulator backend.
     getExecutionManager()->initializeState(targets, state.internal.get());
   }
-  //===--------------------------------------------------------------------===//
-  // qvector with an initial state
-  //===--------------------------------------------------------------------===//
-  // TODO: selectively enable these as needed. In this implementation, only
-  // using `state` by-value is needed (state is a thin wrapper around the
-  // SimulationState pointer).
-  //
-  // explicit qvector(const state *);
-  // explicit qvector(const state &)
-  // explicit qvector(state *);
-  // explicit qvector(state &);
-  // explicit qvector(state &&);
+  explicit qvector(const state *ptr) : qvector(*ptr){};
+  explicit qvector(state *ptr) : qvector(*ptr){};
+  explicit qvector(state &s) : qvector(const_cast<const state &>(s)){};
+  explicit qvector(state &&s) : qvector(s){};
 
   /// @brief `qvectors` cannot be copied
   qvector(qvector const &) = delete;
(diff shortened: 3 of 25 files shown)


runtime/cudaq/qis/execution_manager.h
```diff
@@ -117,7 +118,10 @@ class ExecutionManager {
   virtual void initializeState(const std::vector<QuditInfo> &targets,
                                const void *state,
                                simulation_precision precision) = 0;
-
+  /// @brief Initialize the state of the given qudits to the provided
+  /// simulation state.
+  virtual void initializeState(const std::vector<QuditInfo> &targets,
+                               const SimulationState *state) = 0;
```
schweitzpgi, line 124: Do you want to provide the overload
```c++
virtual void initializeState(const std::vector<QuditInfo> &targets,
                               const SimulationState &state) {
    initializeState(targets, &state);
}
```

runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
```diff
@@ -86,6 +86,25 @@ class DefaultExecutionManager : public cudaq::BasicExecutionManager {
...
+    // need a dedicated code path.
+    // TODO: simplify/combine the two code paths (raw vector and state).
+    if (!requestedAllocations.empty() &&
+        targets.size() != requestedAllocations.size()) {
+      assert(targets.size() < requestedAllocations.size());
+      const auto numDefaultAllocs =
+          requestedAllocations.size() - targets.size();
+      simulator()->allocateQubits(numDefaultAllocs);
+      // The targets will be allocated in a specific state.
+      simulator()->allocateQubits(targets.size(), state);
```
schweitzpgi, line 101: Do these two allocations need to be grouped back together in any way? Does the order of the allocations here matter?

I'm thinking about computing offsets into a "register/array/vector" of qubits.

runtime/cudaq/qis/qvector.h
```diff
@@ -84,12 +85,20 @@ class qvector {
...
+  /// @brief Construct a `qvector` from a pre-existing `state`.
+  /// This `state` could be constructed with `state::from_data` or retrieved
+  /// from an cudaq::get_state.
+  explicit qvector(const state &state) : qudits(state.get_num_qubits()) {
+    std::vector<QuditInfo> targets;
+    for (auto &q : qudits)
+      targets.emplace_back(QuditInfo{Levels, q.id()});
+    // Note: the internal state data will be cloned by the simulator backend.
+    getExecutionManager()->initializeState(targets, state.internal.get());
+  }
```
schweitzpgi, line 97: I suppose this might be defined in the .cpp file and just declared here. Then you would only have to include the state.h file from the .cpp file...

runtime/cudaq/qis/qvector.h
```diff
@@ -84,12 +85,20 @@ class qvector {
...
+  explicit qvector(const state *ptr) : qvector(*ptr){};
+  explicit qvector(state *ptr) : qvector(*ptr){};
+  explicit qvector(state &s) : qvector(const_cast<const state &>(s)){};
+  explicit qvector(state &&s) : qvector(s){};
```
schweitzpgi, line 101:
```suggestion
  explicit qvector(state &&s) : qvector(std::move(s)){};
```
Does qvector allow for moves?

runtime/nvqir/CircuitSimulator.h
```diff
@@ -616,6 +618,10 @@ class CircuitSimulatorBase : public CircuitSimulator {
       addQubitToState();
   }
 
+  /// @brief Add (appending) the given simulation state to the current simulator
+  /// state.
+  virtual void addQubitsToState(const cudaq::SimulationState &state) = 0;
```
schweitzpgi, line 623: Random thought: could this be a NOP default implementation instead of `= 0` here? That way the derived class wouldn't have to define this method if for example it didn't support initial states for some reason.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -884,6 +890,51 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
+    }
+
+    cudaq::info("Allocating {} new qubits.", count);
+
+    previousStateDimension = stateDimension;
+    nQubitsAllocated += count;
+    stateDimension = calculateStateDim(nQubitsAllocated);
+
+    // Tell the subtype to allocate more qubits
+    addQubitsToState(*state);
```
schweitzpgi, line 928: Is `state` always freed in this call?

runtime/nvqir/qpp/QppCircuitSimulator.cpp
```diff
@@ -223,6 +223,19 @@ class QppCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
...
+    const QppState *const casted = dynamic_cast<const QppState *>(&in_state);
+    if (!casted)
+      throw std::invalid_argument(
+          "[QppCircuitSimulator] Incompatible state input");
+
+    if (state.size() == 0)
+      state = casted->state;
+    else
+      state = qpp::kron(casted->state, state);
+    return;
```
schweitzpgi, line 236:
```suggestion

```
nit: not needed.

runtime/nvqir/qpp/QppDMCircuitSimulator.cpp
```diff
@@ -220,6 +220,18 @@ class QppNoiseCircuitSimulator : public nvqir::QppCircuitSimulator<qpp::cmat> {
...
+  void addQubitsToState(const cudaq::SimulationState &in_state) override {
+    const QppDmState *const casted =
+        dynamic_cast<const QppDmState *>(&in_state);
+    if (!casted)
+      throw std::invalid_argument(
+          "[QppNoiseCircuitSimulator] Incompatible state input");
+
+    state =
+        (state.size() == 0) ? casted->state : qpp::kron(casted->state, state);
+    return;
```
schweitzpgi, line 232:
```suggestion
    
```

unittests/qir/NVQIRTester.cpp
```diff
@@ -71,6 +71,14 @@ void __quantum__qis__apply__general(Array *data, int64_t n_qubits, ...);
 
 // Qubit array allocation / deallocation
 Array *__quantum__rt__qubit_allocate_array(uint64_t idx);
+Array *
+__quantum__rt__qubit_allocate_array_with_state_fp64(uint64_t size,
```
schweitzpgi, line 75:
```suggestion
__quantum__rt__qubit_allocate_array_with_state_fp64(std::uint64_t size,
```

schweitzpgi (commented):
OK. I think this means that the simulator returns state data in a unique_ptr. Ownership then passes to the caller at that point. The the state can be passed back to the simulator. No deep copies since the unique_ptr can use a swap to avoid it.

We discussed letting the user force a deep copy. Will that be a different PR? Or are we abandoning that for now?

1tnguyen:
> OK. I think this means that the simulator returns state data in a unique_ptr. Ownership then passes to the caller at that point. The the state can be passed back to the simulator. No deep copies since the unique_ptr can use a swap to avoid it.
> 
> We discussed letting the user force a deep copy. Will that be a different PR? Or are we abandoning that for now?

We abandoned the state-> #1542 

> OK. I think this means that the simulator returns state data in a unique_ptr. Ownership then passes to the caller at that point. The the state can be passed back to the simulator. No deep copies since the unique_ptr can use a swap to avoid it.
> 
> We discussed letting the user force a deep copy. Will that be a different PR? Or are we abandoning that for now?
> 
> [Ed: abandon the move/swap of the unique_ptr for now.]

Yea, we abandon the move/swap approach proposed in the closed PR #1542.

runtime/nvqir/CircuitSimulator.h
1tnguyen: We don't free the `state` in this call. It is passed as `const cudaq::SimulationState &` and the simulator will make its copy of the state. The lifetime is managed by the `shared_ptr` data member of `cudaq::state`.

runtime/nvqir/CircuitSimulator.h
1tnguyen: Sure, we can do that. I made it `=0` mainly for my benefit to know where I need to implement it. 
I'll update the code.

runtime/cudaq/qis/execution_manager.h
1tnguyen: Good idea. Added in https://github.com/NVIDIA/cuda-quantum/pull/1582/commits/97c8317595e1c473b4411a9a5338b62eb46bd8f1

runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
1tnguyen: No, currently we don't group them back in any way. These calls will flow to simulator backends as `addQubitsToState` calls to grow the state. 
We preserve the order of `qvector` allocation while supporting batching of default qvector allocations, e.g.

```
qvector q1(2); // <- qubit id 0, 1 in the simulator
qvector q2(2); // <- qubit id 2, 3 in the simulator
qvector q3(state); // the above allocation requests for a total of 4 qubits in |0000> state were batched -> need to flush before adding some qubits in a specific state, e.g., qubits 4, 5 (if the state is a 2-qubit state).
```
`qvector` holds a `qudits` vector to track these indices as well.

runtime/cudaq/qis/qvector.h
1tnguyen: I'd like to do it as well. Unfortunately, `qvector` is a class template so we need to define the implementation in the header file.

runtime/nvqir/CircuitSimulator.h
schweitzpgi: Either way is ok by me.

runtime/nvqir/CircuitSimulator.h
schweitzpgi: OK. It looked like it was being freed, but I might have been looking in the wrong place.

schweitzpgi (approved):
+1

Issue #1580, opened 2024-04-26
[Python] Incorrect return value with a bool parameter


annagrin:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The code below outputs an incorrect value:

```python
@cudaq.kernel(verbose=True)
def kernel(c: int, b: bool) -> int:
    return c
    
print(kernel(1, True)) # Expected 1
```

**Output**
```
module attributes {quake.mangled_name_map = {__nvqpp__mlirgen__kernel = "__nvqpp__mlirgen__kernel_PyKernelEntryPointRewrite"}} {
  func.func @__nvqpp__mlirgen__kernel(%arg0: i64, %arg1: i1) -> i64 attributes {"cudaq-entrypoint"} {
    %0 = cc.alloca i64
    cc.store %arg0, %0 : !cc.ptr<i64>
    %1 = cc.alloca i1
    cc.store %arg1, %1 : !cc.ptr<i1>
    %2 = cc.load %0 : !cc.ptr<i64>
    return %2 : i64
  }
}

72057594037927936
```

**Note**
Passing two ints or two bools seems to work correctly:

```python
  @cudaq.kernel(verbose=True)
  def kernel(b:bool, b2: bool) -> bool:
      return b
  
  assert kernel(True, False) == True
```

```python
  @cudaq.kernel(verbose=True)
  def kernel(b:int, b2: int) -> int:
      return b
  
  assert kernel(42, 53) == 42
```

### Steps to reproduce the bug

```python
@cudaq.kernel(verbose=True)
def kernel(c: int, b: bool) -> int:
    return c
    
print(kernel(1, True))
```

*Output*

72057594037927936

### Expected behavior

1

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

[...]

PR #1584, opened 2024-04-29
Fix incorrect parameter copy when passing parameters to python kernel functions

annagrin:
### Description

**Problem**
The structure we use for collecting arguments to pass to python is in non-packed form, but we rely on arguments to follow each other immediately (without alignment padding) when we try to find the return argument (last in that structure). This results in incorrect parameter passing to python when alignment padding is inserted (for example, in case of one bool and one int  argument).

**Solution**

Use `genReturnOffsetFunction` to create a `returnOffset` function and call it to compute the return offset for python kernels instead of performing incorrect manual calculations.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1580

### DIFF
--- a/lib/Optimizer/Builder/Factory.cpp
+++ b/lib/Optimizer/Builder/Factory.cpp
@@ -73,7 +73,7 @@ cudaq::cc::StructType factory::buildInvokeStructType(FunctionType funcTy) {
     eleTys.push_back(genBufferType</*isOutput=*/false>(inTy));
   for (auto outTy : funcTy.getResults())
     eleTys.push_back(genBufferType</*isOutput=*/true>(outTy));
-  return cudaq::cc::StructType::get(ctx, eleTys, true);
+  return cudaq::cc::StructType::get(ctx, eleTys /*packed=false*/);
 }
 
 Value factory::packIsArrayAndLengthArray(Location loc,
--- a/lib/Optimizer/Builder/Factory.cpp
+++ b/lib/Optimizer/Builder/Factory.cpp
@@ -73,7 +73,7 @@ cudaq::cc::StructType factory::buildInvokeStructType(FunctionType funcTy) {
     eleTys.push_back(genBufferType</*isOutput=*/false>(inTy));
   for (auto outTy : funcTy.getResults())
     eleTys.push_back(genBufferType</*isOutput=*/true>(outTy));
-  return cudaq::cc::StructType::get(ctx, eleTys /*packed=false*/);
+  return cudaq::cc::StructType::get(ctx, eleTys, true);
 }
 
 Value factory::packIsArrayAndLengthArray(Location loc,
--- a/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
+++ b/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
@@ -48,7 +48,7 @@ using PyStateVectorStorage = std::map<std::string, PyStateVectorData>;
 static std::unique_ptr<PyStateVectorStorage> stateStorage =
     std::make_unique<PyStateVectorStorage>();
 
-std::tuple<ExecutionEngine *, void *, std::size_t>
+std::tuple<ExecutionEngine *, void *, std::size_t, std::size_t>
 jitAndCreateArgs(const std::string &name, MlirModule module,
                  cudaq::OpaqueArguments &runtimeArgs,
                  const std::vector<std::string> &names, Type returnType) {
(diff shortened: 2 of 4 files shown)


lib/Optimizer/Builder/Factory.cpp
```diff
@@ -73,7 +73,7 @@ cudaq::cc::StructType factory::buildInvokeStructType(FunctionType funcTy) {
     eleTys.push_back(genBufferType</*isOutput=*/false>(inTy));
   for (auto outTy : funcTy.getResults())
     eleTys.push_back(genBufferType</*isOutput=*/true>(outTy));
-  return cudaq::cc::StructType::get(ctx, eleTys /*packed=false*/);
+  return cudaq::cc::StructType::get(ctx, eleTys, true);
```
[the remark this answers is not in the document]
annagrin: Here is my current understanding (please correct me if I missed something):

It is used to create a structure that contains all the arguments in `GenerateKernelExecution` : https://github.com/NVIDIA/cuda-quantum/blob/e35041833e1ea81c5ae57ea9a4fa798e7870af45/lib/Optimizer/Transforms/GenKernelExecution.cpp#L1428, and we copy the python args into it I think in `packArgs`:  https://github.com/NVIDIA/cuda-quantum/blob/e35041833e1ea81c5ae57ea9a4fa798e7870af45/python/utils/OpaqueArguments.h#L188

After the execution, we need to read the return value so we calculate where it is located in `rawArgs` (which is the structure in question as a pointer to data) in `pyAltLaunchKernelR`, assuming we know the length of all the args. This breaks with alignment padding of course: https://github.com/NVIDIA/cuda-quantum/blob/e35041833e1ea81c5ae57ea9a4fa798e7870af45/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp#L269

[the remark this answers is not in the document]
annagrin: Changed to calling `genReturnOffsetFunction` instead, thank you!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -183,14 +183,26 @@ jitAndCreateArgs(const std::string &name, MlirModule module,
...
+  std::size_t returnOffset = 0;
+  if (runtimeArgs.size()) {
+    auto expectedPtr = jit->lookup(name + ".returnOffset");
+    if (!expectedPtr) {
+      throw std::runtime_error(
+          "cudaq::builder failed to get returnOffset function.");
+    }
+    auto returnOffsetCreator =
+        reinterpret_cast<std::size_t (*)()>(*expectedPtr);
+    returnOffset = returnOffsetCreator();
```
schweitzpgi, line 196: Do you want to check for NoReturnOffset being returned here?

annagrin: Do you know what is returned in the case or error?

annagrin: Got it, updated!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -36,6 +36,8 @@ namespace py = pybind11;
 using namespace mlir;
 
 namespace cudaq {
+static constexpr std::int32_t NoResultOffset =
```
schweitzpgi, line 39:
```suggestion
// TODO: unify with the definition in GenKernelExec.cpp
static constexpr std::int32_t NoResultOffset =
```

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -183,14 +185,29 @@ jitAndCreateArgs(const std::string &name, MlirModule module,
...
+          "cudaq::builder failed to get returnOffset function.");
+    }
+    auto returnOffsetCalculator =
+        reinterpret_cast<std::int64_t (*)()>(*expectedPtr);
+    returnOffset = (std::int32_t)returnOffsetCalculator();
+    if (returnOffset == NoResultOffset) {
+      returnOffset = 0;
+    }
+  }
+  return std::make_tuple(jit, rawArgs, size, returnOffset);
```
schweitzpgi, line 203:
```suggestion
  return {jit, rawArgs, size, returnOffset};
```
should work too.

schweitzpgi (approved):
LGTM. Thanks for hooking this up to the new function, Anna.

PR #1585, opened 2024-04-29
[C++ bridge] Add conversion of pair and tuple to cc.struct.

### LINKED ISSUE #644: Add support for `std::pair` and `std::tuple` when composed of arithmetic types to the bridge
Explanation by example, we want to be able to generate code for
```c++
  std::pair<int, double> p;
  std::tuple<double, float, short> t;
```

On the other hand, types like the following should raise a clang error.
```c++
  std::pair<int*, void(*)()> invalid1;
  std::tuple<double, int, char*> invalid2;
```

schweitzpgi:
Allows std::pair and std::tuple to be passed as arguments to kernels.

Closes #644


khalatepradnya (approved):
Looks good to me!

PR #1588, opened 2024-04-30
[test] Add missing test for #1557:

fabianmcg:
This patch adds a test for the change introduced in #1557.
This test verifies that `cudaq-translate --emit-llvm=false` works as intended and produces MLIR instead of LLVM IR.


schweitzpgi:
Was this failure in a Release mode build?

PR #1596, opened 2024-05-02
[Core] Add a function to determine the offset of the return value in the

schweitzpgi:
buffer when calling altLaunchKernel().

- update tests


annagrin (approved):
Looks perfect, thanks!

PR #1605, opened 2024-05-02
Cherry pick complex numbers support into main

annagrin:
### Description

Cherry-pick changes from https://github.com/NVIDIA/cuda-quantum/commit/a89e0d2f535a256603e3be0f420d9e89087fcc9f to main.

Note: I had to disable a couple of failing tests - not sure what is missing in main that causes them to fail.


python/tests/kernel/test_kernel_float.py
```diff
@@ -0,0 +1,238 @@
...
+
+import sys
+
+import pytest
+import numpy as np
+
+import cudaq
+
+## [PYTHON_VERSION_FIX]
+skipIfPythonLessThan39 = pytest.mark.skipif(
```
khalatepradnya, line 17: Ditto

khalatepradnya (commented):
Q: Is the support for complex numbers already available in the builder mode?

annagrin:
> Q: Is the support for complex numbers already available in the builder mode?

I think so - that is just using regular python.

annagrin:
I am stashing this PR for now, got too hard with the failures in main to investigate. We have a workaround of using kernel builder for now.

annagrin:
Reopened for failure investigation

schweitzpgi (commented):
Skimming through this and it LGTM. Tests are still chugging though.

@khalatepradnya does this get you unblocked?

khalatepradnya:
> Skimming through this and it LGTM. Tests are still chugging though.
> 
> @khalatepradnya does this get you unblocked?

Yes, it does. (I need one more commit pertaining to `np.exp()`)

schweitzpgi (approved):
Approving to unblock Pradnya.

PR #1606, opened 2024-05-02
[GKE] refactor to share common code

schweitzpgi:
This is a follow up PR to #1604 

Refactor common code so it is easier to maintain.


annagrin (approved):
LGTM!

PR #1607, opened 2024-05-03
A few perf. improvements for state handling

1tnguyen:
### Description

- Make sure that we don't actually allocate state in tracer mode (e.g., in library mode to detect mid-circuit measurements).

- Skip state vector validation by default.

- Add some more timing traces to custatevec backend for future analysis.


runtime/nvqir/CircuitSimulator.h
```diff
@@ -807,7 +807,8 @@ class CircuitSimulatorBase : public CircuitSimulator {
   std::size_t allocateQubit() override {
     // Get a new qubit index
     auto newIdx = tracker.getNextIndex();
-
+    const bool inTraceMode =
+        (executionContext && executionContext->name == "tracer");
```
schweitzpgi, line 811: Looks like you could make this a small `static` function in this file.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -298,9 +298,14 @@ protected:
       // FIXME handle case where pointer is a device pointer
 
       // First allocation, so just set the user provided data here
-      HANDLE_CUDA_ERROR(cudaMemcpy(deviceStateVector, state,
-                                   stateDimension * sizeof(CudaDataType),
-                                   cudaMemcpyHostToDevice));
+      {
```
schweitzpgi, line 301: Do you really need this block (`{ ... }`)?

runtime/nvqir/CircuitSimulator.h
1tnguyen: Indeed, fixed in https://github.com/NVIDIA/cuda-quantum/pull/1607/commits/96b1112502e95b34cec0c9271b05ffe4035667cf

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
1tnguyen: No, we don't. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1607/commits/96b1112502e95b34cec0c9271b05ffe4035667cf

runtime/cudaq/qis/qvector.h
```diff
@@ -60,6 +61,7 @@ class qvector {
                          : simulation_precision::fp64;
     getExecutionManager()->initializeState(targets, vector.data(), precision);
   }
+  qvector(const std::vector<complex> &vector) : qvector(vector, false){};
```
schweitzpgi, line 64:
```suggestion
  qvector(const std::vector<complex> &vector) : qvector(vector, /*validate=*/false){};
```

schweitzpgi (approved):
LGTM. (Small nit.)

Issue #1608, opened 2024-05-03
[Python] Support returning complex numbers from python kernels


annagrin:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

Support returning complex numbers from python kernel functions, for example:

```python
  @cudaq.kernel
  def kernel(c: complex) -> complex:
      return c

  assert kernel(1 + 2j) == 1 + 2j
```

annagrin:
Added in https://github.com/NVIDIA/cuda-quantum/pull/1610

PR #1610, opened 2024-05-03
[python] Support returning complex numbers from python

annagrin:
### Description

Add the following features to Python kernel functions:

- Support for returning `complex`, `numpy.complex128`, `numpy.complex64`, `numpy.float64`, `numpy.float32` numbers.
- Handle returning arguments from an inner scope.
- Allow casting of list elements to complex and float types on parameter passing.
- Add tests.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1608


lib/Optimizer/Builder/Factory.cpp
```diff
@@ -73,7 +73,7 @@ cudaq::cc::StructType factory::buildInvokeStructType(FunctionType funcTy) {
     eleTys.push_back(genBufferType</*isOutput=*/false>(inTy));
   for (auto outTy : funcTy.getResults())
     eleTys.push_back(genBufferType</*isOutput=*/true>(outTy));
-  return cudaq::cc::StructType::get(ctx, eleTys /*packed=false*/);
+  return cudaq::cc::StructType::get(ctx, eleTys /*isPacked = false*/);
```
schweitzpgi, line 76:
```suggestion
  return cudaq::cc::StructType::get(ctx, eleTys);
```
I think we're ok to just remove this comment altogether here. Typically, the convention is to use `/*flagName=*/true` when a `true` or `false` is required. Just having magic values appear as arguments is discouraged.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -171,6 +173,40 @@ def __str__(self):
...
+    def isCastable(self, fromTy, toTy):
+        if F64Type.isinstance(toTy):
+            return F32Type.isinstance(fromTy) or IntegerType.isinstance(fromTy)
+
+        if F32Type.isinstance(toTy):
+            return F64Type.isinstance(fromTy) or IntegerType.isinstance(fromTy)
+
+        if ComplexType.isinstance(toTy):
+            floatToType = ComplexType(toTy).element_type
+            if ComplexType.isinstance(fromTy):
```
schweitzpgi, line 185: Are we allowed to have complex of integral types here?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -149,6 +150,14 @@ jitAndCreateArgs(const std::string &name, MlirModule module,
             delete static_cast<long *>(ptr);
           });
         })
+        .Case([&](ComplexType type) {
+          Py_complex *ourAllocatedArg = new Py_complex();
+          ourAllocatedArg->real = 0.0;
+          ourAllocatedArg->imag = 0.0;
```
schweitzpgi, line 156: Does python return a zero value by default?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -308,17 +317,38 @@ py::object pyAltLaunchKernelR(const std::string &name, MlirModule module,
         std::free(rawArgs);
         return py::int_(concrete);
       })
+      .Case([&](ComplexType ty) -> py::object {
+        auto eleTy = ty.getElementType();
+        return llvm::TypeSwitch<mlir::Type, py::object>(eleTy)
+            .Case([&](Float64Type eTy) -> py::object {
+              std::complex<double> concrete;
+              std::memcpy(&concrete, ((char *)rawArgs) + returnOffset, 16);
```
schweitzpgi, line 325: You could share some code here. The FloatXXType classes have a `getIntOrFloatBitWidth()` member function. There is also a `cudaq::opt::convertBitsToBytes()` function. So,
```c++
cudaq::opt::convertBitsToBytes(someTy.getIntOrFloatBitWidth())`
```
would be fully generic here.

python/utils/OpaqueArguments.h
```diff
@@ -17,7 +17,10 @@
 #include "mlir/CAPI/IR.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
```
schweitzpgi, line 19:
```suggestion
```
nit: blank lines

python/utils/PyTypes.h
```diff
@@ -0,0 +1,168 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include <complex>
+#include <pybind11/complex.h>
+#include <pybind11/pybind11.h>
+
```
schweitzpgi, line 14:
```suggestion
```

python/utils/PyTypes.h
```diff
@@ -0,0 +1,168 @@
...
+#include <iostream>
+
+namespace py = pybind11;
```
schweitzpgi, line 17: Should omit namespace aliases and especially `using namespace ship;` from headers.

python/utils/PyTypes.h
```diff
@@ -0,0 +1,168 @@
...
+  }
+};
+
+template <typename T>
+inline std::string typeName() {
+  return {typeid(T).name()};
+}
+template <>
+inline std::string typeName<py_ext::Float>() {
+  return "float";
```
schweitzpgi, line 135: Curious: do these functions have to return `std::string` objects? Or can they just as easily return `char const *` values that may (or may not) need to be copied to the heap in the calling code?

schweitzpgi (commented):
Looking good, Anna. A few suggestions and minor nits.

python/utils/PyTypes.h
annagrin: They definitely can return char*, will change!

python/utils/PyTypes.h
annagrin: will do, thanks!

python/utils/OpaqueArguments.h
```diff
@@ -13,11 +13,13 @@
 #include "cudaq/builder/kernel_builder.h"
 #include "cudaq/qis/pauli_word.h"
```
schweitzpgi, line 15: nit: remove blank lines between `#include` lines.

python/utils/OpaqueArguments.h
```diff
@@ -156,47 +182,33 @@ packArgs(OpaqueArguments &argData, py::args args,
...
             });
           } else {
-            *ourAllocatedArg = std::complex<float>(
-                arg.attr("real").cast<float>(), arg.attr("imag").cast<float>());
+            std::complex<float> *ourAllocatedArg = new std::complex<float>();
+            *ourAllocatedArg = arg.cast<std::complex<float>>();
 
             argData.emplace_back(ourAllocatedArg, [](void *ptr) {
               delete static_cast<std::complex<float> *>(ptr);
             });
```
schweitzpgi, line 199: I wonder if there is a simpler way to do this pattern with the deallocations. It just seems really repetitive here...

schweitzpgi (approved):
Looks good. Thanks, Anna.

python/utils/OpaqueArguments.h
annagrin: I'll think of something... could the fix be in a different pr?

python/utils/OpaqueArguments.h
```diff
@@ -156,115 +186,63 @@ packArgs(OpaqueArguments &argData, py::args args,
...
+          checkArgumentType<py_ext::Complex>(arg, i);
           if (isa<Float64Type>(ty.getElementType())) {
-            *ourAllocatedArg =
-                std::complex<double>(PyFloat_AsDouble(arg.attr("real").ptr()),
-                                     PyFloat_AsDouble(arg.attr("imag").ptr()));
-
-            argData.emplace_back(ourAllocatedArg, [](void *ptr) {
-              delete static_cast<std::complex<double> *>(ptr);
-            });
+            auto *ourAllocatedArg =
```
amccaskey, line 191: since you've created `addArgument<T>` can't you merge the `ourAllocatedArg` allocation into that new function as well and just have a one-liner (by also taking the `arg` as input)?

annagrin: There are differences in how we initialize the argument - sometimes we need to cast it but sometimes not... Let me see if I can add two functions here

annagrin: Done!

PR #1612, opened 2024-05-06
[Core] Completes making quantum values a linear type.

schweitzpgi:
A quantum type (!quake.wire) is a linear type. While this constraint was widely enforced previously, it was exempted when the linear value reached a control-flow conditional. This patch closes that loophole.

- Add a new Linear type trait to be used with high-level conditionals that are conceptually conditional control-flow. (cc.if)
- Modify cc.if to use the linear type trait. Implement parser and pretty printer, to support the sigma arguments on the blocks.
- Add roundtrip test.
- Fix check lines in existing tests.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCOps.td
+++ b/include/cudaq/Optimizer/Dialect/CC/CCOps.td
@@ -12,7 +12,9 @@
 include "mlir/Interfaces/InferTypeOpInterface.td"
 include "cudaq/Optimizer/Dialect/Common/Traits.td"
 include "cudaq/Optimizer/Dialect/CC/CCDialect.td"
+include "cudaq/Optimizer/Dialect/CC/CCInterfaces.td"
 include "cudaq/Optimizer/Dialect/CC/CCTypes.td"
+include "cudaq/Optimizer/Dialect/Quake/QuakeTypes.td"
 
 def AnyPointerType : Type<CPred<"$_self.isa<mlir::LLVM::LLVMPointerType,"
                           "cudaq::cc::PointerType>()">, "any pointer type">;
@@ -340,7 +342,7 @@ def cc_IfOp : CCOp<"if",
     [DeclareOpInterfaceMethods<RegionBranchOpInterface,
                                ["getNumRegionInvocations",
                                 "getRegionInvocationBounds"]>,
-     RecursiveMemoryEffects, NoRegionArguments]> {
+     RecursiveMemoryEffects, LinearTypeArgsTrait]> {
   let summary = "if-then-else operation";
   let description = [{
     An IfOp is a C-like if statement that supports single-entry, multiple-exit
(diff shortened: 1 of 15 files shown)


include/cudaq/Optimizer/Dialect/CC/CCOps.td
```diff
@@ -388,6 +396,23 @@ def cc_IfOp : CCOp<"if",
         mlir::Location, mlir::Region &)>;
 
     bool hasResults() { return getResults().size(); }
+
+    // A cc.if must have a valid then region.
+    bool hasThen() { return !getThenRegion().empty(); }
+    mlir::Block *getThenEntryBlock() { return &getThenRegion().front(); }
+    mlir::Block::BlockArgListType getThenEntryArguments() {
+      return getThenEntryBlock()->getArguments();
+    }
```
bmhowe23, line 405: Will this handle a case like this:
```cpp
if (condition) {}
else {}
```
It seems like line 404 might segfault, but I might be missing something. (Lines 407-415 seem to handle this correctly/differently for the `else` block.)

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -1143,16 +1151,43 @@ ParseResult cudaq::cc::IfOp::parse(OpAsmParser &parser,
...
+    }
+  }
   if (parser.parseOptionalArrowTypeList(result.types))
     return failure();
-  if (parser.parseRegion(*thenRegion, /*arguments=*/{}, /*argTypes=*/{}))
+  if (!regionArgs.empty()) {
+    // Check that the result.types is compatible with the regionArgs. To be
+    // compatible, it must have at least as many linear types as there are
+    // region arguments. (It can have more.)
+    std::int64_t numRegionArgs = regionArgs.size();
```
bmhowe23, line 1175: I'm surprised this doesn't cause a warning with implicit casting unsigned to signed. 🤷‍♂️

include/cudaq/Optimizer/Dialect/CC/CCOps.td
schweitzpgi: The `then` region is required to be non-empty. So it has a block and the block has a terminator. For the example given, a bridge could generate a similar `else` region as well or omit it.

At that point, it is up to the optimizer to discover, "hey, this if statement really doesn't do anything" and erase it.

A segfault isn't possible here because the block is required and a block has a list of arguments which is 0 or more BlockArguments in length.

lib/Optimizer/Dialect/CC/CCOps.cpp
schweitzpgi: Fortunately, it doesn't. I think warnings like that come into play with truncation though, so assigning an `unsigned long` to an `int` would generate a warning.

bmhowe23: Some `unsigned long` elements cannot be mapped to `std::int_t` elements (i.e. 2^63). That being said, I'm not going to lose any sleep at night about us having >2^63 elements in `regionArgs`, so I'll mark this as resolved. 😆

bmhowe23 (approved):

Issue #1618, opened 2024-05-07
`MidCircuitMeasurementAnalyzer` doesn't handle all cases of measurement assigned variables 


1tnguyen:
There is an oversight in this line:

https://github.com/NVIDIA/cuda-quantum/blob/035bd3ba0b2a5e6b90b23c2bb6dfdcdc3d7bf40c/python/cudaq/kernel/analysis.py#L49

There are two cases:

- For simple variable comparison (`var == True/False`): the `id` field is actually in the `left` node, i.e., we need to look for `condition.left.id`.

- For subscript variables, e.g., `results[0]`, the `left` is an `ast.Subscript` object, hence we need to look for the `value` field to find the variable name.

PR #1619, opened 2024-05-07
Fixes for https://github.com/NVIDIA/cuda-quantum/issues/1618

1tnguyen:
### Description

The current Python `MidCircuitMeasurementAnalyzer` visitor only supports certain expressions, e.g., `if val` but not `if val == boolVar`, where `val = mz(q)`.

Hence, it may mis-identify some kernels as non-measurement feedback, i.e., not setting `qubitMeasurementFeedback=true`, causing wrong sampling mode being used.

This PR:

- Add more handling cases: variable name or variable subscript; `CompareOp`; handle variables in `BoolOp` and `UnaryOp`.

- Add IR checks and target tests.

Note: this is not an exhaustive handling of all possible scenarios.

Resolves https://github.com/NVIDIA/cuda-quantum/issues/1618

### DIFF
--- a/python/cudaq/kernel/analysis.py
+++ b/python/cudaq/kernel/analysis.py
@@ -38,24 +38,45 @@ class MidCircuitMeasurementAnalyzer(ast.NodeVisitor):
         ]:
             self.measureResultsVars.append(target.id)
 
+    # Get the variable name from a variable node.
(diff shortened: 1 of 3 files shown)


bmhowe23 (approved):
+1

schweitzpgi:
IMO, this analysis shouldn't really be done on the AST.

PR #1621, opened 2024-05-07
[docs] Remove tensornet-mps limitation that is no longer accurate

bmhowe23:
This warning is no longer accurate since #1420 was merged.


khalatepradnya:
I did get this warning (albeit a on developer branch, but aligned with `main`) - `RuntimeError: MPS simulator: Gates on 3 or more qubits are unsupported.`

1tnguyen:
@khalatepradnya Do you have a specific case for that? 
It could be that the `applyGate` method of `SimulatorMPS` was called directly. `SimulatorMPS` decomposes all multi-controlled runtime gates before pushing them to the queue.

khalatepradnya:
> @khalatepradnya Do you have a specific case for that? It could be that the `applyGate` method of `SimulatorMPS` was called directly. `SimulatorMPS` decomposes all multi-controlled runtime gates before pushing them to the queue.

Yes, `applyGate` was called directly in my case.

1tnguyen (approved):

1tnguyen:
> > @khalatepradnya Do you have a specific case for that? It could be that the `applyGate` method of `SimulatorMPS` was called directly. `SimulatorMPS` decomposes all multi-controlled runtime gates before pushing them to the queue.
> 
> Yes, `applyGate` was called directly in my case.

We could sync up offline about this use case. With cuquantum 24.03 update, MPS would handle control gates a bit differently (https://github.com/NVIDIA/cuda-quantum/pull/1524). If there is a gap to support unitary synthesis (e.g., `cutensornetStateApplyControlledTensorOperator` only supports single target), we'll find a solution for it :-)

PR #1632, opened 2024-05-08
Support results polling interval override for RemoteRESTQPUs

kukushechkin:
CUDA Quantum runtime polls for a `RemoteRESTQPU` results quite aggressively, 100 microseconds is the hardcoded value. While this is ok for a QPU HTTP API located in the local network, when executing CUDA Quantum programs against a QPU HTTP API located in a cloud, it is very likely to hit a rate-limiter of sorts, like, for example, in case of IQM Resonance.

### DIFF
--- a/runtime/common/Future.cpp
+++ b/runtime/common/Future.cpp
@@ -34,7 +34,8 @@ sample_result future::get() {
     cudaq::info("Future got job retrieval path as {}.", jobGetPath);
     auto resultResponse = client.get(jobGetPath, "", headers);
     while (!serverHelper->jobIsDone(resultResponse)) {
-      std::this_thread::sleep_for(std::chrono::microseconds(100));
+      auto polling_interval = serverHelper->nextResultPolingInterval(resultResponse);
+      std::this_thread::sleep_for(polling_interval);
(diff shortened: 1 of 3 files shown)


bmhowe23:
Hi @kukushechkin - thanks for the contribution! You can run `bash scripts/run_clang_format.sh` to get your code formatted in such a way that will make the CI happy.

bmhowe23 (approved):

runtime/common/Future.cpp
```diff
@@ -34,7 +34,9 @@ sample_result future::get() {
     cudaq::info("Future got job retrieval path as {}.", jobGetPath);
     auto resultResponse = client.get(jobGetPath, "", headers);
     while (!serverHelper->jobIsDone(resultResponse)) {
-      std::this_thread::sleep_for(std::chrono::microseconds(100));
+      auto polling_interval =
+          serverHelper->nextResultPolingInterval(resultResponse);
```
schweitzpgi, line 38: Should be `Polling` rather than `Poling` throughout.

bmhowe23: Thanks for catching that. https://github.com/NVIDIA/cuda-quantum/pull/1644 posted.

PR #1637, opened 2024-05-08
IQM Aphrodite QPU support

kukushechkin:
Connectivity mapping for IQM 54qubits QPU.


bmhowe23 (approved):
Thanks, @kukushechkin!

PR #1645, opened 2024-05-08
Pin ARM's `-march` flag for OpenMPI build

1tnguyen:
### Description

This PR pins the `-march=` compiler flag for OpenMPI build on ARM (similar to the approach taken for x86).
Currently, we're using `-march=native`, which is too broad, hence prone to runner's hardware update.

ARMv8.0-A was chosen as it's the first one in the v8 series (with 64-bit support).

### Issue & Root cause 

Illegal instruction exception when using cuda-q docker container on some specific ARM platforms, e.g., failed on ARM Neoverse-N1, passed on Grace Hopper.

The illegal instruction was ARM's `STLUR`, an ARMv8.4 instruction.

The CI runner, `linux-arm64-cpu8`, is an AWS's `m7g.2xlarge` instance, which has a Graviton3 (ARMv8.5) processor.

**Notes:**

- CUDA-Q's v0.6 image doesn't have `STLUR` in the OpenMPI's binary. We might have switched the runner hardware some time in between that release and now.
 
**Tested by:** 

Using a temp pipeline job (in https://github.com/NVIDIA/cuda-quantum/pull/1645/commits/2309b753db68712716185742535519397fe9c5f4, has been reverted) to build the OpenMPI image with the new flag and copy over the binary to verify the illegal instruction is now fixed (on an Neoverse-N1, ARMv8.2, machine)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,22 @@ jobs:
           echo "llvm_commit=$(git rev-parse @:./tpls/llvm)" >> $GITHUB_OUTPUT
           echo "pybind11_commit=$(git rev-parse @:./tpls/pybind11)" >> $GITHUB_OUTPUT
 
+##   DEBUG ####
+  openmpi:
+    name: Build Open MPI
+    strategy:
+      matrix:
+        platform: [arm64]
+      fail-fast: false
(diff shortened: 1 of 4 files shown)


bmhowe23 (approved):
Nice find, Thien.

Issue #1622, opened 2024-05-07
reset of qubit in kernel mode is giving error 


marwafar:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When running this code snippet from the [doc](https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/specification/cudaq/algorithmic_primitives.rst)

``` 
@cudaq.kernel()
def kernel():
   q = cudaq.qubit()
   h(q)
   reg1 = mz(q)
   reset(q)
   x(q)

cudaq.sample(kernel).dump()
```

It gives this error

`cudaq.kernel.ast_bridge.CompilerError: reset.py:8: error: unhandled function call - reset, known kernels are dict_keys([])
         (offending source -> reset(q))`

However, when commenting reset, it works and give the expected answer.

### Steps to reproduce the bug

``` 
@cudaq.kernel()
def kernel():
   q = cudaq.qubit()
   h(q)
   reg1 = mz(q)
   reset(q)
   x(q)

cudaq.sample(kernel).dump()
```

### Expected behavior

It should reset the qubit and print 
{ 
  __global__ : { 1:1000 }
  reg1 : { 0:501 1:499 }
}

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: latest (May 6)
- **Python version**: python 3
- **C++ compiler**: 
- **Operating system**:

PR #1649, opened 2024-05-09
[Fix] [Python] Enable 'reset' in Python kernel mode

khalatepradnya:
### Description
Addresses issue captured in #1622

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1471,21 +1471,26 @@ class PyASTBridge(ast.NodeVisitor):
 
             if node.func.id == 'reset':
                 target = self.popValue()
-                # single qubit
-                if not quake.VeqType.isinstance(target.type):
+                if quake.RefType.isinstance(target.type):
                     quake.ResetOp([], target)
                     return
-                # target is a VeqType
-                def bodyBuilder(iterVal):
-                    q = quake.ExtractRefOp(self.getRefType(),
-                                           target,
-                                           -1, # kDynamicIndex
-                                           index=iterVal).result
-                    quake.ResetOp([], q)
+                if quake.VeqType.isinstance(target.type):
 
-                veqSize = quake.VeqSizeOp(self.getIntegerType(), target).result
-                self.createInvariantForLoop(veqSize, bodyBuilder)
-                return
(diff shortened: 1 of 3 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -1469,6 +1469,24 @@ def bodyBuilder(iterVal):
...
+                target = self.popValue()
+                # single qubit
+                if not quake.VeqType.isinstance(target.type):
+                    quake.ResetOp([], target)
+                    return
+                # target is a VeqType
+                def bodyBuilder(iterVal):
+                    q = quake.ExtractRefOp(self.getRefType(),
+                                           target,
+                                           -1,
```
schweitzpgi, line 1482: This magic value looks a bit odd. I believe it corresponds to `quake::ExtractOp::kDynamicIndex`, which isn't `-1` but has the same bits as `-1`.
```suggestion
                                           -1, # kDynamicIndex
```

On the C++ side, we have both a constructor for using a constant index and another for using an SSA-value as an index. The internal encoding isn't exposed in this way. It might make sense to do something similar in python.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1469,6 +1469,24 @@ def bodyBuilder(iterVal):
                 opCtor([], [], [], [qubitA, qubitB])
                 return
 
+            if node.func.id == 'reset':
+                target = self.popValue()
+                # single qubit
+                if not quake.VeqType.isinstance(target.type):
```
schweitzpgi, line 1475: Is there a reason this isn't testing for a `RefType` explicitly? It has to be either a `RefType` or a `VeqType` at this point, eh? While `WireType` is allowed in the IR, I hope we don't want to allow that in the bridge.

schweitzpgi (commented):
LGTM, Pradnya.

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Thanks... I had considered only `RefType` and `VeqType` as the two mutually exclusive allowed types. 
Updated to explicitly check for these two types.

schweitzpgi (approved):
+1

PR #1652, opened 2024-05-10
[DON'T MERGE][Reference] Dockerfile to build MPICH compatible image

1tnguyen:
### Description

**This is for reference only.**

Target use case: HPC data center whereby an image with MPICH-compatible libraries are needed. 

Instruction: build this Dockerfile against the CUDA-Q's `base_image` of interest, default to the nightly image.
MPICH version can be changed by `mpich` Docker build argument. Default is 4.1.1.


1tnguyen:
This is generally not needed anymore with the CUDA-Q MPI plugin infrastructure.

Issue #1620, opened 2024-05-07
[Python Kernel] `bool` to `int` type casting issue


1tnguyen:
The below kernel

```
@cudaq.kernel
def kernel():
    data = cudaq.qvector(2)
    aux = cudaq.qubit()
    x(aux)
    bit = mz(aux)
    # Compare to 1
    if bit == 1:
        x(data[0])
    mz(data)
```
 results in the following IR:
 
 ```
  %6 = bitcast %Result* %5 to i1*
  %7 = load i1, i1* %6, align 1
  %8 = sext i1 %7 to i64
  %9 = icmp eq i64 %8, 1
```

If `bit==True` (`%7 == true` in the above IR), it looks like the `icmp` returns `false`. A quick search of LLVM doc indicated that `sext i1 ... i64` may convert `true` to -1.

In fact, the below kernel resulted in the correct distribution (may need the fixes from https://github.com/NVIDIA/cuda-quantum/pull/1619 to get the measurement feedback detection working properly).

```
@cudaq.kernel
def kernel5():
    data = cudaq.qvector(2)
    aux = cudaq.qubit()
    x(aux)
    bit = mz(aux)
    # Compare to -1
    if bit == -1:
        x(data[0])
    mz(data)
```

1tnguyen:
@schweitzpgi suggested that we should use `cc::CastOpMode::Unsigned` in this case.

PR #1654, opened 2024-05-10
[Python AST] Always perform unsigned extension for i1 (bool) type when doing integer compare

1tnguyen:
### Description

When we need to check `if var == 1`, where `var` is a bool type and 1 is an integer type, we need to do an unsigned extension of i1 since a signed extension of bool True will result in -1 rather than 1.
 
Also, add the type casting logic to `NotEq` as well since it should be equivalent to `Eq`. 

Resolve: https://github.com/NVIDIA/cuda-quantum/issues/1620


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2720,6 +2720,16 @@ def visit_Compare(self, node):
             return
 
         if isinstance(op, ast.NotEq):
+            if F64Type.isinstance(left.type) and IntegerType.isinstance(
+                    comparator.type):
+                left = arith.FPToSIOp(comparator.type, left).result
```
schweitzpgi, line 2725: I'm not following why the LHS, if it's a floating point value, is integer truncated here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2720,6 +2720,16 @@ def visit_Compare(self, node):
...
+            if IntegerType(left.type).width < IntegerType(
+                    comparator.type).width:
+                if IntegerType(left.type).width == 1:
+                    # For i1 (bool), always performs an unsigned extension.
+                    left = arith.ExtUIOp(comparator.type, left).result
+                else:
+                    left = arith.ExtSIOp(comparator.type, left).result
```
schweitzpgi, line 2732: Using `cc.CastOp` will simplify things potentially. It's one operation that can do sign-extension, zero-extension, truncation, from/to floating-point conversions, etc.

1tnguyen: Done in https://github.com/NVIDIA/cuda-quantum/pull/1654/commits/130ead60f4035e7cd1548164d71cb733434bb4b1
Agreed, this would also make the IR printing better, e.g., `cc.cast unsigned` is more readable than before.

python/cudaq/kernel/ast_bridge.py
1tnguyen: This is existing code from the `isinstance(op, ast.Eq)` below, which, in principle, should also be handled in the `ast.NotEq` case here.

I don't know the context of it so didn't change it or migrate it to `cc.CastOp` yet. 
@khalatepradnya Do you know the intent of this float -> int conversion in equal check?

khalatepradnya: nope, sorry. Tagging @amccaskey for details.
I see that the `comparator` is of integer type and that the integer comparator is being used (`arith.cmpi`) on line#2740

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2720,6 +2720,22 @@ def visit_Compare(self, node):
...
+                    # For i1 (`bool`), always performs an unsigned extension.
+                    left = cc.CastOp(comparator.type,
+                                     left,
+                                     sint=False,
+                                     zint=True).result
+                else:
+                    left = cc.CastOp(comparator.type,
+                                     left,
+                                     sint=True,
+                                     zint=False).result
```
schweitzpgi, line 2738:
```suggestion
                zeroext = IntegerType(left.type).width == 1
                left = cc.CastOp(comparator.type,
                                     left,
                                     sint=not zeroext,
                                     zint=zeroext).result
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2731,7 +2747,17 @@ def visit_Compare(self, node):
                 left = arith.FPToSIOp(comparator.type, left).result
             if IntegerType(left.type).width < IntegerType(
                     comparator.type).width:
-                left = arith.ExtSIOp(comparator.type, left).result
+                if IntegerType(left.type).width == 1:
```
schweitzpgi, line 2750: similar to above

schweitzpgi (approved):
LGTM

python/cudaq/kernel/ast_bridge.py
1tnguyen: Applied the same change as recommended (5e1bc36df1f8a3000b32157fd723872f4fff05c2)

python/cudaq/kernel/ast_bridge.py
schweitzpgi: I'm wondering if this isn't a bug. Consider this python code
```python
>>> a=1
>>> b=1.2
>>> c= b==a
>>> print(c)
False
```

PR #848, opened 2023-11-01
Add a script to validate jupiter notebooks

abhiram6121:
### Description
 Add Jupyter notebooks to CI #726 

Checklist:
- [x] I have added tests to cover my changes.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.

PR #1656, opened 2024-05-10
Tools to enable notebook validation and pipeline fix

bettinaheim:
I only realized now that tee cause the exit code of the command to not be properly reflected unless pipefail is set...

This PR should facilitate future notebook changes, and enable to finally pull in https://github.com/NVIDIA/cuda-quantum/pull/848.


scripts/validate_container.sh
```diff
@@ -245,6 +216,38 @@ do
     echo "============================="
 done
 
+echo "============================="
```
bmhowe23, line 219: I can't tell - did anything change in this snippet, or was it just moved?

bmhowe23: Never mind, I compared using an external diff tool. I see that 🏳️ , ✅ , and  ❌ were added and no other changes were made (other than moving it). Just curious what the reason for moving it was.

bettinaheim: It was moved and I added output to make it also show up in the checklist in the step summary.

bmhowe23 (approved):

scripts/validate_container.sh
bettinaheim: Not much - I added the notebook validation below (separate PR), so I moved the Python part down as well.

PR #1658, opened 2024-05-10
[Core] Cherry pick cc#GlobalOp.

schweitzpgi:
This functionality was stuck on the feature branch but is more generally useful. Cherry pick the diffs and merge with main.


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -41,8 +41,7 @@ cudaq::cc::AddressOfOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
       getParentOfType<ModuleOp>(getOperation()), getGlobalNameAttr());
 
   // TODO: add globals?
```
khalatepradnya, line 43: `TODO` can be removed?

khalatepradnya (commented):
Thank you for these changes! 
Are there any tests to cover the added functionality?

schweitzpgi:
> Thank you for these changes! Are there any tests to cover the added functionality?

In the other branch. But they depend on the qalloc features, which only exist in the branch.

khalatepradnya (approved):

PR #1661, opened 2024-05-10
removing unused map

sacpis:
Removing unused map from the function

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCOps.td
+++ b/include/cudaq/Optimizer/Dialect/CC/CCOps.td
@@ -890,6 +890,34 @@ def cc_AddressOfOp : CCOp<"address_of", [Pure,
   }];
 }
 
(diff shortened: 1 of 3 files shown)


runtime/common/MeasureCounts.cpp
```diff
@@ -102,7 +102,6 @@ void ExecutionResult::deserialize(std::vector<std::size_t> &data) {
       name += std::string(1, char(data[stride + i]));
 
     stride += nChars;
-    std::unordered_map<std::string, std::size_t> localCounts;
```
schweitzpgi, line 105: The code here, `ExecutionResult::deserialize` appears to be mostly a block copy of the code below in `sample_result::deserialize`. It would be nice to combine these into a single shared implementation by way of a template function or something similar.

schweitzpgi (approved):
LGTM

PR #1664, opened 2024-05-11
Refactoring measurecounts

sacpis:
# Refactoring MeasureCounts for efficiency

1. Moved similar code for ExecutionResult and sample_result deserialization in a separate deserialize method as per @schweitzpgi's suggestion
2. Used emplace instead of insert to avoid copies
3. Directly constructed a string into a vector instead of using for loops, which involves repeatedly concatenating strings
4. Used move utility to allow the transferring of resources from one object to another efficiently

# Test result

100% tests passed, 0 tests failed out of 564

Label Time Summary:
gpu_required    = 133.59 sec*proc (312 tests)

Total Test time (real) = 302.33 sec

The following tests did not run:
        515 - iqm-tests (Skipped)

### DIFF
--- a/runtime/common/MeasureCounts.cpp
+++ b/runtime/common/MeasureCounts.cpp
@@ -149,9 +149,12 @@ void sample_result::deserialize(std::vector<std::size_t> &data) {
     deserializeCounts(data, stride, localCounts);
 
     sampleResults.insert({name, ExecutionResult{localCounts, name}});
-    totalShots = std::accumulate(
-        localCounts.begin(), localCounts.end(), 0,
-        [](std::size_t sum, const auto &pair) { return sum + pair.second; });
+
+    if (stride >= data.size()) {
+      totalShots = std::accumulate(
+          localCounts.begin(), localCounts.end(), 0,
+          [](std::size_t sum, const auto &pair) { return sum + pair.second; });
+    }
   }
 }
 


runtime/common/MeasureCounts.cpp
```diff
@@ -40,25 +65,25 @@ ExecutionResult::ExecutionResult(CountsDictionary c, std::string name, double e)
...
 
 ExecutionResult &ExecutionResult::operator=(const ExecutionResult &other) {
-  counts = other.counts;
-  expectationValue = other.expectationValue;
-  registerName = other.registerName;
-  sequentialData = other.sequentialData;
+  counts = std::move(other.counts);
+  expectationValue = std::move(other.expectationValue);
+  registerName = std::move(other.registerName);
+  sequentialData = std::move(other.sequentialData);
```
bmhowe23, line 77: Invoking `std::move() ` inside these constructors is undesirable because the input parameters (`other.*`) for these constructors are `const` and therefore need to retain ownership over their data. In this case, I think the compiler is saving you by _preventing_ the move constructor of the underlying objects from being called, which is a good thing, but it also means that the `std::move` is having no effect. Unless I'm missing something, I think these particular changes should be removed.

runtime/common/MeasureCounts.cpp
```diff
@@ -127,31 +139,19 @@ std::vector<std::size_t> sample_result::serialize() const {
...
+    std::unordered_map<std::string, std::size_t> localCounts;
+    deserializeCounts(data, stride, localCounts);
 
-    stride += nBs * 3;
-    totalShots = localShots;
+    sampleResults.insert({name, ExecutionResult{std::move(localCounts), name}});
+    totalShots += std::accumulate(
+        localCounts.begin(), localCounts.end(), 0,
+        [](std::size_t sum, const auto &pair) { return sum + pair.second; });
+    stride += localCounts.size() * 3;
```
bmhowe23, line 154: It is unsafe to use `localCounts` after invoking `std::move()` on it because it is undefined whether or not it will retain its values after being "moved".

runtime/common/MeasureCounts.cpp
```diff
@@ -88,31 +113,18 @@ std::vector<std::size_t> ExecutionResult::serialize() const {
...
-      auto bitstring_as_long = data[j];
-      auto size_of_bitstring = data[j + 1];
-      auto count = data[j + 2];
-      auto bs = longToBitString(size_of_bitstring, bitstring_as_long);
-      counts.insert({bs, count});
-    }
-    stride += nBs * 3;
+    deserializeCounts(data, stride, localCounts);
+
+    counts = std::move(localCounts);
```
bmhowe23, line 127: Wouldn't `counts` get overwritten each time through the loop here? Is that intentional?

runtime/common/MeasureCounts.cpp
```diff
@@ -88,31 +112,20 @@ std::vector<std::size_t> ExecutionResult::serialize() const {
...
-    for (std::size_t j = stride; j < stride + nBs * 3; j += 3) {
-      auto bitstring_as_long = data[j];
-      auto size_of_bitstring = data[j + 1];
-      auto count = data[j + 2];
-      auto bs = longToBitString(size_of_bitstring, bitstring_as_long);
-      counts.insert({bs, count});
+    deserializeCounts(data, stride, localCounts);
+
+    for (const auto &entry : localCounts) {
+      counts.insert_or_assign(entry.first, entry.second);
```
bmhowe23, line 127: Why the call to `insert_or_assign` here vs just `insert`? It seems like a potential change in behavior. Is that intentional?

runtime/common/MeasureCounts.cpp
```diff
@@ -127,31 +140,18 @@ std::vector<std::size_t> sample_result::serialize() const {
...
-      localShots += count;
-      localCounts.insert({bs, count});
-    }
+    deserializeCounts(data, stride, localCounts);
 
     sampleResults.insert({name, ExecutionResult{localCounts, name}});
-
-    stride += nBs * 3;
-    totalShots = localShots;
+    totalShots += std::accumulate(
```
bmhowe23, line 152: This seems like a change in behavior. In the old code, `totalShots` was assigned based on the last time through the `while` loop whereas in the new code, `totalShots` is the sum over all times through the `while` loop. Is that intentional?

bmhowe23 (approved):
LGTM. Thanks, @sacpis.

PR #1672, opened 2024-05-14
Publishing fixes for single-GPU runner

bmhowe23:
Fix the notebooks that were failing in this run: https://github.com/NVIDIA/cuda-quantum/actions/runs/9068351041/job/24917322153.


docs/sphinx/examples/python/tutorials/multi_gpu_workflows.ipynb
```diff
@@ -339,15 +339,20 @@
...
+    "num_gpus = cudaq.num_available_gpus()\n",
     "\n",
     "for i in range(len(xi)):\n",
     "    for j in range(xi[i].shape[0]):\n",
+    "        qpu_id = i * num_gpus // len(xi)\n",
     "        asyncresults.append(\n",
-    "            cudaq.observe_async(kernel, h, xi[i][j, :], qpu_id=i))"
+    "            cudaq.observe_async(kernel, h, xi[i][j, :], qpu_id=qpu_id))\n",
+    "\n",
+    "print([res.get().expectation() for res in asyncresults])"
```
bettinaheim, line 355:
```suggestion
    "result = [res.get() for res in asyncresults]"
```

scripts/validate_container.sh
```diff
@@ -61,6 +61,7 @@ installed_backends=`\
 
 # remote_rest targets are automatically filtered, 
 # so is execution on the photonics backend
+# FIXME - remove items from $available_backends based on how many GPUs are installed?
```
bettinaheim, line 64: No, for one, I think it is good if the examples and tutorial we have don't just crash if you don't have the hardware to run them, but instead give a comprehensive message. Furthermore, it is still good to check that the backends that were primarily intended for multi-gpu usage still work on a single gpu (they should). 

I updated the executing_circuits and the multi-gpu notebooks to check the number of gpus first. I was a bit confused about the multi-gpu notebook, since the cell seemed to never actually wait for the async results, and I am not sure what would be a nice way to give some output beyond the timing.  
```suggestion
```

bmhowe23: OK, I reverted this placeholder change in lieu of your changes to the notebooks themselves.

bettinaheim (approved):

PR #1673, opened 2024-05-14
fixing notebook in tutorial

sacpis:
Co-authored with @bmhowe23

This will fix the following error seen in the [tutorial](https://nvidia.github.io/cuda-quantum/latest/examples/python/tutorials/vqe.html) 

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[2], line 7
      4 import numpy as np
      6 # Single precision
----> 7 cudaq.set_target("nvidia")
      8 # Double precision
      9 #cudaq.set_target("nvidia-fp64")

RuntimeError: [custatevec] %CUDA driver version is insufficient for CUDA runtime version in CuStateVecCircuitSimulator (line 334)

### DIFF
--- a/runtime/common/MeasureCounts.cpp
+++ b/runtime/common/MeasureCounts.cpp
@@ -28,6 +28,32 @@ std::string longToBitString(int size, long x) {
   return s;
 }
 
+void deserializeCounts(
+    std::vector<std::size_t> &data, std::size_t &stride,
+    std::unordered_map<std::string, std::size_t> &localCounts) {
+  auto nBs = data[stride];
+  stride++;
+
+  for (std::size_t j = stride; j < stride + nBs * 3; j += 3) {
+    auto bitstring_as_long = data[j];
+    auto size_of_bitstring = data[j + 1];
+    auto count = data[j + 2];
+    auto bs = longToBitString(size_of_bitstring, bitstring_as_long);
+    localCounts.insert({bs, count});
+  }
+  stride += nBs * 3;
+}
+
+std::string extractNameFromData(std::vector<std::size_t> &data,
+                                std::size_t &stride) {
+  auto nChars = data[stride++];
+
+  std::string name(data.begin() + stride, data.begin() + stride + nChars);
+
+  stride += nChars;
+  return name;
+}
+
 ExecutionResult::ExecutionResult(CountsDictionary c) : counts(c) {}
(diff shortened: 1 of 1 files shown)


zohimchandani:
Looks good. 

While you are on this, from this line: 

```
molecule, data = cudaq.chemistry.create_molecular_hamiltonian(geometry, 'sto-3g', 1, 0)
```

Can you please verify that `data` is a spinop type and change it to `hamiltonian`.

Thanks

docs/sphinx/examples/python/tutorials/vqe.ipynb
```diff
@@ -1,210 +1,239 @@
...
+        "execution_count":
+            4,
+        "metadata": {},
+        "outputs": [{
+            "name":
+                "stdout",
+            "output_type":
+                "stream",
+            "text": [
+                "Requirement already satisfied: openfermionpyscf==0.5 in /usr/local/lib/python3.10/dist-packages (0.5)\n",
```
bettinaheim, line 26: If we can omit this output, I think that would be nicer.

bettinaheim (approved):
The docs preview looks good to me. 
Do you know what caused the docs generation to try and execute the cells, leading to the error in the docs?

docs/sphinx/examples/python/tutorials/vqe.ipynb
```diff
@@ -1,210 +1,239 @@
...
+            "code",
+        "execution_count":
+            5,
+        "metadata": {},
+        "outputs": [],
+        "source": [
+            "import cudaq\n", "import matplotlib.pyplot as plt\n",
+            "from scipy.optimize import minimize\n", "import numpy as np\n",
+            "\n", "# Single precision\n", "cudaq.set_target(\"nvidia\")\n",
+            "# Double precision\n", "#cudaq.set_target(\"nvidia-fp64\")"
```
bmhowe23, line 84: It's hard for me to tell what actually changed in these diffs. Would you mind highlighting the actual changes?

I'm concerned that these changes aren't addressing the root cause of why the CI is running the Python notebooks and then including those "just ran" cells in the output.

bmhowe23: Marking this as resolved because we now know that that sphinx was running this notebook during the CI because before this PR, all the outputs were cleared in PR #1631.

Issue #1670, opened 2024-05-13
[c++] qvector initialization from vector of doubles fails for f32 simulator


annagrin:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

```cpp
__qpu__ void test(std::vector<double> inState) {
  cudaq::qvector q = inState;
}

int main() {
  std::vector<double> vec{M_SQRT1_2, 0., 0., M_SQRT1_2};
  auto counts = cudaq::sample(test, vec);
  counts.dump();

  printf("size %zu\n", counts.size());
}
```

**Running**

Running this example on a f32 simulator results in a runtime failure:

`nvq++ --enable-mlir -v from_state.cpp -o temp && ./temp`

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid user-provided state data. Simulator is FP32 but state data is FP64.
Aborted
```

**MLIR**

`cudaq-quake -D CUDAQ_SIMULATION_SCALAR_FP32  from_state.cpp |cudaq-opt`

```mlir
module attributes {llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.triple = "x86_64-unknown-linux-gnu", quake.mangled_name_map = {__nvqpp__mlirgen__function_test._Z4testSt6vectorIdSaIdEE = "_Z4testSt6vectorIdSaIdEE"}} {
  func.func @__nvqpp__mlirgen__function_test._Z4testSt6vectorIdSaIdEE(%arg0: !cc.stdvec<f64>) attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
    %0 = cc.stdvec_size %arg0 : (!cc.stdvec<f64>) -> i64
    %1 = math.cttz %0 : i64
    %2 = cc.stdvec_data %arg0 : (!cc.stdvec<f64>) -> !cc.ptr<f64>
[...]

PR #1674, opened 2024-05-14
Call the correct QIS function.

schweitzpgi:
Must call the correct qubit_allocate_array_with_state function. See #1670

### DIFF
--- a/lib/Optimizer/CodeGen/ConvertToQIR.cpp
+++ b/lib/Optimizer/CodeGen/ConvertToQIR.cpp
@@ -140,17 +140,19 @@ public:
 
     // Inspect the element type of the complex data, need to
     // know if its f32 or f64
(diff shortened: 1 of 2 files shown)


amccaskey:
This fix is needed, but I do not think it fixes 1670. You may already know that but I just wanted to get it out in writing. For the case of `stdvec<f64>` input compiled to a `f32` simulator backend, we'll need some sort of copy to get the data in the right precision.

schweitzpgi:
> This fix is needed, but I do not think it fixes 1670. You may already know that but I just wanted to get it out in writing. For the case of `stdvec<f64>` input compiled to a `f32` simulator backend, we'll need some sort of copy to get the data in the right precision.

Yes, that isn't the only issue with regards to the codegen path.

amccaskey (approved):

PR #1676, opened 2024-05-14
Heuristic threshold to prevent OpenMP slow-down in simple `spin_op` `*=`

1tnguyen:
### Description

Add a threshold for simple cases of `spin_op::operator*=` so that OpenMP doesn't start unnecessary multi-threading, causing a significant slow-down.

With some simple cases, like below (single term * single term)

```
const int n_spins = 50;
  const auto hamGen = [=]() -> cudaq::spin_op {
    cudaq::spin_op tdOp(n_spins);
    for (int i = 0; i < n_spins - 1; ++i) {
      tdOp += (cudaq::spin::x(i) * cudaq::spin::x(i + 1));
      tdOp += (cudaq::spin::y(i) * cudaq::spin::y(i + 1));
      tdOp += (cudaq::spin::z(i) * cudaq::spin::z(i + 1));
    }
    return tdOp;
  };

  constexpr int num_runs = 100;
  std::vector<double> runtimeUs;
  for (int run = 0; run < num_runs; ++run) {
    const auto start = std::chrono::high_resolution_clock::now();
    auto ham = hamGen();
    const auto stop = std::chrono::high_resolution_clock::now();
    auto duration =
        std::chrono::duration_cast<std::chrono::microseconds>(stop - start);
    runtimeUs.emplace_back(duration.count());
  }

  const double avg = std::reduce(runtimeUs.begin(), runtimeUs.end()) / runtimeUs.size();
  std::cout << "Avg: " << avg << "[us]\n";
  ```

Without this threshold, unnecessary OpenMP multithreading made it really slow (up to ~100x).


amccaskey (approved):
+1

PR #1680, opened 2024-05-14
Fix mismatches between input state and simulation engine precision.

### LINKED ISSUE #1670: [c++] qvector initialization from vector of doubles fails for f32 simulator

schweitzpgi:
It may be the case that the state vector data is in a precision that does not match that of the simulator. The runtime library knows precisely what simulator has been selected and can therefore convert the state vector to the expected format at runtime. This conversion may be *expensive* if the state vector is quite long. As a result, the runtime may want to emit a warning when a conversion path is taken.

The following conversions happen at runtime with this change:
```
  vector<float> -> vector<complex<float>>
  vector<float> -> vector<complex<double>>
  vector<double> -> vector<complex<float>>
  vector<double> -> vector<complex<double>>
  vector<complex<float>> -> vector<complex<double>>
  vector<complex<double>> -> vector<complex<float>>
```
If the input is `vector<complex<float>>` or `vector<complex<double>>` and the simulator's precision is a precise match, no conversions or runtime overhead is incurred.

Fix #1670

### DIFF
--- a/runtime/nvqir/NVQIR.cpp
+++ b/runtime/nvqir/NVQIR.cpp
@@ -199,6 +199,7 @@ Array *__quantum__rt__qubit_allocate_array_with_state_complex64(
   }
   // Input was complex<double> but we prefer complex<float>. Make a copy,
   // truncating the values.
+  cudaq::info("copying {} complex values from double to float.", size);
   auto convertData = std::make_unique<std::complex<float>[]>(size);
   for (uint64_t i = 0; i < size; ++i)
     convertData[i] = std::complex<float>{static_cast<float>(data[i].real()),
@@ -212,6 +213,7 @@ Array *__quantum__rt__qubit_allocate_array_with_state_fp64(uint64_t size,
                                                            double *data) {
   ScopedTraceWithContext("NVQIR::qubit_allocate_array_with_data_fp64", size);
   if (nvqir::getCircuitSimulatorInternal()->isDoublePrecision()) {
+    cudaq::info("copying {} double values to complex<double>.", size);
     auto convertData = std::make_unique<std::complex<double>[]>(size);
     for (uint64_t i = 0; i < size; ++i)
       convertData[i] = std::complex<double>{data[i], 0.0};
@@ -220,6 +222,7 @@ Array *__quantum__rt__qubit_allocate_array_with_state_fp64(uint64_t size,
   }
   // Input was double but we prefer complex<float>. Make a copy, truncating the
   // values.
+  cudaq::info("copying {} double values to complex<float>.", size);
   auto convertData = std::make_unique<std::complex<float>[]>(size);
   for (uint64_t i = 0; i < size; ++i)
     convertData[i] = std::complex<float>{static_cast<float>(data[i]), 0.0f};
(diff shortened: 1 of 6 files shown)


runtime/nvqir/NVQIR.cpp
```diff
@@ -184,15 +184,49 @@ Array *__quantum__rt__qubit_allocate_array(uint64_t size) {
...
+                         size);
   __quantum__rt__initialize(0, nullptr);
+  if (nvqir::getCircuitSimulatorInternal()->isDoublePrecision()) {
+    auto qubitIdxs = nvqir::getCircuitSimulatorInternal()->allocateQubits(
+        size, data, cudaq::simulation_precision::fp64);
+    return vectorSizetToArray(qubitIdxs);
+  }
+  // Input was complex<double> but we prefer complex<float>. Make a copy,
+  // truncating the values.
+  auto convertData = std::make_unique<std::complex<float>[]>(size);
```
amccaskey, line 202: I think it would be good to add some `cudaq::info(...)` messages when you make these copies

schweitzpgi: Sure.

1tnguyen (approved):

PR #1690, opened 2024-05-16
ORCA Python support

khalatepradnya:
* Expose the ORCA-specific APIs to Python front-end

  -  Documentation to be added later (Omar) 
  -  Integration tests will be covered separately (Pradnya)

### DIFF
--- a/python/CUDAQuantumExtension.cpp
+++ b/python/CUDAQuantumExtension.cpp
@@ -13,6 +13,7 @@
 #include <pybind11/pytypes.h>
 #include <pybind11/stl.h>
 
+#include "cudaq/Optimizer/CAPI/Dialects.h"
 #include "runtime/common/py_ExecutionContext.h"
 #include "runtime/common/py_NoiseModel.h"
 #include "runtime/common/py_ObserveResult.h"
@@ -33,10 +34,11 @@
 #include "runtime/mlir/py_register_dialects.h"
 #include "utils/LinkedLibraryHolder.h"
 #include "utils/OpaqueArguments.h"
-#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
 
-#include "cudaq/Optimizer/CAPI/Dialects.h"
+#include "../runtime/cudaq/platform/orca/orca_qpu.h"
+
 #include "mlir/Bindings/Python/PybindAdaptors.h"
+#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
 
 namespace py = pybind11;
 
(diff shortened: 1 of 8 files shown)


khalatepradnya:
Tagging @Omar-ORCA  for visibility.

python/CUDAQuantumExtension.cpp
```diff
@@ -33,10 +34,11 @@
 #include "runtime/mlir/py_register_dialects.h"
 #include "utils/LinkedLibraryHolder.h"
 #include "utils/OpaqueArguments.h"
-#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
 
-#include "cudaq/Optimizer/CAPI/Dialects.h"
+#include "../runtime/cudaq/platform/orca/orca_qpu.h"
```
bmhowe23, line 38: It looks like lines 21-34 are all `runtime/cudaq/...`. Can the leading `../` of line 38 be removed and then move the include up there?

python/CUDAQuantumExtension.cpp
```diff
@@ -33,10 +34,11 @@
...
+
 #include "mlir/Bindings/Python/PybindAdaptors.h"
+#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
```
bmhowe23, line 41: Is this new include (and the one on line 16) still needed? I looked at `runtime/cudaq/platform/orca/orca_qpu.h`, and I can't figure out why they would be new dependencies as a result of these changes. (It's to leave it if it is needed ... I just figured I'd double check just in case they were leftovers from an earlier WIP?)

bmhowe23 (approved):
Thanks, Pradnya. This LGTM. Just a few minor questions...

python/CUDAQuantumExtension.cpp
khalatepradnya: The header files on lines# 21-34 reside within the 'python' sub-directory, `<root>/python/runtime/cudaq` for example, https://github.com/NVIDIA/cuda-quantum/blob/main/python/runtime/cudaq/algorithms/py_draw.h. 
The new include comes from `<root>/runtime/cudaq`, i.e. https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/platform/orca/orca_qpu.h. Hence I put it at the bottom of the list.

python/CUDAQuantumExtension.cpp
khalatepradnya: Thanks, I will confirm this...

python/CUDAQuantumExtension.cpp
bmhowe23: OK, thanks. I think the following path should work then: `#include "cudaq/platform/orca/orca_qpu.h"`

python/CUDAQuantumExtension.cpp
bmhowe23: Based on my quick testing, it looks like mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h may be required but "cudaq/Optimizer/CAPI/Dialects.h" is not.

python/CUDAQuantumExtension.cpp
khalatepradnya: Done. 8e6e94aca7e98ad31e16ba62286bb1d86c81cda9

python/CUDAQuantumExtension.cpp
khalatepradnya: Done. 8e6e94aca7e98ad31e16ba62286bb1d86c81cda9

PR #1691, opened 2024-05-17
[MPS Simulator] Optimization for XX, YY, ZZ rotations

1tnguyen:
### Description

These are 2-body ops that are commonly used in quantum applications. 
Without frontend exposure of (Rxx/XXPhase, Ryy/YYPhase, Rzz/ZZPhase) gates, users need to write them in the decomposed form or using `exp_pauli`. Both results in 2 CNOT gates. 

This is not ideal for MPS since we want to minimize the number of 2-q gates.

Hence, add a special handling in MPS for these cases.


runtime/nvqir/cutensornet/simulator_mps_register.cpp
```diff
@@ -84,6 +84,78 @@ class SimulatorMPS : public SimulatorTensorNetBase {
...
+                             const std::vector<std::size_t> &qubitIds,
+                             const cudaq::spin_op &op) override {
+    // Special handling for equivalence of Rxx(theta), Ryy(theta), Rzz(theta)
+    // expressed as exp_pauli.
+    //  Note: for MPS, the runtime is ~ linear with the number of 2-body gates
+    //  (gate split procedure).
+    // Hence, we check if this is a Rxx(theta), Ryy(theta), or Rzz(theta), which
+    // are commonly-used gates and apply the operation directly (the base
+    // decomposition will result in 2 CNOT gates).
+    const auto shouldHandlePauliOp =
```
bmhowe23, line 125: What would happen if `op = "XXXX"`? Is that the same as `"XXII"` followed by `"IIXX"`? If so, should we attempt to handle that by iterating over the elements in the `op` looking for neighboring identical Paulis, and just sending the remaining Paulis to the base class implementation?

1tnguyen: If `op = "XXXX"`, we don't handle it and let the base class performs the decomposition (with CNOT ladders before and after). It's not the same as `"XXII"` followed by `"IIXX"` though if users wrote the circuit in terms of `exp_pauli`, e.g.,
`exp_pauli(theta, 'XX', q0, q1)` followed by `exp_pauli(theta, 'XX', q2, q3)`.

zohimchandani:
Team, any updates on this, please? This is holding a benchmark for a collaboration currently so would be nice to get it moving if we can. 

Thanks :)

bmhowe23 (approved):

zohimchandani:
Do we need to add these gates in the docs?

1tnguyen:
> Do we need to add these gates in the docs?

We don't actually add any 'new' gates. I think the expression of `Rxx`/`XXPhase`, `Ryy`/`YYPhase`, or `Rzz`/`ZZPhase` in terms of `exp_pauli` could be added to our docs as a footnote/example. Most other frameworks that have these native gates should already have equations for the gate matrix in terms of Pauli exponential.

PR #1692, opened 2024-05-17
Allow math dialect to lower and run on NVQC

### LINKED ISSUE #1663: Python version on Quantum Cloud does not allow exponentiation operator
### Required prerequisites

### Describe the bug

When using the Python version on Quantum Cloud, the exponentiation operator ** gives the error:
        error: failed to legalize operation 'math.ipowi'
        RuntimeError: Remote rest platform: applying IR passes failed.

The workaround is to apply multiplication N times in a loop

Here is a simple code example which, when run gives an error.
Commenting the offending line, and uncommenting the WORKAROUND lines enables this to run on quantum cloud with the --target nvqc argument passed.

        #######################################################
        
        # PYTHON KERNEL - EXPONENTIATION BUG - QUANTUM CLOUD
        
        import cudaq
        
        # Inverse Quantum Fourier Transform
        @cudaq.kernel
        def iqft(register: cudaq.qview):
            M_PI = 3.1415926
            
            N = register.size()
            for i in range(N / 2):
                swap(register[i], register[N - i - 1])
            
            for i in range(N - 1):
                h(register[i])
                j = i + 1
                for y in range(i, -1, -1):
                
                    denom = 2 ** (j - y)       # DEVNOTE: Issue ??? This does not work on Quantum Cloud
                    
                    '''
                    X = j - y                   # this WORKAROUND works
                    denom = 1
                    for i in range(X):
                        denom *= 2
                    '''
                    
                    theta = -M_PI / denom
                    r1.ctrl(theta, register[j], register[y]);
                  
            h(register[N - 1])
         
        @cudaq.kernel
        def bug_cloud_exp():
        
            # Allocate a number of qubits
            counting_qubits = cudaq.qvector(4)
        
            # Place some bits into test state
            h(counting_qubits)
        
            # Apply inverse quantum Fourier transform
            iqft(counting_qubits)
            
            # Measure to gather sampling statistics
            # (excludes the auxillary qubit).
            mz(counting_qubits)
            
            
        #######################
        # MAIN
         
        if __name__ == "__main__":
        
            print(f"Python Kernel Bug - Exponentiation - Quantum Cloud")
            
            result = cudaq.sample(bug_cloud_exp)
            
            print(f"result: {result}")

### Expected behavior

There should be no error when using the ** operator.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.1
- **Python version**: 3.10.12

### Suggestions

This is an important bug as the exponentiation operator is critical to many kernels such as QFT
The WOARKAROUND is OK, but this will frustrate users initially

bmhowe23:
Resolves #1663 

Note: this requires updates on the NVQC servers in order to fully resolve the issue, and that is not planned until 0.8 is released.

### DIFF
--- a/runtime/common/BaseRestRemoteClient.h
+++ b/runtime/common/BaseRestRemoteClient.h
@@ -94,6 +94,8 @@ protected:
       "func.func(combine-quantum-alloc)",
       "canonicalize",
       "cse",
+      "convert-math-to-llvm",
+      "convert-math-to-funcs",
       "quake-to-qir"};
   static inline const std::vector<std::string> serverPasses = {};
   // Random number generator.
--- a/runtime/common/BaseRestRemoteClient.h
+++ b/runtime/common/BaseRestRemoteClient.h
@@ -19,6 +19,7 @@
 #include "cudaq/Frontend/nvqpp/AttributeNames.h"
 #include "cudaq/Optimizer/CodeGen/OpenQASMEmitter.h"
 #include "cudaq/Optimizer/CodeGen/Passes.h"
+#include "cudaq/Optimizer/CodeGen/Pipelines.h"
 #include "cudaq/Optimizer/Dialect/CC/CCDialect.h"
 #include "cudaq/Optimizer/Dialect/Quake/QuakeDialect.h"
 #include "cudaq/Optimizer/Transforms/Passes.h"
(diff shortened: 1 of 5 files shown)


schweitzpgi (approved):
LGTM. Thanks, Ben.

runtime/common/BaseRestRemoteClient.h
```diff
@@ -94,6 +94,8 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
       "func.func(combine-quantum-alloc)",
       "canonicalize",
       "cse",
+      "convert-math-to-llvm",
```
bmhowe23, line 97: At a minimum, this line probably isn't needed `populateMathToLLVMConversionPatterns()` is being called in `ConvertToQIR.cpp`.

I'm still investigating whether or not there are other options like using `addPipelineConvertToQIR()`, which is common code.

bmhowe23: Updated in 017a196c5

Issue #1682, opened 2024-05-15
break within two loops is not handled right.


marwafar:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

I am getting this error when running my code (code is in next section):

error: 'cc.unwind_break' op arity of arguments and loop result mismatch
RuntimeError: Failure while executing pass pipeline.

During handling of the above exception, another exception occurred:

RuntimeError: could not compile code for 'qrbm_reuse_ancilla'.

### Steps to reproduce the bug

```
import cudaq
import numpy as np

cudaq.set_target("nvidia")

@cudaq.kernel
def qrbm_reuse_ancilla(v_nodes:int, h_nodes:int, theta:list[float], coupling:list[float]):

    qubits_num=v_nodes+h_nodes
    qubits=cudaq.qvector(qubits_num)
    ancilla=cudaq.qubit()

    count=0
    for i in range(v_nodes+h_nodes):
        ry(theta[count],qubits[i])
        count+=1

    count=0

    for v in range(v_nodes):
        for h in range(v_nodes,v_nodes+h_nodes):
            
            while True:
                ry.ctrl(coupling[count],qubits[v],qubits[h],ancilla)
                x(qubits[v])
                ry.ctrl(coupling[count+1],qubits[v],qubits[h],ancilla)
                x(qubits[v])
                x(qubits[h])
                ry.ctrl(coupling[count+1],qubits[v],qubits[h],ancilla)
                x(qubits[v])
                ry.ctrl(coupling[count],qubits[v],qubits[h],ancilla)
                x(qubits[v])
                x(qubits[h])

                res=mz(ancilla)

                if res:
                    x(ancilla)
                    break   
        
            count+=2

    mz(qubits)



v_nodes=2
h_nodes=2
anc=v_nodes*h_nodes

num_parameters=v_nodes+h_nodes+anc
np.random.seed(42)
[...]

PR #1693, opened 2024-05-17
[Bug-fix] [Python] Correctly handle 'break' within nested loops

khalatepradnya:
* Root cause: The 'break' statement visitor in Python AST bridge did not account for multiple nested loops
* Also fixed the same issue for 'continue' statement
* Added the example from issue #1682  as a test


bmhowe23 (approved):
LGTM 👍

PR #1694, opened 2024-05-17
[Core] Add cc.sizeof

schweitzpgi:
And a distinct operation for determining the memory size in bytes of a type.

- Change GKE to use new sizeof op.
- Add to codegen.
- Add roundtrip.
- Fix tests.


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -1437,22 +1456,37 @@ void cudaq::cc::ReturnOp::getCanonicalizationPatterns(
...
+    auto inpTy = sizeOp.getInputType();
+    std::optional<unsigned> bitsize;
+    if (isa<IntegerType, FloatType>(inpTy)) {
+      bitsize = inpTy.getIntOrFloatBitWidth();
+    } else if (auto strTy = dyn_cast<cudaq::cc::StructType>(inpTy)) {
+      if (auto bits = strTy.getBitSize())
+        bitsize = bits;
+    }
+    if (!bitsize)
+      return failure();
```
bmhowe23, line 1479: Will this show a nice user error message if the type isn't supported yet? I.e., what would happen if they try to take the sizeof a qubit?

Edit: Actually, will this new operator get invoked if the user adds `sizeof` to their code? I'm guessing that since it's not in the bridge, then it won't happen.

bmhowe23 (approved):
LGTM. Just once question about potential error messages.

lib/Optimizer/Dialect/CC/CCOps.cpp
schweitzpgi: This is a quirk of the MLIR interface, but "failure" in the canonicalization pattern means "we don't have anything here to update" rather than "the train went over the cliff".

`sizeof qubit` is nonsense and will be caught in codegen by generating a `poison` value. So don't do it.

PR #1702, opened 2024-05-21
New tutorials 

zohimchandani:
1. new qft tutorial 
2. new qaoa tutorial by @mmvandieren 
3. changed variable name in VQE


anthony-santana (approved):
LGTM 👍

anthony-santana:
@sacpis Would you mind lending a second set of eyes on the changes to the python notebooks here? I've ensured all cells are executed and in proper order, but just want to double check that we're not breaking your recent fix 😄

bmhowe23:
> @sacpis Would you mind lending a second set of eyes on the changes to the python notebooks here? I've ensured all cells are executed and in proper order, but just want to double check that we're not breaking your recent fix 😄

I think @sacpis's recent fix should still be ok. The main thing is that we don't want to modify `docs/sphinx/conf.py` to have `nbsphinx_allow_errors = True` as that masks errors until Publishing. @sacpis's change is still in effect, and it looks like this PR doesn't change that, so we should be good.

I'll also be uploading some other fixes that seem to be required to get past the current failing CI jobs.

bmhowe23 (commented):
Well, unfortunately the https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2FNVIDIA%2Fcuda-quantum%2Fdocs%2Fpreview%2Fpr-1702%2Fexamples%2Fpython%2Ftutorials%2Fquantum_fourier_transform.html= link doesn't look so good. IIRC, we don't have LaTeX integrated with Sphinx, so we should probably re-evaluate what to do with the tutorials that have LaTeX.

bmhowe23:
The preview documentation doesn't look right, but it looks right when I run it locally and view it locally. We decided to merge this PR and see what the real docs end up looking like.

PR #1706, opened 2024-05-22
[Python] Qvector init from state

annagrin:
### Description

Add initialization of qvector from `cudaq::State`


include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
```diff
@@ -93,6 +95,11 @@ inline mlir::Type getResultType(mlir::MLIRContext *context) {
       getQuantumTypeByName("Result", context));
 }
 
+inline mlir::Type getStateType(mlir::MLIRContext *context) {
```
annagrin, line 98: unused

PR #1708, opened 2024-05-23
Enable custom unitary operations in C++ library mode

amccaskey:
Add `CUDAQ_REGISTER_OPERATION` macro that supports fully parameterized unitary matrices. This first PR focuses solely on library mode simulation, and a follow up PR will hook up the auto-generated C function to our MLIR pipeline.


runtime/cudaq/qis/qubit_qis.h
```diff
@@ -18,661 +18,244 @@
...
-  // apply an x.
-  if (!controls.empty())
-    for (std::size_t i = 0; i < controls.size(); i++)
-      if (qubitIsNegated[i])
-        getExecutionManager()->apply("x", {}, {}, {controls[i]});
+// --------------------------
+// Useful C++17 compliant concept checks (note we re-implement
+// std::remove_cvref since its a C++20 thing)
+template <typename T>
+using remove_cvref = std::remove_cv_t<std::remove_reference_t<T>>;
```
schweitzpgi, line 44:
```suggestion
#ifdef CUDAQ_USE_STD20
template <typename T>
using remove_cvref = std::remove_cvref<T>;
#else
template <typename T>
using remove_cvref = std::remove_cv_t<std::remove_reference_t<T>>;
#endif
```
makes it more explicit.

schweitzpgi:
Are we going to punt on using concepts in the headers then?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -780,11 +781,62 @@ struct MergeRotationPattern : public OpRewritePattern<OP> {
...
+
+// %4 = Something that returns float value
+// %5 = cc.alloca f64
+// cc.store %4, %5 : !cc.ptr<f64>
+// %6 = cc.load %5 : !cc.ptr<f64>
+// quake.ry (%6) %1 : (f64, !quake.ref) -> ()
+//
+// to
+//
+// quake.ry(%4) ...
```
schweitzpgi, line 793: memtoreg should already do this.

test/AST-Quake/single_qubit_ctor.cpp
```diff
@@ -6,7 +6,7 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: cudaq-quake %cpp_std %s | FileCheck %s
+// RUN: cudaq-quake %cpp_std %s | cudaq-opt | FileCheck %s
```
schweitzpgi, line 9: why this change?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -780,11 +781,62 @@ struct MergeRotationPattern : public OpRewritePattern<OP> {
...
+template <typename OP>
+class RewritePassByReferenceRotations : public OpRewritePattern<OP> {
```
schweitzpgi, line 795: This store-forwarding shouldn't be replicated in the canonicalizer.

test/AST-Quake/ctrl_vector.cpp
```diff
@@ -33,86 +33,3 @@ struct lower_ctrl_as_qreg {
...
-
-struct test_two_control_call {
-  void operator()() __qpu__ {
-    auto lambda = [](cudaq::qubit &qb) __qpu__ {
-      h<cudaq::ctrl>(qb);
-      x<cudaq::ctrl>(qb);
-    };
-    cudaq::qarray<4> qs;
-    cudaq::qubit qb;
-    cudaq::control(lambda, qs, qb);
```
schweitzpgi, line 45: Why are we discarding all the tests for `cudaq::control()`? Is CUDA-Q dropping that from the API? If not, we need to continue to test it.

unittests/qis/QubitQISTester.cpp
```diff
@@ -424,4 +424,127 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
...
+    };
+
+    counts = cudaq::sample(rotation_adjoint_test);
+    counts.dump();
+    for (auto &[bits, count] : counts) {
+      EXPECT_TRUE(bits == "0");
+    }
+  }
+}
+
```
khalatepradnya, line 549: Tried to run this locally, getting failures for the following three simulators (they don't run in CI, AFAIK):
```
275 - custatevec_fp32_CustomUnitaryTester.checkSimple (Failed)
389 - tensornet_CustomUnitaryTester.checkSimple (Failed)
473 - tensornet_mps_CustomUnitaryTester.checkSimple (Failed)
```

test/AST-Quake/single_qubit_ctor.cpp
amccaskey: The latest update to `qubit_qis.h` takes all quantum operation parameters by reference now. Previously, any scalar rotation parameters were passed by value. This change caused the bridge to produce `cc.ptr<f64>` for the parameter instead of just `f64`. This also lead to some cases of `alloca : ptr<f64>` and so I made a canonicalization pass to remove those patterns in an effort to minimize changes to the FileCheck tests. Interested in other opinions here though.

test/AST-Quake/ctrl_vector.cpp
amccaskey: These are tests have code that is not adherent to the language specification. We don't specify ctrl on single target ops with no controls. The new qubit_qis.h caught this language spec violation

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
amccaskey: True. I can remove this if you think its ok to update the FileCheck tests because of the issue in my other comment.

unittests/qis/QubitQISTester.cpp
amccaskey: Thanks for catching me here. Will look into it

khalatepradnya: I think this issue pertains to the default qubit ordering. 
For uniformity, the targets should be reversed for qpp and tensornet simulators.

amccaskey: Yea check me on that last commit to make sure I didn't miss something

unittests/qis/QubitQISTester.cpp
```diff
@@ -424,4 +424,127 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
   }
 }
 
+// Test someone can build a library of custom operations
+CUDAQ_REGISTER_OPERATION(
+    /* Name */ CustomHadamard, /*NumTargets*/ 1, /*NumParameters*/ 0,
+    /* Unitary Generator */ {M_SQRT1_2, M_SQRT1_2, M_SQRT1_2, -M_SQRT1_2});
+CUDAQ_REGISTER_OPERATION(CustomX, 1, 0, {0, 1, 1, 0});
+CUDAQ_REGISTER_OPERATION(CustomCNOT, 2, 0,
+                         {1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0});
```
khalatepradnya, line 433: This is the Big-endian representation of CNOT. Hence it applies as `CustomCNOT(target, control)` (not the CUDA-Q convention of `cnot(control, target)`.

amccaskey: The latest update will ask the simulators about their bit ordering and convert if necessary

runtime/nvqir/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ if (CUSTATEVEC_ROOT AND CUDA_FOUND)
   add_subdirectory(custatevec)
 endif() 
 if (CUTENSORNET_ROOT AND CUDA_FOUND)
-  add_subdirectory(cutensornet)  
+  # add_subdirectory(cutensornet)
```
khalatepradnya, line 45: This should be uncommented.

unittests/qis/QubitQISTester.cpp
khalatepradnya: The tenornet simulators follow the same convention as qpp (MSB), hence they should override the `getQubitOrdering` method.

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -990,4 +573,38 @@ std::vector<T> slice_vector(std::vector<T> &original, std::size_t start,
...
-} // namespace cudaq
\ No newline at end of file
+} // namespace cudaq
+
+#define __qop__ __attribute__((annotate("user_custom_quantum_operation")))
+
+/// Register a new custom unitary operation providing a unique name,
+/// the number of target qubits, the number of rotation parameters (can be 0),
+/// and the unitary matrix as a 1D row-major `std::vector<complex>`
+/// representation following a MSB qubit ordering.
```
khalatepradnya, line 583: Curious... why require MSB when the internal default is LSB?

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -990,4 +573,38 @@ std::vector<T> slice_vector(std::vector<T> &original, std::size_t start,
...
+#define CUDAQ_REGISTER_OPERATION(NAME, NUMT, NUMP, ...)                        \
```
khalatepradnya, line 584: If interpret the count of targets by accepting unitary as a 2-D matrix, and then interpret any additional argument as a parameter to get count of parameters, can we drop the `NUMT` and `NUMP` arguments?

unittests/qis/QubitQISTester.cpp
```diff
@@ -424,4 +424,127 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
...
+CUDAQ_REGISTER_OPERATION(
+    CustomU3, 1, 3,
+    {std::cos(parameters[0] / 2.),
```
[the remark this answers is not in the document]
amccaskey: As it is implemented now, yes. I went that way since the CircuitSimulator takes custom unitaries like this

runtime/nvqir/CMakeLists.txt
amccaskey: Sorry :/

runtime/cudaq/qis/qubit_qis.h
amccaskey: I'm open to this part changing. I went with that since I'm most familiar with the QC textbook ordering as opposed to the physicist ordering (e.g. qiskit).

runtime/cudaq/qis/qubit_qis.h
amccaskey: For the current implementation, we need the integers at compile time (for the tuple processing), so I'm not sure we can use the unitary data which is provided as a std::vector. Maybe I'm wrong though.

test/AST-Quake/ctrl_vector.cpp
schweitzpgi: Can we repair the violations to meet the spirit of the tests, rather than delete the tests? It looks like the space being explored here is the use of the `cudaq::ctrl` modifier and the `cudaq::control()` function and how they interact.

Additionally, we should probably move anything that is incorrect and a standards violation to a negative test rather than throwing it away. We want to make sure the new templates continue to check for bad usage and don't regress too.

test/AST-Quake/single_qubit_ctor.cpp
schweitzpgi: Curious, but why change to use references? That seems like it will be inefficient for C++ and generate stack clutter (which we will have to manage better), but I'm assuming we're doing this because of Python.

It should be possible to introduce an intrinsic and/or runtime C++ entry point that will convert values to stack slots, bounds the lifetime of these ephemeral references, etc.

khalatepradnya:
These changes are looking good. I will leave it to Eric / Thien for final approval.
I think it is still missing the override of `getQubitOrdering` method by the tenornet simulators since they follow the same convention as qpp (MSB).
Got the following two failures as of commit 0efb23bcd97e992b4d770fb94f7f878fb63621fa
```
The following tests FAILED:
        389 - tensornet_CustomUnitaryTester.checkSimple (Failed)
        473 - tensornet_mps_CustomUnitaryTester.checkSimple (Failed)
```

khalatepradnya (approved):
👍🏽

PR #1709, opened 2024-05-23
NVQC: Auto-select NCA ID based on API key

bmhowe23:
This change is intended to have no effect on production usage of NVQC while making it simpler for developers to use the dev organization. (I.e. the `NVQC_NCA_ID` environment shouldn't be needed for most use cases anymore.)


1tnguyen (approved):
This is great 👍

PR #1710, opened 2024-05-24
[stateHandling] Remove swap semantics optimization support.

schweitzpgi:
Follows the latest design change for ownership management of state objects. Remove the support for "move semantics" to allow the logical lifetime of a state object to escape into called functions and the swapping of any resources it may contain.
  - remove the move constructor from qvector
  - remove move attribute from init_state op
  - remove explicit test for move semantics
  - simplify the handling of init_state data: it has to be a pointer to a supported data type, floating-point, complex, or state.


1tnguyen (approved):
👍

annagrin (approved):

PR #1711, opened 2024-05-24
Create MPS Simulation State from Tensors

1tnguyen:
### Description

Ability to do `state::from_data` with a list of MPS tensors (could be host or device memory)

  - Tensors retrieved from the MPS simulator backend.

   - Externally-provided tensors. For Python, these could be numpy or cupy `ndarray`s.


python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -158,6 +158,70 @@ void bindPyState(py::module &mod) {
                 reinterpret_cast<std::complex<double> *>(info.ptr), info.size));
           },
           "Return a state from data.")
+      .def_static(
+          "from_data",
+          [](std::vector<py::buffer> tensors) {
```
amccaskey, line 163: Is `py::buffer` shared_ptr like? Should this be pass by reference?

python/tests/kernel/test_state_mps.py
```diff
@@ -0,0 +1,133 @@
...
+    print("random state: ", state_vec)
+    stacked_state_vec = state_vec.reshape(2, 2).transpose()
+    # Do SVD
+    U, S, Vh = np.linalg.svd(stacked_state_vec)
+    # Important: Tensor data must be in column major.
+    left_tensor = np.asfortranarray(U)
+    right_tensor = np.asfortranarray(np.dot(np.diag(S), Vh))
+    state = cudaq.State.from_data([left_tensor, right_tensor])
+    state.dump()
+    assert np.isclose(state[0], state_vec[0])
```
amccaskey, line 80: does `state[0]` automatically generate the state-vector-like representation from the MPS tensors?

1tnguyen: No, it will be delegated to the `amplitude` method.

amccaskey (approved):
LGTM

python/runtime/cudaq/algorithms/py_state.cpp
1tnguyen: absolutely. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1711/commits/a1f4cb93b1991d52988975d8d8b2ea094fc335f0

PR #1713, opened 2024-05-24
[Python] Qvector init from state update

annagrin:
### Description
Add initialization of `qvecto`r from `cudaq::State`


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3330,6 +3361,54 @@ def visit_BinOp(self, node):
         else:
             self.emitFatalError(f"unhandled binary operator - {node.op}", node)
 
+    def __store_state(self, value: State):
+        # Compute a unique hash string for the state data
+        hashValue = hashlib.sha1(value).hexdigest(
```
annagrin, line 3366: This reads from the state's buffer which causes a copy of the state from the gpu. Maybe just use a counter as an id?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3330,6 +3361,54 @@ def visit_BinOp(self, node):
...
+        )[:10] + self.name.removeprefix('__nvqppBuilderKernel_')
+        print(f'state has value for {self.name}: {hashValue}')
```
annagrin, line 3368: remove print

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -233,24 +242,49 @@ pyAltLaunchKernelBase(const std::string &name, MlirModule module,
...
-    if (svdata.kernelName != name)
+  for (auto &[stateHash, stateData] : *stateStorage) {
+    if (stateData.kernelName != name)
       continue;
+
     auto setStateFPtr = jit->lookup("nvqpp.set.state." + stateHash);
-    if (!setStateFPtr)
-      throw std::runtime_error(
-          "python alt_launch_kernel failed to get set state function.");
+    if (auto error = setStateFPtr.takeError()) {
```
annagrin, line 250: Note: using `!setStateFPtr` just fails with a message about forgetting to check the result of `Expected<E>`. Updated to take the error and report the actual error message (here and the new code below)

python/cudaq/kernel/ast_bridge.py
annagrin: done

python/cudaq/kernel/ast_bridge.py
annagrin: replaced by an id (just a counter)

1tnguyen (approved):
LGTM 👍

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -119,6 +119,12 @@ inline mlir::LLVM::LLVMStructType stdVectorImplType(mlir::Type eleTy) {
   return mlir::LLVM::LLVMStructType::getLiteral(ctx, eleTys);
 }
 
+inline mlir::Type stateImplType(mlir::Type eleTy) {
+  auto *ctx = eleTy.getContext();
+  auto eTy = cudaq::opt::factory::getCharType(ctx);
+  return cudaq::opt::factory::getPointerType(eTy);
```
schweitzpgi, line 125: Can be simplified to just:
```c++
return cudaq::opt::factory::getPointerType(eleTy.getContext());
```
On the other hand, we don't use the `eleTy`, so we could erase this function and just use `getPointerType` at the call site.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -2026,6 +2033,9 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
     if (isa<NoneType>(eleTy))
       return factory::getPointerType(type.getContext());
     eleTy = typeConverter.convertType(eleTy);
+    if (isa<NoneType>(eleTy))
+      return factory::getPointerType(type.getContext());
+
```
schweitzpgi, line 2038: Why doesn't the recursion handle this? The recursive call on line 2035 ought to handle pointers to pointers.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -434,8 +434,10 @@ class GenerateKernelExecution
         hasTrailingData = true;
         continue;
       }
-      if (isa<cudaq::cc::PointerType>(currEleTy))
+      if (isa<cudaq::cc::PointerType>(currEleTy) &&
+          !isStatePointerType(currEleTy)) {
```
schweitzpgi, line 438: nit: don't need braces.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -933,6 +935,13 @@ class GenerateKernelExecution
     builder.create<cudaq::cc::StoreOp>(loc, endPtr, sret2);
   }
 
+  static bool isStatePointerType(mlir::Type ty) {
+    if (auto ptrTy = dyn_cast<cudaq::cc::PointerType>(ty)) {
```
schweitzpgi, line 939: nit: no braces

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -941,7 +950,8 @@ class GenerateKernelExecution
         cudaq::cc::numberOfHiddenArgs(hasThisPointer, hiddenSRet);
     if (count > 0 && args.size() >= count &&
         std::all_of(args.begin(), args.begin() + count, [](auto i) {
-          return isa<cudaq::cc::PointerType>(i.getType());
+          return isa<cudaq::cc::PointerType>(i.getType()) &&
+                 !isStatePointerType(i.getType());
```
schweitzpgi, line 954: I don't get this one. We're dropping hidden arguments. Why would a `cudaq::state*` appear as a `this` or `sret`? Do we have an example?

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1207,8 +1217,9 @@ class GenerateKernelExecution
         hasTrailingData = true;
         continue;
       }
-      if (isa<cudaq::cc::PointerType>(inTy))
+      if (isa<cudaq::cc::PointerType>(inTy) && !isStatePointerType(inTy))
```
schweitzpgi, line 1220: This doesn't look correct. We cannot pass a `cudaq::state*` as a pointer-free value.

PR #1715, opened 2024-05-24
[stateHandling] Correct assignment operator.

schweitzpgi:
The assignment operator in state was declared as
```c++
  state &operator=(state);
```
but used move/swap semantics.

This means that the compiler was obligated to make a copy of the state object on the RHS to pass to the assignment. That temporary would then be swapped with the LHS resources. This behavior is not expected for this assignment operator and introduces a copy defeating the performance gain of a using std::swap.

Here we change the assignment operator to the following and discard the original.
```c++
  state &operator=(state &&);
```
This version eliminates the RHS copy. The LHS and RHS exchange resources, which is expected with this move-assignment signature. Specifically, a state object holds a shared_ptr. So the shared_ptr resources are simply exchanged and the reference count updates can be omitted.


1tnguyen (approved):
👍

PR #1720, opened 2024-05-28
Reduce NVQC server log file clutter

bmhowe23:
Our NVQC logs have too much clutter originating from the utils container talking to the health endpoint, which we don't have any control over. Use this change to help reduce log clutter on the health endpoint.


1tnguyen (approved):
LGTM 👍

PR #1722, opened 2024-05-28
Add C++ `kernel_builder::qalloc(state&)`

1tnguyen:
### Description

Adding support for C++ `kernel_builder` to allocate `qvector` from a state.


Adrian999999999 (approved):

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -619,6 +625,36 @@ QuakeValue qalloc(ImplicitLocOpBuilder &builder,
...
+      throw std::runtime_error("loading callbacks should never fail");
+  }
+  auto statePtrTy = cudaq::cc::PointerType::get(cc::StateType::get(context));
+  std::intptr_t vecStor = reinterpret_cast<std::intptr_t>(&stateVectorStorage);
+  auto vecPtr = builder.create<arith::ConstantIntOp>(vecStor, 64);
+  auto idxOp = builder.create<arith::ConstantIntOp>(index, 64);
+  auto statePtr = builder
+                      .create<func::CallOp>(statePtrTy, getStatePtrCallback,
+                                            ValueRange{vecPtr, idxOp})
+                      .getResult(0);
```
schweitzpgi, line 649: I think the newerish plan is to just pass to `State*` itself and not have a callback to unwrap the `State` and get a pointer to the vector data. Can you eliminate this call and pass `vecPtr` instead of this `statePtr` on line 654?

1tnguyen: Thanks for the pointer :) I've revised it to be much simpler in https://github.com/NVIDIA/cuda-quantum/pull/1722/commits/83c0c11a275750f69eaee0c4c4f8eab4d2abbe2d.

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -619,6 +619,22 @@ QuakeValue qalloc(ImplicitLocOpBuilder &builder,
   return QuakeValue(builder, qubits);
 }
 
+QuakeValue qalloc(mlir::ImplicitLocOpBuilder &builder, cudaq::state *state,
+                  StateVectorStorage &stateVectorStorage) {
+  auto *context = builder.getContext();
+  auto statePtrTy = cudaq::cc::PointerType::get(cc::StateType::get(context));
+  auto statePtr = builder.create<cc::CastOp>(
```
annagrin, line 626: Is the cast needed? I added the `AllocaOp/InitializeStateOp` combo optimization case with the state pointer in ConverToQIR (https://github.com/NVIDIA/cuda-quantum/blob/a6ca2ddf522b3f9018fb59329369fd48bddd334d/lib/Optimizer/CodeGen/ConvertToQIR.cpp#L156)  and converted StateType to  ptr<i8> there as well (https://github.com/NVIDIA/cuda-quantum/blob/a6ca2ddf522b3f9018fb59329369fd48bddd334d/lib/Optimizer/CodeGen/ConvertToQIR.cpp#L2020) - will those work for the builder mode here as well?

Adrian999999999 (approved):

runtime/cudaq/builder/kernel_builder.cpp
annagrin: Hmm is the cast needed because the type of `state` is not `ptr<StateType>`? If yes, approved!

annagrin (approved):
LGTM with a question

schweitzpgi (approved):

runtime/cudaq/builder/kernel_builder.cpp
1tnguyen: The cast is needed because I use a `ConstantIntOp` to inject the captured state pointer value into the IR, then cast it back to a `ptr<cc::StateType>`.

Issue #1638, opened 2024-05-08
Enable emitting circuit diagrams for unitary quantum kernels in LaTeX


bettinaheim:
# Description
The CUDA-Q platform allows users to build and execute hybrid quantum-classical kernels.  While learning CUDA-Q and also while developing with CUDA-Q, it is helpful to visualize quantum circuits and states to gain intuition, to spot check code, and to communicate with others. We need to expand the visualization functionality into CUDA-Q. CUDA-Q currently contains the ability to visualize unitary quantum kernels as circuit diagrams. These diagrams are emitted in ASCII. To facilitate incorporating such diagrams in papers and other materials, we would like to add support for emitting them in LaTeX instead. 

## Details 

The documentation of the current `draw` function can be found [here (Python)](https://nvidia.github.io/cuda-quantum/latest/api/languages/python_api.html#cudaq.draw) and [here (C++)](https://nvidia.github.io/cuda-quantum/latest/api/languages/cpp_api.html#_CPPv4I0DpEN5cudaq4drawENSt6stringERR13QuantumKernelDpRR4Args).
The API is defined in [draw.h](https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/algorithms/draw.h), with the corresponding implementation of this functionality is in [draw.cpp](https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/algorithms/draw.cpp). This implementation should be extended to take an optional argument `format` , i.e. an additional overload

```
template<typename QuantumKernel, typename ...Args>
std::string cudaq::draw(std::string format, QuantumKernel &&kernel, Args&&... args)
```
should be created, where valid values for `format` are "ascii" or "latex", and a new Python binding should be created for that overload in [py_draw.cpp](https://github.com/NVIDIA/cuda-quantum/blob/main/python/runtime/cudaq/algorithms/py_draw.cpp).

Possible options for LaTeX package to generate circuit diagrams are [qcircuit](https://ctan.org/pkg/qcircuit?lang=en) and [yquant](https://github.com/projekter/yquant).

freifrauvonbleifrei:
Hi, I have started working on this for unitary hack!
 I would gravitate towards [quantikz](https://ctan.org/pkg/quantikz) for the LaTeX output; I believe it is the most commonly used package. Is there a specific reason why you would prefer the other packages?

boschmitt:
Hi, thanks for working on this!

[...]

PR #1723, opened 2024-05-29
Circuit diagrams for unitary quantum kernels in LaTeX

freifrauvonbleifrei:
Add LaTeX circuit diagrams to the cudaq draw() function.

### Description

this PR addresses https://github.com/NVIDIA/cuda-quantum/issues/1638


freifrauvonbleifrei:
recheck

freifrauvonbleifrei:
I am currently building the latex reference output for testing, to match this string output: 
https://github.com/NVIDIA/cuda-quantum/blob/1c76ab2e459eba86e8aa147acf7a23963cf21abd/unittests/integration/draw_tester.cpp#L66-L86

This is the current draft:
```
\documentclass{minimal}
\usepackage{quantikz}
\begin{document}
\begin{quantikz}
  \lstick{$q_0$} & \gate{h} & \ctrl{1} & \ctrl{2} & \ctrl{1} & \gate{y}  & \gate{r_1(3.142)} & \qw     & \swap{2} & \qw      & \swap{1} & \swap{2} & \qw      & \targX{} & \ctrl{2} & \swap{3} & \swap{3} & \gate{r_x(2.718)} & \gate{\text{sdg}} & \gate{r_x(-2.718)} & \qw \\
  \lstick{$q_1$} & \gate{h} & \gate{x} & \ctrl{1} & \gate{y} & \ctrl{-1} & \gate{\text{tdg}} & \qw     & \qw      & \swap{1} & \targX{} & \qw      & \swap{1} & \targX{} & \targX{} & \ctrl{2} & \ctrl{2} & \gate{r_y(3.142)} & \ctrl{-1}         & \gate{r_y(-3.142)} & \qw \\
  \lstick{$q_2$} & \gate{h} & \qw      & \gate{y} & \ctrl{-1}& \ctrl{-2} & \gate{z}          & \gate{s}& \targX{} & \targX{} & \qw      & \targX{} & \targX{} & \qw      & \targX{} & \qw      & \ctrl{-2}& \gate{r_z(-3.142)}& \gate{r_z(3.142)} & \qw                & \qw \\
  \lstick{$q_3$} & \gate{h} & \qw      & \qw      & \qw      & \qw       & \qw               & \qw     & \qw      & \qw      & \qw      & \qw      & \qw      & \ctrl{-3}& \ctrl{-2}& \targX{} & \targX{} & \qw               & \qw               & \qw                & \qw \\
\end{quantikz}
\end{document}
```
which produces this
when run with `latex` and `dvisvgm`.

I think that `.svg` output should be the easiest to integrate with jupyter notebooks.

My svg viewer shows weird triangles for the numerical dots (they aren't there when I run `dvisvgm --no-fonts`, but then all semantics is lost, too). I would just trust that these display issues will be fixed somewhere along the toolchain, and accept them for now :shrug: 

@bettinaheim @amccaskey @boschmitt does this look somewhat like the thing you'd like?

freifrauvonbleifrei:
I think this PR fulfills the requirements stated in issue #1638. 

However, I would like to do some more notebook integration. 
This would mean
* having a Python function that invokes subprocesses to produce the `.dvi` and `.svg` intermediate files, and returns the contents of the `.svg` as string. (Questions: Where does it belong in the python module? What happens in case of failure, e.g. if `latex` or `dvisvgm` are not installed, or files are not writable?)
* wrapping this in a `_repr_svg_` function [that can be understood by IPython](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display) . But here is the tricky part: which class should this function belong to? To obtain the trace, both the kernel and its arguments are required.

Opinions are highly appreciated!

freifrauvonbleifrei:
The latest commit lets you display the svg in a jupyter notebook. Here, for instance, I added `displaySVG` to the quantum Fourier transform tutorial:
image
I am not entirely happy about the Python import structure though, but cannot think of a more elegant way to do it right now. Feedback is appreciated!

freifrauvonbleifrei:
And I added a `_repr_svg_` for `PyKernelDecorator`, which will output the svg if the kernel needs no parameters. 
Here in the `quantum_bits.ipynb` tutorial:
image

boschmitt:
Hi @freifrauvonbleifrei, thanks for all the work put into this.

I noticed that you are using C++20 concepts in the code. We also use it, however, you need to provide an alternative so that the code can also compile with C++17. For example:

https://github.com/NVIDIA/cuda-quantum/blob/b522e18b0bfe0c931db3e2ee243dd01752597e19/runtime/cudaq/algorithms/sample.h#L242-L249

freifrauvonbleifrei:
Hi @boschmitt , thank you for having a look. I was not aware that C++17 compatibility would be required, and it is especially surprising to me that other places in the code use concepts unguarded:
https://github.com/NVIDIA/cuda-quantum/blob/6c736fa3ea133cdf1dd86fa612b45c1a01eef566/include/cudaq/Optimizer/Builder/Factory.h#L29
How are these other parts ensured to be compiled with C++20? 

I can add some guards for sure, but I would like to make sure this is what is needed in the project, as readability would really suffer.

boschmitt:
> I was not aware that C++17 compatibility would be required, and it is especially surprising to me that other places in the code use concepts unguarded:
> 
> https://github.com/NVIDIA/cuda-quantum/blob/6c736fa3ea133cdf1dd86fa612b45c1a01eef566/include/cudaq/Optimizer/Builder/Factory.h#L29
> 
> 
> How are these other parts ensured to be compiled with C++20?

I was not sufficiently precise in my observation. The thing is: not all parts of the codebase have the same requirements regarding C++17 support. We need C++17 support on the public headers, i.e., the headers users would import (directly or indirectly) to develop their quantum applications. (In other words, to build the project, one needs a C++20 capable compiler. However, to use it, you only need C++17.)

The code you showed is not part of the public API and is not installed (distributed with CUDAQ). Thus, there is no need to comply with the C++17 requirement.

python/tests/display/test_draw.py
```diff
@@ -0,0 +1,113 @@
...
+
+import cudaq
+import numpy as np
+import os
+import pytest
+
+
+@cudaq.kernel
+def bar(qvec: cudaq.qview):
+    # FIXME https://github.com/NVIDIA/cuda-quantum/issues/1734
```
bettinaheim, line 17: Thanks for filing the issue, and submitting PR 1756 as well!

bettinaheim (approved):
Nice! Overall this looks very nice. 
Let's follow up in a separate PR for the C++ 17 support in the public headers, so that we can stick to the Unitary Hack time lines.

Issue #1726, opened 2024-05-29
Cannot call C++ __qpu__ function from within another kernel if parameters contain vector of cudaq::measure_results


bmhowe23:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Calling a function with `const std::vector<cudaq::measure_result> &result_vec` parameters seems to not work when running with `--enable-mlir`.

### Steps to reproduce the bug

Consider the following file:
```cpp
#include "cudaq.h"

bool xor_result(const std::vector<cudaq::measure_result> &result_vec) __qpu__ {
  bool result = false;
  for (auto x : result_vec)
    result ^= x;
  return result;
}

bool kernel() __qpu__ {
  cudaq::qvector q(7);
  std::vector<cudaq::measure_result> mz_res = mz(q);
  bool res = xor_result(mz_res);
  return res;
}

int main(int argc, char *argv[]) {
  printf("Result: %d\n", static_cast<int>(kernel()));
  return 0;
}
```
That produces the following compilation error:
```bash
$ nvq++ --enable-mlir issue.cpp
issue.cpp:21:14: error: 'func.call' op kernel call was not inlined, possible recursion in call tree
issue.cpp:21:14: note: see current operation: %3 = "func.call"(%2) {callee = @__nvqpp__mlirgen__function_xor_result._Z10xor_resultRKSt6vectorIbSaIbEE} : (!cc.stdvec<i1>) -> i1
failed: "/workspaces/cuda-quantum/build/bin/cudaq-opt --pass-pipeline=builtin.module(func.func(unwind-lowering),canonicalize,lambda-lifting,func.func(memtoreg{quantum=0}),canonicalize,apply-op-specialization,kernel-execution,aggressive-early-inlining,func.func(quake-add-metadata),device-code-loader{use-quake=1},expand-measurements,func.func(lower-to-cfg),canonicalize,cse) issue.qke -o issue.qke.v8G01q"
```
[...]

Issue #1725, opened 2024-05-29
Using static_cast on measurement results can cause the nvq++ compiler to crash despite the compiler knowing how to do that cast.


bmhowe23:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Performing `static_cast<bool>(measure_result)` causes `cudaq-quake` (part of `nvq++`) to crash with an unhelpful error message.

Currently the problem only happens when using a target that uses MLIR mode (hardware backends, NVQC, and any time the user runs with `--enable-mlir`).

The problem is technically twofold:
- [ ] The compiler should accept the static_cast<bool>() operation, especially since the compiler can convert between measure_result and bool if the `static_cast` is not performed. (See below for example.)
- [ ] The compiler gives an unhelpful error message

### Steps to reproduce the bug

The following code reproduces the problem:
```cpp
#include "cudaq.h"

bool xor_kernel() __qpu__ {
  cudaq::qvector qvec(7);
  x(qvec);
  auto res = mz(qvec);
  bool xor_result = false;
  
  // This fails
  for (auto x : res)
    xor_result ^= static_cast<bool>(x);

  // This works
  // for (auto x : res)
  //   xor_result ^= x;

  return xor_result;
}

int main(int argc, char *argv[]) {
  bool result = xor_kernel();
  printf("Result = %d\n", static_cast<int>(result));
  return 0;
}
```

```bash
$ nvq++ --enable-mlir issue.cpp
cudaq-quake: /workspaces/cuda-quantum/lib/Frontend/nvqpp/ConvertExpr.cpp:615: bool cudaq::details::QuakeBridgeVisitor::TraverseImplicitCastExpr(clang::ImplicitCastExpr*, clang::RecursiveASTVisitor<cudaq::details::QuakeBridgeVisitor>::DataRecursionQueue*): Assertion `typeStack.size() == typeStackDepth && "must be original depth"' failed.
[...]

PR #1728, opened 2024-05-29
Enable MLIR compiler by default.

schweitzpgi:
Change the default mode of the compiler from "library mode" to "MLIR mode". Both modes will remain available. This only changes which mode is selected by default.

(This replaces a previous PR that had gotten very cluttered.)


bmhowe23:
Just a note: this PR would change the conditions under which #1725 and #1726 occur from not-the-default configuration to the default configuration.

schweitzpgi:
Closing since it looks like I need to create a new fork to work around other problems with the CI.

PR #1731, opened 2024-05-29
[stateHandling] Future proofing codegen dialect.

schweitzpgi:
Typed pointers have disappeared in the LLVM-IR dialect, so we needed a more coherent way of identifying the type of data passed to the initialization state. This change extends the op to keep track of the information so it isn't lost when lowering to opaque pointers.

Add a separate pass for testing and a regression test to make sure the RAII ops are being formed correctly.

### DIFF
--- a/lib/Optimizer/CodeGen/ConvertToQIR.cpp
+++ b/lib/Optimizer/CodeGen/ConvertToQIR.cpp
@@ -6,17 +6,8 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-#if (defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER))
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-overflow"
-#pragma GCC diagnostic ignored "-Wrestrict"
-#endif
-#include "PassDetails.h"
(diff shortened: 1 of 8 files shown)


lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -141,30 +138,26 @@ class QmemRAIIOpRewrite
...
+    if (auto elePtrTy = dyn_cast<cudaq::cc::PointerType>(eleTy))
+      eleTy = elePtrTy.getElementType();
+    if (auto arrayTy = dyn_cast<cudaq::cc::ArrayType>(eleTy))
+      eleTy = arrayTy.getElementType();
+    bool fromComplex = false;
+    if (auto complexTy = dyn_cast<ComplexType>(eleTy)) {
+      fromComplex = true;
+      eleTy = complexTy.getElementType();
     }
+    if (isa<cudaq::cc::StateType>(eleTy))
```
annagrin, line 151: I added a conversion from `cc::StateType` to a `ptr<i8>` - does it need to be removed for this to work?

schweitzpgi: (chatted offline)

annagrin (approved):
LGTM

PR #1732, opened 2024-05-30
[StateHandling] Fixes and improvements to tensornet states

1tnguyen:
### Description

- Add a qubit count threshold for these state impls to fully contract the state vector rather than performing amplitude contraction (one-by-one)

- Fix sporadic canonicalization errors when initializing MPS tensors: we should keep those init tensors in the most compact form.

- Add host workspace check as well. We don't need it with the default MPS configuration, but it might be required.


runtime/nvqir/cutensornet/mps_simulation_state.h
```diff
@@ -96,6 +96,11 @@ class MPSSimulationState : public cudaq::SimulationState {
   std::unique_ptr<TensorNetState> state;
   std::vector<MPSTensor> m_mpsTensors;
   ScratchDeviceMem m_scratchPad;
+  // Max number of qubits whereby the tensor network state should be contracted
+  // and cached into a state vector.
+  // This speeds up sequential state amplitude accessors for small states.
+  static inline constexpr std::size_t g_maxQubitsForStateContraction = 30;
```
schweitzpgi, line 102:
```suggestion
  static constexpr std::size_t g_maxQubitsForStateContraction = 30;
```
inline isn't needed here.

runtime/nvqir/cutensornet/tensornet_state.cpp
```diff
@@ -440,6 +474,9 @@ TensorNetState::factorizeMPS(int64_t maxExtent, double absCutoff,
   HANDLE_CUTN_ERROR(cutensornetStateCompute(
       m_cutnHandle, m_quantumState, workDesc, extentsPtr.data(),
       /*strides=*/nullptr, allData.data(), 0));
+
+  if (hostWork)
+    free(hostWork);
```
schweitzpgi, line 479: nit: Add braces? (I think the style in `runtime/` is to use braces as much as possible. :)

runtime/nvqir/cutensornet/tn_simulation_state.cpp
```diff
@@ -152,6 +152,33 @@ TensorNetSimulationState::overlap(const cudaq::SimulationState &other) {
 
 std::complex<double>
 TensorNetSimulationState::getAmplitude(const std::vector<int> &basisState) {
+  if (getNumQubits() != basisState.size())
+    throw std::runtime_error(
+        fmt::format("[tensornet-state] getAmplitude with an invalid number "
+                    "of bits in the "
+                    "basis state: expected {}, provided {}.",
+                    getNumQubits(), basisState.size()));
```
schweitzpgi, line 160: not this PR. I wonder if we can't remove most/all of our dependence on the `fmt` package by just building the strings with regular `std` functions.

runtime/nvqir/cutensornet/tn_simulation_state.h
```diff
@@ -71,5 +72,10 @@ class TensorNetSimulationState : public cudaq::SimulationState {
   std::unique_ptr<TensorNetState> m_state;
   cutensornetHandle_t m_cutnHandle;
   ScratchDeviceMem m_scratchPad;
+  // Max number of qubits whereby the tensor network state should be contracted
+  // and cached into a state vector.
+  // This speeds up sequential state amplitude accessors for small states.
+  static inline constexpr std::size_t g_maxQubitsForStateContraction = 30;
```
schweitzpgi, line 78:
```suggestion
  static  constexpr std::size_t g_maxQubitsForStateContraction = 30;
```

schweitzpgi (commented):
LGTM.

runtime/nvqir/cutensornet/tensornet_state.cpp
1tnguyen: oh, now I know :-) Fixed https://github.com/NVIDIA/cuda-quantum/pull/1732/commits/973ed7aef64bcb4b709dc85adeb1b27b2251ce4a

runtime/nvqir/cutensornet/tn_simulation_state.cpp
1tnguyen: I think we can. It's just more convenient: our `Logger.h` makes `fmt` available, which is much nicer to use to construct these messages.

schweitzpgi (approved):

Issue #1107, opened 2024-01-18
[RFC] Improve the internal data implementation for cudaq::state


amccaskey:
Background 
------------
Currently, state just wraps a tuple containing the shape of the state data and a vector to the state data. This is sub-optimal because it forces the creation of copies, especially for state vector data managed by the GPU. 

We need to rethink the internal data representation for `cudaq::state` to remove this tuple and replace with a new data type specific to simulation backends (i.e. some interface that backend simulators must implement) that allows `cudaq::state` to track the state data in a backend-native manner, i.e. holding the GPU device pointer. 

Proposal
---------
This RFC introduces the following new types and type changes: 

```cpp
namespace cudaq {

/// @brief state_data is a variant type
/// encoding different forms of user state vector data
/// we support.
using state_data = std::variant<
    std::vector<cudaq::complex>,
    std::pair<cudaq::complex *, std::size_t>,
    std::vector<cudaq::complex *>, std::shared_ptr<TensorNetworkState>>;

class SimulationState {
public: 

  /// @brief Simulation states are made up of a
  /// vector of Tensors. Each tensor tracks its data pointer
  /// and the tensor extents.
  struct Tensor {
    cudaq::complex *data = nullptr;

    std::vector<std::size_t> extents;
    precision fp_precision;
    std::size_t get_rank() const { return extents.size(); }
    std::size_t get_num_elements() const;
    std::size_t element_size() const ;
  };

  /// @brief Create a new subclass specific SimulationState
  /// from the user provided data set.
  virtual std::unique_ptr<cudaq::SimulationState>
  createFromData(const state_data &data) ;

  /// @brief Return the tensor at the given index. Throws
  /// for an invalid tensor index.
  virtual Tensor getTensor(std::size_t tensorIdx = 0) const = 0;

  /// @brief Return all tensors that represent this state
  virtual std::vector<Tensor> getTensors() const = 0;

  /// @brief Return the number of tensors that represent this state.
  virtual std::size_t getNumTensors() const = 0;

  /// @brief Return the number of qubits this state represents.
  virtual std::size_t getNumQubits() const = 0;

  /// @brief Compute the overlap of this state representation with
  /// the provided `other` state, e.g. `<this | other>`.
  virtual cudaq::complex overlap(const SimulationState &other) = 0;

  /// @brief Return the amplitude of the given computational
  /// basis state.
  virtual cudaq::complex
  getAmplitude(const std::vector<int> &basisState) = 0;

[...]

PR #1733, opened 2024-05-30
[python] add qalloc from state in kernel builder mode

annagrin:
### Description
Support initializing `qalloc` from state in builder mode in python.

- Added common code `CapturedDataStorage` to generate code for capturing states and numpy arrays
- Use it to generate MLIR for `qalloc(state)` in builder mode
- Migrate AST bridge code to use `CapturedDataStorage` in kernel mode as well.
- Add tests
- See examples in https://github.com/NVIDIA/cuda-quantum/issues/1086

Towards: 
https://github.com/NVIDIA/cuda-quantum/issues/1086
https://github.com/NVIDIA/cuda-quantum/issues/1107


schweitzpgi (commented):
LGTM. Lots and lots of tests, which is great.

python/tests/builder/test_qalloc_init.py
```diff
@@ -0,0 +1,615 @@
...
+    reason='Could not find nvidia in installation')
+
+# float
+
+
+@skipIfNvidiaFP64NotInstalled
+def test_kernel_float_params_f64():
+
+    cudaq.reset_target()
+    cudaq.set_target('nvidia-fp64')
```
1tnguyen, line 28: It's not for this PR, but at some point, we may need to update these new Python tests for state handling so that they can run on the `qpp` target, mainly to get them running as part of the PR CI pipeline (no GPU available).

1tnguyen (approved):
LGTM 👍

python/tests/builder/test_qalloc_init.py
annagrin: Sure!

PR #1736, opened 2024-05-30
Allow inlined calls to cudaq-entrypoint funcs as long as not in a thunk

### LINKED ISSUE #1726: Cannot call C++ __qpu__ function from within another kernel if parameters contain vector of cudaq::measure_results

bmhowe23:
Fixes #1726 

The underlying problem is that `cudaq-entrypoint` attribute is applied to all `__qpu__` functions, but it is removed if there are any `cudaq::qubit` parameters. Downstream code was trying to prevent inlining of `cudaq-entrypoint` kernels for thunk functions, but it was preventing inlining of ALL `cudaq-entrypoint` kernels, which was a bit heavy handed.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Common/InlinerInterface.h
+++ b/include/cudaq/Optimizer/Dialect/Common/InlinerInterface.h
@@ -18,21 +18,21 @@ namespace cudaq {
 
 /// Generic opt-in with inlining any operation. Inlining kernels is important to
 /// build up larger functions to translate into quantum circuits.
-/// Do not allow kernel entry points to be inlined into their callers, typically
+/// Do not allow kernel entry points to be inlined into their callers while in
 /// the thunk function.
 struct EnableInlinerInterface : public mlir::DialectInlinerInterface {
   using DialectInlinerInterface::DialectInlinerInterface;
 
-  bool isLegalToInline(mlir::Region *, mlir::Region *src, bool,
+  bool isLegalToInline(mlir::Region *dest, mlir::Region *src, bool,
                        mlir::IRMapping &) const final {
-    if (auto func = src->getParentOfType<mlir::func::FuncOp>())
-      return !(func->hasAttr(cudaq::entryPointAttrName));
+    if (auto destFunc = dest->getParentOfType<mlir::func::FuncOp>())
+      if (destFunc.getName().ends_with(".thunk"))
+        if (auto srcFunc = src->getParentOfType<mlir::func::FuncOp>())
+          return !(srcFunc->hasAttr(cudaq::entryPointAttrName));
     return true;
   }
   bool isLegalToInline(mlir::Operation *op, mlir::Region *, bool,
                        mlir::IRMapping &) const final {
-    if (auto func = op->getParentOfType<mlir::func::FuncOp>())
-      return !(func->hasAttr(cudaq::entryPointAttrName));
     return true;
   }
   bool isLegalToInline(mlir::Operation *call, mlir::Operation *callable,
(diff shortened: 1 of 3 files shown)


include/cudaq/Optimizer/Dialect/Common/InlinerInterface.h
```diff
@@ -18,21 +18,21 @@ namespace cudaq {
...
+    if (auto destFunc = dest->getParentOfType<mlir::func::FuncOp>())
+      if (destFunc.getName().ends_with(".thunk"))
+        if (auto srcFunc = src->getParentOfType<mlir::func::FuncOp>())
+          return !(srcFunc->hasAttr(cudaq::entryPointAttrName));
     return true;
   }
   bool isLegalToInline(mlir::Operation *op, mlir::Region *, bool,
                        mlir::IRMapping &) const final {
-    if (auto func = op->getParentOfType<mlir::func::FuncOp>())
-      return !(func->hasAttr(cudaq::entryPointAttrName));
```
schweitzpgi, line 35: Shouldn't this one mirror the one above?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -914,7 +914,7 @@ bool cudaq::EnableInlinerInterface::isLegalToInline(Operation *call,
   if (auto applyOp = dyn_cast<quake::ApplyOp>(call))
     if (applyOp.applyToVariant())
       return false;
-  return !(callable->hasAttr(cudaq::entryPointAttrName));
+  return true;
```
schweitzpgi, line 917: and here?

include/cudaq/Optimizer/Dialect/Common/InlinerInterface.h
bmhowe23: Updated in c2e116edc

schweitzpgi (approved):
+1

PR #1743, opened 2024-05-31
Add pass to convert cc dialect to LLVM-IR dialect.

schweitzpgi:
Part 2 of #1727

### DIFF
--- a/include/cudaq/Optimizer/CodeGen/Passes.h
+++ b/include/cudaq/Optimizer/CodeGen/Passes.h
@@ -17,6 +17,7 @@
 #include "mlir/Pass/PassRegistry.h"
 
 namespace mlir {
+class LLVMTypeConverter;
 class MLIRContext;
 class Pass;
 class PassManager;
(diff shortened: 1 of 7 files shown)


amccaskey (approved):

amccaskey (changes requested):
I just realized the intrinsic functions for delegating to the ExecutionManager were added to the DefaultExecutionManager. This means that if you had lowered CC code and wanted to compile / link to an executable you'd have to link to the libcudaq-em-default library, but we really want to just link to libcudaq. Can you move those functions to execution_manager.cpp?

schweitzpgi:
> I just realized the intrinsic functions for delegating to the ExecutionManager were added to the DefaultExecutionManager. This means that if you had lowered CC code and wanted to compile / link to an executable you'd have to link to the libcudaq-em-default library, but we really want to just link to libcudaq. Can you move those functions to execution_manager.cpp?

Not easily. These functions require `cudaq::getExecutionManager()` which isn't a part of `libcudaq.so` and therefore causes the build to fail.

amccaskey (approved):

schweitzpgi:
Another possibility would be to provide a weak definition in libcudaq.so that just emits an error message and aborts, forcing the user to link a real execution manager implementation.

PR #1744, opened 2024-05-31
Move the glue code to execution_manager.cpp.

schweitzpgi:
This throws in a weak definition of cudaq::getExecutionManager() to let everything link in libcudaq.so. In order to really use this however, a proper execution manager must be linked to override the weak definition.

Ed: looks like this idea is DOA for Python.


schweitzpgi:
This looks like it isn't going to work for us.

PR #1746, opened 2024-05-31
Add python and cpp kernel mode example for Trotterization using quantum states

annagrin:
### Description
Add  python and c++ kernel mode examples to demonstrate usage of quantum states.

Addition changes needed to make the examples work:
- C++: Allow passing `pauli_word` as a parameter to kernels
- Python: Allow creating an identity `cudaq.SpinOperator` with a number of qubits

Towards:
https://github.com/NVIDIA/cuda-quantum/issues/1086
https://github.com/NVIDIA/cuda-quantum/issues/1107


dictionary.dic
1tnguyen, line 1: Is this a new file?

docs/sphinx/examples/cpp/other/trotter_kernel_mode.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// Compile and run with:
+// ```
+// nvq++ -std=c++17 --enable-mlir -v trotter_kernel_mode.cpp -o temp && ./temp
```
1tnguyen, line 11: Is `-std=c++17` compulsory? Otherwise, I'd think it's better to use the default (not specifying the standard).

1tnguyen: Also, I'd recommend using a more descriptive output, like `trotter.x`.

docs/sphinx/examples/python/trotter_kernel_mode.py
```diff
@@ -0,0 +1,106 @@
...
+    def termWords(op: cudaq.SpinOperator) -> list[str]:
+        result = []
+        ham.for_each_term(lambda term: result.append(term.to_string(False)))
+        return result
+
+    # Observe the average magnetization of all spins (<Z>)
+    average_magnetization = 0.0
+    for i in range(0, n_spins):
+        average_magnetization += ((1.0 / n_spins) * cudaq.spin.z(i))
+    average_magnetization -= 1.0
```
1tnguyen, line 78: Could you please check (e.g., printing out `average_magnetization`) that at the end of this code, the `spin_op` is `(1/N)*(Z0 + Z1 + ...)`?

The reason the C++ code has a -1 is because the [default constructor](https://github.com/NVIDIA/cuda-quantum/blob/2da801bc343a9cd5b62003e03cb24beb817b98c5/runtime/cudaq/spin/spin_op.cpp#L109) of `spin_op` is an identity term.

I'm not sure if it's also the case here.

docs/sphinx/examples/cpp/other/trotter_kernel_mode.cpp
annagrin: Not compulsory. just copied from out tests in github:) Will do!

docs/sphinx/examples/python/trotter_kernel_mode.py
annagrin: Will do, thanks!

dictionary.dic
annagrin: oops, will remove, thanks!

docs/sphinx/examples/python/trotter_kernel_mode.py
annagrin: I had a couple of bugs - forgot to call the constructor (instead of initializing the `average_magnetization` to 0) and had a copy-paste bug in the hamiltonian. Fixed both, now the `average_magnetization` looks the same in c++ and python, and the results match as well.

dictionary.dic
annagrin: done

1tnguyen (approved):
LGTM 👍

include/cudaq/Frontend/nvqpp/ASTBridge.h
```diff
@@ -779,6 +779,17 @@ inline bool isCharPointerType(mlir::Type t) {
   return false;
 }
 
+/// Is \p t of type `pauli_word`?
+inline bool isCharspanPointerType(mlir::Type t) {
+  if (auto ptrTy = dyn_cast<cc::PointerType>(t)) {
+    mlir::Type eleTy = ptrTy.getElementType();
+    if (auto ptrTy = dyn_cast<cc::CharspanType>(eleTy)) {
+      return true;
+    }
```
schweitzpgi, line 788:
```suggestion
    return isa<cc::CharspanType>(eleTy);
```

lib/Optimizer/Builder/Factory.cpp
```diff
@@ -21,6 +21,10 @@ cc::StateType factory::getCudaqStateType(MLIRContext *ctx) {
   return cc::StateType::get(ctx);
 }
 
+cudaq::cc::CharspanType factory::getCharspanType(MLIRContext *ctx) {
+  return cudaq::cc::CharspanType::get(ctx);
+}
```
schweitzpgi, line 26: nit: this function doesn't add anything. can be removed.

PR #1751, opened 2024-06-03
Fix #1749 - bad cfg conversion

### LINKED ISSUE #1749: `do { ... } while (false)` causes infinite loop when `--enable-mlir` is set
### Required prerequisites

### Describe the bug

The `do { ... } while (0)` construct is common in C/C++ macros, and if used in a `__qpu__` kernel, it causes the program to go in an apparent infinite loop.

Consider the following test program
```cpp
#include "cudaq.h"

void kernel() __qpu__ {
  cudaq::qubit q;
  do {
    x(q);
  } while (0);
}

  auto counts = cudaq::sample(kernel);
  counts.dump();
  return 0;
}
```
Compile and run it like this to see the error:
```bash
$ nvq++ --enable-mlir issue.cpp && ./a.out
# This hangs during program execution
```

### Expected behavior

The program should run the same as when not using `--enable-mlir`:
```bash
$ nvq++ issue.cpp && ./a.out
{ 1:1000 }
```

Not a regression

### Environment

- **CUDA Quantum version**: latest (2da801bc3)
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

schweitzpgi:
Fix #1749. The CFG for do-while was incorrect.

### DIFF
--- a/lib/Optimizer/Transforms/LowerToCFG.cpp
+++ b/lib/Optimizer/Transforms/LowerToCFG.cpp
@@ -168,10 +168,7 @@ public:
     }
     auto comparison = whileCond.getCondition();
     auto *bodyBlock = loopOp.getDoEntryBlock();
-    auto *condBlock =
-        loopOp.hasStep()
-            ? loopOp.getStepBlock()
-            : (loopOp.isPostConditional() ? bodyBlock : whileBlock);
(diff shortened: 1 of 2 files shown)


bmhowe23 (approved):
Thanks, Eric.

PR #1755, opened 2024-06-03
Fix conversion of state vectors at runtime

annagrin:
### Description
Correct conversion of vectors at runtime

- Make sure we copy correct size from the state when converting the data to the simulation precision.
- Remove MLIR-based conversion from python code and rely on conversion in the runtime code instead.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1670


targettests/execution/from_state_double.cpp
```diff
@@ -0,0 +1,28 @@
...
+// CHECK: size 2
+
+int main() {
+  std::vector<double> vec{M_SQRT1_2, 0., 0., M_SQRT1_2};
+  auto counts = cudaq::sample(test, vec);
+  counts.dump();
+
+  printf("size %zu\n", counts.size());
+
+
```
schweitzpgi, line 27: nit:
```suggestion
```

runtime/nvqir/NVQIR.cpp
```diff
@@ -135,6 +136,55 @@ std::size_t qubitToSizeT(Qubit *q) {
...
+  for (uint64_t i = 0; i < stateSize; ++i)
+    convertData[i] = std::complex<To>{static_cast<To>(data[i].real()),
+                                      static_cast<To>(data[i].imag())};
+  return convertData;
+}
+
+/// Input was complex<float>/complex<double> but we prefer
+/// complex<double>/complex<float>. Make a copy, extending or truncating the
+/// values.
+template <typename To, typename From>
```
schweitzpgi, line 170: Can we use concepts here to ensure that `To` and `From` are both in the set `{ float, double }`?

runtime/nvqir/NVQIR.cpp
```diff
@@ -135,6 +136,55 @@ std::size_t qubitToSizeT(Qubit *q) {
...
+
+  auto stateSize = pow(2, size);
+  constexpr auto toType = typeName<To>();
+  constexpr auto fromType = typeName<From>();
+
+  cudaq::info("copying {} {} values to complex<{}>", stateSize, fromType,
+              toType);
+
+  auto convertData = std::make_unique<std::complex<To>[]>(stateSize);
+  for (uint64_t i = 0; i < stateSize; ++i)
```
schweitzpgi, line 182:
```suggestion
  for (std::size_t i = 0; i < stateSize; ++i)
```

schweitzpgi (commented):
Looks good to me. It really simplifies a bunch of the code.

runtime/nvqir/NVQIR.cpp
annagrin: Done!

targettests/execution/from_state_double.cpp
annagrin: Done offline

schweitzpgi (approved):
+2

runtime/nvqir/NVQIR.cpp
```diff
@@ -135,6 +136,63 @@ std::size_t qubitToSizeT(Qubit *q) {
...
+std::unique_ptr<std::complex<To>[]> convertToComplex(std::complex<From> *data,
+                                                     std::size_t size) {
+  auto stateSize = pow(2, size);
+  constexpr auto toType = typeName<To>();
+  constexpr auto fromType = typeName<From>();
+  cudaq::info("copying {} complex<{}> values to complex<{}>", stateSize,
+              fromType, toType);
+
+  auto convertData = std::make_unique<std::complex<To>[]>(stateSize);
+  for (uint64_t i = 0; i < stateSize; ++i)
```
schweitzpgi, line 171:
```suggestion
  for (std::size_t_t i = 0; i < stateSize; ++i)
```
to match the declaration of stateSize.

runtime/nvqir/NVQIR.cpp
```diff
@@ -135,6 +136,63 @@ std::size_t qubitToSizeT(Qubit *q) {
...
+  return "double";
+}
+
+/// Input was complex<float>/complex<double> but we prefer
+/// complex<double>/complex<float>. Make a copy, extending or truncating the
+/// values.
+template <SimPrecisionType To, SimPrecisionType From>
+std::unique_ptr<std::complex<To>[]> convertToComplex(std::complex<From> *data,
+                                                     std::size_t size) {
+  auto stateSize = pow(2, size);
```
schweitzpgi, line 164: Should the loop be over `size` or `stateSize`? I guess `size` must be the number of qubits here.

runtime/nvqir/NVQIR.cpp
```diff
@@ -198,13 +256,7 @@ Array *__quantum__rt__qubit_allocate_array_with_state_complex64(
...
     return vectorSizetToArray(qubitIdxs);
   }
-  // Input was complex<double> but we prefer complex<float>. Make a copy,
-  // truncating the values.
-  cudaq::info("copying {} complex values from double to float.", size);
-  auto convertData = std::make_unique<std::complex<float>[]>(size);
-  for (uint64_t i = 0; i < size; ++i)
-    convertData[i] = std::complex<float>{static_cast<float>(data[i].real()),
-                                         static_cast<float>(data[i].imag())};
+  auto convertData = convertToComplex<float>(data, size);
```
schweitzpgi, line 259: but here `size` looks like it is the number of complex values in the list?

runtime/nvqir/NVQIR.cpp
annagrin: stateSize:)

runtime/nvqir/NVQIR.cpp
annagrin: size here is a number of qubits - I printed it out to confirm. And we are using it to allocate qubits above as well.

targettests/execution/from_state_double.cpp
```diff
@@ -0,0 +1,27 @@
...
+// CHECK: size 2
+
+int main() {
+  std::vector<double> vec{M_SQRT1_2, 0., 0., M_SQRT1_2};
+  auto counts = cudaq::sample(test, vec);
+  counts.dump();
+
+  printf("size %zu\n", counts.size());
+
```
schweitzpgi, line 26: nit: return 0; ?

Issue #1734, opened 2024-05-30
draw() : Python and C++ not doing the same for Euler's number


freifrauvonbleifrei:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

As part of working on #1723, I wanted to mirror the `draw()` integration test in `draw_tester.cpp`:
https://github.com/NVIDIA/cuda-quantum/blob/1edc8a858e6f2edd754afb61b3439f733f52b251/unittests/integration/draw_tester.cpp#L21-L94

But I am having two problems here: 
  * First, I cannot find how to call the adjoint function to `rz`, used in kernel `bar`. The line `cudaq.adjoint(rz, np.pi, qvec[2])` throws a jit compilation error. Is this intended? There appears to be no `rzdg` function, and of course one can write a wrapper, but that would be a lot uglier than in C++ . So I hope there already is a good solution?
  * In the test code, if you comment the offending line (for adjoint `rz`, see above), then the `rx` gates on qubit 0 are not printed at all in the string output. They are displayed if I replace `np.e` by `np.pi`.

### Steps to reproduce the bug

https://github.com/freifrauvonbleifrei/cuda-quantum/blob/96d573106fee3edcaf7edfb6f7187cdf86181b7d/python/tests/display/test_draw.py#L15-L84

### Expected behavior

Python test should produce the same string representation of the kernel as the C++ test.

But! here is the failing assertion 
```
E       AssertionError: assert '     ╭───╮  ...           \n' == '     ╭───╮  ...           \n'
E         
E         Skipping 802 identical leading characters in diff, use -v to show
E           
E         - ╭───────╮             ╭─────╮              
E         - ┤>      ├─────────────┤ sdg ├──────────────
E         + ╭───────╮╭───────────╮    ╭─────╮   ╭────────────╮
E         + ┤>      ├┤ rx(2.718) ├────┤ sdg ├───┤ rx(-2.718) ├...
E         
[...]

PR #1756, opened 2024-06-04
`PyASTBridge` : add more numpy constants to `visit_Attribute`

freifrauvonbleifrei:
### Description

related to #1734 , such that Euler's number can be resolved (and an error is thrown if no named math constant is found).


freifrauvonbleifrei:
@annagrin feel free to have a look, it's just a small change ;)

python/tests/display/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+# # This source code and the accompanying materials are made available under #
+# the terms of the Apache License 2.0 which accompanies this distribution. #
+# ============================================================================ #
+
+add_test(
```
amccaskey, line 8: someone else can correct me if I'm wrong, but I'm not sure we need this CMakeLists file. The CI already globs up all  pytests in the `tests/*` folder so this new `test_draw.py` file should be run already. See here https://github.com/NVIDIA/cuda-quantum/blob/main/.github/workflows/test_in_devenv.yml#L209

freifrauvonbleifrei: This makes sense. I had tried out a weird workaround for running this test apparently -- deleted!

python/tests/visualization/test_draw.py
```diff
@@ -24,13 +24,9 @@ def test_draw():
 
     @cudaq.kernel
     def bar(qvec: cudaq.qview):
-        # FIXME https://github.com/NVIDIA/cuda-quantum/issues/1734
-        # rx(np.e, qvec[0])
-        rx(2.71828182845904523536028, qvec[0])
+        rx(np.e, qvec[0])
```
annagrin, line 27: Can we add a test for `euler_gamma` as well? and maybe a negative test for something that is not supported?

khalatepradnya: Addressed in 96d4ad8310c72779d7a22a262c8c24f026d21c46

annagrin (approved):
Thanks Pradnya, LGTM!

Issue #1265, opened 2024-02-21
[RFC] Standardize the option names in nvq++


schweitzpgi:
At the moment, we have a lot of command line options in nvq++ that take both the form `-my-option` and `--my-option`. We also have options that are strictly only `-my-single-option` or `--my-double-option`.

While that all functions correctly, there hasn't really been any coherent guidance on this. As such users may be surprised when the driver accepts one spelling or another then rejects something that seems very similar.

Some strategies:

- Use 1 dash for distinct options. Self-explanatory.
- Use 1 dash short options with (optional) 2 dash long spelling of _equivalent_ options. `-o` and `--object-file` for instance.
    - Long options without short equivalents could either be 1 or 2 dashes
- Use 1 dash for all options and use a prefix letter to categorize related groups. For example, clang and gcc use `-f<option>` for controls relating to optimizations, etc., and `-W<option>` to control diagnostics.

schweitzpgi:
Proposal:

Remove:
  --execution-manager[=<name>]
  -em[=<name>]
  --target
  --emit-qir
  --emulate
  --disable-qubit-mapping
  --mapping-file
  --platform[=<name>]
  --load[=<name>]
  --clang-verbose

Rename:
  -disable-qubit-mapping -> -fdisable-qubit-mapping
  --save-temps -> -save-temps 
  --[no-]device-code-loading -> -f[no-]device-code-loading
  --[no-]unwind-lowering -> -f[no-]unwind-lowering
  --[no-]kernel-execution -> -f[no-]kernel-execution
  --[no-]aggressive-early-inline -> -f[no-]agressive-early-inline
  --[no-]apply-specialization -> -f[no-]apply-specialization
  --[no-]lambda-lifting -> -f[no-]lambda-lifting

bmhowe23:
If we're removing `--target`, does that mean we would expect the user to use `-target=<target>` and `--list-targets`? That seems inconsistent to me.

Would all of these changes go through a deprecation period so that users would have time to adapt their build scripts?

bmhowe23:
As noted in https://github.com/NVIDIA/cuda-quantum/pull/1757#discussion_r1626169772, we should also update Python to match nvq++ whenever this issue is worked.

PR #1757, opened 2024-06-04
Make Python support `--target=<target>` option (with `=`)

### LINKED ISSUE #1745: Unable to use all GPUs with Python and `nvidia-mgpu` target
### Required prerequisites

### Describe the bug

The [documentation ](https://nvidia.github.io/cuda-quantum/latest/using/backends/simulators.html#multi-node-multi-gpu) states that the `nvidia-mgpu` target can be used to distribute the state vector between available GPUs. The state vector seems to be correctly distributed when using CUDA-Q for C++. However, I'm unable to achieve such a distribution using CUDA-Q for Python. 

I'm using the Singularity image on a machine with two Tesla V100S GPUs.
Create a `cuda-quantum.def` file with the following content:
```
Bootstrap: docker

From: nvcr.io/nvidia/quantum/cuda-quantum:0.7.0
%runscript
    mount devpts /dev/pts -t devpts
    cp -r /home/cudaq/* .
    bash
```
Build the image with `singularity build --fakeroot cuda-quantum.sif cuda-quantum.def`
Run the image with `singularity shell --nv --no-mount hostfs ./cuda-quantum.sif
`
Create a file `cuquantum_backends.py` copying the code from the [Python examples](https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/examples/python/cuquantum_backends.py).
Create a file `cuquantum_backends.cpp` copying the code from the [C++ examples](https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/examples/cpp/basics/cuquantum_backends.cpp
).
Open another terminal to observe the output of `watch -n 1 nvidia-smi`. 

#### 30 qubits
Edit the two files to use 30 qubits (`qubit_count = 30` and `auto counts = cudaq::sample(/*shots=*/100, ghz{}, 30);`).

Compile and run the C++ version with the following commands:
```
nvq++ cuquantum_backends.cpp -o nvidia-mgpu.x --target nvidia-mgpu
mpirun -np 2 ./nvidia-mgpu.x
```
I can see a process for each GPU, peaking at about 8660MiB VRAM each.

Run the Python version with the following command:
```
mpirun -np 2 python3 cuquantum_backends.py --target=nvidia-mgpu
```
I can see two processes both running on GPU 0, peaking at 8528MiB VRAM each.

#### 31 qubits
Edit the two files to use 31 qubits (`qubit_count = 31` and `auto counts = cudaq::sample(/*shots=*/100, ghz{}, 31);`).

Compile and run the C++ version with the following commands:
```
nvq++ cuquantum_backends.cpp -o nvidia-mgpu.x --target nvidia-mgpu
mpirun -np 2 ./nvidia-mgpu.x
```
I can see a process for each GPU, peaking at about 16854MiB VRAM each.

Run the Python version with the following command:
```
mpirun -np 2 python3 cuquantum_backends.py --target=nvidia-mgpu
```
One process runs using GPU 0, peaking at about 16720MiB VRAM. The other process is terminated with the following error `RuntimeError: [custatevec] %out of memory in addQubitsToState (line 210)`.

### Expected behavior

I would expect the Python version to distribute the state vector among multiple GPUs, like the C++ version, rather than being limited to the memory of one GPU.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.0 (also 0.7.1, with a slightly different memory consumption)
- **Python version**: 3.10.12
- **Operating system**:  RHEL 9.3
- **nvidia-smi**:
```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.54.15              Driver Version: 550.54.15      CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  Tesla V100S-PCIE-32GB          Off |   00000000:01:00.0 Off |                    0 |
| N/A   33C    P0             35W /  250W |       0MiB /  32768MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  Tesla V100S-PCIE-32GB          Off |   00000000:C1:00.0 Off |                    0 |
| N/A   33C    P0             37W /  250W |       0MiB /  32768MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

### Suggestions

_No response_

bmhowe23:
Resolves #1745 

This makes it similar to the `nvq++` driver script so there isn't a discrepancy anymore.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -129,6 +129,14 @@ from .dbg import ast
 
 initKwargs = {}
 
+# Look for --target=<target> options
+for p in sys.argv:
+    split_params = p.split('=')
+    if len(split_params) == 2:
+        if split_params[0] in ['-target', '--target']:
+            initKwargs['target'] = split_params[1]
+
+# Look for --target <target> (with a space)
 if '-target' in sys.argv:
     initKwargs['target'] = sys.argv[sys.argv.index('-target') + 1]
 if '--target' in sys.argv:
--- a/python/tests/mlir/target/callable_kernel_arg.py
+++ b/python/tests/mlir/target/callable_kernel_arg.py
@@ -9,6 +9,9 @@
 # RUN: PYTHONPATH=../../.. python3 %s 
 # RUN: PYTHONPATH=../../.. python3 %s --target quantinuum --emulate 
 
+# Perform a single test with --target=<target>
+# RUN: PYTHONPATH=../../.. python3 %s --target=quantinuum --emulate 
+
 import cudaq
 from typing import Callable
 


python/cudaq/__init__.py
```diff
@@ -129,6 +129,14 @@ def __clearKernelRegistries():
 
 initKwargs = {}
 
+# Look for --target=<target> options
```
schweitzpgi, line 132: Note: I'd like to move away from using both `-` and `--` as option prefixes in `nvq++`. The proposal in #1265 is to use a `-` in most cases.

schweitzpgi (approved):

python/cudaq/__init__.py
bmhowe23: Ah, yes.... I thought I remembered some activity or discussion about something related to this recently, and that issue was probably it. Yes, I imagine we should update this code (and the code right below it) as part of any future PR that is posted for that issue. I'll add a comment to that issue so we don't forget.

PR #1760, opened 2024-06-05
Update quake-dialect.md

johnmaxrin:
Fixed typo: replaced "Quake's memory model" with "QTX's memory model"

### Description
Fixed typo: replaced "Quake's memory model" with "QTX's memory model" (previously incorrectly stated as "Quake's memory model").

### DIFF
--- a/docs/sphinx/specification/quake-dialect.md
+++ b/docs/sphinx/specification/quake-dialect.md
@@ -28,7 +28,7 @@ greater, the extra qubits are considered controls.
 The main motivation behind Quake's value model is to directly expose
 quantum and classical data dependencies for optimization purposes,
 i.e., to represent the dataflow in quantum computations.  In contrast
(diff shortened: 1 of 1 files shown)


johnmaxrin:

johnmaxrin:
recheck

docs/sphinx/specification/quake-dialect.md
```diff
@@ -28,7 +28,7 @@ greater, the extra qubits are considered controls.
 The main motivation behind Quake's value model is to directly expose
 quantum and classical data dependencies for optimization purposes,
 i.e., to represent the dataflow in quantum computations.  In contrast
-to Quake's memory model, which uses memory semantics (quantum
```
amccaskey, line 31: Quake is flexible in that it can represent both the memory and value semantics model (switching via the --memtoreg pass). So the text as-is is correct. Granted, the whole document here could use a refresh probably.

PR #1761, opened 2024-06-05
Add ability to set MPS SVD algorithm

1tnguyen:
### Description

This PR:

- Makes the SVD algorithm (to be forwarded to cutensornet) a configurable parameter (via environment variables)

- Updates the MPS simulator docs accordingly.


runtime/nvqir/cutensornet/tensornet_state.cpp
```diff
@@ -321,10 +321,29 @@ TensorNetState::factorizeMPS(int64_t maxExtent, double absCutoff,
...
+  // Check whether we need host memory workspace
+  int64_t hostWorkspaceSize{0};
+  HANDLE_CUTN_ERROR(cutensornetWorkspaceGetMemorySize(
+      m_cutnHandle, workDesc, CUTENSORNET_WORKSIZE_PREF_RECOMMENDED,
+      CUTENSORNET_MEMSPACE_HOST, CUTENSORNET_WORKSPACE_SCRATCH,
+      &hostWorkspaceSize));
+
+  void *hostWork = nullptr;
+  if (hostWorkspaceSize > 0) {
+    hostWork = malloc(hostWorkspaceSize);
```
bmhowe23, line 333: It would probably be a good idea to check for a nullptr and throw an error message / exception in that case (or perform error mitigation if it's easy).

1tnguyen: It's a good point. Added a null check as bellow (it's best to just abort here since we don't really have an alternative).

runtime/nvqir/cutensornet/simulator_mps_register.cpp
```diff
@@ -65,6 +67,29 @@ class SimulatorMPS : public SimulatorTensorNetBase {
       m_relCutoff = relCutoff;
       cudaq::info("Setting MPS relative cutoff to {}.", m_relCutoff);
     }
+    static const std::unordered_map<std::string, cutensornetTensorSVDAlgo_t>
+        g_stringToAlgoEnum{{"GESVD", CUTENSORNET_TENSOR_SVD_ALGO_GESVD},
+                           {"GESVDJ", CUTENSORNET_TENSOR_SVD_ALGO_GESVDJ},
+                           {"GESVDP", CUTENSORNET_TENSOR_SVD_ALGO_GESVDP},
+                           {"GESVDR", CUTENSORNET_TENSOR_SVD_ALGO_GESVDR}};
```
schweitzpgi, line 74: This could be a `constexpr` table stored with the keys sorted. One can use `std::lower_bound` to find the key without any difference in performance (other than the table doesn't need to be built at load time).

runtime/nvqir/cutensornet/tensornet_state.cpp
```diff
@@ -321,10 +321,34 @@ TensorNetState::factorizeMPS(int64_t maxExtent, double absCutoff,
     throw std::runtime_error("ERROR: Insufficient workspace size on Device!");
   }
 
+  // Check whether we need host memory workspace
+  int64_t hostWorkspaceSize{0};
```
schweitzpgi, line 325:
```suggestion
  std::int64_t hostWorkspaceSize{0};
```

runtime/nvqir/cutensornet/simulator_mps_register.cpp
1tnguyen: I gave it a shot in https://github.com/NVIDIA/cuda-quantum/pull/1761/commits/2b6c85a8d2d2fc22573912900bd01ad920eea5b5. 
Could you please give it a second look :) ?

PR #1763, opened 2024-06-05
Blochsphere Visualization

### LINKED ISSUE #1639: Bloch sphere visualization for single-qubit operations
# Description
The CUDA-Q platform allows users to build and execute hybrid quantum-classical kernels.  While learning CUDA-Q and also while developing with CUDA-Q, it is helpful to visualize quantum circuits and states to gain intuition, to spot check code, and to communicate with others. We need to expand the visualization functionality into CUDA-Q. We would like to add functionality to visualize single qubit state vectors on a Bloch sphere.

## Scenarios
A developer who has defined a single qubit quantum kernel and has found the statevector after applying that kernel using `cudaq.get_state(kernel)` should now be able to view the vector representation on the Bloch sphere. Both the ability to graph multiple statevectors onto one Bloch sphere and multiple Bloch spheres side by side should also be made possible.

## Details

We would primarily like to make this available in Jupyter notebooks. For this issue, it is hence sufficient if this is supported in Python only (not in C++). To simplify development, we can take an optional dependency on QuTiP and its tools for [Bloch sphere plotting](https://qutip.org/docs/4.0.2/guide/guide-bloch.html) to add a new member function `add_to_bloch_sphere` on [cudaq.State](https://nvidia.github.io/cuda-quantum/0.7.1/api/languages/python_api.html#cudaq.State). This function should create a new Bloch sphere if no argument is given, and it should add the state vector to an existing Bloch sphere if a Bloch sphere object is passed as argument. 
The `add_to_bloch_sphere` function should check if the qutip package is installed and fail with a comprehensive error to please install qutip if it is not. It should also check if 
the state is a single qubit state and give a comprehensive error otherwise. 
Additionally, there should be a new `cudaq.show` function that takes one or more Bloch spheres and visualizes them side by side.

## Bonus

It would be nice if the Bloch sphere(s) could also be plotted [in LaTeX](https://tex.stackexchange.com/questions/345420/how-to-draw-a-bloch-sphere). However, this functionality is not expected for completion of this issue.

prateekchawla168:
### Description
Fixes  #1639 by adding a visualization for Bloch sphere for single-qubit states.
Significant changes:
- Added [QuTiP](https://qutip.org/) as an optional dependency in [pyproject.toml](pyproject.toml) 
- Added bloch_visualize module in [python/cudaq/visualization](python/cudaq/visualization/bloch_visualize.py) with code to handle Bloch sphere generation and management. Docstrings are contained within the code itself. Tested with Jupyter notebooks.
- Modified Python initialization [__init__.py](python/cudaq/__init__.py) to import QuTiP, and if successful, load the visualization functionality.
- REST server cannot be built on Singularity containers because CMake is never able to find ZLib/Minizip, especially if the program must search manually. Corrected the corresponding [CmakeLists.txt](runtime/cudaq/platform/default/rest_server/CMakeLists.txt) so it searches in the correct places.

This is a submission for Unitary Hack 2024. 


prateekchawla168:
recheck

prateekchawla168:

prateekchawla168:
recheck

pyproject.toml
```diff
@@ -35,7 +35,7 @@ Releases = "https://nvidia.github.io/cuda-quantum/latest/releases.html"
 # We must use h5py<3.11 because 3.11 doesn't include aarch64 Linux wheels.
 # https://github.com/h5py/h5py/issues/2408
 [project.optional-dependencies]
-chemistry = [ "scipy==1.10.1", "openfermionpyscf==0.5", "h5py<3.11" ]
+chemistry = [ "scipy==1.10.1", "openfermionpyscf==0.5", "h5py<3.11", "qutip>=5" ]
```
khalatepradnya, line 38: The new dependency is not strictly related to "chemistry", so better to separate it. Alternatively, you can remove it from dependency and rely on users to explicitly install it.

python/cudaq/__init__.py
```diff
@@ -41,6 +41,15 @@
 from .runtime.observe import observe
 from .mlir._mlir_libs._quakeDialects import cudaq_runtime
 
+try:
+    from qutip import Qobj,Bloch
+except ImportError:
+    raise Exception("The module QuTiP is not installed. Bloch Sphere visualization will not work.")
```
khalatepradnya, line 47: Do you need specific version of QuTip? If so, update the message accordingly, for example, `Please install QuTip v5.0+ to use this feature.`

khalatepradnya (commented):
Thank you for the contribution. This is looking good.
Recommend the following enhancements for completeness:

- Please add a unit test that can check the newly added APIs. You can add to existing suite in this file - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_kernel_features.py
- Add a notebook example to showcase the functionality. You can edit this - https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/examples/python/tutorials/quantum_bits.ipynb. (It currently has an image of Bloch sphere.)

khalatepradnya:
Can you add screenshots to this PR of how the output will look like for - (a) one Bloch sphere, (b) side by side rendering of multiple Bloch spheres?

python/cudaq/__init__.py
prateekchawla168: QuTiP <5 contains an alternate renderer of the Bloch sphere, called `qutip.Bloch3d`, based on Mayavi. This was removed in QuTiP v5.0.0b1 (see [qutip#2306](https://github.com/qutip/qutip/pull/2306)). That said, the current code is backwards compatible with previous versions of QuTiP as well because only the `qutip.Bloch` class is used here.  The only possible change used would be that in case a user is using QuTiP < 5, they will be able to pass a custom figure/axes in their implementation of `qutip.Bloch.render()`. This functionality was also removed in QuTiP v5.

With this background, I think it is good to keep the optional dependency `qutip>=5` for future-proofing in the [pyproject.toml](pyproject.toml), but not add the requirement explicitly while importing.

pyproject.toml
prateekchawla168: I've separated this to a new dependency in the new commit.

prateekchawla168:
Added functionality in the latest commit:

 - Added a separate dependency in [pyproject.toml](pyproject.toml) for qutip, and removed it from chemistry.
 - Added minor comment on how to install qutip in [Python init.py](python/cudaq/__init__.py).
 -  Minor bugfix in [bloch_visualize](python/cudaq/visualization). When a single-qubit state vector is passed to the function, the $\ell_2$-norm was not evaluated properly.
 - Added tests for visualization files in [python/tests/visualization](python/tests/visualization/test_bloch_sphere_visualization.py).
 - Added examples for visualization of Bloch sphere code in [quantum_bits.ipynb](blob/main/docs/sphinx/examples/python/tutorials/quantum_bits.ipynb)

python/cudaq/__init__.py
```diff
@@ -41,6 +41,16 @@
 from .runtime.observe import observe
 from .mlir._mlir_libs._quakeDialects import cudaq_runtime
 
+try:
+    from qutip import Qobj, Bloch
+except ImportError:
+    raise Exception(
```
amccaskey, line 47: maybe my Python is rusty but doesn't this exception here make the optional qutip dependency not optional... Everytime I `import cudaq` with this change I'll see an exception thrown. I don't think this is what we want.

prateekchawla168: Hmm, that's true. Would it be better if the import silently fails if the dependency isn't installed? That way the user will simply get an error that the relevant function is not contained in the module.

prateekchawla168:
Implemented @amccaskey 's suggestion: If `import qutip` fails, then a helpful error message is printed to the screen whenever the user attempts to call any of the visualization functions.

khalatepradnya:
Thanks for the updates and addressing review comments.
Can you fix the 2 CI failures (code formatting, and spell check)?

freifrauvonbleifrei:
Hi, a comment from my work on #1723 : I am introducing a folder `python/cudaq/display/` which semantically looks similar to your `python/cudaq/visualization/`. Maybe we could coordinate on a joint folder name.

prateekchawla168:
@freifrauvonbleifrei Absolutely! I had used the name `visualization` simply because I meant to put in multiple files there, as I had polanning planned to work on the $\LaTeX$ rendering next, but it seems like you've got it handled already, which is awesome!

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -198,10 +202,13 @@ lossy
 lvalue
 makefiles
 merchantability
+mlir
```
khalatepradnya, line 205: Suggest to use `MLIR` it's already in the list.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -198,10 +202,13 @@ lossy
...
 mutex
 namespace
 namespaces
 natively
+ncols
```
khalatepradnya, line 210: Use backticks `ncols` or spell out as `number of columns`

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -198,10 +202,13 @@ lossy
...
+nrows
```
khalatepradnya, line 211: Same as `ncols`

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -214,12 +221,14 @@ prepend
 preprocessor
 programmatically
 pybind
+quakeDialects
```
khalatepradnya, line 224: `Quake` is in the list use `Quake dialects` if required or escape with backticks.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -214,12 +221,14 @@ prepend
...
 quantize
 quantized
 qubit
 qubits
 qudit
 qudits
+qutip
```
khalatepradnya, line 231: Remove this and add `QuTiP`

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -263,4 +272,4 @@ upvote
 variadic
 variational
 verifier
-workflow
+workflow
```
khalatepradnya, line 275: Could be nothing, but, confirm there aren't odd whitespace / line ending changes.

khalatepradnya (changes requested):
Thanks for the updates. Requesting changes on spell check.

python/tests/visualization/test_bloch_sphere_visualization.py
```diff
@@ -0,0 +1,131 @@
...
+    # generate a random density matrix with qutip and add to sphere
+    sph.add_states( qutip.rand_dm(2) )
+    
+    b = cudaq.add_to_bloch_sphere(cudaq.get_state(single_qubit_kernel),existing_sphere=sph)
+    assert isinstance(b, qutip.Bloch)
+
+
+# TODO: refactor this part later, but run the same tests as before with density matrix backend
+
+cudaq.set_target("density-matrix-cpu")
```
khalatepradnya, line 75: This might affect the subsequent tests. Recommend to set / reset the target in the test fixture.

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
```diff
@@ -175,6 +175,215 @@
     "\n"
    ]
   },
+  {
```
khalatepradnya, line 178: This format is difficult to parse. Hence adding review comments in on this file one place here.

- You can install `QuTip` in this notebook. Add a cell `!pip install qutip>=5` at the beginning.
- The error message is not getting displayed if QuTip is no installed, instead following error shows up
```TypeError: install_qutip_request() takes 0 positional arguments but 1 was given```

File: python/tests/visualization/test_bloch_sphere_visualization.py [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+import os 
+import pytest
+import numpy as np
+
+import cudaq
+import qutip
```
khalatepradnya, line 14: Skip these tests if qutip module not available.
`qutip = pytest.importorskip("qutip")` before `import qutip`

freifrauvonbleifrei:
> @freifrauvonbleifrei Absolutely! I had used the name `visualization` simply because I meant to put in multiple files there, as I had planned to work on the LATEX rendering next, but it seems like you've got it handled already, which is awesome!

@prateekchawla168 I had chosen `display` instead of `visualize` because the amount of data shown is rather limited (both for the circuits and the bloch sphere), which better suits the [notion of display](https://eagereyes.org/blog/2012/data-display-vs-data-visualization). If you agree with the naming, feel free to rename your folder to avoid merge conflicts further upstream :)

prateekchawla168:
Changes in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855) : 

- Fixed spellings in `spelling_allowlist.txt`:
      - Changed `bloch` to `Bloch`
      - Removed `kwargs` 
      - Removed `libs` and corrected in file
      - Removed `mlir`
      - Removed `nrows` and `ncols`
      - Removed `quakeDialects`
      - Replaced `qutip` with `QuTiP`
      
- Reset target to original after density matrix-based visualization tests are completed in [test_bloch_sphere_visualization.py](python/tests/visualization/test_bloch_sphere_visualization.py).
- Added statement to install QuTiP in [quantum_bits.ipynb](docs/sphinx/examples/python/tutorials/quantum_bits.ipynb)
- Skipped Bloch sphere tests if QuTiP module is not available

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -108,6 +108,7 @@ arity
 auxillary
 backend
 backends
+bloch
```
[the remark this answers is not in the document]
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -187,9 +188,12 @@ inlined
 inlining
 instantiation
 instantiations
+interoperational
 intrinsics
 iterable
 iteratively
+kwargs
```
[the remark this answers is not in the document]
prateekchawla168: [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855)

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -187,9 +188,12 @@ inlined
...
+libs
```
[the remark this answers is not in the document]
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

.github/workflows/config/spelling_allowlist.txt
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855). I think it was probably a whitespace, but I couldn't find anything odd here.

python/tests/visualization/test_bloch_sphere_visualization.py
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855). I've added a line to reset the target after the  density matrix tests are completed.

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855). I've added a line to install `QuTiP` in the notebook, and also fixed the function to display the error each time if `QuTiP` is not installed. Currently the function simply asks to install `QuTiP` each time the visualization code is called, but it doesn't exit the kernel by throwing an exception. This is by design, so the user doesn't lose progress and the kernel doesn't accidentally die of a thrown exception. So if this is done in a loop, it will show multiple errors.

python/tests/visualization/test_bloch_sphere_visualization.py
prateekchawla168: Fixed in [a3964e2](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/a3964e284910f99cbbddddfa7493747ab9735855).

prateekchawla168:
> > @freifrauvonbleifrei Absolutely! I had used the name `visualization` simply because I meant to put in multiple files there, as I had planned to work on the LATEX rendering next, but it seems like you've got it handled already, which is awesome!
> 
> @prateekchawla168 I had chosen `display` instead of `visualize` because the amount of data shown is rather limited (both for the circuits and the bloch sphere), which better suits the [notion of display](https://eagereyes.org/blog/2012/data-display-vs-data-visualization). If you agree with the naming, feel free to rename your folder to avoid merge conflicts further upstream :)

@freifrauvonbleifrei You're right, it does make sense because we aren't displaying much data, but I'd argue here that showing the quantum circuit that's being executed and the state of the qubit(s) at some point of the execution is pretty much most of the data that can be shown to the user anyway. I don't know of ways by which we can show 2-qubit information (except, perhaps multiple graphs showing the amplitude/phase information in a chosen basis set). Even if we decide to go down that road in the future, it does make sense that we'd be transitioning from display to visualization anyway. This led to why I chose `visualization` instead of `display` - why not choose a future-proof name at the beginning?

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
khalatepradnya: I tried out this notebook locally, even after executing the cell which install QuTip, I'm getting following error -
```The module QuTiP was not detected. Bloch Sphere visualization will not work.
Please install QuTip to use this feature. You can run "pip install qutip" or in a conda environment, run "conda -c conda-forge install qutip" to install qutip. 
```

khalatepradnya (changes requested):
Thank you the updates. Requesting changes for completion.

python/tests/visualization/test_bloch_sphere_visualization.py
```diff
@@ -0,0 +1,134 @@
...
+        cudaq.add_to_bloch_sphere(np.array([1,0]))
+
+def test_visualization_invalid_state():
+    with pytest.raises(Exception) as err:
+        cudaq.add_to_bloch_sphere(cudaq.get_state(two_qubit_kernel) )
+        
+
+def test_visualization_single_qubit_no_sphere():
+    b = cudaq.add_to_bloch_sphere(cudaq.get_state(single_qubit_kernel))
+    assert isinstance(b, qutip.Bloch)
```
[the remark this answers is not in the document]
prateekchawla168: I'm not very sure I understand your comment, exactly. Do you mean that the test must raise an exception (as in, it must fail), or do you mean it as a formatting issue?

If it is the latter, the `assert` statement will anyway throw an error. If it's the former, well, the docstring for the `cudaq.add_to_bloch_sphere()` states that `If a valid qutip.Bloch object is not supplied, then creates a new sphere with the vector representing the supplied state.` Supplying a `None` variable counts as a case of invalid object, so the function will create and return a new  `qutip.Bloch` object. I do not quite see why an exception should be raised here.

prateekchawla168:
Changes in [e64fd71](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/e64fd71fed9adbd17e8521c5e76ae66ea0c27b05) : 
- Sorted [spelling_allowlist.txt](https://github.com/NVIDIA/cuda-quantum/blob/main/.github/workflows/config/spelling_allowlist.txt)
- Edited example notebook to clarify about installing Qutip.

prateekchawla168:
@khalatepradnya  I couldn't reply to [your comment](https://github.com/NVIDIA/cuda-quantum/pull/1763#discussion_r1633566093) on [quantum_bits.ipynb](docs/sphinx/examples/python/tutorials/quantum_bits.ipynb) as a conversation, but whenever you install/uninstall some package inside a Jupyter kernel, you must restart the kernel for the package changes to take effect. I have added a comment in the file regarding this in [e64fd71](https://github.com/NVIDIA/cuda-quantum/pull/1763/commits/e64fd71fed9adbd17e8521c5e76ae66ea0c27b05). I have verified that restarting the kernel works in both Jupyter and VSCode environments.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -108,6 +108,7 @@ arity
 auxillary
 backend
 backends
+Bloch
```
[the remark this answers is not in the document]
prateekchawla168: Thanks! This script was really helpful! I have fixed the sorting issues in a recent commit. Hope it works!

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
```diff
@@ -175,6 +175,228 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let us try to showcase the functionality to render such a 3D representation with Cuda-Q."
```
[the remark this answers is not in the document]
prateekchawla168: I have updated this in the same commit as the spelling changes.

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
prateekchawla168: Hi, you need to restart the notebook kernel after executing the install statement, else it will not load the packages. This is unfortunately a limitation of Jupyter. I've added a comment in the code informing the user to do this as well.

python/tests/visualization/test_bloch_sphere_visualization.py
```diff
@@ -0,0 +1,147 @@
...
+    y(qubit)
+    z(qubit)
+    t(qubit)
+    s(qubit)
+
+    mz(qubit)
+
+
+@cudaq.kernel
+def two_qubit_kernel():
```
khalatepradnya, line 45:
```
@cudaq.kernel
def two_qubit_kernel():
    qvector = cudaq.qvector(2)
    x(qvector[0])
    ry(np.pi / 2., qvector[1])
    mz(qvector)
```

freifrauvonbleifrei:
> > > @freifrauvonbleifrei Absolutely! I had used the name `visualization` simply because I meant to put in multiple files there, as I had planned to work on the LATEX rendering next, but it seems like you've got it handled already, which is awesome!
> > 
> > 
> > @prateekchawla168 I had chosen `display` instead of `visualize` because the amount of data shown is rather limited (both for the circuits and the bloch sphere), which better suits the [notion of display](https://eagereyes.org/blog/2012/data-display-vs-data-visualization). If you agree with the naming, feel free to rename your folder to avoid merge conflicts further upstream :)
> 
> @freifrauvonbleifrei You're right, it does make sense because we aren't displaying much data, but I'd argue here that showing the quantum circuit that's being executed and the state of the qubit(s) at some point of the execution is pretty much most of the data that can be shown to the user anyway. I don't know of ways by which we can show 2-qubit information (except, perhaps multiple graphs showing the amplitude/phase information in a chosen basis set). Even if we decide to go down that road in the future, it does make sense that we'd be transitioning from display to visualization anyway. This led to why I chose `visualization` instead of `display` - why not choose a future-proof name at the beginning?

@prateekchawla168  Thank you for your thoughts on this. I think it is very likely that more complex simulation/measurement data for larger experiments and higher-order correlations will be visualized in/for CUDA-Q at some point (maybe even interactive visualization). It feels like a separate place would make sense for this "non-analytic" data--visualization` would be open for that exact purpose then, if we don't clutter it now.

pyproject.toml
```diff
@@ -35,7 +35,8 @@ Releases = "https://nvidia.github.io/cuda-quantum/latest/releases.html"
 # We must use h5py<3.11 because 3.11 doesn't include aarch64 Linux wheels.
 # https://github.com/h5py/h5py/issues/2408
 [project.optional-dependencies]
-chemistry = [ "scipy==1.10.1", "openfermionpyscf==0.5", "h5py<3.11" ]
+chemistry = [ "scipy==1.10.1", "openfermionpyscf==0.5", "h5py<3.11"  ]
+visualization = [ "qutip>=5" ]
```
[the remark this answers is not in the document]
prateekchawla168: Hi, I've added Matplotlib as a dependency here. 
Normally it is installed while installing qutip, but it may not get installed in some edge case or some minimal system, so I had not put it in. 😅

python/tests/visualization/test_bloch_sphere_visualization.py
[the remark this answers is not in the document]
prateekchawla168: I've simplified the test kernels as you suggested, and I've also ensured that kernel is called and the state is created in a different line, to improve readability (and possibly help with debugging this).

python/tests/visualization/test_bloch_sphere_visualization.py
prateekchawla168: I've simplified the kernel as requested.

python/tests/visualization/test_bloch_sphere_visualization.py
```diff
@@ -0,0 +1,147 @@
...
+def do_something():
+    yield
+    cudaq.__clearKernelRegistries()
+
+
+# write sample kernels
+
+
+@cudaq.kernel
+def single_qubit_kernel():
```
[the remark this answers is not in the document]
prateekchawla168: Simplified the kernel in a new commit

python/cudaq/visualization/bloch_visualize.py
```diff
@@ -0,0 +1,121 @@
...
+    if (not isinstance(existing_sphere, Bloch)) and (not existing_sphere
+                                                     == None):
+        existing_sphere = None
+        print(
+            "Existing sphere object is not a valid Bloch sphere. A new sphere will be created."
+        )
+
+    b = Bloch(**kwargs) if existing_sphere == None else existing_sphere
+    st_rep = np.array(psi)
+    if (st_rep.shape == (2,) and isclose(abs(st_rep.dot(st_rep))**2, 1.0)) or (
```
khalatepradnya, line 54: This check seems too restrictive, is it imposed by QuTip?
One reason for the intermittent test failures is that the randomly generated angles may cause the value of `abs(st_rep.dot(st_rep))**2` be much less than 1.0 and throwing exception.

khalatepradnya (approved):
Thank you for the contribution.
One open Q in comments.
CI failure will be addressed separately (see 6328008dc5d27b025a7c443df6db19c88354c6c8)

docs/sphinx/examples/python/tutorials/quantum_bits.ipynb
khalatepradnya: CI failures owing to Jupyter notebook not auto-reloading.

python/cudaq/visualization/bloch_visualize.py
prateekchawla168: Actually, this is imposed by quantum mechanics. Let me explain. There can be two cases where a single-qubit state is returned by the kernel : Either it is a state vector or a density matrix.

[line 54]
- If it is a state vector, then the state must be normalized as it is necessarily a pure state. In that case, $\lvert \langle \psi | \psi \rangle \rvert^2$ must be 1. This is calculated with `abs(st_rep.dot(st_rep))**2`, where `st_rep` is the NumPy array representation of the  statevector $\psi$ returned by the simulator. 
    -  The 3 random rotations in the example code of [quantum_bits example](docs/sphinx/examples/python/tutorials/quantum_bits.ipynb) are simply unitary single-qubit rotations, which cannot affect the normalization.
    - If norm of the state is much less than 1, it means one of two things:
         -  Some nonunitary operation is being applied which is affecting the normalization of the state. In this case, the user shouldn't be using state vectors anyway, as the output will not be valid state vectors. The user must use density matrices to obtain the correct output.
         - The state was not normalized properly to start with, which is also bad. A single-qubit state expressed as a statevector must be normalized as it is necessarily a pure state, or if one wishes to use a mixed state, they should be using density matrices in the first place.

[line 55]
- If it is a density matrix, then a single qubit density matrix must have a trace of 1. This is checked by `isclose(st_rep.trace().real, 1.0))`.

prateekchawla168: Actually, this is imposed by quantum mechanics. Let me explain. There can be two cases where a single-qubit state is returned by the kernel : Either it is a state vector or a density matrix.

[line 54]
- If it is a state vector, then the state must be normalized as it is necessarily a pure state. In that case, $\lvert \langle \psi | \psi \rangle \rvert^2$ must be 1. This is calculated with `abs(st_rep.dot(st_rep))**2`, where `st_rep` is the NumPy array representation of the  statevector $\psi$ returned by the simulator. 
    -  The 3 random rotations in the example code of [quantum_bits example](docs/sphinx/examples/python/tutorials/quantum_bits.ipynb) are simply unitary single-qubit rotations, which cannot affect the normalization.
    - If norm of the state is much less than 1, it means one of two things:
         -  Some nonunitary operation is being applied which is affecting the normalization of the state. In this case, the user shouldn't be using state vectors anyway, as the output will not be valid state vectors. The user must use density matrices to obtain the correct output.
         - The state was not normalized properly to start with, which is also bad. A single-qubit state expressed as a statevector must be normalized as it is necessarily a pure state, or if one wishes to use a mixed state, they should be using density matrices in the first place.

[line 55]
- If it is a density matrix, then a single qubit density matrix must have a trace of 1. This is checked by `isclose(st_rep.trace().real, 1.0))`.

khalatepradnya: Thanks for the explanation. There is an error in the formula that computes norm.
```suggestion
    if (st_rep.shape == (2,) and isclose(
            np.sqrt(abs(st_rep.dot(st_rep.conjugate()))), 1.0,
            abs_tol=1e-6)) or (st_rep.shape == (2, 2) and
                               isclose(st_rep.trace().real, 1.0)):
```
Please fix it. Also, recommend reducing tolerance to account for precision error(s), if any.

Thanks @justinlietz for your inputs!

prateekchawla168: Hi, I've added the conjugation operation while taking the dot product and the absolute tolerance in the new commit, but I'm not very sure about the need for taking square root. I mean, if $\lvert\langle \psi \mid \psi\rangle \rvert^2 \approx 1 $, then this implies that $\lvert\langle \psi \mid \psi\rangle \rvert \approx 1$. This is due to the fact that since Hilbert spaces are inner product spaces, $\langle \psi \mid \psi\rangle \geq 0$ is guaranteed due to positive definiteness and conjugate symmetry, so $\lvert\langle \psi \mid \psi\rangle \rvert \neq -1$ for a pure state. The only real purpose the `abs()` is serving is to quickly convert the `complex` output of the dot product into a `float` type. It also helps with the accidental edge case of float representation errors where a number too close to 0 might be represented as a negative number. Adding the square root for a comparison to 1 feels a bit unnecessary and uses up more CPU cycles for the same output.

khalatepradnya: I think I follow why you could drop the square root in this case, but, I don't understand why square is necessary... 

```
isclose(abs(st_rep.dot(st_rep.conjugate())), 1.0, abs_tol=1e-6)
```

prateekchawla168: Hmm, you're right. I'd kept the square because $\lvert\langle \psi \mid \psi\rangle \rvert^2$ is related to a physical quantity, and it's indeed not quite needed when you're testing for pure states. I've removed it in the new commit and replaced with just $\lvert\langle \psi \mid \psi\rangle \rvert$

khalatepradnya: Thanks!

khalatepradnya:
@prateekchawla168: Please add a comment on the issue https://github.com/NVIDIA/cuda-quantum/issues/1639 so it can be assigned to you.

PR #1768, opened 2024-06-06
Fixes #1712: failures when using libc++

### LINKED ISSUE #1712: Failing tests when building against libc++
### Required prerequisites

### Describe the bug

CUDA-Q currently is built against libstdc++. Building against a different standard library generally requires updates to the compiler. 
Specifically, when building against libc++ (built from source along with the other LLVM dependencies), the following llvm-lit tests fail:

In test folder:
- [ ] AST-Quake/reverse.cpp
- [ ] AST-Quake/vector_ctor_initlist.cpp
- [ ] AST-Quake/vector_ctor_initlist_int.cpp
- [ ] AST-Quake/vector_ctor_sized.cpp
- [ ] AST-Quake/vector_front_back.cpp

In targettests folder:
- [ ] execution/mapping_test-1-cpp17.cpp
- [ ] execution/mapping_test-1.cpp
- [ ] execution/qir_cond_for_loop-3.cpp
- [ ] execution/sim_gate_timing-cpp17.cpp
- [ ]  execution/sim_gate_timing.cpp
- [ ] CUDAQ-Target :: Kernel/inline-qpu-func.cpp (fails with a type traversal issue in TraverseStmt)

In the unittests folder:
- [ ] tensornet_CustomUnitaryTester.checkSimple and the for tensornet_mps (a unitary is created in a cudaq lib compiled against libc++ and then currently passed directly to the simulator .so compiled against libstdc++ - we need to properly "convert")

Check out the PR branch https://github.com/NVIDIA/cuda-quantum/pull/1603, and remove the excludes listed under this bug in docker/build/assets.Dockerfile. 

### Expected behavior

All tests should pass. 

Possibly, but we haven't tested for libc++ so far

### Environment

ghcr.io/nvidia/cuda-quantum-assets docker image

### Suggestions

_No response_

schweitzpgi:
- Adds code to the bridge to handle differences in the standard header implementations
- Modifies tests that used unsupported features

Fixes #1712

### DIFF
--- a/lib/Frontend/nvqpp/ConvertDecl.cpp
+++ b/lib/Frontend/nvqpp/ConvertDecl.cpp
@@ -238,6 +238,12 @@ bool QuakeBridgeVisitor::interceptRecordDecl(clang::RecordDecl *x) {
       TODO_x(toLocation(x), x, mangler, "std::string type");
       return false;
     }
+    if (name.equals("__wrap_iter")) {
+      auto *cts = cast<clang::ClassTemplateSpecializationDecl>(x);
+      if (!TraverseType(cts->getTemplateArgs()[0].getAsType()))
+        return false;
+      return true;
+    }
     if (name.equals("pair")) {
       auto *cts = cast<clang::ClassTemplateSpecializationDecl>(x);
       SmallVector<Type> members;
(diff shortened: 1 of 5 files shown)


schweitzpgi:
See #1603

bmhowe23 (approved):
Thanks, Eric

PR #1771, opened 2024-06-07
Fixing state handling experimental branch Publishing 

1tnguyen:
### Description

- Some downstream updates are required to maintain backward compatibility with some existing Publishing validation. Hence, pump the commit hash.

- Modify our asset injection into the Python wheel to be generic: don't enforce .config and .so coupling, use the file extension to install the asset file to the proper location (`lib` or `targets`).

- Handle the newly-added `nvidia-mqpu-mps` target in the experimental branch in the validation script.

Publishing pipeline: https://github.com/NVIDIA/cuda-quantum/actions/runs/9421685906/


bmhowe23 (approved):
Thanks, Thien

PR #1780, opened 2024-06-08
Fix `targettests` `RemoteSim/get_state`

1tnguyen:
### Description

The state data from the remote target is double precision. When the local sim is single-precision, we need to do the conversion.

This didn't show up in CI b/c we use `qpp` (double precision) as the default host simulator. 

(I thought I had committed these changes but apparently, it was lost...)


bmhowe23 (approved):
Thanks, Thien. I think this explains at least one of the issues w/ our nightly integration tests after the stateHandling merge. https://github.com/NVIDIA/cuda-quantum/actions/runs/9425701729/job/25967900102

schweitzpgi (approved):

PR #1786, opened 2024-06-10
[unittest] Remove std parallel execution policy.

boschmitt:
### Description

In `stdlib++` the use of parallel exection policy with standard algortihms requires the TBB. As we don't want to add another dependecy and our codebase already uses OpenMP, its better to either find a way of forcing `stdlib++` to use OpenMP backend or simply not use these policies.

Since this particular use of parallelism seems unnecessary (the size of the state vectors being generated are quite small), we opt to remove its use.


1tnguyen (approved):
Thanks, @boschmitt!

PR #1787, opened 2024-06-10
Minor change for consistency in Doxygen config

duydl:
### Description

After building cudaq with the provided dev container, I tried running the `build_docs.sh` script and got this error: 
```
error: problems opening rtf file 'registry_8h_source.rtf' for reading
error: An error occurred during post-processing the RTF files!
Failed to generate documentation using doxygen.
Doxygen exit code: 1
Building CUDA-Q documentation using Sphinx...
Documentation generation failed with exit code 11.
```
The error might not matter as sphinx only uses the XML output, but changing the GENERATE_RTF tag removed the problem for me. And the default value in description is NO anyway.


duydl:

PR #1789, opened 2024-06-10
Add a pass to assign unique ID attributes to qubit allocations

atgeller:
### Description

The `AssignIDs` pass adds attributes to `alloca` QIR instructions that attaches a unique ID attribute. The unique ID is then updated based on the size of the allocation. Unique ID attributes are also added to references to the allocated qvector based on the index of the reference.

Uncommenting #lib/Optimizer/Transforms/AssignIDs.cpp:61 will display the allocations with attached attributes, and similar uncommenting #lib/Optimizer/Transforms/AssignIDs.cpp:70 will display the references with attached attributes.


lib/Optimizer/Transforms/AssignIDs.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+//===----------------------------------------------------------------------===//
+// Generated logic
+//===----------------------------------------------------------------------===//
+namespace cudaq::opt {
+#define GEN_PASS_DEF_ASSIGNIDS
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+namespace {
```
schweitzpgi, line 30: nit: move namespace to below this function

lib/Optimizer/Transforms/AssignIDs.cpp
```diff
@@ -0,0 +1,116 @@
...
+static int allocaSize(quake::AllocaOp alloc) {
+  if (isa<quake::RefType>(alloc.getType()))
+    return 1;
+  if (auto veqTy = dyn_cast<quake::VeqType>(alloc.getType()))
+    return veqTy.getSize();
+  else
```
schweitzpgi, line 36: nit: coding convention. https://www.llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

lib/Optimizer/Transforms/AssignIDs.cpp
```diff
@@ -0,0 +1,116 @@
...
+#define GEN_PASS_DEF_ASSIGNIDS
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+namespace {
+static int allocaSize(quake::AllocaOp alloc) {
+  if (isa<quake::RefType>(alloc.getType()))
+    return 1;
+  if (auto veqTy = dyn_cast<quake::VeqType>(alloc.getType()))
+    return veqTy.getSize();
```
schweitzpgi, line 35: Should use hasSpecifiedSize() instead.

Note that not having a constant/specific size does NOT mean the vector has no size at all.

schweitzpgi (commented):
It should be clear that every qubit that is allocated has an intrinsic identity which is conveyed unambiguously by the SSA value of the operation. Aggregate allocations have identity as well, with each qubit having a distinct offset within the sequence.

schweitzpgi:
See #1864 which provides a way to "assign integers" in the value semantics form of the IR and in a way that will be consistent with the rest of the IR.

atgeller:
This was addressed by #2058

PR #1791, opened 2024-06-10
Fixes for backward compatibility with NVQC

1tnguyen:
### Description

The `ExecutionContext::simulationData` (pre-state handling) type change from `std::tuple<std::vector<std::size_t>, std::vector<std::complex<double>>>` to `std::unique_ptr<SimulationState>` needs to be handle properly during JSON serialization.

In particular, an old/existing server instance adds `simulationData` with empty dim when no state data is available rather than non-data as we switch to the `unique_ptr` data type. 

### Tested by

- Manually run the `nvqc_integration_docker_test` of `nvqc_regression_tests.yml` with the locally-build cuda-q client against the prod nvqc service. 

Notes: 

- Will monitor wheel check post-merge.

- The docker test does include Python examples.

### DIFF
--- a/runtime/common/JsonConvert.h
+++ b/runtime/common/JsonConvert.h
@@ -151,18 +151,23 @@ inline void from_json(const json &j, ExecutionContext &context) {
     j["simulationData"]["dim"].get_to(stateDim);
     j["simulationData"]["data"].get_to(stateData);
 
-    // Create the simulation specific SimulationState
-    auto *simulator = cudaq::get_simulator();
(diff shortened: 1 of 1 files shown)


schweitzpgi (approved):
Thanks for the quick fix.

1tnguyen:
Note: the NVQC regression pipeline has passed with this change: https://github.com/NVIDIA/cuda-quantum/actions/runs/9458875060/
Latest nightly image (`nvcr.io/nvidia/nightly/cuda-quantum:latest`) contains the fix.

PR #1793, opened 2024-06-10
Fix #1705

### LINKED ISSUE #1705: `nvq++` does not respect user's choice for output object filename
### Required prerequisites

### Describe the bug

When compile object files using `nvq++`, the driver does not honors the filename provided by the user.

Try compiling an object file using `nvq++`:
```sh
nvq++ -c foo.cpp -o my_file.o
```

The output will be an object file named `foo.o` instead of `my_file.o`

### Expected behavior

`nvq++` should generate an object file with name `my_file.o`

Not a regression

### Environment

- **CUDA Quantum version**: nvq++ Version proto-0.8.0 (https://github.com/NVIDIA/cuda-quantum 6016e2f23216b64f0fca083bef22fb8266ea9049)
- **C++ compiler**: c++ (GCC) 13.2.1 20240316 (Red Hat 13.2.1-7)
- **Operating system**: Linux fedora 6.8.9-200.fc39.x86_64 

### Suggestions

_No response_

schweitzpgi:
Merge object files into a file as named by the user on the command line.

Add test.

### DIFF
--- a/test/NVQPP/obj_file.cpp
+++ b/test/NVQPP/obj_file.cpp
@@ -1,0 +1,14 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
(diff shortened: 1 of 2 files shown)


khalatepradnya (approved):
LGTM

Issue #1475, opened 2024-04-04
[RFC] Unitary Synthesis


khalatepradnya:
#### Problem 
Given a user provided arbitrary quantum unitary, synthesize it into a sequence of quantum gates.

#### Expectations
- User provides an arbitrary unitary matrix as a custom quantum operation. 
- The custom operation can be used as a regular CUDA-Q supported quantum operation. 
  - Q: Broadcast (same operation on multiple qubits): Out of scope
- The allowed set of quantum gates for synthesis depends on the backend target.
  - Q: Allow user to specify set of allowed gates: Out of scope
- CUDA-Q throws error if a unitary cannot be synthesized (reasonably).
  - 'reasonably' to account for time limit (timeout), gate count limit (upper threshold), and how close the synthesized "circuit" is to the input unitary (tolerance)

#### User API
- Python
```
import cudaq 

custom_h = cudaq.register_operation(1. / np.sqrt(2.) *  np.array([[1, 1], [1, -1]])) 
custom_x = cudaq.register_operation(np.array([[0, 1], [1, 0]])) 

@cudaq.kernel 
def bell(): 
  qubits = cudaq.qvector(2) 
  custom_h(qubits[0]) 
  custom_x.ctrl(qubits[0], qubits[1]) 

counts = cudaq.sample(bell) 
counts.dump()

```

- C++
```
// Macro to specify the custom unitary operation
cudaq_register_operation(custom_h,
                         (std::vector<std::vector<std::complex<double>>>{
                             {M_SQRT1_2, M_SQRT1_2}, {M_SQRT1_2, -M_SQRT1_2}}));
cudaq_register_operation(
    custom_x, (std::vector<std::vector<std::complex<double>>>{{0, 1}, {1, 0}}));

void custom_operation() __qpu__ {
  cudaq::qvector qubits(2);
  custom_h(qubits[0]);
  custom_x.ctrl(qubits[0], qubits[1]);
}

int main() {
  auto result = cudaq::sample(custom_operation);
  std::cout << result.most_probable() << '\n';
  return 0;
}
```

- The user must provide valid unitary matrix (CUDA-Q will not check / enforce this requirement)
- Ordering: The user provided matrix must be in row-major format
- Endianness: The user provided matrix is interpreted as little-endian 

#### Constraints
- Size of unitary matrix: limit to 8 qubits, (2^8 = 256), 256 x 256
[...]

PR #1794, opened 2024-06-10
Enable custom unitary operations in MLIR mode

khalatepradnya:
Towards issue https://github.com/NVIDIA/cuda-quantum/issues/1475.

-------------
This PR covers the support custom operations in simulation for Python & C++ 
It builds on the work previously merged https://github.com/NVIDIA/cuda-quantum/pull/1708

------
Not covered in this PR: 

- Decomposition of the custom operation into discrete gates (synthesis)
- Full support of parameterized custom operations


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1157,4 +1157,12 @@ def ZOp : OneTargetOp<"z", [Hermitian]> {
   }];
 }
 
+def CustomUnitarySymbolOp : QuakeOperator<"unitary"> {
```
schweitzpgi, line 1160: Will this replace the Op in the other PR then?

schweitzpgi (commented):
Need a roundtrip test added for new op, of course.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
khalatepradnya: Yes, that is the intention. I will add a comment there.

khalatepradnya:
> Need a roundtrip test added for new op, of course.

Addressed in https://github.com/NVIDIA/cuda-quantum/pull/1794/commits/e315589931a651caf6a7399da6c03723e4016310

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1595,6 +1611,43 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+        }
+        return count;
+      }();
+      buildOp<quake::CustomUnitarySymbolOp>(
+          builder, loc, args, negations, reportNegateError, isAdjoint,
+          isControl, paramCount, targetCount, true);
+      auto &customUnitary = builder.getBlock()->back();
+      auto srefAttr =
+          SymbolRefAttr::get(StringAttr::get(builder.getContext(), opName));
+      customUnitary.setAttr("generator", srefAttr);
```
amccaskey, line 1647: @schweitzpgi Are you cool with this (note it was my idea)? This was the only way I could figure out how to add a `SymbolRefAttr` to this operation, and inherit from `QuakeOperator` in the tablegen so we get all the semantics of a quake op.

schweitzpgi: Ideally we'd have a distinct argument in the `inp` list for it. (Adding a layer of safety, since it can't be omitted or lost should the op get cloned for example.)  I think we can do that, but I have to do a little experimenting.

schweitzpgi: Patches sent.

khalatepradnya: Addressed in 3dba0f518f7e5e9c62fc8296e31e867152dd0753

include/cudaq/Frontend/nvqpp/ASTBridge.h
```diff
@@ -144,18 +144,19 @@ class QuakeBridgeVisitor
...
-                              clang::CompilerInstance &ci,
-                              clang::ItaniumMangleContext *mangler)
+  explicit QuakeBridgeVisitor(
+      clang::ASTContext *astCtx, mlir::MLIRContext *mlirCtx,
+      mlir::OpBuilder &bldr, mlir::ModuleOp module, SymbolTable &symTab,
+      EmittedFunctionsCollection &funcsToEmit,
+      llvm::ArrayRef<clang::Decl *> reachableFuncs,
+      MangledKernelNamesMap &namesMap, clang::CompilerInstance &ci,
+      clang::ItaniumMangleContext *mangler,
+      llvm::SmallVector<llvm::StringRef> &customOperations)
```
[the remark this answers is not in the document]
amccaskey: Since we are identifying custom operations by their extern "C" generator functions, and those functions are marked with a new function attributed defined in qubit_qis.h, we decided to follow the same pattern the bridge takes for finding QPU kernels. There we find the FunctionDecls marked with __qpu__ and add them to the list to be processed to MLIR. It is the QPUCodeFinder that does that, not the QuakeBridgeVisitor, so we have to send the data over to the QuakeBridgeVisitor. Then we have something to check for when we see an operation that is not in the default set. 

Definitely open to any simpler ways you can think of though.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,79 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
                                          reportNegateError, isAdjoint,
                                          isControl, /*paramCount=*/3);
 
+    // See if this is a custom unitary.
+    std::string maybeUnitaryGenerator = funcName.str() + "_generator_";
+    // Extract number of targets
+    size_t targetCount = 0;
```
schweitzpgi, line 1619:
```suggestion
    std::size_t targetCount = 0;
```

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,79 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+    }
+    if (targetCount > 0) {
+      // Extract number of parameters
+      std::size_t paramCount = [&]() {
+        std::size_t count = 0;
+        for (auto arg : args) {
+          auto argTy = arg.getType();
+          if (isa<FloatType>(argTy))
+            count++;
+          else if (auto ptrTy = dyn_cast<cc::PointerType>(argTy)) {
```
schweitzpgi, line 1637:
```suggestion
          if (isa<FloatType>(argTy)) {
            count++;
          } else if (auto ptrTy = dyn_cast<cc::PointerType>(argTy)) {
```

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,79 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+            if (isa<FloatType>(ptrTy.getElementType()))
```
schweitzpgi, line 1638: I think passing parameters by reference (or pointer) should be a hard error here and unsupported. We want these functions to be pure.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,79 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+        return count;
+      }();
+      auto srefAttr = SymbolRefAttr::get(
+          StringAttr::get(builder.getContext(), genFuncName));
+      ValueRange operands(args);
+      assert(operands.size() >= 1 && "must be at least 1 operand");
+      if ((operands.size() == 1) &&
+          operands[0].getType().isa<quake::VeqType>()) {
+        auto target = operands[0];
+        if (!negations.empty())
```
schweitzpgi, line 1651: Rather than block copying buildOp, buildOp should be generalized to allow for the symbol to be an optional argument.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,79 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+              negations.end())
+            reportNegateError();
+        auto negs =
+            negatedControlsAttribute(builder.getContext(), ctrls, negations);
+        SmallVector<Value> params;
+        for (auto p : operands.take_front(paramCount)) {
+          if (p.getType().isa<cudaq::cc::PointerType>()) {
+            params.push_back(builder.create<cudaq::cc::LoadOp>(loc, p));
+          }
+        }
```
schweitzpgi, line 1682:
```suggestion
        for (auto p : operands.take_front(paramCount)) 
          if (p.getType().isa<cudaq::cc::PointerType>()) 
            params.push_back(builder.create<cudaq::cc::LoadOp>(loc, p));
```

runtime/nvqir/NVQIR.cpp
```diff
@@ -569,6 +569,48 @@ void __quantum__qis__exp_pauli(double theta, Array *qubits, char *pauliWord) {
 
 void __quantum__rt__result_record_output(Result *, int8_t *) {}
 
+void __quantum__qis__custom_unitary(std::complex<double> *unitary,
+                                    Array *controls, Array *targets) {
+  auto ctrlsVec = arrayToVectorSizeT(controls);
+  auto tgtsVec = arrayToVectorSizeT(targets);
+  auto numQubits = tgtsVec.size();
+  auto nToPowTwo = (1ULL << numQubits);
```
bmhowe23, line 577: It would probably be good to throw a `std::invalid_argument()` exception if `numQubits >= 64' to make sure this doesn't overflow. It might save some debugging time in the future if the memory ever gets corrupted.

runtime/nvqir/NVQIR.cpp
```diff
@@ -569,6 +569,48 @@ void __quantum__qis__exp_pauli(double theta, Array *qubits, char *pauliWord) {
...
+                                                             ctrlsVec, tgtsVec);
+}
+
+void __quantum__qis__custom_unitary__adj(std::complex<double> *unitary,
+                                         Array *controls, Array *targets) {
+
+  auto ctrlsVec = arrayToVectorSizeT(controls);
+  auto tgtsVec = arrayToVectorSizeT(targets);
+  auto numQubits = tgtsVec.size();
+  auto nToPowTwo = (1ULL << numQubits);
```
bmhowe23, line 591: Same here

lib/Frontend/nvqpp/ASTBridge.cpp
```diff
@@ -223,6 +225,11 @@ class QPUCodeFinder : public clang::RecursiveASTVisitor<QPUCodeFinder> {
       return true;
     func = func->getDefinition();
     if (func) {
+      if (auto attr = func->getAttr<clang::AnnotateAttr>())
+        if (attr->getAnnotation().str() == "user_custom_quantum_operation") {
```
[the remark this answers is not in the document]
schweitzpgi: This string should be moved to a file, say, `cudaq/Frontend/nvqpp/AttributeNames.h`. Same for the one in `isQuantum()` below.

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
...
+        cudaq::opt::QIRArrayCreateArray, qirArrayTy,
+        {rewriter.getI32Type(), rewriter.getI64Type()}, parentModule);
+    FlatSymbolRefAttr getSymbolRef =
+        cudaq::opt::factory::createLLVMFunctionSymbol(
+            cudaq::opt::QIRArrayGetElementPtr1d, i8PtrTy,
+            {qirArrayTy, rewriter.getIntegerType(64)}, parentModule);
+    Value zero = rewriter.create<arith::ConstantIntOp>(loc, 0, 64);
+    Value one = rewriter.create<arith::ConstantIntOp>(loc, 1, 64);
+    // FIXME: 8 bytes is assumed to be the sizeof(char*) on the target machine.
+    Value eight = rewriter.create<arith::ConstantIntOp>(loc, 8, 32);
```
schweitzpgi, line 1240: Can we make both arguments 64 bit in the call below?

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
+    auto parentModule = op->getParentOfType<ModuleOp>();
+    auto context = op->getContext();
+    auto typeConverter = this->getTypeConverter();
+
+    auto numParameters = op.getParameters().size();
+    if (numParameters != 0) {
+      op.emitOpError("Parameterized custom operations not yet supported.");
+    }
```
schweitzpgi, line 1266:
```suggestion
    if (numParameters) 
      op.emitOpError("Parameterized custom operations not yet supported.");
```

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -7,6 +7,7 @@
  ******************************************************************************/
 
 #include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
```
schweitzpgi, line 10: From the changes, I don't see why this is needed. It's a bit odd to include dialect B into the implementation specifics for dialect A.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -28,6 +28,9 @@ namespace cudaq::opt {
 using namespace mlir;
 
 namespace {
+
+static constexpr const char ReadOnlySuffix[] = ".rodata";
```
schweitzpgi, line 32: This change to omit any uniquing will result in collisions in the symbol namespace as it is entirely possible to have more than 1 array being constructed in a kernel.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -65,10 +68,11 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
     Value conArr;
     Value conGlobal;
     if (toGlobal) {
-      static unsigned counter = 0;
+      // static unsigned counter = 0;
       auto ptrTy = cudaq::cc::PointerType::get(arrTy);
       // Build a new name based on the kernel name.
-      std::string name = funcName + ".rodata_" + std::to_string(counter++);
+      /// FIXME: Add a unique identifier per set of parameters
```
schweitzpgi, line 74: Any parameters, for the transformation to proceed, will be indistinguishable from other constant data.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1605,6 +1613,77 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+        SmallVector<Value> params;
+        for (auto p : operands.take_front(paramCount)) 
+          if (p.getType().isa<cudaq::cc::PointerType>()) 
+            params.push_back(builder.create<cudaq::cc::LoadOp>(loc, p));
+        builder.create<quake::CustomUnitarySymbolOp>(
+            loc, srefAttr, isAdjoint, params, ctrls, targets, negs);
+      }
+      return true;
+    }
+
```
schweitzpgi, line 1686: Where is the handling of complex expressions?

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
schweitzpgi: This probably needs to return here instead of generating bad code.

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
...
+      auto complex64Ty =
+          typeConverter->convertType(ComplexType::get(rewriter.getF64Type()));
+      auto complex64PtrTy = LLVM::LLVMPointerType::get(complex64Ty);
+      Type type = getTypeConverter()->convertType(globalOp.getType());
+      auto addrOp =
+          rewriter.create<LLVM::AddressOfOp>(loc, type, generatorName);
+      auto unitaryData =
+          rewriter.create<LLVM::BitcastOp>(loc, complex64PtrTy, addrOp);
+
+      auto qirFunctionName = "__quantum__qis__custom_unitary";
```
schweitzpgi, line 1327: These qis function names should be put in cudaq/Optimizer/CodeGen/QIRFunctionNames.h

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
...
+            loc, qirArrayTy, concatFunc, ArrayRef<Value>{controlArr, backArr});
+        controlArr = glue.getResult();
+      }
+    }
+    // Fetch the unitary matrix generator for this custom operation
+    auto sref = op.getGenerator();
+    StringRef generatorName = sref.getRootReference();
+    auto globalOp =
+        parentModule.lookupSymbol<cudaq::cc::GlobalOp>(generatorName);
+    if (globalOp) {
```
schweitzpgi, line 1317:
```suggestion
    if (!globalOp)
      return op.emitOpError("global not found for custom op"); 
```
then can unindent the rest of this function.

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
     return success();
   }
 };
+
+class CustomUnitaryOpRewrite
```
schweitzpgi, line 1218: Add a comment here that this pattern is reached if and only if the target is a simulator. The QIS functions being generated here are nonsensical on hardware.

We probably want to test that explicitly in fact.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -321,6 +325,30 @@ class ComplexCreatePattern : public OpRewritePattern<complex::CreateOp> {
...
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(quake::CustomUnitarySymbolOp customOp,
+                                PatternRewriter &rewriter) const override {
+    auto sref = customOp.getGenerator();
+    StringRef generatorName = sref.getRootReference();
+    auto parentModule = customOp->getParentOfType<ModuleOp>();
+
+    auto ccGlobalOp = parentModule.lookupSymbol<cudaq::cc::GlobalOp>(
+        generatorName.str() + ReadOnlySuffix);
```
schweitzpgi, line 341: This lookup isn't going to be general enough because when we lift array values, we don't just look for the special unitary functions. And we cannot restrict it that way.

This is a general pass and shouldn't be a one-off just for unitary matrices. So we need a new, separate pass to rewrite the custom unitaries.

I think the algorithm here has to be:
  - look up the generator function, which should be the symbol before we rewrite anything.
  - scan the function's body for the address_of operation and the symbol it is using.
  - use that symbol to replace the current op with a new custom op. (yes, we want to replace it, otherwise we can have mysterious crashes in the transformer)

python/runtime/mlir/py_register_dialects.cpp
```diff
@@ -13,13 +13,15 @@
 #include "cudaq/Optimizer/CodeGen/Passes.h"
 #include "cudaq/Optimizer/CodeGen/Pipelines.h"
 #include "cudaq/Optimizer/Dialect/CC/CCDialect.h"
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
 #include "cudaq/Optimizer/Dialect/CC/CCTypes.h"
 #include "cudaq/Optimizer/Dialect/Quake/QuakeDialect.h"
 #include "cudaq/Optimizer/Dialect/Quake/QuakeTypes.h"
 #include "cudaq/Optimizer/Transforms/Passes.h"
 #include "mlir/InitAllDialects.h"
```
schweitzpgi, line 22:
```suggestion
```
nit: no whitespace

schweitzpgi (commented):
I didn't go over the python side of this.

schweitzpgi:
Generalities: the macro is defining a "generator" function and these are being translated exactly as if they were kernels. They are not.

Since they aren't kernels
  - they should not appear in the mangled name map for the module
  - they should not have attributes of `cudaq-entrypoint` or `cudaq-kernel`
  - the generation of the C++ side function does not make sense; GKE should not be rewriting these to call `altLaunchKernel` and doing so may cause issues
  - - eventually these functions will necessarily need to be specialized before constant folding

lib/Frontend/nvqpp/ASTBridge.cpp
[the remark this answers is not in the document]
khalatepradnya: I tried that. These strings are needed in `runtime/cudaq/qis/qubit_qis.h` where the attribute is defined. Okay to modify CMake in `runtime`  to include `Frontend` directory?

runtime/nvqir/NVQIR.cpp
khalatepradnya: Addressed in b3678f7324f795a6f48455904c5789c27033b255

lib/Frontend/nvqpp/ASTBridge.cpp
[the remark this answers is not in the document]
amccaskey: You'll need it defined in both qubit_qis.h and in the Frontend, we don't want a dependency on the Frontend in the runtime. This is what we do for the `__qpu__` tag

include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -36,6 +36,7 @@ createApplyOpSpecializationPass(bool computeActionOpt);
 std::unique_ptr<mlir::Pass> createDelayMeasurementsPass();
 std::unique_ptr<mlir::Pass> createExpandMeasurementsPass();
 std::unique_ptr<mlir::Pass> createLambdaLiftingPass();
+std::unique_ptr<mlir::Pass> createLiftArrayAllocPass();
```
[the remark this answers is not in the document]
khalatepradnya: Addressed in 133c0a29cf1ffe1a4e5681d4f6814dcd4efdadf8

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
khalatepradnya: Python test present - python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
khalatepradnya: Addressed in 9ded69b1946782c6a7e42876173de7eabf75ca68

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
khalatepradnya: Leftover code. Removed in 98e4ceb295f76eb76396c571a6b44e9a91b55562

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Will be handled later.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: The `create` function for `CustomUnitarySymbolOp` differs which resulted in multiple conditions in the `buildOp` function, hence kept it separate here.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
khalatepradnya: Addressed in d16052be0a40ca2b4f8ce19b8b8b9df53cd8a537

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: It differs by one argument, right?

That difference is small as compared to block copying all this functionality into a method that is already too large.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: Does this mean in a separate PR? (This PR is already overly large, so I hope the answer is yes.)

lib/Frontend/nvqpp/ASTBridge.cpp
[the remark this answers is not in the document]
schweitzpgi: This is exactly the reason why these strings should be in a header file and not just hard coded in random files. The latter is bad coding practice.

I thought the runtime already included files from the compiler, but maybe I am wrong about that.

Open to any suggestions on where to put universal constants, including string data.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: yes, I will add in follow-up PR.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Addressed in ec6973b042fd692236c417feacf35ba893e2b5dd.

lib/Frontend/nvqpp/ASTBridge.cpp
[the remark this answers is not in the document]
khalatepradnya: I'm leaning towards handling this separately outside of this PR, okay if I create an issue to address this and reference it here?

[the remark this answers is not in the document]
khalatepradnya: https://github.com/NVIDIA/cuda-quantum/issues/1911

include/cudaq/Frontend/nvqpp/ASTBridge.h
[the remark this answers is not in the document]
khalatepradnya: This was updated in the latest commits. We do use the additional argument for custom operations, but, we are now using a map of the generator function and its mangled name (no longer using `cudaq::unitary_operation` since it cannot be handled by kernel (yet).
The general comment to handle macro expansion separately from `__qpu__` still applies and needs refactoring which I will do as a follow-up.

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1213,6 +1214,137 @@ class RemoveRelaxSizeRewrite : public OpConversionPattern<quake::RelaxSizeOp> {
     return success();
   }
 };
+
+/// NOTE: This class currently targets simulator backends. On hardware targets
+/// the custom operations ought to be decomposed by a separate pass and should
+/// never reach here.
+class CustomUnitaryOpRewrite
```
amccaskey, line 1221: We should probably add a FileCheck test for this in `test/Quake-QIR`

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+#define DEBUG_TYPE "get-concrete-matrix"
+
+using namespace mlir;
+
+namespace {
+
+class CustomUnitaryPattern
```
amccaskey, line 32: Documentation / comments throughout this file would be helpful

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+    auto generator = customOp.getGenerator();
+
+    auto parentModule = customOp->getParentOfType<ModuleOp>();
+    auto funcOp = parentModule.lookupSymbol<func::FuncOp>(generator);
+    if (!funcOp) {
+      return failure();
+    }
+
+    StringRef concreteMatrix;
+    mlir::Region &body = funcOp.getBody();
```
amccaskey, line 49: Shouldn't this just be a `funcOp.walk([&](AddressOfOp op) {...})`

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
```diff
@@ -0,0 +1,95 @@
...
+      for (auto &op : block.getOperations()) {
+        auto addrOp = dyn_cast<cudaq::cc::AddressOfOp>(op);
+        if (addrOp) {
+          concreteMatrix = addrOp.getGlobalName();
+          break;
+        }
+      }
+    }
+    if (concreteMatrix.empty()) {
+      return failure();
```
amccaskey, line 60: `return op.emitError("message")`

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
khalatepradnya: Addressed in e137ebef99049d0cd363821a857064b683840069

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
khalatepradnya: I didn't know this was possible, I will update. Thanks!

khalatepradnya: Addressed in 85acc1fe90e553571cf17f1a04e52afa6172b1a8

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
khalatepradnya: Addressed in 85acc1fe90e553571cf17f1a04e52afa6172b1a8

lib/Optimizer/Transforms/GetConcreteMatrix.cpp
khalatepradnya: Addressed in 85acc1fe90e553571cf17f1a04e52afa6172b1a8

amccaskey (commented):
I'm good with this when Eric is. Just be sure that whatever we have here lends itself to the parameterized unitary implementation.

bmhowe23 (commented):
This is looking good to me.
Just to confirm - it looks like the unitary operation is always stored in double precision. And I assume that if we're running with single precision simulator, there must be a conversion step occurring during simulation. Is that right?

python/tests/synthesis/test_custom_operations.py
```diff
@@ -0,0 +1,197 @@
...
+        toffoli(q[0], q[1], q[2])
+
+    counts = cudaq.sample(test_toffoli)
+    print(counts)
+    assert counts["110"] == 1000
+
+
+# NOTE / [SKIP_TEST]: The following test crashes in the 'Validate Python wheel (amd64 / x86)'
+# stage on Ubuntu, RedHat and OpenSuse for 'tensornet' and 'tensornet-mps' backends
+# (works on Debian and Fedora, and on all for arm64 in CI, and locally).
```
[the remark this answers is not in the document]
khalatepradnya: I don't yet. I'm checking if this is a generic problem unrelated to this PR (and I happened to stumble on it) or something I introduced. Will add a new issue accordingly.

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
```diff
@@ -27,6 +27,8 @@ static constexpr const char QIRCphase[] = "__quantum__qis__cphase";
 static constexpr const char QIRReadResultBody[] =
     "__quantum__qis__read_result__body";
 
+static constexpr const char QIRCustomOp[] = "__quantum__qis__custom_unitary";
```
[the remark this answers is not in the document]
khalatepradnya: Isn't the current check at line#550 sufficient? Sorry, I don't follow where to put the additional check...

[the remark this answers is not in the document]
bmhowe23: That check on line 550 allows `__quantum_` functions to be called because they are part of the QIR standard, but this new functions is - to my knowledge - not really supported by any QIR-enabled backend, so I think it should be disallowed.
So that if check should probably be updated to something like `if (!funcName.startswith("__quantum_") || funcName.equals(QIRCustomOp))`.

python/tests/remote/test_remote_platform.py
```diff
@@ -399,6 +401,20 @@ def exp_kernel():
     count = cudaq.sample(exp_kernel)
 
 
+def test_arbitrary_unitary_synthesis():
```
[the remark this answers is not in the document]
khalatepradnya: Addressed in b7389a5cc01c92b9941f231e00269bc10e3be796

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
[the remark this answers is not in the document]
khalatepradnya: Understood. Thanks. Updated in b7389a5cc01c92b9941f231e00269bc10e3be796

khalatepradnya:
> Just to confirm - it looks like the unitary operation is always stored in double precision. And I assume that if we're running with single precision simulator, there must be a conversion step occurring during simulation. Is that right?

Yes, it is implicit. Tried to check that with this test - https://github.com/khalatepradnya/cuda-quantum/blob/cpp_mlir_custom_2/python/tests/synthesis/test_custom_operations.py#L112, running same application on different simulators.
Is that sufficient?

python/tests/remote/test_remote_platform.py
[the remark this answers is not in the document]
khalatepradnya: Updated in ddba3cf.
New issue - https://github.com/NVIDIA/cuda-quantum/issues/1924

schweitzpgi (commented):
This is looking good, Pradnya. Thanks for doing all the heavy lifting.

I'm good with the C++ side. I understand there are some outstanding issues. I know some/all of them are logged as issues.

I did not look too closely at the python language frontend. So if someone else takes a look at that, go ahead and approve this.

khalatepradnya:
> This is looking good, Pradnya. Thanks for doing all the heavy lifting.
> 
> I'm good with the C++ side. I understand there are some outstanding issues. I know some/all of them are logged as issues.
> 
> I did not look too closely at the python language frontend. So if someone else takes a look at that, go ahead and approve this.

Thanks, Eric!
I think @amccaskey did that... or perhaps @bmhowe23 ?

bmhowe23 (approved):
👍

schweitzpgi (approved):
Approval on C++ changes

PR #1797, opened 2024-06-11
Add a strong validation for qubit allocation batching 

1tnguyen:
### Description

The simulator expects linear state expansion (incremental qubit indices) for these `allocateQubits` calls. 
Hence, add a validation check in the `DefaultExecutionManager`, i.e., the one for simulator backends.


bmhowe23 (approved):
Looks like this PR may have slipped through the cracks (or perhaps it was overridden by another PR?). In any case, assuming it merges cleanly, it is still wanted, and the final tests run correctly, it looks good to me.

1tnguyen:
> Looks like this PR may have slipped through the cracks (or perhaps it was overridden by another PR?). In any case, assuming it merges cleanly, it is still wanted, and the final tests run correctly, it looks good to me.

Thanks, Ben! 
Indeed, it slipped through the cracks. This was to address a post-merge comment for the state handling work.

PR #1800, opened 2024-06-11
Add JSON serializers/deserializers for optimizers and gradients

bmhowe23:
## Description

`cudaq::optimizer` and `cudaq::gradient` are abstract virtual classes, so they need special handling for JSON serialization and deserialization. This PR adds that support.

These changes should be backwards compatible with the existing NVQC deployment as long as the new `std::optional<RestRequestOptFields> opt` field in the `RestRequest` is not populated. (And it is indeed not populated in any of our current code.)

This is all in support of remote VQE/optimizer execution on NVQC.

### DIFF
--- a/runtime/common/JsonConvert.h
+++ b/runtime/common/JsonConvert.h
@@ -290,40 +290,42 @@ inline void from_json(const nlohmann::json &j,
 inline std::unique_ptr<cudaq::optimizer>
 make_optimizer_from_json(const nlohmann::json &j,
                          const OptimizerEnum optimizer_type) {
-  if (optimizer_type == OptimizerEnum::COBYLA) {
+  switch (optimizer_type) {
+  case OptimizerEnum::COBYLA: {
     auto ret_ptr = std::make_unique<cudaq::optimizers::cobyla>();
-    from_json(j, dynamic_cast<cudaq::optimizers::base_nlopt &>(*ret_ptr));
+    from_json(j, *ret_ptr);
     return ret_ptr;
   }
-  if (optimizer_type == OptimizerEnum::NELDERMEAD) {
+  case OptimizerEnum::NELDERMEAD: {
     auto ret_ptr = std::make_unique<cudaq::optimizers::neldermead>();
-    from_json(j, dynamic_cast<cudaq::optimizers::base_nlopt &>(*ret_ptr));
+    from_json(j, *ret_ptr);
     return ret_ptr;
   }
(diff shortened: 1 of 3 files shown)


runtime/common/JsonConvert.h
```diff
@@ -178,10 +197,236 @@ inline void from_json(const json &j, ExecutionContext &context) {
...
+  if (optimizer_type == OptimizerEnum::GRAD_DESC) {
+    auto ret_ptr = std::make_unique<cudaq::optimizers::gradient_descent>();
+    from_json(j, dynamic_cast<cudaq::optimizers::BaseEnsmallen &>(*ret_ptr));
+    return ret_ptr;
+  }
+  if (optimizer_type == OptimizerEnum::SGD) {
+    auto ret_ptr = std::make_unique<cudaq::optimizers::sgd>();
+    from_json(j, dynamic_cast<cudaq::optimizers::BaseEnsmallen &>(*ret_ptr));
+    return ret_ptr;
+  }
```
1tnguyen, line 327: nit: is there a reason we don't use `switch case` here?

runtime/common/JsonConvert.h
```diff
@@ -178,10 +197,236 @@ inline void from_json(const json &j, ExecutionContext &context) {
...
+    auto ret_ptr = std::make_unique<cudaq::gradients::forward_difference>();
+    from_json(j,
+              dynamic_cast<cudaq::gradients::forward_difference &>(*ret_ptr));
+    return ret_ptr;
+  }
+  if (gradient_type == GradientEnum::PARAMETER_SHIFT) {
+    auto ret_ptr = std::make_unique<cudaq::gradients::parameter_shift>();
+    from_json(j, dynamic_cast<cudaq::gradients::parameter_shift &>(*ret_ptr));
+    return ret_ptr;
+  }
```
1tnguyen, line 390: same question as https://github.com/NVIDIA/cuda-quantum/pull/1800/files#r1635394341

1tnguyen (approved):
LGTM 👍

runtime/common/JsonConvert.h
bmhowe23: Updated in e8a16e4db. I also removed the unnecessary dynamic_cast just to tidy it up.

runtime/common/JsonConvert.h
bmhowe23: Updated in e8a16e4db. I also removed the unnecessary dynamic_cast just to tidy it up.

PR #1801, opened 2024-06-11
[rt] Move `measureSpinOp` impl. to the simulator.

boschmitt:
### Description

Measuring a `cudaq::spin_op` operation requires knowledge about the inner workings to the backend simulator, specifically we need to guarantee that the operation are applied and not left in the operation queue. To decouple the execution menager from the simulator, we move the implementation of this method to the simulator base class.


amccaskey (approved):
+1

PR #1653, opened 2024-05-10
[StateHandling] State API for remote platforms 

1tnguyen:
### Description

State handling for remote platforms:

- Introduce `SimulationState` subclass for remote platform (`RemoteSimulationState` and `PyRemoteSimulationState`) which is essentially a holder of the kernel IR.

- Perform a lazy evaluation to resolve the state:
    
    (1) For state accessor, e.g., `amplitudes`: adding a configurable qubit number threshold. If the number of qubits is less than the threshold, resolve the full state vector to the client. Otherwise, send on amplitude access request. Adding an API to `cudaq::state` for batched amplitude access.
    
    (2) For overlap, sending job requests with both kernel IRs.
    
     
**Note**: the remote state is implemented for MLIR kernels only (C++ MLIR mode, kernel builder and Python).
Library mode C++ kernel is evaluated as-is (v0.7 mode): state is resolved to a vector by the server for each `get_state` call.

PR #1803, opened 2024-06-11
Remote state handling rebase

1tnguyen:
### Description

State handling for remote platforms:

Introduce `SimulationState` subclass for remote platform (`RemoteSimulationState` and `PyRemoteSimulationState`) which is essentially a holder of the kernel IR.

Perform a lazy evaluation to resolve the state:

(1) For state accessor, e.g., amplitudes: adding a configurable qubit number threshold. If the number of qubits is less than the threshold, resolve the full state vector to the client. Otherwise, send on amplitude access request. Adding an API to `cudaq::state` for batched amplitude access.

(2) For overlap, sending job requests with both kernel IRs.

Note: the remote state is implemented for MLIR kernels only (C++ MLIR mode, kernel builder and Python).
Library mode C++ kernel is evaluated as-is (v0.7 mode): state is resolved to a vector by the server for each `get_state` call.

Adjust the GPU backend validation: skip a few long running examples for cutensornet-based backends. This brings total GPU validation time to ~2hr.

This is migrated from https://github.com/NVIDIA/cuda-quantum/pull/1653: revise the implementation to make it compatible with pre-existing NVQC services.

### Compatibility

Except for **remote** state overlap and amplitude accessor (large state), it is compatible with existing NVQC service.

### CI 

- Publishing: https://github.com/NVIDIA/cuda-quantum/actions/runs/9476728549

(image: `ghcr.io/nvidia/cuda-quantum:workflow-stateHandling`)

- NVQC regression run: https://github.com/NVIDIA/cuda-quantum/actions/runs/9478140066
(testing against the existing prod NVQC service)

- Nightly integration: https://github.com/NVIDIA/cuda-quantum/actions/runs/9478073356
(deploy `ghcr.io/nvidia/cuda-quantum:workflow-stateHandling` as a test function)


python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -321,6 +418,22 @@ index pair.
...
+          
+.. code-block:: python
+
+  # Example:
+  import numpy as np
+
+  # Create a simulation state.
+  state = cudaq.get_state(kernel)
+  # Return the amplitude of |0101>, assuming this is a 4-qubit state.
+  amplitude = state.amplitude([0,1,0,1])#")
```
bettinaheim, line 436: Why not go with a string here that is converted to the appropriate integer index/indices during submission? I find this a bit awkward to have to put commas for a large number of qubits (bitvector would also be a natural choice).

bettinaheim (approved):
Approving to unblock the code ownership, but someone else should take a look at the state changes as well.

runtime/common/BaseRestRemoteClient.h
```diff
@@ -217,8 +217,40 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
+      cudaq::IRPayLoad stateIrPayload1, stateIrPayload2;
+
+      stateIrPayload1.entryPoint = kernelName1;
+      stateIrPayload1.ir = constructKernelPayload(mlirContext, kernelName1,
+                                                  nullptr, args1, argsSize1);
+      stateIrPayload2.entryPoint = kernelName2;
+      stateIrPayload2.ir = constructKernelPayload(mlirContext, kernelName2,
+                                                  nullptr, args2, argsSize2);
+
+      {
```
bmhowe23, line 241: What is the point of the scope braces on lines 241, 245, 246, and 249?

runtime/common/ExecutionContext.h
```diff
@@ -61,6 +61,18 @@ class ExecutionContext {
   /// @brief Pointer to simulation-specific simulation data.
   std::unique_ptr<SimulationState> simulationState;
 
+  /// @brief A map of basis-state amplitudes
+  // The list of basis state is set before kernel launch and the map is filled
+  // by the executor platform.
+  std::map<std::vector<int>, std::complex<double>> amplitudeMaps;
```
bmhowe23, line 67: Just curious ... why is this one non-optional but the other new ones in lines 70 and 74 are optional? If amplitudeMaps is optional as well, I'd recommend making that explicit.

runtime/common/JIT.cpp
```diff
@@ -67,9 +67,10 @@ void invokeWrappedKernel(std::string_view irString,
     // Hence, fix the linkage.
     const auto fixUpLinkage = [](auto &func) {
       if (func.hasInternalLinkage()) {
-        llvm::dbgs() << "Change linkage type for symbol " << func.getName()
-                     << " internal to "
-                        "external linkage.";
+        LLVM_DEBUG(llvm::dbgs()
+                   << "Change linkage type for symbol " << func.getName()
+                   << " internal to "
+                      "external linkage.");
```
bmhowe23, line 73: nit:
```suggestion
                   << " internal to external linkage.");
```

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -567,18 +567,67 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
...
+          throw std::runtime_error("Failed to decode input IR");
+        }
+        std::string_view codeStr(decodedCodeIr.data(), decodedCodeIr.size());
+        handleRequest(reqId, request.executionContext, request.simulator,
+                      codeStr, request.entryPoint, request.args.data(),
+                      request.args.size(), request.seed);
+
+        // If specific amplitudes are requested.
+        // Note: this could be the case whereby the state vector is too large
+        // for full retrieval.
```
bmhowe23, line 618: It might be helpful to add a reference to `RemoteSimulationState::maxQubitCountForFullStateTransfer` here for anybody searching the code. (It took me a little while to search where "too large" was explicitly defined.)

runtime/cudaq/qis/remote_state.cpp
```diff
@@ -0,0 +1,144 @@
...
+  platform.set_exec_ctx(&context);
+  platform.launchKernel(kernelName, nullptr, nullptr, 0, 0);
+  platform.reset_exec_ctx();
+  assert(context.overlapResult.has_value());
+  return context.overlapResult.value();
+}
+
+std::size_t RemoteSimulationState::maxQubitCountForFullStateTransfer() {
+  // Default number of qubits for full state vector transfer.
+  constexpr std::size_t NUM_QUBITS_STATE_TRANSFER = 25; // (~100MB of data)
```
bmhowe23, line 133: Since this is server-side code (I think), what happens if the user tries to submit a request exceeding this limit without providing any specific amplitude basis vectors? Do they get an error message or warning message somewhere?

runtime/cudaq/qis/remote_state.h
```diff
@@ -0,0 +1,103 @@
...
+// The state is represented by a quantum kernel.
+// For accessor APIs, we may resolve the state to a state vector by executing
+// the kernel on the remote simulator. For overlap API b/w 2 remote states, we
+// can send both kernels to the remote backend for execution and compute the
+// overlap.
+class RemoteSimulationState : public cudaq::SimulationState {
+protected:
+  std::string kernelName;
+  mutable std::unique_ptr<cudaq::SimulationState> state;
+  mutable std::vector<char> argsBuffer;
```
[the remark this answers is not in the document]
1tnguyen: I added a code comment in https://github.com/NVIDIA/cuda-quantum/pull/1803/commits/8ab176faf836e7ff11a77bd548c1c91f836fecb3 to document this.

This mainly because of the lazy evaluation. The base API from `cudaq::SimulationState` is expected to be `const` but we would perform evaluation at those points. 

The `argsBuffer` data needs to be passed as non-const pointer for those `launchKernel` evaluation calls as we;ll.

runtime/cudaq/qis/remote_state.cpp
1tnguyen: It's actually client-side code. The limit here is mainly an optimization for amplitude access (resolving the whole state vector rather than sending amplitude requests). We'll have a doc update to note about the benefit of using batched amplitudes API rather than sequential amplitude calls for large number of qubits using NVQC.

If users want to resolve the whole state vector (not an amplitude access), we treat them as the current v0.7 mode, sending back the whole state vector (memory allocation failures if the state is too big).

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
1tnguyen: Good point. Added in 0549fd7fe1f8f2a653223143028bacdf30839596

runtime/common/BaseRestRemoteClient.h
```diff
@@ -217,8 +217,40 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
+        // First kernel of the overlap calculation
+        request.code = stateIrPayload1.ir;
+        request.entryPoint = stateIrPayload1.entryPoint;
+      }
+      {
+        // Second kernel of the overlap calculation
+        request.overlapKernel = stateIrPayload2;
```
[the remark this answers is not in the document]
1tnguyen: I'll defer this change since it'd require further changes to the server-side version check and client-side filter like you said. (more compatibility validation is needed).

PR #1808, opened 2024-06-13
[core] Teach synthesis how to unpack a vector of complex values.

schweitzpgi:
This PR also addresses some issues in QIR codegen that were found doing ad hoc testing. It could be the case that trying to use QIR with state handling is not supposed to work at all, however. To wit, there are checks made the verify that the code conforms with the QIR specification and these *will* be violated when using qubit allocations with state. For example, the state is not a literal constant but rather a buffer of values. The buffer will be in memory, which is not a constant. The LLVM optimizer may chose LLVM instructions and intrinsics that QIR doesn't consider "legal" as well.


amccaskey (approved):
+1

PR #1810, opened 2024-06-13
[Do not merge] Testing CI on notebook

khalatepradnya:
* Attempt to fix the CI failures seen on PR#1763


scripts/validate_container.sh
```diff
@@ -251,6 +251,9 @@ done
 
 if [ -n "$(find $(pwd) -name '*.ipynb')" ]; then
     echo "Validating notebooks:"
+    echo "Installing packages required for notebook validation"
+    ## Ref: https://github.com/qutip/qutip/issues/2412
+    python3 -m pip install --no-cache-dir "qutip<5" matplotlib
```
bettinaheim, line 256: Instead of installing this here, it makes sense to install it by default in the docker image, specifically along with the notebook package: https://github.com/NVIDIA/cuda-quantum/blob/9dbfec4ec2cc6afb0f0c5657ccffc3f578417e46/docker/release/cudaq.Dockerfile#L65
Then also the arm validation should pass (qutip is failing to install on arm here, since it doesn't have gcc to build it)

bettinaheim: I took the liberty of updating your branch to confirm that this works.

khalatepradnya:
Closing this as the required changes were pushed directly to https://github.com/NVIDIA/cuda-quantum/pull/1763

PR #1817, opened 2024-06-13
[core] Add handling of cudaq::state* arguments to marshalling and

schweitzpgi:
materialization.

With a cudaq::state*, quake intentionally sees a completely opaque data type. Quake has no dependence on the implementation of this type. It is only responsible for accepting these pointers from the CUDA-Q runtime and passing them back to the CUDA-Q runtime.

Since cudaq::state* is a valid type of kernel argument, this patch adds a fake marshalling of these arguments. The value of the object itself, unlike all other marshalling, is not copied for potential transport via a message. Instead, a copy of the pointer is made and it is assumed that the pointer will be rematerialized in the same address space. Effectively, we're backdooring the message passing paradigm and simply passing by reference in this case.

### DIFF
--- a/lib/Optimizer/Transforms/GenKernelExecution.cpp
+++ b/lib/Optimizer/Transforms/GenKernelExecution.cpp
@@ -434,8 +434,16 @@ public:
         hasTrailingData = true;
         continue;
       }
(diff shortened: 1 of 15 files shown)


annagrin (approved):
LGTM!

PR #1818, opened 2024-06-14
Fix a bug with allocations then noop kernel in library mode

1tnguyen:
### Description

The simulator state needs to be consistent with allocations even though the circuit is empty.
In library mode (`DefaultExecutionManager`), we flush those pending allocations (due to no gates applied) before execution context ends.


bmhowe23 (approved):
👍

PR #1824, opened 2024-06-16
Fully Remote Execution for Python VQE and Optimizers

sacpis:
This PR adds the ability to run `cudaq.vqe` and `optimizer.optimize` fully remotely on the remote-mqpu server (and eventually NVQC) rather than going back and forth between the client and the server. This greatly reduces the wall-clock execution time of these routines when using `nvqc` and `remote-mqpu`.

## Overview of the Changes
* The server that handles the requests is not `cudaq-qpud`, as that executable doesn't have a Python environment; rather there is a Python server servicing the requests (nvqc_proxy.py calls json_request_runner.py)
* The `requests` package was added as a dependency for our wheel to service the HTTP requests.
* A `std::optional<cudaq::optimization_result> optResult` was added to the `ExecutionContext`
* A `std::optional<SerializedCodeExecutionContext> serializedCodeExecutionContext` was added to the `RestRequest`
* The feature is currently gated behind an environment variable for the `nvqc` target because the current NVQC deployment does not support this feature yet.
* ~~Nested objective functions and arg mappers are not supported yet. I.e. the only user-provided source code sent to the server is the objective function or arg mapper itself. It does not find nested functions yet~~. _Edit: added on June 20th._

## Current Status
* Server-side validation scripts will be put into a separate PR.
* Docs will go into a separate PR.
* C++ changes will go into a separate PR.


tools/cudaq-qpud/nvqc_proxy.py
bmhowe23, line 1: Unfortunately, this file had to be moved from its original location of `scripts/nvqc_proxy.py`, and GitHub isn't showing the diffs for the file content. See https://github.com/NVIDIA/cuda-quantum/compare/main...bmhowe23:cuda-quantum:TEMP-REVIEW for an easy way to see the actual code diffs.

ikkoham (commented):
Thank you for the great work! I added some minor comments.

python/runtime/cudaq/algorithms/py_utils.cpp
```diff
@@ -0,0 +1,166 @@
...
+      auto value = item.second;
+
+      py::bytes serialized_value = pickle.attr("dumps")(value);
+      serialized_dict[key] = serialized_value;
+    } catch (const py::error_already_set &e) {
+      // Uncomment the following lines for debug, but all this really means is
+      // that we won't send this to the remote server.
+
+      // std::cout << "Failed to pickle key: " + std::string(e.what())
+      //           << std::endl;
```
1tnguyen, line 33: nit: do we want to include `key` in the debug message?

tools/cudaq-qpud/json_request_runner.py
```diff
@@ -0,0 +1,132 @@
...
+        try:
+            cmd_result = subprocess.run(['cudaq-qpud', '--cuda-properties'],
+                                        capture_output=True,
+                                        text=True)
+            deviceProps = json.loads(cmd_result.stdout)
+        except:
+            deviceProps = dict()
+
+        # We don't have visibility into the difference between `requestStart`
+        # and `simulationStart`, so simply use `simulationStart` for both
```
1tnguyen, line 108: I suppose we can use the timestamp at the start of `if __name__ == "__main__":` as `requestStart`. 
It'd help to measure the preprocessing time.

1tnguyen:
This looks good to me 👍 (validation in `json_request_runner.py` is to be added in a separate PR as I understood)

tools/cudaq-qpud/json_request_runner.py
```diff
@@ -0,0 +1,132 @@
...
+import subprocess
+from datetime import datetime
+
+
+def get_deserialized_dict(scoped_dict):
+    deserialized_dict = {}
+
+    for key, serialized_bytes in scoped_dict.items():
+        try:
+            deserialized_value = pickle.loads(serialized_bytes)
```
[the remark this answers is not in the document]
bmhowe23: We will address the code validation in a separate PR, but I replaced the pickle usage with custom json serialization/deserialization code for limited types here: 87f0c0255fa27ad0694e6663105a5d124391b705

python/cudaq/kernel/analysis.py
```diff
@@ -315,3 +315,106 @@ def visit_FunctionDef(self, node):
...
+        for f in vis.func_names:
+            if f not in visit_set:
+                childFuncObj = FetchDepFuncsSourceCode._getFuncObj(
+                    f, calling_frame)
+                if childFuncObj:
+                    FetchDepFuncsSourceCode._getChildFuncNames(
+                        childFuncObj, calling_frame, f, full_list, visit_set,
+                        nest_level + 1)
+        full_list.append(name)
+        return full_list
```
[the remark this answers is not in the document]
bmhowe23: Yes. I double checked by adding that scenario to one of the tests: b732ab3cf.

tools/cudaq-qpud/json_request_runner.py
bmhowe23: Updated in d892aec59.

1tnguyen (approved):
LGTM 👍

bmhowe23 (approved):

Issue #1811, opened 2024-06-13
T1 Kraus channel failing CPTP test


jin-sungk:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

I am trying to define a T1 decay channel using cudaq.KrausChannel and the cuda-q and density-matrix-cpu simulator. After I define the Kraus operators and initialize the KrausChannel, I get the following runtime error: Provided kraus_ops are not completely positive and trace preserving.

![image](https://github.com/NVIDIA/cuda-quantum/assets/127429095/7e9f12b4-ed04-4e26-a3b7-7891dd171d76)

### Steps to reproduce the bug

```
import cudaq
import numpy as np

cudaq.set_target('density-matrix-cpu')

def error_prob(t):
    T1=1000 #ns
    T1_decay=1-np.exp(-t/T1)
    return T1_decay

t=10

#T1 event
kraus_0=np.array([[1.0,0.0],[0.0,np.sqrt(1-error_prob(t))]],dtype=np.complex128)
kraus_1=np.array([[0.0,np.sqrt(error_prob(t))],[0.0,0.0]],dtype=np.complex128)

t1_channel=cudaq.KrausChannel([kraus_0,kraus_1])
```

### Expected behavior

I expect the given kraus ops to pass the CPTP test. I've manually checked sum K_dag K gives the identity out to 10 decimal places.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.1
- **Python version**: 3.10.12
- **C++ compiler**: 
- **Operating system**: macos Sonoma 14.5

PR #1825, opened 2024-06-17
Fix storage order of noise kraus ops

1tnguyen:
### Description

The row-major data storage was not consistently handled in our `NoiseModel`. 

This is especially problematic for amplitude damping channel, whose Kraus matrices are not self-adjoint.

This PR:

- Makes sure that we properly map flattened row-major data to Eigen matrices for validation as well as application.
Note: the base Qpp gate application has done this row-major mapping for gate matrices correctly. We need to fix the apply noise channel path.

- Fixes a unit tests: these unit tests are using column-major data for amplitude damping channels, probably to make it to work with the underlying inconsistency.

- Adds 'u3' as a supported gate for noise modeling. We also need to let the QIR runtime (`nvqir`) propagate the gate to the simulator for the noise channel look-up to work.

- Makes the `py::buffer` (numpy ndarray) to flattened row-major Kraus matrix mapping generic for both row and column major storage in numpy.

- Adds unit test.

Fixed https://github.com/NVIDIA/cuda-quantum/issues/1811


python/runtime/common/py_NoiseModel.cpp
```diff
@@ -5,20 +5,19 @@
...
-
 #include <iostream>
+#include <pybind11/complex.h>
+#include <pybind11/stl.h>
 
 namespace cudaq {
 
 /// @brief Extract the array data from a buffer_info into our
 /// own allocated data pointer.
+// This supports 2-d array in either row or column major.
```
schweitzpgi, line 20:
```suggestion
/// This supports 2-d array in either row or column major.
```
Not sure, but...

runtime/common/NoiseModel.h
```diff
@@ -179,8 +179,8 @@ class noise_model {
   using NoiseModelOpMap =
       std::unordered_map<KeyT, std::vector<kraus_channel>, KeyTHash>;
 
-  static const constexpr std::array<const char *, 10> availableOps{
-      "x", "y", "z", "h", "s", "t", "rx", "ry", "rz", "r1"};
+  static const constexpr std::array<const char *, 11> availableOps{
```
schweitzpgi, line 182: Why not use a regular C-style array?

schweitzpgi (approved):
Seems plausible to me. Might want someone else to check though.

PR #1828, opened 2024-06-17
[ci] Whitelist microsoft and visualstudio.

schweitzpgi:
These servers go up and down enough to cause the CI to fail frequently enough to cause a lot of lost productivity. Just whitelist the domain to keep the CI from failing.


khalatepradnya:
Yes please!

bmhowe23 (approved):

PR #1834, opened 2024-06-17
State handling: reify state vectors as constants in .rodata.

schweitzpgi:
Instead of building state vectors on the stack, let them be constants that can be materialized in the .rodata section.


lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -103,6 +105,11 @@ class GenerateDeviceCodeLoader
           LLVM_DEBUG(llvm::dbgs() << "adding declaration: " << op);
           declarations.push_back(&op);
         }
+      } else if (auto global = dyn_cast<cudaq::cc::GlobalOp>(op)) {
+        if (global) {
```
1tnguyen, line 109: nit: do we need this check?

1tnguyen (approved):
LGTM 👍

amccaskey (commented):
Can you provide a test showing this? Before this was creating a cc const_array and now this will elevate the known data to a global on the module?

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
schweitzpgi: No. Thanks for catching this.

schweitzpgi:
> Can you provide a test showing this? Before this was creating a cc const_array and now this will elevate the known data to a global on the module?

I've verified that the unittests for synthesis exercise this code.

PR #1836, opened 2024-06-18
Support NVQC info logging stream customization for remote state

1tnguyen:
### Description

For the remote state, due to lazy evaluation mechanism, the remote execution may coincide with user's logging of state accessor API calls (amplitude, overlap, etc.)

We need to defer this logging (enabled by default) till finishing the remote state lifetime to not interfere with user-level logging.

**Notes**: 

- This info log is only logged **once** for the entire program. Hence, deferring this log if using remote state should be reasonable.

- We only target the default `info` log (enabled by default). For trace level logging, the console logging may be interrupted since we want real-time logging for tracing (probably expected for users when they turn on trace).


runtime/cudaq/qis/remote_state.h
```diff
@@ -44,7 +47,7 @@ class RemoteSimulationState : public cudaq::SimulationState {
     argsBuffer = cudaq::serializeArgs(std::forward<Args>(args)...);
   }
   RemoteSimulationState() = default;
-
+  ~RemoteSimulationState();
```
schweitzpgi, line 50:
```suggestion
  virtual ~RemoteSimulationState();
```

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -146,6 +146,12 @@ void quantum_platform::onRandomSeedSet(std::size_t seed) {
   for (auto &qpu : platformQPUs)
     qpu->onRandomSeedSet(seed);
 }
+
+void quantum_platform::resetLogStream() { platformLogStream = nullptr; }
+std::ostream *quantum_platform::getLogStream() { return platformLogStream; }
+void quantum_platform::setLogStream(std::ostream &logStream) {
```
schweitzpgi, line 152:
```suggestion
void quantum_platform::resetLogStream() { platformLogStream = nullptr; }

std::ostream *quantum_platform::getLogStream() { return platformLogStream; }

void quantum_platform::setLogStream(std::ostream &logStream) {
```
nit: Make a bit more readable with whitespace.

schweitzpgi (approved):
LGTM, Thien

PR #1844, opened 2024-06-21
Adding AFQMC tutorial

zohimchandani:


.github/workflows/config/spelling_allowlist.txt
bmhowe23, line 1: This file is required to be sorted using the `sort` command, which means that all capitalized words need to be at the beginning of the file. Once that is fixed, then it will also be easier to review which words added to this file.

bmhowe23: Also, just for future reference, if any words were added that aren't truly words but instead are code references (like `veq`), then they should probably be surrounded with backticks to tell the spellchecker that they are code artifacts and should not be spellchecked.

bmhowe23: Regarding the `elec`, it's complaining about this string:
```python
    """
    :param final_state_vector: State vector from a VQE simulation
    :param n_active_elec: Number of total electrons in active space
    :param spin: spin
    :param thres: Threshold for coefficients to keep from VQE wavefunction
    :returns: Input for ipie trial: coefficients, list of occupied alpha, list of occupied bets
    """
```
It can be fixed by adding these backticks:
```python
    """
    :param `final_state_vector`: State vector from a VQE simulation
    :param `n_active_elec`: Number of total electrons in active space
    :param `spin`: spin
    :param `thres`: Threshold for coefficients to keep from VQE wavefunction
    :returns: Input for ipie trial: coefficients, list of occupied alpha, list of occupied bets
    """
```
Hopefully that can be used to limit the added words to be added to actual words.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -132,7 +135,10 @@ canonicalization
 canonicalize
 canonicalizer
 canonicalizes
+chol
+ci
```
bmhowe23, line 139: I don't think these are words and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -132,7 +135,10 @@ canonicalization
...
 codebase
+coeffs
```
bmhowe23, line 141: I don't think these are words and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -164,6 +171,7 @@ discoverable
 discretization
 discretize
 discretized
+ecore
```
bmhowe23, line 174: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -189,6 +197,7 @@ extremal
 fermionic
 frontend
 frontends
+hamiltonian
```
bmhowe23, line 200: We already have Hamiltonian in this list, so I don't know why we would add the uncapitalized verison here. If it is a variable name, then it should not trigger the spell check. If this is in the example's text, then it needs to be capitalized. If it is in the text but referencing a code variable, then putting the variable name in backticks should be sufficient to make the spell checker happy.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -214,19 +224,27 @@ lossy
 lvalue
 makefiles
 merchantability
+mgpu
```
bmhowe23, line 227: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -214,19 +224,27 @@ lossy
...
 namespace
 namespaces
 natively
 normalization
+np
 nullary
+nvidia
+occa
+occb
+openfermion
```
bmhowe23, line 239: These all look like code items and probably shouldn't be added to the spell check allow list.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -214,19 +224,27 @@ lossy
...
 optimizer
 optimizers
 parallelization
 parallelize
 parallelizing
+param
 parameterization
+params
```
bmhowe23, line 247: I don't think `param` and `params` are words and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -236,19 +254,21 @@ preprocessor
 probability
 programmatically
 pybind
+pyscf
```
bmhowe23, line 257: This looks like a code item that should not be added to the allow list.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -236,19 +254,21 @@ preprocessor
...
 quantize
 quantized
 qubit
 qubits
 qudit
 qudits
-reStructuredText
 reconfigurable
+reStructuredText
```
bmhowe23, line 265: This reordering of the items in this file will cause the CI to fail. It is expecting the items to be the order that the `sort` command would output.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -236,19 +254,21 @@ preprocessor
...
 runtime
 runtimes
 rvalue
 scalability
 scalable
+scf
```
bmhowe23, line 271: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -279,6 +299,7 @@ superpositions
 symplectic
 tablegen
 templated
+thres
```
bmhowe23, line 302: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -291,8 +312,10 @@ unitarity
 unitary
 unoptimized
 upvote
+utils
```
bmhowe23, line 315: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -291,8 +312,10 @@ unitarity
...
 variadic
 variational
+veq
```
bmhowe23, line 318: I don't think this is a word and therefore should probably not be added to the allowlist.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -1,35 +1,36 @@
+AFQMC
 API
-APIs
 AST
 Aer
 AlmaLinux
+APIs
```
bmhowe23, line 6: This reordering of the items in this file will cause the CI to fail. It is expecting the items to be the order that the `sort` command would output.

bmhowe23 (commented):
Some continued comments about the spelling allow list....

docs/sphinx/examples/python/tutorials/readout_error_mitigation.ipynb
```diff
@@ -29,7 +29,60 @@
    "execution_count": 1,
    "id": "04b58854-7dab-4f1b-8c59-d8b9506bce44",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
bmhowe23, line 32: This output should probably be cleared. Leaving it in there just muddies the docs page in my opinion.

And I'm not really sure why changes to the "readout_error_mitigation.ipynb" tutorial are being done under the "Adding AFQMC tutorial". Are they related to each other?

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -1,4 +1,5 @@
 ABI
+AFQMC
```
bmhowe23, line 2: Most of these words (with the possible exception of this one) should not be added to the allow list. Rather, they need to marked as source code artifacts by annotating them with backticks in the original file that is producing this error. The spell checker will ignore items that are annotated with backticks.

.github/workflows/python_wheels.yml
```diff
@@ -316,7 +316,7 @@ jobs:
           docker run --rm -dit --name wheel-validation-examples wheel_validation:local
           status_sum=0
 
-          for ex in `find docs/sphinx/examples/python -name '*.py' -not -path '*/providers/*' -not -path '*/divisive_clustering_src/*'`; do
+          for ex in `find docs/sphinx/examples/python -name '*.py' -not -path '*/providers/*' -not -path '*/divisive_clustering_src/*' -not -path '*/utils_ipie.py' -not -path '*/vqe_cudaq_qnp.py'`; do
```
bmhowe23, line 319: Perhaps not for this PR, but we probably need to restructure some of these "support" files to put them all in a single location. They are causing our CI grief (see #2078), so perhaps they should all go into a directory called `docs/sphinx/examples/python/tutorials/support`, possibly with per-tutorial subdirectories for each one, just so we can keep them straight.

docs/notebook_validation.py
```diff
@@ -35,6 +35,10 @@ def validate(notebook_filename, available_backends):
         match = re.search('--target ([^ ]+)', notebook_content)
         if match and (match.group(1) not in available_backends):
             return False
+    for notebook_content in lines:
+        match = re.search('(config.update_option)', notebook_content)
```
bmhowe23, line 39: Just curious ... what is `config.update_opton`? Maybe I'll keep reading the rest of the files and find out, but I figured I'd ask anyway.

docs/sphinx/examples/python/tutorials/vqe.ipynb
```diff
@@ -19,7 +19,63 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
bmhowe23, line 22: Similar to above, I think the outputs for this cell should be cleared. They add lots of text to the docs page without adding much value in my opinion.

And also similar to above, why are we changing this one under the "Adding AFQMC tutorial" PR?

docs/notebook_validation.py
bmhowe23: So does that mean we are disabling all validation of this new tutorial, even though that line is disabling GPU support? If so, why does validation need to be disabled if no GPUs are required on the validation runner?

docs/notebook_validation.py
```diff
@@ -31,10 +31,6 @@ def validate(notebook_filename, available_backends):
         match = re.search('set_target[\\\s\(]+"(.+)\\\\"[)]', notebook_content)
         if match and (match.group(1) not in available_backends):
             return False
-    for notebook_content in lines:
-        match = re.search('--target ([^ ]+)', notebook_content)
-        if match and (match.group(1) not in available_backends):
-            return False
```
bmhowe23, line 37: Was the removal of these lines intentional? If so, why?

.github/workflows/docker_images.yml
```diff
@@ -778,7 +778,7 @@ jobs:
           docker cp docs/notebook_validation.py cuda-quantum:"/home/cudaq/notebook_validation.py"
           # In containers without GPU support, UCX does not work properly since it is configured to work with GPU-support.
           # Hence, don't enforce UCX when running these tests.
-          docker exec cuda-quantum bash -c "python3 -m pip install pandas scipy pandas seaborn"
+          docker exec cuda-quantum bash -c "python3 -m pip install pandas scipy pandas seaborn h5py==3.10.0"
```
bmhowe23, line 781: We have a few other places that have h5py restrictions (like devdeps.Dockerfile and pyproject.toml). It would probably be best to specify it in the same way as them.
```suggestion
          docker exec cuda-quantum bash -c "python3 -m pip install pandas scipy pandas seaborn h5py<3.11"
```

bmhowe23 (approved):
LGTM. Just one minor nit about the h5py versioning specification.

PR #1847, opened 2024-06-24
Remove the redundant header from optimizer file 

khalatepradnya:
This allows the optimizer logic to be reused within the CUDA-Q source tree.

(No side effect observed on tests / application / user code)


bmhowe23 (approved):
👍

PR #1848, opened 2024-06-24
Draft of dependency graph generation algorithm

atgeller:
### Description

This is a draft of an analysis to determine dependency information between qubit operations in a quake program.


atgeller:

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+  };
+
+  DependencyNode(DependencyNode *from)
+      : successors({from}), dependencies(), qindex(from->qindex),
+        associated(nullptr) {
+    addSuccessor(from);
+  };
+
+  inline void addSuccessor(DependencyNode *other) {
```
schweitzpgi, line 69: nit: the `inline` is redundant here

schweitzpgi (commented):
I'll have to try this out.

atgeller:
Can be used like so: `cudaq-quake FILE | cudaq-opt --add-dealloc --factor-quantum-alloc --memtoreg --cse --canonicalize --assign-ids --dep-analysis`. Use `--debug` to see lifetime info printed. An example test case can be found in `test/Quake/dependency.qke`.

The additional `assign-ids` pass is a necessary first step to assign matching unique identifiers to each qubit alloc/dealloc.

The `dep-analysis` pass builds dependency graphs for straightline code, performs scheduling, determines lifetimes and finds reuse opportunities, and then emits a new program based on the new schedule and reuses.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.h
```diff
@@ -70,6 +70,15 @@ void genericOpPrinter(mlir::OpAsmPrinter &_odsPrinter, mlir::Operation *op,
 // Utility functions to test the form of an operation.
 //===----------------------------------------------------------------------===//
 
+// Is \p op in the Quake dialect?
+// TODO: Is this StringRef comparison faster than calling MLIRContext::
+// getLoadedDialect("quake")?
```
schweitzpgi, line 75: I believe it must be. This does a O(c) —where c is at most 5—comparison on two character sequences. The getLoadedDialect call does an O(log_2(n)) search over all loaded dialects with the same constant factors.

lib/Optimizer/Transforms/AssignIDs.cpp
```diff
@@ -0,0 +1,171 @@
...
+  else if (hasClassicalInput(defop))
+    i++;
+
+  return findQid(defop->getOperand(i));
+}
+
+class SinkOpPat : public OpRewritePattern<quake::SinkOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  using Base = OpRewritePattern<quake::SinkOp>;
```
schweitzpgi, line 110: nit: can this line be deleted?

lib/Optimizer/Transforms/AssignIDs.cpp
```diff
@@ -0,0 +1,171 @@
...
+    auto *ctx = &getContext();
+    func::FuncOp func = getOperation();
+    RewritePatternSet patterns(ctx);
+    unsigned x = 0;
+    patterns.insert<NullWirePat>(ctx, &x);
+    patterns.insert<SinkOpPat>(ctx);
+    ConversionTarget target(*ctx);
+    target.addLegalDialect<quake::QuakeDialect>();
+    target.addDynamicallyLegalOp<quake::NullWireOp>(
+        [&](quake::NullWireOp alloc) { return alloc->hasAttr("qid"); });
```
schweitzpgi, line 160: The attribute approach is brittle as other transformations will not necessarily keep the information. We should really use WireSet, BorrowWire, ReturnWire for explicitly mapping the identity of a qubit from an infinite space (virtual registers) to a finite space (physical registers).

atgeller: Makes sense, right now Assign IDs and Dependency Analysis are very tangled passes, so they pretty much have to be run one right after another, and Dependency Analysis has to check that the attributes are present.

lib/Optimizer/Transforms/AssignIDs.cpp
atgeller: Yes, done

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.h
atgeller: Gotcha, I brought the comment over with the rest of the function so I didn't know.

atgeller:
Closing in lieu of #2163

Issue #698, opened 2023-09-27
Potential symbol collision when pre-loading multiple sim libs in `LinkedLibraryHolder` (Python)


1tnguyen:
In `LinkedLibraryHolder.cpp`, we [load (`dlopen`)](https://github.com/NVIDIA/cuda-quantum/blob/219bc57c8d55328e6c2ee209859eb6f6efb4ee5d/python/utils/LinkedLibraryHolder.cpp#L170) all the nvqir simulator libs on construction. 

This could potentially create issues like what we saw [here](https://github.com/NVIDIA/cuda-quantum/pull/660#issuecomment-1736178805) (the linker switches between simulator libs at will)

Note: this is the same issue as what we've fixed for OpenBLAS: https://github.com/NVIDIA/cuda-quantum/issues/186

One potential fix is to make `LinkedLibraryHolder.cpp` to do proper load/unload of polymorphic plugin libs (e.g., backend simulators) as requested.

PR #1849, opened 2024-06-24
Make LinkedLibraryHolder.cpp ordering consistent

bmhowe23:
The simulator library loading order currently ends up being unspecified (due to directory_iterator order being unspecified by the C++ standard), and due to #698, this can have negative effects. This PR doesn't totally resolve any ordering issues, but it at least makes the consistent for different users and different platforms by sorting the library names alphabetically before processing them.


1tnguyen (approved):
LGTM 👍

PR #1853, opened 2024-06-26
[core] Fix semantic problems with cc.compute_ptr

schweitzpgi:
The CC dialect has a strong semantics on sequence types. Unlike C (and LLVM), a pointer to a scalar type is not conflated with a pointer to an array of scalars. The types are explicitly distinct.

As a consequence, the following has no semantics:

```mlir
  %1 = ... : !cc.ptr<i32>
  %2 = cc.compute_ptr %1[5] : (!cc.ptr<i32>) -> !cc.ptr<i32>
```

This construct violates the semantics of cc.compute_ptr and will now fail to pass the verifier. The correct way to do this is:

```mlir
  %1 = ... : !cc.ptr<!cc.array<i32 x ?>>
  %2 = cc.compute_ptr %1[5] : (!cc.ptr<!cc.array<i32 x ?>>) -> !cc.ptr<i32>
```

These patches add the checks and fix all the places where the semantics were being violated.

As a consequence, it will be easier to reason about cc.compute_ptr, to debug issues, and to write code that uses cc.compute_ptr.

Reimplement the quake add metadata pass so that it isn't doing ad hoc pattern matching and is prone to failure when small changes to the IR happen. (Is there an issue for this?)

Add more canonicalizations for cc.compute_ptr.

### DIFF
--- a/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
+++ b/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
--- a/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
+++ b/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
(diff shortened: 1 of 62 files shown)


schweitzpgi:
ping: can we get this merged so we can stop writing broken IR?

lib/Optimizer/CodeGen/QuakeToCodegen.cpp
```diff
@@ -0,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
1tnguyen, line 2: nit:
```suggestion
 * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
```

lib/Optimizer/CodeGen/QuakeToCodegen.h
```diff
@@ -0,0 +1,17 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
1tnguyen, line 2: nit:
```suggestion
 * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
```

1tnguyen (approved):
LGTM 👍

PR #1854, opened 2024-06-26
Add a state preparation pass before synthesis

annagrin:
### Description
<Initializing qvectors from data required conversion of data to quantum gates for quantum backends>

- Add a pass during synthesis (before QuakeSynthesizer) to convert state data to quantum gates
- Add tests


schweitzpgi:
Q: Is there some reason this needs to happen before parameter synthesis?

annagrin:
> Q: Is there some reason this needs to happen before parameter synthesis?
> 
> Corollary: we can't know what the state vector's value is until parameter synthesis during JIT compilation, so I'm curious if there is a technical issue why we need to synthesize gates before we get the vector's value...

Not really - just a choice in this case (i.e. I preferred to use the parameter data rather than read from the transformation of the parameter created by synthesis, like attributes from a global constant array). As we figure out how to support initializing vectors from inside the kernel from a vector created in the kernel (as supposed to passed as a parameter), I might move it around and update. If we end up using a const array for that case, then the IR would look very similar to the IR after synthesis and we only might need one implementation, which then could run after synthesis instead. The main part of this PR is the transformation to gates itself, and it should stay the same in both cases.

schweitzpgi:
I think this should be done by working at the IR level. The state vector will not necessarily always be some on-the-side information. To wit, the question about inline initialization lists in a kernel.

annagrin:
Replaced by https://github.com/NVIDIA/cuda-quantum/pull/1891

PR #1857, opened 2024-06-26
Update trotter example with progress printing

annagrin:
### Description
Update trotter example with progress printing.

- Make sure c++ and python examples print similar messages
- Print times and results of all steps while running


docs/sphinx/examples/cpp/other/trotter_kernel_mode.cpp
```diff
@@ -121,23 +122,31 @@ int run_steps(int steps, int spins) {
...
+    expResults.emplace_back(result);
     state = cudaq::get_state(trotter{}, &state, coefficients, words, dt);
     const auto stop = std::chrono::high_resolution_clock::now();
     auto duration =
         std::chrono::duration_cast<std::chrono::microseconds>(stop - start);
-    runtimeMs.emplace_back(duration.count() / 1000.0);
+    auto timeInSeconds = duration.count() / 1000.0 / 1000.0;
+    runtimeMs.emplace_back(timeInSeconds);
+    std::cout << "Step " << i << ": time [s]: " << timeInSeconds
+              << ", result: " << result << std::endl;
```
bmhowe23, line 134: Just a note: we typically favor `'\n'` instead of `std::endl` because `std::endl` forces a flush, and that flush has a throughput cost. However, in this example case, it's probably _better_ to force the flush to happen so the user is guaranteed to see the output as it happens. That's a long way of saying: you don't need to change anything here 😄

docs/sphinx/examples/python/trotter_kernel_mode.py
```diff
@@ -33,7 +33,7 @@
 # This is because the CPU-only backend has difficulty handling
 # 30+ qubit simulations.
 
-spins = 5  # set to around 25 qubits for `nvidia` target
+spins = 11  # set to around 25 qubits for `nvidia` target
```
bmhowe23, line 36: The C++ example says to set spins = 100 and steps = 25 for nvidia target, but they are reversed here. Which one is the better recommendation?

bmhowe23 (approved):
LGTM. Just one minor question about recommended values in the comments.

docs/sphinx/examples/python/trotter_kernel_mode.py
annagrin: looks like a typo in c++ one, thanks!

docs/sphinx/examples/cpp/other/trotter_kernel_mode.cpp
annagrin: Got it, thanks!

docs/sphinx/examples/python/trotter_kernel_mode.py
annagrin: fixed!

bmhowe23 (approved):

PR #1861, opened 2024-06-27
Fix variable capture issue for remote NVQC VQE

bmhowe23:
This fixes an issue were local and global variables were not being capturable by cudaq.kernels when submitted to NVQC.


1tnguyen (approved):
LGTM 👍

PR #1862, opened 2024-06-27
[core] Merge the get_const_element and extract_value operations.

schweitzpgi:
These are semantically the same operation. The distinction was a
manifestation of how LLVM works rather than what the operation did.

While we're here, add semantics verification tests. Also split apart the
get_const_element fold operation from a canonicalization. Make sure we
run canonicalization *after* parameter synthesis. Update affected tests.

Depends on merge of #1853

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCOps.h
+++ b/include/cudaq/Optimizer/Dialect/CC/CCOps.h
@@ -22,24 +22,31 @@
 #include "mlir/Interfaces/LoopLikeInterface.h"
 
 namespace cudaq::cc {
-constexpr int kComputePtrConstantBitWidth = 29;
-using ComputePtrConstantIndex =
-    llvm::PointerEmbeddedInt<std::int32_t, kComputePtrConstantBitWidth>;
+constexpr int kInterleavedArgumentConstantBitWidth = 29;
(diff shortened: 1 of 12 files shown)


lib/Optimizer/CodeGen/QuakeToCodegen.cpp
```diff
@@ -0,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
1tnguyen, line 2: nit:
```suggestion
 * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
```

lib/Optimizer/CodeGen/QuakeToCodegen.h
```diff
@@ -0,0 +1,17 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                  *
```
1tnguyen, line 2: nit:
```suggestion
 * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
```

1tnguyen (approved):
LGTM 👍

PR #1863, opened 2024-06-27
[NFC] Reorganize the quake.td file

schweitzpgi:
This file was getting pretty scattered with related operations thrown in all over the place. This change just sorts the deck of cards so related operations are grouped together.

PR #1864, opened 2024-06-28
[core] Add the concept of wire sets.

schweitzpgi:
A wire set is a new abstraction in Quake. A wire set is a finite set of wires (a.k.a. qubits). Each wire has a unique identity. All wires remain linear types.

The usage model is to borrow a wire from a set, use it, and return it to the set.

A wire set can also model other physical properties of qubits such as topology via a sparse adjecency map (for copious generality, including unidirectional links).

~Requires #1863 - the diffs will be clearer once the referenced PR is merged.~


schweitzpgi:
ping: needs review to unblock the interns

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -546,6 +546,89 @@ def quake_SinkOp : QuakeOp<"sink"> {
...
+    ```
+  }];
+
+  let arguments = (ins
+    FlatSymbolRefAttr:$set_name,
+    I32Attr:$identity
+  );
+  let results = (outs WireType);
+  
+  //let hasVerifier = 1;
```
bmhowe23, line 612: Why is this commented out?

schweitzpgi: Added with 38d0b76

bmhowe23 (approved):
LGTM

PR #1865, opened 2024-06-28
Use a single scratchpad for all tensornet operations

1tnguyen:
### Description

Scratchpad (to use with tensornet 'compute' API) is created a single time, attached to the simulator class.
All other sub-classes (`state`-like objects) just keep a reference to this scratch pad.

**Rationale**:
We observed some performance degradation running these compute API (e.g., state/observe, etc.) in a loop. Theoretically, the performance should be better since the system is warmed up. 
The root cause is this scratch pad dealloc-realloc operations.


schweitzpgi (approved):
+1

PR #1866, opened 2024-06-28
Add initialization_list test.

schweitzpgi:
Tests that the distinction between lowering an initialization list to be used for a state initialization vs. an initialization list that we expect to constant fold into the quantum circuit.


annagrin (approved):
LGTM!

PR #1868, opened 2024-06-28
[core] Fix bug in lower-to-cfg

schweitzpgi:
The pass needed to be updated to handle cc.scope which allows arbitrary control flow regions nested within it now.

Add a test.


schweitzpgi:
As reported by @atgeller

khalatepradnya (approved):

PR #1869, opened 2024-06-28
[core] Add -lift-array-value pass

schweitzpgi:
This pass will promote an array constructed inline on the stack to a cc.const_array op.

TODO: when the array is used by quake.init_state, the verifier will fail. Nothing in the code base is built to accept an array *value* as an input to quake.init_state's second argument.

~Builds on #1862~

This pass will be useful to both the unitary custom op and state handling projects.


annagrin (commented):
LGTM, thanks!

1tnguyen (approved):
LGTM 👍

PR #1870, opened 2024-06-29
[core] Add to the factory's IRBuilder the creating of vector<complex>

schweitzpgi:
These changes add a pair of methods to correctly build a GlobalOp that corresponds with a `vector<complex<REAL>>` constant block. Using these functions will generate a GlobalOp which has the attributes that MLIR will lower to LLVM correctly (and without crashing).

### DIFF
--- a/lib/Optimizer/Builder/Intrinsics.cpp
+++ b/lib/Optimizer/Builder/Intrinsics.cpp
@@ -9,6 +9,7 @@
 #include "cudaq/Optimizer/Builder/Intrinsics.h"
 #include "cudaq/Optimizer/Builder/Runtime.h"
 #include "cudaq/Optimizer/CodeGen/CudaqFunctionNames.h"
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/CommandLine.h"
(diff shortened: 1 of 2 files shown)


lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -374,4 +375,40 @@ LogicalResult IRBuilder::loadIntrinsic(ModuleOp module, StringRef intrinName) {
...
+                             const std::vector<std::complex<A>> &values,
+                             IRBuilder &builder, Type ty) {
+  if (auto glob = module.lookupSymbol<cc::GlobalOp>(name))
+    return glob;
+  auto *ctx = builder.getContext();
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToEnd(module.getBody());
+  auto complexTy = ComplexType::get(ty);
+  auto globalTy = cc::ArrayType::get(ctx, complexTy, values.size());
+  SmallVector<std::complex<APFloat>> newValues;
```
khalatepradnya, line 390: 👍🏽
I was missing this!

khalatepradnya (approved):
LGTM!
(I imagine tests can be added when combined with either the state handling logic or custom operations logic).

Issue #1814, opened 2024-06-13
Unclear error message on passing ndarray to kernels that accept cudaq.State


annagrin:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Python gives unclear message when I pass a `ndarray` to a kernel that takers `cudaq.State` as a parameter"

```
RuntimeError: error: Cannot infer CUDA-Q type from provided Python type (!cc.ptr<!cc.state>)

Offending code:
  File "/home/annagrin/source/cuda-quantum/program.py", line 24, in <module>
    counts = cudaq.sample(kernel, c)
```

### Steps to reproduce the bug

```python
import cudaq
import numpy as np

c = np.array([1, 2j, 3, 4j, 5, 6j, 7, 8j], dtype=np.complex64)

@cudaq.kernel
def kernel(vec: cudaq.State):
    q = cudaq.qvector(vec)

counts = cudaq.sample(kernel, c)
```

### Expected behavior

A better error message would mention the actual type (`ndarray`) not matching the formal type (`cudaq.State`)

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #1872, opened 2024-07-01
Add `State` type to `mlirTypeToPyType`

1tnguyen:
### Description

The unclear error message reported in https://github.com/NVIDIA/cuda-quantum/issues/1814 was due to this error message:

https://github.com/NVIDIA/cuda-quantum/blob/ea0f950bc31c1365708f45c6ffbdeb44d9939c67/python/cudaq/kernel/kernel_decorator.py#L387

When trying to construct the error message ("Invalid runtime argument type....."), the `mlirTypeToPyType` call throwed because we don't support the `State` type yet, leading to an unclear error message.

This PR adds support for `State` type and a unit test case for this scenario.

Resolved https://github.com/NVIDIA/cuda-quantum/issues/1814


annagrin (commented):
Thank you Thien, LGTM!

annagrin (approved):

python/tests/kernel/test_kernel_qvector_state_init.py
```diff
@@ -410,3 +410,17 @@ def kernel2():
     assert '1100' in counts
     assert '0011' in counts
     assert '0000' in counts
+
+def test_invalid_arg_error_msg():
+    cudaq.reset_target()
+
+    c = np.array([1. / np.sqrt(2.) + 0j, 0., 0., 1. / np.sqrt(2.)],
+                 dtype=complex)
```
schweitzpgi, line 418: Is there a way to wrap this in the `State` class? Or is that already a test somewhere?

1tnguyen: Yes, normally users would need to wrap it in a `State` class to call this kernel signature. There are tests for this case, e.g. 

https://github.com/NVIDIA/cuda-quantum/blob/ea0f950bc31c1365708f45c6ffbdeb44d9939c67/python/tests/kernel/test_kernel_qvector_state_init.py#L75-L81

PR #1877, opened 2024-07-01
[C++ bridge] Add lowering for vector<complex<REAL>> and

schweitzpgi:
for vector<struct{...}> when the struct size is recorded from the AST.

### DIFF
--- a/lib/Frontend/nvqpp/ConvertStmt.cpp
+++ b/lib/Frontend/nvqpp/ConvertStmt.cpp
@@ -341,7 +341,7 @@ bool QuakeBridgeVisitor::VisitReturnStmt(clang::ReturnStmt *x) {
           TODO_x(toLocation(x), x, mangler, "unhandled complex element type");
           return false;
         }
-        std::size_t byteWidth = (eleTy.getIntOrFloatBitWidth() + 3) / 4;
+        std::size_t byteWidth = ((eleTy.getIntOrFloatBitWidth() + 7) / 8) * 2;
         result = createVectorInit(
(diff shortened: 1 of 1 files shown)


lib/Frontend/nvqpp/ConvertStmt.cpp
```diff
@@ -317,19 +317,47 @@ bool QuakeBridgeVisitor::VisitReturnStmt(clang::ReturnStmt *stmt) {
...
+        std::size_t byteWidth = (eleTy.getIntOrFloatBitWidth() + 7) / 8;
+        result = createVectorInit(
+            builder.create<arith::ConstantIntOp>(loc, byteWidth, 64));
+      } else if (auto complexTy = dyn_cast<ComplexType>(eleTy)) {
+        eleTy = complexTy.getElementType();
+        if (!isa<IntegerType, FloatType>(eleTy)) {
+          TODO_x(toLocation(x), x, mangler, "unhandled complex element type");
+          return false;
+        }
+        std::size_t byteWidth = (eleTy.getIntOrFloatBitWidth() + 3) / 4;
```
khalatepradnya, line 344: Q: Can you help me understand ``` ... + 3 ) / 4```? 
(Don't we need twice the size to account for real and imaginary parts)

schweitzpgi: ((x + 7) / 8) * 2 = (x + 3) / 4 in 2's complement.

khalatepradnya (approved):
LGTM 👍🏽

PR #1879, opened 2024-07-01
[core] Add offsetof operation to CC dialect.

schweitzpgi:
Add the op with a verifier and canonicalizer. Add to roundtrip test. Will add codegen for this next.

Depends of #1862

### DIFF
--- a/lib/Optimizer/Dialect/CC/CCOps.cpp
+++ b/lib/Optimizer/Dialect/CC/CCOps.cpp
@@ -345,7 +345,8 @@ LogicalResult cudaq::cc::ComputePtrOp::verify() {
   auto resultTy = cast<cc::PointerType>(getResult().getType());
   for (std::int32_t i : getRawConstantIndices()) {
     if (auto arrTy = dyn_cast<cc::ArrayType>(eleTy)) {
-      if (i != kDynamicIndex && !arrTy.isUnknownSize() && i > arrTy.getSize()) {
+      if (i != kDynamicIndex && !arrTy.isUnknownSize() &&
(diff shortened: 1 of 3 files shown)


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -1788,6 +1797,72 @@ MutableOperandRange cudaq::cc::ConditionOp::getMutableSuccessorOperands(
...
+  LogicalResult matchAndRewrite(cudaq::cc::OffsetOfOp offOp,
+                                PatternRewriter &rewriter) const override {
+    // If there are no offsets, the offset is 0.
+    if (offOp.getConstantIndices().empty()) {
+      rewriter.replaceOpWithNewOp<arith::ConstantIntOp>(offOp, 0,
+                                                        offOp.getType());
+      return success();
+    }
+
+    // If all indices are 0, then the offset is necessarily 0.
```
bmhowe23, line 1847: What if you have a struct of `{int16_t, int32_t}`? The compiler will often insert 16 bits of padding, and depending on endianness, it seems like the padding could be prior to the 1st `int16_t`, and in that case, the offset to the first element could be `2`, no?

schweitzpgi: The endianness of the machine will affect the byte order vs. significance, but not the struct layout. See https://stackoverflow.com/questions/38992147/does-the-endianness-affect-how-structure-members-are-stored-into-the-memory for more discussion.

bmhowe23 (approved):

PR #1881, opened 2024-07-02
Add more performance traces to `tensornet` backends

1tnguyen:
### Description

Adds fine-grained trace logs for `cutensornet` API calls to assist future performance analysis/debugging.

### DIFF
--- a/runtime/nvqir/cutensornet/mps_simulation_state.cpp
+++ b/runtime/nvqir/cutensornet/mps_simulation_state.cpp
@@ -35,7 +35,7 @@ MPSSimulationState::~MPSSimulationState() { deallocate(); }
 std::complex<double> MPSSimulationState::computeOverlap(
     const std::vector<MPSTensor> &m_mpsTensors,
     const std::vector<MPSTensor> &mpsOtherTensors) {
-
+  LOG_API_TIME();
   auto dataDevice = deviceFromPointer(m_mpsTensors[0].deviceData);
   auto otherDevice = deviceFromPointer(mpsOtherTensors[0].deviceData);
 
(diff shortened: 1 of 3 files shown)


bmhowe23 (approved):
LGTM.

You don't have to make this change because I don't know if it is actually a good idea or not, but if you are timing nearly every function that is wrapped with `HANDLE_CUTN_ERROR`, then I can't help but wonder if the `ScopedTraceWithContext` should be wrapped into that macro.

I.e.
FROM:
`HANDLE_CUTN_ERROR(cutensornetCreateContractionOptimizerInfo(args))`
TO:
`HANDLE_CUTN_ERROR(cutensornetCreateContractionOptimizerInfo, args))`

1tnguyen:
> You don't have to make this change because I don't know if it is actually a good idea or not, but if you are timing nearly every function that is wrapped with `HANDLE_CUTN_ERROR`, then I can't help but wonder if the `ScopedTraceWithContext` should be wrapped into that macro.
> 
> I.e. FROM: `HANDLE_CUTN_ERROR(cutensornetCreateContractionOptimizerInfo(args))` TO: `HANDLE_CUTN_ERROR(cutensornetCreateContractionOptimizerInfo, args))`

I've had the same thought. After some consideration, I think it might be a bit 'too much' to time-trace every call (e.g., any runtime overhead and the number of log lines to read/analyze when needed). 

I'd say we keep this option in mind and merge this PR in. If we saw the need for more time trace logging of cutensornet API usage beyond these create-config-prepare-compute APIs, we'd go to the full logging route via the `HANDLE_CUTN_ERROR` macro.

PR #1882, opened 2024-07-02
Fix grover example

bettinaheim:
The fact that grover was not actually producing the value we are looking for bothered me. Here is a slightly more general version that reliably produces the correct output.


schweitzpgi (approved):
Thanks for splitting this one.

PR #1883, opened 2024-07-02
Follow-up to PR#1877 - handle complex number vector in the kernel execution generation pass

khalatepradnya:


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1353,11 +1353,22 @@ class GenerateKernelExecution
...
-          genStdvecTFromInitList(loc, builder,
-                                 rewriteEntryBlock->getArguments().front(),
-                                 dataPtr, tSize, vecLen);
+          if (isa<IntegerType, FloatType>(eleTy)) {
+            auto size = (eleTy.getIntOrFloatBitWidth() + 7) / 8;
+            Value tSize = builder.create<arith::ConstantIntOp>(loc, size, 64);
+            genStdvecTFromInitList(loc, builder,
+                                   rewriteEntryBlock->getArguments().front(),
+                                   dataPtr, tSize, vecLen);
+          } else if (auto complexTy = dyn_cast<ComplexType>(eleTy)) {
```
schweitzpgi, line 1362: Should handle `StructType` as well.

Also, let's do a little refactoring instead of block copy.

It would be good to think about emitting a warning for unhandled cases instead of silently ignoring the return values.

khalatepradnya: Addressed in https://github.com/NVIDIA/cuda-quantum/pull/1883/commits/e4abdbe77771ec656b27f869d16f7b387c919fdf

schweitzpgi: I'm ok with converting to an issue with these.

schweitzpgi (approved):

lib/Optimizer/Transforms/GenKernelExecution.cpp
khalatepradnya: https://github.com/NVIDIA/cuda-quantum/issues/1886

PR #1888, opened 2024-07-03
Fix #1886: refactoring of common code.

### LINKED ISSUE #1886: Code duplication in `ConvertStmt.cpp` and `GenKernelExecution.cpp`
Also, let's do a little refactoring instead of block copy.

_Originally posted by @schweitzpgi in https://github.com/NVIDIA/cuda-quantum/pull/1883#discussion_r1662893802_

schweitzpgi:
This refactors the common code that will generate expressions to determine the size of an object with a type that can be reified in memory statically. The code is moved to the IRBuilder and out from ConvertStmt and GenKernelExecution. (There may be other places that ought to use this new IRBuilder code.)


khalatepradnya (approved):
Thanks, Eric! 👍🏽

PR #1889, opened 2024-07-05
[llvm] Fix region simplification bug

### LINKED ISSUE #1799: Incorrect circuit compilation when using some nested loops
### Required prerequisites

### Describe the bug

When using nested loops where the inner loop doesn't start at 0, the circuit compilation can produce incorrect circuits.

Note that the ultimate root cause of this issue is https://github.com/llvm/llvm-project/issues/94520, but there are more details below.

Also note that the current manifestation of the bug was accidentally fixed in #1692, so **the latest from main will not show this bad behavior**, but see the "suggestions" for more details.

The issue can best be seen with the following test case (thanks to @poojarao8 for the clean example):

Using `nvcr.io/nvidia/quantum/cuda-quantum:0.7.1`...

```python
import cudaq

N = 4 # Number of qubits

@cudaq.kernel
def ansatz():
    qr = cudaq.qvector(N)
    
    for j in range(N-1):
        for k in range(j+1,N):
            x.ctrl(qr[j], qr[k])
    
print("Circuit from the decorator (INCORRECT)")
print(cudaq.draw(ansatz))

ansatz = cudaq.make_kernel()
qr = ansatz.qalloc(N)

for j in range(N-1):
    for k in range(j+1,N):
        ansatz.cx(qr[j], qr[k])
      
print("Circuit from the builder (CORRECT)")
print(cudaq.draw(ansatz))
```
```bash
cudaq@b4d004c1f8e8:~$ python3 test.py
Circuit from the decorator (INCORRECT)

q0 : ──●────────────
     ╭─┴─╮
q1 : ┤ x ├──●───────
     ╰───╯╭─┴─╮
q2 : ─────┤ x ├──●──
          ╰───╯╭─┴─╮
q3 : ──────────┤ x ├
               ╰───╯

Circuit from the builder (CORRECT)

q0 : ──●────●────●─────────────────
     ╭─┴─╮  │    │
q1 : ┤ x ├──┼────┼────●────●───────
     ╰───╯╭─┴─╮  │  ╭─┴─╮  │
q2 : ─────┤ x ├──┼──┤ x ├──┼────●──
          ╰───╯╭─┴─╮╰───╯╭─┴─╮╭─┴─╮
q3 : ──────────┤ x ├─────┤ x ├┤ x ├
               ╰───╯     ╰───╯╰───╯
```

### Expected behavior

The expected behavior is shown above.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.1
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

This was actually fixed with #1692 (**which will be part of 0.8.0**), despite #1692 having no intention of fixing this bug. That PR simply reordered some of the QIR passes, and subsequent research determined that the reorder was masking an underlying LLVM/MLIR issue. The underlying root cause was reported here: https://github.com/llvm/llvm-project/issues/94520.

Since the issue still exists in LLVM upstream as of 19.0, I think we need to perform some or all of the mitigation steps in CUDA-Q. The _exact_ list that we choose to implement is up for debate.

- [ ] Add canonicalizer pass after every lower-to-cfg pass. (The upstream LLM problem doesn't seem to occur with canonicalized code.)
- [ ] Add canonicalizer pass before every cc-loop-normalize pass. (The upstream LLM problem doesn't seem to occur with canonicalized code.)
- [ ] Disable region simplification in calls to `applyPatternsAndFoldGreedily` by doing `applyPatternsAndFoldGreedily(..., {.enableRegionSimplification = false})`.
- [ ] Fix the underlying LLVM bug and/or patch upstream LLVM. (This eliminates the need for 1-3.)
- [ ] Create UCCSD tests that check for specific answers using some sort of golden truth data.

bmhowe23:
Fixes #1799 

This fixes the underlying MLIR bug by applying a patch to our version of LLVM. The problem is that it was merging regions despite their block arguments being used differently downstream. It should not do that, so this patch fixes that.

We will work on getting it merged upstream in the coming weeks.

### DIFF
--- a/docker/build/devdeps.Dockerfile
+++ b/docker/build/devdeps.Dockerfile
@@ -52,6 +52,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends git \
 ADD ./tpls/customizations/llvm/llvm_pr71968_mod.diff /
 RUN cd /llvm-project && git apply /llvm_pr71968_mod.diff && rm /llvm_pr71968_mod.diff
 
+# Apply customization for https://github.com/NVIDIA/cuda-quantum/issues/1799
+ADD ./tpls/customizations/llvm/fix_region_simplification.diff /
+RUN cd /llvm-project && git apply /fix_region_simplification.diff && rm /fix_region_simplification.diff
+
 # Build the the LLVM libraries and compiler toolchain needed to build CUDA-Q;
 # The safest option to avoid any compatibility issues is to build an application using these libraries 
 # with the same compiler toolchain that the libraries were compiled with.
--- a/tpls/customizations/llvm/fix_region_simplification.diff
+++ b/tpls/customizations/llvm/fix_region_simplification.diff
@@ -1,0 +1,20 @@
+diff --git a/mlir/lib/Transforms/Utils/RegionUtils.cpp b/mlir/lib/Transforms/Utils/RegionUtils.cpp
+index 996588243..63ab385be 100644
+--- a/mlir/lib/Transforms/Utils/RegionUtils.cpp
++++ b/mlir/lib/Transforms/Utils/RegionUtils.cpp
+@@ -679,6 +679,15 @@ static LogicalResult mergeIdenticalBlocks(RewriterBase &rewriter,
+       if (hasNonEmptyRegion)
+         continue;
+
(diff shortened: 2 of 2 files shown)


schweitzpgi (approved):
+1

PR #1890, opened 2024-07-05
[WIP] Work on new target configuration setup

1tnguyen:
### Description

[TODO] Add description....


runtime/cudaq/platform/mqpu/remote-mqpu.yml
```diff
@@ -0,0 +1,56 @@
...
+  platform-qpu: RemoteSimulatorQPU
+  preprocessor-defines: ["-D CUDAQ_REMOTE_SIM"]
+  link-libs: ["-lcudaq-remote-simulator-qpu"]
+  # Ability to inject bash commands to be executed by the nvq++ main driver script
+  bash: |
+    if ${LIBRARY_MODE}; then
+      # In library mode, we enforce that no custom fallback host compiler has been set via `cmake-host-compiler` option.
+      # Note: even if another Clang-based compiler is set, we cannot guarantee that it is compatible with the LLVM version
+      # that the CUDA-Q runtime is compiled against. Hence, we don't support custom host compiler for this platform.
+      if [[ "$CMAKE_FALLBACK_HOST_CXX" != "$CXX" ]]; then
```
bettinaheim, line 25: As briefly discussed offline, the option to configure a host compiler will be removed with PR 1603. After this change, the bash code here can be removed such that we have this all be configuration flags only.

docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
```diff
@@ -1,6 +1,6 @@
 // Compile and run with:
 // ```
-// nvq++ bernstein_vazirani.cpp -o bv.x --target nvidia-mgpu
+// nvq++ bernstein_vazirani.cpp -o bv.x --target nvidia --nvidia-option mgpu
```
bmhowe23, line 3: I can't quite put my finger on it, but `--nvidia-option` doesn't quite sound right to me. The whole command line argument is a command line "option", so it seems redundant to put it in the name.

Anyway, I don't have a better suggestion, and I realize we're trying to wrap up this release, so I realize we may not do anything about this, but I figured I'd make a note just in case it spurred any better naming ideas from anybody else.

include/cudaq/Support/TargetConfig.h
```diff
@@ -0,0 +1,190 @@
...
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace cudaq {
+namespace config {
+/// Flag to enable feature(s) of the unified NVIDIA target.
+// Use bitset so that we can combine different options, e.g., multi-gpu with
+// fp32/64.
+enum TargetFeatureFlag : unsigned {
```
bmhowe23, line 21: Recommend using scoped enums here ("enum class") if possible (to avoid enum names leaking into the parent namespace). If that makes some of the bit arithmetic more complicated, then feel free to disregard this comment.

include/cudaq/Support/TargetConfig.h
```diff
@@ -0,0 +1,190 @@
...
+  flagsFP32 = 0x0001,
+  flagsFP64 = 0x0002,
+  flagsMgpu = 0x0004,
+  flagsMqpu = 0x0008,
+};
+
+/// @brief Configuration argument type annotation
+// e.g., to support type validation.
+enum ArgumentType { String, Int, UUID, FeatureFlag };
```
bmhowe23, line 30: Recommend using scoped enums here ("enum class") if possible (to avoid enum names leaking into the parent namespace).

lib/Support/Config/TargetConfig.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+/// @brief Convert the backend config entry into nvq++ compatible script.
+static std::string processSimBackendConfig(
+    const std::string &targetName,
+    const cudaq::config::BackendEndConfigEntry &configValue) {
+  std::stringstream output;
+  if (configValue.GenTargetBackend.has_value()) {
+    output << "GEN_TARGET_BACKEND="
+           << (configValue.GenTargetBackend.value() ? "true" : "false") << "\n";
+  }
```
bmhowe23, line 46: I'm willing to be overridden here, but there are many instances in this file and other files of unnecessary braces, and I was under the impression that we were following this guidance: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (configValue.GenTargetBackend.has_value())
    output << "GEN_TARGET_BACKEND="
           << (configValue.GenTargetBackend.value() ? "true" : "false") << "\n";
```
I'll hold off on littering the review with other comments like this until we make a decision here.

lib/Support/Config/TargetConfig.cpp
```diff
@@ -0,0 +1,338 @@
...
+          return (nvqppArgKey == argsStr) || (argsStr == argConfig.KeyName);
+        });
+    if (iter != config.TargetArguments.end()) {
+      if (iter->Type != cudaq::config::ArgumentType::FeatureFlag) {
+        // If this is a platform option (platform argument key is provide),
+        // forward the value to the platform extra arguments.
+        if (!iter->PlatformArgKey.empty()) {
+          platformExtraArgs << ";" << iter->PlatformArgKey << ";"
+                            << targetArgv[idx + 1];
+        }
```
bmhowe23, line 171: Suggest adding this range check here to be sure we don't access outside of the size of `targetArgv`.
```suggestion
        if (!iter->PlatformArgKey.empty() && idx + 1 < targetArgv.size()) {
          platformExtraArgs << ";" << iter->PlatformArgKey << ";"
                            << targetArgv[idx + 1];
        }
```

lib/Support/Config/TargetConfig.cpp
```diff
@@ -0,0 +1,338 @@
...
+      } else {
+        // This is an option flag, construct the value for mapping selection.
+        const auto featureFlags = targetArgv[idx + 1];
```
bmhowe23, line 174: Same here. Perhaps even add a new `else` branch to throw an error.
```suggestion
      } else if (idx + 1 < targetArgv.size()) {
        // This is an option flag, construct the value for mapping selection.
        const auto featureFlags = targetArgv[idx + 1];
```

python/cudaq/__init__.py
```diff
@@ -183,6 +183,8 @@ def __clearKernelRegistries():
     initKwargs['target'] = sys.argv[sys.argv.index('-target') + 1]
 if '--target' in sys.argv:
     initKwargs['target'] = sys.argv[sys.argv.index('--target') + 1]
+if '--option' in sys.argv:
+    initKwargs['option'] = sys.argv[sys.argv.index('--option') + 1]
```
bmhowe23, line 187: This would be a different command-line interface than C++, no?
Just off the cuff, I wonder if we should choose `--target-option` for both?

docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
khalatepradnya: I think LLVM / MLIR passes follow the `--<name>-option` style. So, it will change per target. How about we call it generically `--target-option`? 
So this would be - ```nvq++ bernstein_vazirani.cpp -o bv.x --target nvidia --target-option mgpu```.

scripts/validate_container.sh
```diff
@@ -55,23 +55,25 @@ requested_backends=`\
...
     echo "default"
-    for file in $(ls $CUDA_QUANTUM_PATH/targets/*.config); \
+    for file in $(ls $CUDA_QUANTUM_PATH/targets/*.yml); \
     do basename $file | cut -d "." -f 1; \
     done`
 
 # remote_rest targets are automatically filtered, 
 # so is execution on the photonics backend
+# TODO: This will test all NVIDIA-derivative targets in the legacy mode,
+# i.e., nvidia-fp64, nvidia-mgpu, nvidia-mqpu, etc.
```
bmhowe23, line 65: Is this still a `TODO`?

test/TargetConfig/RegressionValidation/oqc.config
bmhowe23, line 1: That was a good idea to test all these config files like this 👍

docs/sphinx/examples/cpp/algorithms/bernstein_vazirani.cpp
bmhowe23: > How about we call it generically `--target-option`?

Lol, I think I was coming to that idea later in the review :)

> I think LLVM / MLIR passes follow the --<name>-option style. 

I don't think I've seen that before. I've seen (and used) this style before, though: `--canonicalize='enable-patterns=AnyPattern region-simplify=aggressive'`. I'm always curious about the various ways to set those CLI options for LLVM/MLIR. Do you know of a good resource to bookmark on those? And are those `--<name>-option` or `--<name>-<option>`?

1tnguyen: I'd vote for `--target-option`. 

We have the convention that target-specific options are prefixed with the target name for filtering purposes (not sending to the compiler).
`--target-option` seems synonymous with `--<target name>-option`, hence would be reasonable to be added to the CLI options filter logic.

> I'd like to get rid of the -- prefixes in our driver to be more like other C++ compilers.
That's a good suggestion. We might circle back to it after wrapping up this release since this would be a breaking change regarding existing options.

include/cudaq/Support/TargetConfig.h
1tnguyen: Yea, this would be tricky to convert to an `enum class`. I added a code comment for this fact.

lib/Support/Config/TargetConfig.cpp
1tnguyen: Agreed. This code is part of the LLVM part of the CUDA-Q codebase so should follow LLVM coding standards. 

Fixed in https://github.com/NVIDIA/cuda-quantum/pull/1890/commits/6b49f72f2a6e2831a3060526ce3dbc9115e8aeb8

python/cudaq/__init__.py
1tnguyen: Agree. It is now `--target-option` for both (C++ and Python CLI).

bmhowe23 (approved):
I reviewed the changes since last time. This LGTM.

khalatepradnya:
Great work, @1tnguyen! 👏🏼

PR #892, opened 2023-11-09
[rt] Reimplementation of `from_state`

boschmitt:
### Description
This PR reimplements `from_state` following the description in https://arxiv.org/pdf/quant-ph/0407010.pdf

Closes #819 and #906

PR #1891, opened 2024-07-05
State preparation after synthesis

annagrin:
### Description
- Add a State Preparation pass after synthesis to convert state data to quantum gates
  - Used algorithm from https://github.com/NVIDIA/cuda-quantum/pull/892 for gate building
- Use `LiftAlloc` pass to convert array initialization into a global constant array
  - Add a few const prop patterns to const prop truncations and extensions
  - Make sure we handle cases of arrays created inside kernels and passed as parameters
- Create global arrays with dense element attributes for arrays flowing into quake.init_state
  - Update `LiftAlloc`, `QuakeSynthesis`
  - Allows lowering the code to LLVM in scenarios with or without state prep
- Unify common code, add helpers in `intrinsics.cpp` for global constant array creation for various types
- Add tests


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -476,6 +476,10 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
   PassManager pm(context);
   pm.addPass(cudaq::opt::createQuakeSynthesizer(name, rawArgs));
   pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
+  pm.addPass(cudaq::opt::createLiftArrayAllocPass());
+  pm.addPass(cudaq::opt::createStatePreparation(name));
```
amccaskey, line 481: I think that this code implies that this state decomposition pass will be run for any Python kernel with qvector init from state, even for simulators. But I would think we only want to run this when we are targeting physical QPUs. For simulators we can just keep forwarding the data like we've done before.

targettests/execution/from_state_complex.cpp
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
```
amccaskey, line 9: How does this test that the state decomposition pass has run? Correct me if I'm wrong, but as-is I think this just compiles with MLIR and runs against the simulator, and the RemoteRESTQPU is never used. Do you not have to test with something like `--target quantinuum --emulate` ? 

I think it would also be good to have a FileCheck test in the `tests/Quake` folder that explictly tests that the init_state op was decomposed into quake quantum operations.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: I will try to figure out what the backend is and add a check

targettests/execution/from_state_complex.cpp
annagrin: Will do, thanks!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: Alternatively, you can add the pass in the hardware configuration file, like `runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.config`

targettests/execution/from_state_complex.cpp
annagrin: Updated the test and also added a test in tests/Quake

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Added condition to avoid running state prep for simulators.

annagrin: >Alternatively, you can add the pass in the hardware configuration file, like 

I have this too, but this API is just running synthesis without loading the pipelines from the config files. So added a condition for now.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -381,6 +382,12 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
     moduleOp.push_back(func.clone());
     moduleOp->setAttrs(m_module->getAttrDictionary());
 
+    for (auto &op : m_module.getOps()) {
+      if (auto globalOp = dyn_cast<cudaq::cc::GlobalOp>(op)) {
+        moduleOp.push_back(globalOp.clone());
+      }
+    }
```
bmhowe23, line 389: Nit:
```suggestion
    for (auto &op : m_module.getOps())
      if (auto globalOp = dyn_cast<cudaq::cc::GlobalOp>(op))
        moduleOp.push_back(globalOp.clone());
```

include/cudaq/Optimizer/Builder/Intrinsics.h
```diff
@@ -86,13 +86,40 @@ class IRBuilder : public mlir::OpBuilder {
...
+                       const std::vector<std::complex<double>> &values);
   cc::GlobalOp
-  genVectorOfComplexConstant(mlir::Location loc, mlir::ModuleOp module,
-                             mlir::StringRef name,
-                             const std::vector<std::complex<float>> &values);
+  genVectorOfConstants(mlir::Location loc, mlir::ModuleOp module,
+                       mlir::StringRef name,
+                       const std::vector<std::complex<float>> &values);
+
+  cc::GlobalOp genVectorOfConstants(mlir::Location loc, mlir::ModuleOp module,
```
schweitzpgi, line 97: This looks like a perfect case for a template.

include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -36,10 +36,12 @@ createApplyOpSpecializationPass(bool computeActionOpt);
 std::unique_ptr<mlir::Pass> createDelayMeasurementsPass();
 std::unique_ptr<mlir::Pass> createExpandMeasurementsPass();
 std::unique_ptr<mlir::Pass> createLambdaLiftingPass();
+std::unique_ptr<mlir::Pass> createLiftArrayAllocPass();
```
schweitzpgi, line 39:
```suggestion
```
Use `createLiftArrayAlloc()`. It's already declared.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -532,6 +532,52 @@ def ObserveAnsatz : Pass<"observe-ansatz", "mlir::func::FuncOp"> {
...
+        %cst_1 = arith.constant 0.78539816339744839 : f64
+        quake.ry (%cst_1) %2 : (f64, !quake.ref) -> ()
+        quake.x [%1] %2 : (!quake.ref, !quake.ref) -> ()
+        return
+      }
+    }
+    ```
+  }];
+
+  let constructor = "cudaq::opt::createStatePreparation()";
```
schweitzpgi, line 578:
```suggestion
```
The old way is no longer needed.

include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -36,10 +36,12 @@ createApplyOpSpecializationPass(bool computeActionOpt);
...
 std::unique_ptr<mlir::Pass> createLowerToCFGPass();
 std::unique_ptr<mlir::Pass> createObserveAnsatzPass(std::vector<bool> &);
 std::unique_ptr<mlir::Pass> createQuakeAddMetadata();
 std::unique_ptr<mlir::Pass> createQuakeAddDeallocs();
+std::unique_ptr<mlir::Pass> createStatePreparation();
```
schweitzpgi, line 44:
```suggestion
```
Use `createPrepareState()` which is already declared and defined. (If you prefer `createStatePreparation()` (I sort of do), change the name of the pass in Passes.td accordingly.

lib/Optimizer/Transforms/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ add_cudaq_library(OptTransforms
   QuakeSynthesizer.cpp
   RefToVeqAlloc.cpp
   RegToMem.cpp
+  StateDecomposer.cpp
+  StatePreparation.cpp
   PySynthCallableBlockArgs.cpp
```
schweitzpgi, line 49: Can you move the PySynth... file name to maintain alphabetical order here? Thanks in advance.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -38,11 +96,12 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
     SmallVector<Operation *> stores;
     bool toGlobal = false;
-    if (!isGoodCandidate(alloc, stores, dom, toGlobal))
+    if (!isGoodCandidate(alloc, stores, dom, toGlobal)) {
       return failure();
+    }
 
     LLVM_DEBUG(llvm::dbgs() << "Candidate was found\n");
-    auto arrTy = cast<cudaq::cc::ArrayType>(alloc.getElementType());
+    auto arrTy = cast<cudaq::cc::ArrayType>(alloc.getType().getElementType());
```
schweitzpgi, line 104: This change can be reverted. The AllocaOp class has a getElementType(), which is the call we want to use here.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -69,54 +128,18 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
-        } else {
-          OpBuilder::InsertionGuard guard(rewriter);
-          rewriter.setInsertionPointToEnd(module.getBody());
-          rewriter.create<cudaq::cc::GlobalOp>(loc, arrTy, name, valuesAttr,
-                                               /*isConstant=*/true,
-                                               /*isExternal=*/false);
-        }
-      }
+      cudaq::IRBuilder irBuilder(rewriter.getContext());
+      genVectorOfConstantsFromAttributes(irBuilder, loc, module, name, values,
```
schweitzpgi, line 132: I'm not sure I like this change or not.

This code was converting all the vectors, regardless of type to globals of vector of complex of double here at compile time. That costs compilation time, some memory, and some disk/file space for the executable, etc.

It looks like these changes do not do that conversion, but kick the can down the road. So we can end up with a family of vectors of all sorts of floating point types. That may save a some disk space, but every time the executable runs, those arrays will need to be converted to the API of the runtime, which I believe is vector of complex of double.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -128,30 +151,44 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
     } else {
-      rewriter.eraseOp(alloc);
+      toErase.push_back(alloc);
+    }
+
+    for (auto *op : toErase) {
+      if (op->getUses().empty()) {
+        rewriter.eraseOp(op);
+      } else {
+        module.emitOpError("LiftArrayAlloc failed to remove quake.init_state "
```
schweitzpgi, line 188: Can you provide a test case that reaches this error?

It appears the only ops on the list are cc::StoreOp which will never have users.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -182,8 +219,8 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
     if (std::distance(alloc->getUses().begin(), alloc->getUses().end()) < size)
       return false;
 
-    // Keep a scoreboard for every element in the array. Every element *must* be
-    // stored to with a constant exactly one time.
+    //  Keep a scoreboard for every element in the array. Every element *must*
```
schweitzpgi, line 222: Can revert these changes. No need to add an extra space hee.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -321,6 +366,88 @@ class ComplexCreatePattern : public OpRewritePattern<complex::CreateOp> {
   }
 };
 
+// Fold arith.trunc ops if the argument is constant.
+class FloatTruncatePattern : public OpRewritePattern<arith::TruncFOp> {
```
schweitzpgi, line 370: I'm surprised these arith ops are showing up at all. Where are they coming from? Why isn't cc::CastOp used? Or handled?

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -338,6 +465,10 @@ class LiftArrayAllocPass
       RewritePatternSet patterns(ctx);
       patterns.insert<AllocaPattern>(ctx, domInfo, funcName, module);
       patterns.insert<ComplexCreatePattern>(ctx);
+      patterns.insert<FloatExtendPattern>(ctx);
+      patterns.insert<FloatTruncatePattern>(ctx);
+      patterns.insert<ComplexRePattern>(ctx);
+      patterns.insert<ComplexImPattern>(ctx);
```
schweitzpgi, line 471: These patterns seem like they should already be part of canonicalization. Are they? If not, is it because we're using an old version of MLIR?

If MLIR isn't canonicalizing these, we should use cc::CastOp and add these patterns to its canonicalization patterns. This functionality isn't conceptually part of lifting array values.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -249,11 +286,19 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
               scoreboard[0] = w;
               continue;
             }
-          return false;
+          // can be a cast only used for a quake.init_state or vector init
```
schweitzpgi, line 288: Generally speaking, the answer is no. A cast could be used to recast the data for a call to some other function, etc.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -249,11 +286,19 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
+          continue;
+        } else {
+          if (getWriteOp(cast, 0)) {
+            LLVM_DEBUG(llvm::dbgs()
+                       << "another cast used in store: " << *op << '\n');
+            return false;
+          }
+          // can be a cast only used for a quake.init_state or vector init
+          toGlobal = true;
+          continue;
```
schweitzpgi, line 298: These changes look suspect to me. I assume they are to workaround the problems of floating-point conversions being kicked down the road and having floating point extensions and truncations being inserted?

There is only 1 truly safe cast that should be allowed. Specifically, casting the array of element types pointer to a singleton of element type for the single intention of writing a value of element type to that position in memory. Casting to arbitrary types other than the element type or having random other operations on the casting of the array's address would open the door to bugs.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Not sure I understand - I just moved the code into helpers and used a common API (plus made vectors of float treated the same way as vectors of complex). I don't think any ordering should have changed, please correct me if I am wrong.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I've run into it with loops that create arrays (like when python inlines the code that converts arrays to the simulation precision, or if a function that has a loop for initializing array is inlined)

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I encountered them when running for 32 bit precision (ie quantinuum) - the IR had float constants of double precision and truncation operation on them before creating complex

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: The Trunc and Ext are definitely not handled (my new tests will fail without them) but I Re and IM pattern was my try to const prop conversion to the correct simulation precision (which is not working yet, plan to address it in another PR).

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I haven't seen casts before, but I can add a pattern when I encounter them

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: I agree and I even tried:) That made a lot of helper code move to intrinsics.h - probably most of the cpp file - would this be ok?

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
schweitzpgi: I'm getting a different result regarding this. I'll send you the test.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: No transformations should be moved from their original places, and I added a removal of ext and trunc so it is done before we get here. I just had more test cases coming from my tests and made sure they all worked (ie globals coming from ConvertExpr, LiftArray, QuakeSynth, as well as the test you added for unitary synthesis. Those changes make all of those test cases work. This specific line is only needed for the vector init, which is the unitary synthesis case.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: > we should use cc::CastOp
This would be a change in ConvertExpr which I would like to do separately

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Examples of an extra cast:

Unitary synthesis:
```
%9 = cc.cast %4 : (!cc.ptr<!cc.array<complex<f64> x 4>>) -> !cc.ptr<i8>
%10 = call @__nvqpp_vectorCopyCtor(%9, %c4_i64, %c16_i64) : (!cc.ptr<i8>, i64, i64) -> !cc.ptr<i8>
%11 = cc.stdvec_init %10, %c4_i64 : (!cc.ptr<i8>, i64) -> !cc.stdvec<complex<f64>>
```

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -321,6 +366,88 @@ class ComplexCreatePattern : public OpRewritePattern<complex::CreateOp> {
...
+      rewriter.replaceOpWithNewOp<arith::ConstantFloatOp>(
+          extend, APFloat{static_cast<double>(v)}, fTy);
+      return success();
+    }
+    return failure();
+  }
+};
+
+// Fold complex.re ops if the argument is constant.
+class ComplexRePattern : public OpRewritePattern<complex::ReOp> {
```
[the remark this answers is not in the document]
annagrin: I think not - see below, thanks!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -321,6 +366,88 @@ class ComplexCreatePattern : public OpRewritePattern<complex::CreateOp> {
...
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(complex::ImOp im,
+                                PatternRewriter &rewriter) const override {
+    auto val = im.getOperand();
+    auto valCon = val.getDefiningOp<complex::ConstantOp>();
+    if (valCon) {
+      auto attr = valCon.getValue();
+      auto real = cast<FloatAttr>(attr[0]).getValue();
```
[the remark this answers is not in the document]
annagrin: Thanks, you caught a copy-paste error! Fixed.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -462,6 +462,18 @@ py::object pyAltLaunchKernelR(const std::string &name, MlirModule module,
   return returnValue;
 }
 
+/// @brief Helper function to get boolean environment variable
+bool getEnvBool(const char *envName, bool defaultVal = false) {
```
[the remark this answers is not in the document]
annagrin: Done, thanks!

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: Will leave this to next PR as discussed.

include/cudaq/Optimizer/Transforms/Passes.h
annagrin: Done, thanks!

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: Removed, thanks!

include/cudaq/Optimizer/Transforms/Passes.h
annagrin: Renamed, thanks!

lib/Optimizer/Transforms/CMakeLists.txt
annagrin: done!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Done!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: done!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Will add those to canonicalizer, among other things, in a separate PR, as discussed. For now, I moved all the patterns not related to lifting arrays to a new pass, ConstPropComplex

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Will add those to canonicalizer, among other things, in a separate PR, as discussed.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Reverted, thanks!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Reverted the changes and added a separate pass that removes the extra cast.

targettests/execution/state_preparation_vector.cpp
```diff
@@ -0,0 +1,159 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --enable-mlir --target quantinuum --emulate %s -o %t && %t | FileCheck %s
```
bmhowe23, line 10: Nit: `--enable-mlir` is not needed when specifying the hardware targets because the hardware targets' config files say `LIBRARY_MODE=false`.

bmhowe23, line 10: This is a nice test for a something new, so it might be nice to run this test against all of our hardware targets, like this:
```
// RUN: nvq++ --target ionq                     --emulate %s -o %t && %t | FileCheck %s
// 2 different IQM machines for 2 different topologies
// RUN: nvq++ --target iqm --iqm-machine Adonis --emulate %s -o %t && %t | FileCheck %s
// RUN: nvq++ --target iqm --iqm-machine Apollo --emulate %s -o %t && %t | FileCheck %s
// RUN: nvq++ --target oqc                      --emulate %s -o %t && %t | FileCheck %s
```

bmhowe23 (approved):
This LGTM. Thanks, Anna.

targettests/execution/state_preparation_vector.cpp
annagrin: Thanks, will remove

annagrin: thanks, will add!

annagrin: done

annagrin: done

include/cudaq/Optimizer/Builder/Intrinsics.h
```diff
@@ -86,13 +86,40 @@ class IRBuilder : public mlir::OpBuilder {
...
+                                    mlir::StringRef name,
+                                    const std::vector<std::int16_t> &values);
+
+  cc::GlobalOp genVectorOfConstants(mlir::Location loc, mlir::ModuleOp module,
+                                    mlir::StringRef name,
+                                    const std::vector<std::int8_t> &values);
+
+  cc::GlobalOp genVectorOfConstants(mlir::Location loc, mlir::ModuleOp module,
```
schweitzpgi, line 120: Are there specific tests for all of these?

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,210 @@
...
+namespace {
+
+// Replace array ptr casts that throw away the size by a cast to element
+// pointer.
+//
+//%1 = cc.cast %0 : (!cc.ptr<!cc.array<complex<f32> x 4>>) ->
+//! cc.ptr<!cc.array<complex<f32> x ?>>
+// ->
+//%1 = cc.cast %0 : (!cc.ptr<!cc.array<complex<f32> x 4>>) ->
+//! cc.ptr<complex<f32>>
```
schweitzpgi, line 39: Looks like clang-format may have formatted this oddly? Could use some whitespace between the `//` and the code here.

nit: we conventionally describe a rewrite using a horizontal line. (See e.g. Peephole.td)

```
      2
    + 1
  ─────
      3
```

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,210 @@
...
+    auto toTy = cast.getType();
+
+    if (auto ptrFromTy = dyn_cast<cudaq::cc::PointerType>(fromTy)) {
+      if (auto arrayFromTy =
+              dyn_cast<cudaq::cc::ArrayType>(ptrFromTy.getElementType())) {
+        if (auto ptrToTy = dyn_cast<cudaq::cc::PointerType>(toTy)) {
+          if (auto arrayToTy =
+                  dyn_cast<cudaq::cc::ArrayType>(ptrToTy.getElementType())) {
+            if (arrayFromTy.getElementType() == arrayToTy.getElementType()) {
+              auto eleTy = arrayFromTy.getElementType();
```
schweitzpgi, line 57: The type `cc.ptr<cc.array<T x ?>>` and `cc.ptr<T>` are not the same and not interchangeable. This rewrite may possibly fail as a consequence.

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,210 @@
...
+          ArrayRef<Attribute>{reCon.getValue(), imCon.getValue()});
+      rewriter.replaceOpWithNewOp<complex::ConstantOp>(create, create.getType(),
+                                                       aa);
+      return success();
+    }
+    return failure();
+  }
+};
+
+// Fold arith.trunc ops if the argument is constant.
```
schweitzpgi, line 94: Where are the arith.trunc ops being generated? We need to file an issue against that code. It should be generating cc.cast.

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,210 @@
...
+      auto fTy = dyn_cast<FloatType>(truncate.getType());
+      rewriter.replaceOpWithNewOp<arith::ConstantFloatOp>(
+          truncate, APFloat{static_cast<float>(v)}, fTy);
+      return success();
+    }
+    return failure();
+  }
+};
+
+// Fold arith.ext ops if the argument is constant.
```
schweitzpgi, line 114: Same as previous comment.

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,210 @@
...
+        continue;
+      DominanceInfo domInfo(func);
+      std::string funcName = func.getName().str();
+      RewritePatternSet patterns(ctx);
+      patterns.insert<ComplexCreatePattern>(ctx);
+      patterns.insert<FloatExtendPattern>(ctx);
+      patterns.insert<FloatTruncatePattern>(ctx);
+      patterns.insert<ComplexRePattern>(ctx);
+      patterns.insert<ComplexImPattern>(ctx);
+      patterns.insert<CastArrayPtrPattern>(ctx);
```
schweitzpgi, line 196: nit: These can be combined into a single line.

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
```diff
@@ -105,8 +105,8 @@ class GenerateDeviceCodeLoader
           LLVM_DEBUG(llvm::dbgs() << "adding declaration: " << op);
           declarations.push_back(&op);
         }
-      } else if (auto ccGlobalOp = dyn_cast<cudaq::cc::GlobalOp>(op)) {
-        LLVM_DEBUG(llvm::dbgs() << "adding global constants: " << op);
+      } else if (auto globalOp = dyn_cast<cudaq::cc::GlobalOp>(op)) {
+        LLVM_DEBUG(llvm::dbgs() << "adding global: " << op);
```
schweitzpgi, line 109: Is this not a constant any longer? That's a departure, since we shouldn't be creating global variables AFAIK.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -27,6 +27,64 @@ namespace cudaq::opt {
...
+      auto vals = readConstantValues<double>(values, floatTy);
+      irBuilder.genVectorOfConstants(loc, module, name, vals);
+      return;
+    } else if (floatTy == irBuilder.getF32Type()) {
+      auto vals = readConstantValues<float>(values, floatTy);
+      irBuilder.genVectorOfConstants(loc, module, name, vals);
+      return;
+    }
+  }
+  assert(false && "unexpected element type in constant array");
```
schweitzpgi, line 84: I don't think this pass should crash if it sees more general code that it is unsafe to transform in this way.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
schweitzpgi: Don't the helpers create one of `vector<float>`, `vector<double>`, `vector<complex<float>>`, etc.?

Previously this created only `vector<complex<double>>` and the other possibilities did not exist intentionally.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -128,30 +150,43 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
     if (toGlobal) {
+      rewriter.setInsertionPointAfter(alloc);
       rewriter.replaceOp(alloc, conGlobal);
     } else {
-      rewriter.eraseOp(alloc);
+      toErase.push_back(alloc);
+    }
+
+    for (auto *op : toErase) {
+      if (op->getUses().empty()) {
```
schweitzpgi, line 183: Stores will never have uses.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -182,8 +217,8 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
     if (std::distance(alloc->getUses().begin(), alloc->getUses().end()) < size)
       return false;
 
-    // Keep a scoreboard for every element in the array. Every element *must* be
-    // stored to with a constant exactly one time.
+    // Keep a scoreboard for every element in the array. Every element *must*
+    // be stored to with a constant exactly one time.
```
schweitzpgi, line 221: clang-format appears to have unformatted this along the way. (Should revert this one, it appears.)

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -104,63 +104,58 @@ Value makeComplexElement(OpBuilder &builder, Location argLoc,
   return builder.create<complex::ConstantOp>(argLoc, eleTy, complexVal);
 }
 
-/// returns true if and only if \p argument is used by a `quake.init_state`
-/// operation.
-static bool hasInitStateUse(BlockArgument argument) {
```
schweitzpgi, line 109: We definitely do not want all arrays to be converted to global constants.

The original purpose of this code still holds. Namely, we want to lift a vector of constants to a value so that all the constant propagation works correctly and we can build a proper quantum circuit.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -230,7 +225,7 @@ synthesizeVectorArgument(OpBuilder &builder, ModuleOp module, unsigned &counter,
           Value runtimeParam =
               makeElementValue(builder, argLoc, vec[index], eleTy);
           if (failed(replaceLoads(elePtrOp, runtimeParam))) {
-            Value memArr = getArrayInMemory();
+            auto [memArr, _] = getArrayInMemory();
```
schweitzpgi, line 228: Prefer not to use Python-like style.

lib/Optimizer/Transforms/StateDecomposer.cpp
```diff
@@ -0,0 +1,128 @@
+/****************************************************************-*- C++ -*-****
```
schweitzpgi, line 1:
```suggestion
/*******************************************************************************
```
nit: use the emacs tag in files ending in `.h`, not `.cpp`

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -565,6 +573,50 @@ def ObserveAnsatz : Pass<"observe-ansatz", "mlir::func::FuncOp"> {
   ];
 }
 
+def StatePreparation : Pass<"state-prep", "mlir::ModuleOp"> {
```
schweitzpgi, line 576: Why is StateDecomposer not listed here?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+          auto globalName = addr.getGlobalName();
+          auto symbol = module.lookupSymbol(globalName);
+          if (auto global = dyn_cast<cudaq::cc::GlobalOp>(symbol)) {
+            // Read state initialization data from the global array.
+            auto vec = readConstantArray(builder, global);
+
+            // Prepare state from vector data.
+            auto gateBuilder = StateGateBuilder(builder, loc, qubits);
+            auto decomposer = StateDecomposer(gateBuilder, vec);
```
schweitzpgi, line 152: It looks like the only place this is used is in the StatePrep pass. Are there plans to use it anywhere else? If not, why not put the code in StatePrep.cpp?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,209 @@
...
+  return result;
+}
+
+LogicalResult transform(ModuleOp module, func::FuncOp funcOp) {
+  auto builder = OpBuilder::atBlockBegin(&funcOp.getBody().front());
+  auto toErase = std::vector<mlir::Operation *>();
+  auto hasInitState = false;
+  auto replacedInitState = false;
+
+  funcOp->walk([&](Operation *op) {
```
schweitzpgi, line 125: Why are we using `walk` and building a state machine here?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,209 @@
...
+///     %0 = cc.address_of @foo.rodata_0 : !cc.ptr<!cc.array<complex<f32> x 4>>
+///     %1 = quake.alloca !quake.veq<2>
+///     %2 = quake.init_state %1, %0 : (!quake.veq<2>,
+///       !cc.ptr<!cc.array<complex<f32> x 4>>) -> !quake.veq<2> return
+///  }
+///  cc.global constant @foo.rodata_0 (dense<[(0.707106769,0.000000e+00),
+///      (0.707106769,0.000000e+00), (0.000000e+00,0.000000e+00),
+///      (0.000000e+00,0.000000e+00)]> : tensor<4xcomplex<f32>>) :
+///    !cc.array<complex<f32> x 4>
+/// }
```
schweitzpgi, line 56: I'm a bit surprised this unmarked code block in the doxygen comments doesn't choke the spelling checker in the CI.

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  if (hasInitState && !replacedInitState) {
+    funcOp.emitOpError("StatePreparation failed to replace quake.init_state");
+    return failure();
+  }
+
+  for (auto &op : toErase) {
+    if (op->getUses().empty()) {
+      op->erase();
```
schweitzpgi, line 172: It isn't necessary for a pass to keep track of and delete ops that are unused in general. It's less tedious and clearer code to just let region simplification get rid of the dead code.

That said, an operation such as cc::StoreOp has side-effects and as such region simplification will not erase those.

include/cudaq/Optimizer/Transforms/Passes.td
schweitzpgi: It looks like it is a part of StatePrep and not its own pass.

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: Yes, I had some tests that caught bugs in this code: https://github.com/NVIDIA/cuda-quantum/blob/9b9124c9747f1c11fdfd587b4c2bb68596db4832/targettests/Remote-Sim/args_synthesis.cpp#L1

They are very thorough, so I didn't add anything else.

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: will update, thanks!

test/Quake/state_prep.qke
```diff
@@ -0,0 +1,114 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -state-prep %s | FileCheck %s
```
schweitzpgi, line 9:
```suggestion
// RUN: cudaq-opt -state-prep -canonicalize %s | FileCheck %s
```
This would simplify the output and reduce the number of lines, I think.

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: I can try just preserving the original type (ie removing the cast that removes array size) and see if it works.

test/AST-Quake/custom_op_concrete_matrix.cpp
```diff
@@ -6,7 +6,7 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-// RUN: cudaq-quake %cpp_std %s | cudaq-opt -lift-array-value -get-concrete-matrix | FileCheck %s
+// RUN: cudaq-quake %cpp_std %s | cudaq-opt -const-prop-complex -lift-array-value -get-concrete-matrix | FileCheck %s
```
schweitzpgi, line 9: This doesn't appear to change the test? Is it not needed?

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: I suspect this code: https://github.com/NVIDIA/cuda-quantum/blob/9b9124c9747f1c11fdfd587b4c2bb68596db4832/lib/Frontend/nvqpp/ConvertExpr.cpp#L659

Will file an issue (+ adding complex const prop to the canonicalizer)

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: ack

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I think we are removing `alloc` sometimes, that's where I saw it fail once. Just wanted an error message to find if it happens again, but I can remove this if you prefer.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: this is done by the clang format though :)

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
annagrin: I'll add the check back.

lib/Optimizer/Transforms/StateDecomposer.cpp
annagrin: will update

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: It is not a pass - it is helper code that converts vectors to gates using a builder

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: i think it does not erase the `quake.init_state` either (it remained and failed to lower to llvm in my tests), and this is the main purpose here (plus all its dependencies)

test/Quake/state_prep.qke
annagrin: will do, thanks

test/AST-Quake/custom_op_concrete_matrix.cpp
annagrin: The test failed without it because it needs some complex const prop opts that I moved from `lift-array-value` to `const-prop-complex`

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: Btw we the IR for constant arrays (without the new const prop pass) uses them interchangeably it seems:

```
// -----// IR Dump Before LiftArrayAlloc (lift-array-value) ('builtin.module' operation) //----- //
module attributes {llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.triple = "x86_64-unknown-linux-gnu", quake.mangled_name_map = {__nvqpp__mlirgen__function_test_complex_array_param._Z24test_complex_array_paramSt6vectorISt7complexIfESaIS1_EE = "_Z24test_complex_array_paramSt6vectorISt7complexIfESaIS1_EE", __nvqpp__mlirgen__function_test_complex_constant_array._Z27test_complex_constant_arrayv = "_Z27test_complex_constant_arrayv", __nvqpp__mlirgen__function_test_complex_constant_array2._Z28test_complex_constant_array2v = "_Z28test_complex_constant_array2v", __nvqpp__mlirgen__function_test_complex_constant_array3._Z28test_complex_constant_array3v = "_Z28test_complex_constant_array3v", __nvqpp__mlirgen__function_test_real_array_param._Z21test_real_array_paramSt6vectorIfSaIfEE = "_Z21test_real_array_paramSt6vectorIfSaIfEE", __nvqpp__mlirgen__function_test_real_constant_array._Z24test_real_constant_arrayv = "_Z24test_real_constant_arrayv"}} {
  func.func @__nvqpp__mlirgen__function_test_real_constant_array._Z24test_real_constant_arrayv() attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
    %cst = arith.constant 0.70710678118654757 : f64
    %cst_0 = arith.constant 0.000000e+00 : f64
    %0 = cc.alloca !cc.array<f64 x 4>
    %1 = cc.cast %0 : (!cc.ptr<!cc.array<f64 x 4>>) -> !cc.ptr<!cc.array<f64 x ?>>
    %2 = cc.cast %0 : (!cc.ptr<!cc.array<f64 x 4>>) -> !cc.ptr<f64>
    cc.store %cst, %2 : !cc.ptr<f64>
    %3 = cc.compute_ptr %0[1] : (!cc.ptr<!cc.array<f64 x 4>>) -> !cc.ptr<f64>
    cc.store %cst, %3 : !cc.ptr<f64>
    %4 = cc.compute_ptr %0[2] : (!cc.ptr<!cc.array<f64 x 4>>) -> !cc.ptr<f64>
    cc.store %cst_0, %4 : !cc.ptr<f64>
    %5 = cc.compute_ptr %0[3] : (!cc.ptr<!cc.array<f64 x 4>>) -> !cc.ptr<f64>
    cc.store %cst_0, %5 : !cc.ptr<f64>
    %6 = quake.alloca !quake.veq<2>
    %7 = quake.init_state %6, %1 : (!quake.veq<2>, !cc.ptr<!cc.array<f64 x ?>>) -> !quake.veq<2>
    return
  }```

File: lib/Optimizer/Transforms/StateDecomposer.cpp [hunk 41]
```diff
@@ -0,0 +1,128 @@
...
+    if (denominator == 0.0) {
+      assert(numerator == 0.0 &&
+             "If the denominator is zero, the numerator must also be zero.");
+      angles.push_back(0.0);
+      continue;
+    }
+    angles.push_back(2.0 * std::asin(std::sqrt(numerator / denominator)));
+  }
+  return angles;
+}
```
schweitzpgi, line 127: Did not review the specifics in this implementation. Maybe @boschmitt has some familiarity with the state -> gate synthesis algorithm.

lib/Optimizer/Transforms/StateDecomposer.h
```diff
@@ -0,0 +1,175 @@
...
+      : builder(b), amplitudes(a), numQubits(log2(a.size())) {}
+
+  /// @brief Decompose the input state vector data to a set of controlled
+  /// operations and rotations. This function takes as input a `OpBuilder`
+  /// and appends the operations of the decomposition to its internal
+  /// representation. This implementation follows the algorithm defined in
+  /// `https://arxiv.org/pdf/quant-ph/0407010.pdf`.
+  void decompose() {
+
+    // Decompose the state into phases and magnitudes.
```
schweitzpgi, line 106: Is it possible to move any more of this code out of the header file?

On the third hand, all of this could be in the StatePreparation file.

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,209 @@
...
+std::vector<std::complex<double>>
+readConstantArray(mlir::OpBuilder &builder, cudaq::cc::GlobalOp &global) {
+  std::vector<std::complex<double>> result{};
+
+  auto attr = global.getValue();
+  auto type = global.getType().getElementType();
+
+  auto arrayTy = dyn_cast<cudaq::cc::ArrayType>(type);
+  assert(arrayTy);
+  assert(attr.has_value());
```
schweitzpgi, line 90: nit: might use `&& "must be cc.array"` as more of a description.

I assume this is purely for debug purposes. If so, these can be removed and the `dyn_cast` above converted to `cast`. It gives the same level of assurance, etc.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -561,7 +561,7 @@ def bodyBuilder(iterVar):
                                       rawIndex).result
             loadedEle = cc.LoadOp(eleAddr).result
             castedEle = self.promoteOperandType(targetEleType, loadedEle)
-            targetEleAddr = cc.ComputePtrOp(targetElePtrType, targetPtr,
+            targetEleAddr = cc.ComputePtrOp(targetElePtrTy, targetPtr,
```
schweitzpgi, line 564: Changes in this file look like simple respellings of identifier and NFC. Could be moved to a separate PR.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -493,6 +493,18 @@ py::object pyAltLaunchKernelR(const std::string &name, MlirModule module,
   return returnValue;
 }
 
+/// @brief Helper function to get boolean environment variable
+static bool getEnvBool(const char *envName, bool defaultVal = false) {
```
schweitzpgi, line 497: Is this function copied from somewhere else? It looks familiar.

If it is a copy, we should just put it someplace "common" and expose the function to all the libraries that use it.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -493,6 +493,18 @@ py::object pyAltLaunchKernelR(const std::string &name, MlirModule module,
...
+  if (auto envVal = std::getenv(envName)) {
+    std::string tmp(envVal);
+    std::transform(tmp.begin(), tmp.end(), tmp.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+    if (tmp == "1" || tmp == "on" || tmp == "true" || tmp == "yes")
```
schweitzpgi, line 502: Should this allow anything that starts with "t" or "y"? Or is that too permissive? CMake accepts "y", for example.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -493,6 +493,18 @@ py::object pyAltLaunchKernelR(const std::string &name, MlirModule module,
...
+      return true;
```
schweitzpgi, line 503: Should this be `!defaultVal`? If the `defaultVal` is `true`, then there will be no distinction between success and failure.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -504,9 +516,26 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
...
   pm.addPass(cudaq::opt::createQuakeSynthesizer(name, rawArgs));
   pm.addPass(createCanonicalizerPass());
+
+  // Run state preparation for quantum devices only.
+  // Simulators have direct implementation of state initialization
+  // in their runtime.
+  auto &platform = cudaq::get_platform();
+  if (!platform.is_simulator() || platform.is_emulated()) {
+    pm.addPass(cudaq::opt::createConstPropComplex());
+    pm.addPass(createCSEPass());
```
schweitzpgi, line 534: Do we need CSE or will Canonicalizer suffice?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -517,6 +546,10 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
   tm.setEnabled(cudaq::isTimingTagEnabled(cudaq::TIMING_JIT_PASSES));
   auto timingScope = tm.getRootScope(); // starts the timer
   pm.enableTiming(timingScope);         // do this right before pm.run
+  if (disableMLIRthreading || enablePrintMLIREachPass)
+    context->disableMultithreading();
+  if (enablePrintMLIREachPass)
```
schweitzpgi, line 551: Hooray!

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -382,12 +382,13 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
...
     for (auto &op : m_module.getOps()) {
-      if (auto ccGlobalOp = dyn_cast<cudaq::cc::GlobalOp>(op)) {
-        moduleOp.push_back(ccGlobalOp.clone());
-      }
+      // Add any global symbols, including global constant arrays.
+      // Global constant arrays can be created during compilation,
+      // `lift-array-value`, `quake-synthesizer`, and `get-concrete-matrix`
+      // passes.
+      if (auto globalOp = dyn_cast<cudaq::cc::GlobalOp>(op))
+        moduleOp.push_back(globalOp.clone());
```
schweitzpgi, line 391: Not this PR, but a random comment. I'm not sure why we're trying to cherry-pick pieces out of the module here. It seems to create problems and if we just copied the entire module, what would be the downside? (Asking for a friend.)

test/Quake/const_prop_complex.qke
```diff
@@ -0,0 +1,133 @@
...
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -const-prop-complex %s | FileCheck %s
+
+func.func @__nvqpp__mlirgen__function_test_complex_constant_array._Z27test_complex_constant_arrayv() attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
+    %cst = arith.constant 0.000000e+00 : f32
+    %cst_0 = arith.constant 0.70710678118654757 : f64
+    %0 = arith.truncf %cst_0 : f64 to f32
```
schweitzpgi, line 14: `cc.cast`?

schweitzpgi (commented):
Ok, made it to the bottom.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
annagrin: added the check back, so globals are only generated for args used in quake.init_state.

python/cudaq/kernel/ast_bridge.py
annagrin: Ah it was a bug that sneaked into main and was fixed separately here in another PR, and then merged:) Will undo

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: will do, thanks

runtime/common/BaseRemoteRESTQPU.h
annagrin: Pradnya had the same changes too - the code was already picking a function for a new module, but did not copy globals, which we need now. Note for future - we might want to only add globals used in the function. Could this be because we send one function at a time somewhere? 

@bmhowe23 do you know why we are not just using the whole module?

targettests/execution/from_state_complex.cpp
```diff
@@ -0,0 +1,27 @@
...
+// RUN: nvq++ %cpp_std %s -o %t && %t | FileCheck %s
+
+#include <cudaq.h>
+
+__qpu__ void test(std::vector<cudaq::complex> inState) {
```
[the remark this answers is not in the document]
annagrin: of course.

targettests/execution/state_preparation_vector.cpp
```diff
@@ -0,0 +1,164 @@
...
+          printCounts(counts);
+
+          counts = cudaq::sample(kernel, vec1);
+          printCounts(counts);
+      }
+    }
+
+    {
+      std::vector<cudaq::real> vec{M_SQRT1_2, M_SQRT1_2, 0., 0.};
+      std::vector<cudaq::real> vec1{0., 0., M_SQRT1_2, M_SQRT1_2};
```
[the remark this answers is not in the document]
annagrin: will add, thanks!

test/Quake/const_prop_complex.qke
annagrin: I couldn't generate code from examples that uses a cc.cast here, so no idea how to properly test that. I'd prefer to leave it for another PR that moves all of this to the canonicalizer and replaces arith.truncf with cc.cast.

schweitzpgi: Added an issue.

targettests/execution/state_preparation_vector.cpp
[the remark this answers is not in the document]
annagrin: Looks like unmatching combinations do not compile - ie, cannot pass a vector of float to the qvector constructor if `CUDAQ_SIMULATION_SCALAR_FP32` not defined, for example, the following code does not compile, so I can only test matching combinations
```
__qpu__ void test_real_array_param(std::vector<float> inState) {
  cudaq::qvector q1 = inState;
}
```

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: will try without CSE to check it out. I think I needed it for some reason...

runtime/common/BaseRemoteRESTQPU.h
bmhowe23: > @bmhowe23 do you know why we are not just using the whole module?

I don't know for sure, but it might be that the server doesn't like receiving multiple entry-point functions. That's pure speculation, though. I would need to do more investigation to be sure.

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: removed the pattern

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: I removed the pattern and a added the case handling this cast directly in LifArrayAlloc

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: https://github.com/NVIDIA/cuda-quantum/issues/1955

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: https://github.com/NVIDIA/cuda-quantum/issues/1955

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: removed

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: run clang format again

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
annagrin: updated

lib/Optimizer/Transforms/StateDecomposer.cpp
annagrin: removed this file

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: I moved the code from StateDecomposer.h and StateDecomposer.cpp into StatePreparation.cpp

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: moved

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: Some state machine is needed - the pass succeeds if all init state instructions are removed, so I need to track it somehow. I modified the code to be less state-machine-like, let me know if it looks better.

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: added quotes

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: Had to keep it - if I don't remove any of them we either end up with invalid IR or quantum passes fail

test/Quake/state_prep.qke
annagrin: done

lib/Optimizer/Transforms/StateDecomposer.h
annagrin: moved to StatePreparation.cpp

lib/Optimizer/Transforms/StatePreparation.cpp
annagrin: updated

python/cudaq/kernel/ast_bridge.py
annagrin: done

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: done

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Hmm I presume env value, if defined, should take precedence over the default value? Updated.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Actually, neither of them is needed here anymore. I previously had to insert a CSE to cleanup my (incorrect) cast replacement that would sometimes create duplicate casts. Removed, thanks!

targettests/execution/from_state_complex.cpp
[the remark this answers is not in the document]
annagrin: added

targettests/execution/state_preparation_vector.cpp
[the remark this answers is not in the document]
annagrin: I added some tests for float and double vectors under `#ifdef` so compilation can succeed

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
annagrin: It is a constant still. I believe this is just a merge of the same change Pradnya and I had to do, so just a name change

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Thanks, will update it to continue

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I updated to only create globals if we see an argument flowing into a` quake.state_init`, so nothing should change for previously handled code

test/Quake/const_prop_complex.qke
annagrin: https://github.com/NVIDIA/cuda-quantum/issues/1955

Also added a cast const prop pattern and a quake test

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: added

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: Actually, they are missing cases for synthesizing vectors of complex. Will add tests in another PR.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Fixed! Now the pass will just create a const array instead of failing to create a global.

lib/Optimizer/Transforms/GenDeviceCodeLoader.cpp
annagrin: updated to match the current version in main

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Removed the error

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,440 @@
...
+#define GEN_PASS_DEF_STATEPREPARATION
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+#define DEBUG_TYPE "state-preparation"
+
+using namespace mlir;
+
+namespace cudaq::details {
+
```
annagrin, line 39: @boschmitt Appreciate your review of the code ported from your implementation of from_state in https://github.com/NVIDIA/cuda-quantum/pull/892

lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -0,0 +1,200 @@
...
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(cudaq::cc::CastOp cast,
+                                PatternRewriter &rewriter) const override {
+    auto val = cast.getOperand();
+    auto valCon = val.getDefiningOp<arith::ConstantFloatOp>();
+    if (valCon) {
+      auto fTy = dyn_cast<FloatType>(cast.getType());
+      if (fTy == rewriter.getF64Type()) {
+        auto v = valCon.value().convertToFloat();
```
schweitzpgi, line 67: Do we have code coverage on this line? I seem to recall some issue with this call. Unfortunately, I don't remember the specifics.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -128,30 +160,45 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
+      rewriter.setInsertionPointAfter(alloc);
       rewriter.replaceOp(alloc, conGlobal);
     } else {
-      rewriter.eraseOp(alloc);
+      toErase.push_back(alloc);
+    }
+
+    for (auto *op : toErase) {
+      rewriter.eraseOp(op);
     }
```
schweitzpgi, line 202:
```suggestion
    for (auto *op : toErase) 
      rewriter.eraseOp(op);
```

lib/Optimizer/Transforms/StatePreparation.cpp
schweitzpgi: Was it actually getting through? That's nothing short of amazing. heh

lib/Optimizer/Transforms/StatePreparation.cpp
schweitzpgi: Letting this go for now, but let's open an issue.

We should be able to write a partial conversion that will necessarily eliminate all instances of whatever operations we tell it we want eliminated. It will remove the boilerplate, apply simplifications, etc. all under the hood for us.

schweitzpgi (approved):
There may be some things to look at in a cleanup sort of way and it looks like this PR fell in the mud pit...

But thanks for all the hard work and heavy lifting on this one, Anna. It's going to be great.

boschmitt:
Before looking deeper in the PR I tried running the following code:
```c++
#include <cudaq.h>
#include <iostream>

__qpu__ void test(std::vector<cudaq::complex> inState) {
  cudaq::qvector q1 = inState;
}

void printCounts(cudaq::sample_result &result) {
  std::vector<std::string> values{};
  for (auto &&[bits, counts] : result) {
    values.push_back(bits);
  }

  std::sort(values.begin(), values.end());
  for (auto &&bits : values) {
    // Reverse the bits so that the output is a binary number.
    std::reverse(bits.begin(), bits.end());
    std::cout << bits << '\n';
  }
}

void printState(const std::vector<cudaq::complex> &amplitudes) {
  std::cout << "state: (size: " << amplitudes.size() << ") { ";
  for (auto a : amplitudes)
    std::cout << a << ' ';
  std::cout << "}\n";
}

int main() {
  constexpr auto kNUM_QUBITS = 5u;
  for (auto n = 0u; n < kNUM_QUBITS; ++n) {
    const auto dimension = (1ULL << (n + 1));
    for (auto i = 0u; i < dimension; ++i) {
      std::vector<cudaq::complex> amplitudes(dimension, 0.);
      amplitudes[i] = 1.;
      printState(amplitudes);
      auto counts = cudaq::sample(test, amplitudes);
      printCounts(counts);
      std::cout << '\n';
    }
  }
}
```

This is basically initializing trivial states in order to check correct bit ordering. I then compiled with: `~/.cudaq/bin/nvq++ --enable-mlir --target quantinuum --emulate`.

I got the following result:
```
state: (size: 2) { (1,0) (0,0) }
0

state: (size: 2) { (0,0) (1,0) }
1

state: (size: 4) { (1,0) (0,0) (0,0) (0,0) }
00

state: (size: 4) { (0,0) (1,0) (0,0) (0,0) }
01

state: (size: 4) { (0,0) (0,0) (1,0) (0,0) }
10

state: (size: 4) { (0,0) (0,0) (0,0) (1,0) }
11

state: (size: 8) { (1,0) (0,0) (0,0) (0,0) (0,0) (0,0) (0,0) (0,0) }
loc("state_preparation_vector.cpp":5:23): error: 'quake.init_state' op initialize state vector must be power of 2, but is 3 instead.
fish: Job 1, './a.out' terminated by signal SIGABRT (Abort)
```

lib/Optimizer/Transforms/ConstPropComplex.cpp
annagrin: I have tests in quake for this, AFAIK we don't generate this in code yet. But it will help after https://github.com/NVIDIA/cuda-quantum/issues/1955 is addressed.

But I think I am missing a safety check for the operand's type here. 

Added another test (I only had a test for one direction of the cast before). I remember seeing problems with those functions too - I think `valCon.value().convertToFloat()` fails if the value is double, I added an extra check to ensure that it is indeed a float.

annagrin:
> ```c++
> #include <cudaq.h>
> #include <iostream>
> 
> __qpu__ void test(std::vector<cudaq::complex> inState) {
>   cudaq::qvector q1 = inState;
> }
> 
> void printCounts(cudaq::sample_result &result) {
>   std::vector<std::string> values{};
>   for (auto &&[bits, counts] : result) {
>     values.push_back(bits);
>   }
> 
>   std::sort(values.begin(), values.end());
>   for (auto &&bits : values) {
>     // Reverse the bits so that the output is a binary number.
>     std::reverse(bits.begin(), bits.end());
>     std::cout << bits << '\n';
>   }
> }
> 
> void printState(const std::vector<cudaq::complex> &amplitudes) {
>   std::cout << "state: (size: " << amplitudes.size() << ") { ";
>   for (auto a : amplitudes)
>     std::cout << a << ' ';
>   std::cout << "}\n";
> }
> 
> int main() {
>   constexpr auto kNUM_QUBITS = 5u;
>   for (auto n = 0u; n < kNUM_QUBITS; ++n) {
>     const auto dimension = (1ULL << (n + 1));
>     for (auto i = 0u; i < dimension; ++i) {
>       std::vector<cudaq::complex> amplitudes(dimension, 0.);
>       amplitudes[i] = 1.;
>       printState(amplitudes);
>       auto counts = cudaq::sample(test, amplitudes);
>       printCounts(counts);
>       std::cout << '\n';
>     }
>   }
> }
> ```

Wow, looks like bug! I suspect this check is incorrect in https://github.com/NVIDIA/cuda-quantum/blob/80bfb35872811045630405ff19c2ac523e0d2e39/lib/Optimizer/Dialect/Quake/QuakeOps.cpp#L515, because the veq length is the number of qubits :)

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,440 @@
...
+    // Decompose the state into phases and magnitudes.
+    bool needsPhaseEqualization = false;
+    std::vector<double> phases;
+    std::vector<double> magnitudes;
+    for (const auto &a : amplitudes) {
+      phases.push_back(std::arg(a));
+      magnitudes.push_back(std::abs(a));
+      // FIXME: remove magic number.
+      needsPhaseEqualization |= std::abs(phases.back()) > 1e-10;
```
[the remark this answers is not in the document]
annagrin: will do, thanks!

annagrin:
> I focused the review in the state preparation algorithm. It looks good to me. I just left a couple of smaller comments. Great work!
> 
> It might be good to add at least one python test that uses a bigger state vector.

done, thanks!

lib/Optimizer/Transforms/StatePreparation.cpp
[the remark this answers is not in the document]
annagrin: done

lib/Optimizer/Transforms/StateDecomposer.cpp
annagrin: Thanks, @boschmitt is currently reviewing the changes

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -568,6 +576,55 @@ def ObserveAnsatz : Pass<"observe-ansatz", "mlir::func::FuncOp"> {
...
+        quake.x [%1] %2 : (!quake.ref, !quake.ref) -> ()
+        return
+      }
+    }
+    ```
+  }];
+
+  let options = [
+    Option<"phaseThreshold", "threshold", "double",
+      /*default=*/"1e-10", "Equalize the state if larger than the threshold">,
```
boschmitt, line 624:
```suggestion
      /*default=*/"1e-10", "Threshold to trigger phase equalization">,
```

boschmitt (approved):

khalatepradnya:
Great work, @annagrin ! 👏🏼

PR #1897, opened 2024-07-08
Improving the timeout error message

sacpis:
fetching the default WATCHDOG_TIMEOUT_SEC and improving the timeout error message

Error message looks like this

root@7837e016f2c7:/workspaces/cuda-quantum# python3 examples/python/advanced_vqe.py 
RuntimeError: Failed to launch kernel. Error: Failed to execute the kernel on the remote server: "json_request_runner.py returned an error"
Error message: "Timeout occurred during execution."

Note: Watchdog timer is working as expected. We just need to set the environment variable WATCHDOG_TIMEOUT_SEC on the server.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCOps.td
+++ b/include/cudaq/Optimizer/Dialect/CC/CCOps.td
@@ -1066,10 +1066,10 @@ def cc_ComputePtrOp : CCOp<"compute_ptr", [Pure]> {
 def cc_SizeOfOp : CCOp<"sizeof", [Pure]> {
   let summary = "Return the size, in bytes, of a type.";
   let description = [{
-    This operation is the size, in bytes, of a type that can be reified. It can
-    be used to determine how many bytes to allocate or move when a value is in
-    memory, for example. If the type cannot be reified, it returns a size of
-    poison.
+    This operation returns the size, in bytes, of a type that can be reified.
+    It can be used to determine how many bytes to allocate or move when a value
+    is in memory, for example. If the type cannot be reified, it returns a size
+    of poison.
 
     Example:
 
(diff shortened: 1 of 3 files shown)


tools/cudaq-qpud/nvqc_proxy.py
```diff
@@ -53,7 +53,9 @@ def build_command_list(temp_file_name: str) -> list[str]:
     if WATCHDOG_TIMEOUT_SEC > 0:
         cmd_list = ['timeout', str(WATCHDOG_TIMEOUT_SEC)] + cmd_list
     if RUN_AS_NOBODY:
-        cmd_list = ['su', '-s', '/bin/bash', 'nobody', '-c', ' '.join(cmd_list)]
+        cmd_list = [
+            'su', '-s', '/bin/bash', 'nobody', '-c', ' '.join(cmd_list)
+        ]
```
bmhowe23, line 58: I think the formatter isn't happy with this change because I think you need to run `yapf -i --style google tools/cudaq-qpud/nvqc_proxy.py`.

tools/cudaq-qpud/nvqc_proxy.py
```diff
@@ -256,7 +262,8 @@ def do_POST(self):
         NUM_GPUS = 0
     MPI_FOUND = (shutil.which('mpiexec') != None)
     SUDO_FOUND = (shutil.which('sudo') != None)
-    WATCHDOG_TIMEOUT_SEC = int(os.environ.get('WATCHDOG_TIMEOUT_SEC', 0))
+    WATCHDOG_TIMEOUT_SEC = int(
+        os.environ.get('WATCHDOG_TIMEOUT_SEC', WATCHDOG_TIMEOUT_SEC))
```
bmhowe23, line 266: I'm curious if this change was required or not. Wasn't WATCHDOG_TIMEOUT_SEC initialized to 0 anyway? If so, it seems like it would be functionally the same.

bmhowe23 (approved):
LGTM....thanks!

I would recommend reducing the title of the PR (and the resulting commit message) down to ~50 characters if possible. (It's not a strict requirement but the currently length of 108 characters is probably a little too long.) Given that the WATCHDOG_TIMEOUT_SEC change wasn't strictly required, I'd recommend omitting that part from the title. You could also shorten QuantumCloud down to NVQC.

https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

Issue #1898, opened 2024-07-08
the plot the sample testing images in the hybrid nn doesn't work when run on gpu


mmvandieren:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

if the `device` is `cuda:0` instead of a `cpu` then the notebook gives an error when it tries to plot the sample testing data

### Steps to reproduce the bug

run hybrid_qnns notebook on a gpu

### Expected behavior

plot the images

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: gory

### Suggestions

check for device and move data to cpu if necessary

PR #1899, opened 2024-07-08
fixed bug when notebook was run on a GPU

mmvandieren:
sample images were not plotted and gave an error message when on GPU.  Added a few lines to check the device and move the sample to be plotted back to the cpu if necessary.

### Description

Fixes https://github.com/NVIDIA/cuda-quantum/issues/1898


bmhowe23 (approved):
LGTM. Thanks @mmvandieren.

PR #1903, opened 2024-07-09
Refactor github.actor to replace 'fixed' strings and add context checks of buildx for sequential compilation and partial failures scenario

Quleaf:
### Description

1. **Refactor github.actor:**
Replace 'fixed' strings with the more flexible github.actor. This change ensures that workflows accommodate any general actor who forks the repository and utilizes GitHub Actions, rather than being limited to fixed string values.

2. **Context Checks for Buildx:**
Implement context checks in the buildx step, especially when using a single linux-xxx-cpu8 for parallel workflows. This ensures the build context is correctly validated, addressing potential issues in sequential compilation and scenarios where partial failures might occur.


.github/workflows/ci.yml
```diff
@@ -36,17 +36,26 @@ jobs:
       llvm_commit: ${{ steps.repo_info.outputs.llvm_commit }}
       pybind11_commit: ${{ steps.repo_info.outputs.pybind11_commit }}
       platform_config: ${{ steps.config.outputs.platforms }}
+      actor_lower: ${{ steps.set_actor_lower.outputs.actor_lower }}
```
bettinaheim, line 39: rather than the lower case of the actor (which is incorrect in the case of the cudaq repository), I believe what you actually want is to just use ${{ github.repository_owner }}.

.github/workflows/ci.yml
```diff
@@ -36,17 +36,26 @@ jobs:
...
+          echo "ACTOR_LOWER=$actor_lower" >> $GITHUB_ENV
+          echo "actor_lower=$actor_lower" >> $GITHUB_OUTPUT 
+      
       - id: config
         run: |
           platforms="{}"
           for platform_id in amd64 arm64; do
-            if [ "$platform_id" == "amd64" ]; then minimal_base_image=ghcr.io/nvidia/amd64/almalinux:8
-            elif [ "$platform_id" == "arm64" ]; then minimal_base_image=ghcr.io/nvidia/arm64v8/almalinux:8
+            if [ "$platform_id" == "amd64" ]; then minimal_base_image=ghcr.io/${ACTOR_LOWER}/amd64/almalinux:8
```
bettinaheim, line 57:
```suggestion
            if [ "$platform_id" == "amd64" ]; then minimal_base_image=ghcr.io/${{ github.repository_owner }}/amd64/almalinux:8
```

bettinaheim (changes requested):

PR #1905, opened 2024-07-09
Fix bug in `memtoreg` pass with classical values in `if`s

atgeller:
### Description
There was a small bug in the `memtoreg` pass where non-linear classical values would incorrectly result in additional arguments being added to the `then` and `else` blocks of `if` statements due to a missing check. This PR adds the missing check, as well as a test case based on the below example, to address this bug.

#### Example:
```
cudaq::qubit q;
int i = 0;

if (true) {
  x(q);
  i = 1;
}
 ```


schweitzpgi (approved):
LGTM. Thanks for finding and fixing this, Adam.

PR #1910, opened 2024-07-10
Fix #17: don't generate loops in the kernel_builder for measurements

### LINKED ISSUE #17: Update kernel_builder measurement functions to leverage quake.mz(qvec)
Do this instead of looping over quake.qextract calls.

schweitzpgi:
This simplifies kernel_builder and leverages that Quake already knows what to do with veq types.

Fix reset while we are here. There is no need to test the types of these arguments. The verifier already does that.

Fix std::string arguments to not make copies of the string at each call site.


annagrin (approved):
LGTM!

PR #1912, opened 2024-07-10
[kernel_builder] Use quake.reset with a veq argument.

schweitzpgi:
Simplify the kernel_builder to use both qubit and veq forms of the reset op. Add a pattern to expand measurements that also expands resets.

~Requires #1910 to be merged first.~


amccaskey (approved):
nice simplification

bmhowe23 (approved):
👍

PR #1913, opened 2024-07-10
[NFC] Add tools/cudaq-qpud/README.md

bmhowe23:
Add some developer docs for `remote-mqpu` and `nvqc`.

### DIFF
--- a/tools/cudaq-qpud/README.md
+++ b/tools/cudaq-qpud/README.md
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable MD013 -->
 # `remote-mqpu` / `nvqc` Debugging Tips
 
 This file contains tips and tricks for when you are performing manual testing/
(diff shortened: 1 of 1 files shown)


tools/cudaq-qpud/README.md
```diff
@@ -0,0 +1,92 @@
...
+   correct IP address obtained in step 3:
+   - If you are using Python, change your `cudaq.set_target` line to be
+   something like this: `cudaq.set_target('remote-mqpu', url='172.31.123.45:3030')`.
+   - If you are using C++, change your `nvq++` command to something like this:
+   `nvq++ --target remote-mqpu --remote-mqpu-url 172.31.123.45:3030`.
+
+## Running your own image on the NVQC server
+
+1. After building your `nvcr.io/pnyjrcojiblh/cuda-quantum/cuda-quantum:custom`
+   image, you can `docker push` it (assuming you have authorized credentials).
```
khalatepradnya, line 53: Is this section supposed to be internal-only?

khalatepradnya (approved):
Thanks, Ben! This is helpful.

tools/cudaq-qpud/README.md
bmhowe23: > Is this section supposed to be internal-only?

The `pnyjrcojiblh` identifier is also in our `.github/workflows/integration_tests.yml` workflow, and it certainly isn't a secret. However, I can add a sentence or two explaining this clarifying tag and providing an alternative if people want to name it something else.

khalatepradnya: I see. I didn't mean that it was confidential, but it may not be useful to anyone external?

bmhowe23: > I see. I didn't mean that it was confidential, but it may not be useful to anyone external?

Yeah, I think this whole README.md file would be of limited utility to anyone external (because hopefully they don't need to debug issues like this), but IMO, it seems more convenient to place the docs near the code anyway.

bmhowe23: Clarified in 106b8e9e1

PR #1914, opened 2024-07-10
Reading the temp_file for the subprocess result

sacpis:
Checked with adding print() function in the kernel.

```
RuntimeError: Failed to launch kernel. Error: Failed to execute the kernel on the remote server: "Failed to process incoming request"
Error message: "advanced_vqe.py:20: error: unhandled function call - print, known kernels are dict_keys([])\n\t (offending source -> print())"
```

Note: Will handle timeout exception in a different PR.

### DIFF
--- a/docker/release/cudaq.nvqc.Dockerfile
+++ b/docker/release/cudaq.nvqc.Dockerfile
@@ -16,29 +16,14 @@
 ARG base_image=nvcr.io/nvidia/nightly/cuda-quantum:latest
 FROM $base_image as nvcf_image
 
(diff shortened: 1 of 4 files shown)


bmhowe23:
> Note: Will handle timeout exception in a different PR.

Timeout is something we had working before. Is there a reason we can't retain that functionality here? I suspect it would only be a little bit more effort to retain that functionality here. Let me know if I'm missing something, though.

bmhowe23 (approved):
👍

PR #1915, opened 2024-07-10
[NFCI] Fix compiler warning: `[-Wdefaulted-function-deleted]`

1tnguyen:
### Description

This should be a deleted function as per the compiler.


1tnguyen:
Fixed by https://github.com/NVIDIA/cuda-quantum/pull/1603

PR #1916, opened 2024-07-11
Add cudaq.translate API to docs

annagrin:
### Description
Add `cudaq.translate` API to docs

image


python/runtime/cudaq/algorithms/py_translate.cpp
```diff
@@ -75,10 +75,11 @@ path, i.e., the trace, of the provided `kernel`.
 
 Args:
   format (str): format to translate to. Available formats: `qir`, `qir-base`,
-    `qir-adaptive`, `openqasm`.
+    `qir-adaptive`, `openqasm2`.
   kernel (:class:`Kernel`): The :class:`Kernel` to translate.
   *arguments (Optional[Any]): The concrete values to evaluate the kernel
     function at. Leave empty if the kernel doesn't accept any arguments.
+  Note: Translating functions with arguments to OpenQasm 2.0 is not supported.
```
khalatepradnya, line 82: I've seen this stylized as `OpenQASM 2.0`, we should probably use the same?

annagrin: Done, thanks!

khalatepradnya (approved):
LGTM! Thanks, Anna.

Issue #1927, opened 2024-07-13
[python] Array capture in kernels does not follow pythons semantics


annagrin:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Arrays and lists captured by kernels don't change after the first compilation, so sampling of the kernel generates the same results even if the array is changed.

### Steps to reproduce the bug

```python

import cudaq
import numpy as np

def swap(arr, index1, index2):
    t = arr[index2]
    arr[index2] = arr[index1]
    arr[index1] = t

cudaq.reset_target()
cudaq.set_target('nvidia')

c = [1. / np.sqrt(2.) + 0j, 0., 0., 1. / np.sqrt(2.)]

@cudaq.kernel
def kernel():
    q = cudaq.qvector(np.array(c))

counts = cudaq.sample(kernel)
print(counts)
assert '11' in counts
assert '00' in counts

swap(c, 1, 3)

counts = cudaq.sample(kernel)
print(counts)

# The following assert fails
assert '10' in counts
assert '00' in counts
```

### Expected behavior

Asserts do not fail

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #1928, opened 2024-07-13
Fix array and list incorrect capture in python

annagrin:
### Description
Fix array and list incorrect capture in python

Details:
Recompile kernels at every execution if they capture arrays or lists.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1927


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -184,7 +184,11 @@ def compile(self):
                 }
                 if self.dependentCaptures != None:
                     for k, v in self.dependentCaptures.items():
-                        if self.globalScopedVars[k] != v:
+                        if (isinstance(v, (list, np.ndarray))):
+                            # We can't detect a change in an array, always recompile.
+                            self.module = None
```
bmhowe23, line 189: If one were running kernels doing `cudaq.vqe()` or with `optimzer.optimize`, does this mean it would have to recompile every time through the optimization loop? If so, that might have a negative performance impact.

annagrin: I think so. I think the problem is that we don't save the list argument to compare to... Let me see if I can just copy it into a kernel decorator for comparison, so I can compare the list elements and skip the recompilation. The comparison would have to happen every time, which less than recompilation. I want to eventually avoid that too (https://github.com/NVIDIA/cuda-quantum/issues/1929) but for now just would like to fix the bug. Will update.

annagrin: Done!

bmhowe23 (approved):
LGTM. Thanks, Anna.

Issue #1852, opened 2024-06-26
[workflows] The workflows should default to disabled in forks of the repository


schweitzpgi:
There should be some way to disable the workflows in people's forks of the repo by default. They may not have the correct permissions to run them and not want all the failure emails every time they update their `main` branch.

PR #1932, opened 2024-07-14
Disable workflows for a forked repository

sacpis:
For issue [#1852](https://github.com/NVIDIA/cuda-quantum/issues/1852) raised by @schweitzpgi

Test ([PR](https://github.com/sacpis/cuda-quantum/pull/15) on a forked repository)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
   metadata:
     name: Retrieve PR info
     runs-on: ubuntu-latest
-    if: ${{ ! github.event.repository.fork }}
+    if: ${{ github.event_name == 'workflow_dispatch' || ! github.event.repository.fork }}
     permissions:
       pull-requests: read
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
   metadata:
     name: Retrieve PR info
     runs-on: ubuntu-latest
+    if: ${{ ! github.event.repository.fork }}
     permissions:
       pull-requests: read
 
--- a/.github/workflows/cla_assistant.yml
+++ b/.github/workflows/cla_assistant.yml
@@ -9,6 +9,7 @@ name: "CLA bot"
 jobs:
   cla-acknowledgement:
     runs-on: ubuntu-latest
+    if: ${{ ! github.event.repository.fork }}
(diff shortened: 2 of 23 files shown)


bmhowe23:
> > > Also, I know I sometimes run the integration_tests.yml from my account, and I'd like to be able to continue to do that as
> > > needed. Is there any way to only disable the scheduled runs in forks and retain the ability to run them on demand?
> 
> Yes, there is a way to run any tests in the local PR by removing the condition for forked repo.

Maybe I'm in the minority here, but I'd still like to be able to manually run the nightly integration tests from my repo (screenshot below) without having to make my main branch diverge from the upstream main branch by committing a change to revert integration_tests.yml back to how it is right now.

E.g.,
image

Would it be possible to retain this capability by adding structuring the `if` check like this?
```
if: ${{ github.event_name == 'workflow_dispatch' || ! github.event.repository.fork }})
```

I believe this would only need to be done for the ones that have this:
```
on:
  workflow_dispatch:
```

We should also get @bettinaheim to weigh in here.

schweitzpgi:
Does this check prevent manually running the yml file?

The original issue was logged because these scripts are "on by default". So if you have a fork, you have to go through them all one by one and disable them, or set up email filters to send all the failures to your deleted folder or both. For a fork, these scripts should all be "off by default" since many/most won't have access to the runners, etc. anyway.

bmhowe23:
Perhaps relevant: https://github.com/orgs/community/discussions/26704#discussioncomment-5268498

That thread makes it look like users have to manually accept workflows for new forks, at least as of March 2023.

bmhowe23:
FWIW, I picked 3 random forks that were opened in the last few months, and they all contain 0 Actions runs:
* https://github.com/InderpalSuthar/cuda-quantum/actions
* https://github.com/prateekchawla168/cuda-quantum/actions
* https://github.com/mletras89/cuda-quantum/actions

So maybe GitHub's changes supersede the need for this change?

bmhowe23:
> @bmhowe23 Should I proceed with the changes as per the suggestions made here?

Perhaps we should discuss at the team meeting. I think that the original issue (#1852) isn't relevant anymore due to GitHub's changes mentioned above. For example, the 3 freshly forked repos I mentioned above have 0 Actions runs, so it seems like the underlying issue won't appear for any new forks.

schweitzpgi:
So the solution is for original authors to create new forks?

PR #1933, opened 2024-07-14
Fix custom operations on remote-mqpu (library mode)

### LINKED ISSUE #1922: [C++] [library mode] Custom operations not working on 'remote' target
### Required prerequisites

### Describe the bug

* Attempting to use custom operation in C++ library mode with remote target fails.

Error:
``` Caught signal 11 (Segmentation fault: address not mapped to object at address 0x7efe6150eed8)```

```
nvq++ --target remote-mqpu --remote-mqpu-auto-launch 1 custom_op.cpp && ./aout
```

custom_op.cpp:

```
#include <cudaq.h>

CUDAQ_REGISTER_OPERATION(custom_h, 1, 0,
                         {M_SQRT1_2, M_SQRT1_2, M_SQRT1_2, -M_SQRT1_2})

__qpu__ void bell_pair() {
  cudaq::qubit q, r;
  custom_h(q);
  x<cudaq::ctrl>(q, r);
}

int main() {
  auto counts = cudaq::sample(bell_pair);
  for (auto &[bits, count] : counts) {
    printf("%s\n", bits.data());
  }
}
```

### Expected behavior

Expected output: Program runs successfully.

```
11
00
```

Not a regression

### Environment

- **CUDA Quantum version**: `main` branch as of f09014dc8715bf0ba1b8afca4b56302e0eab9ad1
- **Python version**: N/A
- **C++ compiler**: 

### Suggestions

Potentially missing section in payload.

bmhowe23:
Fixes #1922 

The problem was that the LLVM JIT objects were being destroyed right at the end of `invokeWrappedKernel()`, and the ExecutionManager's `registeredOperations` contained there were dangling pointers into object code created during that JIT (specifically the vtables for the custom operations' structs). The fix is to make the lifetime of the LLVM JIT last longer than it previously did.

Also, we have to provide a mechanism to clear the registered operations so that shutdown of `cudaq-qpud` doesn't attempt to call destructors in object code that was created during a JIT (again, vtables).

### DIFF
--- a/runtime/common/JIT.cpp
+++ b/runtime/common/JIT.cpp
@@ -31,11 +31,10 @@
 #define DEBUG_TYPE "cudaq-qpud"
 
 namespace cudaq {
-
-void invokeWrappedKernel(std::string_view irString,
-                         const std::string &entryPointFn, void *args,
-                         std::uint64_t argsSize, std::size_t numTimes,
-                         std::function<void(std::size_t)> postExecCallback) {
+std::unique_ptr<llvm::orc::LLJIT>
+invokeWrappedKernel(std::string_view irString, const std::string &entryPointFn,
+                    void *args, std::uint64_t argsSize, std::size_t numTimes,
+                    std::function<void(std::size_t)> postExecCallback) {
 
   std::unique_ptr<llvm::LLVMContext> ctx(new llvm::LLVMContext);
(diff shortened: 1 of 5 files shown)


1tnguyen (approved):
Thanks, Ben!

Issue #1860, opened 2024-06-27
Build dependency missing for python?


schweitzpgi:
When doing iterative development on files in `python/` one will notice that source files may be copied to their working `build/python/` tree. These are a second copy of the files from the source tree, `cuda-quantum/python/`. Making edits to the files in the source tree may not update the files in the build tree when running `make` (or `ninja`) again. This can lead to confusing results as, for example, debug print statements remain only in the source tree and are thus not included in the build tree code, which may be what is executed.

I suspect that there may be missing boilerplate regarding copying the `.py` files in our cmake recipes.

schweitzpgi:
Verified.

PR #1935, opened 2024-07-15
Adding a new custom target to copy python files

sacpis:
For issue #1860 raised by @schweitzpgi

Adding a new custom target to copy python files from python/ to python/build/ on running make or ninja.

Tested with create example.py file in python/ and then making modifications to it. Checked that modified example.py file is getting copied in the build/python directory after running make and ninja.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1673,6 +1673,25 @@ class PyASTBridge(ast.NodeVisitor):
                     is_adj=False)
                 return
 
+            # Handle the case where we are capturing an opaque kernel
+            # function. It has to be in the capture vars and it has to
+            # be a PyKernelDecorator.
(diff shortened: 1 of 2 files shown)


python/CMakeLists.txt
```diff
@@ -132,6 +132,16 @@ add_mlir_python_common_capi_library(CUDAQuantumPythonCAPI
 # Instantiation of all Python modules
 ################################################################################
 
+file(GLOB_RECURSE PYTHON_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/cudaq/*.py")
```
bmhowe23, line 135: In addition to the python/cudaq directory, we also have the python/{runtime,tests,utils} directories. Should those be marked as dependencies and copied over as well? Or are they covered by something else already?

bmhowe23 (approved):
Thanks, Sachin. LGTM.

Issue #1804, opened 2024-06-12
[python] Combining `cudaq` module with `multiprocesssing` module can result in deadlock


bmhowe23:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Some applications need to use the `multiprocessing.Process` function (or the `ProcessPoolExecutor.run_in_executor` function) to spawn child processes to perform work. By default, Python on Linux creates the child process using the "fork" method (as described in the [Python docs](https://docs.python.org/3/library/multiprocessing.html)). The docs say the following (emphasis mine):

> The parent process uses [os.fork()](https://docs.python.org/3/library/os.html#os.fork) to fork the Python interpreter. The child process, when it begins, is effectively identical to the parent process. All resources of the parent are inherited by the child process. **Note that safely forking a multithreaded process is problematic.**

cudaq is indeed a multithreaded process, so as it stands right now, this is problematic.

Also note that according to the docs, Python 3.14 will change this behavior.

> The default start method will change away from fork in Python 3.14. Code that requires fork should explicitly specify that via [get_context()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) or [set_start_method()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method).

### Steps to reproduce the bug

The following example demonstrates the problem.

```python
import cudaq
import multiprocessing

def myFunc():
    cudaq.set_target('qpp-cpu')

if __name__ == '__main__':
    # Uncommenting either *ONE* of the following lines "fixes" this problematic
    # example.
    # multiprocessing.set_start_method('spawn')
    # multiprocessing.set_start_method('forkserver')
    
    cudaq.set_target('qpp-cpu')
    print('Spawning another process to run myFunc')
    process = multiprocessing.Process(target=myFunc)
    process.start()
    print('Joining...')
    process.join()
    print('DONE!') # never gets here unless set_start_method is called
```
[...]

PR #1936, opened 2024-07-15
Setting multiprocessing start method to forkserver

sacpis:
For issue #1804 raised by @bmhowe23 

Example used to test this (Thanks @bmhowe23 for providing the same in the issue)
```
import cudaq
import multiprocessing

def myFunc():
    cudaq.set_target('qpp-cpu')

if __name__ == '__main__':
    # Uncommenting either *ONE* of the following lines "fixes" this problematic
    # example.
    # multiprocessing.set_start_method('spawn')
    # multiprocessing.set_start_method('forkserver')
    
    cudaq.set_target('qpp-cpu')
    print('Spawning another process to run myFunc')
    process = multiprocessing.Process(target=myFunc)
    process.start()
    print('Joining...')
    process.join()
    print('DONE!') # never gets here unless set_start_method is called
```

# Test for test.py file

Old behavior

root@d3f2e548b5a6:/workspaces/cuda-quantum# python3 examples/python/test.py 
Spawning another process to run myFunc
Joining...

New behavior

root@d3f2e548b5a6:/workspaces/cuda-quantum# python3 examples/python/test.py 
Spawning another process to run myFunc
Joining...
DONE!

# Test for test.ipynb file

- If we use the entire code in a single file (even in different code cells), the child process throws AttributeError, as it is unable to find the function.
- To handle multiprocessing in Python notebook, the function used should be defined in an external module. This makes the function importable in the main .ipynb file. The child process (created using spawn method) can then find the function.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -192,7 +192,7 @@ jobs:
           create_function_result=$(ngc-cli/ngc cloud-function function create \
             --container-image nvcr.io/${{ env.NGC_QUANTUM_ORG }}/${{ env.NGC_QUANTUM_TEAM }}/cuda-quantum:nightly \
             --container-environment-variable NUM_GPUS:1 \
-            --container-environment-variable NVQC_REST_PAYLOAD_VERSION:1 \
+            --container-environment-variable NVQC_REST_PAYLOAD_VERSION:1.1 \
             --api-body-format CUSTOM \
             --inference-port 3030 \
             --health-uri / \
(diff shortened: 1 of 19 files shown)


bmhowe23:
Any ideas why this CI test hung: https://github.com/NVIDIA/cuda-quantum/actions/runs/9948953673/job/27512188192?pr=1936

Is that just a fluke?

bmhowe23 (approved):
LGTM. Thanks @sacpis

Issue #1921, opened 2024-07-12
[folding] Strange behavior in MLIR folding


schweitzpgi:
If we have this code
```mlir
func.func @f() -> f32 {
  %0 = arith.constant 1.0 : f64
  %1 = arith.truncf %0 : f64 to f32
  return %1 : f32
}
```
And run it through MLIR's canonicalizer, we get back
```mlir
  func.func @f() -> f32 {
    %cst = arith.constant 1.000000e+00 : f32
    return %cst : f32
  }
```

But if we have this code
```mlir
func.func @f() {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 0.70710678118654757 : f64
    %0 = arith.truncf %cst_0 : f64 to f32
    %1 = complex.create %0, %cst : complex<f32>
    %2 = arith.truncf %cst_0 : f64 to f32
    %3 = complex.create %2, %cst : complex<f32>
    %4 = complex.create %cst, %cst : complex<f32>
    %5 = complex.create %cst, %cst : complex<f32>
    %6 = cc.alloca !cc.array<complex<f32> x 4>
    %7 = cc.cast %6 : (!cc.ptr<!cc.array<complex<f32> x 4>>) -> !cc.ptr<complex<f32>>
    cc.store %1, %7 : !cc.ptr<complex<f32>>
    return
}
```
the canonicalizer will return
```mlir
  func.func @f() {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 0.70710678118654757 : f64
    %0 = arith.truncf %cst_0 : f64 to f32
    %1 = complex.create %0, %cst : complex<f32>
    %2 = cc.alloca !cc.array<complex<f32> x 4>
    %3 = cc.cast %2 : (!cc.ptr<!cc.array<complex<f32> x 4>>) -> !cc.ptr<complex<f32>>
    cc.store %1, %3 : !cc.ptr<complex<f32>>
    return
  }
```
and the double floating point constants are *not* folded.

Related to this issue is teaching the CC dialect's cast operation to do (more) constant folding and teaching it to understand the complex type.

annagrin:
Related: https://github.com/NVIDIA/cuda-quantum/issues/1955

schweitzpgi:
The canonicalizer does not fold this `arith.truncf`, which then inhibits the constant folding of the `complex.create` because of a loss of precision.

PR #1938, opened 2024-07-16
[core] Add constant folding of cc.cast operations.

schweitzpgi:
These constant folding operations will eliminate cc.cast operations when they are converting the type of a constant input op.


schweitzpgi:
See also #1921

test/Quake/cast_fold.qke
```diff
@@ -0,0 +1,122 @@
...
+// CHECK:           return %[[VAL_0]] : i64
+// CHECK:         }
+
+// CHECK-LABEL:   func.func @b9() -> i32 {
+// CHECK:           %[[VAL_0:.*]] = arith.constant 7 : i32
+// CHECK:           return %[[VAL_0]] : i32
+// CHECK:         }
+
+// On amd64, the result is 4294967289 (-7) but on arm64 it folds to 0.
+// I've verified this behavior on an arm64 machine with g++ and clang++.
```
bmhowe23, line 117: Ouch. I don't know what the precedence is here, but as a user, I'd rather it throw an error than produce platform-specific behavior. If this is really used for constant folding when all the values are concrete values, would it be possible to throw an error or simply not fold it instead?

schweitzpgi: We're folding this exactly like the C++ compilers do on those systems. That's not a bug. It's a bit strange that it's inconsistent, I agree, but this is the correct thing to do and has been verified.

bmhowe23: `clang++` also supports `-fsanitize=undefined` to notify the user if they encounter this behavior. Would that flag be honored if/when running it in Quake, too?

E.g.
```cpp
  double x = -7.11;
  unsigned y = static_cast<unsigned>(x);
  printf("x = %f   y = %u\n", x, y);
  return 0;
}
```
```bash
$ clang++ -fsanitize=undefined test.cpp
$ ./a.out
test.cpp:14:38: runtime error: -7.11 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior test.cpp:14:38 in 
x = -7.110000   y = 4294967289
```

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -187,10 +187,73 @@ void cudaq::cc::AllocaOp::getCanonicalizationPatterns(
...
+        auto width = ty.getIntOrFloatBitWidth();
+        if (getZint()) {
+          std::uint64_t v = val.convertToDouble();
+          return builder.create<arith::ConstantIntOp>(loc, v, width)
+              .getResult();
+        }
+        if (getSint()) {
+          std::int64_t v = val.convertToDouble();
+          return builder.create<arith::ConstantIntOp>(loc, v, width)
+              .getResult();
```
bmhowe23, line 243: Would there be any harm in doing this?
```suggestion
          if (auto d = val.convertToDouble(); d >= 0.0) {
            std::int64_t v = d;
            return builder.create<arith::ConstantIntOp>(loc, v, width)
                .getResult();
          }
```

bmhowe23: The reason I ask is that if the different platforms behave differently due to the underlying hardware implementations, the compilers may be stuck with that, but with constant folding, it seems like there is no benefit to being stuck with a bad cast if it can simply be detected and avoided.

schweitzpgi: Casting floating-point values to integers is rife with undefined behavior. It isn't the job of `nvq++` to protect a programmer from C++'s undefined behavior as far as I understand it. Adding error checks for undefined behavior doesn't seem like a good use of our limited resources, in my opinion.

Again, this isn't a bug. Or, if you'd rather, `nvq++` is faithfully "bug for bug compatible" in this regard.

The only other option is to drop constant folding floating-point to integral casts completely here, which seems like a mistake to some degree. On the other hand, it will mean LLVM will do the exact same constant folding later should this code go through LLVM.

test/Quake/cast_fold.qke
schweitzpgi: This isn't `constexpr`.

What you're suggesting here is that library mode, AOT compilation, and JIT compilation all work differently. That is, identical code that runs in library mode, compiles with the AOT compiler without C++ warnings, will crash/raise an error in the JIT compiler. Worse, the only reason the JIT compile would fail is because I added constant folding earlier in the pipeline.

While I appreciate that undefined behavior is something the user may not expect, it may also be something they fully expect because they've read the standard and know their platform.

bmhowe23: > What you're suggesting here is that library mode, AOT compilation, and JIT compilation all work differently.

Yes, that's a good point, but I think it's an acceptable deviation in this case because we are informing the user about their use of unsafe code. Users should not be writing code that makes ~~architecture-dependent~~ erroneous assumptions [*], and if we have the opportunity to warn them about it, we should. If we had the ability to do this in library mode, I would want us to do that, too, but I don't think the fact that we can't do it in library mode should prevent us from helping the user out in JIT compilation.

[*] I just discovered that even without `constexpr`, the results of the above test program change based on optimization level (on amd64).

| Compiler | -O0 | -O1 | -O2 | -O3 |
| --- | --- | --- | --- | --- |
| g++ | 4,294,967,289 | 0 | 0 | 0 |
| clang++ | 4,294,967,289 | 3,583,640,872 | 3,841,045,096 | 2,228,773,352 |

I think that nearly every user would prefer to be notified about that bug in their code if the compiler is able to conclusively determine that they are doing things incorrectly, and in this case, I would argue that we can conclusively determine if they are doing something wrong for the price of a single if check.

That's just my two cents.

schweitzpgi: If the programmer writes this code
```c++
   double d = ...;
   unsigned u = (unsigned)d;
```
which is standard conforming code, then we will produce correct results where this is well-defined and results 100% consistent with the C++ compiler and library mode otherwise. I disagree that heroic efforts to sanitize the floating-point values is necessary.

schweitzpgi: Some other cast situations that may do something unexpected include
```c++
int i = ...;
unsigned u = i;

char ch = ...;
short sh = ch;

double d = ....;
float f = d;

long ll = ...;
int i = ll;

char *cp = ...;
*(int*)cp = 42;
```
I don't think nvq++ should be trying to warn about the conditions and ways these might go wrong.

bmhowe23: Unless I'm mistaken, I believe all of those examples follow well-defined behaviors with the possible exception of the pointer handling. The pointer example would not be handled by this pass because the pointers are not constant values anyway. The `char` example highlights that some compilers choose signed char while others choose unsigned char by default, but for any given compiler, there *is* a default sign value, and the spec says what should happen depending on what the sign of the char is, so that results in well-defined behavior (unlike the example that started this thread).

I disagree that if checks against constant values are heroic efforts. If they were, I suspect both clang++ and g++ wouldn't have bothered with checking for this exact error condition when handling `constexpr` items, but they clearly do; it's trivially easy to do when you're working with constants.

At a minimum, it appears the comment about "On amd64, the result is 4294967289 (-7)" isn't strictly correct because the actual values depend on host platform, host compiler, and compilation optimization settings, and perhaps what the phase of the moon is 😄.

If it's more convenient, I can write an issue and circle back to this whole thing later.

bmhowe23: #1959 written as an enhancement issue.

bmhowe23, line 117:
```suggestion
// This is undefined behavior, so answers are not checked.
// See also: https://github.com/NVIDIA/cuda-quantum/issues/1959
```

PR #1939, opened 2024-07-16
Enable user-custom data structs in python kernels

amccaskey:
Enable the use of custom data class types in Python CUDA-Q kernel code 
```python
import cudaq
from dataclasses import dataclass 

@dataclass
class CustomIntAndFloatType:
    integer : int 
    floatingPoint : float 

@cudaq.kernel
def test(input : CustomIntAndFloatType):
    qubits = cudaq.qvector(input.integer)
    ry(input.floatingPoint, qubits[0])
    rx(input.floatingPoint * 2, qubits[0])
    x.ctrl(qubits[0], qubits[1])
    
instance = CustomIntAndFloatType(2, np.pi/2.)
counts = cudaq.sample(test, instance)
```


python/cudaq/kernel/ast_bridge.py
```diff
@@ -808,7 +808,8 @@ def needsStackSlot(self, type):
         """
         # FIXME add more as we need them
         return ComplexType.isinstance(type) or F64Type.isinstance(
-            type) or F32Type.isinstance(type) or IntegerType.isinstance(type)
+            type) or F32Type.isinstance(type) or IntegerType.isinstance(
+                type) or cc.StructType.isinstance(type)
```
schweitzpgi, line 812: [not an action] The way the formatter breaks lines is rather odd.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1127,6 +1128,33 @@ def visit_Attribute(self, node):
...
+                    memberName = node.attr
+                    structName = cc.StructType.getName(eleType)
+                    structIdx = None
+                    _, userType = globalRegisteredTypes[structName]
+                    for i, (k, _) in enumerate(userType.items()):
+                        if k == memberName:
+                            structIdx = i
+                    if structIdx == None:
+                        self.emitFatalError(
+                            f'Invalid struct member: {structName}.{memberName} (members={[k for k,_ in userType.items()]})'
```
schweitzpgi, line 1144: Does this happen? Can a negative test generate this error?

python/tests/kernel/mock/hello/__init__.py
```diff
@@ -0,0 +1,8 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+from .dtype import TestClass
```
schweitzpgi, line 8: nit: end-of-file warnings

python/utils/OpaqueArguments.h
```diff
@@ -185,6 +189,87 @@ inline std::string mlirTypeToString(mlir::Type ty) {
   return msg;
 }
 
+/// @brief Return the size and member variable offsets for the input struct.
+inline std::pair<std::size_t, std::vector<std::size_t>>
+getTargetLayout(func::FuncOp func, cudaq::cc::StructType structTy) {
```
schweitzpgi, line 194: We should move this function to a "common" place instead of copying it. The C++ code may need to do this as well in more places than it does at the moment.

schweitzpgi: Note: we should also start using the size of and offset of operations in the IR when appropriate and when offset of is completed.

python/utils/OpaqueArguments.h
```diff
@@ -185,6 +189,87 @@ inline std::string mlirTypeToString(mlir::Type ty) {
...
+          std::vector<double> *values = new std::vector<double>(asList.size());
+          for (std::size_t i = 0; auto &v : asList)
+            (*values)[i++] = v.cast<double>();
+
+          std::memcpy(((char *)data) + offset, values, 16);
+        };
+
+        TypeSwitch<Type, void>(ty.getElementType())
+            .Case([&](IntegerType type) {
+              if (type.getIntOrFloatBitWidth() == 1) {
```
schweitzpgi, line 248: `type == IntegerType::get(context, 1)` should also work here.

python/utils/OpaqueArguments.h
```diff
@@ -185,6 +189,87 @@ inline std::string mlirTypeToString(mlir::Type ty) {
...
+                appendVectorValue(value, data, offset, bool());
+                return;
+              }
+
+              appendVectorValue(value, data, offset, std::size_t());
+              return;
+            })
+            .Case([&](FloatType type) {
+              if (type.getIntOrFloatBitWidth() == 32) {
```
schweitzpgi, line 257: `type == builder.getF32Type()` might be a bit safer here. I think a 32 bit floating-point in MLIR is always IEEE single-precision at the moment, but it is possible that a collection of 32 bits might be use some other encoding. This already happens for half-precision formats.

schweitzpgi (commented):
LGTM.

python/cudaq/kernel/ast_bridge.py
amccaskey: Well since the kernel body is just treated as a string that is parsed to an AST, technically one could put anything in there. This is the thing I'm trying to throw an error on 
```python
class TestClass:
    i : int 

@cudaq.kernel 
def test(input : TestClass):
   localVar = input.thisMemberVarDoesNotExist
```

I added a negative test for it.

python/utils/OpaqueArguments.h
amccaskey: Will do.

amccaskey: Actually I think I'll leave this as a follow on as I'm not sure where to put it. I tried to put in the Factory.h/Factory.cpp but I ran into linking issues bc this code requires the CodeGen library.

amccaskey: #1951

python/tests/kernel/test_kernel_features.py
```diff
@@ -1617,6 +1617,102 @@ def kd():
     counts = cudaq.sample(kd)
     assert len(counts) == 2 and '0' in counts and '1' in counts 
     
+def test_custom_kernel_type():
```
khalatepradnya, line 1620: Can you add one representative test to 'remote' backend as well?

python/tests/kernel/mock/hello/dtype.py
```diff
@@ -0,0 +1,14 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+import cudaq 
+from dataclasses import dataclass
+
+@dataclass
+class TestClass:
+    i : int 
+    d : float
```
khalatepradnya, line 14: Q: Only `int` and `float` members are supported, right? What about boolean? (If already supported, perhaps add one test to cover that)

khalatepradnya:
I see this PR covers Python frontend, do we want to have same support on C++ side? (in the future)

schweitzpgi:
> I see this PR covers Python frontend, do we want to have same support on C++ side? (in the future)

We already support (some) data structures in C++. I don't think we want or need to register them as in this PR.

amccaskey:
> > I see this PR covers Python frontend, do we want to have same support on C++ side? (in the future)
> 
> We already support (some) data structures in C++. I don't think we want or need to register them as in this PR.

After your comment the other day Eric - this PR no longer requires user-registration of the custom structs.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1127,6 +1128,33 @@ def visit_Attribute(self, node):
...
+                eleType = cc.PointerType.getElementType(value.type)
+                if cc.StructType.isinstance(eleType):
+                    # Handle the case where we have a struct member extraction
+                    memberName = node.attr
+                    structName = cc.StructType.getName(eleType)
+                    structIdx = None
+                    _, userType = globalRegisteredTypes[structName]
+                    for i, (k, _) in enumerate(userType.items()):
+                        if k == memberName:
+                            structIdx = i
```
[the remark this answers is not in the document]
amccaskey: Yep. Will update

python/cudaq/kernel/utils.py
```diff
@@ -111,27 +115,28 @@ def emitFatalErrorOverride(msg):
...
-                return cc.StdvecType.get(ctx, F64Type.get())
-            if annotation.attr == 'complex128':
-                return ComplexType.get(F64Type.get())
-            if annotation.attr == 'complex64':
-                return ComplexType.get(F32Type.get())
-            if annotation.attr == 'float64':
-                return F64Type.get()
-            if annotation.attr == 'float32':
-                return F32Type.get()
+        if hasattr(annotation.value, 'id'):
```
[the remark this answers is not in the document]
amccaskey: Yep.

python/tests/kernel/mock/hello/dtype.py
amccaskey: Right now I have int, float, bool, and list supported.

python/tests/kernel/test_kernel_features.py
amccaskey: Does the C++ frontend support custom user structs for remote execution?

amccaskey:
```
QuakeSynthesizer.cpp:608
// The struct type ends up as a i64 in the thunk kernel args pointer, so
// just skip ahead. TODO: add support for struct types!
```

amccaskey: This should be a follow-on work. See issue #1987

python/tests/kernel/test_kernel_features.py
```diff
@@ -1620,7 +1620,103 @@ def kd():
...
+        ry(input.floatingPoint, qubits[0])
+        rx(input.floatingPoint * 2, qubits[0])
+        x.ctrl(qubits[0], qubits[1])
+    
+    instance = CustomIntAndFloatType(2, np.pi/2.)
+    counts = cudaq.sample(test, instance)
+    counts.dump()
+    assert len(counts) == 2 and '00' in counts and '11' in counts 
+
+    @dataclass
```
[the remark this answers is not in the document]
amccaskey: This is just a convenience. `dataclasses` is a common Python library for defining simple data structs. You annotate your class and this thing automates the addition of the constructor. You could just not use it and write your own struct `__init__(...)` method.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -13,7 +13,7 @@
 from collections import deque
 import numpy as np
 from .analysis import FindDepKernelsVisitor
-from .utils import globalAstRegistry, globalKernelRegistry, globalRegisteredOperations, nvqppPrefix, mlirTypeFromAnnotation, mlirTypeFromPyType, Color, mlirTypeToPyType
+from .utils import globalAstRegistry, globalKernelRegistry, globalRegisteredOperations, nvqppPrefix, mlirTypeFromAnnotation, mlirTypeFromPyType, Color, mlirTypeToPyType, globalRegisteredTypes
```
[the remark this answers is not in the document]
amccaskey: yes this is still necessary for the python ast bridge. we need to know what custom data types are defined outside of our compiler pipeline. the kernel decorator will use this dictionary to store custom data types from its parent scope.

python/tests/kernel/test_kernel_features.py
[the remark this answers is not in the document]
schweitzpgi: Ah. OK, that makes sense. Thanks.

python/cudaq/kernel/ast_bridge.py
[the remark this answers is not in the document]
schweitzpgi: OK. So, I think these are not really "registered" any longer, but the python bridge will stash the ones it encounters and will need in this dictionary.

Does that sound right?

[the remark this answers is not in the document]
amccaskey: That is correct.

annagrin (approved):
LGTM, thanks Alex!

amccaskey:
> LGTM, thanks Alex!

Thanks Anna. I talked to Eric yesterday and today and this still needs a bit more work (as well as the C++ PR). I hope to add these new updates tomorrow.

python/tests/mlir/quantum_type.py
```diff
@@ -0,0 +1,87 @@
...
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# RUN: PYTHONPATH=../../ pytest -rP  %s | FileCheck %s
+
+import pytest
+
+import cudaq
+
+def test_custom_quantum_type():
```
amccaskey, line 15: @schweitzpgi It would be great to get your review on this again. I've updated to treat custom classes in Python that contain quantum memory as value types (`cc.struct` is allocated with `cc.undef` and values inserted and pulled with `cc.insertvalue / cc.extractvalue`. See the test FileCheck code below. 

Note the purely classical custom classes will continue using memory semantics (`cc.alloca` + `cc.load / cc.store`).

Issue #1855, opened 2024-06-26
unclear error message when calling cudaq.adjoint outside of a kernel


mmvandieren:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

There is no (or difficult to find) documentation on the `cudaq.adjoint` operator applied to a kernel to indicate that the behaviors below might not be allowed (unless one has a solid understanding of decorators, which I don't think all of our user base will necessarily have).  The error message produced doesn't really help since it seems to indicate that this function doesn't even exist.

error message: module 'cudaq' has no attribute 'adjoint'

### Steps to reproduce the bug

```@cudaq.kernel
def quantum_fourier_transform(qubits: cudaq.qview):
    qubit_count = len(qubits)
    # Apply Hadamard gates and controlled rotation gates.
    for i in range(qubit_count):
        h(qubits[i])
        for j in range(i + 1, qubit_count):
            angle = (2 * np.pi) / (2**(j - i + 1))
            cr1(angle, [qubits[j]], qubits[i])

@cudaq.kernel
def test(q:int):
    qubits = cudaq.qvector(q)
    quantum_fourier_transform(qubits)

test_adj = cudaq.adjoint(test,3)

print(cudaq.draw(test_adj))
```

### Expected behavior

An error message that better explains the problem and doesn't mislead the user that this operation isn't defined at all.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.1
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #1942, opened 2024-07-17
Adding a function to handle attribute access dynamically

sacpis:
For issue #1855 raised by @mmvandieren 

Adding a function __getattr__ to handle attribute access dynamically

Test

```
root@d3f2e548b5a6:/workspaces/cuda-quantum# python3 examples/python/test.py 
AttributeError: 'adjoint' is not supported for CUDA-Q in Python
```


python/cudaq/__init__.py
```diff
@@ -54,6 +54,15 @@
     from .visualization.bloch_visualize import add_to_bloch_sphere
     from .visualization.bloch_visualize import show_bloch_sphere as show
 
+
+def __getattr__(name):
+    if hasattr(cudaq_runtime, name):
+        return getattr(cudaq_runtime, name)
+    raise AttributeError(f"'{name}' is not supported for CUDA-Q in Python")
```
amccaskey, line 61: The real issue is that there are functions that can only be called within cuda-q kernel code and not from host code. its not really about the function not being supported in Python.

PR #1945, opened 2024-07-17
[core] Add a canonicalization rule for quake.init_state

schweitzpgi:
There is no need to cast the state vector to unknown length, so this pattern will forward the vector and drop the use of an unused cast.


bmhowe23 (approved):
👍

PR #1952, opened 2024-07-17
[Python] [AST Bridge] Correctly distinguish measurement result and boolean variable

khalatepradnya:
* If a boolean variable is not the result of measurement then treat it as regular variable.
* Update the value of boolean if seen in outer scope.

### DIFF
--- a/python/tests/mlir/bool_var_scope.py
+++ b/python/tests/mlir/bool_var_scope.py
@@ -1,0 +1,32 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# RUN: PYTHONPATH=../../ pytest -rP  %s | FileCheck %s
+
+import cudaq
+
+
+@cudaq.kernel
+def test():
+    qubits = cudaq.qvector(2)
+    var_int = 42
+    var_bool = True
+
+
+print(test)
+
+# CHECK-LABEL:   func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint"} {
+# CHECK:           %[[VAL_0:.*]] = arith.constant true
+# CHECK:           %[[VAL_1:.*]] = arith.constant 42 : i64
+# CHECK:           %[[VAL_2:.*]] = quake.alloca !quake.veq<2>
+# CHECK:           %[[VAL_3:.*]] = cc.alloca i64
+# CHECK:           cc.store %[[VAL_1]], %[[VAL_3]] : !cc.ptr<i64>
+# CHECK:           %[[VAL_4:.*]] = cc.alloca i1
+# CHECK:           cc.store %[[VAL_0]], %[[VAL_4]] : !cc.ptr<i1>
+# CHECK:           return
+# CHECK:         }
--- a/python/tests/mlir/bool_var_scope.py
+++ b/python/tests/mlir/bool_var_scope.py
@@ -10,17 +10,20 @@
 
 import cudaq
 
+import pytest
 
-@cudaq.kernel
-def test():
-    qubits = cudaq.qvector(2)
-    var_int = 42
-    var_bool = True
+def test_var_scope():
+
+    @cudaq.kernel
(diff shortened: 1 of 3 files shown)


python/tests/mlir/bool_var_scope.py
```diff
@@ -0,0 +1,32 @@
...
+# CHECK-LABEL:   func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint"} {
+# CHECK:           %[[VAL_0:.*]] = arith.constant true
+# CHECK:           %[[VAL_1:.*]] = arith.constant 42 : i64
+# CHECK:           %[[VAL_2:.*]] = quake.alloca !quake.veq<2>
+# CHECK:           %[[VAL_3:.*]] = cc.alloca i64
+# CHECK:           cc.store %[[VAL_1]], %[[VAL_3]] : !cc.ptr<i64>
+# CHECK:           %[[VAL_4:.*]] = cc.alloca i1
+# CHECK:           cc.store %[[VAL_0]], %[[VAL_4]] : !cc.ptr<i1>
+# CHECK:           return
+# CHECK:         }
```
khalatepradnya, line 32: For reference, MLIR before this change looked like -
```
  func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint"} {
    %c42_i64 = arith.constant 42 : i64
    %1 = cc.alloca i64
    cc.store %c42_i64, %1 : !cc.ptr<i64>
    return
  }

```

bmhowe23 (approved):
LGTM

python/tests/mlir/bool_var_scope.py
```diff
@@ -0,0 +1,35 @@
...
+    print(kernel)
+
+
+# CHECK-LABEL:   func.func @__nvqpp__mlirgen__kernel() attributes {"cudaq-entrypoint"} {
+# CHECK:           %[[VAL_0:.*]] = arith.constant true
+# CHECK:           %[[VAL_1:.*]] = arith.constant 42 : i64
+# CHECK:           %[[VAL_2:.*]] = quake.alloca !quake.veq<2>
+# CHECK:           %[[VAL_3:.*]] = cc.alloca i64
+# CHECK:           cc.store %[[VAL_1]], %[[VAL_3]] : !cc.ptr<i64>
+# CHECK:           %[[VAL_4:.*]] = cc.alloca i1
```
schweitzpgi, line 32: All the variables should be inserted at the top of the entry block, so this alloca, store, alloca pattern would ought to look like alloca, alloca, store, ...

python/tests/mlir/bug_1777.py
```diff
@@ -0,0 +1,81 @@
...
+# CHECK:             %[[VAL_12:.*]] = quake.mz %[[VAL_11]] name "res" : (!quake.ref) -> !quake.measure
+# CHECK:             %[[VAL_13:.*]] = quake.discriminate %[[VAL_12]] : (!quake.measure) -> i1
+# CHECK:             cc.store %[[VAL_13]], %[[VAL_6]] : !cc.ptr<i1>
+# CHECK:             %[[VAL_14:.*]] = cc.load %[[VAL_6]] : !cc.ptr<i1>
+# CHECK:             %[[VAL_15:.*]] = arith.cmpi eq, %[[VAL_14]], %[[VAL_0]] : i1
+# CHECK:             cc.if(%[[VAL_15]]) {
+# CHECK:               %[[VAL_16:.*]] = quake.mz %[[VAL_5]] name "inner_mz" : (!quake.veq<2>) -> !cc.stdvec<!quake.measure>
+# CHECK:               %[[VAL_17:.*]] = quake.discriminate %[[VAL_16]] : (!cc.stdvec<!quake.measure>) -> !cc.stdvec<i1>
+# CHECK:               %[[VAL_18:.*]] = cc.alloca !cc.stdvec<i1>
+# CHECK:               cc.store %[[VAL_17]], %[[VAL_18]] : !cc.ptr<!cc.stdvec<i1>>
```
schweitzpgi, line 60: This code doesn't create a vector, doesn't create any storage to save a vector's data, doesn't copy the vector, does potentially create a memory leak on the stack, and in the end is a dead store to a memory location that will never be read.

python/tests/mlir/bool_var_scope.py
khalatepradnya: This seems like a larger issue, now captured in https://github.com/NVIDIA/cuda-quantum/issues/1963

PR #1953, opened 2024-07-17
[core] Fix memtoreg bug with loop body

schweitzpgi:
When the body of a loop defines a value that was live-in and live-out of that loop, the region arguments have to be added for all loop regions.

These arguments must be added to balance all the region arguments even if the argument is technically not live-in to the region because it is defined unconditionally therein.

### DIFF
--- a/test/Quake/memtoreg-7.qke
+++ b/test/Quake/memtoreg-7.qke
@@ -1,0 +1,102 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --memtoreg %s | FileCheck %s
+
+func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint", qubitMeasurementFeedback = true} {
+  %false = arith.constant false
+  %c1_i64 = arith.constant 1 : i64
+  %c0_i64 = arith.constant 0 : i64
+  %true = arith.constant true
+  %c2_i64 = arith.constant 2 : i64
+  %0 = quake.alloca !quake.veq<2>
+  %1 = cc.alloca i1
+  cc.store %true, %1 : !cc.ptr<i1>
+  %2 = cc.loop while ((%arg0 = %c0_i64) -> (i64)) {
+    %5 = arith.cmpi slt, %arg0, %c2_i64 : i64
+    cc.condition %5(%arg0 : i64)
+  } do {
+  ^bb0(%arg0: i64):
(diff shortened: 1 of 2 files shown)


test/Quake/memtoreg-7.qke
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint", qubitMeasurementFeedback = true} {
+// CHECK-DAG:       %[[VAL_0:.*]] = arith.constant false
+// CHECK-DAG:       %[[VAL_1:.*]] = arith.constant 1 : i64
+// CHECK-DAG:       %[[VAL_2:.*]] = arith.constant 0 : i64
+// CHECK-DAG:       %[[VAL_3:.*]] = arith.constant true
+// CHECK-DAG:       %[[VAL_4:.*]] = arith.constant 2 : i64
+// CHECK-DAG:       %[[VAL_5:.*]] = quake.alloca !quake.veq<2>
+// CHECK-DAG:       %[[VAL_6:.*]] = cc.undef i1
```
khalatepradnya, line 65: This test doesn't run on `main` so it's a difficult for me to gauge what the new changes are... of course it passes now.
For example, I don't understand how / why this is `undef`.

test/Quake/memtoreg-7.qke
```diff
@@ -0,0 +1,102 @@
...
+// CHECK:           %[[VAL_7:.*]]:2 = cc.loop while ((%[[VAL_8:.*]] = %[[VAL_2]], %[[VAL_9:.*]] = %[[VAL_3]]) -> (i64, i1)) {
+// CHECK:             %[[VAL_10:.*]] = arith.cmpi slt, %[[VAL_8]], %[[VAL_4]] : i64
+// CHECK:             cc.condition %[[VAL_10]](%[[VAL_8]], %[[VAL_9]] : i64, i1)
+// CHECK:           } do {
+// CHECK:           ^bb0(%[[VAL_11:.*]]: i64, %[[VAL_12:.*]]: i1):
+// CHECK:             %[[VAL_13:.*]] = quake.extract_ref %[[VAL_5]]{{\[}}%[[VAL_11]]] : (!quake.veq<2>, i64) -> !quake.ref
+// CHECK:             %[[VAL_14:.*]] = quake.unwrap %[[VAL_13]] : (!quake.ref) -> !quake.wire
```
khalatepradnya, line 72: [note to self] `quake.unwrap` and quake.wrap` got inserted here.

khalatepradnya (approved):

test/Quake/memtoreg-7.qke
schweitzpgi: That is part of mem2reg. It takes references and turns them into values.

test/Quake/memtoreg-7.qke
schweitzpgi: This is mem2reg. We turn references into values. Like a memory location is undefined until we store something to it, a value chain starts out as undefined until we compute something.

PR #1964, opened 2024-07-18
[NFC] [devdeps] Enhance the development dependencies image by pre-installing regularly used tools

khalatepradnya:
Following developer tools which come in handy during CUDA-Q development.
 - yapf: For formatting Python files
 - gdb, lldb : Debuggers

Hence, making them available in the devdeps image.


schweitzpgi (commented):
Seems like a good idea. +1

docker/build/devdeps.Dockerfile
```diff
@@ -189,12 +189,12 @@ ENV PATH="${PATH}:/usr/local/cmake-3.26/bin"
 # We must use h5py<3.11 because 3.11 doesn't include aarch64 Linux wheels.
 # https://github.com/h5py/h5py/issues/2408
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        git ninja-build file \
+        git gdb ninja-build file \
```
bmhowe23, line 192: Would it be ok to install lldb, too? Previously, I found that I get much better performance w/ lldb.

khalatepradnya: Sure. Should I keep both or just `lldb`?

bmhowe23: If you added gdb, I'm assuming that's because you use gdb, and that's certainly fine with me. So I'd be fine keeping both.

khalatepradnya: BTW do we want to install a specific version of `lldb`? I checked that in our dev image, if I simply run `apt install lldb`, then, `lldb version 14.0.0` gets installed.

bmhowe23: > BTW do we want to install a specific version of `lldb`? I checked that in our dev image, if I simply run `apt install lldb`, then, `lldb version 14.0.0` gets installed.

I'm ok with how you did it in 29a49a3e1ce025df36f72ad4e5adc066c828d2cc.

bmhowe23 (approved):
LGTM

Issue #1597, opened 2024-05-02
random_walk_qpe.cpp not returning correct phase when run through NVQC


qci-petrenko:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When running the random walk phase estimation example (random_walk_qpe.cpp) in WSL (on a Windows 11 machine - CPU simulation) using the 0.7.1 cuda-quantum image, the expected phase is returned: `Phase = 0.487390`.

If running through Nvidia's cloud service (`--target nvqc`), `Phase = 0.000000`. This is not expected. It's not clear whether this is a GPU simulation issue or something with the cloud service.

### Steps to reproduce the bug

The full terminal i/o is below:
```
cudaq@add20b7f7934:~/examples/cpp/other$ nvq++ random_walk_qpe.cpp
cudaq@add20b7f7934:~/examples/cpp/other$ ./a.out
Phase = 0.487390
cudaq@add20b7f7934:~/examples/cpp/other$ nvq++ random_walk_qpe.cpp --target nvqc
cudaq@add20b7f7934:~/examples/cpp/other$ ./a.out
[2024-05-02 00:29:36.675] Submitting jobs to NVQC service with 1 GPU(s). Max execution time: 3600 seconds (excluding queue wait time).

================ NVQC Device Info ================
GPU Device Name: "NVIDIA H100 80GB HBM3"
CUDA Driver Version / Runtime Version: 12.2 / 11.8
Total global memory (GB): 79.1
Memory Clock Rate (MHz): 2619.000
GPU Clock Rate (MHz): 1980.000
==================================================
Phase = 0.000000
```

### Expected behavior

I expect to have the local simulation and the cloud simulation return the same result.

### Is this a regression? If it is, put the last known working version (or commit) here.

Unknown.

### Environment

- **CUDA Quantum version**: 0.7.1
- **C++ compiler**: nvq++
- **Operating system**: Windows 11
- **WSL version**: 2

[...]

PR #1969, opened 2024-07-19
[NVQC] Support direct kernel invocation with a return value

1tnguyen:
### Description

For MLIR-based kernels, the kernel launch is bootstrapped; hence technically we can pack the return value back to users.

This PR enables that:

- Add a flat-buffer in `ExecutionContext` to store this data.

- Server side: introspect the entry-point kernel IR to know what type is being return. In this PR, only simple data types (float/int/bool) are supported.

- Client side: check if buffer data is being returned from the server and the kernel launch is expecting some return value (via the return offset). If so, copy the data to return to users.

- Adds tests for supported return types.

### Compatibility 

This is not a breaking change. 
When both the client and service are updated, the return value will be available. Otherwise, no return value would be available (same as the existing behavior)

- New client - old service: the service won't populate the buffer field in the return JSON. 

- Old client - new service: the buffer field in the return JSON would be dropped/ignored by the client.
 

Resolved https://github.com/NVIDIA/cuda-quantum/issues/1597


runtime/common/BaseRemoteSimulatorQPU.h
```diff
@@ -126,15 +126,38 @@ class BaseRemoteSimulatorQPU : public cudaq::QPU {
     // set up a single-shot execution context for this case.
     static thread_local cudaq::ExecutionContext defaultContext("sample",
                                                                /*shots=*/1);
+    // This is a kernel invocation outside the CUDA-Q APIs (sample/observe).
+    const bool isDirectInvocation = executionContextPtr == nullptr;
```
schweitzpgi, line 130: No worries here.
```suggestion
    const bool isDirectInvocation = !executionContextPtr;
```
I think way back we decided to not write out comparisons like `== 0` and `!= 0` in the code. I'm sure we've meandered off that style in many places. At any rate, we may want to discuss again.

runtime/common/ExecutionContext.h
```diff
@@ -100,6 +100,11 @@ class ExecutionContext {
   /// register after execution. Empty means no reordering.
   std::vector<std::size_t> reorderIdx;
 
+  /// @brief A buffer containing the return value of a kernel invocation.
+  // Note: this is only needed for invocation not able to return a
+  // sample_result.
```
schweitzpgi, line 105:
```suggestion
  /// @brief A buffer containing the return value of a kernel invocation.
  /// Note: this is only needed for invocation not able to return a
  /// `sample_result`.
```

runtime/common/ExecutionContext.h
```diff
@@ -100,6 +100,11 @@ class ExecutionContext {
...
+  std::vector<char> invocationResultBuffer;
```
schweitzpgi, line 106: I'm wondering if we'll need to think about endianness issues here. OTOH, our assumption to date is that the endianness is the same.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -549,21 +550,55 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
     if (!module)
       throw std::runtime_error("Failed to parse the input MLIR code");
     auto engine = jitMlirCode(*module, passes);
+    llvm::SmallVector<void *> returnArg;
```
schweitzpgi, line 553: I notice this is a `runtime` file. Should this be `std::vector`?

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -549,21 +550,55 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
...
+      auto returnTy = funcTy.getReturnType();
+      // These are the returned types that we support.
+      if (returnTy.isF32()) {
+        io_context.invocationResultBuffer.resize(sizeof(float));
+        returnArg.push_back(io_context.invocationResultBuffer.data());
+      } else if (returnTy.isF64()) {
+        io_context.invocationResultBuffer.resize(sizeof(double));
+        returnArg.push_back(io_context.invocationResultBuffer.data());
+      } else if (returnTy.isInteger(1)) {
+        io_context.invocationResultBuffer.resize(sizeof(bool));
```
schweitzpgi, line 567: FYI: On the device side, we always treat bools as chars in memory.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -549,21 +550,55 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
...
-        postExecCallback(i);
+    // Note: currently, we only return data from kernel on single-shot
+    // execution. Once we enable arbitrary sample return type, we can run this
+    // in a loop and return a vector of return type.
+    if (numTimes == 1 && !returnArg.empty()) {
+      simulationStart = std::chrono::high_resolution_clock::now();
+      llvm::Error error = engine->invokePacked(entryPointFunc, returnArg);
+      if (error) {
+        throw std::runtime_error("JIT invocation failed");
+      }
```
schweitzpgi, line 584:
```suggestion
      if (error) 
        throw std::runtime_error("JIT invocation failed");
```
Consistent style with next statement.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -549,21 +550,55 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
...
+        throw std::runtime_error("Failed to get entry function");
+
+      auto fn = reinterpret_cast<void (*)()>(fnPtr);
+      simulationStart = std::chrono::high_resolution_clock::now();
+      for (std::size_t i = 0; i < numTimes; ++i) {
+        // Invoke the kernel
+        fn();
+        if (postExecCallback) {
+          postExecCallback(i);
+        }
```
schweitzpgi, line 601:
```suggestion
        if (postExecCallback) 
          postExecCallback(i);
        
```

schweitzpgi (commented):
Do we have an RFC someplace for this that scopes out the work, esp. the limitations?

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -549,21 +550,54 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
...
         std::string(cudaq::runtime::cudaqGenPrefixName) + entryPointFn;
-    auto fnPtr =
-        getValueOrThrow(engine->lookup(entryPointFunc),
-                        "Failed to look up entry-point function symbol");
-    if (!fnPtr)
-      throw std::runtime_error("Failed to get entry function");
+    if (auto funcOp = module->lookupSymbol<LLVM::LLVMFuncOp>(entryPointFunc)) {
+      auto funcTy = funcOp.getFunctionType();
+      auto returnTy = funcTy.getReturnType();
+      // These are the returned types that we support.
```
bmhowe23, line 559: I think it would be worth updating https://nvidia.github.io/cuda-quantum/latest/using/backends/platform.html#supported-kernel-arguments in accordance with the supported return types listed here.

bmhowe23 (commented):
This looks like a good improvement to me, and it would be nice to be able to close the associated issue, so I'm in favor of putting this into the release.

runtime/common/ExecutionContext.h
1tnguyen: I added an endian check in https://github.com/NVIDIA/cuda-quantum/pull/1969/commits/026834c4a5e7382e0a8e7caf5750ceb19be21b96. 

Yea, our assumption about endianness is kind of loose at the moment, hence I think throwing an error is sufficient (rather than trying to handle it).

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
1tnguyen: This file used LLVM extensively, so I think using `llvm::SmallVector` is a better choice :-)

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
1tnguyen: Thanks! I added a `static_assert` to enforce this assumption.

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
1tnguyen: Added a note about supported return types in https://github.com/NVIDIA/cuda-quantum/pull/1969/commits/026834c4a5e7382e0a8e7caf5750ceb19be21b96.

bmhowe23 (approved):
This LGTM. Thanks, Thien!

PR #1970, opened 2024-07-19
Fix tutorials so they can run in publishing pipeline

bmhowe23:
Fixing failures from here: https://github.com/NVIDIA/cuda-quantum/actions/runs/10006741614/job/27662295778


schweitzpgi (approved):
+1

PR #1972, opened 2024-07-19
Update Hadamard test tutorial

ikkoham:
### Description

I'd like to suggest some points.
- Use observe primitive instead of sample. It's simpler and easy to extend the imaginary part.
- Removed diagonalization part. It's not Hadamard test itself but one of the applications of Hadamard test.
- Format and improved naming.
- Use `@` instead of np.dot.

@marwafar Feel free to comment on this PR.


marwafar:
I am fine with the modification. Looks good to me.

I recommend to keep the matrix diagonalization example (I am aware that it is one of the Hadamard test application). The goal of this application is to to show an example for the nvidia-mQpu backend. In chemistry, the quantum filter diagonalization (Hadamrd test type) can be use to build the matrix element of the Hamiltonian and the wavefunction overlap, then we use classical algorithm to diagonalize the matrix.

ikkoham:
Thank you. I'll revert the diagonalization part and modify it.

ikkoham:
Sorry. I didn't understand what you meant by diagonalizing random expectations in a matrix. It might make sense if the parameters were correlated, but that would be a generalized eigenvalue problem, which would be more complicated.

At any rate, I went back to computing the expectation values in parallel. I believe I have included what we want in this notebook.

marwafar:
@ikkoham  Notebook looks good to me.

ikkoham:
Thank you. I believe this PR is mergeable now.

docs/sphinx/examples/python/tutorials/hadamard_test.ipynb
```diff
@@ -4,11 +4,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Hadamard Test\n",
+    "# Hadamard Test and application\n",
```
bmhowe23, line 7:
```suggestion
    "# Hadamard Test and Application\n",
```

bmhowe23 (approved):
👍

PR #1973, opened 2024-07-19
[test][NFC] Move target config tests to `targettests` directory

1tnguyen:
### Description

These tests, despite not doing any execution, are 'target'-related and, hence are now relocated to the `targettests` directory.

The one that performs compilations for all targets in both library and mlir mode is a slow test.


schweitzpgi (approved):
LGTM.

My only question is whether these tests can and ought to be run even though they aren't being run. Running them would provide more test coverage obviously.

1tnguyen:
> LGTM.
> 
> My only question is whether these tests can and ought to be run even though they aren't being run. Running them would provide more test coverage obviously.

They will be run in the CI, just not under the 'test' umbrella, which we prefer to be really quick :-)

Issue #1931, opened 2024-07-13
Support cudaq::state synthesis for remote simulators


annagrin:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

We currently only support cudaq::state argument synthesis for executions that are running in the same address space as the synthesis pass, which is not true for remote simulators.

Suggestion: 
- Same-address space simulation: Synthesize state pointers as constant integers (supported)
- Remote simulation: Synthesize state pointers as constant global vectors of data
  - rewrite `quake.init_state` instructions to take either the vector or a new state created by `cudaq::state::from_data`.
 - Quantum devices and emulators: for now, add an error message (preferably during compilation) to indicate that state synthesis is not supported. (see issue https://github.com/NVIDIA/cuda-quantum/issues/1930 )

PR #1977, opened 2024-07-19
Add remote sim tests

annagrin:
### Description
Support `state*` synthesis on remote simulators.

### Details

- Add a `SimulationData` class that owns the data pointer and contains length and element size of the data.
  - avoids circular dependency between the compiler and the runtime
- Pass a function that reads the data to the QuakeSynthesizer
- Update quake synthesizer to 
  - read the data from the state
  - synthesize data vector
  - update the `quake.state_init` instruction to use the vector instead of the state
  - const prop calls to `getNumQubits` runtime call
- Add tests

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1931


annagrin:
Closing in favor of https://github.com/NVIDIA/cuda-quantum/pull/2035

PR #1978, opened 2024-07-19
[C++ bridge] Switch from using arith ops to cc.cast

schweitzpgi:
The arith dialect provides a big menu of casting/conversion operations that we can use but which have their own rules and semantics. For example, the generation of a quantum circuit can involve aggressive constant propagation. By using an operation in a dialect we control (CC), we can implement folding and combing these operations in ways that are suitable to our domain and do not rely on decisions in the MLIR code that may be too conservative or too aggressive for our needs.

[core] Add constant folding of cc.cast operations.

These constant folding operations will eliminate cc.cast operations when they are converting the type of a constant input op.

Update the tests.

Add pattern to eliminate integer casts into an integer comparison.

Add constant folding of complex.create. Apparently MLIR doesn't do this.

Add cast complex to complex and various constant foldings.

Add a new test.

Add codegen expansion pattern.


test/Quake/complex.qke
```diff
@@ -0,0 +1,46 @@
+// ========================================================================== //
```
annagrin, line 1: Do we still need const-prop-complex pass? Maybe we can just replace it with the canonicalizer pass when needed?

schweitzpgi: I left it for now, but we might be able to replace it (soon?).

annagrin (approved):
LGTM, thanks Eric!

Issue #1871, opened 2024-06-30
Calling the same quantum kernel in cudaq.control and cudaq.adjoint in the main kernel give error.


marwafar:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

If I run the code in the example 1 below with both 
```
cudaq.control(my_func,ancilla,q,theta)
cudaq.adjoint(my_func,q,theta)
```
I got this error: 
`error: 'func.call' op '__nvqpp__mlirgen__my_func.adj' does not reference a valid function`

However if I comment one of them, for example:
```
cudaq.control(my_func,ancilla,q,theta)
#cudaq.adjoint(my_func,q,theta)
```
It works.

In Example 2, I am trying first to apply a cudaq.control on a given quantum kernel and then apply control-adjoint of the quantum kernel, but it gives error. 
`error: 'func.call' op '__nvqpp__mlirgen__my_func.adj.ctrl' does not reference a valid function`

A work around it is shown in example 3. Example 3 works.

This does not make sense. Because, I have to explicitly rewrite the kernel using adj() for each applied gate instead of using the cudaq.adjoint() and then cudaq.control(). 

### Steps to reproduce the bug

Example 1:

```
import cudaq

@cudaq.kernel
def my_func(q:cudaq.qubit, theta:float):
    ry(theta,q)
    rz(theta,q)

@cudaq.kernel
def kernel(theta:float):
    ancilla=cudaq.qubit()
    q=cudaq.qubit()
    
    h(ancilla)
    cudaq.control(my_func,ancilla,q,theta)
    cudaq.adjoint(my_func,q,theta)

theta=1.5
count=cudaq.sample(kernel,theta)
print(count)
```

Example 2:

```
import cudaq

@cudaq.kernel
def my_func(q:cudaq.qubit, theta:float):
    ry(theta,q)
    rz(theta,q)

@cudaq.kernel
[...]

PR #1979, opened 2024-07-20
[Bug-fix] Support use of `control` and `adjoint` variants of a kernel

khalatepradnya:
* Handle case where both `control` and `adjoint` variants of a kernel are used.

* Fixes one of the issues captured in issue #1871

### DIFF
--- a/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
+++ b/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
@@ -68,8 +68,14 @@ private:
       ApplyVariants variant;
       auto callee = lookupCallee(apply);
       auto iter = infoMap.find(callee);
-      if (iter != infoMap.end())
+      if (iter != infoMap.end()) {
         variant = iter->second;
+        /// NOTE: As per the `DenseMap` API  in
+        /// llvm/include/llvm/ADT/DenseMap.h, an `insert` operation doesn't
+        /// update the value if key is already in the map. Hence, remove it and
+        /// insert a new entry with updated key.
+        infoMap.erase(iter);
+      }
       if (apply.getIsAdj() && !apply.getControls().empty())
         variant.needsAdjointControlVariant = true;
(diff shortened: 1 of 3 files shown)


test/Quake/control_and_adjoint.qke
```diff
@@ -0,0 +1,88 @@
+// ========================================================================== //
```
schweitzpgi, line 1: We could merge this with adjoint-1.qke as they are testing the same passes.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -68,8 +68,14 @@ struct ApplyOpAnalysis {
...
       auto callee = lookupCallee(apply);
       auto iter = infoMap.find(callee);
-      if (iter != infoMap.end())
+      if (iter != infoMap.end()) {
         variant = iter->second;
+        /// NOTE: As per the `DenseMap` API  in
+        /// llvm/include/llvm/ADT/DenseMap.h, an `insert` operation doesn't
+        /// update the value if key is already in the map. Hence, remove it and
+        /// insert a new entry with updated key.
+        infoMap.erase(iter);
```
schweitzpgi, line 77: I don't think we need to erase anything in the infoMap. Just change the insert line to use `operator[]` and update the map in place. That'll avoid freeing and allocating these structs in the map and be slightly more efficient.

khalatepradnya: Addressed in 9e6fc7e3cf41b4d1cd529384078170664446faa7

test/Quake/control_and_adjoint.qke
khalatepradnya: There's another problem in issue#1871 which I was planning to expand in this file as combinations of `control` and `adjoint`, hence prefer to keep it separate.

schweitzpgi (approved):
+1

PR #1981, opened 2024-07-20
Fix cudaq::slice_vector synthesis bug and add test

### LINKED ISSUE #1975: Synthesis fails for some kernels that use `cudaq::slice_vector`
### Required prerequisites

### Describe the bug

As a result of #1853 (merged on July 2), examples like vqe_h2.cpp can no longer be run through the full synthesizer and submitted to the current NVQC servers. The problem has nothing to do with VQE, per se, and in fact, the upcoming release of 0.8.0 should side-step this problem for VQE because synthesis will bypass the first argument of VQE kernels. The problem appears to be related to the usage of the `cudaq::slice_vector()` function, and it can be reproduced using a simple `cudaq::observe`. That is - VQE is not required to reproduce this problem.

In the latest nightly Docker container, run
```bash
$ nvq++ --target nvqc examples/cpp/algorithms/vqe_h2.cpp
$ export NVQC_API_KEY=nvapi-<key>
$ ./a.out
[2024-07-19 20:34:45.081] Submitting jobs to NVQC service with 1 GPU(s). Max execution time: 3600 seconds (excluding queue wait time).
4 qubit Hamiltonian -> 36 parameters
loc("vqe_h2.cpp":68:27): error: Unknown gep/load configuration for synthesis.
loc("vqe_h2.cpp":60:27): error: Unknown gep/load configuration for synthesis.
loc("vqe_h2.cpp":47:3): error: 'cc.compute_ptr' op too many indices (2) for the source pointer
Aborted (core dumped)
```

**The problem can also be reproduced by simply calling `cudaq::observe(ansatz, H, init_params, n_qubits, n_layers)`** (i.e. no VQE required).

### Expected behavior

Running the same thing with a 0.7.1 client produces valid results.

Yes, I believe PR #1853 introduced the problem

### Environment

- **CUDA Quantum version**: Everything including and after f444e01870e47da618c309d778aa4964410d2a4d
- **Python version**: N/A
- **C++ compiler**: N/A
- **Operating system**: CUDA-Q Docker image

### Suggestions

_No response_

bmhowe23:
Fixes #1975 

This should fix the `Test failed (failed to execute): targettests/Remote-Sim/vqe_h2.cpp` errors in the [NVQC Regression Tests](https://github.com/NVIDIA/cuda-quantum/actions/workflows/nvqc_regression_tests.yml) and the [Nightly Integration Tests](https://github.com/NVIDIA/cuda-quantum/actions/workflows/integration_tests.yml). (Note that the bugfix has nothing to do with VQE, per-se.)


targettests/Remote-Sim/observe_h2.cpp
```diff
@@ -0,0 +1,128 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// REQUIRES: remote-sim
+
+// clang-format off
+// RUN: nvq++ %cpp_std --target remote-mqpu --remote-mqpu-auto-launch 1 %s -o %t && %t
```
schweitzpgi, line 12: Will this command use quake synthesis?

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -166,15 +166,14 @@ synthesizeVectorArgument(OpBuilder &builder, ModuleOp module, unsigned &counter,
...
         loadOp.replaceAllUsesWith(newVal);
-        continue;
-      }
-      return elePtrOp.emitError(
-          "Unknown gep/load configuration for synthesis.");
+      else
+        allLoadUsers = false;
     }
-    return success();
+    return allLoadUsers ? success() : failure();
```
schweitzpgi, line 176:
```suggestion
    return success(allLoadUsers);
```
I believe this syntax is still ok in MLIR...

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -166,15 +166,14 @@ synthesizeVectorArgument(OpBuilder &builder, ModuleOp module, unsigned &counter,
...
                           Value newVal) -> LogicalResult {
+    bool allLoadUsers = true;
     for (auto *u : elePtrOp->getUsers()) {
-      if (auto loadOp = dyn_cast<cudaq::cc::LoadOp>(u)) {
+      if (auto loadOp = dyn_cast<cudaq::cc::LoadOp>(u))
         loadOp.replaceAllUsesWith(newVal);
-        continue;
-      }
-      return elePtrOp.emitError(
-          "Unknown gep/load configuration for synthesis.");
```
schweitzpgi, line 175: This is better than having the pass fail. Nice improvement.

targettests/Remote-Sim/observe_h2.cpp
bmhowe23: No, I put it in there to exercise/demonstrate that the answers don't change between runs that require synthesis and runs that don't. If desired, I could remove it and put a comment near the `assert` line requesting that developers manually run library mode in the future if the `assert` line ever needs to change, but this seemed like the safest way to guarantee that the answers were correct.

targettests/Remote-Sim/observe_h2.cpp
```diff
@@ -0,0 +1,128 @@
...
+  for (int i = 0; i < n_params; i++)
+    init_params[i] = -1 + i * (2.0 / n_params);
+
+  so4_fabric ansatz;
+
+  auto energy =
+      cudaq::observe(ansatz, H, init_params, n_qubits, n_layers).expectation();
+  printf("energy %f\n", energy);
+  assert(std::abs(energy - -0.320848) < 1e-3);
+}
```
schweitzpgi, line 128: `main` should have a `return <int>;` and not just fall off the end, esp. since we're checking the return code in the test.

bmhowe23: > `main` should have a `return <int>;` and not just fall off the end, esp. since we're checking the return code in the test.

Updated in cbe57e8cd

schweitzpgi (approved):
Looks great. Thanks, Ben.

PR #1982, opened 2024-07-21
[RFC] Handle device pointer for state initialization

### LINKED ISSUE #1788: [StateHandling] Implement custatevec state initialization from state vector in device memory
Implement: `FIXME handle case where pointer is a device pointer` in `CuStateVecCircuitSimulator::addQubitsToState`

sacpis:
Fixes #1788 

This PR checks if the pointer is a device pointer and then does memcpy from device to device.

### DIFF
--- a/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cpp
+++ b/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cpp
@@ -210,22 +210,22 @@ protected:
 
           // Perform device to device copy
           HANDLE_CUDA_ERROR(cudaMemcpy(deviceStateVector, state,
-                                    stateDimension * sizeof(CudaDataType),
-                                    cudaMemcpyDeviceToDevice));
+                                       stateDimension * sizeof(CudaDataType),
+                                       cudaMemcpyDeviceToDevice));
 
           // Restore the current device
           HANDLE_CUDA_ERROR(cudaSetDevice(currentDevice));
         } else {
           // Memory is on the same GPU
           HANDLE_CUDA_ERROR(cudaMemcpy(deviceStateVector, state,
-                                    stateDimension * sizeof(CudaDataType),
-                                    cudaMemcpyDeviceToDevice));
+                                       stateDimension * sizeof(CudaDataType),
+                                       cudaMemcpyDeviceToDevice));
         }
       } else {
         // Else, copy from host to device
         HANDLE_CUDA_ERROR(cudaMemcpy(deviceStateVector, state,
-                                   stateDimension * sizeof(CudaDataType),
-                                   cudaMemcpyHostToDevice));
+                                     stateDimension * sizeof(CudaDataType),
+                                     cudaMemcpyHostToDevice));
       }
       return;
     }
@@ -263,8 +263,8 @@ protected:
 
           // Perform device to device copy
           HANDLE_CUDA_ERROR(cudaMemcpy(otherState, state,
-                                    stateDimension * sizeof(CudaDataType),
-                                    cudaMemcpyDeviceToDevice));
+                                       stateDimension * sizeof(CudaDataType),
+                                       cudaMemcpyDeviceToDevice));
 
           // Restore the current device
           HANDLE_CUDA_ERROR(cudaSetDevice(currentDevice));
@@ -272,8 +272,8 @@ protected:
       } else {
         // Else, copy from host to device
         HANDLE_CUDA_ERROR(cudaMemcpy(otherState, state,
-                                   (1UL << count) * sizeof(CudaDataType),
-                                   cudaMemcpyHostToDevice));
+                                     (1UL << count) * sizeof(CudaDataType),
+                                     cudaMemcpyHostToDevice));
       }
     }
 


runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -324,11 +335,22 @@ protected:
...
           reinterpret_cast<CudaDataType *>(otherState), (1UL << count));
     } else {
-
-      // FIXME Handle case where data is already on GPU
-      HANDLE_CUDA_ERROR(cudaMemcpy(otherState, state,
+      // Check if the pointer is a device pointer
+      cudaPointerAttributes attributes;
+      cudaPointerGetAttributes(&attributes, state);
+
+      if (attributes.type == cudaMemoryTypeDevice) {
```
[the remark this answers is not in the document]
schweitzpgi: Thanks, Sachin.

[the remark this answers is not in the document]
bmhowe23: @schweitzpgi - just curious, why do we need to see if the memory is on a different GPU? Maybe I'm just missing something, but it looks like the `cudaMemcpy` docs don't describe any special requirements other than setting the correct `cudaMemcpyDeviceToDevice` flag. It looks like @sacpis changed it to call `cudaSetDevice` to the device of the source memory before the copy if it was on a different device, so I was just curious where the requirement to do that came from.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -294,16 +294,43 @@ protected:
       }
 
       // User state provided...
-
-      // FIXME handle case where pointer is a device pointer
+      // Check if the pointer is a device pointer
+      cudaPointerAttributes attributes;
+      cudaPointerGetAttributes(&attributes, state);
```
1tnguyen, line 299: Could we add a `HANDLE_CUDA_ERROR` wrapper for this call as well?

1tnguyen (approved):
LGTM 👍 Thanks, @sacpis.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -324,11 +351,38 @@ protected:
...
+                                    stateDimension * sizeof(CudaDataType),
+                                    cudaMemcpyDeviceToDevice));
+
+          // Restore the current device
+          HANDLE_CUDA_ERROR(cudaSetDevice(currentDevice));
+        } else {
+          // Memory is on the same GPU
+          HANDLE_CUDA_ERROR(cudaMemcpy(otherState, state,
+                                    stateDimension * sizeof(CudaDataType),
+                                    cudaMemcpyDeviceToDevice));
```
1tnguyen, line 378: I wonder if we need a copy here. `otherState` is only needed for the below kron function where it is passed as a const argument anyway.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -294,16 +294,43 @@ protected:
...
       ScopedTraceWithContext(
           "CuStateVecCircuitSimulator::addQubitsToState cudaMemcpy",
           stateDimension * sizeof(CudaDataType));
-      HANDLE_CUDA_ERROR(cudaMemcpy(deviceStateVector, state,
+
+      if (attributes.type == cudaMemoryTypeDevice) {
+        int currentDevice;
+        HANDLE_CUDA_ERROR(cudaGetDevice(&currentDevice));
+
+        if (attributes.device != currentDevice) {
```
bmhowe23, line 310: This was discussed offline, but just to capture it here, we think it would be good to add a multi-GPU test somewhere to exercise this new code in this if branch; otherwise, the code in here would never have been tested, and there is some question about whether or not the calls to `cudaSetDevice` would work as intended here.

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -324,11 +351,33 @@ protected:
...
-      HANDLE_CUDA_ERROR(cudaMemcpy(otherState, state,
+      // Check if the pointer is a device pointer
+      cudaPointerAttributes attributes;
+      HANDLE_CUDA_ERROR(cudaPointerGetAttributes(&attributes, state));
+
+      if (attributes.type == cudaMemoryTypeDevice) {
+        int currentDevice;
+        HANDLE_CUDA_ERROR(cudaGetDevice(&currentDevice));
+
+        if (attributes.device != currentDevice) {
```
bmhowe23, line 362: Same comment here.

bmhowe23 (changes requested):
Just capturing the offline discussion amongst the 4 of us...

runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
bmhowe23: Another alternative would be to throw an "unsupported" error message here, but I think we should still add a test for the "same GPU" functionality as well.

bmhowe23:
> Here I am blocked on including nvqir/CircuitSimulator in get_state_tester. Seems like the builder is not happy including it in the file as it throws the below error.
> 
> ```
> In file included from /workspaces/cuda-quantum/runtime/common/Logger.h:14,
>                  from /workspaces/cuda-quantum/runtime/nvqir/CircuitSimulator.h:13,
>                  from /workspaces/cuda-quantum/unittests/gpu/get_state_tester.cu:12:
> /workspaces/cuda-quantum/runtime/common/FmtCore.h:16:10: fatal error: fmt/chrono.h: No such file or directory
>    16 | #include <fmt/chrono.h>
>       |          ^~~~~~~~~~~~~~
> compilation terminated.
> ```

Thanks, Sachin. I'm definitely open to any other ideas for testing here. I just think that we need to have some sort of test to prove that code changes are working as intended. If a unit test isn't working out, perhaps an application-level test (one where the user is providing a pointer to GPU memory) would work better than a unit test?

PR #1983, opened 2024-07-22
Downstream `mgpu` updates

1tnguyen:
### Description

- Update `mgpu` commit hash.

- Update `auditwheel` to exclude the NVIDIA Management Library (NVML) library.
This is a new dependency of the `mgpu` backend. 

**Note**: `libnvidia-ml.so.1` should be included when installing an NVIDIA driver (`nvidia-smi` command line tool is present).

- Update docs with information about new features/configurations of the `nvidia` target: host-device state vector migration, new configurations, etc.

### Checklist

- [ ] Run Publishing pipeline (GPU backend validation)

- [x] Final update of `mgpu` commit hash after merging downstream MR.


docs/sphinx/using/backends/simulators.rst
```diff
@@ -42,24 +65,64 @@ To execute a program on the :code:`nvidia` target, use the following commands:
...
 
+In the single-GPU mode, the :code:`nvidia` target provides the following environment variable options.
+
+.. list-table:: **Environment variable options supported in single-GPU mode**
+  :widths: 20 30 50
+
+  * - Option
+    - Value
+    - Description
+  * - ``CUDAQ_FUSION_MAX_QUBITS``
```
bmhowe23, line 102: Is this in addition to, or is it replacing, `CUDAQ_MGPU_FUSE`? That environment variable was advertised in a published blog with the 0.7.1 release, so unless there is a pressing need to deprecate it, I would recommend we retain that same name.

1tnguyen: It's in addition to (as an alias since the single-GPU simulator, i.e., no MPI required, now also supports gate fusing). 

Thanks for the information. Yea, we should retain that exact config name in this case: I added `CUDAQ_MGPU_FUSE` to the table for the multi-node section and revert the performance tuning note below to be the same as before.

bmhowe23 (approved):
The changes look good to me and should be ready to merge as soon as the [Publishing run](https://github.com/NVIDIA/cuda-quantum/actions/runs/10104435235) finishes successfully. Thanks, Thien!

bmhowe23 (approved):
Still looks good to me ... hopefully https://github.com/NVIDIA/cuda-quantum/actions/runs/10116290104 finishes up without issues this time.

1tnguyen:
> Still looks good to me ... hopefully https://github.com/NVIDIA/cuda-quantum/actions/runs/10116290104 finishes up without issues this time.

The pipeline finished without any issues. The OMP setting does seem to help reduce/stabilize notebook validation time.

PR #1990, opened 2024-07-22
New Tutorial - Divisive Clustering

mawolf2023:
### Description

I am uploading a new Jupyter notebook tutorial based on this https://arxiv.org/abs/2402.01529 paper focusing on divisive clustering.  The tutorial was jointly developed with one of the authors.  I placed the notebook in the docs section with all of the other notebooks.   This example has a couple of auxiliary files, I created a source folder for them and use `wget` to grab the files if someone uses the notebook.  If anything is placed in the wrong spot or needs to be modified, please let me know.


mawolf2023:

bmhowe23:
recheck

mawolf2023:
@bmhowe23 , Thanks so much for this. I just saw the preview and it looks like the LaTeX in the notebooks does not render. And I also noticed the divisive clustering function now shows up a bullet point in the tutorial page, but clicking on it does not open anything so I think we would want that hidden.

bmhowe23 (commented):
The current set of commits pass the CI, but the new notebook has to be skipped during the regular per-PR validation since it requires GPUs to execute. In fact, it requires the `nvidia-mgpu` backend to really run it will rarely get exercised by the CI if we were to merge this PR right now. We should probably discuss what the best path forward is. The options that I can think of are:
* Leave it as is right now and just know that the CI will not exercise the notebook on a regular basis
* Remove the `nvidia-mgpu` dependency to make it run during Deployment (but still not during the per-PR CI)
* Remove all `nvidia` target dependencies (somehow?) so that it runs with every PR.

bmhowe23:
> @bmhowe23 , Thanks so much for this. I just saw the preview and it looks like the LaTeX in the notebooks does not render.

Yeah, the LaTeX doesn't render in the GH docs preview for some reason...that's a known issue. The best way to review it is to run `bash scripts/build_docs.sh` locally and then view it locally on your machine.

> And I also noticed the divisive clustering function now shows up a bullet point in the tutorial page, but clicking on it does not open anything so I think we would want that hidden.

That's strange, it works for me. Maybe we should sync up offline to make sure we're looking at the same thing.

PR #1991, opened 2024-07-22
[Python] Follow-up to PR#1968

khalatepradnya:
* Fix the Python `SyntaxWarning: assertion is always true, perhaps remove parentheses?`
* Update comment in tests to point to GitHub issue


1tnguyen (approved):
LGTM 👍

PR #1994, opened 2024-07-22
[Bug-fix] [Python] Checking for mid circuit measurement

### LINKED ISSUE #1875: issue with if condition after mid circuit measurement
### Required prerequisites

### Describe the bug

If I run example 1, it will work and print the register `a`. However, if I run example 2, it will not print the register `a`

Example 1:

```
import cudaq

@cudaq.kernel
def kernel_break():
    ancilla_a=cudaq.qubit()
    ancilla_b=cudaq.qubit()
    q=cudaq.qubit()
    
    
    
    h(ancilla_a)
    h(ancilla_b)
    x(q)
        
    aux_1=mz(ancilla_a)
    aux_2=mz(ancilla_b)
    if aux_1==0:
        if aux_2==0:
            x.ctrl(ancilla_a,q)
            a=mz(q)
         
            


count=cudaq.sample(kernel_break,shots_count=1000)
print(count)
```

Example 2:
```
import cudaq

@cudaq.kernel
def kernel_break():
    ancilla_a=cudaq.qubit()
    ancilla_b=cudaq.qubit()
    q=cudaq.qubit()
    
    
    
    h(ancilla_a)
    h(ancilla_b)
    x(q)
        
    aux_1=mz(ancilla_a)
    aux_2=mz(ancilla_b)
    if aux_1==0 and aux_2==0:
        #if aux_2==0:
        x.ctrl(ancilla_a,q)
        a=mz(q)
         
            


count=cudaq.sample(kernel_break,shots_count=1000)
print(count)
```

### Expected behavior

```
if aux_1==0:
   if aux_2==0:
      ......
```

and 

```
if aux_1==0 and axu_2==0:
   ....
```

I expect the two if statement should work.

Not a regression

### Environment

- **CUDA Quantum version**: latest
- **Python version**: 3.10
- **C++ compiler**: 
- **Operating system**: linux

### Suggestions

_No response_

khalatepradnya:
* Handle more condition(s)
* Refactor common code

Fixes #1875


python/cudaq/kernel/analysis.py
```diff
@@ -52,38 +52,45 @@ def getVariableName(self, node):
             return self.getVariableName(node.value)
         return ''
 
+    def checkForMeasureResult(self, value):
```
bmhowe23, line 55: Is there a reason we can't use the `quake-add-metadata` pass to determine if it has mid-circuit measurements? It seems like some (or all?) of this is duplicated logic from `lib/Optimizer/Transforms/QuakeAddMetadata.cpp`. It's possible that Python needs to know this earlier than C++, and if so, then perhaps running that pass would be too late, but I figured I would ask. My very quick scan of the Python code's use of `hasMidCircuitMeasures`/ `conditionalOnMeasure` makes it seem like its whole job in life is setting `qubitMeasurementFeedback`, and that aforementioned pass does the same thing.

schweitzpgi: I have the same question. We should not duplicate this functionality in Python unless there is an extremely compelling argument why the pass can't be run.

khalatepradnya: Great question! I know this is invoked before the compile to MLIR stage in Python, but I will confirm if it can be moved down to make use of the existing pass.

python/cudaq/kernel/analysis.py
```diff
@@ -52,38 +52,45 @@ def getVariableName(self, node):
...
+        if self.isMeasureCallOp(value) or self.getVariableName(
+                value) in self.measureResultsVars:
+            return True
+        return False
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 51cec310330ad2187858bd1f18c1581bafa57851

python/cudaq/kernel/analysis.py
khalatepradnya: I think I will have to compile all the kernels beforehand to make use of the `quake-add-metadata` function pass in Python workflow instead of the `MidCircuitMeasurementAnalyzer` class.
@amccaskey: Can you confirm this?

amccaskey: This analysis pass may actually be a relic from the original implementation - originally we had a library mode for Python and this pass was necessary to inform the runtime that the kernel had mid circuit measurement. We can probably drop it now for the MLIR pass. But you will have to update `sample.py` to check the Module attribute instead of the python kernel metadata attribute.

khalatepradnya: WIP...

bmhowe23: > WIP...

IMO, if it's unreasonable to get this done before the release, I'd be ok deferring this work until after the release. I'd hate to rush some refactoring at the last minute, as that could add its own risk to the release.

khalatepradnya: Sounds good, I will make the refactoring a separate PR.

python/cudaq/kernel/analysis.py
```diff
@@ -52,38 +52,45 @@ def getVariableName(self, node):
...
         # Catch `if something BoolOp mz(q)` or `something BoolOp var` (`var = mz(q)`)
-        elif isinstance(condition,
-                        ast.BoolOp) and 'values' in condition.__dict__:
-            for node in condition.__dict__['values']:
-                if self.isMeasureCallOp(node):
+        if isinstance(condition, ast.BoolOp) and 'values' in condition.__dict__:
+
+            for value in condition.__dict__['values']:
+                if self.checkForMeasureResult(value):
                     self.hasMidCircuitMeasures = True
```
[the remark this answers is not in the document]
khalatepradnya: I don't think it is necessary.

PR #1996, opened 2024-07-23
Enable quantum struct member types in kernel code 

amccaskey:
Enable expressions in C++ like this 
```cpp
struct MyStructWithQuantumType {
  int i;
  double d;
  cudaq::qview<> q;
};

__qpu__ void applyHadamardOnQubit(cudaq::qubit &q) { h(q); }

__qpu__ void pureDeviceKernelTakesQuantumStruct(MyStructWithQuantumType t) {
  h(t.q);
  for (int i = 0; i < t.i; i++)
    applyHadamardOnQubit(t.q[i]);
}

__qpu__ void entryPoint(int i) {
  cudaq::qvector q(i);
  MyStructWithQuantumType t{i, 2.2, q};
  pureDeviceKernelTakesQuantumStruct(t);
}
```

### DIFF
--- a/lib/Frontend/nvqpp/ASTBridge.cpp
+++ b/lib/Frontend/nvqpp/ASTBridge.cpp
@@ -50,8 +50,9 @@ listReachableFunctions(clang::CallGraphNode *cgn) {
   return result;
 }
 
-// Does `ty` refer to a Quake quantum type? This also checks custom recursive
-// types. It does not check builtin recursive types; e.g., `!llvm.ptr<T>`.
+// Does `ty` refer to a Quake quantum type? This also checks CC Dialect
+// recursive types. It does not check builtin recursive types; e.g.,
+// `!llvm.ptr<T>`.
 static bool isQubitType(Type ty) {
   if (isa<quake::RefType, quake::VeqType>(ty))
     return true;
@@ -360,6 +361,18 @@ private:
 
 #ifndef NDEBUG
 namespace cudaq::details {
+bool QuakeBridgeVisitor::isQuantumStructType(Type ty) {
+  auto structTy = dyn_cast<cc::StructType>(ty);
+  if (!structTy)
+    return false;
+
+  for (auto member : structTy.getMembers())
+    if (quake::isQuantumType(member))
+      return true;
+
+  return false;
+}
+
 bool QuakeBridgeVisitor::pushValue(Value v) {
   LLVM_DEBUG(llvm::dbgs() << std::string(valueStack.size(), ' ')
                           << "+push value: ";
--- a/lib/Frontend/nvqpp/ASTBridge.cpp
+++ b/lib/Frontend/nvqpp/ASTBridge.cpp
@@ -53,8 +53,14 @@ listReachableFunctions(clang::CallGraphNode *cgn) {
 // Does `ty` refer to a Quake quantum type? This also checks custom recursive
 // types. It does not check builtin recursive types; e.g., `!llvm.ptr<T>`.
 static bool isQubitType(Type ty) {
-  if (ty.isa<quake::RefType, quake::VeqType>())
+  if (isa<quake::RefType, quake::VeqType>(ty))
     return true;
+
+  if (auto structTy = dyn_cast<cudaq::cc::StructType>(ty))
+    for (auto memberTy : structTy.getMembers())
+      if (isQubitType(memberTy))
+        return true;
+
   // FIXME: next if case is a bug.
   if (auto vecTy = dyn_cast<cudaq::cc::StdvecType>(ty))
     return isQubitType(vecTy.getElementType());
@@ -70,6 +76,7 @@ static bool hasAnyQubitTypes(FunctionType funcTy) {
   for (auto ty : funcTy.getResults())
     if (isQubitType(ty))
       return true;
+
   return false;
 }
 
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -2873,9 +2873,16 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
     return pushValue(builder.create<cc::LoadOp>(loc, copyObj));
   }
 
-  // Struct construction handled elsewhere in the traversal
-  if (auto structTy = dyn_cast<cc::StructType>(ctorTy))
+  // Note we only screen for structs that contain data members.
+  // Revisit this if we ever enable Kernel structs with data members.
+  if (auto structTy = dyn_cast<cc::StructType>(ctorTy);
+      structTy && !structTy.getMembers().empty()) {
+    if (ctor->isDefaultConstructor())
+      return pushValue(builder.create<cc::AllocaOp>(loc, ctorTy));
+
+    // Struct construction handled elsewhere in the traversal.
     return true;
+  }
 
   // TODO: remove this when we can handle ctors more generally.
   if (!ctor->isDefaultConstructor()) {
--- a/lib/Frontend/nvqpp/ConvertExpr.cpp
+++ b/lib/Frontend/nvqpp/ConvertExpr.cpp
@@ -1490,6 +1490,21 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
           builder.create<quake::DiscriminateOp>(loc, resTy, measure));
     }
 
+    // We may have qubit values that come from struct members, in this
+    // case the value may be provided as a pointer, need to load it here.
+    SmallVector<Value> localArgs;
+    for (auto arg : args) {
+      if (auto ptrTy = dyn_cast<cudaq::cc::PointerType>(arg.getType())) {
+        if (isa<quake::VeqType, quake::RefType>(ptrTy.getElementType())) {
+          localArgs.push_back(builder.create<cudaq::cc::LoadOp>(loc, arg));
+          continue;
+        }
+      }
+      localArgs.push_back(arg);
+    }
+
+    args = localArgs;
+
     // Handle the quantum gate set.
     auto reportNegateError = [&]() {
       reportClangError(x, mangler, "target qubit cannot be negated");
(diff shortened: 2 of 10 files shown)


test/AST-Quake/struct-1.cpp
```diff
@@ -221,7 +221,6 @@ struct S5 {
 // CHECK:           %[[VAL_10:.*]] = cc.load %[[VAL_9]] : !cc.ptr<f64>
 // CHECK:           call @_Z15debug_the_thingiid(%[[VAL_6]], %[[VAL_8]], %[[VAL_10]]) : (i32, i32, f64) -> ()
 // CHECK:           %[[VAL_11:.*]] = cc.alloca !cc.struct<"ResultThing" {i1, i1, i1, i32} [64,4]>
-// CHECK:           call @_ZN11ResultThingC1Ev(%[[VAL_11]]) : (!cc.ptr<!cc.struct<"ResultThing" {i1, i1, i1, i32} [64,4]>>) -> ()
```
bmhowe23, line 224: It looks like this is removing the call to the `ResultThing` constructor. That may (?) be safe to do in this test case where all values are set after the variable is declared, but is that really desired for all structs?

lib/Frontend/nvqpp/ASTBridge.cpp
```diff
@@ -53,8 +53,14 @@ listReachableFunctions(clang::CallGraphNode *cgn) {
 // Does `ty` refer to a Quake quantum type? This also checks custom recursive
 // types. It does not check builtin recursive types; e.g., `!llvm.ptr<T>`.
```
schweitzpgi, line 54: Should update this comment. "custom recursive types" might be better stated as something like CC dialect types? (I wonder just how old this comment is. heh)

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1490,6 +1490,21 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
           builder.create<quake::DiscriminateOp>(loc, resTy, measure));
     }
 
+    // We may have qubit values that come from struct members, in this
+    // case the value may be provided as a pointer, need to load it here.
```
schweitzpgi, line 1494: I don't think we really want classical pointers to quantum references. (We always get rid of the C++ references.)

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -2168,6 +2183,24 @@ bool QuakeBridgeVisitor::VisitCXXOperatorCallExpr(
       auto idx_var = popValue();
       auto qreg_var = popValue();
 
+      // We could have the case where the qvector is coming from a struct
+      // and is not in the symbol table.
```
schweitzpgi, line 2187: Why wouldn't this be an expression like `var.qview_member`?

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -2829,6 +2862,17 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
     return pushValue(builder.create<cc::LoadOp>(loc, copyObj));
   }
 
+  // Note we only screen for structs that contain data members.
+  // Revisit this if we ever enable Kernel structs with data members.
+  if (auto structTy = dyn_cast<cc::StructType>(ctorTy);
```
schweitzpgi, line 2867: Why are we generating constructors someplace other than when the constructor is visited?

targettests/execution/quantum_struct_exec.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+struct test {
+  int i;
+  double d;
+  cudaq::qview<> q;
+};
+
+__qpu__ void hello(cudaq::qubit &q) { h(q); }
+
+__qpu__ void kernel(test t) {
```
schweitzpgi, line 21: Passing a copy of an object that contains a `qview<>` here ... seems like it ought to be passing this by reference.

test/AST-Quake/struct-1.cpp
schweitzpgi: Not really, no.

lib/Frontend/nvqpp/ConvertExpr.cpp
amccaskey: When you extract a member from a struct you use ComputePtrOp, which returns a pointer to the member. How do we get around that?

lib/Frontend/nvqpp/ConvertExpr.cpp
amccaskey: I would have thought so. When I ran the code as is (`h(var.qview_member[2]);`, ), the bridge threw an error here - it was looking for the variable name (`qview_member`) in they symbol table.

lib/Frontend/nvqpp/ConvertExpr.cpp
amccaskey: I originally constructed the full struct here, but for something like `MyStruct s(var0, var1, var1, ...)`, the result ended up with 2 struct creations and filling the struct. Not sure where the first one is created, but I figured you'd implemented it to handle construction somewhere already. Note that for `MyStruct s` (no constructor args), you are required to alloca here.

targettests/execution/quantum_struct_exec.cpp
amccaskey: `qview<>` is meant to be passed by value (`std::span` like)

test/AST-Quake/struct-1.cpp
amccaskey: This was a question I had - the updates here (in my head) were to allocate the struct and populate its members (`MyStruct s(var0, var1,...)`). I forgot about the case of `MyStruct s`, but I wonder what the semantics of alloca are? Does `alloca struct<"ResultThing" {i32}>` allocate the struct with a default integer of 0? If not, then need some extra code to initialize the members. Also should we have been originally calling the host-side constructor from quantum device code?

schweitzpgi: I'll update the description, but cc.alloca has always been pure allocation. To load from such a location without first initializing with a value will be loading garbage.

schweitzpgi: We can and do make such calls in "pseudo-library" mode. If the call is still present when we convert to QIR, the QIR checks fail.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: I was overly terse.

This is a huge, huge change. Lots of stuff in the compiler is predicated on the fact that we use `quake.alloca` to allocate qubits. Those qubits are allocated and used and have either memory-reference semantics or value semantics, etc. Everything is structured so we can do inlining, loop unrolling, qubit factoring, scheduling, etc. to eventually end up with a small(ish) cardinal set of qubits and a DAG (quantum circuit).

Stashing qubit references (which would semantically be storing values with quantum memory semantics in variables with classical memory semantics) will break the previous paragraph. Instead of linear chains of def-use on qubits that can be discriminated from one another, this would mean qubits and quantum memory references can alias, at a minimum. If one could bound the number of qubits (total $n$) to $k$ but not their identities in a region of code, this could mean one could have something like $n \choose k$ path replications to create the DAG in the worst case.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: That's, well, weird. Need to figure that out. (I thought members were working fine.)

lib/Frontend/nvqpp/ConvertExpr.cpp
amccaskey: I will remove this and convert structs with quantum members to value semantics (insert / extract value)

test/AST-Quake/struct-1.cpp
amccaskey: reverted this change

amccaskey:
@schweitzpgi I updated everything here to treat quantum structs with the value semantic operations `cc.undef`, `cc.insertvalue`, and `cc.extractvalue`. Please take a look and see if this fits better with the quake model.

targettests/execution/quantum_struct_exec.cpp
```diff
@@ -0,0 +1,37 @@
...
+  h(t.q);
+  for (int i = 0; i < t.i; i++)
+    hello(t.q[i]);
+}
+
+__qpu__ void entry(int i) {
+  cudaq::qvector q(i);
+  test tt{i, 2.2, q};
+  kernel(tt);
+}
```
bmhowe23, line 31: Thanks, Alex. I downloaded this example to play with it a bit more. I can't get it to work for any of the target hardware. In essence, it looks like the loop unrolling passes can't figure out what to do with the `cudaq::qview<> q` struct member (or perhaps other members too).

Trying to simplify as much as possible ... if you remove the device `kernel` function and the `for` loop from the example, you're left with something like this:
```cpp
__qpu__ void entry(int i) {
  cudaq::qvector q(i);
  test tt{i, 2.2, q};

  // This works
  h(q);
  
  // This fails with error: 'cc.loop' op not a constant upper bound loop
  h(tt.q);
}
```
It seems like the above example should, in principle, be able to work, but when compiling and running with `nvq++ --target quantinuum --emulate ../targettests/execution/quantum_struct_exec.cpp`, it fails with the loop unrolling error.

Do you think there is a reasonable path forward to making this work? Would the constant propagation logic need to be updated to handle this? Or something else?

amccaskey: I've added a canonicalizer to address this. Loops unroll now.

schweitzpgi: Just to document this: few of the transformation passes were built to work with aggregate typed values. (None to work with loading and storing quantum typed reference values.)

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -970,11 +970,64 @@ struct FuseWithConstantArray
     return success();
   }
 };
+
+struct FoldQuantumStructMember
```
schweitzpgi, line 974: I think we should just implement scalar replacement. "A rising tide lifts all boats."

PR #2008, opened 2024-07-25
Use standard function to search

sacpis:
* Use standard function std::find to search, instead of loops
* Remove redundant checks
* Return early if g_plugin is found, to avoid unnecessary checks

# Test result

```
100% tests passed, 0 tests failed out of 636

Label Time Summary:
gpu_required    =  88.00 sec*proc (361 tests)

Total Test time (real) = 418.11 sec
```

### DIFF
--- a/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
+++ b/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
@@ -290,6 +290,7 @@ protected:
         initializeDeviceStateVector<<<n_blocks, threads_per_block>>>(
             reinterpret_cast<CudaDataType *>(deviceStateVector),
             stateDimension);
+        HANDLE_CUDA_ERROR(cudaGetLastError());
         return;
       }
 
@@ -323,6 +324,7 @@ protected:
     if (state == nullptr) {
       initializeDeviceStateVector<<<n_blocks, threads_per_block>>>(
           reinterpret_cast<CudaDataType *>(otherState), (1UL << count));
+      HANDLE_CUDA_ERROR(cudaGetLastError());
     } else {
 
       // FIXME Handle case where data is already on GPU
@@ -482,6 +484,7 @@ protected:
         (stateDimension + threads_per_block - 1) / threads_per_block;
     initializeDeviceStateVector<<<n_blocks, threads_per_block>>>(
         reinterpret_cast<CudaDataType *>(deviceStateVector), stateDimension);
+    HANDLE_CUDA_ERROR(cudaGetLastError());
   }
 
 public:


runtime/cudaq/cudaq.cpp
```diff
@@ -300,27 +291,31 @@ std::string get_quake_by_name(const std::string &kernelName,
...
-      } else {
-        result = v;
-      }
-    }
-  }
+
+  std::find_if(
+      quakeRegistry.begin(), quakeRegistry.end(), [&](const auto &pair) {
+        if (pair.first == kernelName) {
+          result = pair.second;
```
bmhowe23, line 298: I think it would be worth retaining the comment that previously existed here IMO.
```suggestion
          // Exact match. Return the code.
          result = pair.second;
```

runtime/cudaq/cudaq.cpp
```diff
@@ -300,27 +291,31 @@ std::string get_quake_by_name(const std::string &kernelName,
...
+          return true;
+        }
+
+        if (pair.first.starts_with(kernelNamePrefix)) {
+          if (result.has_value()) {
```
bmhowe23, line 303: I think it would be worth retaining the comment that previously existed here IMO.
```suggestion
          // Prefix match. Record it and make sure that it is a unique prefix.
          if (result.has_value()) {
```

runtime/cudaq/cudaq.cpp
```diff
@@ -36,98 +36,91 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
...
-  return commPlugin != nullptr;
-}
+bool available() { return getMpiPlugin(/*unsafe=*/true) != nullptr; }
 
 void initialize() {
-  auto *commPlugin = getMpiPlugin();
-  commPlugin->initialize();
+  if (auto *commPlugin = getMpiPlugin()) {
+    commPlugin->initialize();
+  }
```
bmhowe23, line 108: I'm not sure I follow the need for the change here. `getMpiPlugin()` is guaranteed to return a valid pointer (unless we're setting `unsafe` to true, and we're not doing that), so the added check seems unnecessary IMO.

runtime/cudaq/cudaq.cpp
```diff
@@ -36,98 +36,91 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
...
-  const auto np = commPlugin->num_ranks();
-  if (pid == 0)
-    cudaq::info("MPI Initialized, nRanks = {}", np);
+  if (auto *commPlugin = getMpiPlugin()) {
+    commPlugin->initialize(argc, argv);
+    const auto pid = commPlugin->rank();
+    const auto np = commPlugin->num_ranks();
+    if (pid == 0)
+      cudaq::info("MPI Initialized, nRanks = {}", np);
+  }
```
bmhowe23, line 118: I'm not sure I follow the need for the change here. `getMpiPlugin()` is guaranteed to return a valid pointer (unless we're setting `unsafe` to true, and we're not doing that), so the added check seems unnecessary IMO.

runtime/cudaq/cudaq.cpp
```diff
@@ -36,98 +36,91 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
...
-    // with MPI)
-    const char *mpiLibPath = std::getenv("CUDAQ_MPI_COMM_LIB");
-    if (mpiLibPath) {
-      // The user has set the environment variable.
-      cudaq::info("Load MPI comm plugin from CUDAQ_MPI_COMM_LIB environment "
-                  "variable at '{}'",
-                  mpiLibPath);
-      g_plugin = std::make_unique<cudaq::MPIPlugin>(mpiLibPath);
+  if (g_plugin)
+    return g_plugin.get();
```
bmhowe23, line 40: Nice simplification here.

runtime/cudaq/cudaq.cpp
```diff
@@ -300,27 +291,31 @@ std::string get_quake_by_name(const std::string &kernelName,
...
+        if (pair.first.starts_with(kernelNamePrefix)) {
+          if (result.has_value()) {
+            if (throwException)
+              throw std::runtime_error("Quake code for '" + kernelName +
+                                       "' has multiple matches.\n");
+          } else {
+            result = pair.second;
+          }
+        }
+        return false;
```
bmhowe23, line 311: If we're always returning false here, then it seems a bit unintuitive to me to be using `find_if` here because there is no path that returns true inside the `find_if` lambda.

bmhowe23 (commented):
Thanks, Sachin ... just a few comments/questions....

runtime/cudaq/cudaq.cpp
```diff
@@ -300,27 +289,31 @@ std::string get_quake_by_name(const std::string &kernelName,
...
+    if (pair.first.starts_with(kernelNamePrefix)) {
       // Prefix match. Record it and make sure that it is a unique prefix.
       if (result.has_value()) {
-        if (throwException)
+        if (throwException) {
           throw std::runtime_error("Quake code for '" + kernelName +
                                    "' has multiple matches.\n");
-      } else {
-        result = v;
+        } else {
```
bmhowe23, line 305: This `else` used to be paired with the `if (result.has_value())`, but with the brace change, it looks like it's now paired with `if (throwException)`. Is that intentional?

runtime/cudaq/cudaq.cpp
1tnguyen: seconded 💯  Thanks, @sacpis!

1tnguyen (approved):
This LGTM 👍

runtime/cudaq.h
```diff
@@ -96,25 +91,17 @@ std::string get_kernel_template_member_name() {
 }
 
 /// Get the name of a plain old function that is marked as a quantum kernel.
-inline std::string get_kernel_function_name(std::string &&name) {
+inline std::string get_kernel_function_name(std::string name) {
   return "function_" + std::move(name);
 }
-inline std::string get_kernel_function_name(const std::string &name) {
-  return "function_" + name;
-}
```
schweitzpgi, line 104: These changes should be reverted.

The first overload is the r-value reference overload where std::move() makes perfect sense. The second overload covers cases where there isn't an r-value reference and, more importantly, doesn't make a copy.

This change forces a tax on every call to make a copy of the string first.

runtime/cudaq.h
```diff
@@ -96,25 +91,17 @@ std::string get_kernel_template_member_name() {
...
-std::string get_kernel_template_function_name(std::string &&funcName) {
+std::string get_kernel_template_function_name(std::string funcName) {
   std::string name = internal::expand_parameter_pack<Args...>();
   return "instance_function_" + std::move(funcName) + name;
 }
-template <typename... Args>
-std::string get_kernel_template_function_name(const std::string &funcName) {
-  std::string name = internal::expand_parameter_pack<Args...>();
-  return "instance_function_" + funcName + name;
-}
```
schweitzpgi, line 117: Same here.

runtime/cudaq/cudaq.cpp
```diff
@@ -36,74 +36,71 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
...
-        cudaq::info("Load MPI comm plugin from '{}'",
-                    activatedInterfaceLibFile.c_str());
-        g_plugin = std::make_unique<cudaq::MPIPlugin>(
-            activatedInterfaceLibFile.c_str());
-      } else {
-        const auto pluginsPath = cudaqLibPath.parent_path() / "plugins";
+      const auto pluginsPath = cudaqLibPath.parent_path() / "plugins";
 #if defined(__APPLE__) && defined(__MACH__)
-        const std::string libSuffix = "dylib";
+      const std::string libSuffix = "dylib";
```
schweitzpgi, line 76: Could possibly use `constexpr std::string_view` here to avoid constructing a string on the heap that is just data to be copied into another string object.

runtime/cudaq/cudaq.cpp
```diff
@@ -36,74 +36,71 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
...
   }
 
   return g_plugin.get();
 };
 
-bool available() {
-  auto *commPlugin = getMpiPlugin(/*unsafe=*/true);
-  return commPlugin != nullptr;
-}
+bool available() { return getMpiPlugin(/*unsafe=*/true) != nullptr; }
```
schweitzpgi, line 103: Don't need to compare with 0 and nullptr.

PR #2013, opened 2024-07-26
Update remote-mqpu tests to cleanly abort during assertion checks

bmhowe23:
Update `targettests/Remote-Sim/*.cpp` to use a new macro instead of `assert()` because the standard `assert()` calls are causing hangs during test failures rather than cleanly aborting with an error message. I manually tested this change by modifying one of the assertion checks to intentionally fail while running `llvm-lit`, and it failed as desired now.

### DIFF
--- a/targettests/Remote-Sim/remote_test_assert.h
+++ b/targettests/Remote-Sim/remote_test_assert.h
@@ -1,0 +1,19 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// llvm-lit tests for remote-mqpu tend to hang rather than abort if they hit an
+// assertion error using the regular `assert()` call. Use this instead to
+// cleanly exit/abort from your main function while printing an error message
+// about where the error occurred.
+#define REMOTE_TEST_ASSERT(x)                                                  \
+  do {                                                                         \
+    if (!(x)) {                                                                \
+      printf("Assertion failure in %s:%d\nExiting\n", __FILE__, __LINE__);     \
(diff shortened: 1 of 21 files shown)


targettests/Remote-Sim/remote_test_assert.h
```diff
@@ -0,0 +1,19 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// llvm-lit tests for remote-mqpu tend to hang rather than abort if they hit an
+// assertion error using the regular `assert()` call. Use this instead to
+// cleanly exit/abort from your main function while printing an error message
+// about where the error occurred.
```
khalatepradnya, line 12: Q: Do we know why it hangs?

khalatepradnya (approved):
Changes look good to me. Thanks, Ben!

targettests/Remote-Sim/remote_test_assert.h
bmhowe23: I don't know for sure. I suspect it has something to do with auto-launch the server in the background. I will try to investigate further.

bmhowe23: The problem appears to be that the `AutoLaunchRestServerProcess` destructor is not called when the assertion fails. This is similar to what would happen if `abort()` were called (*). If the `AutoLaunchRestServerProcess` destructor is not called, then the `cudaq-qpud` stays running in the background, and the `llvm-lit` wrapper thinks the test has not completed, so it just hangs waiting for completion.

(*) https://man7.org/linux/man-pages/man3/abort.3.html:
> As with other cases of abnormal termination the functions registered with [atexit(3)](https://man7.org/linux/man-pages/man3/atexit.3.html) and [on_exit(3)](https://man7.org/linux/man-pages/man3/on_exit.3.html) are not called.

I'll update the comment in this file to provide this additional information.

bmhowe23: Comment updated in da01ba463

PR #2014, opened 2024-07-26
Minor updates for downstream projects. 

amccaskey:
* Create imported CMake targets for cudaq targets.
* Enable downstream CMake to set the target. 
* Fix bug where fmtlib not found for downstream project import NVQIR config.
* Fix issue printing out call error in python bridge.
* Add serialize / deserialize to sample result in python bindings.
* carry the custom operator name through to the GateApplicationTask.

### DIFF
--- a/cmake/Modules/CMakeLists.txt
+++ b/cmake/Modules/CMakeLists.txt
@@ -22,6 +22,20 @@ set(LANG_FILES
   CMakeTestCUDAQCompiler.cmake
 )
 
+execute_process(
+  COMMAND git rev-parse HEAD
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tpls/fmt
+  OUTPUT_VARIABLE CUDAQ_FMTLIB_HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Configure the file
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/NVQIRConfig.cmake.in
+  ${CMAKE_BINARY_DIR}/cmake/Modules/NVQIRConfig.cmake
+  @ONLY
+)
+
 install(FILES ${CONFIG_FILES} DESTINATION lib/cmake/cudaq)
 install(FILES ${LANG_FILES} DESTINATION lib/cmake/cudaq)
-install(FILES NVQIRConfig.cmake DESTINATION lib/cmake/nvqir)
+install(FILES ${CMAKE_BINARY_DIR}/cmake/Modules/NVQIRConfig.cmake DESTINATION lib/cmake/nvqir)
--- a/cmake/Modules/CUDAQConfig.cmake
+++ b/cmake/Modules/CUDAQConfig.cmake
@@ -126,42 +126,15 @@ set_target_properties(cudaq::cudaq-builder PROPERTIES
 # Check for the presence of NVIDIA GPUs, if none
 # are found set the default target to qpp-cpu
 set(__tmp_cudaq_target "qpp-cpu")
-include(CheckLanguage)
-check_language(CUDA)
-if(CMAKE_CUDA_COMPILER)
-  enable_language(CUDA)
-  message(STATUS "CUDA compiler found: ${CMAKE_CUDA_COMPILER}")
-  set(CUDA_TEST_SOURCE "
-    #include <cuda_runtime.h>
-    #include <stdio.h>
-    int main() {
-      int deviceCount;
-      cudaError_t error = cudaGetDeviceCount(&deviceCount);
-      if (error != cudaSuccess) {
-        printf(\"cudaGetDeviceCount returned error %d: %s\\n\", error, cudaGetErrorString(error));
-        return 1;
-      }
-      printf(\"%d\", deviceCount);
-      return 0;
-    }
-  ")
-
-  set(CUDA_TEST_FILE "${CMAKE_BINARY_DIR}/platform/cuda_test.cu")
-  file(WRITE "${CUDA_TEST_FILE}" "${CUDA_TEST_SOURCE}")
-
-  try_run(RUN_RESULT COMPILE_RESULT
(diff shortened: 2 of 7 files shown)


schweitzpgi (commented):
LGTM

cmake/Modules/NVQIRConfig.cmake
```diff
@@ -13,7 +13,15 @@ get_filename_component(PARENT_DIRECTORY ${NVQIR_CMAKE_DIR} DIRECTORY)
...
 find_dependency(CUDAQSpin REQUIRED HINTS "${PARENT_DIRECTORY}/cudaq")
 find_dependency(CUDAQCommon REQUIRED HINTS "${PARENT_DIRECTORY}/cudaq")
-find_dependency(fmt REQUIRED HINTS "${PARENT_DIRECTORY}/cudaq")
+find_package(fmt QUIET)
+if (NOT fmt_FOUND)
+  include(FetchContent)
+  FetchContent_Declare(
+    fmt
+    GIT_REPOSITORY https://github.com/fmtlib/fmt
+    GIT_TAG        e69e5f977d458f2650bb346dadf2ad30c5320281) # 10.2.1
```
bmhowe23, line 22: This is different than our current submodule hash for `fmt`. Should we update the submodule one to match?

cmake/Modules/CUDAQConfig.cmake
```diff
@@ -44,11 +44,39 @@ if (NOT CUDAQ_LIBRARY_MODE)
...
+endfunction()
+
 add_library(cudaq::cudaq-builder SHARED IMPORTED)
 set_target_properties(cudaq::cudaq-builder PROPERTIES
   IMPORTED_LOCATION "${CUDAQ_LIBRARY_DIR}/libcudaq-builder${CMAKE_SHARED_LIBRARY_SUFFIX}"
-  IMPORTED_SONAME "libcudaq-builder${CMAKE_SHARED_LIBRARY_SUFFIX}")
\ No newline at end of file
+  IMPORTED_SONAME "libcudaq-builder${CMAKE_SHARED_LIBRARY_SUFFIX}")
+
+set(CUDAQ_TARGET "nvidia" CACHE STRING "The CUDA Quantum target to compile for and execute on. Defaults to `nvidia`")
```
bmhowe23, line 81: What happens if a system is using this but doesn't have a GPU that supports the `nvidia` target?

cmake/Modules/CUDAQConfig.cmake
```diff
@@ -44,11 +44,39 @@ if (NOT CUDAQ_LIBRARY_MODE)
   enable_language(CUDAQ)
 endif() 
 
+# ---- TARGET EXPORTS ----
+add_library(cudaq::cudaq-nvidia-target SHARED IMPORTED)
+set_target_properties(cudaq::cudaq-nvidia-target PROPERTIES
+  IMPORTED_LOCATION "${CUDAQ_LIBRARY_DIR}/libnvqir-custatevec-fp32${CMAKE_SHARED_LIBRARY_SUFFIX}"
```
bmhowe23, line 50: Do we need updates for the libnvqir-cusvim-* targets, too?

cmake/Modules/NVQIRConfig.cmake
amccaskey: Addressed in latest commit.

cmake/Modules/CUDAQConfig.cmake
amccaskey: Addressed in latest commit. Let me know if that check works for you

cmake/Modules/CUDAQConfig.cmake
amccaskey: Addressed in latest commit. Let me know what you think

cmake/Modules/CMakeLists.txt
```diff
@@ -22,6 +22,20 @@ set(LANG_FILES
   CMakeTestCUDAQCompiler.cmake
 )
 
+execute_process(
+  COMMAND git rev-parse HEAD
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tpls/fmt
+  OUTPUT_VARIABLE CUDAQ_FMTLIB_HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE
```
bmhowe23, line 29: I'm wondering if there is a way to do this that would work without the submodule being populated, i.e. perhaps something like `git submodule status tpls/fmt`. (Maybe your command already handles .... I don't really know.)

That being said, if it works in the CI, then it probably doesn't matter that much to me. So I'll kick off the CI.

cmake/Modules/CUDAQConfig.cmake
```diff
@@ -78,5 +78,45 @@ set_target_properties(cudaq::cudaq-builder PROPERTIES
...
+check_language(CUDA)
+if(CMAKE_CUDA_COMPILER)
+  enable_language(CUDA)
+  message(STATUS "CUDA compiler found: ${CMAKE_CUDA_COMPILER}")
+  set(CUDA_TEST_SOURCE "
+    #include <cuda_runtime.h>
+    #include <stdio.h>
+    int main() {
+      int deviceCount;
+      cudaError_t error = cudaGetDeviceCount(&deviceCount);
```
bmhowe23, line 139: Just curious ... why not use `nvidia-smi -L | wc -l` instead of this program?

cmake/Modules/CMakeLists.txt
amccaskey: The git submodules are always initialized on the first invocation of CMake. 

https://github.com/NVIDIA/cuda-quantum/blob/main/CMakeLists.txt#L110

cmake/Modules/CUDAQConfig.cmake
amccaskey: updated

bmhowe23 (approved):
LGTM. Thanks, Alex.

Issue #2002, opened 2024-07-24
[docs] Add user facing documentation for custom operations and unitary synthesis


khalatepradnya:
* CUDA-Q documentation must include the API, usage, examples, limitations for the custom operations feature.
* Part of issue #1475

PR #2016, opened 2024-07-27
[docs] Document the API for custom operations

khalatepradnya:
Towards issue #2002 

image

### DIFF
--- a/docs/sphinx/api/default_ops.rst
+++ b/docs/sphinx/api/default_ops.rst
@@ -615,85 +615,18 @@ operations, each operating on 2 qubits.
 
 .. tab:: Python
 
-    .. code-block:: python
-
-        import cudaq
-        import numpy as np
-        
-        # Create and test a custom CNOT operation.
-        cudaq.register_operation("my_cnot", np.array([1, 0, 0, 0, 
-                                                      0, 1, 0, 0, 
-                                                      0, 0, 0, 1, 
-                                                      0, 0, 1, 0]))
-
-
-        @cudaq.kernel
-        def bell_pair():
-            qubits = cudaq.qvector(2)
-            h(qubits[0])
-            my_cnot(qubits[0], qubits[1]) # my_cnot(control, target)
-
-
-        cudaq.sample(bell_pair).dump()
-
-        # Construct a custom unitary matrix for X on the first qubit and Y 
-        # on the second qubit.
-        X = np.array([[0,  1 ], [1 , 0]])
-        Y = np.array([[0, -1j], [1j, 0]])
-        XY = np.kron(X, Y)
-        XY_flat = np.ndarray.flatten(XY)
-
-        # Demonstrate that the flattened array is the same as the hard-coded one
-        # below (simply for demonstration purposes).
-        XY_flat_manual = np.array([0,   0,   0, -1j,
-                                   0,   0,  1j,   0,
-                                   0, -1j,   0,   0,
-                                   1j,  0,   0,   0])
-        assert np.array_equal(XY_flat, XY_flat_manual)
-
-        # Register the custom operation
-        cudaq.register_operation("my_XY", XY_flat_manual)
-
-        @cudaq.kernel
-        def custom_xy_test():
-            qubits = cudaq.qvector(2)
-            my_XY(qubits[0], qubits[1])
-            y(qubits[1]) # undo the prior Y gate on qubit 1
-
-
-        cudaq.sample(custom_xy_test).dump() # prints { 10:1000 }
+    .. literalinclude:: ../snippets/python/using/examples/two_qubit_custom_op.py
+      :language: python
+      :start-after: [Begin Docs]
+      :end-before: [End Docs]
 
 
 .. tab:: C++
 
-    .. code-block:: cpp
-
-            #include <cudaq.h>
-
-            CUDAQ_REGISTER_OPERATION(MyCNOT, 2, 0,
-                                     {1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0});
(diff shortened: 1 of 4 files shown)


docs/sphinx/api/default_ops.rst
```diff
@@ -541,4 +541,73 @@ This is a non-linear transformation, and no template overloads are available.
...
+        int main() {
+            auto counts = cudaq::sample(bell_pair);
+            for (auto &[bits, count] : counts) {
+                printf("%s\n", bits.data());
+            }
+        }
+
+.. note:: 
+
+  Custom operations are currently supported in simulation only.
```
efratshabtai, line 613: It might be worth mentioning the backends that are supported, so it's clear what simulation only means. Maybe we can also add what is the user experience when trying to use a custom operator on non-supported backends (error message?)

khalatepradnya: Addressed in d06bc6a7673bd326270377e3a48f19847fc40ce8

docs/sphinx/api/default_ops.rst
```diff
@@ -541,4 +541,74 @@ This is a non-linear transformation, and no template overloads are available.
     .. code-block:: cpp
 
         cudaq::qubit qubit;
-        my(qubit);
\ No newline at end of file
+        my(qubit);
+
+
+User-defined Custom Operations
```
bmhowe23, line 547: nit: I would vote for "defined" to be capitalized.
```suggestion
User-Defined Custom Operations
```

docs/sphinx/api/default_ops.rst
```diff
@@ -541,4 +541,74 @@ This is a non-linear transformation, and no template overloads are available.
...
+==============================
+
+Users can define a custom quantum operation by its unitary matrix. First use 
+the API to register a custom operation, outside of a CUDA-Q kernel. Then the 
+operation can be used within a CUDA-Q kernel like any of the built-in operations
+defined above.
+Custom operations are supported on qubits only (`qudit` with `level = 2`). For 
+multi-qubit operations, the matrix is interpreted with MSB qubit ordering.
```
bmhowe23, line 555: I'm not sure that everyone will interpret "MSB qubit ordering" the same way unless we provide a more explicit example. I think that in addition to showing what you have that also showing a more sophisticated example like a combined "XY" gate on 2 qubits (X on qubit 0 and Y on qubit 1) would be helpful. The result is not symmetrical, so it should help users get all of the ordering correct.

docs/sphinx/api/default_ops.rst
khalatepradnya: Addressed in cc43fff33919ba6d1aa98985efc8e71863f48e62

docs/sphinx/api/default_ops.rst
khalatepradnya: I added an example to define CNOT operation to clarify this. See cc43fff33919ba6d1aa98985efc8e71863f48e62

docs/sphinx/api/default_ops.rst
```diff
@@ -541,4 +541,123 @@ This is a non-linear transformation, and no template overloads are available.
...
+
+    .. code-block:: python
+
+        import cudaq
+        import numpy as np
+        
+        cudaq.register_operation("my_cnot", np.array([1, 0, 0, 0, 
+                                                      0, 1, 0, 0, 
+                                                      0, 0, 0, 1, 
+                                                      0, 0, 1, 0]))
```
bmhowe23, line 624: Maybe this should be a topic for a different PR, but why do we have to say that a custom operation like this requires controls (i.e. `my_cnot(control, target)`). It seems like a 4x4 matrix should have exactly 2 targets.

bmhowe23: In other words, since the 4x4 matrix was fully realized, it no longer needs to have "controls" specified in my opinion. I believe the only reason "controls" are distinctly specified from targets is that the underlying simulator can optimize the gate application by not reading or writing to half of the state vector (for 1 control). But if we are supplying a full 4x4 matrix, then I would imagine the underlying simulator has to read/write the full state vector.

So I would expect the invocation of `my_cnot` to be: `my_cnot(qubits[0:2])`.

khalatepradnya: `my_cnot(qubits[0:2])` - this should work, I called out control and target since it is a controlled-NOT gate. May be not the best example, but I chose one from standard gate set...

docs/sphinx/api/default_ops.rst
khalatepradnya: Perhaps we should add multiple examples? Can you help me construct this one to convey this...

bmhowe23: Yes, will do. It would be a good reason for me to experiment with it anyway 😄

bmhowe23: OK, here are some draft updates: https://github.com/NVIDIA/cuda-quantum/commit/3bae23e30a944784c8b7c62ecc56b6638d113e6f. If you like them, feel free to cherry pick that commit into your branch.

On a separate note, we should probably also put the code snippets into the `docs/sphinx/snippets` folder and reference them with `.. literalinclude` similar to how that is handled here: https://github.com/NVIDIA/cuda-quantum/blob/ad367c1bc562a11255fce11a0cecebe88a455d7b/docs/sphinx/using/quick_start.rst?plain=1#L78

The CI runs items in the snippets directory, so if the syntax ever changes, this will ensure that we keep the corresponding docs up to date.

khalatepradnya: Addressed in beecd2aa42f953df3c02b1f0ff2f0414af538343

bmhowe23 (approved):
This LGTM. Thanks, Pradnya!

PR #2020, opened 2024-07-29
[NFC] Add newlines to the end of any files missing them.

schweitzpgi:
This stops various tools from complaining about the newlines missing.


lib/Optimizer/CAPI/CMakeLists.txt
```diff
@@ -15,4 +15,4 @@ add_mlir_public_c_api_library(CUDAQuantumMLIRCAPI
   LINK_LIBS PRIVATE 
   QuakeDialect 
   CCDialect
-)
\ No newline at end of file
+)
```
khalatepradnya, line 18: Wonder if this can be enforced by a pre-commit hook?

khalatepradnya (approved):
LGTM. Thanks, Eric!

lib/Optimizer/CAPI/CMakeLists.txt
schweitzpgi: Seems like it could be. 🤷

PR #2021, opened 2024-07-29
Keeping overloaded functions for r-value reference

sacpis:
Addressing Eric's comment in the following PR
https://github.com/NVIDIA/cuda-quantum/pull/2008

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -5,7 +5,7 @@ header:
     software-name: CUDA Quantum
 
     content: |
-      Copyright (c) 2023 NVIDIA Corporation & Affiliates.
+      Copyright (c) 2023 - 2024 NVIDIA Corporation & Affiliates.
       All rights reserved.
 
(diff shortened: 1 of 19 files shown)


runtime/cudaq/cudaq.cpp
```diff
@@ -100,7 +100,7 @@ cudaq::MPIPlugin *getMpiPlugin(bool unsafe) {
   return g_plugin.get();
 };
 
-bool available() { return getMpiPlugin(/*unsafe=*/true) != nullptr; }
+bool available() { return static_cast<bool>(getMpiPlugin(/*unsafe=*/true)); }
```
schweitzpgi, line 103: You shouldn't need the `static_cast` here, and I don't think it helps that much in understanding the code.

schweitzpgi (commented):
Thanks, Sachin.

schweitzpgi (approved):

PR #2024, opened 2024-07-29
Refactor python kernel

sacpis:
Refactoring kernel_builder and kernel_decorator


bmhowe23:
Just a recommendation ... it would probably be best to hold off on requesting reviews until a PR passes the CI. Otherwise, people will likely have to re-review it once additional changes are applied.

PR #2026, opened 2024-07-30
Synthesize state pointers for remote simulation

annagrin:
### Description
Synthesize state pointers for remote simulation


annagrin:
Replaced by https://github.com/NVIDIA/cuda-quantum/pull/2035

Issue #1911, opened 2024-07-10
[refactoring] Move constants into a common header file


khalatepradnya:
* Hardcoded constants, including strings should be moved to a common header file.
* See discussion here -

   

> This is exactly the reason why these strings should be in a header file and not just hard coded in random files. The latter is bad coding practice.
> 
> I thought the runtime already included files from the compiler, but maybe I am wrong about that.
> 
> Open to any suggestions on where to put universal constants, including string data.

_Originally posted by @schweitzpgi in https://github.com/NVIDIA/cuda-quantum/pull/1794#discussion_r1672470040_

ParamThakkar123:
Is anyone working on this one??

ParamThakkar123:
If not, I can take this

khalatepradnya:
> If not, I can take this

Thank you. I will assign this to you.

ParamThakkar123:
for which files this is to be done ??

khalatepradnya:
> for which files this is yet to be done ??

Start with the constant strings defined in `include/cudaq/Frontend/nvqpp/AttributeNames.h` that are duplicated in `runtime/cudaq/qis/qubit_qis.h`.
Refer to the link to comment in description for context.

PR #2028, opened 2024-07-30
Added definitions from header file

ParamThakkar123:
### Description

Addresses issue #1911


ParamThakkar123:

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -19,7 +19,7 @@
 #include <cstring>
 #include <functional>
 
-#define __qpu__ __attribute__((annotate("quantum")))
+#define __qpu__ __attribute__((annotate(cudaq::kernelAnnotation)))
```
khalatepradnya, line 22: This is defined in 'Frontend' directory, hence build error `error: no member named 'kernelAnnotation' in namespace 'cudaq'`.
Need to refactor the constant strings in a common file. See discussion https://github.com/NVIDIA/cuda-quantum/pull/1794#discussion_r1672470040

ParamThakkar123: ok. got it. I will do that.

ParamThakkar123:
image

@khalatepradnya this doesn't give and error in other files but this some thing gives an error in qubit_qis.h . Why is this happening ?

ParamThakkar123:
@khalatepradnya can you please explain me what's the problem here ?? So I can try to fix this

khalatepradnya:
> image
> 
> @khalatepradnya this doesn't give and error in other files but this some thing gives an error in qubit_qis.h . Why is this happening ?

Is this the error message you get when building your branch locally? 
(I see that CI is failing with `error: 'annotate' attribute requires a string`)

khalatepradnya:
> @khalatepradnya can you please explain me what's the problem here ?? So I can try to fix this

I don't see any changes to cmake in this PR. I suspect the build is not able to locate the new file, can you check `target_include_directories`?

ParamThakkar123:
> > @khalatepradnya can you please explain me what's the problem here ?? So I can try to fix this
> 
> I don't see any changes to cmake in this PR. I suspect the build is not able to locate the new file, can you check `target_include_directories`?

Ok I will try this

schweitzpgi:
> > image
> > @khalatepradnya this doesn't give and error in other files but this some thing gives an error in qubit_qis.h . Why is this happening ?
> 
> Is this the error message you get when building your branch locally? (I see that CI is failing with `error: 'annotate' attribute requires a string`)

Unfortunately, `__attribute__((annotate(thing)))` isn't very smart and strictly requires a C style string literal to appear inline for `thing`.  Sad, but true. So we can't use a symbolic for `thing` in these cases.

schweitzpgi:
@ParamThakkar123  do you want to take another stab at this one?

ParamThakkar123:
> @ParamThakkar123 do you want to take another stab at this one?

Yeah would like to work on this. But would need some help from your side not to solve this issue but to get a better understanding of this codebase.

khalatepradnya:
> > @ParamThakkar123 do you want to take another stab at this one?
> 
> Yeah would like to work on this. But would need some help from your side not to solve this issue but to get a better understanding of this codebase.

Hi! If you want to continue investigating this build failure, I recommend first aligning with `main` and re-running the build. If you happen to have error message from your local build environment, please share it so we can debug further. Thanks!

PR #2030, opened 2024-07-30
Config file update legacy nvidia targets

1tnguyen:
### Description

- Enable the new custatevec-based simulators for all legacy targets (`nvidia-fp64`, `nvidia-mqpu`, and `nvidia-mqpu-fp32`) as the first option (to be taken if available).

- Update the config file regression checks.

- Fixed a fallout from https://github.com/NVIDIA/cuda-quantum/pull/1973, these `.config` files were not run in the new location unintentionally.


bmhowe23 (approved):
LGTM

PR #2031, opened 2024-07-30
[core] Implement new argument synthesis pass.

schweitzpgi:
This pass performs function specialization by replacing 1 or more arguments to a function with custom code blocks, which would be constant expressions in the typical application. This functionality is an alternative to some of the functionality in the current quake synthesis pass. Some key differences:

  - It will support substitutions of any argument type.
  - It can be used to replace any permutation of arguments.
  - Substitutions must be type correct.
  - Can be tested from the command-line in full generality.

Add a test that demonstrates the command-line option in action.


amccaskey:
This seems like a really powerful approach - the ability to put arbitrary code in the `arg_subst` body. How do you envision these `cc.arg_subst` operations being created at runtime? Is there also a need for global data to be defined? (I'm thinking about some of the state synthesis stuff).

schweitzpgi:
> This seems like a really powerful approach - the ability to put arbitrary code in the `arg_subst` body. How do you envision these `cc.arg_subst` operations being created at runtime? Is there also a need for global data to be defined? (I'm thinking about some of the state synthesis stuff).

Well, I have code for the first question in my next PR. heh  Basically, I've implemented argument conversion in the runtime library in a general way that includes all our supported types (recursively). (Not `cudaq::state` yet. Sorry, Anna.)

Yes, we need and will have constant global information.

annagrin (approved):
LGTM, thanks Eric!

PR #2035, opened 2024-07-31
Support cudaq::state argument synthesis for remote simulators

annagrin:
### Description
Support `cudaq::state*` argument synthesis for remote simulators

#### Details
- Add a `ArgumentDataStore` class that stores extra information about the arguments
  - number of elements + size + data pointer per each argument stored
- Create and pass the argument store to the `QuakeSynthesis` pass
- Update quake synthesizer to
  - synthesize data vector corresponding to a `state*` argument from the argument store
  - update `quake.init_state` instruction to use the vector instead of the state
  - const prop calls to `getNumQubits` runtime call
- Add tests for remote simulator scenarios
- Add negative tests for local quantinuum emulator

#### Notes  
- Added a helper to get argument offsets to `ConvertToQIR.cpp` (`OptCodeGen` library)
- Added a helper to read simulation state data to `nvqir` library
- Added library dependency on `nvqir` library to python extension library so it can use the above helper.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/1931
Towards: https://github.com/NVIDIA/cuda-quantum/issues/1107
Needs: https://github.com/NVIDIA/cuda-quantum/pull/2027


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -511,18 +529,25 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
...
-  // Simulators have direct implementation of state initialization
-  // in their runtime.
   auto &platform = cudaq::get_platform();
-  if (!platform.is_simulator() || platform.is_emulated()) {
-    pm.addPass(cudaq::opt::createConstPropComplex());
-    pm.addPass(cudaq::opt::createLiftArrayAlloc());
-    pm.addPass(cudaq::opt::createStatePreparation());
-  }
+  auto isRemoteSimulator =
+      platform.get_remote_capabilities().serializedCodeExec;
```
bmhowe23, line 540: NVQC will have this set to `false`, but I believe you would want `isRemoteSimulator` set to true in that case, no?

targettests/execution/state_init_err_runtime.cpp
```diff
@@ -9,8 +9,6 @@
 // clang-format off
 // Note: change |& to 2>&1| if running in bash
 // RUN: nvq++ %cpp_std %s -o %t --target quantinuum --emulate && %t |& FileCheck %s
-// Note: change |& to 2>&1| if running in bash
-// RUN: nvq++ %cpp_std --enable-mlir --target remote-mqpu --remote-mqpu-url localhost:9999 %s -o %t && %t |& FileCheck %s
```
bmhowe23, line 13: If this test is being removed, do we need to update the wording for the error message on line 34 to remove the reference to remote simulators not being able to synthesize it anymore?

In other words, perhaps the error should be updated to say `op synthesis: unsupported argument type for remote devices: state*`

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -204,6 +204,37 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
   cudaq::opt::populateCCTypeConversions(&typeConverter);
 }
 
+std::pair<std::size_t, std::vector<std::size_t>>
```
bmhowe23, line 207: It might be nice to have a Doxygen comment about what exactly is being returned here.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+template <typename T>
+LogicalResult synthesizeStateArgument(OpBuilder &builder, ModuleOp module,
+                                      unsigned &counter, BlockArgument argument,
+                                      Type eleTy, std::vector<T> &vec) {
+  auto *ctx = builder.getContext();
+  auto argLoc = argument.getLoc();
+  auto arrTy = cudaq::cc::ArrayType::get(ctx, eleTy, vec.size());
+
+  builder.setInsertionPointToStart(argument.getOwner());
+  auto toErase = std::vector<mlir::Operation *>();
```
bmhowe23, line 145: Nit: since this is compiler code, I think @schweitzpgi typically prefers MLIR data structures like `SmallVector` instead of `std::vector` (at least he has in the past ... not sure if that still holds).

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -561,12 +632,21 @@ class QuakeSynthesizer
...
+                                               argument, *stateData,
+                                               stateDataOffset)))
+              funcOp.emitOpError("synthesis failed for state*");
+            stateDataOffset++;
           } else {
-            funcOp.emitOpError("synthesis: unsupported argument type for "
-                               "remote devices and simulators: state*");
+            // All other cases are not yet supported (i.e. quantum hardware).
+            funcOp.emitOpError("synthesis: unsupported argument type on "
+                               "quantum devices: state*");
```
bmhowe23, line 646: Oh, it looks like you already updated the error message, 👍 . (Though I'm not sure how that other test was passing if the error message changed?)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Yes, I would need `isRemoteSimulator` to be true for nvqc as well, is there a better way to detect remote simulators?

targettests/execution/state_init_err_runtime.cpp
annagrin: yes, thanks!

lib/Optimizer/CodeGen/ConvertToQIR.cpp
annagrin: will do

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
bmhowe23: I'm open to other ideas, but perhaps it would be better to add a new struct field to `RemoteCapabilities` in `runtime/cudaq/remote_capabilities.h`. You can then update `getRemoteCapabilities()` in `runtime/common/BaseRestRemoteClient.h` to set it to true similar to how we set `vqe` and `stateOverlap`?

That would be functionally equivalent to simply using the `vqe` or `stateOverlap` field, but it would be a bit more explicit.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
annagrin: it is not used in the IR though... There are other opts that do the same.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
annagrin: That test was failing, fixed

lib/Optimizer/CodeGen/ConvertToQIR.cpp
annagrin: done!

targettests/execution/state_init_err_runtime.cpp
annagrin: fixed:)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Added an `isRemoteSimulator` field in the remote capabilities, which I set to true for remote-mqpu and nvqc

annagrin (commented):
Thanks @bmhowe23 for the review, I addressed the comments.

include/cudaq/Optimizer/Transforms/ArgumentDataStore.h
```diff
@@ -0,0 +1,53 @@
+/*******************************************************************************
```
1tnguyen, line 1: nit-pick: we often add C++ here for `.h` files. 
```suggestion
/****************************************************************-*- C++ -*-****
```

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
   return false;
 }
 
+template <typename T>
+Value createGlobalArray(OpBuilder &builder, ModuleOp module, unsigned &counter,
+                        BlockArgument argument, Type arrTy,
+                        std::vector<T> vec) {
```
1tnguyen, line 121: Could we use `const&` here?
```suggestion
                        const std::vector<T> &vec) {
```

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+  auto *ctx = builder.getContext();
+  auto argLoc = argument.getLoc();
+  auto arrTy = cudaq::cc::ArrayType::get(ctx, eleTy, vec.size());
+
+  builder.setInsertionPointToStart(argument.getOwner());
+  auto toErase = std::vector<mlir::Operation *>();
+
+  // Iterate over the users of this state argument.
+  for (auto *argUser : argument.getUsers()) {
+    // Replace a calls to runtime function that reads the number of qubits
```
1tnguyen, line 149:
```suggestion
    // Replace a call to runtime function that reads the number of qubits
```

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+template <typename T>
+static LogicalResult synthesizeStateArgument(OpBuilder &builder,
+                                             ModuleOp module, unsigned &counter,
+                                             BlockArgument argument, Type eleTy,
+                                             void *data, std::size_t size) {
+  std::vector<T> vec;
+  for (std::size_t i = 0; i < size; i++) {
+    auto elePtr = reinterpret_cast<T *>(data) + i;
+    vec.push_back(*elePtr);
+  }
```
1tnguyen, line 190: Perhaps this is more efficient?
```suggestion
  auto *castData =  reinterpret_cast<T *>(data);
  std::vector<T> vec(castData, castData + size);
```

runtime/common/SimulationStateData.h
```diff
@@ -0,0 +1,24 @@
+/*******************************************************************************
```
1tnguyen, line 1: same nit-pick 
```suggestion
/****************************************************************-*- C++ -*-****
```

bmhowe23 (commented):
This LGTM. I believe some of the other synthesis code is getting rewritten, so it would be good for @schweitzpgi to review this as well.

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
```diff
@@ -103,4 +103,8 @@ inline mlir::Type getCharPointerType(mlir::MLIRContext *context) {
 
 void initializeTypeConversions(mlir::LLVMTypeConverter &typeConverter);
 
+std::pair<std::size_t, std::vector<std::size_t>>
```
schweitzpgi, line 106: I think this needs some documentation here in the header file. I'm not sure what it does just from the name and signature.

Is this a QIR defined function?

include/cudaq/Optimizer/Transforms/ArgumentDataStore.h
schweitzpgi: Often = always. 😄

include/cudaq/Optimizer/Transforms/ArgumentDataStore.h
```diff
@@ -0,0 +1,53 @@
...
+#include <cassert>
+#include <numbers>
+#include <tuple>
+#include <vector>
+
+namespace cudaq::opt {
+
+/// Stores argument values.
+/// Stores data, element size, and number of elements for each argument.
+/// If the data pointer is owned, cleans up the data on destruction.
```
schweitzpgi, line 20: Is this here as some sort of wrapper around all argument pointers? I'm not sure why it is needed though. The only argument type that seems relevant to this PR is a `cudaq::state *`.

This seems fully redundant with the new argument synthesis approach, unfortunately.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -204,6 +204,43 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
   cudaq::opt::populateCCTypeConversions(&typeConverter);
 }
 
+/// @brief Compute function argument size in bytes and offsets
+/// of all the arguments that satisfy the filter.
+/// Start computing offsets and size from `startingArgIdx`.
+std::pair<std::size_t, std::vector<std::size_t>>
+cudaq::opt::getFunctionArgumentLayout(mlir::ModuleOp module,
```
schweitzpgi, line 211:
```suggestion
cudaq::opt::getFunctionArgumentLayout(ModuleOp module,
```
I believe this file has a `using namespace mlir;` near the top. So please delete all the `mlir::` namespace prefixes for consistency with the rest of the code.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -204,6 +204,43 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
...
+                                      mlir::func::FuncOp func,
+                                      bool filter(mlir::Type arg),
+                                      std::size_t startingArgIdx) {
+  auto arguments = func.getArguments();
+  auto results = func.getResultTypes();
+  auto funcTy = func.getFunctionType();
+  auto bufferTy =
+      cudaq::opt::factory::buildInvokeStructType(funcTy, startingArgIdx);
+  mlir::StringRef dataLayoutSpec = "";
```
schweitzpgi, line 220:
```suggestion
  StringRef dataLayoutSpec;
```
I believe that creates an empty StringRef.

lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -204,6 +204,43 @@ void cudaq::opt::initializeTypeConversions(LLVMTypeConverter &typeConverter) {
...
+  llvm::DataLayout dataLayout{dataLayoutSpec};
+  // Convert bufferTy to llvm.
+  llvm::LLVMContext context;
+  mlir::LLVMTypeConverter converter(funcTy.getContext());
+  cudaq::opt::initializeTypeConversions(converter);
+  auto llvmDialectTy = converter.convertType(bufferTy);
+  mlir::LLVM::TypeToLLVMIRTranslator translator(context);
+  auto *llvmStructTy =
+      cast<llvm::StructType>(translator.translateType(llvmDialectTy));
+  auto *layout = dataLayout.getStructLayout(llvmStructTy);
```
schweitzpgi, line 233: In one of my PRs, I've already moved this code to a common function for general use.

lib/Optimizer/CodeGen/VerifyNVQIRCalls.cpp
```diff
@@ -43,7 +43,9 @@ struct VerifyNVQIRCallOpsPass
           cudaq::opt::NVQIRInvokeRotationWithControlBits,
           cudaq::opt::NVQIRInvokeWithControlRegisterOrBits,
           cudaq::opt::NVQIRPackSingleQubitInArray,
-          cudaq::opt::NVQIRReleasePackedQubitArray};
+          cudaq::opt::NVQIRReleasePackedQubitArray,
+          cudaq::getNumQubitsFromCudaqState,
```
schweitzpgi, line 47:
```suggestion
          cudaq::getNumQubitsFromCudaqState
```

lib/Optimizer/CodeGen/VerifyNVQIRCalls.cpp
```diff
@@ -71,8 +73,8 @@ struct VerifyNVQIRCallOpsPass
         passFailed = true;
         return WalkResult::interrupt();
       } else if (!isa<LLVM::AddressOfOp, LLVM::AllocaOp, LLVM::BitcastOp,
-                      LLVM::ExtractValueOp, LLVM::GEPOp, LLVM::InsertValueOp,
-                      LLVM::LoadOp, LLVM::StoreOp>(op)) {
+                      LLVM::ExtractValueOp, LLVM::GEPOp, LLVM::IntToPtrOp,
```
schweitzpgi, line 76: @bmhowe23 Is this legit per the QIR spec?

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -966,6 +966,16 @@ struct FuseWithConstantArray
 
           return success();
         }
+        if (auto complexTy = dyn_cast<ComplexType>(extval.getType())) {
+          std::int32_t i = extval.getRawConstantIndices()[0];
+          auto rePart = cast<FloatAttr>(conarr.getConstantValues()[2 * i]);
+          auto imPart = cast<FloatAttr>(conarr.getConstantValues()[2 * i + 1]);
```
schweitzpgi, line 972: I think the default checks make sure the index is within the range of the size of the array. So that check may not be good enough when accessing `2 * i`. An additional check here would be needed in that case. We shouldn't make any assumptions here and risk crashing with a SIGSEGV.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+                        BlockArgument argument, Type arrTy,
+                        std::vector<T> vec) {
+  OpBuilder::InsertionGuard guard(builder);
+  auto argLoc = argument.getLoc();
+
+  // Stick global at end of Module.
+  std::string symbol = "__nvqpp_rodata_init_state." + std::to_string(counter++);
+
+  cudaq::IRBuilder irBuilder(builder);
+  irBuilder.genVectorOfConstants(argLoc, module, symbol, vec);
```
schweitzpgi, line 129: What happens if this is called more than once for the same vector of init state values?

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+  // Iterate over the users of this state argument.
+  for (auto *argUser : argument.getUsers()) {
+    // Replace a calls to runtime function that reads the number of qubits
+    // with the log of the length, which is a synthesized constant.
+    if (auto numOfQubitsOp = dyn_cast<func::CallOp>(argUser)) {
+      if (auto calleeAttr = numOfQubitsOp.getCalleeAttr()) {
+        auto funcName = calleeAttr.getValue().str();
+        if (funcName == cudaq::getNumQubitsFromCudaqState) {
+          Value numOfQubits = builder.create<arith::ConstantIntOp>(
+              argLoc, log2(vec.size()), builder.getI64Type());
```
schweitzpgi, line 156: Can we use count trailing zeros here for (log base 2) instead of a (possible) floating point log function?

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+          numOfQubitsOp.replaceAllUsesWith(ValueRange{numOfQubits});
+          toErase.push_back(numOfQubitsOp);
```
schweitzpgi, line 158: Maintaining lists of dead code in passes just clutters up the code and can lend itself to bugs. Just use `repalceOpWithNewOp` when you can and forget about dead ops otherwise. There is a general DCE framework in MLIR that will do it for us.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+  }
+  return synthesizeStateArgument(builder, module, counter, argument, eleTy,
+                                 vec);
+}
+
+static LogicalResult
+synthesizeStateArgument(OpBuilder &builder, ModuleOp module, unsigned &counter,
+                        BlockArgument argument,
+                        const cudaq::opt::ArgumentDataStore &stateData,
+                        std::size_t idx) {
```
schweitzpgi, line 199: I thought this was all going to be done in a separate pass and not in QuakeSynthesis.

The synthesis of gates from the initial state value can and should be a distinct thing from argument synthesis. They are very different transformations in the IR. Having them separated will make it easy to use them and debug the passes when problems pop up.

It's super unfortunate that all these different transformation got labelled "synthesis". That may be the correct terminology in the literature, but it is rife with confusion and seems to suggest monolithic design instead of modularity perhaps.

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -115,6 +115,103 @@ static bool hasInitStateUse(BlockArgument argument) {
...
+  auto [data, size, elementSize] = stateData.getData(idx);
+
+  if (elementSize == sizeof(std::complex<double>)) {
+    return synthesizeStateArgument<std::complex<double>>(
+        builder, module, counter, argument,
+        ComplexType::get(builder.getF64Type()), data, size);
+  } else if (elementSize == sizeof(std::complex<float>)) {
```
schweitzpgi, line 206: else after return. - don't do this according to the LLVM coding guidelines. https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

lib/Optimizer/Transforms/QuakeSynthesizer.cpp
```diff
@@ -561,12 +632,21 @@ class QuakeSynthesizer
...
+            // space, when we have converted state data.
+            if (failed(synthesizeStateArgument(builder, module, counter,
+                                               argument, *stateData,
+                                               stateDataOffset)))
+              funcOp.emitOpError("synthesis failed for state*");
+            stateDataOffset++;
           } else {
-            funcOp.emitOpError("synthesis: unsupported argument type for "
-                               "remote devices and simulators: state*");
+            // All other cases are not yet supported (i.e. quantum hardware).
```
schweitzpgi, line 644: I don't think we can crash out of the compile if we see a pointer. I'm kind of surprised the tests didn't catch this.

There are perfectly legit cases where an argument to the kernel is a pointer and it is ok. It happens when the pointer isn't really required because the argument was already dealt with in another way.

Strange though that this didn't fail in the tests.

unittests/Optimizer/QuakeSynthTester.cpp
```diff
@@ -424,6 +451,62 @@ TEST(QuakeSynthTests, checkVectorOfPauliWord) {
...
+  auto [args, offset] = cudaq::mapToRawArgs(colonel.name(), initialState);
+
+  EXPECT_TRUE(succeeded(runQuakeSynth(colonel.name(), args, module, true)));
+
+  auto func = module->lookupSymbol<func::FuncOp>(colonelName);
+  EXPECT_TRUE(func);
+  EXPECT_TRUE(func.getArguments().empty());
+  func.dump();
+}
+
```
schweitzpgi, line 509: I think I missed where the gate creation is happening... argh!

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
annagrin: Will add, thanks! This is a helper to get argument struct size and argument offsets from a func op - we currently use it in several places (quake synthesis tests, python alt_launch_kernel, quake synthesis, BaseRestRemoteClient.h). Let me know if there is a better place for it!

include/cudaq/Optimizer/Transforms/ArgumentDataStore.h
annagrin: Yes this is just to store state data from the state* arguments. Let me know if you would like this to do in before the new approach is finished - I can definitely remove it later when not needed anymore.

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
annagrin: done!

annagrin:
Replaced by https://github.com/NVIDIA/cuda-quantum/pull/2094

PR #2038, opened 2024-08-01
Add unintentionally excluded tests to Publishing workflow

bmhowe23:
Some tests have been disabled from our Publishing workflow, essentially because the `validate_wheel.sh` script uses the `find` command to find tests, and that is apparently not following the symlinks that our publishing workflow had setup. Specifically, it was missing the `snippets` and `examples` folders when doing the wheel validation.

Note: I expect this to reveal the following error:
```bash
(cuda-quantum) root@682642a99424:/cuda-quantum# python3 /tmp/snippets/using/cudaq/platform/sample_async_remote.py
Number of virtual QPUs: 2
Sampling jobs launched for asynchronous processing.
terminate called recursively
  what():  Failed to launch kernel. Error: Failed to execute the kernel on the remote server: "Failed to process incoming request"
Error message: "Failed to open simulator backend library: libcutensornet.so.2: cannot open shared object file: No such file or directory."

Aborted (core dumped)
```


bettinaheim (approved):
Thanks for finding this.

PR #2039, opened 2024-08-02
Update Crow to v1.2.0 and remove customization

bmhowe23:
Update to https://github.com/CrowCpp/Crow/releases/tag/v1.2.0, which includes https://github.com/CrowCpp/Crow/pull/784, so we no longer need our customization.


schweitzpgi (commented):
Shouldn't this just be updating the SHA of the Crow submodule? GitHub is presenting file changes here, which is unexpected.

bmhowe23:
> Shouldn't this just be updating the SHA of the Crow submodule? GitHub is presenting file changes here, which is unexpected.

I thought that looked weird as well. I think it might just be a new "feature" for how GitHub is showing diffs for submodules. See https://patch-diff.githubusercontent.com/raw/NVIDIA/cuda-quantum/pull/2039.diff for proof that the submodule SHA is indeed the only "real" update happening there (ignoring the other intentional changes to our build scripts).

schweitzpgi:
> > Shouldn't this just be updating the SHA of the Crow submodule? GitHub is presenting file changes here, which is unexpected.
> 
> I thought that looked weird as well. I think it might just be a new "feature" for how GitHub is showing diffs for submodules. See https://patch-diff.githubusercontent.com/raw/NVIDIA/cuda-quantum/pull/2039.diff for proof that the submodule SHA is indeed the only "real" update happening there (ignoring the other intentional changes to our build scripts).

New feature to baffle old-timers. heh

Thanks, Ben.

PR #2040, opened 2024-08-02
[runtime] Implements argument conversions as a runtime library.

schweitzpgi:
These changes add a new `ArgumentConverter` class which will be used to translate argument values to MLIR code. The code is meant to be used as a library by the runtime. Argument data will arrive as raw pointers, so this uses auxiliary type information to decode the arguments and translate them into constants as MLIR code that are supported and used on the device side.

~Requires #2031 be merged first!~


runtime/common/ArgumentConversion.cpp
```diff
@@ -0,0 +1,360 @@
...
+  auto cast = builder.create<cudaq::cc::CastOp>(loc, i8PtrTy, addr);
+  auto size = builder.create<arith::ConstantIntOp>(loc, v.size(), 64);
+  auto chSpanTy = cudaq::cc::CharspanType::get(builder.getContext());
+  return builder.create<cudaq::cc::StdvecInitOp>(loc, chSpanTy, cast, size);
+}
+
+static Value genConstant(OpBuilder &builder, const cudaq::state *v,
+                         ModuleOp substMod) {
+  // TODO: do we materialize the data here or just add a symbolic reference into
+  // the unknown?
```
1tnguyen, line 89: Should we have a `TODO` exception here so that it might be easier to spot it in the future?

runtime/common/ArgumentConversion.h
```diff
@@ -0,0 +1,51 @@
...
+  /// created.
+  mlir::ModuleOp getSubstitutionModule() { return substModule; }
+
+private:
+  mlir::ModuleOp sourceModule;
+  mlir::ModuleOp substModule;
+  mlir::OpBuilder builder;
+  mlir::StringRef kernelName;
+
+  mlir::SmallVector<cc::ArgumentSubstitutionOp> substitutions;
```
1tnguyen, line 48: Is the extra line intentional?
```suggestion
  mlir::StringRef kernelName;
  mlir::SmallVector<cc::ArgumentSubstitutionOp> substitutions;
```

1tnguyen (approved):
This LGTM 👍

PR #2042, opened 2024-08-02
Add cudaq::draw support for hardware targets

bmhowe23:
Drawing circuits that were targeting hardware-specific targets previously just returned an empty string. This change makes it behave like other draw invocations, which is nice for seeing the effects of hardware-specific passes like mapping, qubit management, and multi-control decomposition.


khalatepradnya (commented):
Thanks, Ben! These changes look good. Can you add one Python test as well?

bmhowe23:
> Thanks, Ben! These changes look good. Can you add one Python test as well?

Sure! Added in f18aa0b81.

khalatepradnya (approved):
👍🏽

PR #2043, opened 2024-08-02
[core] Modify GKE to generate code to be 

schweitzpgi:
used exclusively with argument conversion and argument synthesis.

This is the initial implementation to argument gathering on the C++ side. It changes the autogenerated C++ host entry point to gather the arguments that are passed in and pass them as a `std::vector<void*>` to the runtime for processing. The vector includes only textual arguments for now. All hidden arguments are ignored.

~#2040 must be committed first.~


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1392,6 +1393,91 @@ class GenerateKernelExecution
...
+    Value loadKernName = builder.create<LLVM::AddressOfOp>(
+        loc, cudaq::opt::factory::getPointerType(kernelNameObj.getType()),
+        kernelNameObj.getSymName());
+    auto castKernelNameObj =
+        builder.create<cudaq::cc::CastOp>(loc, i8PtrTy, loadKernName);
+    builder.create<func::CallOp>(
+        loc, std::nullopt, cudaq::runtime::launchKernelVersion2FuncName,
+        ArrayRef<Value>{castKernelNameObj, castStdvec, castResultBuffer});
+
+    // FIXME: Drop any results on the floor for now and return random data left
```
annagrin, line 1469: I thought we support returning values from kernels in c++ and python now?

schweitzpgi: I'll add this to the new flow in a later PR.

schweitzpgi: As to the question, it's nuanced.

Does the kernel launch have the _capability_ to handle return values? Yes. Is that capability fully functional and correct? Almost completely, if you consider C++. Do any of the kernel launchers deal with return values? Almost never. There is a singular path that can return a `int`, `double` and `bool` (only when remote simulation and other qualifiers IIRC).

1tnguyen (approved):
LGTM 👍

PR #2047, opened 2024-08-05
Update factor-quantum-alloc to handle subveqs better

bmhowe23:
This change updates the `factor-quantum-alloc` pass to understand the `quake.subveq` operation with constant operands.


lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +42,53 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+          rewriter.setInsertionPoint(dealloc);
+          auto deloc = dealloc.getLoc();
+          for (std::size_t i = 0; i < size - 1; ++i)
+            rewriter.create<quake::DeallocOp>(deloc, newAllocs[i]);
+          rewriter.replaceOpWithNewOp<quake::DeallocOp>(dealloc,
+                                                        newAllocs[size - 1]);
+          continue;
+        }
+        if (auto subveq = dyn_cast<quake::SubVeqOp>(user)) {
+          std::int64_t lowInt = -1;
```
schweitzpgi, line 59:
```suggestion
          std::optional<std::int64_t> lowInt = std::nullopt;
```

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +42,53 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+          if (auto low =
+                  dyn_cast<arith::ConstantOp>(subveq.getLow().getDefiningOp()))
+            if (auto myInt = dyn_cast<IntegerAttr>(low.getValue()))
+              lowInt = start + myInt.getInt();
+          if (lowInt < 0) {
```
schweitzpgi, line 64:
```suggestion
          if (!lowInt) {
```

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +42,53 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+            subveq.emitOpError("Invalid subveq operator found");
```
schweitzpgi, line 65: This pass should not fail with hard errors here. A subveq operation need not have constant bounds and this pass should not be forced to only run when there are constant bounds.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +42,53 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+            return failure();
+          }
+          for (auto *subUser : subveq->getUsers())
+            if (failed(rewriteOpAndUsers(subUser, lowInt)))
```
schweitzpgi, line 69:
```suggestion
            if (failed(rewriteOpAndUsers(subUser, *lowInt)))
```

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi, line 65:
```suggestion
```

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +42,53 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+          rewriter.replaceOpWithNewOp<quake::DeallocOp>(dealloc,
+                                                        newAllocs[size - 1]);
+          continue;
+        }
+        if (auto subveq = dyn_cast<quake::SubVeqOp>(user)) {
+          std::int64_t lowInt = -1;
+          if (auto low =
+                  dyn_cast<arith::ConstantOp>(subveq.getLow().getDefiningOp()))
+            if (auto myInt = dyn_cast<IntegerAttr>(low.getValue()))
+              lowInt = start + myInt.getInt();
```
schweitzpgi, line 63: Actually, should be using `cudaq::opt:;factory::getIntIfConstant()` here.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: The changes in the `runAnalysis` function have checks to ensure that it has constant bounds. If it does not have constant bounds, the original alloca op will not be added to the list of ops that needs to be rewritten. This is a check to ensure that the op hasn't snuck into the `matchAndRewrite` function after the analysis should've silently ignored that operation. This check should not fail under normal operation, unless somebody updates the `runAnalysis` function in the future without realizing they might break downstream code; this is a guard against that scenario.

bmhowe23: That being said - I can still remove the `emitOpError` call.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Updated in e9a614cec (per suggestion from subsequent comment)

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Updated in e9a614cec

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Updated in e9a614cec

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Updated in e9a614cec

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Thanks. I updated it to use the `getIntIfConstant` method both here and in the `runAnalysis` function (e9a614cec).

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: Updated in e9a614cec

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -42,21 +43,47 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
-    for (auto *user : allocOp->getUsers()) {
-      if (auto dealloc = dyn_cast<quake::DeallocOp>(user)) {
-        rewriter.setInsertionPoint(dealloc);
-        auto deloc = dealloc.getLoc();
-        for (std::size_t i = 0; i < size - 1; ++i)
-          rewriter.create<quake::DeallocOp>(deloc, newAllocs[i]);
-        rewriter.replaceOpWithNewOp<quake::DeallocOp>(dealloc,
-                                                      newAllocs[size - 1]);
-        continue;
+    std::function<LogicalResult(Operation *, std::int64_t)> rewriteOpAndUsers =
```
schweitzpgi, line 46:
```suggestion
    auto rewriteOpAndUsers =
```

schweitzpgi: We normally let the compiler type a lambda expression.

schweitzpgi (approved):
LGTM

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
bmhowe23: > We normally let the compiler type a lambda expression.

Agree, but that won't work in this case because this lambda can call itself inside the lambda, and using `auto` returns this compilation error: `a variable declared with an auto type specifier cannot appear in its own initializer`.

PR #2050, opened 2024-08-06
copy edits on divisive clustering tutorial

mmvandieren:
### Description

One LaTeX command had an extra space which prevented rendering on Safari browser.  Several typos were corrected.


docs/sphinx/examples/python/tutorials/Divisive_clustering.ipynb
```diff
@@ -571,7 +571,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The data can be nicely visualized with a Dendrogram which maps where the bipartitionings occurred. Early splits are generally mark divisions between the least like data."
+    "The data can be nicely visualized with a Dendrogram which maps where the bipartitionings occurred. Early splits generally mark divisions between the least alike data."
```
schweitzpgi, line 574: Seems like we could use a more mathematical term than "alike" here. (I don't have a good suggestion.)

schweitzpgi (approved):
Thanks for the copy edits.

docs/sphinx/examples/python/tutorials/Divisive_clustering.ipynb
mmvandieren: @mawolf2023 any suggestions?  would "similar" be ok?

mawolf2023: @mmvandieren, yes, that is perfectly fine.  That sentence was a mess so thanks for catching.

mmvandieren: since the plotting is part of finding the clusters, it is a little circular to refer to the clusters before they are established.

PR #2054, opened 2024-08-06
Fixing cmake.version error by setting scikit-build-core==0.9.10

sacpis:
Fixing cmake.version error by setting scikit-build-core==0.9.10

Fixes the following CI error
```
ERROR: Cannot set both cmake.minimum_version and cmake.version; use version only for scikit-build-core >= 0.8.
```


bmhowe23 (approved):
LGTM, assuming the rest of the CI passes without issues. Thanks!

Issue #1937, opened 2024-07-15
[RFC] Python frontend for 'photonics' target


khalatepradnya:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

* Add support to use the `photonics` target from Python frontend of CUDA-Q

* Proposed API

```
import cudaq

cudaq.set_target("photonics")


# New decorator; for a given kernel, `qudit` level is constant
@cudaq.photonics_kernel(level=3)
def photonicsKernel():
    qutrits = cudaq.qvector(2)
    plus(qutrits[0])
    plus(qutrits[1])
    plus(qutrits[1])
    mz(qutrits)


counts = cudaq.sample(photonicsKernel)
print(counts)
```

khalatepradnya:
For vis, @Omar-ORCA

Omar-ORCA:
> For vis, @Omar-ORCA

LGTM

khalatepradnya:
Updated the decorator as per inputs from @amccaskey

bettinaheim:
Let's have a quick sync on implementation to make sure we are not missing anything that will lead to some technical pain across the different scenarios in the future.

khalatepradnya:
Updated the decorator - handler named 'orca', qudit level associated with data type, not kernel.

khalatepradnya:
[API docs](https://htmlpreview.github.io/?https://github.com/khalatepradnya/cuda-quantum/blob/api_photonics/docs/photonics_kernel.html)

Omar-ORCA:
The use of the `plus`  gate in photonics circuits is not intuitive, changing it for the set of `create` and `annihilate` gates makes more sense.
The following changes will be needed:

> ### Required prerequisites
> * [x]  Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.
> 
> ### Describe the feature
> * Add support to use the photonics simulator with `photonics-cpu` target from Python frontend of CUDA-Q

* Add support to use the photonics simulator with `photonics` target from Python frontend of CUDA-Q

> * Proposed API
>   
>   * Gates (same as those supported on C++ frontend):
>     
>     1. `plus(q: qudit) -> None`,
>     2. `phase_shift(q: qudit, phi: float) -> None`,
[...]

PR #2055, opened 2024-08-06
[Python] Python frontend for 'photonics' target 

khalatepradnya:
### Description
Towards https://github.com/NVIDIA/cuda-quantum/issues/1937

* Prototype the requirements for photonics simulator support
   * `sample()` API covered; `get_state()` not covered
* Set execution manager from Python
* Infer the handler based on target


runtime/cudaq/qis/execution_manager.h
```diff
@@ -194,17 +194,36 @@ class ExecutionManager {
 
 /// Get the default execution manager instance.
 ExecutionManager *getExecutionManager();
```
khalatepradnya, line 196: Q: Should we replace `getExecutionManager` with the new `get_execution_manager`?

amccaskey: Why do we need a new `get_execution_manager() : ExecutionManager&` when we could just keep using the original one?

khalatepradnya: Discussed offline... The original method is defined by the expansion of the macro below.

python/cudaq/handlers/photonics_kernel.py
```diff
@@ -0,0 +1,212 @@
...
+
+        global globalQuditLevel
+        globalQuditLevel = None
+
+        self.kernelFunction = function
+
+        self.kernelFunction.__globals__['qudit'] = qudit
+        self.kernelFunction.__globals__['plus'] = plus
+        self.kernelFunction.__globals__['phase_shift'] = phase_shift
+        self.kernelFunction.__globals__['beam_splitter'] = beam_splitter
```
bmhowe23, line 208: I assume that these would still produce error messages if they were in a kernel that was accidentally targeting a non-photonics backend. Is that correct?

python/extension/CUDAQuantumExtension.cpp
```diff
@@ -178,6 +178,42 @@ PYBIND11_MODULE(_quakeDialects, m) {
...
+      },
+      "Allocate a qudit of given level.", py::arg("level"));
+  photonicsSubmodule.def(
+      "apply_operation",
+      [](const std::string &name, std::vector<double> &params,
+         std::vector<std::vector<std::size_t>> &targets) {
+        std::vector<cudaq::QuditInfo> targetInfo;
+        for (auto t : targets) {
+          targetInfo.push_back(cudaq::QuditInfo(t[0], t[1]));
+        }
```
bmhowe23, line 196: This would produce fewer intermediate copies.
```suggestion
        for (auto &t : targets)
          targetInfo.emplace_back(t[0], t[1]);
```

python/cudaq/handlers/photonics_kernel.py
khalatepradnya: Yes, and if I understood correctly, this test showcases that scenario - https://github.com/NVIDIA/cuda-quantum/pull/2055/files#diff-2953518ed100672f01a4c48a36b793f3234dcc0c3f62e7776d9c00b46a9fdff3R152.

python/cudaq/handlers/photonics_kernel.py
```diff
@@ -0,0 +1,212 @@
...
+            in the system.
+    """
+    level: int
+    id: int
+
+    def __del__(self):
+        cudaq_runtime.photonics.release_qudit(self.level, self.id)
+
+
+def _check_qudit_type(q):
```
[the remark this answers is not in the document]
khalatepradnya: Do you mean have this one simply return the result of `isinstance` call? Then we could perhaps call `isinstance` directly in the caller and raise exception there... This function encapsulates that behavior.

python/cudaq/handlers/photonics_kernel.py
```diff
@@ -0,0 +1,212 @@
...
+    
+    Returns:
+        Measurement results.
+    
+    Raises:
+        RuntimeError: If the qudit level is not set.
+        Exception: If input argument is not instance of `PyQudit` class.
+    """
+    _check_args(qudits)
+    if isinstance(qudits, PyQudit):
```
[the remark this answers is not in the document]
khalatepradnya: Right. This `measure` function is an overload on whether the input is a single qudit or list of qudits - I don't know if there's a better way to implement that in Python.

python/cudaq/handlers/photonics_kernel.py
```diff
@@ -0,0 +1,212 @@
...
+        return cudaq_runtime.photonics.measure(qudits.level, qudits.id,
+                                               register_name)
+    if isinstance(qudits, List):
```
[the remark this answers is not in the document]
khalatepradnya: Same as above.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -343,9 +344,25 @@ def from_json(jStr, overrideDict=None):
...
-        Invoke the CUDA-Q kernel. JIT compilation of the 
-        kernel AST to MLIR will occur here if it has not already occurred. 
+        Invoke the CUDA-Q kernel. JIT compilation of the kernel AST to MLIR 
+        will occur here if it has not already occurred, except when the target
+        requires custom handling.
         """
+
+        # Check if target is set
+        try:
+            target_name = cudaq_runtime.get_target().name
```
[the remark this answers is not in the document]
khalatepradnya: I think since they are different scenarios they don't need to be combined in one block.
Practically, the way `LinkedLibraryHolder` works, I don't think the target will be `None`, this check is to ensure in case the function gets called otherwise.

docs/sphinx/examples/python/providers/photonics_tbi.py
```diff
@@ -0,0 +1,45 @@
...
+
+@cudaq.kernel
+def TBI(
+    bs_angles: list[float],
+    ps_angles: list[float],
+    input_state: list[int],
+    loop_lengths: list[int],
+):
+    n_modes = len(input_state)
+    d = sum(input_state) + 1  # qudit level
```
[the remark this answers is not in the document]
khalatepradnya: Sure, fixed in 42faba3f1dd55d5feec1fcc2514a9e07782132b8

docs/sphinx/examples/python/providers/photonics_tbi.py
```diff
@@ -0,0 +1,45 @@
...
+
+    quds = [qudit(d) for _ in range(n_modes)]
+
+    for i in range(n_modes):
+        for _ in range(input_state[i]):
+            plus(quds[i])
+
+    c = 0
```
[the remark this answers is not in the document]
khalatepradnya: 42faba3f1dd55d5feec1fcc2514a9e07782132b8

python/cudaq/handlers/photonics_kernel.py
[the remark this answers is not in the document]
khalatepradnya: thanks, understood... updated in 8020f1eba75634e4fca7591e89a9962f6b47ba94

python/cudaq/handlers/photonics_kernel.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    _check_args(qudit)
+    cudaq_runtime.photonics.apply_operation('plusGate', [],
+                                            [[qudit.level, qudit.id]])
+
+
+def phase_shift(qudit: PyQudit, phi: float):
+    """
+    Apply phase shift gate.
+    TBD
```
bmhowe23, line 123: Is resolving these `TBD`s planned for the remainder of #1937? (Same for lines 141-145)

python/extension/CUDAQuantumExtension.cpp
```diff
@@ -178,6 +178,41 @@ PYBIND11_MODULE(_quakeDialects, m) {
...
+        return cudaq::getExecutionManager()->allocateQudit(level);
+      },
+      "Allocate a qudit of given level.", py::arg("level"));
+  photonicsSubmodule.def(
+      "apply_operation",
+      [](const std::string &name, std::vector<double> &params,
+         std::vector<std::vector<std::size_t>> &targets) {
+        std::vector<cudaq::QuditInfo> targetInfo;
+        for (auto &t : targets)
+          targetInfo.emplace_back(t[0], t[1]);
```
bmhowe23, line 195: Do we need any assertion or error checking here to ensure that t.size() == 2? If this is only called from other CUDA-Q code, then perhaps it is ok, but if this is directly called via user code, then it might be worth adding error checking.

runtime/cudaq/CMakeLists.txt
```diff
@@ -19,7 +19,8 @@ add_library(${LIBRARY_NAME}
                 algorithms/draw.cpp
                 qis/state.cpp
                 qis/remote_state.cpp
-                platform/quantum_platform.cpp 
+                platform/quantum_platform.cpp
+                qis/execution_manager_c_api.cpp
```
bmhowe23, line 23: nit: it would be preferable to keep the `qis` items together if possible (i.e. move the new `platform` one above the other `qis` entries).

bmhowe23 (approved):
Overall LGTM. Just a few minor comments/questions below.

python/cudaq/handlers/photonics_kernel.py
khalatepradnya: Yes, external contribution expected on the documentation, including gate descriptions.

runtime/cudaq/CMakeLists.txt
khalatepradnya: Fixed in aebe2ece6a061854947bd784413a77ee55440461

Issue #775, opened 2023-10-15
Add `get_state_async` documentation


1tnguyen:
- Add `get_state_async` examples to the `platform` docs page.

- Layout/format examples on the page to improve navigation/readability.

PR #2063, opened 2024-08-08
Add `get_state_async` examples to the `mqpu` platform docs

1tnguyen:
### Description

We've had these two `get_state_async` (py and cpp) examples added and tested for some time.
However, they were not included in the docs. Hence, this PR fixes that.

**Note**: the [linked issue](https://github.com/NVIDIA/cuda-quantum/issues/775) also mentioned layout/format/etc. since it's from a long time ago when we didn't use tabs throughout the docs to show C++ and Python examples. Since then, they have all been reorganized. 

Resolved https://github.com/NVIDIA/cuda-quantum/issues/775


khalatepradnya (approved):
LGTM. Thanks, Thien!

PR #2068, opened 2024-08-08
Add interface to exclude any permutation of arguments

schweitzpgi:
when performing argument conversion.

There is both a "drop_front" style interface to support the legacy style and a fully generalized interface to allow the filtering of any permutation of arguments.

~Requires #2043 to be merged first.~


1tnguyen (approved):
LGTM 👍

PR #2074, opened 2024-08-09
Stitch separate QIR pipelines together

bmhowe23:
This is just a draft to allow some others to review the approach to reconcile the full QIR path with the wireset-base/adaptive-profile QIR path. It tries to use common code where possible, so I updated the wireset-based path to invoke some of the standard QIR passes.

Note: https://github.com/NVIDIA/cuda-quantum/pull/2145 (targeting `main`) does the decomposition updates in a more preferable way, so this PR should be updated once the other one is merged everwhere.


lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
```diff
@@ -214,6 +214,15 @@ struct AddFuncAttribute : public OpRewritePattern<LLVM::LLVMFuncOp> {
     bool isAdaptive = convertTo == "qir-adaptive";
     const char *profileName = isAdaptive ? "adaptive_profile" : "base_profile";
 
+    StringRef requiredQubitsStr(std::to_string(info.nQubits));
```
schweitzpgi, line 217: This looks like it may be unsafe. `to_string` is returning a new `string` object, but that object isn't interned anywhere so the `StringRef` would be set to reference an object that could be deleted at the end of the statement.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
```diff
@@ -254,6 +256,16 @@ struct WireSetToProfileQIRPass
...
         regNameMap[disc.getOperation()] = "?";
     });
+    std::optional<std::uint32_t> highestIdentity;
+    op.walk([&](quake::BorrowWireOp op) {
+      highestIdentity = highestIdentity
+                            ? std::max(*highestIdentity, op.getIdentity())
+                            : op.getIdentity();
+    });
+    if (highestIdentity)
+      op->setAttr(cudaq::opt::QIRRequiredQubitsAttrName,
```
schweitzpgi, line 266: I think this needs to be done over the entire Module and not just this function. We may have more than 1 kernel in the Module.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
```diff
@@ -419,6 +436,9 @@ void cudaq::opt::addWiresetToProfileQIRPipeline(OpPassManager &pm,
     wopt.convertTo = profile.str();
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createWireSetToProfileQIR(wopt));
   pm.addPass(cudaq::opt::createWireSetToProfileQIRPost());
+  // Perform final cleanup for other dialect conversions (like func.func)
+  pm.addPass(cudaq::opt::createConvertToQIR());
+  cudaq::opt::addQIRProfilePipeline(pm, profile, /*performPrep=*/false);
```
schweitzpgi, line 441: My intention was to not run these. I'm not entirely sure why they are needed? This pipeline encodes "these passes must be run together in this order", as a sort of "if you don't do it this way, all bets are off", so this seems like the wrong place from that perspective.

ConvertToQIR's primary purpose is to convert the code to "full QIR" form. And the point of this wireset pass pipeline is to _not_ convert to "full QIR" at all.

There is an independent `cc-to-llvm` pass that does the work of converting the CC dialect to LLVM. That seems like it should be closer to what we really want to use after this pass pipeline.

It's also strange to run the conversion from full QIR to profile _foo_ QIR pass since the input isn't, and will never be, full QIR here.

I assume that there are small side-effects as far as attributes being stuck on the Module, etc. that are the real driver here. It would be nice to know exactly what those are...

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -373,7 +375,16 @@ qirProfileTranslationFunction(const char *qirProfile, mlir::Operation *op,
     pm.enableStatistics();
   std::string errMsg;
   llvm::raw_string_ostream errOs(errMsg);
-  cudaq::opt::addPipelineConvertToQIR(pm, qirProfile);
+  bool containsWireSet =
+      op->walk<mlir::WalkOrder::PreOrder>([](quake::WireSetOp wireSetOp) {
+          return mlir::WalkResult::interrupt();
+        }).wasInterrupted();
```
schweitzpgi, line 381: Won't this be more predictable (perhaps in the future)? It seems like we will know that if the target selected is _foo_ then the compiler will be running qubit management, mapping, etc. and scanning for the presence of an Op in the Module won't be needed or determinative of what has to happen next. If mapping uses wire_set and the output doesn't have a wire_set, then we can just raise a fatal error.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
bmhowe23: > I'm not entirely sure why they are needed?

These were needed for to get everything to lower correctly. Rather than copy/paste the items I needed for the small number of test cases I was dealing with, I tried to err on the side of use common code to ensure that it is a bit more future-proof. (At least that was my thinking.)

The first line (`createConvertToQIR`) was added primarily due to needing to convert the `func.func` to `llvm.func`, but there are bunch of other ones that I believe could pop up, such as the conversions happening in and around here:
https://github.com/NVIDIA/cuda-quantum/blob/515d5b1e967ecdc4549774e001323746179122f3/lib/Optimizer/CodeGen/ConvertToQIR.cpp#L177

The second line adds these items: (code link) https://github.com/bmhowe23/cuda-quantum/blob/cf82cb5a868ff8428777372bb4b65caa757ebca6/lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp#L545-L548

See the manually added inline comments for explanations for each one of them.
```cpp
  createConvertToQIRFuncPass(convertTo); // need QIR function attributes
  createQIRToQIRProfilePass(convertTo); // may not be needed but does no harm
  createVerifyQIRProfile(...); // need to verify the generated QIR as the appropriate base profile or adaptive profile
```

There may be other ways to refactor this so that the full QIR path and the wireset QIR path use common code, but I was initially trying to keep the changeset as small as possible.

lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
bmhowe23: > This looks like it may be unsafe. `to_string` is returning a new `string` object, but that object isn't interned anywhere so the `StringRef` would be set to reference an object that could be deleted at the end of the statement.

Yeah, I was surprised this worked. Here is the relevant section of StringRef.h:
```cpp
    /// Disallow accidental assignment from a temporary std::string.
    ///
    /// The declaration here is extra complicated so that `stringRef = {}`
    /// and `stringRef = "abc"` continue to select the move assignment operator.
    template <typename T>
    std::enable_if_t<std::is_same<T, std::string>::value, StringRef> &
    operator=(T &&Str) = delete;
```
So it looks like it would protect me because it would move a temporary if it could, but I should probably update this to be safe.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
bmhowe23: Addressed in d00b1a001. I also made it do something similar for QIR result numbers, too.

lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
bmhowe23: Updated in 96d09dfe3

bmhowe23:
Closing in lieu of #2163

PR #2077, opened 2024-08-12
Follow-up to #1603 - fix `kronprod` bug occurring with Nvidia simulators

### LINKED ISSUE #2076: `qvector` initialization from state does not work unless no other qubits have been allocated yet
### Required prerequisites

### Describe the bug

There is a bug introduced in PR #1603 that is causing state initialization from vectors to behave incorrectly. This causes incorrect simulation results on the `nvidia-*` targets.

The following example that tries to initialize 20 qubits (`q`) from a state vector after declaring an initial qubit `p` demonstrates the problem.
```python
import cudaq

@cudaq.kernel
def kernel(vec : list[complex]):
    p = cudaq.qubit()
    q = cudaq.qvector(vec)
    mz(p)
    mz(q)

v = [0. + 0j] * (1024*1024)
v[1024*1024-1] = 1.
print(cudaq.sample(kernel, v))
```
When run on a system with Nvidia GPUs, it produces the following result.
```bash
$ python3 test.py
{ 111111111111111111111:1000 }
```

### Expected behavior

The above result is incorrect. It should be `011111111111111111111` since qubit `p` should have been initialized to the `|0>` state.

Yes, it was first broken in ad367c1bc

### Environment

- **CUDA Quantum version**: 0.8.0 (occurred right before the release)
- **Python version**: N/A
- **C++ compiler**: N/A
- **Operating system**: N/A

### Suggestions

PR is forthcoming.

bmhowe23:
Fixes #2076


unittests/integration/get_state_tester.cpp
```diff
@@ -165,3 +165,28 @@ CUDAQ_TEST(GetStateTester, checkOverlapFromHostVector) {
   EXPECT_NEAR(1.0, state.overlap(hostState).real(), 1e-3);
 }
 #endif
+
+#ifdef CUDAQ_SIMULATION_SCALAR_FP64
+using SimScalarType = double;
```
amccaskey, line 170: Can't you use `cudaq::real` ?

bmhowe23: > Can't you use `cudaq::real` ?

Ah, thanks. I didn't see that. Updated in 1e04dd696.

1tnguyen (approved):
LGTM 👍 Thanks, Ben!

PR #2079, opened 2024-08-12
Explanation added to bloch sphere documentation

mmvandieren:
### Description

problems: 
* the explanation for the need for 3D was inaccurate as |i> and -|i> lie on the same 2D plane as |+> and |-> and are also examples of equal superposition so they are subsumed in the previous sentence.  
* there was no connection between the text and the image of the bloch sphere. in particular the variables phi and theta were undefined.

pull request: addressed these 2 problems with edits to the text


khalatepradnya (approved):
LGTM.

image

PR #2083, opened 2024-08-12
[core] Simplify autogenerated code.

schweitzpgi:
Use the cc.offsetof operation to reduce the size of the code.

~Requires #2082 be merged first.~

PR #2084, opened 2024-08-13
[JIT] Hook everything up to use the new argument synthesis

schweitzpgi:
- Add option to nvq++ to use the new kernel launcher.
- Add code to the runtime to implement the new launch sequence. This receives the new vector of arguments protocol, uses the new argument conversion, and then calls the new argument synthesis pass to specialize the kernel for JIT compilation.
-  Add a couple of tests to smoke test this new implementation.

### NOTE BEFORE REVIEWING THIS:
~Requires #2083 to be merged first.~

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include "common/ArgumentConversion.h"
 #include "common/Environment.h"
 #include "common/ExecutionContext.h"
 #include "common/Executor.h"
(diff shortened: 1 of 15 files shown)


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -402,10 +417,26 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
         throw std::runtime_error("Remote rest platform Quake lowering failed.");
     };
 
-    if (updatedArgs) {
-      cudaq::info("Run Quake Synth.\n");
+    if (!rawArgs.empty() || updatedArgs) {
       mlir::PassManager pm(&context);
-      pm.addPass(cudaq::opt::createQuakeSynthesizer(kernelName, updatedArgs));
+      if (!rawArgs.empty()) {
```
annagrin, line 422: Should there be a similar change in py_alt_launch_kernel.cpp (in `synthesizeKernel`)?

schweitzpgi: Yes. This PR gets one of the many flows up to the point it can be tested end-to-end. But we need to tackle all of the different flows. (I think that work can be done in parallel—finally.)

annagrin: sounds good (and looks good to me)

annagrin (approved):
LGTM, thanks Eric!

PR #2094, opened 2024-08-15
Add argument substitution for state pointers for local and remote simulators

annagrin:
### Description
Add argument substitution for state pointers for local and remote simulators.

#### Details

Substitute `cudaq::state*` arguments for simulators:

- For hardware or hardware emulators, issue an error message
- For simulators, read the data from the state and convert it to a constant array
  - Create a new state from the synthesized data using new `__nvqpp_cudaq_state_createFromData_fpXX` runtime functions
    - Convert data to remote sim precision if needed
    - Error if number of qubits in the state is more than 14 (we would need to use a different approach for larger states)
- Add tests

Towards: https://github.com/NVIDIA/cuda-quantum/issues/1931


runtime/common/ArgumentConversion.cpp
```diff
@@ -17,6 +17,34 @@
...
+  auto stateVector = state->get_tensor();
+  auto numElements = stateVector.get_num_elements();
+  auto elementSize = stateVector.element_size();
+
+  if (state->is_on_gpu()) {
+    if (precision == cudaq::SimulationState::precision::fp32) {
+      assert(elementSize == sizeof(std::complex<float>) &&
+             "Incorrect complex<float> element size");
+      auto *hostData = new std::complex<float>[numElements];
+      state->to_host(hostData, numElements);
```
schweitzpgi, line 32: I'm kind of lost why we need to do a deep copy of the data here.

At this point, we're in the process of launching the kernel. The bits in the `cudaq::state` object are final and cannot change at this point. So why copy them? Instead of copying them, we should be able to do the $n$ dereferences to the actual floating point bits and either construct a copy in MLIR code or skip the last dereference and pass the pointer to the block of memory.

So I don't think we want to pass a reference to the `cudaq::state` object (which is a wrapper around a shared_ptr) or reference to the shared_ptr, but the pointer inside the shared_ptr. And then we have to choice of using that pointer or dereferencing it to read out the data. What is the simulator expecting here? A vector? Which is yet another wrapper class in the way. Or maybe just something like a raw `double*`?

At any rate, I'm not sure we need this StateData object if we don't need to make a copy of the bits or create an object to manage that copy.

annagrin: The code will execute in a different process or on a different machine, so we need to copy and materialize the data so the server code will know what to use. Another option is to replace the quake.init_state by a call to a kernel that generates the data but it has a lot of work to do, and might not be ideal for remote simulation scenario (to be determined)

runtime/common/ArgumentConversion.cpp
```diff
@@ -84,27 +112,55 @@ static Value genConstant(OpBuilder &builder, const std::string &v,
...
-  // the unknown?
-  TODO("cudaq::state* argument synthesis");
-  return {};
-}
-
 // Forward declare aggregate type builder as they can be recursive.
 static Value genConstant(OpBuilder &, cudaq::cc::StdvecType, void *,
-                         ModuleOp substMod, llvm::DataLayout &);
+                         ModuleOp substMod, llvm::DataLayout &,
+                         const cudaq::opt::PlatformSettings &platform);
```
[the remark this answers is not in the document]
annagrin: I am going to remove the local simulation scenario, and copy the state data, but we would still have simulator/hardware options (need to emit a "TODO" in hardware case). So is passing a boolean (isSimulator) ok?

runtime/common/ArgumentConversion.cpp
```diff
@@ -84,27 +112,55 @@ static Value genConstant(OpBuilder &builder, const std::string &v,
...
+
+static Value genConstant(OpBuilder &builder, const cudaq::state *v,
+                         ModuleOp substMod, llvm::DataLayout &layout,
+                         const cudaq::opt::PlatformSettings &platform) {
+  if (platform.isSimulator && !platform.isRemote) {
+    // The program is executed in the same address space, use the pointer
+    // directly.
+    auto loc = builder.getUnknownLoc();
+    Value rawPtr = builder.create<arith::ConstantIntOp>(
+        loc, reinterpret_cast<std::intptr_t>(v), sizeof(void *) * 8);
```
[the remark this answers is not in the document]
annagrin: Going to remove this scenario to simplify things (only used from python's  cudaq::synthesize API)

runtime/common/ArgumentConversion.cpp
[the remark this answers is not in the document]
annagrin: Will try removing the nested state* treatment and see if it allows me to remove the platfrom/boolean from other calls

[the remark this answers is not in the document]
annagrin: Removed the nested state*, now only need to pass a boolean `isSimulator` args in a couple of places.

runtime/common/ArgumentConversion.cpp
```diff
@@ -147,27 +203,28 @@ Value dispatchSubtype(OpBuilder &builder, Type ty, void *p, ModuleOp substMod,
       .Case([&](cudaq::cc::PointerType ptrTy) -> Value {
         if (ptrTy.getElementType() == cudaq::cc::StateType::get(ctx))
           return genConstant(builder, static_cast<const cudaq::state *>(p),
-                             substMod);
+                             substMod, layout, platform);
```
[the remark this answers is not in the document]
annagrin: remove, thanks!

runtime/common/ArgumentConversion.cpp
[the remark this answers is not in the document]
annagrin: I removed special optimization for the local optimization for now.

runtime/common/ArgumentConversion.cpp
annagrin: Removed the `StateData` class, replaced by getting data values from the state using `state::operator()` instead.

runtime/common/ArgumentConversion.cpp
```diff
@@ -84,27 +112,55 @@ static Value genConstant(OpBuilder &builder, const std::string &v,
...
+                         llvm::DataLayout &,
+                         const cudaq::opt::PlatformSettings &platform);
 static Value genConstant(OpBuilder &, cudaq::cc::ArrayType, void *,
-                         ModuleOp substMod, llvm::DataLayout &);
+                         ModuleOp substMod, llvm::DataLayout &,
+                         const cudaq::opt::PlatformSettings &platform);
+
+static Value genConstant(OpBuilder &builder, const cudaq::state *v,
+                         ModuleOp substMod, llvm::DataLayout &layout,
+                         const cudaq::opt::PlatformSettings &platform) {
```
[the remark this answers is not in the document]
annagrin: Replaced by a `isSimulator` argument.

runtime/common/ArgumentConversion.cpp
```diff
@@ -342,20 +403,20 @@ void cudaq::opt::ArgumentConverter::gen(const std::vector<void *> &arguments) {
             .Case([&](cc::PointerType ptrTy) -> cc::ArgumentSubstitutionOp {
               if (ptrTy.getElementType() == cc::StateType::get(ctx))
                 return buildSubst(static_cast<const state *>(argPtr),
-                                  substModule);
+                                  substModule, dataLayout, platform);
```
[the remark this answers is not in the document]
annagrin: Done, thanks! Replaced the platform by `isSimulator` argument as well.

runtime/common/ArgumentConversion.cpp
```diff
@@ -100,6 +92,60 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+    auto buffer = builder.create<cudaq::cc::AllocaOp>(loc, arrTy);
+    builder.create<cudaq::cc::StoreOp>(loc, conArr, buffer);
+
+    auto cast = builder.create<cudaq::cc::CastOp>(loc, i8PtrTy, buffer);
+    auto statePtr = builder
+                        .create<func::CallOp>(loc, statePtrTy, createState,
+                                              ValueRange{cast, arrSize})
+                        .getResult(0);
+
+    // TODO: Delete the new state before function exit.
```
annagrin, line 140: @schweitzpgi what would be the best way to handle cleanup of the substitution? I'd like to add a function call at the function exit that deletes the new state.

runtime/cudaq/qis/state.cpp
```diff
@@ -48,12 +48,12 @@ void state::dump(std::ostream &os) const { internal->dump(os); }
...
 std::complex<double>
 state::operator()(const std::initializer_list<std::size_t> &indices,
-                  std::size_t tensorIdx) {
+                  std::size_t tensorIdx) const {
   std::vector<std::size_t> idxVec(indices.begin(), indices.end());
   return (*internal)(tensorIdx, idxVec);
 }
 
-std::complex<double> state::operator[](std::size_t idx) {
+std::complex<double> state::operator[](std::size_t idx) const {
```
annagrin, line 56: @1tnguyen I had to add a const to a couple of methods there to allow accessing state elements in ArgumentConversion.cpp. Does not seem to break anything...

runtime/common/ArgumentConversion.cpp
```diff
@@ -100,6 +92,60 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+    auto is64Bit =
+        v->get_precision() == cudaq::SimulationState::precision::fp64;
+    auto eleTy = is64Bit ? ComplexType::get(Float64Type::get(ctx))
+                         : ComplexType::get(Float32Type::get(ctx));
+    auto arrTy = cudaq::cc::ArrayType::get(ctx, eleTy, size);
+
+    auto genConArray = [&]<typename T>() -> Value {
+      std::complex<T> data[size];
+      for (std::size_t i = 0; i < size; i++) {
+        data[i] = (*v)({i}, 0);
```
annagrin, line 113: @1tnguyen accessing the state elements here (using the const functions from state.cpp).

runtime/common/BaseRestRemoteClient.h
```diff
@@ -913,6 +913,7 @@ class BaseNvcfRuntimeClient : public cudaq::BaseRemoteRestRuntimeClient {
         (funcEnv.majorVersion >= 1 && funcEnv.minorVersion >= 1);
     capabilities.vqe = funcEnv.majorVersion > 1 ||
                        (funcEnv.majorVersion >= 1 && funcEnv.minorVersion >= 1);
+    capabilities.isRemoteSimulator = true;
```
annagrin, line 916: @bmhowe23 added a new capability so we can pass it to argument conversion during synthesis.

runtime/cudaq/qis/state.cpp
```diff
@@ -111,6 +111,33 @@ extern "C" {
 std::int64_t __nvqpp_cudaq_state_numberOfQubits(state *obj) {
   return obj->get_num_qubits();
 }
+
+state *__nvqpp_cudaq_state_createFromData_fp64(void *data, std::size_t size) {
+  auto d = reinterpret_cast<std::complex<double> *>(data);
+  auto current = std::vector<std::complex<float>>(d, d + size);
+
+  auto *simulator = cudaq::get_simulator();
+  if (simulator->isSinglePrecision()) {
```
annagrin, line 120: @1tnguyen @bmhowe23 Here is where we allow to create a new remote sim state from data that could be coming from a simulation state from the host machine and have different precision.

runtime/cudaq/qis/state.cpp
```diff
@@ -111,6 +111,33 @@ extern "C" {
...
 }
+
+state *__nvqpp_cudaq_state_createFromData_fp64(void *data, std::size_t size) {
+  auto d = reinterpret_cast<std::complex<double> *>(data);
+  auto current = std::vector<std::complex<double>>(d, d + size);
+
+  auto *simulator = cudaq::get_simulator();
+  if (simulator->isSinglePrecision()) {
+    auto converted =
+        std::vector<std::complex<float>>(current.begin(), current.end());
```
1tnguyen, line 122: Would this work?
```suggestion
    auto converted =
        std::vector<std::complex<float>>(d, d + size);
```
and move the `current` constructor to after the `if` block

runtime/cudaq/qis/state.cpp
```diff
@@ -111,6 +111,33 @@ extern "C" {
...
+    return new state(state::from_data(converted));
+  }
+
+  return new state(state::from_data(current));
 }
 
+state *__nvqpp_cudaq_state_createFromData_fp32(void *data, std::size_t size) {
+  auto d = reinterpret_cast<std::complex<float> *>(data);
+  auto current = std::vector<std::complex<float>>(d, d + size);
```
1tnguyen, line 131: Same question as [above](https://github.com/NVIDIA/cuda-quantum/pull/2094/files#r1724473842).

targettests/Remote-Sim/state_init.cpp
1tnguyen, line 1: I wonder if we need to escape this file and the one below in `.github/workflows/nvqc_regression_tests.yml`. 
While we've synthesized the state argument, the new `__nvqpp_cudaq_state_createFromData` functions may not be defined in the old runtime (circa 0.7 version).

runtime/cudaq/qis/state.cpp
schweitzpgi: Let's try to use "left-handed" declarations in these cases rather than `auto`.
```suggestion
    std::vector<std::complex<float>> converted (current.begin(), current.end());
```

annagrin: Thanks, will do in the PR update.

runtime/cudaq/qis/state.h
```diff
@@ -31,6 +31,7 @@ class state {
   friend class state_helper;
 
 public:
+  explicit state() : internal(std::shared_ptr<SimulationState>()) {}
```
annagrin, line 34: remove:)

targettests/Remote-Sim/state_init.cpp
annagrin: done, thanks!

runtime/cudaq/qis/state.cpp
annagrin: done, thanks!

runtime/cudaq/qis/state.cpp
annagrin: Done, thanks!

.github/workflows/nvqc_regression_tests.yml
```diff
@@ -128,7 +128,8 @@ jobs:
             # nested_vectors: related to vector of pauli_words (https://github.com/NVIDIA/cuda-quantum/issues/1957)
             # custom_operation: https://github.com/NVIDIA/cuda-quantum/issues/1985
             # return_values: only supported in 0.8 NVQC service.
-            if [[ "$filename" != *"unsupport_args"* ]] && [[ "$filename" != *"state_overlap"* ]] && [[ "$filename" != *"compile_errors"* ]] && [[ "$filename" != *"nested_vectors"* ]] && [[ "$filename" != *"pauli_word"* ]] && [[ "$filename" != *"custom_operation"* ]] && [[ "$filename" != *"return_values"* ]]; then
+            # state_init: only supported in 0.8 NVQC service.
```
bmhowe23, line 131: Hopefully 0.8 will be deployed any day now. When it is deployed, will this change be needed?

runtime/common/ArgumentConversion.cpp
```diff
@@ -100,6 +92,60 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
 
+static Value genConstant(OpBuilder &builder, const cudaq::state *v,
+                         ModuleOp substMod, llvm::DataLayout &layout,
+                         bool isSimulator) {
+  if (isSimulator) {
+    // The program is executed remotely, materialize the simulation data
+    // into an array and create a new state from it.
+    auto ctx = builder.getContext();
+    auto loc = builder.getUnknownLoc();
+    auto size = 1ULL << v->get_num_qubits();
```
bmhowe23, line 103: Could we please range check `v->get_num_qubits()`? I've seen too many cases where the `<<` operand is >= 64 and then stuff silently goes haywire. (It's especially concerning in this case because `v` could point to invalid data in some erroneous cases.) It would be better to get a clear error message here rather than snowballing into something bigger.

runtime/common/ArgumentConversion.cpp
```diff
@@ -100,6 +92,60 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+    auto is64Bit =
+        v->get_precision() == cudaq::SimulationState::precision::fp64;
+    auto eleTy = is64Bit ? ComplexType::get(Float64Type::get(ctx))
+                         : ComplexType::get(Float32Type::get(ctx));
+    auto arrTy = cudaq::cc::ArrayType::get(ctx, eleTy, size);
+
+    auto genConArray = [&]<typename T>() -> Value {
+      std::complex<T> data[size];
```
bmhowe23, line 111: Is this really putting the data on the stack? That seems like we might encounter a stack overflow, with >= 20 qubits.

targettests/Remote-Sim/state_init_vector.cpp
```diff
@@ -0,0 +1,213 @@
...
+    cudaq::complex(0.0),
+    cudaq::complex(0.0)
+  });
+}
+
+__qpu__ void test_complex_array_param(std::vector<cudaq::complex> inState) {
+  cudaq::qvector q1 = inState;
+}
+
+__qpu__ void test_real_constant_array() {
```
bmhowe23, line 41: Would it be possible to make some bigger tests here, say, with 20-25 qubits?

runtime/common/ArgumentConversion.cpp
annagrin: Discussed offline - will add another pass to:
- replace state* by the data in `cudaq.init_state` instruction
- remove `getNumberOfQubits(state*)` calls (replace by constant)
- if the state is no longer used, remove the `createFromData` call
- otherwise, add a call to delete the state before every exit from the function

targettests/Remote-Sim/state_init_vector.cpp
annagrin: Hmm let me think how to best generate the constant that big:)

bmhowe23: Can it be generated in your test? I.e. (untested)
```cpp
std::vector<std::complex<double>> myVec(1<<20); 
for (int i = 0; i < myVec.size(); i++)
  myVec[i] = 1.0 / myVec.size();
```

annagrin: I gave it a try - please let me know if you have any suggestions. Max number of qubits I could use was 19 (20 caused a seg fault on the server)

.github/workflows/nvqc_regression_tests.yml
annagrin: Not sure - it depends on how much of the new synthesis implementation is in 0.8 - it would need support for `-fkernel-exec-kind=2` flag, specifically this PR: https://github.com/NVIDIA/cuda-quantum/pull/2101 and before that.

runtime/common/ArgumentConversion.cpp
annagrin: For now, yes, but I can also run the `LiftArrayAlloc` pass to move it to a global after synthesis.

annagrin: Let me try it out and see if I can run a larger test then.

annagrin: Unfortunately this vector is not a constant in the MLIR so we don't lift it to a global currently. I think this would need to be addressed separately.

targettests/Remote-Sim/state_init_vector.cpp
bmhowe23: > I gave it a try - please let me know if you have any suggestions. Max number of qubits I could use was 19 (20 caused a seg fault on the server)

20 qubits sounds like the number of qubits where the state vector approaches ~8MB, which is near the stack size. I don't think we should be putting the state onto the stack. Should we malloc/free it?

runtime/common/ArgumentConversion.cpp
bmhowe23: > For now, yes, but I can also run the `LiftArrayAlloc` pass to move it to a global after synthesis.

Whoever is calling `genConArray` would probably segfault no matter what is run afterwards (assuming `size` was large enough).

annagrin: Tried various approaches here and discovered a few things, thanks for stress testing:)
- putting data in the stack does not cause problems but I've changed it to a `unique_ptr` just in case
- creating a constant local array for data (by calling `genConstant`) works only up to 14 qubits
- creating a global constant array for data instead of calling `genConstant` works for up to 14 qubits as well but works faster

So I think it I will just bail out for now if we have more than 14 qubits, and continue working on other approaches (in later PRs)

runtime/common/ArgumentConversion.cpp
annagrin: Done, thanks!

runtime/common/ArgumentConversion.cpp
annagrin: Done, thanks!

targettests/Remote-Sim/state_init_vector.cpp
annagrin: Added a test with 19 qubits.

annagrin: Removed for now, will send another PR adding the test and the error during synthesis.

.github/workflows/nvqc_regression_tests.yml
bmhowe23: > Not sure - it depends on how much of the new synthesis implementation is in 0.8 - it would need support for `-fkernel-exec-kind=2` flag, specifically this PR: #2101 and before that.

Isn't synthesis running locally? If so, wouldn't a fully synthesized kernel be able to run on the NVQC server regardless of what version is on the server? (Perhaps it depends on the runtime functions used by that synthesized kernel?)

In any case, I'm mainly trying to figure out if we should create an issue to track the fact that we should remove this test from the exclusion list.

targettests/Remote-Sim/state_init.cpp
```diff
@@ -0,0 +1,84 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// REQUIRES: remote-sim
+// REQUIRES: c++20
+
+// clang-format off
+// RUN: nvq++ %cpp_std --enable-mlir --target remote-mqpu -fkernel-exec-kind=2 %s -o %t && %t | FileCheck %s
```
bmhowe23, line 13: I haven't been tracking all the new stuff regarding kernel launches. Will a user be expected to know when to do `-fkernel-exec-kind=2`? If so, it should probably be documented in the user-facing docs, and if not, all the spots that need edits should probably be tracked with some sort of `TODO-FIX-KERNEL-EXEC` (or something like that).

runtime/test/test_argument_conversion.cpp
```diff
@@ -289,6 +290,49 @@ void test_recursive(mlir::MLIRContext *ctx) {
...
+    std::vector<void *> v = {static_cast<void *>(&x)};
+    doSimpleTest(ctx, "!cc.ptr<!cc.state>", v);
+  }
+
+  // clang-format off
+// CHECK:       Source module:
+// CHECK:         func.func private @callee(!cc.ptr<!cc.state>)
+// CHECK:       Substitution module:
+
+// CHECK-LABEL:   cc.arg_subst[0] {
```
bmhowe23, line 308: It looks like this test is failing: https://github.com/NVIDIA/cuda-quantum/actions/runs/10499383135/job/29086484483#step:8:742

python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -166,6 +166,20 @@ def kernel(vec: List[complex]):
     assert not '111' in counts
 
 
+def test_quantinuum_state_synthesis():
```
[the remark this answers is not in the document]
bmhowe23: Oops, this is just checking for an expected error message. That's great! (Sorry for the oversight 😄 ). Marking as resolved.

python/tests/kernel/test_kernel_qvector_state_init.py
```diff
@@ -18,6 +18,27 @@
     not (cudaq.num_available_gpus() > 0 and cudaq.has_target('nvidia')),
     reason='Could not find nvidia in installation')
 
+# synthesis
+
+
+def test_kernel_synthesis_complex():
```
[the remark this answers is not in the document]
bmhowe23: Disregard this comment. I will mark as resolved. Sorry for the confusion.

runtime/common/ArgumentConversion.cpp
```diff
@@ -100,6 +92,75 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+
+    cudaq::IRBuilder irBuilder(ctx);
+    auto genConArray = [&]<typename T>() -> Value {
+      auto data = std::make_unique<std::complex<T>[]>(size);
+      for (std::size_t i = 0; i < size; i++) {
+        data[i] = (*v)({i}, 0);
+      }
+      std::string name =
+          kernelName.str() + ".rodata_synth_" + std::to_string(counter++);
+      std::vector<std::complex<T>> vec(data.get(), data.get() + size);
```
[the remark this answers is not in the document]
annagrin: I could create a vector and push_back things to it - maybe it will look better, but we will still allocate memory... Will do that.

targettests/Remote-Sim/state_init.cpp
annagrin: This should be temporary and `-fkernel-exec-kind=2` will be removed after we switch to the new launch by default. Will add the markings and an item in https://github.com/NVIDIA/cuda-quantum/issues/1931

runtime/common/ArgumentConversion.cpp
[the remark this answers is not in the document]
bmhowe23: In the current code, we are allocating 2x the memory (once for `data` and once for `vec`). I think it could be reduced to a single allocation, and push_back could be avoided, if done like this:
```cpp
std::vector<std::complex<T>> vec(size);
for (std::size_t i = 0; i < size; i++)
  vec[i] = (*v)({i}, 0);
```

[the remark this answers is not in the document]
bmhowe23: In other words, `std::vector<std::complex<T>> vec(data.get(), data.get() + size);` is simply making a copy of `data`, not just pointing to its underlying data, so that's where the duplicate allocation is coming from.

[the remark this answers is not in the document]
annagrin: Thanks! Updated.

runtime/test/test_argument_conversion.cpp
annagrin: fixed now, thanks!

.github/workflows/nvqc_regression_tests.yml
annagrin: Yes, the synthesis is running locally but the server will need new runtime API for creating states.
I added removal of it to https://github.com/NVIDIA/cuda-quantum/issues/1931.

targettests/Remote-Sim/state_init.cpp
annagrin: Added `// TODO-FIX-KERNEL-EXEC` in all tests that use the new launch and added an item in #1931

bmhowe23 (approved):
LGTM. Thanks, @annagrin!

PR #2097, opened 2024-08-16
Fix for non-commutativity check in spin_op multiplication

### LINKED ISSUE #2092: Incorrect spin op expression evaluation
### Required prerequisites

### Describe the bug

The computation of spin operators is incorrect. This bug affects cudaq.observe. A specific example where it is incorrect is that spin_op = (cudaq.spin.z(0) + cudaq.spin.z(1)) * cudaq.spin.x(0) should produce the matrix 
      [[ 0,  2,  0,  0],
       [ 0,  0,  0,  0],
       [ 0,  0,  0,  0],
       [ 0,  0, -2,  0]]
I can give more similar cases, but they are likely all the same bug. A good testbed is a new set of features that will be added in the near future.

spin_op = (cudaq.spin.z(0) + cudaq.spin.z(1)) * cudaq.spin.x(0)
spin_op.to_matrix()

### Expected behavior

The produced matrix should be 
      [[ 0,  2,  0,  0],
       [ 0,  0,  0,  0],
       [ 0,  0,  0,  0],
       [ 0,  0, -2,  0]]

Not a regression

### Environment

- **CUDA Quantum version**: 0.8.0 and previous
- **Python version**: N/A
- **C++ compiler**: N/A
- **Operating system**: N/A

### Suggestions

_No response_

sacpis:
Fixes #2092 

Fix for non-commutativity check in spin_op multiplication

# Test result

`spin_op = (cudaq.spin.z(0) + cudaq.spin.z(1)) * cudaq.spin.x(0)`
`print(spin_op.to_matrix())`

```
 (0,0)  (2,0)  (0,0)  (0,0)
 (0,0)  (0,0)  (0,0)  (0,0)
 (0,0)  (0,0)  (0,0)  (0,0)
 (0,0)  (0,0) (-2,0)  (0,0)
```

`spin_op = cudaq.spin.x(0) * (cudaq.spin.z(0) + cudaq.spin.z(1))`
`print(spin_op.to_matrix())`

```
 (0,0)  (2,0)  (0,0)  (0,0)
 (0,0)  (0,0)  (0,0)  (0,0)
 (0,0)  (0,0)  (0,0)  (0,0)
 (0,0)  (0,0) (-2,0)  (0,0)
```

# cTest result

```
100% tests passed, 0 tests failed out of 644

Label Time Summary:
gpu_required    =  89.12 sec*proc (367 tests)

Total Test time (real) = 539.95 sec
```

### DIFF
--- a/runtime/common/BaseRemoteSimulatorQPU.h
+++ b/runtime/common/BaseRemoteSimulatorQPU.h
@@ -107,12 +107,20 @@ public:
 
   void launchKernel(const std::string &name,
                     const std::vector<void *> &rawArgs) override {
-    throw std::runtime_error("launch kernel on raw args not implemented");
+    launchKernelImpl(name, nullptr, nullptr, 0, 0, &rawArgs);
   }
 
   void launchKernel(const std::string &name, void (*kernelFunc)(void *),
                     void *args, std::uint64_t voidStarSize,
                     std::uint64_t resultOffset) override {
+    launchKernelImpl(name, kernelFunc, args, voidStarSize, resultOffset,
+                     nullptr);
+  }
+
+  void launchKernelImpl(const std::string &name, void (*kernelFunc)(void *),
+                        void *args, std::uint64_t voidStarSize,
+                        std::uint64_t resultOffset,
+                        const std::vector<void *> *rawArgs) {
     cudaq::info(
         "BaseRemoteSimulatorQPU: Launch kernel named '{}' remote QPU {} "
         "(simulator = {})",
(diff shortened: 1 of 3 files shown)


runtime/cudaq/spin/spin_op.cpp
```diff
@@ -73,7 +73,8 @@ mult(std::vector<bool> row, std::vector<bool> other_row,
     tmp[i] = row[i] ^ other_row[i];
 
   for (std::size_t i = 0; i < numQubits; i++)
-    tmp2[i] = row[i] && other_row[numQubits + i];
+    tmp2[i] = (row[i] && other_row[numQubits + i]) ||
+              (row[i + numQubits] && other_row[i]);
```
bmhowe23, line 77: Are we sure we shouldn't be using an xor instead of or here? They both give the same answer for the test case in the original issue, and an xor feels a bit more intuitive here since we're doing binary arithmetic, but I could be wrong.
```suggestion
    tmp2[i] = (row[i] && other_row[numQubits + i]) ^
              (row[i + numQubits] && other_row[i]);
```

runtime/cudaq/spin/spin_op.cpp
bmhowe23, line 1: Could you please add a test to demonstrate the fix for this issue? That will help ensure we don't encounter any regressions.

runtime/cudaq/spin/spin_op.cpp
bmhowe23: > Thanks @bmhowe23. But using ^ does fails the test for y(1) * y(1) (test for commutativity for like terms).

OK, thanks for checking. If that is a pre-existing test, would you mind sharing a link to it?

anthony-santana (approved):

PR #2098, opened 2024-08-16
[python runtime] Use argument synthesis pass

schweitzpgi:
These changes convert the python runtime over to use the new argument synthesis pass.


annagrin (commented):
LGTM. Thank Eric!

annagrin (approved):

PR #2100, opened 2024-08-16
Demonstrate simple runnable example with dep-analysis pass

bmhowe23:
This bypasses the REST platform in order to use kernel return values. In turn, all the passes are listed on the command line.

Also, update regtomem to be able comprehend wire sets.

The aforementioned example is purely a proof of concept right now. This should not be checked into any real branches yet.


bmhowe23:
Replaced with other PR(s) to the feature branch.

PR #2101, opened 2024-08-16
Add hooks for remote launch to use argument synthesis.

schweitzpgi:

### DIFF
--- a/runtime/common/BaseRestRemoteClient.h
+++ b/runtime/common/BaseRestRemoteClient.h
@@ -175,16 +175,17 @@ public:
             moduleOp.push_back(funcOp.clone());
         }
         // Add globals defined in the module.
-        if (auto globalOp = dyn_cast<cudaq::cc::GlobalOp>(op))
+        if (auto globalOp = dyn_cast<cc::GlobalOp>(op))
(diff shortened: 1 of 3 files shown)


runtime/common/BaseRestRemoteClient.h
```diff
@@ -320,20 +334,20 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
       stateIrPayload1.entryPoint = kernelName1;
       stateIrPayload1.ir =
           constructKernelPayload(mlirContext, kernelName1, nullptr, args1,
-                                 argsSize1, /*startingArgIdx=*/0);
+                                 argsSize1, /*startingArgIdx=*/0, nullptr);
```
annagrin, line 337: Looks like this will always cause old synthesis to be used - are we implementing this later?

schweitzpgi: It doesn't change the current hookup, other than to add this extra unused pointer argument.

The new path enters via the `streamlinedLaunchKernel` entry point.

1tnguyen (approved):
LGTM 👍

PR #2102, opened 2024-08-19
Update the qubit mapping pass to use wire sets

bmhowe23:
This PR contains a collection of changes that were originally developed on the `features/qubit-mgmt` branch, but they have been isolated such that they can be merged into `main`, independent of that feature branch. The changes have not been previously reviewed, so they need to be reviewed just like normal in this PR.

* ~~Make the `add-wireset` pass add one `quake.wire_set` per function. The problem with the old approach of one per module is shown in a newly added test case (`test/Quake/wires_to_wireset.qke`). (The problem was that nested functions could borrow the same wire twice, which is not a good idea.) I backed these changes out.~~ Edit: Adam is going to make a separate PR to resolve the issue.
* Make the `qubit-mapping` pass operate on wires from a wire set rather than unnumbered null wires. This is the main point of the PR. This makes the qubit IDs used in the pass explicit rather than implicit. (However, the IDs are still dropped on the floor if the IR goes through the `regtomem` pass, which it does for now; this is no worse than it was before.)
* The `qubit-mapping` pass now manipulates the top level module by adding a `mapped_wireset` symbol in accordance with the user-provided device topology. Hence, the pass was split into a "prep" pass (to operate on the module) and a "func" pass (to operate on each function).
* Update `regtomem` to understand wire sets.
* Update CircuitCheck to understand wire sets to enable the existing mapping tests to continue working.
* Fix a bug in the `quake::WireSetOp::parse` function. The bug only occurred when it was parsing an op with that was set to private.

### DIFF
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -583,9 +583,9 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
     performs mapping of the qubits by inserting the necessary swap operations
     into the Quake IR.
 
-    Note 1: this pass requires strictly value semantics for any quantum
-    operations. It will throw an error if any memory reference semantics are
-    present on quantum operations.
+    Note 1: this pass requires strictly value semantics (in the form of wire
+    sets) for any quantum operations. It will throw an error if any memory
+    reference semantics are present on quantum operations.
 
     Note 2: this pass can introduce ancilla qubits.
 
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -12,10 +12,10 @@
 include "mlir/Pass/PassBase.td"
 
 def AddWireset : Pass<"add-wireset", "mlir::ModuleOp"> {
-  let summary = "Adds topology-less `quake.wire_set`s to the module.";
+  let summary = "Adds a topology-less `quake.wire_set` to the module.";
   let description = [{
-    Adds `quake.wire_set` operations without any topological information to the
-    ModuleOp (one for each function).
+    Adds a `quake.wire_set` operation without any topological information to the
+    ModuleOp.
   }];
 }
 
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -576,7 +576,7 @@ def LowerToCFG : Pass<"lower-to-cfg", "mlir::func::FuncOp"> {
   let constructor = "cudaq::opt::createLowerToCFGPass()";
 }
 
-def MappingPass: Pass<"qubit-mapping", "mlir::func::FuncOp"> {
+def MappingPass: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
   let summary = "Perform qubit mapping to account for connectivity constraints";
   let description = [{
     Some backends cannot support any-to-any multi-qubit operations, so this pass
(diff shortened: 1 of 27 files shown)


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -576,16 +576,16 @@ def LowerToCFG : Pass<"lower-to-cfg", "mlir::func::FuncOp"> {
...
   let description = [{
     Some backends cannot support any-to-any multi-qubit operations, so this pass
     performs mapping of the qubits by inserting the necessary swap operations
     into the Quake IR.
 
-    Note 1: this pass requires strictly value semantics for any quantum
-    operations. It will throw an error if any memory reference semantics are
-    present on quantum operations.
+    Note 1: this pass requires strictly value semantics (in the form of wire
+    sets) for any quantum operations. It will throw an error if any memory
```
schweitzpgi, line 587: How much complexity and effort is it for mapping to be flexible here? That is, it might take null_wire, borrow_wire, or a combination of the two?

targettests/execution/mapping_test-1-cpp17.cpp
```diff
@@ -7,7 +7,7 @@
  ******************************************************************************/
 
 // REQUIRES: c++17
-// RUN: nvq++ %cpp_std %s -o %t --target oqc --emulate && CUDAQ_DUMP_JIT_IR=1 %t 2> %t.txt | FileCheck %s && FileCheck --check-prefix=QUAKE %s < %t.txt
```
schweitzpgi, line 10: Why are the QIR checks dropped on the floor now?

include/cudaq/Optimizer/Transforms/Passes.td
bmhowe23: > How much complexity and effort is it for mapping to be flexible here? That is, it might take null_wire, borrow_wire, or a combination of the two?

I think that would be opening a can of worms that we might not want to open. It would have infer IDs for some of the qubits but not all of them. Some questions you might have to answer are: where do the inferred IDs start? It might be easy for cases like this:

```mlir
%0 = quake.borrow_wire @wires[0] : !quake.wire
// This would probably infer ID 1
%1 = quake.null_wire
```
But it would quickly get ill-defined. For example, consider this case
```mlir
%0 = quake.borrow_wire @wires[1] : !quake.wire
// Would this infer ID 0 or ID 2?
%1 = quake.null_wire
```
And even worse for cases like this:
```mlir
%0 = quake.borrow_wire @wires[2] : !quake.wire
// Would this infer ID 1 or ID 3? Or something else?
%1 = quake.null_wire
%0 = quake.borrow_wire @wires[0] : !quake.wire
```
So I think we should take the opportunity to take the simple path here.

targettests/execution/mapping_test-1-cpp17.cpp
bmhowe23: The QIR checks were overly restrictive. There were cases where swapping both result numbers and result strings would result in semantically the same answer, but the QIR checks couldn't verify that. I replaced the QIR checks with answer checks where required.
(Side note: those updates aren't strictly required for this PR, but they will be required if we ever run Adam's pass prior to mapping. I could remove them if you want, but we would probably have to add them back in to support Adam's other changes.)

schweitzpgi: I'm thinking let's leave them in then, but add the new checks as well. When we get to the point they have to come out, we can revisit.

include/cudaq/Optimizer/Transforms/Passes.td
schweitzpgi: So, if we take the combination off the table, then this will work either strictly borrow_wire or strictly null_wire?

bmhowe23: > So, if we take the combination off the table, then this will work either strictly borrow_wire or strictly null_wire?

No, not as written. I thought wire sets were the "new way" and were strictly better because they were more precise, so I was not intending for anybody to pass null wires into this pass anymore.

targettests/execution/mapping_test-1-cpp17.cpp
bmhowe23: > I'm thinking let's leave them in then, but add the new checks as well. When we get to the point they have to come out, we can revisit.

OK, I will add them back in.

bmhowe23: > > I'm thinking let's leave them in then, but add the new checks as well. When we get to the point they have to come out, we can revisit.
> 
> OK, I will add them back in.

Added back in a5ae54374925ff45ce0dd67b933cee637b381141

include/cudaq/Optimizer/Transforms/Passes.td
schweitzpgi: Just to capture the discussion from the meeting: the mapping pass will require that the input IR be in value semantics form using wire set notation on entry.  The wire set will be defined to correspond to the target machine's model. This should mean a topology was already selected, the wire set has an accurate description of the total number of qubits, $n$, and the qubit borrowing shall never be greater than $n$. If more than $n$ qubits are required, the compilation will terminate with an error.

For the mapper pass proper, if there are any `quake.alloca` or `quake.null_wire`, the above implies we can just emit an error and stop processing.

test/Transforms/mapping_with_meas-1.qke
```diff
@@ -6,30 +6,31 @@
...
+    quake.return_wire %wires_1 : !quake.wire
+    quake.return_wire %wires_3 : !quake.wire
     return
   }
 }
 
 // CHECK-LABEL:   func.func @__nvqpp__mlirgen__function_foo._Z3foov() attributes {"cudaq-entrypoint", "cudaq-kernel", mapping_reorder_idx = [1, 0, 2], mapping_v2p = [1, 0, 2], no_this} {
-// CHECK:           %[[VAL_0:.*]] = quake.null_wire
-// CHECK:           %[[VAL_1:.*]] = quake.null_wire
-// CHECK:           %[[VAL_2:.*]] = quake.null_wire
```
schweitzpgi, line 32: We should probably have a negative test that checks that mapping gives an error when it sees a `null_wire` or `quake.alloca` for that matter.

bmhowe23: > We should probably have a negative test that checks that mapping gives an error when it sees a `null_wire` or `quake.alloca` for that matter.

Added in 248b978f4

schweitzpgi (approved):
LGTM.

schweitzpgi:
I assume the empty file will be removed.

PR #2106, opened 2024-08-19
Only check links in files that changed for any given PR

bmhowe23:
We're getting too many sporadic link check failures for link checks that haven't actually changed and aren't permanently down. Update the workflow to only check the links in the files that have changed in any given PR.


bmhowe23:
The link checker doesn't seem to catch failures if `check-modified-files-only` is set to `"yes"`. Closing this PR.

PR #2107, opened 2024-08-19
Make assign-wire-indices skip functions with calls

atgeller:
### Description

Multiple functions that allocate qubits will cause `assign-wire-indices` to produce duplicate `borrow_wire` ops (i.e., two `borrow_wire` ops with the same index). This change ensures that `assign-wire-indices` refuses to run on kernels with multiple functions to ensure that each `borrow_wire` is unique.

### DIFF
--- a/lib/Optimizer/Transforms/WiresToWiresets.cpp
+++ b/lib/Optimizer/Transforms/WiresToWiresets.cpp
@@ -8,6 +8,7 @@
 
 #include "cudaq/Frontend/nvqpp/AttributeNames.h"
 #include "cudaq/Optimizer/Dialect/CC/CCDialect.h"
+#include "cudaq/Optimizer/Dialect/Characteristics.h"
 #include "cudaq/Optimizer/Dialect/Quake/QuakeDialect.h"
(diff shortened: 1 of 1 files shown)


schweitzpgi (approved):

lib/Optimizer/Transforms/WiresToWiresets.cpp
```diff
@@ -74,6 +75,18 @@ struct AssignWireIndicesPass
...
+    // TODO: someday we may want to allow calls to non-quantum functions
+    if (cudaq::opt::hasCallOp(func)) {
+      func.emitRemark(
+          "AssignWireIndicesPass function has calls, pass will not be run.");
+      return;
+    }
+
+    // Only run on the entrypoint, the expectation is that inlining has been
+    // done already, so there should only be one kernel remaining.
+    if (!func->hasAttr("cudaq-entrypoint"))
```
bmhowe23, line 87:
```suggestion
    if (!func->hasAttr(cudaq::entryPointAttrName))
```

PR #2109, opened 2024-08-19
[core] Add dynamic tuple layout handling.

schweitzpgi:
A std::tuple can be laid out in memory either like a struct or like the reverse of a struct. We need to be able to detect which one of these schemes is being used and it depends on the C++ headers being used. (libstdc++ is reversed, while libc++ is like a struct.)

Update tests.

### DIFF
--- a/lib/Frontend/nvqpp/ASTBridge.cpp
+++ b/lib/Frontend/nvqpp/ASTBridge.cpp
@@ -300,7 +300,24 @@ public:
     return true;
   }
 
+  bool isTupleReverseVar(clang::VarDecl *decl) {
+    if (cudaq::isInNamespace(decl, "cudaq")) {
+      auto name = decl->getName();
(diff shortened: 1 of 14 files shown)


lib/Frontend/nvqpp/ASTBridge.cpp
```diff
@@ -300,7 +300,24 @@ class QPUCodeFinder : public clang::RecursiveASTVisitor<QPUCodeFinder> {
     return true;
   }
 
+  bool isTupleReverseVar(clang::VarDecl *decl) {
+    if (cudaq::isInNamespace(decl, "cudaq")) {
+      auto name = decl->getName();
```
amccaskey, line 305: Possible simplification - `return decl->getName().equals("TupleIsReverse")` ?

khalatepradnya (approved):
Looks great! Thanks, Eric.

PR #2112, opened 2024-08-20
[arg synth] Test that address of variable isn't taken.

schweitzpgi:
New code pattern exposed that memtoreg was trying to promote a value in memory even though its address was taken. This fix is needed for the new argument synthesis work.


1tnguyen (approved):
LGTM

PR #2113, opened 2024-08-20
Fixes bug in lower to cfg.

schweitzpgi:
cc.if now has linear block arguments, which have to be threaded. Extend test coverage to cc.if and cc.loop.


bmhowe23 (approved):
👍

PR #2114, opened 2024-08-20
Updates to control flow support in dependency analysis

atgeller:
### Description

Refactors the analysis to work in the expected order, so now allocation and lifting work as expected.


test/Quake/dependency-1.qke
```diff
@@ -6,11 +6,11 @@
 // the terms of the Apache License 2.0 which accompanies this distribution.   //
 // ========================================================================== //
 
-// RUN: cudaq-opt --assign-ids --dep-analysis %s | FileCheck %s
+// RUN: cudaq-opt --add-wireset --assign-wire-indices %s | FileCheck %s
```
bmhowe23, line 9: Are these tests passing for you? It looks like they fail for me:
```
Failed Tests (6):
  CUDAQ :: Quake/dependency-1.qke
  CUDAQ :: Quake/dependency-2.qke
  CUDAQ :: Quake/dependency-3.qke
  CUDAQ :: Quake/dependency-4.qke
  CUDAQ :: Quake/dependency-classical.qke
```

bmhowe23 (commented):
Thanks, Adam. I'll download this PR to start playing with it, but would you mind updating it to address the compilation errors when using the CI's compiler: https://github.com/NVIDIA/cuda-quantum/actions/runs/10490052305/job/29060785461#step:7:921

bmhowe23 (approved):
I'm ok pulling this change in now and working the comments afterwards. I also think we need a fresh merge from main after incorporating this change into the feature branch.

test/Quake/dependency-1.qke
schweitzpgi: Those failures look familiar. I thought there was a patch to fix them, but maybe I'm recalling the wrong thing?

PR #2122, opened 2024-08-21
Add `opt-test` target

bmhowe23:
Add an `opt-test` target intended for compiler optimization testing.

Depends on #2120.

One can exercise the new `dep-analsysis` pass by doing `nvq++ --target opt-test test.cpp`.


targettests/execution/qubit_management/if-1.cpp
```diff
@@ -9,10 +9,8 @@
 #include <cudaq.h>
```
schweitzpgi, line 9: nit: can we remove this extra `qubit_management` directory here? rename the tests `qubit_management_if-0.cpp`, etc.

schweitzpgi (approved):

targettests/execution/qubit_management/if-1.cpp
bmhowe23: Yes, but I'd prefer that Adam do that in his upcoming PR since he is planning on adding some of his tests anyway. (If I rename them out from underneath him, then that might complicate his merge.)

PR #2124, opened 2024-08-21
Add destructors and fix up tests

atgeller:


lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -130,6 +130,7 @@ class LifeTimeAnalysis {
     if (best_reuse) {
       auto physical = best_reuse.value();
       lifetimes[physical]->combine(lifetime);
+      delete lifetime;
```
bmhowe23, line 133: Is it safe to conditionally delete a pointer in this function? How will the caller know if it was deleted or not?
Semi-related question: does it also need to be deleted after line 140?

atgeller: I think this is a side effect of the fact that the lifetimes here probably shouldn't be pointers, which will simplify away these issues.

bmhowe23 (approved):
👍

PR #2129, opened 2024-08-22
[Kernel Launch] Remote state to use new launch mode

1tnguyen:
### Description

- Use the new `vector<void*>` arguments for streamline launch in `remote_state`.

- For Python, previously, we pack `void*` into an `ArgWrapper*` primarily to send on the `mlir::ModuleOp`. In the new streamline launch approach, use the first element in `vector<void*>` to store `mlir::ModuleOp*`, i.e., the rest will be actual arguments.

### DIFF
--- a/python/runtime/utils/PyRemoteSimulatorQPU.cpp
+++ b/python/runtime/utils/PyRemoteSimulatorQPU.cpp
@@ -16,12 +16,12 @@ namespace {
 
 // This is a helper function to help reduce duplicated code across
 // PyRemoteSimulatorQPU and PyNvcfSimulatorQPU.
-static void _launchVQE(cudaq::ExecutionContext *executionContextPtr,
-                       std::unique_ptr<cudaq::RemoteRuntimeClient> &m_client,
-                       const std::string &m_simName, const std::string &name,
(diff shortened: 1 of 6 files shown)


python/runtime/utils/PyRemoteSimulatorQPU.cpp
```diff
@@ -74,6 +74,41 @@ static void _launchKernel(cudaq::ExecutionContext *executionContextPtr,
     throw std::runtime_error("Failed to launch kernel. Error: " + errorMsg);
 }
 
+static void
+_launchKernelStreamline(cudaq::ExecutionContext *executionContextPtr,
```
schweitzpgi, line 78: nit: shouldn't really start library implementation functions with an `_`.

annagrin (approved):
LGTM, thanks Thien!

python/runtime/utils/PyRemoteSimulatorQPU.cpp
1tnguyen: Good point. I renamed all of those functions to use the `Impl` suffix instead.

PR #2135, opened 2024-08-23
Update regtomem to handle more cases.

schweitzpgi:
The regtomem pass has pretty limited applicability. It can, under the correct preconditions, recover the quantum reference semantics model. However, it could fail in the presence of control-flow, so was only reliable when the kernel was a single block.

This PR reimplements much of regtomem to make it work with control-flow and the CC dialect's IfOp.

Add a test. The pass will no longer raise a fatal error if the input is in a form that shares quake operations ambiguously between multiple qubits.

A note on borrow/return: running this pass on borrow/return will *not* preserve the semantics of that form. The map from the virtual reference domain to a borrow set is a surjective function. We make this a quasi-bijective function but re-resolving the scope of the borrow to be global. The extra qubits can be cleaned up by another pass.

### DIFF
--- a/lib/Optimizer/Transforms/RegToMem.cpp
+++ b/lib/Optimizer/Transforms/RegToMem.cpp
@@ -39,6 +39,8 @@ static unsigned successorIndex(Operation *terminator, Block *successor) {
   return ~0u;
 }
 
+using BlockSet = SmallPtrSet<Block *, 4>;
+
 namespace {
 /// Register to memory analysis.
 ///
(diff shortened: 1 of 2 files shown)


lib/Optimizer/Transforms/RegToMem.cpp
```diff
@@ -287,36 +461,32 @@ class RegToMemPass : public cudaq::opt::impl::RegToMemBase<RegToMemPass> {
...
     // 2) Replace each quake.null_wire with a quake.alloca.
     SmallVector<Value> allocas(analysis.getCardinality());
-    for (auto nwire : analysis.getNullWires()) {
-      OpBuilder builder(nwire);
+    SmallVector<Value> borrowAllocas;
+    for (auto *nwire : analysis.getWires()) {
+      OpBuilder builder(ctx);
+      const bool fromWire = isa<quake::BorrowWireOp>(nwire);
+      if (fromWire)
+        builder.setInsertionPointToStart(&func.getBody().front());
```
bmhowe23, line 478: Won't this place the `quake.alloca`s coming from borrow wires in reverse order? If so, I would expect some of the existing tests to fail (e.g. `targettests/execution/mapping_test-1.cpp`).

bmhowe23 (approved):

PR #2140, opened 2024-08-25
Document C++ API for get_state, sample and observe

### LINKED ISSUE #2139: [docs] Missing cudaq::{observe, sample, get_state} entries in C++ API docs
Hi CUDA-Q team!

Thanks for this amazing project and the great work! It's been very fun and easy to get into it.

Looking at the [C++ API docs](https://nvidia.github.io/cuda-quantum/latest/api/languages/cpp_api.html), I've noticed that the functions `cudaq::observe`, `cudaq::sample` and `cudaq::get_state` (there might be more) do not have an entry. Are they listed on some other page I might have missed?

This is a minor issue, as the Python equivalents of these functions are well documented. If I have a moment tonight I might try to fix this myself.

lmondada:
closes #2139 

### Description

Added entries for the three functions to `docs/sphinx/api/languages/cpp_api.rst`, along with very minor modifications of two of the relevant header files. Sphinx now generates entries for these functions in the C++ API reference.

For `observe` and `sample`, I was unsure about how many overloads should be exposed, both to avoid unnecessary clutter in the docs and also to avoid clashes in Sphinx's name resolution. ATM all overloads (3) of `sample()` are exposed (this requires `sample_options` to be exposed, too), whereas for `observe()` the overloads that either:
- take a `DistributionType` template (afaik Doxygen cannot distinguish overloads that only differ in their return type), or
- take a `ArgumentSet` parameter (would require exposing this type synonym)

are excluded.

Let me know if you want me to adjust the list of exposed overloads, the ordering of the declarations or any other changes. Happy to modify it as you prefer.

Note: There are further related functions that are not exposed (e.g. their `_async` counterparts). Let me know if I should raise a further issue about those.


lmondada:

lmondada:
Hi @schweitzpgi and @bmhowe23 !

Thanks for your reviews. All CI checks are currently passing, so I won't be touching this PR anymore. Feel free to have a look whenever you have time, this is very low priority.

Bettina mentioned to me that a lot of the C++ API isn't documented on purpose, as future breaking changes aren't excluded -- feel free to ignore and close this PR in this case.

Of course also happy to modify and address any comments you may have.

bettinaheim (approved):

Issue #2146, opened 2024-08-26
[NVQC] New argument synthesis is not running yet for NVQC


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

A code with state pointer argument (`targettests/Remote-Sim/state_init.cpp`) currently fails for NVQC with a message from old argument synthesis:
```
error: 'func.func' op synthesis: unsupported argument type for remote devices and simulators: state* 
```
Note that command line tells to execute the new argument synthesis:
```
nvq++ %cpp_std --enable-mlir --target remote-mqpu -fkernel-exec-kind=2 %s -o %t && %t | FileCheck %s
```

### Steps to reproduce the bug

Run `targettests/Remote-Sim/state_init.cpp` for NVQC

### Expected behavior

The test should execute successfully.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

annagrin:
Note: `nvq++  --enable-mlir --target nvqc -fkernel-exec-kind=2 targettests/Remote-Sim/qvector_init_from_state.cpp -o tmp.x && ./tmp.x` runs successfully on my machine, looks like the issue is that nvqc tests in CI skip the `-fkernel-exec-kind=2` argument. 

Note: This will get fixed automatically when the new argument synthesis becomes default. 

For now all tests that pass `cudaq::state*` as a parameter should be disabled after https://github.com/NVIDIA/cuda-quantum/pull/2224.

schweitzpgi:
The default is `0` which supports return values (not just histograms over shots). I think that work had some bugs and some of #2216 tries to address those. Specifically, the hybrid mode (`0`) wasn't using argument synthesis but just falling back on the old quake synthesis path,

[...]

PR #2147, opened 2024-08-26
Fix failing NVQC test

annagrin:
### Description
Disable NVQC test that runs old argument synthesis until the new argument synthesis is plugged in into NVQC as well.

See https://github.com/NVIDIA/cuda-quantum/issues/2146
Task to enable the test is listed in https://github.com/NVIDIA/cuda-quantum/issues/1931


.github/workflows/integration_tests.yml
```diff
@@ -553,7 +553,8 @@ jobs:
           for filename in `find targettests/Remote-Sim -name '*.cpp'`; do
             # unsupport_args and compile_errors are compile error tests
             # nested_vectors: Compiler fails to synthesize nested vector parameters (https://github.com/NVIDIA/cuda-quantum/issues/2001)
-            if [[ "$filename" != *"unsupport_args"* ]] && [[ "$filename" != *"compile_errors"* ]] && [[ "$filename" != *"nested_vectors"* ]]; then
+            # state_init: New argument synthesis is not executed for nvqc (https://github.com/NVIDIA/cuda-quantum/issues/2146)
+            if [[ "$filename" != *"unsupport_args"* ]] && [[ "$filename" != *"compile_errors"* ]] && [[ "$filename" != *"nested_vectors"* ]] && [[ "$filename" != *"state_init"* ]]; then
```
bmhowe23, line 557: Wouldn't this also disable targettests/Remote-Sim/state_init_vector.cpp, too? Is that necessary/desired?

annagrin: yes, thanks! Renamed that test since the name was confusing.

annagrin: Now I am thinking both names are confusing. Will name them `qvector_init_vector.cpp` and `qvector_init_state.cpp` to indicate that we are initializing a `qvector` from vector or state arguments.

annagrin: done!

bmhowe23 (approved):
Thanks, @annagrin

Issue #2132, opened 2024-08-22
Accessing qview gives incorrect results


annagrin:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Accessing `qview[0:1]` seems to extract the first element of the original vector the `qview` is created from, not the `start of qview + 0` element of the original vector.

### Steps to reproduce the bug

```python
@cudaq.kernel
def kernel1(N: int):
    q = cudaq.qvector(N+N)
    x(q[0])

    qv = q[N:]
    x(qv[0:1])

counts = cudaq.sample(kernel1, 3)
print(counts)
```

### Expected behavior

Actual: `{ 000000:1000 }`

Expected: `{ 100100:1000 }`

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #2149, opened 2024-08-26
Fix runtime bug for span qview access where the span is of length 1

annagrin:
### Description
Fix runtime bug for span `qview` access where the span is of length 1.

The following example generated a runtime call to `@__quantum__rt__array_slice` with a range of element from 0 to 0 with step 1. The runtime function had incorrect optimization for a `start_idx == end_idx` case where it returned the `end_idx` instead of an index of the qubit at `end_idx`.

```
...
q = cudaq.qvector(N + N)
qv = q[N:]
 x(qv[0:1])
```

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2132


runtime/nvqir/QIRTypes.cpp
```diff
@@ -73,10 +73,6 @@ std::vector<int64_t> getRangeValues(Array *in_array, const Range &in_range) {
   // Convert to absolute index.
   const auto start_idx = convertIndex(in_range.start);
   const auto end_idx = convertIndex(in_range.end);
-  // start == end
-  if (start_idx == end_idx) {
-    return {end_idx};
```
schweitzpgi, line 78: Instead of deleting this completely, why not fix it to perform the indirection and return the qubit value? Seems easy enough to do here.
```
      auto qubit = *reinterpret_cast<Qubit **>((*in_array)[end_idx]);
      result.emplace_back(qubit->idx);
      return result;
```

annagrin: Isn't this exactly what happens below, in case `start_idx == end_idx`? Not sure if we are adding any optimization here...

schweitzpgi: Correct.

schweitzpgi (approved):

PR #2150, opened 2024-08-26
State synthesis for remote simulators - delete newly created states

annagrin:
### Description
Add a pass to delete new states added by synthesis.

After https://github.com/NVIDIA/cuda-quantum/pull/2094, argument synthesis for `cudaq::state*` creates new states from synthesized data using runtime functions. The current PR is the second phase - we can either optimize the states away (i.e. replace them by underlying data if the state is only use in `quake.state_init` instruction, or add a runtime call deleting the new state before function exits.

This PR adds a new pass, `DeleteStates`, for this purpose.


include/cudaq/Optimizer/Builder/Intrinsics.h
```diff
@@ -47,11 +47,14 @@ static constexpr const char getNumQubitsFromCudaqState[] =
     "__nvqpp_cudaq_state_numberOfQubits";
 
 // Create a new state from data.
-static constexpr const char createCudaqStateFromDataFP64[] =
+static constexpr const char createCudaqStateFromData64[] =
```
schweitzpgi, line 50: curious: why get rid of the "fp" part of the name?

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -669,6 +669,15 @@ def ObserveAnsatz : Pass<"observe-ansatz", "mlir::func::FuncOp"> {
   ];
 }
 
+def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
+  let summary =
+    "Remove unnecessary creation of new states from data";
+  let description = [{
+    - Replace state by its data if possible
+    - Add state deletion in before exit if not possible to remove a new state
```
schweitzpgi, line 677: I think this could be expanded a bit to be more informative as to what the pass is doing. An example of the problem it is solving might be useful here.

lib/Optimizer/Transforms/DeleteStates.cpp
```diff
@@ -0,0 +1,239 @@
...
+namespace cudaq::opt {
+#define GEN_PASS_DEF_DELETESTATES
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+#define DEBUG_TYPE "delete-states"
+
+using namespace mlir;
+
+// clang-format off
```
schweitzpgi, line 31: Something like this belongs in the .td file rather than here. (The .td file is the source-of-truth in LLVM parlance, so that is where one would go first to figure out what is what.)

schweitzpgi (commented):
This, LGTM. Thank you, Anna.

I think you could not remove the "FP" parts of the names and avoid touching several files here.

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: I was trying to match the rest of the API but then realized that they are not consistent anyway:) I'll bring them back.

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: will do, thanks!

lib/Optimizer/Transforms/DeleteStates.cpp
annagrin: will move, thanks!

test/Quake/delete_states.qke
```diff
@@ -0,0 +1,84 @@
...
+
+  func.func @__nvqpp__mlirgen__function_test_state_param1._Z16test_state_paramPN5cudaq5stateE() attributes {"cudaq-entrypoint", "cudaq-kernel", no_this} {
+    %c8_i64 = arith.constant 8 : i64
+    %0 = cc.address_of @function_test_state_param1._Z16test_state_paramPN5cudaq5stateE.rodata_synth_0 : !cc.ptr<!cc.array<complex<f32> x 8>>
+    %1 = cc.load %0 : !cc.ptr<!cc.array<complex<f32> x 8>>
+    %2 = cc.alloca !cc.array<complex<f32> x 8>
+    cc.store %1, %2 : !cc.ptr<!cc.array<complex<f32> x 8>>
+    %3 = cc.cast %2 : (!cc.ptr<!cc.array<complex<f32> x 8>>) -> !cc.ptr<i8>
+    %4 = call @__nvqpp_cudaq_state_createFromData_fp32(%3, %c8_i64) : (!cc.ptr<i8>, i64) -> !cc.ptr<!cc.state>
+    call @__nvqpp__mlirgen__sub_kernel(%4) : (!cc.ptr<!cc.state>) -> () // Why is this getting removed?
```
annagrin, line 55: remove comment:)

annagrin: done!

lib/Optimizer/Transforms/DeleteStates.cpp
annagrin: done - moved and added details

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: reverted

schweitzpgi (approved):

PR #2154, opened 2024-08-26
[custom op] Single qubit operation synthesis

khalatepradnya:
Towards: #1475 

Synthesize single qubit custom operations using Euler angle decomposition


lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+        std::atan2(special_unitary[3].imag(), special_unitary[3].real());
+    auto half_diff_alpha_gamma =
+        std::atan2(special_unitary[2].imag(), special_unitary[2].real());
+
+    angles.alpha = half_sum_alpha_gamma + half_diff_alpha_gamma;
+    angles.gamma = half_sum_alpha_gamma - half_diff_alpha_gamma;
+  }
+
+  BasisZYZ(std::vector<std::complex<double>> vec) {
+    std::move(vec.begin(), vec.end(), matrix.begin());
```
1tnguyen, line 86: Should we add a check here to make sure that `vec` is of length 4?

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+      func.setPrivate();
+      rewriter.restoreInsertionPoint(insPt);
+
+      insPt = rewriter.saveInsertionPoint();
+      auto *block = func.addEntryBlock();
+      rewriter.setInsertionPointToStart(block);
+
+      /// Use Euler angle decomposition for single qubit operation
+      auto zyz = BasisZYZ(unitary);
+      zyz.decompose();
```
1tnguyen, line 141: Does it make sense to make the `decompose` call part of the constructor so that we don't need to call it manually?

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+      /// NOTE: Typically global phase can be ignored but, if this decomposition
+      /// is applied in a kernel that is called with `cudaq::control`, the
+      /// global phase will become a local phase and give a wrong result if we
+      /// don't keep track of that.
+      if (isAboveThreshold(zyz.phase)) {
+        auto phase = cudaq::opt::factory::createFloatConstant(
+            loc, rewriter, zyz.phase, floatTy);
+        Value negPhase = rewriter.create<arith::NegFOp>(loc, phase);
+        rewriter.create<quake::R1Op>(loc, phase, ValueRange{}, arguments[0]);
+        rewriter.create<quake::RzOp>(loc, negPhase, ValueRange{}, arguments[0]);
```
1tnguyen, line 180: Do we have a test for this?
My back-of-the-envelope calculation is that a R1-Rz pair results in a theta/2 global phase, i.e., we need to multiply the angle by 2.

python/tests/backends/test_IQM.py
```diff
@@ -196,6 +200,43 @@ def test_IQM_state_preparation_builder():
     assert assert_close(counts["11"], 0., 2)
 
 
+def test_arbitrary_unitary_synthesis():
```
1tnguyen, line 203: For more coverage, could we also add a random unitary test (with fixed random seed)?
e.g., with a few random unitary matrices generated by `scipy.stats.unitary_group.rvs(2)`, check that the sample statistics matches the first column if the initial state is |0>.

python/tests/backends/test_IQM.py
```diff
@@ -196,6 +200,43 @@ def test_IQM_state_preparation_builder():
...
+    cudaq.register_operation("custom_x", np.array([0, 1, 1, 0]))
+
+    @cudaq.kernel
+    def basic_x():
+        qubit = cudaq.qubit()
+        custom_x(qubit)
+
+    counts = cudaq.sample(basic_x)
+    counts.dump()
+    # Gives result like { 1:999 0:0 }
```
1tnguyen, line 216: nit: is the default shot count 1000 :) ?
```suggestion
    # Gives result like { 1:1000 0:0 }
```

python/tests/backends/test_IQM.py
```diff
@@ -196,6 +200,43 @@ def test_IQM_state_preparation_builder():
...
+
+    @cudaq.kernel
+    def bell():
+        qubits = cudaq.qvector(2)
+        custom_h(qubits[0])
+        custom_x.ctrl(qubits[0], qubits[1])
+
+    counts = cudaq.sample(bell)
+    counts.dump()
+    # Gives result like { 11:499 10:0 01:0 00:499 }
```
1tnguyen, line 236:
```suggestion
    # Gives result like { 11:500 10:0 01:0 00:500 }
```

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+    auto half_sum_alpha_gamma =
+        std::atan2(special_unitary[3].imag(), special_unitary[3].real());
+    auto half_diff_alpha_gamma =
+        std::atan2(special_unitary[2].imag(), special_unitary[2].real());
+
+    angles.alpha = half_sum_alpha_gamma + half_diff_alpha_gamma;
+    angles.gamma = half_sum_alpha_gamma - half_diff_alpha_gamma;
+  }
+
+  BasisZYZ(std::vector<std::complex<double>> vec) {
```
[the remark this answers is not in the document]
schweitzpgi:
```suggestion
  BasisZYZ(const std::vector<std::complex<double>>& vec) {
```
and use `std::copy()` below is another optimized version that avoids the copy-on-call overhead.

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -35,6 +35,11 @@ namespace cudaq::opt {
 
 using namespace mlir;
 
+namespace cudaq {
+std::vector<std::complex<double>>
+readGlobalConstantArray(cudaq::cc::GlobalOp &global);
```
schweitzpgi, line 40: Let's put the declaration in a header file. I believe this one belongs in Factory.h, which is where the other constant extracting functions are.

lib/Optimizer/Transforms/CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ add_cudaq_library(OptTransforms
   RefToVeqAlloc.cpp
   RegToMem.cpp
   StatePreparation.cpp
+  UnitarySynthesis.cpp
+  Utils.cpp
```
schweitzpgi, line 54: No. Let's not have a "Utils" module. It ends up being a miscellaneous dumping ground of poorly organized code over time.

lib/Optimizer/Transforms/Utils.cpp
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
+
```
schweitzpgi, line 10: nit: blank line between include directives.

lib/Optimizer/Transforms/Utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <complex>
+#include <vector>
+
+using namespace mlir;
+
+namespace cudaq {
```
schweitzpgi, line 16: nit: use LLVM style.

lib/Optimizer/Transforms/Utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+std::vector<std::complex<double>>
+readGlobalConstantArray(cudaq::cc::GlobalOp &global) {
```
schweitzpgi, line 19:
```suggestion
cudaq::factory::readGlobalConstantArray(cudaq::cc::GlobalOp &global) {
```
(see LLVM style guide)

lib/Optimizer/Transforms/Utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
+
+#include <complex>
+#include <vector>
+
+using namespace mlir;
+
+namespace cudaq {
+
+std::vector<std::complex<double>>
```
schweitzpgi, line 18: Move this function to Factory.cpp. Thanks.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+#define DEBUG_TYPE "unitary-synthesis"
+
+using namespace mlir;
+
+namespace cudaq {
+std::vector<std::complex<double>>
+readGlobalConstantArray(cudaq::cc::GlobalOp &global);
```
schweitzpgi, line 34: Move to Factory.h instead of using block copies.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+namespace {
+
+struct EulerAngles {
+  double alpha;
+  double beta;
+  double gamma;
+};
+
+struct BasisZYZ {
+
```
schweitzpgi, line 46: nit: blank line

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+  void decompose() {
+    using namespace std::complex_literals;
+
+    /// Rescale the input unitary matrix, `u`, to be special unitary.
+    /// Extract a phase factor, `phase`, so that `det(inv_phase * unitary) = 1`
+    auto det = (matrix[0] * matrix[3]) - (matrix[1] * matrix[2]);
+    phase = 0.5 * std::atan2(det.imag(), det.real());
+    auto inv_phase = std::exp(-1i * phase);
+
+    std::array<std::complex<double>, 4> special_unitary;
```
schweitzpgi, line 64: nit: we use camelCase for variables, per the MLIR style guide.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+    std::transform(matrix.begin(), matrix.end(), special_unitary.begin(),
+                   [&](auto element) { return element * inv_phase; });
+
+    auto abs_00 = std::abs(special_unitary[0]);
```
schweitzpgi, line 68: camelCase

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+    auto abs_01 = std::abs(special_unitary[1]);
```
schweitzpgi, line 69: nit: No snake_case, Thanks. There is more snake_case here in this function, but I will stop with the nits here.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+    auto globalOp =
+        parentModule.lookupSymbol<cudaq::cc::GlobalOp>(generatorName);
+
+    /// The decomposed sequence of quantum operations are in a function
+    auto pair = generatorName.split(".rodata");
+    std::string funcName = pair.first.str() + ".kernel" + pair.second.str();
+
+    /// If the replacement function doesn't exist, create it here
+    if (!parentModule.lookupSymbol<LLVM::LLVMFuncOp>(funcName)) {
+
```
schweitzpgi, line 117: nit: blank line

It appears we have a lot of blank lines after an opening brace. That's not the style you'll find in the vast majority of the code base.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+      auto unitary = cudaq::readGlobalConstantArray(globalOp);
+
+      /// TODO: Expand the logic to decompose upto 4-qubit operations
+      if (unitary.size() != 4)
+        return customOp.emitError(
+            "Decomposition of only single qubit custom operations supported.");
```
schweitzpgi, line 123: I'm thinking the pass raising a fatal error here might be too Draconian in general. Do we really want the compiler to crash if the matrix isn't 2x2?

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+    StringRef generatorName = customOp.getGenerator().getRootReference();
+    auto globalOp =
+        parentModule.lookupSymbol<cudaq::cc::GlobalOp>(generatorName);
+
+    /// The decomposed sequence of quantum operations are in a function
+    auto pair = generatorName.split(".rodata");
+    std::string funcName = pair.first.str() + ".kernel" + pair.second.str();
+
+    /// If the replacement function doesn't exist, create it here
+    if (!parentModule.lookupSymbol<LLVM::LLVMFuncOp>(funcName)) {
```
schweitzpgi, line 116: Why is this an LLVMFuncOp? It should be a `func::FuncOp` at this point.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+      if (unitary.size() != 4)
+        return customOp.emitError(
+            "Decomposition of only single qubit custom operations supported.");
+
+      /// Controls are handled via apply specialization
+      auto funcTy =
+          FunctionType::get(parentModule.getContext(), targets.getTypes(), {});
+      auto insPt = rewriter.saveInsertionPoint();
+      rewriter.setInsertionPointToStart(parentModule.getBody());
+      auto func = rewriter.create<func::FuncOp>(parentModule->getLoc(),
```
schweitzpgi, line 130: This looks like a lurking multithreading bug. Since this pass is specified as a `func::FuncOp` pass, we shouldn't be modifying the Module here. Doing so is a silent race condition which will cause mysterious random crashes.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+                                                funcName, funcTy);
+      func.setPrivate();
+      rewriter.restoreInsertionPoint(insPt);
+
+      insPt = rewriter.saveInsertionPoint();
```
schweitzpgi, line 135: Why are we restoring and then saving the insertion point back to back here?

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+      auto *block = func.addEntryBlock();
+      rewriter.setInsertionPointToStart(block);
+
+      /// Use Euler angle decomposition for single qubit operation
+      auto zyz = BasisZYZ(unitary);
+      zyz.decompose();
+
+      /// For 1-qubit operation, apply on 'all' the targets
+      auto arguments = func.getArguments();
+      auto floatTy = cast<FloatType>(rewriter.getF64Type());
```
schweitzpgi, line 145:
```suggestion
      FloatType floatTy = rewriter.getF64Type();
```

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+      /// Ignore angles less than some threshold
+      double epsilon = 1e-9;
+      auto isAboveThreshold = [&](auto value) {
```
schweitzpgi, line 149:
```suggestion
      auto isAboveThreshold = [&](auto value) {
        const double epsilon = 1e-9;
```

targettests/execution/custom_operation_adj.cpp
```diff
@@ -7,6 +7,10 @@
  ******************************************************************************/
 
 // RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target ionq --emulate %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target iqm --emulate --iqm-machine Apollo %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target oqc --emulate %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t | FileCheck %s
```
schweitzpgi, line 13: It seems odd that none of these change the output in any way or if they do that we wouldn't check that the changes are correct and what is expected.
Corollary: if these new command lines don't exercise any new functionality, there is no need to add them to the regression tests.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+      rewriter.create<func::ReturnOp>(loc);
+      rewriter.restoreInsertionPoint(insPt);
+    }
+
+    rewriter.create<quake::ApplyOp>(
+        loc, TypeRange{}, SymbolRefAttr::get(rewriter.getContext(), funcName),
+        customOp.isAdj(), controls, targets);
+
+    rewriter.eraseOp(customOp);
```
schweitzpgi, line 191: Can we use `replaceOpWithNewOp` instead of `create` followed by `eraseOp`?

targettests/execution/custom_operation_adj.cpp
khalatepradnya: These should haven failing test(s) prior to the changes in this PR. There was a failing target test in Python but missing in C++. The test here is that the result is same across simulation targets which directly apply the custom matrix and hardware emulation targets which now decompose that custom matrix.

python/tests/backends/test_IQM.py
khalatepradnya: I started with checking that the count for '1' is '1000' and got failures locally. On multiple runs I noticed that the output was `{ 1:999 0:0 }`, rather than `{ 1:1000 0:0 }`,  hence the comment... So instead, I check that count for '0' is '0'. 
I don't know if there is an underlying issue.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: In my current tests, applying global phase had no effect on the results. I will add a test such that this is tested.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: The hardware targets cannot process the `quake.custom_op` unless it is decomposed, thus cannot proceed further.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Thanks! Addressed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7 by changing to a module pass.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Addressed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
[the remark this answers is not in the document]
khalatepradnya: Addressed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/StatePreparation.cpp
khalatepradnya: Addressed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/Utils.cpp
khalatepradnya: Removed this file.

lib/Optimizer/Transforms/Utils.cpp
khalatepradnya: Removed this file.

lib/Optimizer/Transforms/Utils.cpp
khalatepradnya: Removed this file, moved to `Factory.cpp` in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/Utils.cpp
khalatepradnya: Done. 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Removed. 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Fixed. 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Fixed. 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Fixed.

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: This was an oversight. Fixed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Fixed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Applied in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Applied changes in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Fixed in 3e71dee865f261ba1ba9379e172eb4b68ccab2d7

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Covered by newly added tests in ecb3c0c1119c0cb418106625b03a2ca36342012b

python/tests/backends/test_IQM.py
khalatepradnya: If I understood correctly, here's how I implemented this suggestion - newly added tests in ecb3c0c1119c0cb418106625b03a2ca36342012b

lib/Optimizer/Transforms/UnitarySynthesis.cpp
schweitzpgi: Sure. But there are other places, like target codegen, that can make that decision, or potentially make some other transformation.

Additionally, this pass is written as a greedy pass, which connotates that it is a best effort pass. That is, it is not an all-or-nothing showstopper (like the codegen pipeline necessarily is).

schweitzpgi (commented):
Looks really good. Thanks for the nice work. Some minor comments.

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -16,6 +17,9 @@
 #include "mlir/IR/Types.h"
 #include "mlir/Transforms/DialectConversion.h"
```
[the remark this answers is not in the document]
khalatepradnya: 🤦🏼‍♀️Fixed in a0630832cfc57b43b7744e9ff50e96c233e8d742

lib/Optimizer/Transforms/UnitarySynthesis.cpp
khalatepradnya: Okay, updated in a0630832cfc57b43b7744e9ff50e96c233e8d742, sot his pass only returns 'failure' (with warning) and then the error message comes from codegen.

targettests/execution/custom_operation_adj.cpp
schweitzpgi: I forgot that these tests are run in both a normal and a weird way.

schweitzpgi (approved):
Thank you, Pradnya.

PR #2162, opened 2024-08-27
Fix publishing issue for Shor's notebook

sacpis:
Adding pip install for contfrac

This will fix publishing error
https://github.com/NVIDIA/cuda-quantum/actions/runs/10571307755/job/29288632701

### DIFF
--- a/docs/sphinx/examples/python/tutorials/Shors.ipynb
+++ b/docs/sphinx/examples/python/tutorials/Shors.ipynb
@@ -29,7 +29,17 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
+   "id": "367c0921",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install contfrac"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 20,
    "id": "88619b35-624e-4853-8d2e-6d9952390e5e",
    "metadata": {
     "tags": []
@@ -77,7 +87,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": 21,
    "id": "7dd05729",
    "metadata": {},
    "outputs": [],
@@ -175,7 +185,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": 22,
    "id": "6c8eed12",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


docs/sphinx/examples/python/tutorials/Shors.ipynb
```diff
@@ -29,7 +29,17 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
+   "id": "367c0921",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install contfrac"
```
bmhowe23, line 37: I thought we determined that doing `pip install` from within the notebook didn't reliably work because we were running the notebook from within the context of `docs/notebook_validation.py`. Am I misremembering that?

docs/sphinx/examples/python/tutorials/Shors.ipynb
```diff
@@ -829,30 +839,31 @@
...
       "testing order 2\n",
       "testing order 3\n",
       "Found order: 3\n",
       "The order of a is 3\n",
       "No non-trivial factor found\n",
       "retrying...\n",
-      "Trying a = 18\n",
-      "Found factor by chance: 3 7\n"
+      "Trying a = 0\n",
+      "Found factor by chance: 21 1\n"
```
bmhowe23, line 857: `21` and `1` are trivial factors. I don't think we want the main docs page showing those trivial factors. (I.e. I think we would prefer that it find the 3 and 7, as it was before.)

bmhowe23: Yeah, it must be different every time (which is unfortunate for those of us who like repeatability 😄 ). I'm just saying that it would be better if the one that we publish in the docs didn't say 21 and 1, so could you re-run it, or revert the output cells to what they were before?

docs/sphinx/examples/python/tutorials/Shors.ipynb
bmhowe23: FWIW, we added `contfrac` to one of the CI images here during the PR:
https://github.com/NVIDIA/cuda-quantum/blame/c4001f66a8f90c0367126b0fdc5df59f24cd8cdf/.github/workflows/docker_images.yml#L781

So perhaps we need to also add it to the Publishing workflow somehow (perhaps as opposed to putting the `pip install` in the notebook itself).

PR #2163, opened 2024-08-27
Add `dep-analysis` pass (a.k.a. `features/qubit-mgmt`)

bmhowe23:
Description and code from @atgeller. Thanks, Adam!

### Description
DependencyAnalysis assigns physical qubits to virtual qubits and reorders the quake code to better reflect the ordering of operations implicit in the quake code, to try to open up additional parallelism. First, it constructs a data structure representing the quake code with explicit operation ordering constraints (in the form of dependency relations between operations) and performs several analyses/optimizations on it to try to reduce the total number of physical qubits used and the overall cycle-length of the circuit.

The first optimization is to lower virtual qubits allocation/deallocation pairs into the inner-most blocks they are used (deleting them in branches where they are not used). This opens up more physical qubit reuse opportunities within inner scopes and gives the other analyses/optimizations more flexibility.

The second optimization is to lift common operations at the beginning/end of the then and else blocks within an `if` to the parent scope. This optimization can reduce the circuit length. Also, because DependencyAnalysis currently treats `if`s as a "solid barrier" that the analyses cannot look inside (for scalability), also gives the parent context more flexibility in reusing qubits, since it has more information available that was previously "hidden" within the `if`.
The first analysis is scheduling: assigning a cycle to every quantum operation, such that the cycle respects the explicit ordering constraints. That is, an operation must be scheduled at a cycle after each of its dependencies, and any operation that depends on it must be likewise be scheduled after it.

The second analysis calculates lifetimes for virtual qubits. The lifetime of a qubit is from the first cycle in which it is used through the last cycle it is used in.

The third and final analysis assigns physical qubits to virtual allocations, using the lifetime information to find opportunities to assign the same physical qubit to two different virtual qubits. Because this may reduce the number of qubits used, this is also an optimization.

Currently, the second optimization and the three analyses are run together in an inside-out optimization pass. In this pass, inner `if`s are first resolved, then the three analyses are run on the block. Parent `if`s combine the physical allocations between the then and else blocks, respecting the reuse decisions they make. After the inner physical qubit assignments are combined, common operations are lifted from the inner blocks of the `if`. There are more opportunities to lift operations from the `if` at this point, since different virtual qubits may have been assigned the same physical qubit, so more operations will be "common". Finally, now that the inner `if` is resolved, the pass now runs on the parent block of the `if`.

After the analyses and optimizations are performed, new quake code is generated based on the resulting data structure.


lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
bmhowe23, line 1: We did not end up needing this file for the approach we ended up taking, so I'm wondering if it should be pulled from this PR. (We definitely want to hold onto this file for the future; I'm just not sure if it should be included in this PR since it is non-functional when hooked up end-to-end. Thoughts?)

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -290,6 +296,19 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
     passPipelineConfig = std::string("cc-loop-unroll{allow-early-exit=") +
                          allowEarlyExitSetting + "}," + passPipelineConfig;
 
+    if (getEnvBool("DISABLE_QUBIT_MANAGEMENT", false)) {
```
bmhowe23, line 299: @atgeller Is this still valid, or should it be removed?

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
schweitzpgi: I say we leave it in along with the patches. It is the correct way to do this, just couldn't be finished in the timeframe given.

runtime/common/BaseRemoteRESTQPU.h
atgeller: That should probably be removed or at least updated. Can we add an `fno-qubit-mgmt` option to `nvq++` instead perhaps?

bmhowe23: > That should probably be removed or at least updated. Can we add an `fno-qubit-mgmt` option to `nvq++` instead perhaps?

I don't think that is necessary because `qubit-mgmt` is not being run unless a user explicitly chooses the `opt-test` target with the `dep-analysis` option.

test/Quake/dependency-if-bug-lowering.qke
```diff
@@ -0,0 +1,49 @@
+// ========================================================================== //
```
schweitzpgi, line 1: nit: The `-` in the filename should be `_` for consistency with the rest of the tests.

The filenames ought to follow a snake case convention, with a `-<n>` suffix for several variant tests of the same/similar thing.

test/Quake/dependency-if-bug-lowering.qke
```diff
@@ -0,0 +1,49 @@
...
+  quake.sink %2#1 : !quake.wire
+  return %3 : i1
+}
+
+// CHECK-LABEL:   quake.wire_set @wires[2147483647] attributes {sym_visibility = "private"}
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__run_test() -> i1 attributes {"cudaq-entrypoint", "cudaq-kernel"} {
+// CHECK:           %[[VAL_0:.*]] = arith.constant true
+// CHECK:           %[[VAL_1:.*]] = quake.borrow_wire @wires[1] : !quake.wire
+// CHECK:           %[[VAL_2:.*]] = quake.borrow_wire @wires[0] : !quake.wire
```
schweitzpgi, line 35: Should these 2 lines be `CHECK-DAG`, assuming the order is irrelevant?

test/Quake/dependency-if-bug-lifting.qke
```diff
@@ -0,0 +1,124 @@
...
+// Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --add-wireset --assign-wire-indices --dep-analysis -split-input-file %s | FileCheck %s
+
+// The `quake.h` on `%arg0` will be lifted, causing a conflict as the same qubit will be reused for r and q
```
schweitzpgi, line 11: Is this comment out of date? It looks like the `h` is hoisted out of the conditional without an error.

test/Quake/dependency-if-bug-classical.qke
```diff
@@ -0,0 +1,136 @@
...
+    cc.continue %8 : !quake.wire
+  }
+  %measOut_1, %wires_2 = quake.mz %6 : (!quake.wire) -> (!quake.measure, !quake.wire)
+  %7 = quake.discriminate %measOut_1 : (!quake.measure) -> i1
+  quake.sink %wires : !quake.wire
+  quake.sink %wires_2 : !quake.wire
+  return %7 : i1
+}
+
+// CHECK-LABEL:   quake.wire_set @wires[2147483647] attributes {sym_visibility = "private"}
```
schweitzpgi, line 116: noting: we need to fix this. qubit management should not be using an absurd physical machine definition.

test/Quake/dependency-if-bug-classical.qke
```diff
@@ -0,0 +1,136 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --add-wireset --assign-wire-indices --dep-analysis -split-input-file %s | FileCheck %s
```
schweitzpgi, line 9: curious: why are these using `split-file`? Does the pass fall apart if there is more than 1 func.func in the module?

runtime/common/BaseRemoteRESTQPU.h
atgeller: Then it should be removed.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
bmhowe23: > I say we leave it in along with the patches. It is the correct way to do this, just couldn't be finished in the timeframe given.

Re: "along with the patches", how do you want those included? Should I just merge the relevant fixes from #2074 into this PR?

test/Quake/dependency-if-bug-lifting.qke
atgeller: It is indeed, the bug the test case show cases has since been fixed :)

test/Quake/dependency-if-bug-classical.qke
atgeller: It relies on `assign-wire-indices` which wants only one entrypoint.

test/Quake/dependency-if-bug-lowering.qke
atgeller: A lot of the tests could have `CHECK-DAG`, but even `CHECK-DAG` is a bit restrictive with the reordering that is possible with `dep-analysis`, so it's likely that any change to the ordering will break the tests in other places too.

runtime/common/BaseRemoteRESTQPU.h
bmhowe23: Removed in 82671ba68

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -214,6 +215,49 @@ def DecompositionPass: Pass<"decomposition", "mlir::ModuleOp"> {
...
+    scheduling algorithm tries to pack operations as densely as possible,
+    minimizing the number of cycles between operations.
+
+    From this dependency graph, we can calculate the lifetime of a qubit: from
+    the cycle in which it is first used through the cycle in which it is last
+    used. If two virtual qubits have non-overlapping lifetimes, they can be
+    assigned to the same physical qubit, as every virtual qubit is assumed to be
+    fully reset before release. Failure to fully reset virtual qubits before
+    release is undefinied behavior, and will likely lead to incorrect output
+    when running DependencyAnalysis.
```
bmhowe23, line 246: Noting for future: we need to highlight this caveat in any eventual user-facing documentation.

targettests/execution/qubit_management_if-1.cpp
```diff
@@ -0,0 +1,35 @@
...
+struct run_test {
+  __qpu__ auto operator()() {
+    cudaq::qubit q;
+
+    if (true)
+      x(q);
+    else
+      y(q);
+
+    bool b = mz(q);
```
bmhowe23, line 25: If possible, I think we should add `reset` calls after every `mz` in order to comply with the pass assumptions listed in the pass description. Otherwise, our own tests are testing with undefined behavior.

targettests/execution/qubit_management_if-2.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    h(q);
+    bool b = mz(q);
+    
+    // Should be able to lift x(p/r)
+    if (b) {
+      cudaq::qubit p;
+      x(p);
+      y(p);
+      res = mz(p);
```
bmhowe23, line 27: Same comment regarding `reset` here. (I won't comment any more about that.)

test/Quake/dependency-if-bug-classical.qke
bmhowe23: > It relies on `assign-wire-indices` which wants only one entrypoint.

I don't think `assign-wire-indices` prohibits multiple functions from having the `cudaq::entryPointAttrName` attribute set, does it? (My quick local test suggests that correctly assigns wires just fine.)

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+// Generated logic
+//===----------------------------------------------------------------------===//
+namespace cudaq::opt {
+#define GEN_PASS_DEF_DEPENDENCYANALYSIS
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+namespace {
+typedef size_t PhysicalQID;
+typedef size_t VirtualQID;
```
bmhowe23, line 38: It would be helpful to have descriptions on what you mean by "physical" and "virtual" for these types. (I know you explained offline, so just putting that in writing would be good.)

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+
+size_t getOperandIDXFromResultIDX(size_t resultidx, Operation *op) {
+  if (isa<RAW_MEASURE_OPS>(op))
+    return 0;
+  if (isa<quake::SwapOp>(op))
+    return (resultidx == 0 ? 1 : 0);
```
bmhowe23, line 44: Why is the swap gate not treated like any other 2-qubit gate? Is this attempting to reconstruct something from swap gates that you have inserted? If not, then I think this should simply fall through into the for loop below.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+      break;
+  }
+  return resultidx;
+}
+
+size_t getResultIDXFromOperandIDX(size_t operand_idx, Operation *op) {
+  if (isa<RAW_MEASURE_OPS>(op))
+    return 1;
+  if (isa<quake::SwapOp>(op))
+    return (operand_idx == 0 ? 1 : 0);
```
bmhowe23, line 59: Same here.

test/Quake/dependency-if-bug-classical.qke
schweitzpgi: Well, we don't want that pass to require that there be 1 func in the module at any rate. That'll need to be fixed if it is still an issue.

Each function need not use its own wire set operation, btw. A wire set is a global resource available to all the functions in the module.

targettests/execution/qubit_management_if-classical.cpp
```diff
@@ -0,0 +1,34 @@
...
+    }
+    auto res = mz(p);
+    return res;
+  }
+};
+
+int main() {
+  bool result = run_test{}();
+  printf("Result = %b\n", result);
+  return 0;
```
schweitzpgi, line 33: Should this return non-zero if the value of result is incorrect?

targettests/execution/qubit_management_if-5.cpp
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    return res;
+  }
+};
+
+int main() {
+  bool result = run_test{}();
+  printf("Result = %b\n", result);
+  return 0;
```
schweitzpgi, line 40: likewise

targettests/execution/qubit_management_if-1.cpp
```diff
@@ -0,0 +1,35 @@
...
+// TODO: filecheck with statistics
+
+// RUN: nvq++ --target opt-test --target-option dep-analysis,qpp %s -o %t && %t
+
+// Simple test, shouldn't affect anything
+struct run_test {
+  __qpu__ auto operator()() {
+    cudaq::qubit q;
+
+    if (true)
```
schweitzpgi, line 20:
```suggestion
    if (condition)
```

targettests/execution/qubit_management_if-1.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <cudaq.h>
+
+// TODO: filecheck with statistics
+
+// RUN: nvq++ --target opt-test --target-option dep-analysis,qpp %s -o %t && %t
+
+// Simple test, shouldn't affect anything
+struct run_test {
+  __qpu__ auto operator()() {
```
schweitzpgi, line 17:
```suggestion
  __qpu__ auto operator()(bool condition) {
```
just wondering if removing the constant condition breaks anything

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+};
+
+class DependencyGraph {
+private:
+  SetVector<DependencyNode *> roots;
+  DenseMap<VirtualQID, InitDependencyNode *> allocs;
+  DenseMap<VirtualQID, DependencyNode *> leafs;
+  SetVector<VirtualQID> qids;
+  DenseMap<PhysicalQID, DependencyNode *> qubits;
+  uint total_height;
```
bmhowe23, line 779:
```suggestion
  uint total_height = 0;
```

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+
+  ~DependencyGraph() {
+    SetVector<DependencyNode *> nodes;
+    for (auto root : roots)
+      gatherNodes(nodes, root);
+
+    for (auto node : nodes)
+      // ArgDependencyNodes are handled by the block and skipped here
+      if (!node->isLeaf() || !node->isQuantumOp() || node->isAlloc())
+        delete node;
```
bmhowe23, line 991: Where are the leaf nodes, quantum nodes and alloc nodes deleted?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+  }
+};
+
+class ArgDependencyNode : public DependencyNode {
+  friend class DependencyBlock;
+  friend class IfDependencyNode;
+
+protected:
+  BlockArgument barg;
+  uint argNum;
```
bmhowe23, line 1355:
```suggestion
  uint argNum = 0;
```

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+      llvm::outs() << qid;
+      printComma = true;
+    }
+    llvm::outs() << ": ";
+    associated->dump();
+  }
+
+  uint numTicks() override { return 0; }
+
+  bool isQuantumOp() override { return qids.size() > 0; }
```
bmhowe23, line 1484:
```suggestion
  bool isQuantumOp() override { return !qids.empty(); }
```

lib/Optimizer/Transforms/WiresToWiresets.cpp
```diff
@@ -75,18 +75,18 @@ struct AssignWireIndicesPass
   void runOnOperation() override {
     func::FuncOp func = getOperation();
 
+    // Only run on the entrypoint, the expectation is that inlining has been
+    // done already, so there should only be one kernel remaining.
```
schweitzpgi, line 79: Seems like this ought to run on pure device kernels too.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: I've addressed this in local changes I'll add soon.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: I think this was a mis-guided attempt to get my pass to play nicely with measurement ordering, so I don't think it's a necessarily thing to keep (and actually could lead to incorrect behavior even).

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+  uint getBegin() { return begin; }
+  uint getEnd() { return end; }
+};
+
+class LifeTimeAnalysis {
+private:
+  StringRef name;
+  SmallVector<LifeTime *> lifetimes;
+  SetVector<PhysicalQID> frame;
+  // DenseMap<VirtualQID, PhysicalQID> allocMap;
```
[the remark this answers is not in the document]
atgeller: I'll add that to my changes

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
atgeller: Not as currently written, every wire is expected to be a linear type. It is a good idea to check that the pass correctly refuses to run if it encounters a `!quake.control`

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: It's only quantum leaf nodes that are not allocs that are deleted here. As the comment says, `ArgDependencyNodes` are handled by the `DependencyBlock` destructor.

include/cudaq/Optimizer/CodeGen/Passes.h
```diff
@@ -36,6 +35,8 @@ void registerConvertToQIRPass();
 void addQIRProfilePipeline(mlir::OpPassManager &pm, llvm::StringRef convertTo);
 
 void addLowerToCCPipeline(mlir::OpPassManager &pm);
+void addWiresetToProfileQIRPipeline(mlir::OpPassManager &pm,
+                                    llvm::StringRef profile);
```
schweitzpgi, line 39: nit? I think we might use `mlir::StringRef` in the code more often, but I could be totally wrong. It may even be 50/50. Anyway, just a note about conformity.

include/cudaq/Optimizer/CodeGen/QIRAttributeNames.h
```diff
@@ -0,0 +1,27 @@
...
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+/// This file provides some common QIR attribute names for use in code gen.
+
+namespace cudaq::opt {
+
+static constexpr const char QIRRegisterNameAttr[] = "registerName";
```
schweitzpgi, line 15: This is pure genius! heh

include/cudaq/Support/TargetConfig.h
```diff
@@ -24,6 +24,8 @@ enum TargetFeatureFlag : unsigned {
   flagsFP64 = 0x0002,
   flagsMgpu = 0x0004,
   flagsMqpu = 0x0008,
+  flagsDepAnalysis = 0x0010,
+  flagsQPP = 0x0020,
```
schweitzpgi, line 28: Rather than exposing a bitset to the user (which I assume is what this does?), we should think about using strings/names in a comma separate list to make these option comprehensible to mortals.

bmhowe23: > Rather than exposing a bitset to the user (which I assume is what this does?),

That is not what this does.

> we should think about using strings/names in a comma separate list to make these option comprehensible to mortals.

This is what it does. See Adam's new test files to see how this is used.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
schweitzpgi: It's actually a helpful concept wrt scheduling. Obviously, there is no time left, but ...

Given:
```
  x [q0] q1
  y [q0] q2
```
If `q0` is a wire type, it appears y depends upon x. That's not true functionally however. So if `q0` was promoted to `control` type both permutations can be seen as correct orders of execution.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+//===----------------------------------------------------------------------===//
+// Generated logic
+//===----------------------------------------------------------------------===//
+namespace cudaq::opt {
+#define GEN_PASS_DEF_DEPENDENCYANALYSIS
+#include "cudaq/Optimizer/Transforms/Passes.h.inc"
+} // namespace cudaq::opt
+
+namespace {
+typedef size_t PhysicalQID;
```
schweitzpgi, line 37:
```suggestion
typedef std::size_t PhysicalQID;
```
global replace

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+      operand_idx--;
+    else
+      break;
+  }
+  return operand_idx;
+}
+
+class LifeTime {
+protected:
+  uint begin;
```
schweitzpgi, line 72:
```suggestion
  unsigned begin;
```
global replace

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+  }
+  return resultidx;
+}
+
+size_t getResultIDXFromOperandIDX(size_t operand_idx, Operation *op) {
+  if (isa<RAW_MEASURE_OPS>(op))
+    return 1;
+  if (isa<quake::SwapOp>(op))
+    return (operand_idx == 0 ? 1 : 0);
+  // Currently, all classical operands precede all quantum operands
```
schweitzpgi, line 60: This may not be a safe assumption in the long term. Maybe we need an assertion here?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+  }
+
+  uint getBegin() { return begin; }
+  uint getEnd() { return end; }
+};
+
+class LifeTimeAnalysis {
+private:
+  StringRef name;
+  SmallVector<LifeTime *> lifetimes;
```
schweitzpgi, line 104: curious: why is this a vector of raw pointers? what if I use a `SmallVector<LifeTime>` instread?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+    return pqids;
+  }
+
+  void reallocatePhysical(PhysicalQID phys, LifeTime *lifetime) {
+    lifetimes[phys] = lifetime;
+  }
+
+  size_t getCount() { return lifetimes.size(); }
+
+  void print() {
```
schweitzpgi, line 181:
```suggestion
  void dump() {
```
LLVM uses `dump` everywhere.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+class DependencyGraph;
+
+class DependencyNode {
+  friend class DependencyGraph;
+  friend class OpDependencyNode;
+  friend class IfDependencyNode;
+  friend class ArgDependencyNode;
+  friend class RootDependencyNode;
+  friend class TerminatorDependencyNode;
+  friend class InitDependencyNode;
```
schweitzpgi, line 201: This type is very gregarious.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Yes, @bmhowe23 and I had talked about doing that instead, which will prevent some potential memory leaks. I'll work on that after I address the documentation.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: I've made the API for DependencyNodes a little cleaner, so hopefully most of those can be removed.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

include/cudaq/Support/TargetConfig.h
schweitzpgi: OK, good.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -653,6 +855,9 @@ class OpDependencyNode : public DependencyNode {
     dependencies = _dependencies;
 
     quantumOp = isQuakeOperation(op);
+    // TODO: quake.discriminate is currently the only operation in the quake
+    // dialect that doesn't operate on wires. This will need to be updated if
+    // that changes.
```
bmhowe23, line 863: This isn't strictly true. For example, `quake.alloca` doesn't operate on wires either, but I suspect you aren't expecting to see those in this pass?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -785,12 +983,29 @@ class OpDependencyNode : public DependencyNode {
   }
 };
 
+/// A DependencyGraph is a DAG consisting of DependencyNodes somehow related by
+/// interaction. It contains useful metadata and functions for reasoning about
+/// and manipulating the DAG.
 class DependencyGraph {
 private:
+  // The set of root nodes in the DAG
+  // TODO: why is this a Set and `leafs` a Map? I don't really remember
```
bmhowe23, line 995: I suspect that they are `SetVector`s to guarantee repeatable iterating order based on insertion order and not pointer values?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -915,6 +1135,27 @@ class DependencyGraph {
...
+  /// Replaces the leaf for \p old_qid (if \p old_qid is part of the graph) with
+  /// \p new_leaf which has \p new_qid by removing the old leaf for \p old_qid
+  /// from the graph metadata and replacing the dependency on the old leaf with
+  /// \p new_leaf.
+  ///
+  /// If \p old_qid was not part of the graph, this has the effect of adding \p
+  /// new_leaf to the graph.
+  ///
+  /// Cleaning up the old leaf is the responsibility of the caller.
+  // TODO: this, replaceRoot, and replaceLeafAndRoot are all a mess.
```
bmhowe23, line 1150: > this, replaceRoot, and replaceLeafAndRoot are all a mess.

🤣

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -1212,6 +1527,7 @@ class DependencyGraph {
     }
   }
 
+  /// Tells the graph that
```
bmhowe23, line 1527: Incomplete sentence?

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
schweitzpgi: I'm thinking we should merge #2074 into this one. Unless there is a reason not to?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Most of the ones have been removed, and the remaining ones have a decent reason for staying.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+
+  uint numTicks() override { return 0; }
+  bool isQuantumOp() override { return true; }
+
+  Value getResult(uint resultidx) override {
+    assert(resultidx == 0 && "Illegal resultidx");
+    return wire;
+  }
+
+  ValueRange getResults() override { return ValueRange({wire}); }
```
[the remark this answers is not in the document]
atgeller: `getResults` was not used anywhere so I got rid of it.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+    }
+
+    printNode();
+
+    for (auto dependency : dependencies)
+      dependency->printSubGraph(tabIndex + 1);
+  }
+
+  virtual bool isRoot() { return successors.size() == 0; };
+  virtual bool isLeaf() { return dependencies.size() == 0; };
```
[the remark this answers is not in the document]
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed. I made the comment focus on the specific issues with it.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: It's possible, although since we're not worrying about measurement/result ordering I think predictable iterating order has gone out the window. My suspicion is that it is a vestigial trait, and that a map will make finding roots a bit faster. Right now, I think the roots are only really iterated through when trying to find a specific root anyway.

bmhowe23: Having repeatable outputs with repeatable inputs is very important in my opinion, so I'd still recommend keeping the predictable iterating order.

atgeller: I'll see where they are iterated through. Scheduling sorts them currently, so if it's just when looking them up I don't see a difference. If they're iterated through elsewhere too then it would make sense.

atgeller: Ah, it's also used for code generation, so I'll leave it as a SetVector and remove the TODO.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

include/cudaq/Optimizer/CallGraphFix.h
```diff
@@ -0,0 +1,47 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include "mlir/Analysis/CallGraph.h"
+
+namespace llvm {
+// FIXME: `GraphTraits` specialization for `const mlir::CallGraphNode *` in
```
schweitzpgi, line 14: MLIR eschews the use of `const`. https://mlir.llvm.org/docs/Rationale/UsageOfConst/

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -100,7 +100,7 @@ def ArgumentSynthesis : Pass<"argument-synthesis", "mlir::func::FuncOp"> {
     "mlir::cf::ControlFlowDialect"];
 }
 
-def BasisConversionPass : Pass<"basis-conversion", "mlir::ModuleOp"> {
+def BasisConversionPass: Pass<"basis-conversion", "mlir::ModuleOp"> {
```
schweitzpgi, line 103:
```suggestion
def BasisConversionPass : Pass<"basis-conversion", "mlir::ModuleOp"> {
```

include/cudaq/Optimizer/Transforms/Passes.td
schweitzpgi: I think we may need to auto-insert the `reset` ops. I don't know that this happens, per se, but it could be an issue.
```
  // gates
  mz qubit
  return_wire qubit
```
converted to
```
  // ...
  mz qubit
  // reuse qubit now, replacing qubit2
  // error? qubit needs to be set to |0>
  // ... [qubit2\qubit] ... 
  return_wire qubit
```

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+  /// - Step 2: blocks are resolved.
+  /// * First the nodes inside them are scheduled, assigning a cycle to every
+  ///   node such that each node is scheduled after all of its dependencies, and
+  ///   before all of its successors.
+  /// * Then, physical qubits are allocated and assigned to virtual wires based
+  ///   on lifetime information (i.e., which cycles the virtual wire is used
+  ///   in). This algorithm treats `if`s as "solid rectangles", where all qubits
+  ///   in use anywhere in the `if` are considered in use for the entire `if` by
+  ///   the parent scope. In other words, the lifetime analysis does not look
+  ///   inside `if`s.
```
schweitzpgi, line 3088: This considers qubits that are temporally live _only_ within the body of the `if` as "used by the if", right?

Abstractly, that is the biggest difference between an `if` macro operation and a slimmer quake operation. But necessarily imply that their quantum uses are live-in and their quantum outputs are live-out. The "bonus" use concept might be useful for operations that imply ancilla qubits as well.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+  /// - Step 3: Once its blocks are resolved, then the parent `if` is resolved.
+  /// * First, the allocations from the two inner blocks are combined/matched
+  ///   (respecting reuse within the blocks but with re-indexing allowed).
+  /// * Second, equivalent operations at the beginning/end of the then and else
+  ///   blocks are lifted to the parent context, before/after the `if`.
```
schweitzpgi, line 3093: I'm curious how the virtual qids interact with the determination of equivalence. If the two arms use different qid numberings, that doesn't necessarily mean that the prefixes aren't common, for example.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+      dependencies.push_back(visitValue(op->getOperand(i)));
+
+    DependencyNode *newNode;
+
+    if (auto init = dyn_cast<quake::BorrowWireOp>(op)) {
+      newNode = new InitDependencyNode(init);
+      vallocs++;
+    } else if (auto sink = dyn_cast<quake::ReturnWireOp>(op))
+      newNode = new RootDependencyNode(sink, dependencies);
+    else if (auto ifop = dyn_cast<cudaq::cc::IfOp>(op)) {
```
schweitzpgi, line 2972:
```suggestion
    } else if (auto sink = dyn_cast<quake::ReturnWireOp>(op)) {
      newNode = new RootDependencyNode(sink, dependencies);
    } else if (auto ifop = dyn_cast<cudaq::cc::IfOp>(op)) {
```

test/Quake/dependency-if-bug-lifting.qke
schweitzpgi: OK. Let's remove stale comments where we can.

test/Quake/dependency-classical-7.qke
```diff
@@ -0,0 +1,39 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --add-wireset --assign-wire-indices --dep-analysis %s | FileCheck %s
```
schweitzpgi, line 9: note: it looks like many/all of these tests are testing the same command-line. may need to fix the 1 func per module limitation, but should merge these tests when that happens.

targettests/execution/qubit_management_if-classical.cpp
schweitzpgi: Sounds good. But noting that it looks like this command line doesn't actually check the JIT IR dump.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+        if (!body) {
+          LLVM_DEBUG(
+              llvm::dbgs()
+              << "Encountered issue, backing out and skipping function\n");
+          engine.clean();
+          continue;
+        }
+
+        OpBuilder builder(func);
+        LifeTimeAnalysis set(cudaq::opt::topologyAgnosticWiresetName);
```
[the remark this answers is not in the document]
atgeller: First the final assignment of wires is determined (by performAnalysis), and only in the last step is new quake code generated. Also, because of `assign-wire-indices`, it should be `borrow_wire`s in and `borrow_wire`s out.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Right, any qubit used anywhere in the `if` is considered "in use" for the whole `if`, whether it is passed as a linear argument or only allocated in the body.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Virtual wires only allocated inside the `if` will be assigned physical qubits by the analysis, so equivalence should only look at the physical qubit assigned to those. For virtual wires allocated in some parent scope, we look at the qid to ensure that it is the same virtual wire.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+
+struct DependencyAnalysisPass
+    : public cudaq::opt::impl::DependencyAnalysisBase<DependencyAnalysisPass> {
+  using DependencyAnalysisBase::DependencyAnalysisBase;
+
+  /// DependencyAnalysis constructs a data structure representing the
+  /// quake AST, performs several analyses/optimizations, and then generates
+  /// a new quake AST based on the resulting data structure.
+  ///
+  /// First, the quake AST is walked, constructing a DependencyBlock for the
```
schweitzpgi, line 3064: Terminology question: Is "quake AST" just another way of referring to the IR? Or are we to take some addition nuance from it?

There isn't actually an abstract syntax tree (AST) in quake. Quake is a structured IR that is an SSA value graph. That graph is cyclic, has explicit and implicit use/def relations, can use symbolic relations, etc.

My wobbly inference is that this terminology is a way of viewing the order of operations in MLIR's basic blocks as not definitive w.r.t. the actual quantum constraints. That property is always true for classical code as well.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+      return DependencyNode::DependencyEdge{shadow_node, resultidx};
+    }
+
+    return DependencyNode::DependencyEdge{dnode, resultidx};
+  }
+
+  /// Cleans up semi-constructed dependency graph when backing out of running
+  /// DependencyAnalysis because of an encountered error
+  void clean() {
+    // TODO: clean up nodes
```
schweitzpgi, line 3049: Is this leaking memory?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Yes, it is referring to the IR. Would "quake code" (or perhaps "unordered quake code") be a more accurate phrasing?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+  /// This is not necessary to always perform, instead, one could only lift
+  /// allocs when lifting operations on those allocs, but it is very difficult
+  /// to do that safely.
+  ///
+  /// Assumes (and checks) that \p qubit is allocated in either/both the
+  /// then/else blocks.
+  ///
+  /// If the qubit is present in both child blocks, then the extra
+  /// alloc/de-alloc pair is cleaned up here.
+  void liftAlloc(PhysicalQID qubit, DependencyGraph *parent) {
```
schweitzpgi, line 2417: Is this only looking for common prefixes across the arms of `if`s? Or is there someplace that is doing tail fusion and sinking common tail suffixes?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Yeah, a bunch of `DependencyNode`s are getting allocated when building up the `DependencyGraph`, but if there's an illegal operation that causes the analysis to back out, those allocated `DependencyNode`s aren't freed currently. That's why I added the `clean()` function, but I didn't have time to write it. I think it will take an extra data structure to keep track of all the allocated `DependencyNode`s, but also remove the ones inside `IfDependencyNode`s, because the `IfDependencyNode` is responsible for cleaning those up. The tricky part is that some nodes could be inside an `if`, but if there's an illegal operation later in the `if`, then no `IfDependencyNode` would be created and the nodes would have to be explicitly cleaned up by the `DependencyAnalysisEngine`.

atgeller: Thinking about this a bit more, I think it may be as simple as collecting the set of nodes in `visitBlock` and then cleaning up there if an illegal operation is encountered.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: `tryLiftingBefore` looks for common prefixes (and `liftOpBefore` does the lifting), and `tryLiftingAfter` looks for common "postfixes" (and `liftOpAfter` does the lifting). `liftAlloc` lifts all physical allocations to the parent scope.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,3148 @@
...
+    DependencyGraph *new_graph = new DependencyGraph(terminator);
+    auto included = new_graph->getRoots();
+
+    // In debug mode, alert about dead code wires
+    // TODO: If an unused wire flows through an `if` with a useful wire,
+    //       then the unused wire is considered useful as the parent context
+    //       doesn't know that it doesn't interact with anything inside the if,
+    //       it would be nice to have a "hasInteraction" predicate inside `if`s
+    //       to be able to detect this case and do a better job of removing
+    //       unused wires.
```
schweitzpgi, line 2933: Do you actually see such a structure being created?

memtoreg should be generating a pruned SSI form for any wire values. That means any wire value that is live across the `if` but not used within the `if` wouldn't be part of the $\sigma$ or $\phi$ nodes of the `if` by construction.

That said, it's possible that ops get moved around and break the pruned SSI property.

atgeller: The situation I was (trying to) describe is a little trickier. Consider the following program
```
cudaq::qubit q,p;

if (some_condition) {
  x(p);
  x(q);
} else {
  y(p);
  y(q);
}

return mz(q);
```
Here, both `q` and `p` flow into (and are used inside) the `if`, so `p` is not treated as dead code, but it is actually dead code (well, according to DependencyAnalysis' definition of dead code) because it has no impact on the return value. This should be detectable (at least in a simpler case like this).

test/Quake/dependency-if-bug-classical.qke
atgeller: @bmhowe23 I don't think so, no. The requirement was added to prevent multiple kernel functions getting assigned duplicate virtual wires but then calling each other, so the virtual wires indices would no longer be unique. Function calls should be ruled out anyway, so I think it should be fine.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Fixed

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: If that helps to clarify the TODO, I can add the example to the TODO?

schweitzpgi: OK.

We have to stay at the same level of abstraction to discuss the point here.
```mlir
  p = alloca
  q = alloca
  [p', q'] = cc.if (...) (p, q) {
    // p,q used
  } else {
    // p,q used
 }
 mz q'
 dealloc p'  // this is not optional and a use
 dealloc q'
 return
}
```
Because of the dealloc, the only correct representation necessitates that both p and q remain alive across the entire `if`.

Now, the dealloc could be replicated in both arms of the if which would sink that operation into the `if`, but doesn't seem to optimize anything to me.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
schweitzpgi: Thanks, Adam.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
schweitzpgi: Assuming I was reading your mind correctly, I guess "quake IR" is what we are walking across, but we're really interested in the dependency information encoded in the IR. So maybe something like "quake dependency graph"? But then that sort of gets confusing with the on-the-side dependency graph being built here, which is also a dependence graph. (Naming stuff is hard.)

lib/Optimizer/Transforms/DependencyAnalysis.cpp
schweitzpgi: > we look at the qid
Which is the qid assigned in the function scope?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: The issue is that `if` uses both `p` and `q`, producing `p'` and `q'`, so the dependency graph will start searching from the return, see the operation on `p'`, and then see the `if` producing `p'`, from which it will see `q'`, so it will include `q'` in the graph, as if the `if` was any other quake operation which entangled `p` and `q`, but because it's an `if`, `p` and `q` may or may not interact in the body of the `if`, so it's not always the case that the graph needs to chase down all wires flowing through an `if` (there's no harm in doing so, it is just suboptimal).

schweitzpgi: OK.

Yes, the `if` must consider both arms in aggregate, and there has to be both arms. But, by virtue of this being a _conditional_ operation by definition, it need not be the case all qubits are used unconditionally. heh

Coincidentally, we can consider qubits in control positions to be "unused" in the sense that their quantum state is unchanged.

schweitzpgi: > If that helps to clarify the TODO, I can add the example to the TODO?

We can turn this thread into an issue.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
atgeller: Well, the qid assigned in the the parent scope (which may be the function scope).

targettests/execution/qubit_management_if-classical.cpp
atgeller: The JIT IR can't currently be dumped by the target. I think checking the pass statistics would also be a good testing approach (and would be a little easier to set up with nvq++ and the custom target).

schweitzpgi: Oh. Does setting `CUDAQ_DUMP_JIT_IR=1` in the environment not work?

lib/Optimizer/Transforms/DependencyAnalysis.cpp
schweitzpgi: I see. Thanks for clarifying.

test/Quake/dependency-if-bug-lifting.qke
atgeller: The comment has been updated to describe what regression the test is testing for

test/Quake/dependency-classical-7.qke
atgeller: Yeah I think all of `dependency-*` and `dependency-classical-*` can be merged. They can even be merged as is with `-split-input-file`. The other test files test for more specific things, but could also be merged.

test/Quake/dependency-if-bug-classical.qke
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

test/Quake/dependency-if-bug-classical.qke
bmhowe23: @atgeller to resolve this one. (We think it's easy.)

include/cudaq/Optimizer/Transforms/Passes.td
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

targettests/execution/qubit_management_if-1.cpp
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

targettests/execution/qubit_management_if-classical.cpp
bmhowe23: The current approach does not go through JIT. Will create a separate issue for this one (or add to a new grouped issue). The better approach would be to use statistics here, and check the statistics.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
```diff
@@ -0,0 +1,2561 @@
...
+      return physical;
+    }
+
+    // Fall back: allocate a new qubit
+    lifetimes.push_back(lifetime);
+    return lifetimes.size() - 1;
+  }
+
+public:
+  LifeTimeAnalysis(StringRef name) : name(name), lifetimes(), frame() {}
```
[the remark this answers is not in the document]
bmhowe23: @atgeller to do this one (if it's easy)

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue). This one may be challenging because there are many assumptions w.r.t. linear types right now. @atgeller to create short comments in the code for now.

targettests/execution/qubit_management_if-1.cpp
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

targettests/execution/qubit_management_if-1.cpp
bmhowe23: (Same as above.) Will create a separate issue for this one (or add to a new grouped issue).

lib/Optimizer/Transforms/WiresToWiresets.cpp
bmhowe23: @schweitzpgi - will you please write an issue for this one (separate from the other dep-analysis issue we're discussing).

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: @atgeller to add assertion here.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: Will create a separate issue for this one (or add to a new grouped issue).

test/Quake/dependency-if-bug-classical.qke
atgeller: Resolved.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

targettests/execution/qubit_management_if-1.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

targettests/execution/qubit_management_if-1.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

targettests/execution/qubit_management_if-classical.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

targettests/execution/qubit_management_if-1.cpp
bmhowe23: https://github.com/NVIDIA/cuda-quantum/issues/2181 written. Resolving this conversation.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
atgeller: Resolved

lib/Optimizer/Transforms/DependencyAnalysis.cpp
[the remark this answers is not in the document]
atgeller: Comments added. This mostly comes up when moving nodes around (specifically, the `getSuccessorForQID` function and all users of that function), and in the invariants about DependencyGraph construction (specifically, a single leaf/root pair per QID), though the invariants may still hold depending on how controls work.

lib/Optimizer/Transforms/DependencyAnalysis.cpp
bmhowe23: Addressed in f3abaa96f

lib/Optimizer/Transforms/WiresToWiresets.cpp
schweitzpgi: Created issue.

PR #2168, opened 2024-08-28
Noise model enhancements

1tnguyen:
### Description

(1) Noise model enhancements

- Support a match-all condition (noise on a gate, any qubits)

- Support noise defined via callback functions.

(2) Noise mapping for custom operations

- Move the custom operation map to the `common/` location so that it can be accessed from anywhere.

- Make the registration a bit earlier so that we know that an operation is registered before actually running that gate.

- Connect the Python registration to C++.

- Propagate the custom op name all the way back to the runtime, i.e., simulator.


python/cudaq/kernel/register_op.py
```diff
@@ -58,5 +59,8 @@ def kernel():
     # Make available to kernel builder object
     setattr(PyKernel, operation_name,
             partialmethod(__generalCustomOperation, operation_name))
+    # Let the runtime know about this registered operation.
+    # Note: the matrix generator/construction is not known by the ExecutionManager in this case since we don't expect the ExecutionManager to be involved.
+    cudaq_runtime.register_custom_operation(operation_name)
```
khalatepradnya, line 64: Q: Why do we need this?
(Is it for remote execution support?)

1tnguyen: The `NoiseModel` class has a validation check for gate name when adding noise channels. 
Previously, we only check against built-in gates. To support custom ops, I moved the custom gate map/registry to `CustomOp.h` in `common/` so that the `NoiseModel` can check against registered custom ops. Python has a different registration mechanism here; hence we need to propagate this to the C++ registry as well via this call.

khalatepradnya (commented):
These changes look good to me. Thanks, Thien!
I reviewed the tests, refactoring for custom operations, and Python bindings. (I defer on the specifics of the implementation.)
nit: Do we need a C++ example as well?

khalatepradnya (approved):
LGTM. 👍🏽

1tnguyen:
> These changes look good to me. Thanks, Thien! I reviewed the tests, refactoring for custom operations, and Python bindings. (I defer on the specifics of the implementation.) nit: Do we need a C++ example as well?

Thanks @khalatepradnya. I added a C++ example in https://github.com/NVIDIA/cuda-quantum/pull/2168/commits/01130bd1211885c86878f48f9d9ab7fb37741f73, which is similar to the Python example that I added.

PR #2179, opened 2024-08-30
Fix publishing error

sacpis:
This will fix the following publishing error

https://github.com/NVIDIA/cuda-quantum/actions/runs/10634730291/job/29485962342


bmhowe23 (approved):
LGTM. Thanks, @sacpis.

PR #2180, opened 2024-08-30
Fix NVQC regression test

sacpis:
Adding qvector_inti_from_vector to the skiplist.

This will fix NVQC integration test on main
https://github.com/NVIDIA/cuda-quantum/actions/runs/10626085666/job/29457118419

Seems like we missed to add this test in PR #2147.


bmhowe23:
@annagrin I added a note to #1931 to re-enable this test when it is ready. Please let me know if that's ok.

bmhowe23 (approved):

PR #2182, opened 2024-08-30
Update DependencyAnalysis to maintain uniqueness of QIDs when splitting during lowering

atgeller:
### Description

Uses a unique QID counter in `contractAllocsPass` to generate new QIDs for copied virtual allocations when splitting virtual wires between the then and else branch of an `if`. Only the final commit, be0ed707a1b2c1c03971431c3d434a38d42d207c, is different from the changes in #2163. The final commit contains the relevant bug fix.


atgeller:
@bmhowe23 @schweitzpgi here is the bug fix I mentioned. All the tests were passing locally, so hopefully it makes it through the gauntlet ok.

schweitzpgi:
@bmhowe23 Are we going to merge this with the "wile enchilada" PR or do it after?

bmhowe23:
> @bmhowe23 Are we going to merge this with the "whole enchilada" PR or do it after?

Since it passed the CI, I'd be ok cherry-picking the one commit (https://github.com/NVIDIA/cuda-quantum/commit/be0ed707a1b2c1c03971431c3d434a38d42d207c) into #2163 and closing this one.

bmhowe23:
I moved the relevant commits to #2163, so I think we can close this one.

PR #2187, opened 2024-09-03
CUDA 12 installer and wheels

bettinaheim:
This PR sets up the pipelines to build cuda-quantum-cu11 and cuda-quantum-cu12 wheels, and same for the installer. 
In a subsequent change, the goal is to additionally create a source distribution that will detect the CUDA version on the system and install either the -cu11 or -cu12 wheel (not covered in this PR). Some of the builds and checks changed here to account for the new wheel name will be updated to rely on that src distr in the next PR.


.github/workflows/config/validation_config.json
```diff
@@ -1,13 +1,6 @@
...
     [
-        {
-            "version": 3.8,
-            "operating_systems": 
-            [
-                "fedora:38"
-            ]
-        },
         {
             "version": 3.9,
```
1tnguyen, line 5: Could we also drop 3.9 in this PR?

.github/workflows/dev_environment.yml
```diff
@@ -35,10 +35,10 @@ on:
         required: false
         type: string
         description: Any additional parameters that should be passed to --cache-from (see Docker documentation for more detail).
-      toolchain:
+      build_config_id:
         required: false
         type: string
-        description: The toolchain used within the Docker image. The value is passed as build argument to Docker and used in tags and cache locations to distinguish different builds.
+        description: An string that uniquely identfies the toolchain and build configuration used within the Docker image. The value is used in tags and cache locations to distinguish different builds.
```
1tnguyen, line 41: nit: minor typos

```suggestion
        description: A string that uniquely identifies the toolchain and build configuration used within the Docker image. The value is used in tags and cache locations to distinguish different builds.
```

.github/workflows/prebuilt_binaries.yml
```diff
@@ -13,10 +13,14 @@ on:
       build_cache:
         type: string
         required: false
-      toolchain:
+      build_config_id:
+        required: false
+        type: string
+        description: An string that uniquely identfies the toolchain and build configuration used within the Docker image. The value is used in tags and cache locations to distinguish different builds.
```
1tnguyen, line 19: nit: minor typos
```suggestion
        description: A string that uniquely identifies the toolchain and build configuration used within the Docker image. The value is used in tags and cache locations to distinguish different builds.
```

1tnguyen (approved):
Looks great 👍 I just have one request of dropping Python 3.9 as well.

PR #2188, opened 2024-09-03
[core] Fix calling convention differences between libstdc++ and libc++

schweitzpgi:
for small structs.

In libc++, the compiler uses a "zero-cost" model of passing small struct values. In libstdc++, this was never fixed in order to maintain backwards compatibility with older compilers. So small structs are passed on the stack via references.  See https://tinyurl.com/bdhrkx9x for more information.

These patches fix nvq++ to adapt the calling convention to the header files in use at compile time.


bmhowe23 (approved):
LGTM

PR #2193, opened 2024-09-06
Add Clifford-only simulator (Stim)

bmhowe23:
Add https://github.com/quantumlib/Stim (v1.13.0) simulator for Clifford-only circuits.

The user requests it like other targets: `--target stim` and `cudaq.set_target('stim')`.


runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -0,0 +1,192 @@
...
+  std::mt19937_64 randomEngine;
+
+  /// @brief Grow the state vector by one qubit.
+  void addQubitToState() override { addQubitsToState(1); }
+
+  /// @brief Override the default sized allocation of qubits
+  /// here to be a bit more efficient than the default implementation
+  void addQubitsToState(std::size_t qubitCount,
+                        const void *stateDataIn = nullptr) override {
+    return;
```
1tnguyen, line 37: nit: should we throw an error here when `stateDataIn` is not null?

runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -0,0 +1,192 @@
...
+    std::transform(gateName.begin(), gateName.end(), gateName.begin(),
+                   ::toupper);
+    std::vector<std::uint32_t> stimTargets;
+
+    // These CUDA-Q rotation gates have the same name as Stim "reset" gates.
+    // Stim is a Clifford simulator, so it doesn't actually support rotational
+    // gates. Throw exceptions if they are encountered here.
+    if (gateName == "RX" || gateName == "RY" || gateName == "RZ")
+      throw std::runtime_error(fmt::format(
+          "Gate not supported by simulator: {}", task.operationName));
```
1tnguyen, line 86: nit: should we add a TODO comment here about the possibility of handling `+/- pi/2` rotations for future references?

runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -0,0 +1,192 @@
...
+    stimCircuit.safe_append_u(
+        "R", std::vector<std::uint32_t>{static_cast<std::uint32_t>(index)});
+  }
+
+  /// @brief Sample the multi-qubit state.
+  cudaq::ExecutionResult sample(const std::vector<std::size_t> &qubits,
+                                const int shots) override {
+    std::vector<std::uint32_t> stimTargetQubits;
+    for (auto q : qubits)
+      stimTargetQubits.push_back(static_cast<std::uint32_t>(q));
```
1tnguyen, line 147: nit
```suggestion
    std::vector<std::uint32_t> stimTargetQubits(qubits.begin(), qubits.end());
```

1tnguyen (approved):
LGTM 👍

runtime/nvqir/stim/StimCircuitSimulator.cpp
bmhowe23: Updated in 20343fd07

runtime/nvqir/stim/StimCircuitSimulator.cpp
bmhowe23: Updated in 20343fd07

runtime/nvqir/stim/StimCircuitSimulator.cpp
bmhowe23: Updated in 20343fd07

bmhowe23:
> These changes look great! Thanks, Ben. I don't see any new tests... It seems all of the C++ tests will be run except the ones which are skipped with `#ifndef CUDAQ_BACKEND_STIM`. What about Python - may be we should add a couple which showcase which gates are supported and behavior on unsupported gates.

Thanks for the suggestion, Pradnya. I added some Stim-specific backend tests in 20343fd07. Please take a look and let me know if that's what you had in mind, or if you'd like to see something else, too.

khalatepradnya (approved):
👍

Issue #2196, opened 2024-09-06
FuseWithConstantArray on ExtractValueOp always returns success()


amccaskey:
`FuseWithConstantArray::matchAndRewrite` is returning success even if no canonicalization is applied. I'm not 100% sure, but I think that final return statement should be a `return failure();`. I noticed this when I tried to add more canonicalization patterns for `ExtractValueOp` and they would not run, and my guess was that it was because the first pass, `FuseWithConstantArray` was always returning success.

schweitzpgi:
Closing. Should be fixed.

PR #2198, opened 2024-09-06
Fix issue #2196

schweitzpgi:
Canonicalization patterns should return failure if they don't modify the IR.


amccaskey (approved):
+1 Thanks!

PR #2199, opened 2024-09-06
Enable pure quantum struct usage in kernels with restrictions

amccaskey:
This PR addresses the need to support struct types with only quantum members, e.g. 
```cpp
struct test {
  cudaq::qview<> q;
  cudaq::qview<> r;
};
__qpu__ void entry_ctor() {
  cudaq::qvector q(2), r(2);
  test tt(q, r);
  h(tt.r[0]);
}
```
or 
```python 
@dataclass
class patch:
    q : cudaq.qview
    r : cudaq.qview 

@cudaq.kernel 
def entry():
    q = cudaq.qvector(2)
    r = cudaq.qvector(2)
    p = patch(q, r)
    h(p.r[0])
```
Pure quantum struct types are treated as value types in the MLIR (no loading or storing). 

We enforce the following restrictions in both C++ and Python with thrown errors with descriptive messages: 
- hybrid quantum-classical structs (contains both classical and quantum types)
- quantum struct return types 
- recursive usage of quantum structs
- quantum structs with user-defined methods 

A canonicalizer is also added to ensure lowering to static circuit code is possible.


schweitzpgi:
Maybe I forgot to comment before, but I think it will be handy to have a distinct type for these other than a cc.struct type. It would let us fine tune the semantics better. They are "quantum" types in the "quantum memory" space. They only interact with quake dialect artifacts. They can't be loaded and stored from and to regular memory, etc.

PR #2200, opened 2024-09-09
Add enumerated noise_model_type to noise interface

bmhowe23:
This is related to #2235 but will require additional review.

Reference: CUDA-Q Spec Change MR 9 (internal only).
 * Approved by @amccaskey on Oct 1
 * Approved by @bettinaheim on Oct 4


runtime/common/NoiseModel.h
```diff
@@ -106,6 +117,12 @@ class kraus_channel {
   }
 
 public:
+  /// @brief Noise type enumeration
+  noise_model_type noise_type = noise_model_type::unknown;
+
+  /// @brief Noise probability
+  real probability = 0.0;
```
amccaskey, line 124: Do we envision a need for multiple noise parameters?

bmhowe23: > Do we envision a need for multiple noise parameters?

Yes, I suppose that could happen for some future noise models. For example, if we wanted to mimic the [PAULI_CHANNEL_1](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#the-pauli_channel_1-instruction) noise option from Stim, then that would need multiple parameters. Do you think we should change `probability` to a vector of doubles?

amccaskey: Yea I think that would be a good idea, maybe call it parameters

bmhowe23: Updated in aa7982853

1tnguyen (approved):
LGTM 👍

PR #2205, opened 2024-09-11
ORCA asynchronous sampling

Omar-ORCA:
### Description
Support for asynchronous sampling for ORCA backend


khalatepradnya:
> To fix Basic content checks CI job, please run these commands locally to see the issue on your local machine
> 
> ```
> bash scripts/run_clang_format.sh
> bash scripts/run_all_spelling.sh
> ```

Fixed.

docs/sphinx/using/backends/hardware.rst
```diff
@@ -312,6 +312,16 @@ configuration.
 
   export ORCA_ACCESS_URL="https://<ORCA API Server>"
 
+
+Sometimes the requests to the PT-1 require an authentication token. This token can be set as an
```
1tnguyen, line 316: I just want to confirm that the authentication token is only required 'sometimes' (not always).

python/extension/CUDAQuantumExtension.cpp
```diff
@@ -178,6 +178,23 @@ PYBIND11_MODULE(_quakeDialects, m) {
...
       py::arg("n_samples") = 10000);
+  orcaSubmodule.def(
+      "sample_async",
+      py::overload_cast<std::vector<std::size_t> &, std::vector<std::size_t> &,
+                        std::vector<double> &, std::vector<double> &, int>(
+          &cudaq::orca::sample_async),
+      "Performs Time Bin Interferometer (TBI) boson sampling experiments on "
+      "ORCA's backends",
+      py::arg("input_state"), py::arg("loop_lengths"), py::arg("bs_angles"),
+      py::arg("ps_angles") = nullptr, py::arg("n_samples") = 10000);
```
1tnguyen, line 189: I wonder if the default value of `nullptr` is valid in this case. Do we have a test/example _not_ passing `ps_angles` to this call to make sure it works as intended?

runtime/common/Executor.cpp
```diff
@@ -55,4 +55,40 @@ Executor::execute(std::vector<KernelExecution> &codesToExecute) {
...
+  auto config = serverHelper->getConfig();
+
+  std::vector<cudaq::details::future::Job> ids;
+  cudaq::info("Job created, posting to {}", jobPostPath);
+
+  // Post it, get the response
+  auto response = client.post(jobPostPath, "", job, headers);
+  cudaq::info("Job posted, response was {}", response.dump());
+
+  // // Add the job id and the job name.
```
1tnguyen, line 80: nit:
```suggestion
  // Add the job id and the job name.
```

runtime/common/ServerHelper.h
```diff
@@ -103,6 +104,19 @@ class ServerHelper : public registry::RegisteredType<ServerHelper> {
...
+  virtual ServerJobPayload createJob(cudaq::orca::TBIParameters params) {
+    std::vector<ServerMessage> jobs;
+    ServerMessage job;
+    jobs.push_back(job);
+
+    std::map<std::string, std::string> headers;
+
+    // Return a tuple containing the job path, headers, and the job message
+    auto ret = std::make_tuple("", headers, jobs);
+    return ret;
```
1tnguyen, line 117: I think it makes more sense to throw an exception in this case rather than returning a dummy payload like this.

runtime/common/ServerHelper.h
```diff
@@ -14,6 +14,7 @@
 #include "Future.h"
 #include "MeasureCounts.h"
 #include "Registry.h"
+#include "cudaq/orca.h"
```
1tnguyen, line 17: I think the need to add orca-specific dependency at this level (common) is undesirable.

Since we would override the `cudaq::QPU` interface for ORCA anyway, we could encapsulate all changes via inheritance, e.g.,

```
class OrcaServerHelper: public ServerHelper {
public:
// Add a method, not override
ServerJobPayload createJob(cudaq::orca::TBIParameters params);
}

class OrcaExecutor: public Executor {

// Added method, no need to introduce to base clas
details::future execute(cudaq::orca::TBIParameters params,
                          const std::string &kernelName) {

						 ....
auto casted = dynamic_cast<OrcaServerHelper*>(serverHelper);
assert(casted);
auto [jobPostPath, headers, jobs] = casted->createJob(params);						 
						  
				....		  
}

}

class OrcaRemoteRESTQPU : public cudaq::QPU {


  std::unique_ptr<OrcaExecutor> executor; // use OrcaExecutor
	
	
  OrcaRemoteRESTQPU() : QPU() {
    ....
    executor = std::make_unique<OrcaExecutor>();
  }
}
```

At a cost of a `dynamic_cast`, we could encapsulate/localize all ORCA-specific changes. 
@amccaskey interested to hear your opinion as well.

runtime/common/Executor.h
```diff
@@ -42,6 +42,12 @@ class Executor : public registry::RegisteredType<Executor> {
   /// @brief Execute the provided quantum codes and return a future object
   /// The caller can make this synchronous by just immediately calling .get().
   details::future execute(std::vector<KernelExecution> &codesToExecute);
+
+  /// @brief Execute the provided orca quantum parameters and return a future
+  /// object The caller can make this synchronous by just immediately calling
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 93779972c3565f272211b2c625ff7255bbc19c20

runtime/cudaq/platform/orca/OrcaServerHelper.cpp
```diff
@@ -0,0 +1,135 @@
...
+  job["input_state"] = params.input_state;
+  job["loop_lengths"] = params.loop_lengths;
+  job["bs_angles"] = params.bs_angles;
+  job["ps_angles"] = params.ps_angles;
+  job["n_samples"] = params.n_samples;
+
+  jobs.push_back(job);
+
+  // Return a tuple containing the job path, headers, and the job message
+  auto ret = std::make_tuple(baseUrl + "v1/submit", getHeaders(), jobs);
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 93779972c3565f272211b2c625ff7255bbc19c20

runtime/common/ServerHelper.h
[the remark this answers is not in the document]
khalatepradnya: Fixed in 93779972c3565f272211b2c625ff7255bbc19c20

docs/sphinx/using/backends/hardware.rst
Omar-ORCA: I can confirm that is the case

python/extension/CUDAQuantumExtension.cpp
Omar-ORCA: The example files (`examples/cpp/providers/orca.cpp` and python equivalent) are _not_ passing `ps_angles`

1tnguyen: I think the below overload (with no `ps_angles` argument) masked the default value for `ps_angles` as it was probably the overload selected in this case.
Since we have a specific overload for not having the `ps_angles` argument, do we need a default value for it here as that would have no effect (i.e., without providing a value for it, another overload is provided)?

runtime/common/ServerHelper.h
Omar-ORCA: Throwing exception

runtime/common/ServerHelper.h
Omar-ORCA: I am happy with either solution

khalatepradnya: Addressed in 9b979f2bd2ee0c5ad558c70211416a4aff26f442.

runtime/common/Executor.cpp
1tnguyen, line 1: nit: the change below (removing the EOF line break) may be accidental.

1tnguyen (approved):
LGTM 👍 Thanks @Omar-ORCA and @khalatepradnya!

runtime/common/Executor.cpp
khalatepradnya: Fixed in b8b2de129513d9d25dbd00145561d1fa95c5ff0d

docs/sphinx/examples/cpp/providers/orca.cpp
```diff
@@ -73,8 +85,27 @@ int main() {
...
+    std::ofstream out("saveMe.json");
+    out << async_results;
+  }
+
+  // Then come back and read it in later.
+  cudaq::async_result<cudaq::sample_result> readIn;
+  std::ifstream in("saveMe.json");
+  in >> readIn;
+
+  sleep_for(200ms); // wait for the job to be processed
```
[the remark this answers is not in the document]
khalatepradnya: Will address offline.

Issue #2210, opened 2024-09-12
trotter example not running on remote simulators


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following command is failing with a segmentation fault with the new launch kernel:

```
nvq++ --enable-mlir -v trotter_kernel_mode.cpp -o trotter.x  --target remote-mqpu -fkernel-exec-kind=2 && ./trotter.x
```

### Steps to reproduce the bug

see [trotter_kernel_mode.cpp](https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/examples/cpp/other/trotter_kernel_mode.cpp)

Command to run:

`nvq++ --enable-mlir -v trotter_kernel_mode.cpp -o trotter.x  --target remote-mqpu -fkernel-exec-kind=2 && ./trotter.x`

### Expected behavior

Run successfully

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #2211, opened 2024-09-12
Fix argument conversion issues

annagrin:
### Description
Fix issues with argument conversion

#### Issues (discovered while running trotter example on remote sim):
- Passing pauli_word as a parameter caused a seg fault due to casting to string
- Passing a vector of pauli_word as a parameter caused a seg fault due to incorrect size of element calculation
- `ArgumentConverter.get_drop_front` dropping all arguments after the first due to unsigned overflow

#### Added tests
  - for argument conversion, to test for more argument combinations
  - for remote simulation, to test for non-lazy-evaluation scenario

### Note:
  - lazy evaluation scenario is still not working for remote sim (Fix is to be added in a separate PR.)

Towards: https://github.com/NVIDIA/cuda-quantum/issues/2210


runtime/common/ArgumentConversion.cpp
```diff
@@ -199,8 +200,8 @@ Value dispatchSubtype(OpBuilder &builder, Type ty, void *p, ModuleOp substMod,
         return {};
       })
       .Case([&](cudaq::cc::CharspanType strTy) {
-        return genConstant(builder, *static_cast<const std::string *>(p),
-                           substMod);
+        return genConstant(
+            builder, (*static_cast<cudaq::pauli_word *>(p)).str(), substMod);
```
schweitzpgi, line 204:
```suggestion
            builder, static_cast<cudaq::pauli_word *>(p)->str(), substMod);
```
Does this work as well?

runtime/common/ArgumentConversion.cpp
```diff
@@ -224,6 +225,11 @@ Value genConstant(OpBuilder &builder, cudaq::cc::StdvecType vecTy, void *p,
   auto eleTy = vecTy.getElementType();
   auto elePtrTy = cudaq::cc::PointerType::get(eleTy);
   auto eleSize = cudaq::opt::getDataSize(layout, eleTy);
+  if (isa<cudaq::cc::CharspanType>(eleTy)) {
+    // char span type (i.e. pauli word) is a `vector<char>`
```
schweitzpgi, line 229: It shouldn't be, but that hasn't been fixed yet.

runtime/common/ArgumentConversion.cpp
```diff
@@ -406,8 +413,9 @@ void cudaq::opt::ArgumentConverter::gen_drop_front(
   if (numDrop >= arguments.size())
     return;
   std::vector<void *> partialArgs;
+  int drop = numDrop;
   for (void *arg : arguments) {
-    if (numDrop--) {
+    if (drop-- > 0) {
```
schweitzpgi, line 418:
```suggestion
    if (numDrop > 0) {
      numDrop--;
```

runtime/test/test_argument_conversion.cpp
```diff
@@ -15,40 +15,67 @@
...
-func.func @__nvqpp__mlirgen__testy(%0: )#" +
-                     typeName + R"#() {
-  call @callee(%0) : ()#" +
-                     typeName + R"#() -> ()
-  return
-})#";
+void doTest(mlir::MLIRContext *ctx, std::vector<std::string> &typeNames,
+            std::vector<void *> args, std::size_t startingArgIdx = 0) {
+
+  // Create code
```
schweitzpgi, line 26: Why do we need a stringstream?

runtime/test/test_argument_conversion.cpp
```diff
@@ -15,40 +15,67 @@
...
+  auto code = os.str();
 
   // Create the Module
   auto mod = mlir::parseSourceString<mlir::ModuleOp>(code, ctx);
   llvm::outs() << "Source module:\n" << *mod << '\n';
   cudaq::opt::ArgumentConverter ab{"testy", *mod};
 
   // Create the argument conversions
-  ab.gen(args);
+  ab.gen_drop_front(args, startingArgIdx);
```
schweitzpgi, line 60: Why are we removing the original test that uses `gen` here?

schweitzpgi (commented):
Yeah, the pauli_word is implemented incorrectly in the C++ headers. It should be a std::string, and that should be fixed at some point.

I'm not sure why there are so many changes to test_argument_conversion.cpp. They don't look necessary to me and possibly like they aren't even testing the same functionality any longer.

runtime/test/test_argument_conversion.cpp
annagrin: The `gen_drop_front` wasn't tested and had an issue, so I was making sure were are going through all the paths (`gen_drop_front` calls `gen` eventually, but if you would prefer I can create separate tests for `gen_drop_front` that would look the same except calling `gen_drop_front` instead of `gen`.

runtime/test/test_argument_conversion.cpp
annagrin: What would be the preference?

runtime/common/ArgumentConversion.cpp
annagrin: ack.

runtime/common/ArgumentConversion.cpp
annagrin: Yes, updated. Thank!

runtime/common/ArgumentConversion.cpp
annagrin: created an issue: https://github.com/NVIDIA/cuda-quantum/issues/2213

runtime/test/test_argument_conversion.cpp
annagrin: I switched to `llvm::raw_string_ostream` (similar to what we use in creating argument substitutions when calling argumnet conversion). Let me know if there is something better to use.

runtime/test/test_argument_conversion.cpp
annagrin: I restored the original tests and only now use `gen_drop_front` in an additional `test_combinations` set of tests, to make sure the issues are fixed.

runtime/common/ArgumentConversion.cpp
annagrin: Done, thanks!

runtime/test/test_argument_conversion.cpp
schweitzpgi: Originally, it was just concatenating pieces of strings, right?

runtime/test/test_argument_conversion.cpp
```diff
@@ -146,7 +192,7 @@ void test_scalars(mlir::MLIRContext *ctx) {
   // clang-format on
 
   {
-    std::string x = "Hi, there!";
```
schweitzpgi, line 149: Does string not work at all now?

annagrin: I would not but we currently don't allow strings to be passed as arguments.

PR #2212, opened 2024-09-12
Allow state synthesis for remote simulators

annagrin:
### Description
Allow state synthesis for remote simulators.

- Copy pointers that we save in remote simulation state, to avoid them being overridden by other code.
- Add tests for lazy evaluation with state pointers
- Add trotter test on remote simulators.

Note: do not enable the trotter_kernel_mode.cpp yet as it requires using new synthesis on remote platform.

<del>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2211</del>
Closes: https://github.com/NVIDIA/cuda-quantum/issues/2210


1tnguyen (approved):
LGTM 👍 Thanks, @annagrin!

schweitzpgi (approved):
+1

PR #2214, opened 2024-09-17
Enable Python/C++ interop via exposed JIT functionality

amccaskey:
Enable one to pass Python kernels to library entry-point kernels defined in C++. Enable one to expose C++ pure-device kernels to Python.


include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -46,9 +46,10 @@ std::unique_ptr<mlir::Pass> createRaiseToAffinePass();
 std::unique_ptr<mlir::Pass> createUnwindLoweringPass();
 
 std::unique_ptr<mlir::Pass>
-createPySynthCallableBlockArgs(const std::vector<std::string> &);
+createPySynthCallableBlockArgs(const std::vector<std::string> &,
```
schweitzpgi, line 49: Should be using `SmallVector`, `SmallVectorImpl`, or `ArrayRef` here, no? Maybe `StringRef`?

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
```diff
@@ -59,6 +59,41 @@ class ReplaceCallIndirect : public OpConversionPattern<func::CallIndirectOp> {
   }
 };
 
+class ReplaceCallCallable
+    : public OpConversionPattern<cudaq::cc::CallCallableOp> {
+public:
+  const std::vector<std::string> &names;
+  const std::map<std::size_t, std::size_t> &blockArgToNameMap;
```
schweitzpgi, line 66: Ought to use llvm ADTs in this code.

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
```diff
@@ -59,6 +59,41 @@ class ReplaceCallIndirect : public OpConversionPattern<func::CallIndirectOp> {
...
+  LogicalResult
+  matchAndRewrite(cudaq::cc::CallCallableOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto callableOperand = adaptor.getCallee();
+    auto module = op->getParentOp()->getParentOfType<ModuleOp>();
+    if (auto blockArg = dyn_cast<BlockArgument>(callableOperand)) {
+      auto argIdx = blockArg.getArgNumber();
+      auto replacementName = names[blockArgToNameMap.at(argIdx)];
+      auto replacement = module.lookupSymbol<func::FuncOp>(
+          "__nvqpp__mlirgen__" + replacementName);
```
schweitzpgi, line 83: Should use the named string from the header instead of a string literal here. I think I added a helper function somewhere?

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
```diff
@@ -59,6 +59,41 @@ class ReplaceCallIndirect : public OpConversionPattern<func::CallIndirectOp> {
...
+      if (!replacement) {
+        op.emitError("Invalid replacement function " + replacementName);
```
schweitzpgi, line 85: Should the pass itself fail here? Or should it just not do the replacement? In other words, should this pass just be best effort (like most passes) or should it be an ultimatum: it shall be replaced?

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
```diff
@@ -97,10 +132,13 @@ class UpdateQuakeApplyOp : public OpConversionPattern<quake::ApplyOp> {
 class PySynthCallableBlockArgs
     : public cudaq::opt::PySynthCallableBlockArgsBase<
           PySynthCallableBlockArgs> {
+private:
+  bool removeBlockArg = false;
```
schweitzpgi, line 136: Q: callables are always going to be entry-block arguments, so if we're removing them from there, we'll also have to update the FuncOp signature, right?

python/CMakeLists.txt
```diff
@@ -64,3 +64,5 @@ if(CUDAQ_BUILD_TESTS)
     message(FATAL_ERROR "CUDA Quantum Python Warning - CUDAQ_BUILD_TESTS=TRUE but can't find numpy or pytest modules required for testing.")
   endif()
 endif()
+
+add_subdirectory(runtime/interop)
```
schweitzpgi, line 68: nit: whining about eof

python/runtime/interop/PythonCppInterop.cpp
```diff
@@ -0,0 +1,100 @@
...
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include "PythonCppInterop.h"
+#include "cudaq.h"
+
+namespace cudaq {
```
schweitzpgi, line 11: I wonder if it might not be a good idea to put the python specific code in a sub-namespace like `cudaq::python` just to declutter things and prevent us from accidentally reusing the same names for C++ and Python both.

python/CMakeLists.txt
amccaskey: fixed

python/runtime/interop/PythonCppInterop.cpp
amccaskey: good idea, fixed

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
amccaskey: hadn't thought of that originally, updated to just not perform the replacement, no error thrown

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
amccaskey: Fixed

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
amccaskey: Fixed

include/cudaq/Optimizer/Transforms/Passes.h
amccaskey: fixed

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
amccaskey: correct, which is done with the op.eraseArguments() call. These changes build off what was there previously for the python ast bridge / py_alt_launch_kernel. There the funcop update is not required, so I made this an optional flag to retain backwards compatibility

lib/Optimizer/Transforms/PySynthCallableBlockArgs.cpp
```diff
@@ -77,13 +110,11 @@ class UpdateQuakeApplyOp : public OpConversionPattern<quake::ApplyOp> {
     auto ctx = op.getContext();
     if (auto blockArg = dyn_cast<BlockArgument>(callableOperand)) {
       auto argIdx = blockArg.getArgNumber();
-      auto replacementName = names[blockArgToNameMap.at(argIdx)];
+      auto replacementName = names[blockArgToNameMap[argIdx]];
       auto replacement = module.lookupSymbol<func::FuncOp>(
-          "__nvqpp__mlirgen__" + replacementName);
-      if (!replacement) {
-        op.emitError("Invalid replacement function " + replacementName);
+          "__nvqpp__mlirgen__" + replacementName.str());
```
schweitzpgi, line 115:
```suggestion
          cudaq::runtime::cudaqGenPrefixName + replacementName.str());
```

schweitzpgi (approved):
LGTM. Thanks for the updates.

PR #2216, opened 2024-09-18
[core] Separate compilation, supporting C++ host side function references.

schweitzpgi:
These changes add a new wrapper class to support interfacing between kernels. Kernels on the device side have their own names and calling conventions. Meanwhile, the C++ host code can capture references to these kernels. The C++ compiler assumes it can erase the functions completely and degenerate any reference to them as a pointer. Furthermore, it may inline or wrap this code in a thicket of template instantiations. But in order to be useful on the device side, the CUDA-Q runtime must be able to determine which kernel was wrapped in C++ code, lookup the device side code, and "link-time" optimize these calls, doing calling convention conversions, etc. Furthermore, what must be done is flavored by the execution environment.

Update the call paths of hybridLaunchKernel so that we can use the new argument synthesis instead of falling back on and failing with quake synthesis A couple of init state tests are marked XFAIL as they don't seem to behave with the new argument synthesis..

### DIFF
--- a/runtime/cudaq/qis/qkernel_ref.h
+++ b/runtime/cudaq/qis/qkernel_ref.h
@@ -1,0 +1,195 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <type_traits>
+
+#ifdef CUDAQ_LIBRARY_MODE
+
+namespace cudaq {
+
+/// In library mode, the quake compiler is not involved. To streamline things,
+/// just have qkernel_ref alias the std::function template class.
+template <typename Sig>
+using qkernel_ref = std::function<Sig>;
+
+} // namespace cudaq
+
+#else
+
+namespace cudaq {
+
+namespace details {
+class QKernelDummy;
+
+template <typename R, typename... As>
+class QKernelInterface {
+public:
+  virtual ~QKernelInterface() = default;
+
+  virtual R dispatch(As...) = 0;
+  virtual void *getEntry() = 0;
+};
+
+template <typename R, typename F, typename... As>
+class QKernelHolder : public QKernelInterface<R, As...> {
+public:
+  using EntryType =
+      std::conditional_t<std::is_class_v<F>, R (QKernelDummy::*)(As...),
+                         R (*)(As...)>;
+
+  QKernelHolder() : entry{nullptr}, callable{} {}
+  QKernelHolder(const QKernelHolder &) = default;
+  QKernelHolder(QKernelHolder &&) = default;
+
+  explicit QKernelHolder(F &&f) : callable(std::forward<F>(f)) {
+    if constexpr (std::is_same_v<F, R (*)(As...)>) {
+      entry = f;
+    } else {
+      setEntry(&F::operator());
+    }
+  }
+  explicit QKernelHolder(const F &f) : callable(f) {
+    if constexpr (std::is_same_v<F, R (*)(As...)>) {
+      entry = f;
+    } else {
+      setEntry(&F::operator());
+    }
+  }
+
+  QKernelHolder &operator=(const QKernelHolder &) = default;
+  QKernelHolder &operator=(QKernelHolder &&holder) = default;
+
+  R dispatch(As... as) override { return std::invoke(callable, as...); }
+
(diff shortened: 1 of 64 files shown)


targettests/SeparateCompilation/emulate.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+// clang-format off
+// RUN: split-file %s %t && \
+// RUN: nvq++ -target quantinuum -emulate -fno-set-target-backend -c %t/emulib.cpp -o %t/emulibx.o && \
+// RUN: nvq++ -target quantinuum -emulate -c %t/emuuser.cpp -o %t/emuuserx.o && \
+// RUN: nvq++ -target quantinuum -emulate %t/emulibx.o %t/emuuserx.o -o %t/emux.a.out && \
+// RUN: %t/emux.a.out | FileCheck %s
+// clang-format on
+
+// Catch-22. We need to use argument synthesis, but if I enable that in the
```
amccaskey, line 17: Is this comment old? I don't see an XFAIL comment, does the test pass with -emulate?

targettests/SeparateCompilation/emulate.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+__qpu__ void
+userKernel(const cudaq::qkernel_ref<void(cudaq::qvector<> &)> &init) {
+  cudaq::qvector q(2);
+  init(q);
+}
+
+int main() {
+  cudaq::sample(10, userKernel,
+                cudaq::qkernel_ref<void(cudaq::qvector<> &)>{dunkadee});
```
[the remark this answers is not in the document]
schweitzpgi: I had something broken in the deduction guides at one point. I don't think this is required any longer.

targettests/SeparateCompilation/emulate.cpp
schweitzpgi: Yes and yes. I'll remove this. I got the JIT parts all working here now.

1tnguyen (approved):
LGTM 👍

runtime/cudaq/qis/qkernel_ref.h
```diff
@@ -0,0 +1,195 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include <functional>
+#include <iostream>
```
amccaskey, line 12: We should drop the `<iostream>` include, but we will need to add `<memory>` for the `unique_ptr` usage.

targettests/execution/state_init_err_runtime.cpp
```diff
@@ -8,11 +8,14 @@
...
 // clang-format off
 // Note: change |& to 2>&1| if running in bash
-// RUN: nvq++ %cpp_std %s -o %t --target quantinuum --emulate && %t |& FileCheck %s
+// RUN: nvq++ %cpp_std %s -o %t -target quantinuum -emulate && %t |& FileCheck %s
 // Note: change |& to 2>&1| if running in bash
-// RUN: nvq++ %cpp_std --enable-mlir --target remote-mqpu --remote-mqpu-url localhost:9999 %s -o %t && %t |& FileCheck %s
+// RUN: nvq++ %cpp_std --enable-mlir -target remote-mqpu --remote-mqpu-url localhost:9999 %s -o %t && %t |& FileCheck %s
 // clang-format on
 
+// FIXME! Requires init_state to be implemented via new argument synthesis
```
annagrin, line 16: This test checks for error message on quantum devices, ie verifies that it is indeed giving "unsupported " message. The message itself is probably changed between the old and the new synthesis, can the test to be updated with the new message instead of being disabled?

targettests/Remote-Sim/qvector_init_from_state.cpp
```diff
@@ -14,6 +14,9 @@
 // RUN: nvq++ %cpp_std --enable-mlir --target remote-mqpu -fkernel-exec-kind=2 %s -o %t && %t | FileCheck %s
 // clang-format on
 
+// FIXME: state handling with new argument synthesis?
```
annagrin, line 17: This is supported actually (this test is aimed to verify the support). Does this test need to be run without `-fkernel-exec-kind=2` , or does it need to be removed and default is used? (see the command line above).

What was the error btw? If the error gives an "unsupported" message, then we must be running old synthesis...

targettests/execution/state_init_err_runtime.cpp
schweitzpgi: Easy fix. Thanks.

targettests/Remote-Sim/qvector_init_from_state.cpp
schweitzpgi: Intermittent issue in CI? (This test works fine for me. `shrug`)

PR #2217, opened 2024-09-19
[custom op] Two-qubit operation synthesis

khalatepradnya:
Towards: https://github.com/NVIDIA/cuda-quantum/issues/1475

Synthesize two-qubit custom operations using KAK decomposition


lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -28,48 +31,372 @@ namespace cudaq::opt {
...
+  ///                +1 −1 −1 −1
+  ///                +1 −1 +1 +1
+  static Eigen::Matrix4cd GammaT;
+  GammaT << 1, 1, 1, 1, 1, 1, -1, -1, -1, 1, -1, 1, 1, -1, -1, 1;
+  GammaT /= 4;
+  return GammaT;
+}
+
+std::tuple<Eigen::Matrix4d, Eigen::Matrix4cd, Eigen::Matrix4d>
+bidiagonalize(const Eigen::Matrix4cd &matrix) {
```
1tnguyen, line 185: nit: maybe worth adding some comments about what 'bidiagonalize' means and the return tuple.

khalatepradnya: Addressed in 5ea69709ab689d05b8ef2607ac2726ae47e38474.

1tnguyen (approved):
LGTM 👍

PR #2221, opened 2024-09-19
Improve state synthesis tests for remote simulators and fix flakes

annagrin:
Fix flaky tests introduced in previous PRs (#2212 and #2211)


targettests/Remote-Sim/qvector_init_from_state.cpp
```diff
@@ -31,6 +31,7 @@ __qpu__ void test_init_large_state() {
 
 __qpu__ void test_state_param(cudaq::state* state) {
   cudaq::qvector q1(state);
+  x(q1);
```
khalatepradnya, line 34: Q: Does addition of this operation make the test deterministic?

targettests/Remote-Sim/qvector_init_from_state.cpp
```diff
@@ -77,68 +78,57 @@ int main() {
...
   }
-
 // CHECK: Passing state created from data as argument (kernel mode)
-// CHECK: 000
-// CHECK: 100
-
 // CHECK: 011
 // CHECK: 111
 
+// CHECK: 000
```
khalatepradnya, line 85: Q: Do we need `CHECK-DAG` directive if the output order varies?

khalatepradnya (approved):
LGTM. 
Minor questions.

targettests/Remote-Sim/qvector_init_from_state.cpp
annagrin: No, this part is just for making the test better - if it changes the output every time it's called we can test that iterations fip the qubit values and not just get stale values.

targettests/Remote-Sim/qvector_init_from_state.cpp
annagrin: The output order should not vary, the change in the output is due to the addition of the gate x() in `test_state_param`. Apologies for confusion, I'll change the title of the PR to "improve tests and remove flakes"

PR #2224, opened 2024-09-20
Disable tests using new argument synthesis for NVQC

annagrin:
Disable `qvector_init_from_state_lazy.cpp`, `test_trotter.cpp` for NVQC until new argument synthesis becomes default (or we support passing compilation arguments to NVQC tests).

See: https://github.com/NVIDIA/cuda-quantum/issues/2146


annagrin:
> Disable `qvector_init_from_state_lazy.cpp`, `test_trotter.cpp` for NVQC until new argument synthesis becomes default (or we support passing compilation arguments to NVQC tests).
> 
> See: #2146

@bettinaheim @amccaskey @bmhowe23 could you please review? This is a fix that unblock NVQC integration and regression testing.

amccaskey (approved):

PR #2225, opened 2024-09-23
Fermioniq implementation

MarkusPfundstein:
This PR adds integration with the Fermioniq simulator


runtime/common/Future.cpp
```diff
@@ -49,7 +49,8 @@ sample_result future::get() {
     } else {
       // For each register, add the results into result.
       for (auto &regName : c.register_names()) {
-        results.emplace_back(c.to_map(regName), regName);
+        // results.emplace_back(c.to_map(regName), regName);
```
1tnguyen, line 52:
```suggestion
```

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
```diff
@@ -0,0 +1,233 @@
...
+    completeLaunchKernel(kernelName, std::move(codes));
+  }
+
+  /// @brief Extract the Quake representation for the given kernel name and
+  /// lower it to the code format required for the specific backend. The
+  /// lowering process is controllable via the configuration file in the
+  /// platform directory for the targeted backend.
+  std::vector<cudaq::KernelExecution>
+  getQuakeCodes(const std::string &kernelName, void *kernelArgs,
+                 const std::vector<void *> &rawArgs) {
```
1tnguyen, line 66: This can be simplified to reuse the base class' `lowerQuakeCode`, for example,

```
cudaq::ExecutionContext defaultContext("sample", 1);
auto* originalContext = executionContext;
if (executionContext->name == "observe")
  executionContext = &defaultContext; // use this context to skip observe ansatz generation
  
codes = lowerQuakeCode(....);

// Restore execution context 
executionContext =  originalContext;
// We can also check that codes.size() == 1
// Add the spin json 
if (executionContext->name == "observe") {
    auto spin = executionContext->spin.value();
    auto user_data = nlohmann::json::object();
    .....
      
    codes[0].user_data  = user_data;
}
```

This would help reduce a lot of duplication in setting up the IR pipeline below.

MarkusPfundstein: I tried this initially. The problem is that [BaseRemoteRESTQPU::lowerQuakeCode](https://github.com/NVIDIA/cuda-quantum/blob/ebaf92040cc49d1ea91acc25a9d26a792b0c695b/runtime/common/BaseRemoteRESTQPU.h#L478) does some particular things when the executionContext name equals "observe". It splits the circuit into multiple circuits and we don't want that for the FermioniqQPU as we have a different way of calculating the observables. So in my implementation, i just add the module like it would be a "sample" call.

So I had to implement my own function. I named it getQuakeCodes in order to not confuse the compiler.

I agree that a more elegant fix would be to re-use as much code as possible from lowerQuakeCodes. One idea might be to add class methods to BaseRemoteRESTQPU that then can be overwritten in the subclasses. But I guess thats not a trivial change.

MarkusPfundstein:

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
1tnguyen: > I tried this initially. The problem is that [BaseRemoteRESTQPU::lowerQuakeCode](https://github.com/NVIDIA/cuda-quantum/blob/ebaf92040cc49d1ea91acc25a9d26a792b0c695b/runtime/common/BaseRemoteRESTQPU.h#L478) does some very specific things when the executionContext->name == "observe". It splits the circuit into multiple circuits and we don't want that for the FermioniqQPU as we have a different way of calculating the observables. So in my implementation, i add the module in the same way as it would be a "sample" call.


In the [above](https://github.com/NVIDIA/cuda-quantum/pull/2225#discussion_r1786886535) code suggestion, this part of the code was meant to address the above issue

```
cudaq::ExecutionContext defaultContext("sample", 1);
auto* originalContext = executionContext;
if (executionContext->name == "observe")
  executionContext = &defaultContext; // use this context to skip observe ansatz generation
```

In particular, we set the `executionContext` to a placeholder context of name "sample" before calling `BaseRemoteRESTQPU::lowerQuakeCode`.

1tnguyen:
@MarkusPfundstein Just fyi, you can format the Python file with `yapf --style google --recursive -i <filename>`.

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
MarkusPfundstein: ok clear. will try it out.

MarkusPfundstein: Hey Thien, I did it and it worked!

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+}
+
+/// Refresh the api key and refresh-token
+void FermioniqServerHelper::refreshTokens(bool force_refresh) {
+  std::mutex m;
+  std::lock_guard<std::mutex> l(m);
+  RestClient client;
+
+  if (!force_refresh) {
+    auto now = std::chrono::high_resolution_clock::now();
```
1tnguyen, line 287: Is there a reason we use `std::chrono::high_resolution_clock` here but `std::chrono::system_clock` for the `tokenExpTime`.
This caused the compile error with Clang (the installer CI).

docs/sphinx/examples/python/providers/fermioniq.py
```diff
@@ -0,0 +1,25 @@
+import cudaq
+
+# You only have to set the target once! No need to redefine it
+# for every execution call on your kernel.
+cudaq.set_target(
+    "fermioniq", **{
```
1tnguyen, line 6: Could you please put `"fermioniq"` on the same line? We have a regex in `validate_container.sh` to parse Python target name in example files. Looks like it missed this one due to the line break.

lib/Optimizer/CodeGen/Passes.cpp
```diff
@@ -77,6 +77,17 @@ static void addIonQPipeline(OpPassManager &pm) {
   pm.addPass(createBasisConversionPass(options));
 }
 
+static void addFermioniqPipeline(OpPassManager &pm) {
+  // TO-DO Fermioniq: Figure this out!
```
1tnguyen, line 81: Could you please update this comment?

runtime/cudaq/platform/fermioniq/fermioniq.yml
```diff
@@ -0,0 +1,40 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+name: fermioniq
+description: "CUDA-Q target for Orca."
```
1tnguyen, line 10:
```suggestion
description: "CUDA-Q target for Fermioniq."
```

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+  cudaq::sample_result processResults(ServerMessage &postJobResponse,
+                                      std::string &jobId) override;
+
+  void refreshTokens(bool force_refresh);
+
+#if 0
+  /// @brief Update `passPipeline` with architecture-specific pass options
+  void updatePassPipeline(const std::filesystem::path &platformPath,
+                          std::string &passPipeline) override;
+#endif
```
1tnguyen, line 86: Do we still need to keep this block? I'd suggest to remove unused code.

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+  /// @brief Return next results polling interval
+  std::chrono::microseconds
+  nextResultPollingInterval(ServerMessage &postResponse) override;
+
+private:
+  /// @brief RestClient used for HTTP requests.
+  // RestClient client;
```
1tnguyen, line 94:
```suggestion
```

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+                                      std::string &jobID) {
+  cudaq::debug("processResults for job: {}", jobID);
+
+  /*
+  auto &output_names = outputNames[jobID];
+  for (auto &[result, info] : output_names) {
+    cudaq::info("Qubit {} Name {}", info.qubitNum,
+                info.registerName);
+  }
+  */
```
1tnguyen, line 397:
```suggestion
```

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+  auto metadata = response_json.at("metadata");
+  // cudaq::info("metadata: {}", metadata.dump());
+  auto output = response_json.at("emulator_output");
+
+  cudaq::sample_result sample_result;
+
+  // to-do: restrict to 1 circuit
+  for (const auto &it : output.items()) {
+    // cudaq::info("result: {}", it.value().dump());
+    // int circuit_number = it.value().at("circuit_number");
```
1tnguyen, line 420:
```suggestion
```

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+  }
+
+  // We will also make registers for each result using output_names.
+  for (auto &[result, info] : output_names) {
+    cudaq::sample_result singleBitResult = sample_result.get_marginal({result});
+    ExecutionResult executionResult{singleBitResult.to_map(),
+                                    info.registerName};
+    sample_result.append(executionResult);
+  }
+#endif
```
1tnguyen, line 479: We can just remove this block if it's not relevant.

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
```diff
@@ -0,0 +1,512 @@
...
+  if (config.find("noise_model") != config.end()) {
+    backendConfig[CFG_NOISE_MODEL_KEY] = config["noise_model"];
+  }
+
+  if (config.find("bond_dim") != config.end()) {
+    backendConfig[CFG_BOND_DIM_KEY] = config.at("bond_dim");
+  }
+
+  refreshTokens(true);
+}
```
1tnguyen, line 155: Would you mind adding a "Fermioniq" section in https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/using/backends/hardware.rst to describe these configurations? Thanks!

scripts/validate_container.sh
```diff
@@ -81,7 +81,7 @@ available_backends=`\
         qpu=${platform##* }
         requirements=$(cat $file | grep "gpu-requirements:")
         gpus=${requirements##* }
-        if [ "${qpu}" != "remote_rest" ] && [ "${qpu}" != "orca" ] && [ "${qpu}" != "NvcfSimulatorQPU" ] \
+        if [ "${qpu}" != "remote_rest" ] && [ "${qpu}" != "fermioniq" ] && [ "${qpu}" != "orca" ] && [ "${qpu}" != "NvcfSimulatorQPU" ] \
```
1tnguyen, line 84: Since we don't run these examples in the CI, could you please add a job to the `.github/workflows/integration_tests.yml` to run those? 

This is a [sample](https://github.com/NVIDIA/cuda-quantum/blob/b0396b6b75cf3dfcf72515b6d0c249a8085f44c0/.github/workflows/integration_tests.yml#L486) for reference.

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
MarkusPfundstein: thanks. I indeed saw that and will update it.

docs/sphinx/examples/python/providers/fermioniq.py
MarkusPfundstein: check

lib/Optimizer/CodeGen/Passes.cpp
MarkusPfundstein: done

runtime/cudaq/platform/fermioniq/fermioniq.yml
MarkusPfundstein: done

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
MarkusPfundstein: done

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
MarkusPfundstein: done

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
MarkusPfundstein: done

runtime/cudaq/platform/fermioniq/helpers/FermioniqServerHelper.cpp
MarkusPfundstein: done

docs/sphinx/using/backends/simulators.rst
```diff
@@ -479,6 +479,88 @@ Specific aspects of the simulation can be configured by defining the following e
...
+        nvq++ --target fermioniq --fermioniq-project-id <project_id> src.cpp ...
+
+    To specify the bond dimensions, you can pass the ``fermioniq-bond-dim`` parameter.
+
+    .. code:: bash
+
+        nvq++ --target fermioniq --fermioniq-bind-dim 10 src.cpp ...
+
+
+.. tab:: Python
```
1tnguyen, line 527: The rendered doc looks good.
One minor thing: could you please move the Python tab ahead of the C++ tab? This is a CUDA-Q convention of having the Python tab as the default view.

MarkusPfundstein: done

scripts/validate_container.sh
MarkusPfundstein: I added this.

.github/workflows/integration_tests.yml
```diff
@@ -528,6 +528,56 @@ jobs:
           fi
         shell: bash
 
+      - name: Submit to Fermioniq test server
+        if: (success() || failure()) && (inputs.target == 'fermioniq' || github.event_name == 'schedule' || inputs.target == 'nightly')
```
1tnguyen, line 532: Thanks @MarkusPfundstein.
This looks good. Since `inputs.target` is a `choice` option, we'd need to add `fermioniq` to the list (line 25 of this file) as well.

MarkusPfundstein: Ok will done. Thx

1tnguyen:
I've tried the `fermioniq.cpp`,  `fermioniq.py`, and `fermioniq_observables.py` examples. They all worked.

1tnguyen (approved):
LGTM 👍

runtime/common/MeasureCounts.h
```diff
@@ -230,6 +229,9 @@ class sample_result {
   CountsDictionary
   to_map(const std::string_view registerName = GlobalRegisterName) const;
 
+  ExecutionResult get_result_for_register(
+      const std::string_view registerName = GlobalRegisterName) const;
```
1tnguyen, line 233: @amccaskey @boschmitt This is an _added_ method to the `sample_result` type (user-facing), hence I'd like to check if it's okay.

The rationale is that the `to_map` method only returns the count map; this one returns the full `ExecutionResult`, which may contain the `expectationValue`.

boschmitt: Seems a bit weird. At least if the idea behind `sample_result` is to wrap `ExecutionResult` as stated by its documentation. You would be basically breaking the abstraction. @amccaskey might have better insights regarding this.

MarkusPfundstein: Hi. I changed this back and used Thiens solution which doesnt break the abstraction.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -28,6 +28,7 @@ DOI
 Deuteron
 Doxygen
 FPGAs
+Fermioniq
```
bmhowe23, line 31: Nit: if `fermioniq` is an allowed word, then we don't need to explicitly add the capitalized version of this word, so I would recommend removing this line.
```suggestion
```

MarkusPfundstein: Ok that is fine with me.

MarkusPfundstein:
@bmhowe23 Hi, might it be possible to run CI and then merge it if all passes ? :-) Thanks

Issue #2219, opened 2024-09-19
Can `mx` and `my` be used with `sample`?


kshyatt-aws:
I'm trying to translate some simple kernels to OpenQASM2. Sampling a kernel which measures in the Pauli-Z basis works fine:

```c++
#include <cudaq.h>

template <std::size_t N>
struct ghz {
  auto operator()() __qpu__ {

    // Compile-time sized array like std::array
    cudaq::qarray<N> q;
    h(q[0]);
    for (int i = 0; i < N - 1; i++) {
      x<cudaq::ctrl>(q[i], q[i + 1]);
    }
    mz(q);
  }
};

int main() {

  auto kernel = ghz<10>{};
  auto counts = cudaq::sample(kernel);

  if (!cudaq::mpi::is_initialized() || cudaq::mpi::rank() == 0) {
    counts.dump();

    // Fine grain access to the bits and counts
    for (auto &[bits, count] : counts) {
      printf("Observed: %s, %lu\n", bits.data(), count);
    }
  }

  return 0;
}
```

The compile-and-translation pipeline I used was:
```bash
cudaq-quake ghz.cpp | cudaq-opt --pass-pipeline="builtin.module(canonicalize,lambda-lifting,apply-op-specialization,func.func(memtoreg{quantum=0}),cc-loop-normalize,cc-loop-unroll)"  | cudaq-translate --convert-to=openqasm2 &> ghz.qasm
```

However, if I change my kernel to (or `my(q)` on the final line):
```c++
template <std::size_t N>
struct ghz {
  auto operator()() __qpu__ {

    // Compile-time sized array like std::array
    cudaq::qarray<N> q;
    h(q[0]);
    for (int i = 0; i < N - 1; i++) {
      x<cudaq::ctrl>(q[i], q[i + 1]);
    }
    mx(q);
  }
};
```
I see the following error (from the initial `cudaq-quake` pass, I think?):
```bash
ghz.cpp:11:5: error: no matching function for call to 'mx'
    mx(q);
    ^~
/usr/local/llvm/include/c++/v1/__functional/invoke.h:391:28: note: in instantiation of member function 'ghz<10>::operator()' requested here
_LIBCPP_CONSTEXPR decltype(std::declval<_Fp>()(std::declval<_Args>()...))
                           ^
[...]

PR #2231, opened 2024-09-26
[runtime] adding mx and my support for qarrays

cabreraam:
### Description

~~This is a work in progress, but I wanted to go ahead and try~~ This PR  addresses issue #2219.

Upon testing this, I found that the issue arises here:
```
/usr/local/cudaq/include/cudaq/qis/qubit_qis.h:760:23: note: candidate function not viable: no known conversion from 'cudaq::qarray<10UL>' to 'qubit &' (aka 'qudit<2> &') for 1st argument
inline measure_result mx(qubit &q) {
                      ^
```

So, I implemented support for `mx` and `my`. I'm posting this before adding any test cases just to see if I'm on the right track, here.

Feedback appreciated!

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCTypes.td
+++ b/include/cudaq/Optimizer/Dialect/CC/CCTypes.td
@@ -283,6 +283,27 @@ def cc_CharSpanType
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// MeasureHandleType - opaque, word-sized classical value
+//===----------------------------------------------------------------------===//
+
+def cc_MeasureHandleType : CCType<"MeasureHandle", "measure_handle"> {
+  let summary = "An opaque word-sized handle for a measurement event.";
+  let description = [{
+    The IR alias of the C++ `cudaq::measure_handle` class. Semantically an
+    opaque, word-sized classical value (an `i64` payload) whose only meaningful
+    consumer is `quake.discriminate`. The bridge emits this type so that
+    verification passes can reject entry-point kernels that leak handles across
+    the host-device boundary; a late lowering pass replaces it with `i64`
+    before QIR emission.
+
+    ```mlir
+      %mh = quake.mz %q name "mh" : (!quake.ref) -> !cc.measure_handle
+      %b  = quake.discriminate %mh : (!cc.measure_handle) -> i1
+    ```
+  }];
+}
+
 def IsSpanLikePred : CPred<"isa<cudaq::cc::SpanLikeType>($_self)">;
 def SpanningType : Type<IsSpanLikePred, "dynamic length type">;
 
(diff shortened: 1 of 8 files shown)


cabreraam:

runtime/cudaq/qis/qubit_qis.h
```diff
@@ -791,6 +791,45 @@ std::vector<measure_result> mz(QubitRange &q) {
...
+#else
+template <
+    typename QubitRange,
+    typename = std::enable_if_t<!std::is_same_v<
+        std::remove_reference_t<std::remove_cv_t<QubitRange>>, cudaq::qubit>>>
+#endif
+std::vector<measure_result> my(QubitRange &q) {
+  std::vector<measure_result> b;
+  for (auto &qq : q) {
+    r1(-M_PI_2, qq);
```
amccaskey, line 807: Here I think you don't want the basis change operations, since you are calling the single qubit `my` which will apply the basis change already.

amccaskey:
Thanks for looking into this issue @cabreraam . Would you mind also adding a test to `unittests/` and a FileCheck test to the `test/AST-Quake` folder? You can just use the failing snippet from Issue tracker.

amccaskey:
Also while you're here, there are other `mz` overloads that we'll likely want to implement for `mx`, `my`. Lines 777-827

runtime/cudaq/qis/qubit_qis.h
cabreraam: Will do; thanks for the tip!

cabreraam:
> Thanks for looking into this issue @cabreraam . Would you mind also adding a test to `unittests/` and a FileCheck test to the `test/AST-Quake` folder? You can just use the failing snippet from Issue tracker.

I will do this!

> Also while you're here, there are other `mz` overloads that we'll likely want to implement for `mx`, `my`. Lines 777-827

Sounds good; I'll look into it.

cabreraam:
@amccaskey @schweitzpgi 

Added a couple of `lit` tests. I tried to remove all of the superfluous stuff and just check for the most important bit of `cudaq-quake`'s result. Let me know if you want to include all of the other setup MLIR stuff as well.

Additionally, #2219 wants to lower to OpenQASM 2.0. However, I'm not sure of a corresponding OQ2 instruction that measures with respect to `y` and `z`. (I'm using the OpenQASM 2.0 spec and the [relevant `cuda-quantum` source code](https://github.com/NVIDIA/cuda-quantum/blob/f5ab479598deb67c580257ca3d348f9cde4eca44/lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp#L329-L353) to try to deduce if I can lower this.)

cabreraam:
I'll be out for a couple of weeks so I'll be unable to respond to feedback during that time.

PR #2236, opened 2024-09-30
[C++ bridge] Add support for move constructor of std::vector.

schweitzpgi:
Allow the use of the move constructor so that cases such as

	std::vector<int> vectorValue(N);
	...
	return vectorValue;

will compile and simple vectors of arithmetic types may be used.  Note that this requires the CUDA-Q runtime to actually return these values to the caller rather than dropping them on the floor after a launched kernel has deactivated.  That is not always the case, however, since some launchers do not expect any return value(s).

### DIFF
--- a/lib/Optimizer/Transforms/GenKernelExecution.cpp
+++ b/lib/Optimizer/Transforms/GenKernelExecution.cpp
@@ -1811,7 +1811,6 @@ public:
       });
       for (auto *op : deadCalls)
         op->erase();
-      module.dump();
 
       LLVM_DEBUG(llvm::dbgs() << "final module:\n" << module << '\n');
     }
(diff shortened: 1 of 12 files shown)


amccaskey (approved):
tested locally, lgtm

schweitzpgi:
Jobs took so long, the cache was purged out from underneath them.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1776,11 +1778,47 @@ class GenerateKernelExecution
...
+            auto sv = builder.create<cudaq::cc::StdvecInitOp>(loc, ty, buffPtr,
+                                                              buffSize);
+            call.getResult(0).replaceAllUsesWith(sv);
+            deadCalls.push_back(call);
+          }
+        }
+      });
+      for (auto *op : deadCalls)
+        op->erase();
+      module.dump();
```
bmhowe23, line 1814: @schweitzpgi - is this intentional?

PR #2243, opened 2024-10-02
Fix Photonics `get_state`

Omar-ORCA:
Fixes `get_state` of the photonics simulator


docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
```diff
@@ -0,0 +1,87 @@
...
+ * Copyright (c) 2022 - 2024 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// Compile and run with:
+// ```
+// nvq++ --target photonics-cpu photonics_tbi_get_state.cpp -o
```
schweitzpgi, line 11: nit: shouldn't this have a `\` at the end of the line?

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
```diff
@@ -0,0 +1,87 @@
...
+// photonics_tbi_get_state.x && ./photonics_tbi_get_state.x
+// ```
+
+#include "cudaq.h"
+#include "cudaq/photonics.h"
+
+#include <iostream>
+
+// Global variables
+static const std::size_t one{1};
```
schweitzpgi, line 21:
```suggestion
static constexpr std::size_t one{1};
```

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+};
+
+template <typename T>
+void LinearSpacedArray(std::vector<T> &xs, T min, T max, std::size_t N) {
+  T h = (max - min) / static_cast<T>(N - 1);
+  typename std::vector<T>::iterator x;
+  T val;
+  for (x = xs.begin(), val = min; x != xs.end(); ++x, val += h) {
```
schweitzpgi, line 60:
```suggestion
  
  T val;
  for (auto x = xs.begin(), val = min; x != xs.end(); ++x, val += h) {
```
`x` may not be the clearest name, since it is a quantum gate name.

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
```diff
@@ -0,0 +1,87 @@
...
+    *x = val;
+  }
+}
+
+int main() {
+  std::size_t n_loops{2};
```
schweitzpgi, line 66: nit: Much of the code follows using the `= <init-expr>;` style. It would be nice for this example file to stay consistent.

docs/sphinx/examples/cpp/providers/photonics_tbi_sample.cpp
```diff
@@ -76,9 +76,5 @@ int main() {
   auto counts = cudaq::sample(1000000, TBI{}, bs_angles, ps_angles, input_state,
                               loop_lengths);
 
-  for (auto &[k, v] : counts) {
-    std::cout << k << ":" << v << " ";
-  }
-  std::cout << std::endl;
-  return 0;
+  counts.dump();
 }
```
schweitzpgi, line 80: bug: main return an `int` value.

runtime/cudaq/qis/managers/photonics/PhotonicsExecutionManager.cpp
```diff
@@ -151,242 +134,116 @@ class PhotonicsExecutionManager : public cudaq::BasicExecutionManager {
...
+                photonic_simulator()->beam_splitter(parameters[0], localC,
+                                                    localT);
+              })
+        .Case("phase_shift",
+              [&]() {
+                photonic_simulator()->phase_shift(parameters[0], localC,
+                                                  localT[0]);
+              })
+        .Default([&]() {
+          if (cudaq::customOpRegistry::getInstance().isOperationRegistered(
```
khalatepradnya, line 187: Can you add tests to cover custom operations?

runtime/cudaq/qis/managers/photonics/photonics_qis.h
```diff
@@ -10,44 +10,42 @@
 #pragma once
 
 #include "common/ExecutionContext.h"
-#include "cudaq/qis/qarray.h"
+// #include "cudaq/qis/qarray.h"
```
khalatepradnya, line 13: TODO: Clean up such commented-out lines of code.

docs/sphinx/examples/cpp/providers/photonics_tbi_sample.cpp
Omar-ORCA: Applies to most cpp files in folder

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
Omar-ORCA: Need to use `typename std::vector<T>::iterator x;`
rename xs to array and x to array_it

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
Omar-ORCA: Now using the `= <init-expr>; `style for photonics examples

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
Omar-ORCA: `&&` was missing

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
Omar-ORCA: Made changes

runtime/nvqir/photonics/PhotonicCircuitSimulator.h
```diff
@@ -0,0 +1,1217 @@
...
+  bool isDoublePrecision() const { return !isSinglePrecision(); }
+
+}; // PhotonicCircuitSimulator
+
+/// @brief The PhotonicCircuitSimulatorBase is the type that is meant to
+/// be subclassed for new simulation strategies. The separation of
+/// PhotonicCircuitSimulator from PhotonicCircuitSimulatorBase allows simulation
+/// sub-types to specify the floating point precision for the simulation
+template <typename ScalarType>
+class PhotonicCircuitSimulatorBase : public PhotonicCircuitSimulator {
```
khalatepradnya, line 260: Should the parent class be named with suffix 'Base'?

Omar-ORCA: Following the way CircuitSimulator and CircuitSimulatorBase  were implemented in `CircuitSimulator.h`
`class CircuitSimulatorBase : public CircuitSimulator`

runtime/cudaq/cudaq.cpp
```diff
@@ -408,6 +409,17 @@ void set_random_seed(std::size_t seed) {
   platform.onRandomSeedSet(seed);
 }
 
+/// @brief Note: a seed value of 0 will cause broadcast operations to use
+/// std::random_device (or something similar) as a seed for the PRNGs, so this
+/// will not be repeatable for those operations.
+void set_photonic_random_seed(std::size_t seed) {
```
khalatepradnya, line 415: Q: Can we try to reuse the existing `cudaq::set_random_seed` API?
May need to correspondingly update `getSimulator`.

Omar-ORCA: I removed `set_photonic_random_seed`

docs/sphinx/examples/cpp/providers/photonics_tbi_sample.cpp
Omar-ORCA: photonics cpp files now return 0

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
```diff
@@ -0,0 +1,82 @@
...
+  const std::size_t zero = 0;
+  std::size_t sum_loop_lenghts{
+      std::accumulate(loop_lengths.begin(), loop_lengths.end(), zero)};
+
+  std::size_t n_beam_splitters = n_loops * ::n_modes - sum_loop_lenghts;
+
+  std::vector<double> bs_angles(n_beam_splitters);
+  std::vector<double> ps_angles(n_beam_splitters);
+
+  LinearSpacedArray(bs_angles, M_PI / 3, M_PI / 6, n_beam_splitters);
```
amccaskey, line 73: Just reading through this quickly, but can't you use `cudaq::linspace(min, max, N)` instead of writing your own?

runtime/nvqir/photonics/PhotonicGates.h
```diff
@@ -0,0 +1,199 @@
...
+    throw std::invalid_argument("received invalid value, n <= 30");
+  }
+  return FACTORIAL_TABLE[n];
+}
+
+/// @brief Computes the Kronecker delta of two values
+int _kron(int a, int b) {
+  if (a == b)
+    return 1;
+  else
```
amccaskey, line 69: don't need the `else` after the return

amccaskey:
One big question I have is what does the PhotonicsCircuitSimulator add beyond what is in CircuitSimulator? I wonder if you could craft a CircuitSimulator subtype that does what you need? I would have thought so if you craft the execution manager to only use the applyCustomUnitary operation.

docs/sphinx/examples/cpp/providers/photonics_tbi_get_state.cpp
khalatepradnya: Fixed in e64e62c098d96e9461708caced91445e370b9ef6

runtime/nvqir/photonics/PhotonicGates.h
khalatepradnya: Fixed in 14afd309415d6186111e9b955f7d852d3903d3b5

Omar-ORCA:
> One big question I have is what does the PhotonicsCircuitSimulator add beyond what is in CircuitSimulator? I wonder if you could craft a CircuitSimulator subtype that does what you need? I would have thought so if you craft the execution manager to only use the applyCustomUnitary operation.

The main difference is that the PhotonicsCircuitSimulator works with qudits of different levels.
The number of levels depend on the number of photons in the circuit, therefore different kernels will have different qudit levels and the size of the gates will be different as well.

runtime/cudaq/qis/managers/photonics/PhotonicsExecutionManager.cpp
Omar-ORCA: Custom operation were removed in #[7da6e7d](https://github.com/NVIDIA/cuda-quantum/pull/2243/commits/7da6e7dbea71e2b1bb538b99864c7ae36e507842)

boschmitt:
I cannot make much sense of what is going on in this PR. (It seems that the latest changes broke the diff.) That being said, the tile "Fix Photonics `get_state`" doesn't seems to describe what this is doing: from what I can tell this is adding a whole parallel infrastructure for handling `qudit`-based computation.

The previous photonics work was supposed to be a stopgap until qudits became better supported in the stack. This PR seems to be working on the opposite direction: instead of improving the current infrastructure, it is creating a parallel one to handle qudits.

> The ExecutionManager works on Qudits (QuditInfo). Would it be possible to update the CircuitSimulator to operate on QuditInfo instances? @boschmitt what are your thoughts here? should we keep these things separate?

It seems to me that we don't even need to make `CircuitSimulator` operate on `QuditInfo`, we do need to give it the ability to know about qudits, but must of the methods in this new `PhotonicCircuitSimulator` work on qudit indices, which are exactly the same thing as qubit indices. 

The point made about the gates being different is correct, but arguably the simulator should not provide gate methods, e.g. `ccx(..)` or `beam_spliter(...)`, instead it should have a generic method that takes a matrix and the indices for either qubits or qudits.

Measurements are a bit more annoying but we could make so that `CircuitSimulator` layer returns a `int8_t` (or `uint8_t`), and then we properly cast it to `bool` when dealing with qubits, e.g., `cudaq::mz` will have a overload for `cudaq::qubit` that do the casting before returning to the user.

docs/sphinx/api/default_ops.rst
```diff
@@ -638,31 +638,67 @@ operations, each operating on 2 qubits.
 Photonic Operations on Qudits
 =============================
 
-These operations are valid only on the `photonics` target which does not support the quantum operations above.
+These operations are valid only on the `photonics-cpu` target which does not support
+the quantum operations above.
 
-:code:`plus`
+:code:`create`
```
khalatepradnya, line 644: Q: Is this a new gate? or an alias for 'plus'? Doesn't match the examples and API spec...

docs/sphinx/api/default_ops.rst
```diff
@@ -638,31 +638,67 @@ operations, each operating on 2 qubits.
...
 
 .. tab:: C++
 
     .. code-block:: cpp
 
         cudaq::qvector<3> q(1);
-        plus(q[0]);
+        create(q[0]);
+
+:code:`annihilate`
```
khalatepradnya, line 669: Ditto.

docs/sphinx/api/default_ops.rst
Omar-ORCA: This is a new gate replacing the `plus` that can be deprecated, the behaviour of the `plus` gate is not very intuitive (cycling through the number of photons).

Omar-ORCA:
Superseded by #2283

Issue #476, opened 2023-07-24
Rewrite tablegen rewrite patterns as C++ code


schweitzpgi:
As part of code cleanup, it was decided to just write the C++ classes instead of trying to make use of tablegen to generate rewrite patterns. Peephole.td and Canonical.td are sets of rewrite patterns that should be converted from tablegen to C++ code for this cleanup.

PR #2245, opened 2024-10-03
Allow controlled Z gates for profile-specific QIR

bmhowe23:
The `anyon-pgate-set-mapping` pass (used by the `berkeley-25q` machine from the Anyon target) uses the `z(1)` gate as the primary two-qubit gate. No other QIR-based hardware backend was using this gate as the primary two-qubit gate, so it was the first one to encounter this gap in the QIR implementation. This PR fixes the shortcoming and adds a test to prove that it works.

### DIFF
--- a/include/cudaq/Optimizer/CodeGen/Peephole.td
+++ b/include/cudaq/Optimizer/CodeGen/Peephole.td
@@ -33,6 +33,23 @@ def XCtrlOneTargetToCNot : Pat<
     (LLVM_CallOp $callee, $args, $_, $_), (CreateCallCnot $args),
     [(InvokeOnXWithOneControl $callee, $args)]>;
 
+def InvokeOnZWithOneControl : Constraint<CPred<
+    "$0 && callToInvokeWithZCtrlOneTarget($0.getValue(), $1)">>;
+
+def CreateCallCZ : NativeCodeCall<
+    "[&]() -> std::size_t {"
+    "  $_builder.create<mlir::LLVM::CallOp>($_loc,"
+    "    mlir::TypeRange{}, cudaq::opt::QIRCZ, $0.drop_front(2));"
+    "  return 0; }()">;
+
+// %1 = address_of @__quantum__qis__z__ctl
+// %2 = call @invokewithControlBits %1, %ctrl, %targ
+// ─────────────────────────────────────────────────
+// %2 = call __quantum__qis__cz %ctrl, %targ
+def ZCtrlOneTargetToCZ : Pat<
+    (LLVM_CallOp $callee, $args, $_, $_), (CreateCallCZ $args),
+    [(InvokeOnZWithOneControl $callee, $args)]>;
+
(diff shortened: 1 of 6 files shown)


include/cudaq/Optimizer/CodeGen/Peephole.td
```diff
@@ -33,6 +33,23 @@ def XCtrlOneTargetToCNot : Pat<
     (LLVM_CallOp $callee, $args, $_, $_), (CreateCallCnot $args),
     [(InvokeOnXWithOneControl $callee, $args)]>;
 
+def InvokeOnZWithOneControl : Constraint<CPred<
```
schweitzpgi, line 36: Please write this as a C++ class inheriting from a rewriter pattern. MLIR's support for autogenerating these has been kneecapped, and we need to rewrite all of these peepholes.

khalatepradnya (commented):
These changes look good to me. 👍🏽Thanks, Ben!

(I see Eric's comment on the peephole implementation so will defer final approval.)

include/cudaq/Optimizer/CodeGen/Peephole.td
bmhowe23: > Please write this as a C++ class inheriting from a rewriter pattern. MLIR's support for autogenerating these has been kneecapped, and we need to rewrite all of these peepholes.
> 
> #476

Updated in 324588fa6.

lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
```diff
@@ -434,6 +434,34 @@ struct CallAlloc : public OpRewritePattern<LLVM::CallOp> {
...
+    if (args.size() == 4 && call.getCallee() &&
+        call.getCallee()->equals(cudaq::opt::NVQIRInvokeWithControlBits)) {
+      if (auto addrOf = dyn_cast_or_null<mlir::LLVM::AddressOfOp>(
+              args[1].getDefiningOp())) {
+        if (addrOf.getGlobalName().startswith(
+                std::string(cudaq::opt::QIRQISPrefix) + "z__ctl")) {
+          rewriter.replaceOpWithNewOp<LLVM::CallOp>(
+              call, TypeRange{}, cudaq::opt::QIRCZ, args.drop_front(2));
+          if (addrOf.use_empty())
+            addrOf.erase();
```
schweitzpgi, line 456: It's better not to do this since we're actually iterating through the IR at this point. Erasing ops can confuse the iterator data structures and cause intermittent crashes.

schweitzpgi (commented):
LGTM. Thanks, Ben.

lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
bmhowe23: Thanks, Eric. Updated in 37fd88c35.

khalatepradnya (approved):
Thanks, Ben!

PR #2247, opened 2024-10-03
Clarifying the document text

sacpis:
The updated text makes it clear about default value of CPU memory.
https://nvidia.slack.com/archives/C03KV6A81HT/p1727967897285079

image


schweitzpgi (commented):
LGTM

PR #2256, opened 2024-10-06
Create cudaq python wheel

sacpis:
Checking installed CUDA libraries to detect CUDA version and generating wheel file.


cudaq-1.0.0/LICENSE
```diff
@@ -0,0 +1,121 @@
+NVIDIA SOFTWARE LICENSE
```
bettinaheim, line 1: Any changes to the software license would be a separate PR, and we'll want to keep it with the Apache 2.0 license we already have checked in. Was this added to this PR accidentally?

cudaq-1.0.0/cudaq.egg-info/requires.txt
```diff
@@ -0,0 +1 @@
+cudaq-cu11==1.0.0
```
bettinaheim, line 1: Why is this hardcoded to be cu11? 
A lot of things that are checked in here would/should be autogenerated by the wheel build. How are you building the cudaq source distribution (what you call the cudaq-1.0.0)?

bettinaheim:
Superseded by https://github.com/NVIDIA/cuda-quantum/pull/2356

PR #2266, opened 2024-10-10
Fix argument passing issue in C++ bridge.

schweitzpgi:
In C++, pointers to arrays are automatically compatible with pointers to the scalar element type. Teach the bridge this C++ rule. Add a new regression test.


amccaskey (approved):
+1

PR #2268, opened 2024-10-11
[core] Fix various bugs when processing the host-side functions and the

schweitzpgi:
native calling conventions.

As support for more types is added, it is required that the host side code have the same calling conventions as the native clang++ compiler. Otherwise, programs will simply fail to execute correctly.

This is part 1. The launch kernel execution path is broken, and those bugs will be addressed in the next PR.

### DIFF
--- a/lib/Optimizer/Builder/Factory.cpp
+++ b/lib/Optimizer/Builder/Factory.cpp
@@ -18,6 +18,9 @@ using namespace mlir;
 
 namespace cudaq::opt {
 
+// The common small struct limit for architectures cudaq is supporting.
+static constexpr unsigned CommonSmallStructSize = 128;
+
 bool factory::isX86_64(ModuleOp module) {
   std::string triple;
   if (auto ta = module->getAttr(targetTripleAttrName))
@@ -417,7 +420,7 @@ static bool shouldExpand(SmallVectorImpl<Type> &packedTys,
     } else {
       return false;
     }
-    if (bits > 128)
+    if (bits > CommonSmallStructSize)
       return false;
   }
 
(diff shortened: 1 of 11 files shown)


lib/Optimizer/Builder/Factory.cpp
```diff
@@ -403,33 +376,49 @@ static Type convertToHostSideType(Type ty) {
...
         set1.push_back(ty);
-      else
+      } else {
+        bits = std::max(bits, 64u) + addBits;
         set2.push_back(ty);
+      }
     } else {
       return false;
     }
+    if (bits > 128)
```
bmhowe23, line 420: I see ~5 references to `128` in this change set; they seem to suggest that structs smaller than 128 bits are considered "small", at least on x86-64. Is that right? Is it the same on arm64, too? Would it make sense to make that a `#define` or `constexpr` in a shared header file?

test/Quake-QIR/return_values.qke
```diff
@@ -6,20 +6,22 @@
...
+// QIR codegen requires the target triple.
 module attributes{ quake.mangled_name_map = {
   __nvqpp__mlirgen__test_0 = "test_0",
   __nvqpp__mlirgen__test_1 = "test_1",
   __nvqpp__mlirgen__test_2 = "test_2",
   __nvqpp__mlirgen__test_3 = "test_3",
   __nvqpp__mlirgen__test_4 = "test_4",
-  __nvqpp__mlirgen__test_5 = "test_5" }} {
+  __nvqpp__mlirgen__test_5 = "test_5" },
+  llvm.triple = "x86_64-unknown-linux-gnu"} {
```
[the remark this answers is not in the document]
schweitzpgi: I'm adding aarch64 in a separate PR.

lib/Optimizer/Builder/Factory.cpp
schweitzpgi: Yes, it turns out aarch64 is also 128. (Actually, I think the ABI may allow bigger structs for x86_64, but GNU did it their own way.)

bmhowe23 (approved):
LGTM

PR #2269, opened 2024-10-11
Workflow changes to facilitate maintenance

bettinaheim:


.github/workflows/clean_up.yml
```diff
@@ -42,7 +42,7 @@ jobs:
               /repos/nvidia/cuda-quantum/git/refs/heads/$branch 
           done
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.REPO_BOT_ACCESS_TOKEN }}
```
bmhowe23, line 45: Is this new secret pointing back to the public repo? If so, what (if anything) would delete the bot/* branches in the private repo?

bettinaheim: this is a secret on the repository that the workflow runs on. I can make this more fork-friendly by falling back to the gh token if it is not defined.

bmhowe23 (approved):
LGTM

bmhowe23 (approved):
This morning's updates LGTM. I made one comment offline, but I am ok with it as is.

PR #2278, opened 2024-10-14
[core] Add support for aarch64 calling conventions.

schweitzpgi:
Fix calling conventions for aarch64 and add a regression test.

Requires #2268 to be merged.


bmhowe23 (approved):
LGTM

PR #2286, opened 2024-10-16
Allow to configure assets, packages and updates from CUDA-Q repo

bettinaheim:
The changes allow to define a GitHub repository where temporary assets should be created, and to define a package prefix for all packages produced by this repo. 
It also adds a workflow for forks to fetch daily updates from the CUDA-Q parent repository.


1tnguyen (approved):
LGTM 👍

.github/workflows/nvqc_regression_tests.yml
bmhowe23, line 1: It looks like this will have both repos submit jobs to the production server at the same time. That should be fine, but I just wanted to point it out in case there are any anomalies that follow.

.github/workflows/integration_tests.yml
bmhowe23, line 1: Will both repos try to deploy servers to NVQC function ID at the same time? If so, that might not go so well.

bettinaheim: The image name is different to avoid any potential issues.

bmhowe23: > The image name is different to avoid any potential issues.

I verified that in the test script, we are setting the `NVQC_FUNCTION_VERSION_ID` environment variable based on the deployment, so that will help distinguish them as well, so that's good. But additionally, we are resource limited, so simultaneous deployments may exhaust our quota, depending on what other deployments are active. I think it's fine for now, but we'll need to monitor this going forward.

PR #2287, opened 2024-10-16
[core] Refactor the lift array alloc pass into two proper phases.

schweitzpgi:
Start changing core code over to use SmallVector (and be more compatible with MLIR) instead of std::vector.  Refactoring lifting of array values.  Implement globalize array values. Update new pass and pass names across all the config and yaml files.
    
Requires merge of #2277

### DIFF
--- a/lib/Optimizer/Transforms/ConstPropComplex.cpp
+++ b/lib/Optimizer/Transforms/ConstPropComplex.cpp
@@ -179,15 +179,15 @@ public:
                     FloatTruncatePattern, ComplexRePattern, ComplexImPattern>(
         ctx);
 
-    LLVM_DEBUG(llvm::dbgs()
-               << "Before lifting constant array: " << func << '\n');
+    LLVM_DEBUG(llvm::dbgs() << "Before constant propagation of complex values: "
+                            << func << '\n');
 
     if (failed(applyPatternsAndFoldGreedily(func.getOperation(),
                                             std::move(patterns))))
       signalPassFailure();
 
-    LLVM_DEBUG(llvm::dbgs()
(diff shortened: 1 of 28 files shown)


lib/Optimizer/Transforms/ConstPropComplex.cpp
```diff
@@ -172,29 +172,22 @@ class ConstPropComplexPass
...
-    }
+    auto func = getOperation();
+    DominanceInfo domInfo(func);
+    RewritePatternSet patterns(ctx);
+    patterns.insert<ComplexCreatePattern, FloatCastPattern, FloatExtendPattern,
+                    FloatTruncatePattern, ComplexRePattern, ComplexImPattern>(
+        ctx);
+
+    LLVM_DEBUG(llvm::dbgs()
+               << "Before lifting constant array: " << func << '\n');
```
annagrin, line 183: Looks like a copy-and-paste error from my side... should be "Before propagating complex constants" or similar. And below.

targettests/Remote-Sim/qvector_init_from_vector.cpp
```diff
@@ -78,155 +76,177 @@ void printCounts(cudaq::sample_result& result) {
...
-    }
+  {
+    auto counts = cudaq::sample(test_large_double_constant_array);
+    std::cout << "Part 1\n";
+    printCounts(counts);
+  }
 
-// CHECK: 0000000000000000000
-// CHECK: 1000000000000000000
+  // CHECK-LABEL: Part 1
```
annagrin, line 85: those should be already fixed on the main

annagrin (approved):
LGTM!

PR #2290, opened 2024-10-17
ensure quantum_lib is built after nvq++ components are built

amccaskey:
A bad rebase removed the add_dependency line for the quantum_lib test library. Add that back here.


bmhowe23 (approved):
Thanks!

schweitzpgi (approved):

Issue #2271, opened 2024-10-14
Mid-circuit measurements are not handled properly when using `__qpu__` attributes on C++ functions or lambdas


bmhowe23:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Adaptive quantum kernels written as C++ functions or lambdas with `__qpu__` attributes do not work correctly because internal logic in the CUDA-Q runtime cannot lookup the name of C++ functions or lambdas at runtime.

### Steps to reproduce the bug

The following example (based off of [targettests/execution/qir_simple_cond-1.cpp](https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/execution/qir_simple_cond-1.cpp) from the existing CUDA-Q tree, but slightly modified to use functions instead of class operators) reproduces the problem.
```cpp
#include <cudaq.h>
#include <iostream>

struct kernel {
  void operator()() __qpu__ {
    cudaq::qubit q0;
    cudaq::qubit q1;
    h(q0);
    auto q0result = mz(q0);
    if (q0result)
      x(q1);
    auto q1result = mz(q1); // Every q1 measurement will be the same as q0
  }
};

int main() {

  int nShots = 100;
  // Sample
  auto counts = cudaq::sample(/*shots=*/nShots, kernel{});
  counts.dump();
  // Assert that all shots contained "00" or "11", exclusively
  if (counts.count("00") + counts.count("11") != nShots) {
    std::cout << "counts00 (" << counts.count("00") << ") + counts11 ("
              << counts.count("11") << ") != nShots (" << nShots << ")\n";
    return 1;
  }
  std::cout << "SUCCESS\n";
  return 0;
}
```
Save the above file to `test-if.cpp`; then compile and run with `nvq++`:
```bash
$ nvq++ --enable-mlir test-if.cpp
$ ./a.out
...
counts00 (0) + counts11 (0) != nShots (100)
```
[...]

PR #2291, opened 2024-10-17
State synthesis for quantum devices

annagrin:
#### SimulationState:
- Add `hasData` API that returns true iff the vector data exists on the state or can be computed.
- Add `getKernelInfo` API that returns optional kernel name and a list of arguments for the kernel that generated the state.

#### QuantumState:
- Add QuantumState class that inherits from simulation state and keeps kernel information only.

#### ArgumentConversion:
- Substitute state pointer argument with `quake.get_state "callee.num_qubits_N" "callee.init_N"` instruction.
- Add `callee.num_qubits_N` and `callee.init_N` functions that are created from the callee code and compute allocation size and initialize the allocation, respectively.
- Add a list of converters for the callees in `ArgumentConverter`.
- Generate substitutions for modified callee functions, recursively.

#### Passes:
- Add a post-synthesis `ReplaceStateWithKernel` pass that
   -  replaces `quake.get_num_qubits` instructions by a call to `callee.num_qubits_N()`
   -  replaces `quake.init_state` instructions by a call to `callee.init_N()`

#### Synthesis:
- Collect all substitutions for the call tree at the synthesis time for quantum devices.

#### Tests:
- Add fake quantum state and a test for quantum state argument conversion to `test_argument_conversion`
- Add quake tests for `ReplaceStateWithKernel` pass
- Add quake tests for `ArgumentSynthesis` pass with state pointer substitutions
- Add `targettests/execution/qvector_init_from_state.cpp` test that runs for various quantum backends emulation

#### Notes
Currently state pointer synthesis is only supported for kernels implemented as `operator()` inside a struct.

#### TODO (in subsequent PRs)
- c++ tests for quantum backend integration
- Use `cudaq::qkernel` to support c-like kernels in state pointer synthesis
- Tests for synthesis of vectors of pauli words (needs rework of pauli word support)
- python tests for quantum platforms (needs support for new synthesis in python)

<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2354</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2644</strike>


lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,146 @@
...
+#define DEBUG_TYPE "state-initialization"
+
+using namespace mlir;
+
+namespace {
+
+static bool isCall(Operation *callOp, std::vector<const char *> &&names) {
+  if (callOp) {
+    if (auto createStateCall = dyn_cast<func::CallOp>(callOp)) {
+      if (auto calleeAttr = createStateCall.getCalleeAttr()) {
```
1tnguyen, line 36: nit: I think we can rename the variables to make it more generic.

```suggestion
static bool isCall(Operation *op, std::vector<const char *> &&names) {
  if (op) {
    if (auto callOp= dyn_cast<func::CallOp>(op)) {
      if (auto calleeAttr = callOp.getCalleeAttr()) {
```

runtime/common/CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ set_source_files_properties(
     JIT.cpp
     Logger.cpp
     RuntimeMLIR.cpp
-  PROPERTIES COMPILE_FLAGS -fno-rtti
+#  PROPERTIES COMPILE_FLAGS -fno-rtti
```
1tnguyen, line 105: If the compiler is happy with this, I guess we can just delete this line (and the comment above) :)

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
...
+#endif
+#if defined(CUDAQ_QUANTUM_DEVICE)
+  // Store kernel name and arguments for quantum states.
+  if (!cudaq::get_quake_by_name(cudaq::getKernelName(kernel), false).empty()) {
+    return state(new QuantumState(std::forward<QuantumKernel>(kernel),
+                                  std::forward<Args>(args)...));
+  } else {
+    throw std::runtime_error(
+        "cudaq::state* argument synthesis is not supported for quantum hardware"
+        "for c-like functions, use class kernels instead");
```
1tnguyen, line 131: Related to https://github.com/NVIDIA/cuda-quantum/issues/2271

1tnguyen (approved):
LGTM 👍

runtime/common/CMakeLists.txt
annagrin: thanks, will do!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

runtime/common/CMakeLists.txt
annagrin: done.

include/cudaq/Optimizer/Builder/Intrinsics.h
```diff
@@ -55,6 +55,10 @@ static constexpr const char createCudaqStateFromDataFP32[] =
 // Delete a state created by the runtime functions above.
 static constexpr const char deleteCudaqState[] = "__nvqpp_cudaq_state_delete";
 
+// Get state of a kernel (placeholder function, calls are always replaced in
+// opts)
+static constexpr const char getCudaqState[] = "__nvqpp_cudaq_state_get";
```
schweitzpgi, line 60: Do you want to stick with a dummy function or would you rather introduce a NOP "conversion" Op to the IR for this?

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -779,6 +779,44 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
   }];
 }
 
+def StateInitialization : Pass<"state-initialization", "mlir::func::FuncOp"> {
```
schweitzpgi, line 782: I think we should tweak this name a bit to be more descriptive. Something more like "replace state with kernel" perhaps?

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -828,6 +866,17 @@ def StatePreparation : Pass<"state-prep", "mlir::ModuleOp"> {
   ];
 }
 
+def StateValidation : Pass<"state-validation", "mlir::ModuleOp"> {
+  let summary =
+    "Make sure MLIR is valid after synthesis for quantum devices";
+  let description = [{
+    Argument synthesis should replace all `quake.init` from state instructions
```
schweitzpgi, line 873:
```suggestion
    Argument synthesis should replace all `quake.state_init` from state instructions
```
?

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -828,6 +866,17 @@ def StatePreparation : Pass<"state-prep", "mlir::ModuleOp"> {
...
+    and calls to state-related runtime functions.
+    Make sure none of them left, and remove definitions for state-related
```
schweitzpgi, line 875: If the definitions are marked `private` there is nothing special to do. `private` artifacts that are unreferenced will be cleaned up automatically.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -828,6 +866,17 @@ def StatePreparation : Pass<"state-prep", "mlir::ModuleOp"> {
   ];
 }
 
+def StateValidation : Pass<"state-validation", "mlir::ModuleOp"> {
```
schweitzpgi, line 869: Not sure about this name either. It seems to suggest that there is some sort of comparison going on, possibly to validate (verify) that an algorithm is correct.

schweitzpgi (commented):
Still reviewing...

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+/// ```
+///  %0 = cc.string_literal "callee.modified_0" : !cc.ptr<!cc.array<i8 x 27>>
+///  %1 = cc.cast %0 : (!cc.ptr<!cc.array<i8 x 27>>) -> !cc.ptr<i8>
+///  %2 = call @__nvqpp_cudaq_state_get(%1) : (!cc.ptr<i8>) -> !cc.ptr<!cc.state>
+///  %3 = call @__nvqpp_cudaq_state_numberOfQubits(%2) : (!cc.ptr<!cc.state>) -> i64
+///  %4 = quake.alloca !quake.veq<?>[%3 : i64]
+///  %5 = quake.init_state %4, %2 : (!quake.veq<?>, !cc.ptr<!cc.state>) -> !quake.veq<?>
+/// ───────────────────────────────────────────
+/// ...
+///  %5 = call @callee.modified_0() : () -> !quake.veq<?>
```
schweitzpgi, line 65: This violates the semantics of Quake: all quantum memory allocations happen at the top-level and cannot be returned from other kernels.

schweitzpgi (commented):
Still reviewing...

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+/// ```
+// clang-format on
+class StateInitPattern : public OpRewritePattern<quake::InitializeStateOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(quake::InitializeStateOp initState,
+                                PatternRewriter &rewriter) const override {
+    auto loc = initState.getLoc();
+    auto allocaOp = initState.getOperand(0).getDefiningOp();
```
schweitzpgi, line 75:
```suggestion
    auto *allocaOp = initState.getOperand(0).getDefiningOp();
```

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto stateOp = initState.getOperand(1);
+
+    if (isa<cudaq::cc::StateType>(stateOp.getType())) {
+      auto getStateOp = stateOp.getDefiningOp();
+      auto numOfQubits = allocaOp->getOperand(0).getDefiningOp();
```
schweitzpgi, line 80:
```suggestion
      auto *numOfQubits = allocaOp->getOperand(0).getDefiningOp();
```

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(quake::InitializeStateOp initState,
+                                PatternRewriter &rewriter) const override {
+    auto loc = initState.getLoc();
+    auto allocaOp = initState.getOperand(0).getDefiningOp();
+    auto stateOp = initState.getOperand(1);
+
+    if (isa<cudaq::cc::StateType>(stateOp.getType())) {
+      auto getStateOp = stateOp.getDefiningOp();
```
schweitzpgi, line 79:
```suggestion
      auto *getStateOp = stateOp.getDefiningOp();
```

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+      auto numOfQubits = allocaOp->getOperand(0).getDefiningOp();
+
+      if (isGetStateCall(getStateOp)) {
+        auto calleeNameOp = getStateOp->getOperand(0);
+        if (auto cast =
+                dyn_cast<cudaq::cc::CastOp>(calleeNameOp.getDefiningOp())) {
```
schweitzpgi, line 85:
```suggestion
               calleeNameOp.getDefiningOp<cudaq::cc::CastOp>()) {
```

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+          calleeNameOp = cast.getOperand();
+
+          if (auto literal = dyn_cast<cudaq::cc::CreateStringLiteralOp>(
+                  calleeNameOp.getDefiningOp())) {
```
schweitzpgi, line 89:
```suggestion
          if (auto literal = 
                  calleeNameOp.getDefiningOp<cudaq::cc::CreateStringLiteralOp>()) {
```

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+            auto calleeName = literal.getStringLiteral();
+
+            Value result =
+                rewriter
+                    .create<func::CallOp>(loc, initState.getType(), calleeName,
+                                          mlir::ValueRange{})
+                    .getResult(0);
+            rewriter.replaceAllUsesWith(initState, result);
```
schweitzpgi, line 97: Why not use `replaceOpWithNewOp` for this?

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+            initState.erase();
+            allocaOp->dropAllUses();
+            rewriter.eraseOp(allocaOp);
```
schweitzpgi, line 100: None of these erasures are checking that there aren't other uses, etc. That should be fixed.

Note: this is where it would be much better in my opinion to come up with as small a set of primitive Ops as possible to define instead of using calls. An Op without effects that has no use will just automatically be erased as well as the graph of whatever it uses recursively. No need to manually check constraints or try to subvert them.

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+            cast->dropAllUses();
+            rewriter.eraseOp(cast);
+            literal->dropAllUses();
+            rewriter.eraseOp(literal);
+            return success();
+          }
+        }
+      }
+    }
+    return failure();
```
schweitzpgi, line 116: return initState.emitOpError("oh, no! this is bad!");

lib/Optimizer/Transforms/StateValidation.cpp
```diff
@@ -0,0 +1,127 @@
...
+    auto stateOp = initState.getOperand(1);
+    if (isa<cudaq::cc::StateType>(stateOp.getType()))
+      initState.emitError("Synthesis did not remove `quake.init_state <veq> "
+                          "<state>` instruction");
+
+    return failure();
+  }
+};
+
+class StateValidationPass
```
schweitzpgi, line 88: I'm skeptical this pass is really needed.

The other pass looks like it can be modified to fail if it cannot replace all instances of `quake.init_state`. And any sort of residual local definitions will already be cleaned up by MLIR passes like SymbolDCE, so we don't need to implement that.

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: I am fine either way. An OP gives an advantage of not having to add the declaration to the MLIR for this function and having to remove it in StateValidation pass (but there are more to remove so StateValidation is still needed).

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: Sounds good, will do

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: Some later ops are failing because those definitions are still in the MLIR... Will double check if they are private.

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: It is only temporary and does not seem to break anything until the Inlining fixes the semantics... But returning allocations (instead of passing them as a parameter) is much easier to implement. Let me know if it is a showstopper, I can revisit passing allocations as a parameter.

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: it crashes otherwise. The crash does not happen if I postpone erasing...

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: will do, thanks!

annagrin: Btw not everything is getting erased - function calls, alloca ops stay, so I have to remove them manually.

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: Wouldn't this fail the pass? I just want it to continue in the case of `faliure()`

lib/Optimizer/Transforms/StateValidation.cpp
annagrin: I don't see local definitions cleaned up, some later conversion fail because they are still there... When is SymbolDCE is invoked?

lib/Optimizer/Transforms/StateInitialization.cpp
schweitzpgi: Well, it's a bug even if it "works" for now, so I think we ought to deal with it now.

I don't see why returning is any easier. Updating the "init kernel" from something like
```
func.func @init_kernel() {
   %0 = quake.alloca !quake.veq<4>
   ...
}
```
to a "modified init kernel" of
```
func.func @init_kernel(%0 : !quake.veq<4>) {
   ...
}
```
is easy enough.

Also, if we pass in the allocated qubits, we don't have to erase them at the top-level or rely on inlining to get things normalized, etc.

lib/Optimizer/Transforms/StateInitialization.cpp
schweitzpgi: I suspect the crash might be because ops are being erased directly in the IR instead of letting the rewriter deal with them. Walking the IR and erasing Ops out of the linked lists they are in will crash unfortunately.

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: Renamed to `RemoveRuntimeFunctions`, let me know if you have a better naming idea

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: Actually, I was thinking about adding ops for a few runtime functions related to state, so it would be easier to remove those ops in optimizations and only add the runtime functions if needed. I would like to do that in a separate PR, for all the state-related runtime functions:

```cpp
// The internal data of the cudaq::state object must be `2**n` in length. This
// function returns the value `n`.
static constexpr const char getNumQubitsFromCudaqState[] =
    "__nvqpp_cudaq_state_numberOfQubits";

// Create a new state from data.
static constexpr const char createCudaqStateFromDataFP64[] =
    "__nvqpp_cudaq_state_createFromData_fp64";
static constexpr const char createCudaqStateFromDataFP32[] =
    "__nvqpp_cudaq_state_createFromData_fp32";

// Get state of a kernel (placeholder function, calls are always replaced in
// opts)
static constexpr const char getCudaqState[] = "__nvqpp_cudaq_state_get";
```

For the first 3 functions, we would still have the runtime intrinsics defined but calls to them added only if the corresponding op is not removed by opts. The last function would not be needed as the corresponding op is always removed by the opts. WDYT?

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: done, thanks!

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: Added SymbolDCE opt after synthesis and cleanup opts, it worked, thanks! Removed the StateValidation pass entirely.

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+namespace {
+
+static bool isCall(Operation *op, std::vector<const char *> &&names) {
+  if (op) {
+    if (auto callOp = dyn_cast<func::CallOp>(op)) {
+      if (auto calleeAttr = callOp.getCalleeAttr()) {
+        auto funcName = calleeAttr.getValue().str();
+        if (std::find(names.begin(), names.end(), funcName) != names.end())
+          return true;
```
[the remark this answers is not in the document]
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
```diff
@@ -0,0 +1,141 @@
...
+static bool isGetStateCall(Operation *op) {
+  return isCall(op, {cudaq::getCudaqState});
+}
+
+static bool isNumberOfQubitsCall(Operation *op) {
+  return isCall(op, {cudaq::getNumQubitsFromCudaqState});
+}
+
+// clang-format off
+/// Replace `quake.init_state` by a call to a (modified) kernel that produced the state.
```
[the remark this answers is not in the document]
annagrin: yes, thanks! fixed

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: Apologies, I updated my answer above - in this case something went wrong when I was replacing an init_state op with a call, I thought it was due to a different number of results, but it works on the second attempt with some other modifications in the code, so I presume those other changes helped. Updated.

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: Updated

lib/Optimizer/Transforms/StateValidation.cpp
annagrin: Removed that pass

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: Added checks and errors. Planning to send a follow-up PR to create a set of ops for state-related operations.

include/cudaq/Optimizer/Builder/Intrinsics.h
annagrin: Replaced `"__nvqpp_cudaq_state_get"` intrinsic by `cc.get_state` op, this simplified the code quite a bit. I will create a follow-up PR for the rest of the state-related op additions, as it is not related to the current change.

lib/Optimizer/Transforms/StateInitialization.cpp
annagrin: updated to 2 calls (one for getting the number of qubits, the other for initializing the allocation passed as a parameter.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1467,19 +1467,29 @@ def QuakeOp_GetStateOp : QuakeOp<"get_state", [Pure] > {
...
     ```mlir
-      %0 = quake.get_state "callee" : !cc.ptr<!cc.state>
+      %0 = quake.get_state "num_qubits" "init" : !cc.ptr<!cc.state>
     ```
   }];
 
-  let arguments = (ins StrAttr:$calleeName);
+  let arguments = (ins
+    StrAttr:$numQubitsFuncName,
+    StrAttr:$initFuncName
```
schweitzpgi, line 1488:
```suggestion
    FlatSymbolRefAttr:$numQubitsFuncName,
    FlatSymbolRefAttr:$initFuncName
```
If these are truly artifacts that shall be present in the IR, let's make them Symbol attrs.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1467,19 +1467,29 @@ def QuakeOp_GetStateOp : QuakeOp<"get_state", [Pure] > {
...
+  );
   let results = (outs cc_PointerType:$result);
```
schweitzpgi, line 1490: We probably want to force this to be a ptr<state>, no?

include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -58,6 +58,13 @@ std::unique_ptr<mlir::Pass>
 createArgumentSynthesisPass(mlir::ArrayRef<mlir::StringRef> funcNames,
                             mlir::ArrayRef<mlir::StringRef> substitutions);
 
+/// Helper function to build an argument synthesis pass. The names of the
+/// functions and the substitutions text can be built as an unzipped pair of
+/// lists.
+std::unique_ptr<mlir::Pass>
+createArgumentSynthesisPass(const std::vector<std::string> &funcNames,
+                            const std::vector<std::string> &substitutions);
```
schweitzpgi, line 66: Why do we need this? `ArrayRef` should subsume rigid `std::vector` here. I don't think we need this overload?

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
annagrin: If the signatures of those functions are changed by the synthesis, would this instruction get updated with symbols with new signatures?

include/cudaq/Optimizer/Transforms/Passes.h
annagrin: `std::vector<string>` does not get auto-converted to `ArrayRef<StringRef>`...  I can try `ArrayRef<std::string>` instead

runtime/common/ArgumentConversion.h
```diff
@@ -47,13 +48,30 @@ class ArgumentConverter {
...
+    auto converter = ArgumentConverter(calleeName, sourceModule);
+    converter.gen(args);
+    calleeConverters.push_back(converter);
+  }
+
+  std::vector<ArgumentConverter> &getCalleeConverters() {
+    return calleeConverters;
+  }
+
+  std::pair<std::vector<std::string>, std::vector<std::string>>
```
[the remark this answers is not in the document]
annagrin: I think i tried and something went wrong, will try again

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
annagrin: Will do, thanks

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
schweitzpgi: I don't believe that'll be an issue. That is, you can update the signature independently of the symbol.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -446,18 +447,18 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
...
+        // the kernels and their arguments that produced the state.
+        opt::ArgumentConverter argCon(kernelName, moduleOp);
         argCon.gen(rawArgs);
-        std::string kernName = cudaq::runtime::cudaqGenPrefixName + kernelName;
-        mlir::SmallVector<mlir::StringRef> kernels = {kernName};
-        std::string substBuff;
-        llvm::raw_string_ostream ss(substBuff);
-        ss << argCon.getSubstitutionModule();
-        mlir::SmallVector<mlir::StringRef> substs = {substBuff};
+        auto [kernels, substs] = argCon.collectAllSubstitutions();
```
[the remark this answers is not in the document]
schweitzpgi: Let me try to be more clear.

The LLVM way to do this sort of thing is to build the string and/or vector data on the stack in the calling function. This will be reclaimed once the task we're performing completes.

For the called functions (the call tree below the "top" function where the data was built), we use the LLVM ADTs which are efficient lightweight wrappers around that data. This eliminates the use of heap memory, accidentally calling copy-constructors and destructors, etc. At the "leaf layer" where we create, say, MLIR IR, any data that must be interned is captured into the MLIRContext.

Why is it done this way? Because both LLVM and MLIR use a model of invariant pointers to refer to distinct artifacts in the overall representation. If we have an `Op*`, then that pointer value will never change. The `Op` may be logically replaced, by updating all references in the IR, but the `Op` itself will never be moved/copied someplace else in memory. This applies to Type values, string data, etc. Once it's in the IR, think of it as semi-permanent or congealed into a memory location, if you will. Because of all that, using STL containers and `std::string` is likely because the code isn't really structured in the spirit of the underlying LLVM model.

Of course this model imposes a bit of structure on how we write the code and implement things. It limits the types of refactoring we can do. But that's ok, because we can do it the LLVM Way, be efficient, and avoid bugs.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1467,19 +1467,29 @@ def QuakeOp_GetStateOp : QuakeOp<"get_state", [Pure] > {
...
+      - "init" for initializing the state the same way as the original kernel
+        passed to `cudaq::get_state`) as ASCIIZ string literal
+
+    And returns the quantum state of the original kernel passed to
+    `cudaq::get_state`. The operation is replaced by calls to the kernels with
+    the provided names in `ReplaceStateByKernel` pass.
 
     ```mlir
-      %0 = quake.get_state "callee" : !cc.ptr<!cc.state>
+      %0 = quake.get_state "num_qubits" "init" : !cc.ptr<!cc.state>
```
schweitzpgi, line 1482: I think these two parameters need to be an IntegerAttr and a Symbol. The syntax ought to look like
```mlir
  quake.get_state @init_artifact, <n> (attributes { ... })?
```
We don't even need the type, since I don't think it can be anything other that `!cc.ptr<!state>`, right?

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1467,19 +1467,29 @@ def QuakeOp_GetStateOp : QuakeOp<"get_state", [Pure] > {
...
     This operation is created by argument synthesis of state pointer arguments
-    for quantum devices. It takes a kernel name as ASCIIZ string literal value
-    and returns the kernel's quantum state. The operation is replaced by a call
-    to the kernel with the provided name in ReplaceStateByKernel pass.
+    for quantum devices.
+
+    It takes two kernel names as ASCIIZ string literals:
+      - "num_qubits" for determining the size of the allocation to initialize
+      - "init" for initializing the state the same way as the original kernel
+        passed to `cudaq::get_state`) as ASCIIZ string literal
```
schweitzpgi, line 1475: Isn't `init` the name of a kernel? i.e., it needs to be a legit symbol and not just an arbitrary string ("See Spot. See Spot run.").

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1467,19 +1467,29 @@ def QuakeOp_GetStateOp : QuakeOp<"get_state", [Pure] > {
...
 
-  let arguments = (ins StrAttr:$calleeName);
+  let arguments = (ins
+    StrAttr:$numQubitsFuncName,
+    StrAttr:$initFuncName
+  );
   let results = (outs cc_PointerType:$result);
   let assemblyFormat = [{
-     $calleeName `:` qualified(type(results)) attr-dict
+     $numQubitsFuncName $initFuncName `:` qualified(type(results)) attr-dict
```
schweitzpgi, line 1492:
```suggestion
     $initFuncName `,` $numQubitsFuncName attr-dict
```
If we go with what I suggested. We'll want to provide some builders to hardcode the return type as well.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -815,9 +815,47 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
...
+    }
+    ```
+  }];
+}
+
+def ReplaceStateWithKernel : Pass<"replace-state-with-kernel", "mlir::func::FuncOp"> {
+  let summary =
+    "Replace `quake.init_state` instructions with call to the kernel generating the state";
+  let description = [{
+    Argument synthesis for state pointers for quantum devices substitutes state
```
schweitzpgi, line 831:
```suggestion
    Argument synthesis for state pointers for quantum devices will substitute
```

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -815,9 +815,47 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
...
+    argument by a new state created from `__nvqpp_cudaq_state_get` intrinsic, which
```
schweitzpgi, line 832:
```suggestion
    a new state created from the `__nvqpp_cudaq_state_get` intrinsic for the state argument.
```

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -815,9 +815,47 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
...
+    in turn accepts the name for the synthesized kernel that generated the state.
```
schweitzpgi, line 833:
```suggestion
    The `__nvqpp_cudaq_state_get` intrinsic accepts the symbol for the synthesized kernel that generated the state.
```

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -815,9 +815,47 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
...
+
+    This optimization completes the replacement of `quake.init_state` instruction by:
+
+    - Replace `quake.init_state` by a call that `get_state` call refers to.
```
schweitzpgi, line 837:
```suggestion
    - Replacing `quake.init_state` by a call to a kernel to construct the state. e.g., The `cudaq::get_state` call refers to the result of a specific quantum kernel being invoked with a set of parameters.
```

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -815,9 +815,47 @@ def DeleteStates : Pass<"delete-states", "mlir::ModuleOp"> {
...
+    - Remove all unneeded instructions.
```
schweitzpgi, line 838:
```suggestion
    - Remove any unneeded Ops.
```
I think we only really need to remove `Op`s that have side-effects. Any that are pure and unused will just go away with the next DCE that runs.

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
```diff
@@ -163,3 +163,12 @@ cudaq::opt::createArgumentSynthesisPass(ArrayRef<StringRef> funcNames,
   return std::make_unique<ArgumentSynthesisPass>(
       ArgumentSynthesisOptions{pairs});
 }
+
+std::unique_ptr<mlir::Pass> cudaq::opt::createArgumentSynthesisPass(
+    const std::vector<std::string> &funcNames,
```
schweitzpgi, line 168: As discussed, we should see if we can't get rid of this overload and use the original function with the LLVM ADT signature.

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
     return state(new RemoteSimulationState(std::forward<QuantumKernel>(kernel),
                                            std::forward<Args>(args)...));
   }
+#endif
+#if defined(CUDAQ_QUANTUM_DEVICE)
+  // Store kernel name and arguments for quantum states.
+  if (!cudaq::get_quake_by_name(cudaq::getKernelName(kernel), false).empty()) {
+    return state(new QuantumState(std::forward<QuantumKernel>(kernel),
+                                  std::forward<Args>(args)...));
+  } else {
```
schweitzpgi, line 128: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

runtime/cudaq/qis/quantum_state.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  for (std::size_t counter = 0; auto &ptr : args)
+    deleters[counter++](ptr);
+
+  args.clear();
+  deleters.clear();
+}
+
+std::size_t QuantumState::getNumQubits() const {
+  throw std::runtime_error(
```
schweitzpgi, line 29: Seems like all of these stubs are masking coding bugs until runtime and then just cause things to crash.

Can we just make `QuantumState` a pure interface and all of these methods pure virtual? That would force any derived classes to think it through and provide a proper implementation. An alternative would be to use the CRTP and the instantiation would require that the implementation class define all of the methods correctly.

runtime/test/FakeSimulationState.h
```diff
@@ -21,7 +21,7 @@ class FakeSimulationState : public cudaq::SimulationState {
   virtual std::unique_ptr<SimulationState>
   createFromSizeAndPtr(std::size_t size, void *data,
                        std::size_t dataType) override {
-    std::runtime_error("Not implemented");
+    throw std::runtime_error("Not implemented");
     return std::make_unique<FakeSimulationState>(size, data);
```
schweitzpgi, line 25: The return is never reached, right?

runtime/test/test_argument_conversion.cpp
```diff
@@ -80,14 +100,12 @@ void doTest(mlir::MLIRContext *ctx, std::vector<std::string> &typeNames,
...
-
   // Create the argument conversions
   ab.gen_drop_front(args, startingArgIdx);
-
-  // Dump the conversions
-  llvm::outs() << "========================================\n"
-                  "Substitution module:\n"
-               << ab.getSubstitutionModule() << '\n';
+  // Dump the modified source module
+  llvm::outs() << "Source module (after):\n" << *mod << '\n';
```
schweitzpgi, line 106: Is this dumping the same thing? It looks like it was dumping the substitution module but now is dumping the module receiving the substitutions?

schweitzpgi: Since this is testing _argument conversions_, we must care about building the substitutions properly. The code into which they are applied is only interesting at the point of _argument synthesis_, no?

runtime/common/ArgumentConversion.cpp
```diff
@@ -99,12 +99,245 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+  initFunc.setType(funcTy);
+  initFunc.setPrivate();
+
+  OpBuilder newBuilder(ctx);
+
+  auto *entryBlock = &initFunc.getRegion().front();
+  newBuilder.setInsertionPointToStart(entryBlock);
+  auto intType = newBuilder.getI64Type();
+  Value zero = newBuilder.create<arith::ConstantIntOp>(loc, 0, intType);
+  Value one = newBuilder.create<arith::ConstantIntOp>(loc, 1, intType);
```
schweitzpgi, line 145:
```suggestion
  Value zero = newBuilder.create<arith::ConstantIntOp>(loc, 0, 64);
  Value one = newBuilder.create<arith::ConstantIntOp>(loc, 1, 64);
```
nit: I've been hardcoding the size as an integer everywhere.

runtime/common/ArgumentConversion.cpp
```diff
@@ -99,12 +99,245 @@ static Value genConstant(OpBuilder &, cudaq::cc::StructType, void *,
...
+/// !!quake.veq<?> {
+///   %1 = quake.extract_ref %arg0[1] : (!quake.veq<2>) -> !quake.ref
+///   quake.x %1 : (f64, !quake.ref) -> ()
+///   return %arg0: !quake.veq<?>
+/// }
+static void createInitFunc(OpBuilder &builder, ModuleOp sourceMod,
+                           func::FuncOp calleeFunc,
+                           std::string &initKernelName) {
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToEnd(sourceMod.getBody());
```
schweitzpgi, line 124: Are we creating _argument substitution_ information in the substitution module here?

It should not be assumed that the source module can be modified in place or that it is going to be the same module as the final result, after argument synthesis.

So, we have 3 modules:
   1. original source (assume it is const)
   2. a substitution module full of substitution patterns
   3. a result module (to be generated by a different pass)

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
...
+    throw std::runtime_error(
+        "cudaq::state* argument synthesis is not supported for quantum hardware"
+        "for c-like functions, use class kernels instead");
+  }
 #endif
```
schweitzpgi, line 133:
```suggestion
#else
```
and wrap the return in the else conditional block...

runtime/test/FakeQuantumState.h
```diff
@@ -0,0 +1,112 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "cudaq/qis/state.h"
+#include <cassert>
+#include <memory>
+
```
schweitzpgi, line 12:
```suggestion
```
nit

runtime/test/test_argument_conversion.cpp
```diff
@@ -384,6 +402,197 @@ void test_state(mlir::MLIRContext *ctx) {
   // clang-format on
 }
 
+void test_quantum_state(mlir::MLIRContext *ctx) {
+  {
+    auto kernel = "init";
+    auto kernelCode =
+        ""
```
schweitzpgi, line 409:
```suggestion
```

runtime/test/test_argument_conversion.cpp
```diff
@@ -384,6 +402,197 @@ void test_state(mlir::MLIRContext *ctx) {
...
+// CHECK:         init.num_qubits_0
+// CHECK-LABEL:   cc.arg_subst[0] {
+// CHECK:           %[[VAL_0:.*]] = arith.constant 2 : i64
+// CHECK:         }
+  // clang-format on
+
+  {
+    auto kernel = "init";
+    auto kernelCode =
+        ""
```
schweitzpgi, line 485:
```suggestion
```

runtime/common/ArgumentConversion.cpp
```diff
@@ -146,9 +379,136 @@ static Value genConstant(OpBuilder &builder, const cudaq::state *v,
...
+
+    // Create a substitution for the state pointer.
+    auto statePtrTy =
+        cudaq::cc::PointerType::get(cudaq::cc::StateType::get(ctx));
+    return builder.create<quake::GetStateOp>(
+        loc, statePtrTy, builder.getStringAttr(numQubitsKernelName),
+        builder.getStringAttr(initKernelName));
+  }
+
+  TODO("cudaq::state* argument synthesis for quantum hardware for c functions");
```
schweitzpgi, line 511: Why doesn't this use `cudaq::qkernel`? That's the vehicle for capturing quantum kernels in general-purpose C++ code contexts. Since `cudaq::get_state()` may appear anywhere, I don't understand how we know what kernel is being captured.

There is the additional issue of captured arguments as well...

schweitzpgi:
I'm going to have to experiment with this on my machine. I'm not sure I get what's going on... For example,

```c++
int main() {
  // This is not quantum code and will never be seen by our compiler.
  ...
  KernelA a;  // our compiler sees KernelA's call operator.
  a(arg0, arg1, ... argN); // but not this call
  ...
  auto state1 = cudaq::get_state(a);   // is this ok per the spec?
  ...
  // f is a plain old function marked with __qpu__, so compiler sees it
  ...
  auto state2 = cudaq::get_state(f, frg0, frg1, ... frgN);  // I think this must be supported
  ....
  KernelB b;  // our compiler sees KernelB's call operator
  b(brg0, brg1, ..., state1, state2); // but not this call
  ...
}
```

So our implementation must capture the context to fully recreate the kernel launches of `a` and `f` so they can be replayed by the kernel launch of `b`. But we're not using `qkernel` which is the piece that captures `a` and `f` in `main`. I'm not sure how/if the state capture of `a` would work since the arguments aren't present. (Hopefully, that's not a legit case.)

If we focus on `f` here, and we were to capture the kernel in a `qkernel` then we can also capture the args in the `get_state` calls argument list.

Now "argument synthesis" is sort of a misnomer at this point. `KernelB` takes 2 `cudaq::state*` arguments, but there is nothing to synthesize (explicitly in terms of data constants) and substitute into the body here. In fact, the `state*` arguments are dead. The `quake.init_state` Ops that use them in the original code can be rewritten into device-only calls to `KernelA'`::operator() and `f'` which are clones of `KernelA::operator()` and `f` with extra argument(s) stuck on which are the veq (or ref) quake.alloca's from `KernelB` being passed in and which replace the corresponding `quake.alloca` operations in the called (now) device-only kernel (prime variants).

So unless I am missing something, it seems like this can be implemented without argument synthesis and as a standalone compiler pass, no? (Argument conversion would need to know that the `cudaq::state*` are just getting dropped on the floor, I guess.)

test/Quake/arg_subst-5.txt
```diff
@@ -0,0 +1,11 @@
+// ========================================================================== //
```
schweitzpgi, line 1: Should we name these `.txt` files with an extension like `.subst` to help identify them?

test/Quake/arg_subst_func.qke
```diff
@@ -146,3 +146,33 @@ func.func @testy4(%arg0: !cc.stdvec<!cc.struct<{i32, f64, i8, i16}>>) {
...
 // CHECK:         }
+
+func.func @testy5(%arg0: !cc.ptr<!cc.state>) {
+  %0 = quake.get_number_of_qubits %arg0 : (!cc.ptr<!cc.state>) -> i64
+  %1 = quake.alloca !quake.veq<?>[%0 : i64]
+  %5 = quake.init_state %1, %arg0 : (!quake.veq<?>, !cc.ptr<!cc.state>) -> !quake.veq<?>
+  return
+}
+
+func.func @init(%arg0: i64, %arg1: !quake.veq<?>) -> !quake.veq<?> {
```
schweitzpgi, line 157: For these `init` kernels, the qubit vector is passed as references, which means the qubits passed in will be changed by the function in-place (or remain unchanged if unreferenced, I suppose). I don't think returning that same set of qubit references is needed.

test/Quake/arg_subst_func.qke
```diff
@@ -146,3 +146,33 @@ func.func @testy4(%arg0: !cc.stdvec<!cc.struct<{i32, f64, i8, i16}>>) {
...
+  return %arg1 : !quake.veq<?>
+}
+
+func.func @num_qubits(%arg0: i64) -> i64 {
```
schweitzpgi, line 161: When we get to the phase of building the substitutions, we should know the number of qubits as a constant, right?

runtime/test/FakeSimulationState.h
```diff
@@ -21,7 +21,7 @@ class FakeSimulationState : public cudaq::SimulationState {
   virtual std::unique_ptr<SimulationState>
```
schweitzpgi, line 21: This header file and `FakeQuantumState.h` are really just scaffolding for the test and aren't included anywhere else. We could get rid of these .h files and move the code into the file `test_argument_conversion.cpp`.

schweitzpgi: In fact, most of the code for these two classes is the same or very similar and could be shared or templated.

runtime/test/FakeSimulationState.h
annagrin: fixed, thanks!

runtime/test/FakeSimulationState.h
annagrin: done, thanks!

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1436,7 +1436,7 @@ def QuakeOp_DeleteStateOp : QuakeOp<"delete_state", [] > {
     ```
   }];
 
-  let arguments = (ins cc_PointerType:$state);
+  let arguments = (ins AnyStatePointerType:$state);
```
schweitzpgi, line 1439: I know we've had this discussion before somewhere, but this notion of "delete state" still just makes no sense to me at all. We are necessarily talking about kernel code here (this is quake dialect and it can't run anywhere else).

Since it is kernel code, we also have a hard restriction that we don't have a heap. No heap means, we don't have (normal) objects, since objects are memory-bound data structures by definition. No objects at all in the kernel makes the description of this operation in conflict. If we don't have any objects, what object does this think it is deleting?

Operationally speaking, a state vector (regardless of how it is packaged) shouldn't be cloned at the kernel level anyway.

I think this had something to do with some oddball issues with "remote simulators" IIRC. We need to have a meeting with @1tnguyen (I think) to discuss this, because I think this is semantically broken and contrary to the rest of the Quake IR.

runtime/cudaq/qis/quantum_state.h
```diff
@@ -0,0 +1,153 @@
...
+#include "common/SimulationState.h"
+#include "cudaq.h"
+#include "cudaq/utils/cudaq_utils.h"
+
+namespace cudaq {
+/// Implementation of `SimulationState` for quantum device backends.
+// The state is represented by a quantum kernel.
+// Quantum state contains all the information we need to replicate a
+// call to kernel that created the state.
+class QuantumState : public cudaq::SimulationState {
```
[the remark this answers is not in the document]
annagrin: moved to platform and renamed to qpu_state.h and qpu_state.cpp, renamed the class to QPUState

runtime/common/ArgumentConversion.cpp
annagrin: Changed to adding new functions in the parent substitution module (this works after making the synthesis transitive in https://github.com/NVIDIA/cuda-quantum/pull/2644

annagrin:
> ```c++
> int main() {
>   // This is not quantum code and will never be seen by our compiler.
>   ...
>   KernelA a;  // our compiler sees KernelA's call operator.
>   a(arg0, arg1, ... argN); // but not this call

This will execute kernelA with a set of qubits and deallocate them, we never see the result.

>   ...
>   auto state1 = cudaq::get_state(a);   // is this ok per the spec?

No, we cannot call `cudaq::get_state(a)` without arguments it accepts.

>   ...
>   // f is a plain old function marked with __qpu__, so compiler sees it
>   ...
>   auto state2 = cudaq::get_state(f, frg0, frg1, ... frgN);  // I think this must be supported

Yes, agreed, and I plan to support it in one of the next PRs
>   ....
>   KernelB b;  // our compiler sees KernelB's call operator
>   b(brg0, brg1, ..., state1, state2); // but not this call

This one is interesting... I definitely missed that. Currently it crashes, I will try to add an error instead and think if we can support this case in a later PR.

I think this should fold into the discussion of what it means to execute a kernel without `sample` or `run` on a quantum device.
>   ...
> }
> ```

schweitzpgi:
> > ```c++
> > int main() {
> >   // This is not quantum code and will never be seen by our compiler.
> >   ...
> >   KernelA a;  // our compiler sees KernelA's call operator.
> >   a(arg0, arg1, ... argN); // but not this call
> > ```
> 
> On a simulator, this will execute kernelA with a set of qubits and deallocate them, we never see the result. On a quantum device we don't support it currently:
> 

I don't think "never see the result" is a strong/valid argument. It is imperative that one assumes the executing a quantum kernel *does* produce a result. Per the CUDA-Q spec, there is no ambiguity. One can run kernels and kernels can return results. From the target models there are different mechanisms for results to be returned and on hardware devices these results have been by side-channels.

> ```
>  terminate called after throwing an instance of 'std::runtime_error'
>   what():  Remote rest execution can only be performed via cudaq::sample(), cudaq::observe(), or cudaq::draw().
> ```
> 
> > ...
> > auto state1 = cudaq::get_state(a);   // is this ok per the spec?
> 
> No, we cannot call `cudaq::get_state(a)` without arguments `a.operator()` accepts.
> 
> > ...
> > // f is a plain old function marked with **qpu**, so compiler sees it
> > ...
> > auto state2 = cudaq::get_state(f, frg0, frg1, ... frgN);  // I think this must be supported
> 
> Yes, agreed, and I plan to support it in one of the next PRs
> 
> > ....
> > KernelB b;  // our compiler sees KernelB's call operator
> > b(brg0, brg1, ..., state1, state2); // but not this call
> 
> This one runs on a simulator but we never see the results... On quantum devices, it currently throws an error so it seems we did not intend to support it:
> 

I find the argument that something shouldn't be supported by limitations of our early prototypes to not be terribly convincing. Let's take these points to the power that be for further clarification though.

> ```
>  terminate called after throwing an instance of 'std::runtime_error'
>   what():  Remote rest execution can only be performed via cudaq::sample(), cudaq::observe(), or cudaq::draw().
> ```
> 
> I think this should fold into a broader discussion of what it means to execute a kernel without `sample` or `run` on a quantum device and how to return values in that case (I suspect it is possible after `cudaq::run` is implemented?)
> 
> > ...
> > }
> 
> > Now "argument synthesis" is sort of a misnomer at this point. KernelB takes 2 cudaq::state* arguments, but there is nothing to >synthesize (explicitly in terms of data constants) and substitute into the body here. In fact, the state* arguments are dead. The >quake.init_state Ops that use them in the original code can be rewritten into device-only calls to KernelA'::operator() and f' >which are clones of KernelA::operator() and f with extra argument(s) stuck on which are the veq (or ref) quake.alloca's from >KernelB being passed in and which replace the corresponding quake.alloca operations in the called (now) device-only kernel
> > (prime variants).
> 
> `b(brg0, brg1, ..., state1, state2)` is a direct function call that is currently not supported for quantum devices, but it is supported for simulators and remote simulators. I've tried to create a `sample with 1 shot` context for it [here](https://github.com/NVIDIA/cuda-quantum/blob/a79d9abb3206ddb1e4e693569abdc25ab9a10aaf/runtime/common/BaseRemoteRESTQPU.h#L633) and I could successfully run synthesis for it and execute it on a quantum device (with --emulate flag) so it seems I have this case covered if needed.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -1633,30 +1633,43 @@ def quake_GetNumberOfQubitsOp : QuakeOp<"get_number_of_qubits", [Pure] > {
...
-  let arguments = (ins StrAttr:$calleeName);
-  let results = (outs cc_PointerType:$result);
+  let arguments = (ins
+    FlatSymbolRefAttr:$numQubitsFunc,
+    FlatSymbolRefAttr:$initFunc
+  );
+  let results = (outs PointerOf<[cc_StateType]>:$result);
   let assemblyFormat = [{
-     $calleeName `:` qualified(type(results)) attr-dict
+     $numQubitsFunc $initFunc `:` qualified(type(results)) attr-dict
```
schweitzpgi, line 1672:
```suggestion
     $numQubitsFunc `,` $initFunc `:` qualified(type(results)) attr-dict
```
nit: would be nice to use a comma here for consistency.

lib/Optimizer/Transforms/ReplaceStateWithKernel.cpp
```diff
@@ -0,0 +1,129 @@
...
+// clang-format on
+class ReplaceGetNumQubitsPattern
+    : public OpRewritePattern<quake::GetNumberOfQubitsOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(quake::GetNumberOfQubitsOp numQubits,
+                                PatternRewriter &rewriter) const override {
+
+    auto stateOp = numQubits.getOperand();
```
schweitzpgi, line 50:
```suggestion
    auto stateVal = numQubits.getState();
```
Operands are always `mlir::Value` and not `mlir::Op`.

lib/Optimizer/Transforms/ReplaceStateWithKernel.cpp
```diff
@@ -0,0 +1,129 @@
...
+    auto materializeState = stateOp.getDefiningOp<quake::MaterializeStateOp>();
+    if (!materializeState)
+      return numQubits->emitError(
+          "ReplaceStateWithKernel: failed to replace `quake.get_num_qubits`");
```
schweitzpgi, line 54: I wouldn't make this an absolute error. Just returning failure() here will keep the compiler from crashing and allow for future implementations to introduce different tactics cleanly.

lib/Optimizer/Transforms/ReplaceStateWithKernel.cpp
```diff
@@ -0,0 +1,129 @@
...
+    auto allocaOp = initState.getOperand(0);
+    auto stateOp = initState.getOperand(1);
+
+    if (auto ptrTy = dyn_cast<cudaq::cc::PointerType>(stateOp.getType())) {
+      if (isa<cudaq::cc::StateType>(ptrTy.getElementType())) {
+        auto materializeState =
+            stateOp.getDefiningOp<quake::MaterializeStateOp>();
+        if (!materializeState)
+          return initState->emitError(
+              "ReplaceStateWithKernel: failed to replace `quake.init_state`");
```
schweitzpgi, line 91: Again, `return failure();` ought to be enough. The codegen isn't going to generate code ultimately if nothing cleans this up.

That said, adding a `LLVM_DEBUG(llvm::dbgs() << "Whoops!\n");` line would help with debugging the pass.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -543,18 +543,37 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
...
-  llvm::raw_string_ostream ss(substBuff);
-  ss << argCon.getSubstitutionModule();
-  SmallVector<StringRef> substs = {substBuff};
+
+  // Store kernel and substitution strings on the stack.
+  // We pass string references to the `createArgumentSynthesisPass`.
+  mlir::SmallVector<std::string> kernels;
+  mlir::SmallVector<std::string> substs;
+  for (auto &kInfo : argCon.getKernelSubstitutions()) {
+    {
```
schweitzpgi, line 554:
```suggestion
    
```

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -543,18 +543,37 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
...
+      std::string kernName =
+          cudaq::runtime::cudaqGenPrefixName + kInfo.getKernelName().str();
+      kernels.emplace_back(kernName);
+    }
+    {
+      std::string substBuff;
+      llvm::raw_string_ostream ss(substBuff);
+      ss << kInfo.getSubstitutionModule();
+      substs.emplace_back(substBuff);
+    }
```
schweitzpgi, line 564:
```suggestion
  
```
The extra braces aren't needed and are distracting. clang is smart enough to be able to determine object lifetimes and optimize dtor calls, etc. I think we can just let the compiler do what it does best here.

runtime/common/ArgumentConversion.h
```diff
@@ -37,23 +71,39 @@ class ArgumentConverter {
...
+    return kernelSubstitutions.emplace_back(kernelName, substModule);
+  }
+
+  /// Memory to store new kernel names generated during argument conversion.
+  /// Use list here to keep references to those elements valid.
+  std::list<std::string> nameRegistry;
+
+  /// Memory to store new kernel info generated during argument conversion.
+  /// Use list here to keep elements sorted in order of creation.
+  std::list<KernelSubstitutionInfo> kernelSubstitutions;
```
schweitzpgi, line 100:
```suggestion
  SmallVector<KernelSubstitutionInfo> kernelSubstitutions;
```

runtime/common/ArgumentConversion.h
```diff
@@ -37,23 +71,39 @@ class ArgumentConverter {
...
-  /// Some substitutions may generate global constant information. Use this
-  /// interface to access both the substitutions and any global constants
-  /// created.
-  mlir::ModuleOp getSubstitutionModule() { return substModule; }
+  bool isRegisteredKernel(const std::string &kernelName) {
+    return std::find(nameRegistry.begin(), nameRegistry.end(), kernelName) !=
+           nameRegistry.end();
+  }
+
+  std::string &registerKernel(const std::string &kernelName) {
```
schweitzpgi, line 84:
```suggestion
  std::string &registerKernel(StringRef kernelName) {
```

runtime/common/BaseRestRemoteClient.h
```diff
@@ -185,15 +185,36 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
-          ss << argCon.getSubstitutionModule();
-          mlir::SmallVector<mlir::StringRef> substs = {substBuff};
-          pm.addPass(opt::createArgumentSynthesisPass(kernels, substs));
+
+          // Store kernel and substitution strings on the stack.
+          // We pass string references to the `createArgumentSynthesisPass`.
+          mlir::SmallVector<std::string> kernels;
+          mlir::SmallVector<std::string> substs;
+          for (auto &kInfo : argCon.getKernelSubstitutions()) {
+            {
```
schweitzpgi, line 194: nit: braces

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,27 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
     return state(new RemoteSimulationState(std::forward<QuantumKernel>(kernel),
                                            std::forward<Args>(args)...));
   }
+#else
+#if defined(CUDAQ_QUANTUM_DEVICE) && !defined(CUDAQ_LIBRARY_MODE)
```
schweitzpgi, line 123:
```suggestion
#elif defined(CUDAQ_QUANTUM_DEVICE) && !defined(CUDAQ_LIBRARY_MODE)
```

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,27 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
...
+  // Store kernel name and arguments for quantum states.
+  if (!cudaq::get_quake_by_name(cudaq::getKernelName(kernel), false).empty())
+    return state(new QPUState(std::forward<QuantumKernel>(kernel),
+                              std::forward<Args>(args)...));
+  throw std::runtime_error(
+      "cudaq::state* argument synthesis is not supported for quantum hardware"
+      " for c-like functions, use class kernels instead");
+#else
+#if defined(CUDAQ_QUANTUM_DEVICE)
```
schweitzpgi, line 132:
```suggestion
#elif defined(CUDAQ_QUANTUM_DEVICE)
```

runtime/cudaq/algorithms/get_state.h
```diff
@@ -118,6 +119,27 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
...
+  // Kernel builder is MLIR-based kernel.
+  if constexpr (has_name<QuantumKernel>::value)
+    return state(new QPUState(std::forward<QuantumKernel>(kernel),
+                              std::forward<Args>(args)...));
+
+  throw std::runtime_error(
+      "cudaq::state* argument synthesis is not supported for quantum hardware"
+      " for c-like functions in library mode");
+#endif
+#endif
```
schweitzpgi, line 142:
```suggestion
```

runtime/cudaq/cudaq.cpp
```diff
@@ -19,6 +19,7 @@
 #include "distributed/mpi_plugin.h"
 #include <dlfcn.h>
 #include <filesystem>
+#include <list>
```
schweitzpgi, line 22: Since there are no other changes, I assume this can be removed.

schweitzpgi (commented):
Some nits. I'm still looking at some of the changes here...

runtime/common/CMakeLists.txt
```diff
@@ -93,13 +93,11 @@ add_library(cudaq-mlir-runtime
 )
 set_property(GLOBAL APPEND PROPERTY CUDAQ_RUNTIME_LIBS cudaq-mlir-runtime)
 
-# Note: JIT.cpp contains throw statements. Should RTTI be enabled?
 set_source_files_properties(
     Environment.cpp
     JIT.cpp
     Logger.cpp
     RuntimeMLIR.cpp
-  PROPERTIES COMPILE_FLAGS -fno-rtti
```
schweitzpgi, line 102: @amccaskey Do you recall why this had RTTI disabled?

runtime/cudaq/platform/default/rest/helpers/anyon/anyon.yml
```diff
@@ -13,6 +13,8 @@ config:
   platform-qpu: remote_rest
   # Tell NVQ++ to generate glue code to set the target backend name
   gen-target-backend: true
+  # Add preprocessor defines to compilation
+  preprocessor-defines: ["-D CUDAQ_QUANTUM_DEVICE"]
```
schweitzpgi, line 17: Can't this preprocessor macro be added by the `nvq++` driver?

targettests/Remote-Sim/qvector_init_from_state_lazy.cpp
```diff
@@ -16,54 +16,51 @@
 
 #include <cudaq.h>
 #include <iostream>
+#include <string>
+#include <vector>
```
schweitzpgi, line 20: Curious but what does `lazy` in the filename mean?

targettests/execution/qvector_init_from_state.cpp
```diff
@@ -0,0 +1,137 @@
...
+  std::sort(values.begin(), values.end());
+  for (auto &&bits : values) {
+    std::cout << bits << std::endl;
+  }
+}
+
+int main() {
+  std::vector<cudaq::complex> vec{M_SQRT1_2, M_SQRT1_2, 0., 0., 0., 0., 0., 0.};
+  std::vector<cudaq::complex> vec1{0., 0., 0.,        0.,
+                                   0., 0., M_SQRT1_2, M_SQRT1_2};
```
schweitzpgi, line 58: do we need a clang-format here?

runtime/cudaq/platform/qpu_state.h
```diff
@@ -0,0 +1,153 @@
...
+
+#include "common/SimulationState.h"
+#include "cudaq.h"
+#include "cudaq/utils/cudaq_utils.h"
+
+namespace cudaq {
+/// Implementation of `SimulationState` for quantum device backends.
+// The state is represented by a quantum kernel.
+// Quantum state contains all the information we need to replicate a
+// call to kernel that created the state.
```
schweitzpgi, line 19: Is all of this intended to be the doxygen comment?

runtime/common/CMakeLists.txt
amccaskey: i believe this was to get the build working against LLVM/MLIR which by default turns off rtti in favor of its hand-rolled version

runtime/common/ArgumentConversion.h
```diff
@@ -13,21 +13,55 @@
 #include "cudaq/qis/state.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/Types.h"
+#include <list>
```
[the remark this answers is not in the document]
annagrin: replaced by `SmallVector<StringAttr>`, thanks!

runtime/common/ArgumentConversion.h
```diff
@@ -37,23 +71,39 @@ class ArgumentConverter {
...
+  /// Get the kernel info that were collected by `gen()`.
+  std::list<KernelSubstitutionInfo> &getKernelSubstitutions() {
+    return kernelSubstitutions;
   }
 
-  /// Some substitutions may generate global constant information. Use this
-  /// interface to access both the substitutions and any global constants
-  /// created.
-  mlir::ModuleOp getSubstitutionModule() { return substModule; }
+  bool isRegisteredKernel(const std::string &kernelName) {
```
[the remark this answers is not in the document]
annagrin: done!

runtime/common/ArgumentConversion.h
annagrin: done!

runtime/common/ArgumentConversion.h
```diff
@@ -37,23 +71,39 @@ class ArgumentConverter {
...
 
 private:
+  KernelSubstitutionInfo &addKernelInfo(mlir::StringRef kernelName,
+                                        mlir::ModuleOp substModule) {
+    return kernelSubstitutions.emplace_back(kernelName, substModule);
+  }
+
+  /// Memory to store new kernel names generated during argument conversion.
+  /// Use list here to keep references to those elements valid.
+  std::list<std::string> nameRegistry;
```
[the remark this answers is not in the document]
annagrin: Replaced by `SmallVector<StringAttr>`, thanks!

runtime/common/ArgumentConversion.h
annagrin: Replaced by `SmallVector<KernelSubstitutionInfo*>` and added a destructor. This allow me to use a small vector and create new kernel infos in order of creation of the corresponding kernels, and avoid references going invalid when the vector is reallocated.

runtime/cudaq/algorithms/get_state.h
annagrin: done, thanks!

runtime/cudaq/algorithms/get_state.h
annagrin: done!

runtime/cudaq/algorithms/get_state.h
annagrin: done!

runtime/cudaq/cudaq.cpp
annagrin: done, thanks!

runtime/cudaq/platform/qpu_state.h
annagrin: fixed, thanks!

runtime/common/CMakeLists.txt
annagrin: @amccaskey since the build works with rtti enabled now, is this change ok, or are there some hidden problems that we might be concerned about?

runtime/test/test_argument_conversion.cpp
```diff
@@ -11,18 +11,151 @@
...
+  }
+
+  virtual void destroyState() override {}
+
+  virtual std::complex<double>
+  operator()(std::size_t tensorIdx,
+             const std::vector<std::size_t> &indices) override {
+    if (hasData()) {
+      assert(tensorIdx == 0);
+      assert(indices.size() == 1);
```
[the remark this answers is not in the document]
annagrin: I would like the test to crash if some unexpected functions are called. I'll add a throw instead.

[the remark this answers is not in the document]
annagrin: done!

runtime/common/CMakeLists.txt
annagrin: all tests pass so I think we are good!

runtime/cudaq/platform/default/rest/helpers/anyon/anyon.yml
schweitzpgi: I'm not a fan of copying the "same" configuration data/code all over the sources. It's hard to maintain and easily broken.

If you want to open an issue, that's ok.

schweitzpgi (approved):

schweitzpgi:
Wahoo!

runtime/cudaq/platform/default/rest/helpers/anyon/anyon.yml
annagrin: Done: https://github.com/NVIDIA/cuda-quantum/issues/2747

Issue #2299, opened 2024-10-18
New launch kernel is not invoked for python


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Python code using states as arguments fails to synthesize on quantum targets and remote sim due to the new launch kernel not being invoked (old synthesis message is displayed)

### Code 

```python
# RUN: PYTHONPATH=../../.. python3 %s --target  --remote-mqpu
# RUN: PYTHONPATH=../../.. python3 %s --target  --quantinuum --emulate

import cudaq
import numpy as np
def test_init_from_data_state():
    c = np.array([1. / np.sqrt(2.) + 0j, 1. / np.sqrt(2.), 0., 0.],
                 dtype=np.complex128)
    state = cudaq.State.from_data(c)
    @cudaq.kernel
    def kernel(vec: cudaq.State):
        q = cudaq.qvector(vec)
    counts = cudaq.sample(kernel, state)
    assert '10' in counts
    assert '00' in counts
test_init_from_data_state()

def test_init_from_state():
    @cudaq.kernel
    def init(n: int):
        q = cudaq.qvector(n)
        ry(np.pi/2.0, q[0])
    @cudaq.kernel
    def kernel(vec: cudaq.State):
        q = cudaq.qvector(vec)
    state = cudaq.get_state(init, 2)
    counts = cudaq.sample(kernel, state)
    assert '10' in counts
    assert '00' in counts
test_init_from_state()
```

### Output

```
root@f4bab51c0fb2:/workspaces/cuda-quantum# python3 python/tests/mlir/target/tmp.py --target quantinuum
error: 'func.func' op synthesis: unsupported argument type for remote devices and simulators: state*
error: 'func.func' op We cannot synthesize argument(s) of this type.
RuntimeError: Could not successfully apply quake-synth.
```

### Expected behavior

[...]

PR #2300, opened 2024-10-18
Fix failing NVQC integration tests

annagrin:
### Description
Disable `building_kernels.py` test for NVQC integration tests (it requires state pointer argument synthesis to work, which requires a fix for https://github.com/NVIDIA/cuda-quantum/issues/2299)


bmhowe23 (approved):
Thanks, @annagrin. Could you please add a link to https://github.com/NVIDIA/cuda-quantum/issues/2299 in this file itself to explain that is the reason for this test being disabled? (We should probably do that for the others too, but those can probably wait.)

scripts/validate_container.sh
```diff
@@ -181,7 +181,7 @@ do
             # Skipped long-running tests (variational optimization loops) for the "remote-mqpu" target to keep CI runtime managable.
             # A simplified test for these use cases is included in the 'test/Remote-Sim/' test suite. 
             # Skipped tests that require passing kernel callables to entry-point kernels for the "remote-mqpu" target.
-            if [[ "$ex" == *"vqe_h2"* || "$ex" == *"qaoa_maxcut"* || "$ex" == *"gradients"* || "$ex" == *"grover"* || "$ex" == *"multi_controlled_operations"* || "$ex" == *"phase_estimation"* || "$ex" == *"trotter_kernel_mode"* || "$ex" == *"builder.cpp"* ]];
+            if [[ "$ex" == *"vqe_h2"* || "$ex" == *"qaoa_maxcut"* || "$ex" == *"gradients"* || "$ex" == *"grover"* || "$ex" == *"multi_controlled_operations"* || "$ex" == *"phase_estimation"* || "$ex" == *"building_kernels"* || "$ex" == *"trotter_kernel_mode"* || "$ex" == *"builder.cpp"* ]];
```
bmhowe23, line 184: Oops, I meant to put the comment here: Could you please add a link to https://github.com/NVIDIA/cuda-quantum/issues/2299 in this file itself to explain that is the reason for this test being disabled? (We should probably do that for the others too, but those can probably wait.)

bmhowe23: On second thought, I think these changes should be reverted because I think the required changes would be near here instead: https://github.com/NVIDIA/cuda-quantum/blob/94f6ee513565549bd6b8d5f909cc64a46130cab0/.github/workflows/integration_tests.yml#L670.

annagrin:
replaced by other PRs

PR #2301, opened 2024-10-19
Fixes publishing failure

sacpis:
Fixes the following publishing failure
https://github.com/NVIDIA/cuda-quantum/actions/runs/11412281048/job/31762944853#step:5:1293

### DIFF
--- a/.github/workflows/config/spelling_allowlist.txt
+++ b/.github/workflows/config/spelling_allowlist.txt
@@ -5,6 +5,7 @@ APIs
 AST
 Aer
 AlmaLinux
+Anyon
 Asynchronous
 BFGS
 Bloch
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -261,7 +261,7 @@ jobs:
         #if: github.event_name == 'schedule' || inputs.target == 'nightly' || inputs.target == 'anyon'
         if: inputs.target == 'anyon'
         run: |
-          curl -X POST -H "Content Type: application/json" -d '{ "username":"${{ secrets.BACKEND_LOGIN_USERNAME }}","password":"${{ secrets.BACKEND_LOGIN_PASSWORD }}" }' https://api.anyon.cloud/login > credentials.json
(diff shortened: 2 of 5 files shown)


docs/sphinx/applications/python/vqe_advanced.ipynb
```diff
@@ -341,7 +341,7 @@
     "\n",
     "cudaq.set_target(\"nvidia-mqpu\")\n",
     "\n",
-    "num_qpus = 4\n",
+    "num_qpus = 1\n",
```
bmhowe23, line 344: Perhaps this should be:
```suggestion
    "num_qpus = cudaq.get_target().num_qpus()\n",
```
I believe that would allow it to dynamically calculate the number of QPUs to be equal to the number of GPUs on the host system.

docs/sphinx/applications/python/vqe_advanced.ipynb
```diff
@@ -367,7 +367,7 @@
     "                                qubit_count,\n",
     "                                electron_count,\n",
     "                                xplus[i],\n",
-    "                                qpu_id=qpu_counter))\n",
+    "                                qpu_id=qpu_counter%num_qpus))\n",
```
bmhowe23, line 370: I believe this change (and the one on line 379) would make `qpu_counter` an unused variable, so I'm wondering if they should be reverted.

bmhowe23: After looking at this again, I realize my confusion. The use of `qpu_counter` is confusing. The code was originally assuming that num_qpus was always at least 2, and that is obviously a bad assumption. Your change fixes that, which is good.

However, lines 382-383 can potentially mislead someone into thinking that qpu_counter will always reset back to 0 every time it wraps around, but that isn't actually what would happen. For example, if num_qpus was 3, the line 383 would _not_ execute when i==0 but _would_ execute when i==1. This qpu_counter variable is a bit flawed in my opinion.

In order to prevent continued confusion, I would recommend either a) duplicating lines 382-383 right after line 371, or b) removing lines 382-383. It makes no sense to have the wrap-around check after one increment but not another increment.

If you choose a), then lines 370 and 379 should simply be `qpu_id=qpu_counter` (because qpu_counter would be guaranteed to be < num_qpus), and if you choose b), then lines 370 and 379 should be as you have them right now (`qpuid=qpu_counter%num_qpus`).

bmhowe23 (approved):
Thanks, Sachin!

PR #2304, opened 2024-10-21
Add tensor type to cudaq.

amccaskey:


schweitzpgi:
@anthony-santana @sacpis Is this the interface you were looking for? (makes Jedi hand motion.)

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+        return false;
+    return true;
+  }
+
+public:
+  /// @brief Constructor for xtensor
+  /// @param d Pointer to the tensor data
+  /// @param s Shape of the tensor
+  xtensor(const Scalar *d, const std::vector<std::size_t> &s)
+      : m_data(const_cast<Scalar *>(d)), m_shape(s) {}
```
bmhowe23, line 44: This takes ownership of the data, and if that is intentional, the docs should be updated to reflect that. (Otherwise, users may be tempted to pass `theirVec.data()` into the constructor and then be confused when a double free (or something worse) happens.

runtime/cudaq/utils/tensor.h
```diff
@@ -0,0 +1,146 @@
...
+
+/// @brief A tensor class implementing the PIMPL idiom.
+template <typename Scalar = std::complex<double>>
+class tensor {
+private:
+  std::shared_ptr<details::tensor_impl<Scalar>> pimpl;
+
+public:
+  /// @brief Type alias for the scalar type used in the tensor
+  using scalar_type = details::tensor_impl<Scalar>::scalar_type;
```
bmhowe23, line 24: I think this might be needed for C++17 (or, more specifically, if we include this file in an nvcc compilation.
```suggestion
  using scalar_type = typename details::tensor_impl<Scalar>::scalar_type;
```

anthony-santana (commented):
General note: Replace "cudaqx" namespaces and comments in docs with "cudaq"

runtime/cudaq/utils/extension_point.h
```diff
@@ -0,0 +1,168 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <unordered_map>
```
bmhowe23, line 13: I think nvcc needed this change:
```suggestion
#include <stdexcept>
#include <unordered_map>
```

runtime/cudaq/utils/extension_point.h
```diff
@@ -0,0 +1,168 @@
...
+  /// @brief Create an instance of a registered extension.
+  /// @param name The identifier of the registered extension.
+  /// @param args Constructor arguments for the extension.
+  /// @return A unique pointer to the created instance.
+  /// @throws std::runtime_error if the extension is not found.
+  static std::unique_ptr<T> get(const std::string &name, CtorArgs... args) {
+    auto &registry = get_registry();
+    auto iter = registry.find(name);
+    if (iter == registry.end())
+      throw std::runtime_error("Cannot find extension with name = " + name);
```
bmhowe23, line 103: I think nvcc needed this change. The bottom line is that it would be good for somebody to add a test with a .cu file to make sure all of this works with the nvcc compiler.
```suggestion
      throw std::runtime_error(
          std::string("Cannot find extension with name = ") + name);
```

python/cudaq/__init__.py
```diff
@@ -73,7 +74,8 @@
 Target = cudaq_runtime.Target
 State = cudaq_runtime.State
 pauli_word = cudaq_runtime.pauli_word
-Tensor = cudaq_runtime.Tensor
+# FIXME What to do with this Tensor, is it used publicly?
+StateTensor = cudaq_runtime.Tensor
```
schweitzpgi, line 78: @annagrin ?

python/cudaq/utils/Tensor.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Args:
+            data (numpy.ndarray): The data to initialize the tensor with.
+            dtype (numpy.dtype): The data type of the tensor elements.
+
+        Raises:
+            ValueError: If the dtype is unsupported.
+        """
+        self._dtype = dtype
+        if dtype == np.float64:
```
schweitzpgi, line 82: float32, int16 ?

python/extension/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ endif()
 include(HandleLLVMOptions)
 include(AddMLIRPython)
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT APPLE)
+  add_link_options(-Wl,--no-as-needed)
```
schweitzpgi, line 17: Why is this needed?

python/runtime/utils/py_tensor.cpp
```diff
@@ -0,0 +1,139 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include <limits>
+#include <pybind11/complex.h>
+#include <pybind11/numpy.h>
+#include <pybind11/operators.h>
+#include <pybind11/stl.h>
+
```
schweitzpgi, line 13: nit
```suggestion
```

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
+/****************************************************************-*- C++ -*-****
```
schweitzpgi, line 1:
```suggestion
/*******************************************************************************
```

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * Copyright (c) 2024 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "cudaq/utils/details/tensor_impl.h"
+#include "cudaq/utils/type_traits.h"
+
```
schweitzpgi, line 11:
```suggestion
```

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include <xtensor/xadapt.hpp>
+#include <xtensor/xio.hpp>
+
```
schweitzpgi, line 14:
```suggestion
```

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+      : m_data(const_cast<Scalar *>(d)), m_shape(s) {}
+
+  /// @brief Get the rank of the tensor
+  /// @return The rank (number of dimensions) of the tensor
+  std::size_t rank() const override { return m_shape.size(); }
+
+  /// @brief Get the total size of the tensor
+  /// @return The total number of elements in the tensor
+  std::size_t size() const override {
+    return std::accumulate(m_shape.begin(), m_shape.end(), 1,
```
schweitzpgi, line 53: What happens if `m_shape` is empty?

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+#include <xtensor/xadapt.hpp>
+#include <xtensor/xio.hpp>
+
+#include <fmt/ranges.h>
+
+namespace cudaq {
+
+/// @brief An implementation of tensor_impl using xtensor library
+template <typename Scalar>
```
schweitzpgi, line 20: Probably want to use concepts here to force `Scalar` to be a limited set of types.

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /// @brief Copy data into the tensor
+  /// @param d Pointer to the source data
+  /// @param shape The shape of the source data
+  void copy(const Scalar *d, const std::vector<std::size_t> &shape) override {
+    auto size = std::accumulate(shape.begin(), shape.end(), 1,
+                                std::multiplies<size_t>());
+    if (m_data)
+      std::free(m_data);
```
schweitzpgi, line 91:
```suggestion
      delete m_data;
```
This matches the next line, but as noted above the ctor is just assuming ownership of data that was allocated outside of this code.

runtime/cudaq/utils/details/tensor_impl.h
```diff
@@ -0,0 +1,134 @@
...
+ * Copyright (c) 2024 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#pragma once
+
+#include "cudaq/utils/extension_point.h"
+
```
schweitzpgi, line 11:
```suggestion
```

python/cudaq/__init__.py
annagrin: @1tnguyen do you think we can remove it from here?

annagrin: The type was exposed before so technically this change can have effect on the user code, for example:

the following user code
```python 
def foo(t: list[cudaq.Tensor]):
        return t
```
would need to be changed to
```python
def foo(t: list[cudaq.StateTensor]):
        return t
```

I would think that this is a rare (and may be non-existent) occurrence, but I am not really familiar with the user code and our process on changing language... 

@efratshabtai do you know if we can just rename an already existing type if there is low chance of it being already used?

amccaskey: This should not be exposed at the CUDA-Q kernel level. If you are talking about annotations for CUDA-Q kernels, the type to be used there should just be cudaq.State. Why would we allow any general Tensor as input to a kernel? And if the above are just regular Python functions, why annotate those?

runtime/cudaq/utils/details/impls/xtensor_impl.cpp
amccaskey: yea that should be updated to return 0 in that case

python/extension/CMakeLists.txt
amccaskey: This can be dropped, left over from debugging

runtime/cudaq/utils/type_traits.h
```diff
@@ -0,0 +1,119 @@
+/****************************************************************-*- C++ -*-****
```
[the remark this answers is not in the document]
amccaskey: My thought was that this would always be included from relative path `#include "cudaq/utils/type_traits.h"`, thus clearly differentiating from the standard header.

python/cudaq/utils/Tensor.py
amccaskey: definitely ok to add more supported types. Just make sure to add them in py_tensor.cpp and in the bottom of xtensor_impl.cpp

runtime/cudaq/utils/extension_point.h
```diff
@@ -33,47 +33,47 @@ namespace cudaq {
 ///    This class should declare pure virtual methods that extensions will
 ///    implement.
 ///
-/// @code
+/// <code>
 /// class MyExtensionPoint : public cudaq::extension_point<MyExtensionPoint> {
 /// public:
-///   virtual std::string parrotBack(const std::string &msg) const = 0;
+///   virtual std::string parrotBack(const std::string &) const = 0;
```
schweitzpgi, line 39: It looks like spell check is failing on this. Why? It's clearly in a code block?

python/tests/utils/test_tensor.py
```diff
@@ -0,0 +1,115 @@
...
+            t[i, j] = i * 3 + j
+
+    arr = np.array(t, copy=False)
+    assert arr.shape == tuple(shape)
+    assert np.array_equal(arr, np.array([[0, 1, 2], [3, 4, 5]]))
+
+
+def test_numpy_to_tensor():
+    arr = np.array([[0, 1, 2], [3, 4, 5]], dtype=np.complex128)
+    t = cudaq.Tensor(arr.shape, dtype=np.complex128)
```
annagrin, line 56: We should add tests for other dtypes

python/tests/utils/test_tensor.py
```diff
@@ -0,0 +1,115 @@
...
+import pytest
+import numpy as np
+import cudaq
+
+
+def tensor_class(request):
+    return request.param
+
+
+def test_tensor_creation():
```
annagrin, line 18: We should add tests for `take`, `borrow`, `copy`, `dump` and such

runtime/cudaq/utils/extension_point.h
annagrin: looks like pyspelling does not ignore the <code> blocks correctly, I tied to play with configuration to no avail, so I changed them to markdown quotes (```) instead

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -313,3 +317,4 @@ verifier
 vertices
 wavefunction
 workflow
+NumPy
```
schweitzpgi, line 320: Isn't the spelling checker supposed to accept anything in CamelCase as a legal spelling?

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -234,8 +235,10 @@ mutex
 namespace
 namespaces
 natively
+ndarray
```
schweitzpgi, line 238: Isn't this a python type name? If so, I thought we backquoted all of those?

.github/workflows/config/spelling_allowlist.txt
bmhowe23: > Isn't the spelling checker supposed to accept anything in CamelCase as a legal spelling?

Not that I know of. *Perhaps* if "Num" and "Py" were valid words, but those are not valid words either, so I wouldn't expect it to allow "NumPy".

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -176,6 +176,7 @@ discoverable
 discretization
 discretize
 discretized
+dtype
```
bmhowe23, line 179: dtype isn't a word. Wherever it is being used should be backticked (similar to the ndarray comment below).

.github/workflows/config/spelling_allowlist.txt
schweitzpgi: OK. I vaguely recall tweaking a regular expression somewhere once upon a time ... but it could be long gone by now.

runtime/cudaq/utils/tensor.h
anthony-santana, line 1: nit: should this file be in the `integration` folder instead of utils?

python/cudaq/__init__.py
schweitzpgi: Is it referenced/used anywhere? If not, let's just get rid of it.

python/tests/utils/test_tensor.py
annagrin: done

python/tests/utils/test_tensor.py
annagrin: done

annagrin: done

python/extension/CMakeLists.txt
```diff
@@ -138,6 +139,9 @@ add_mlir_python_modules(CUDAQuantumPythonModules
     CUDAQuantumPythonCAPI
   )
 
+set_target_properties(CUDAQuantumPythonModules.extension._quakeDialects.dso PROPERTIES CXX_VISIBILITY_PRESET "default")
+target_compile_options(CUDAQuantumPythonModules.extension._quakeDialects.dso PRIVATE -Wno-attributes)
```
amccaskey, line 143: I don't think the -Wno-attributes is needed, might have been left around from my prototyping

schweitzpgi: Should we remove both of these lines?

python/cudaq/utils/Tensor.py
```diff
@@ -0,0 +1,307 @@
...
+
+    @property
+    def dtype(self):
+        """
+        Get the data type of the tensor.
+
+        Returns:
+            `numpy.dtype`: The data type of the tensor elements.
+        """
+        return self._dtype
```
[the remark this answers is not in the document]
annagrin: we already have a `shape()` method, would you like me to replace it with a property?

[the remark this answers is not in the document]
schweitzpgi: The shape information is buried in the C++ object, so it would seem we need a function, no?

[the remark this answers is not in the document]
amccaskey: Was just thinking it would be convenient to have both 
```python
@property 
def shape(): return self._tensor.shape()
```
maybe its not possible to have both.

unittests/utils/tensor_tester.cpp
```diff
@@ -0,0 +1,282 @@
...
+  try {
+    t.at({2, 0});
+  } catch (...) {
+    eptr = std::current_exception();
+    auto info = eptr.__cxa_exception_type();
+    std::cout << "Caught exception of type : " << info->name() << std::endl;
+    throw std::runtime_error(info->name());
+  }
+
+  EXPECT_THROW(t.at({2, 0}), std::runtime_error);
```
annagrin, line 270: This test fails in CI `Create CUDA Quantum installer (xxx) / Build CUDA Quantum assets` task

annagrin: Trying to disable.

amccaskey:
Before merging this, you should add a test (and get it working) for using cudaq::tensor<T> in an application file compiled with `nvq++`. There are issues with regards to compiler compatibility and extension_point static initialization that may show up there. May need to remove extension_point and just use xtensor directly.

annagrin:
> Before merging this, you should add a test (and get it working) for using cudaq::tensor in an application file compiled with `nvq++`. There are issues with regards to compiler compatibility and extension_point static initialization that may show up there. May need to remove extension_point and just use xtensor directly.

I gave it a try in https://github.com/amccaskey/cuda-quantum/pull/6, let me know if this is in the right direction.

runtime/cudaq/utils/details/tensor_impl.h
```diff
@@ -0,0 +1,183 @@
...
+  // Double-dispatch hooks. We use double dispatch to ensure that both arguments
+  // are in fact the same derived class of `tensor_impl`.
+  virtual tensor_impl<Scalar> *
+  dd_multiply(const tensor_impl<Scalar> &left) const = 0;
+  virtual tensor_impl<Scalar> *
+  dd_add(const tensor_impl<Scalar> &left) const = 0;
+
+  // Terminal implementation of operators.
+  virtual tensor_impl<Scalar> *multiply(const xtensor<Scalar> &right) const = 0;
```
amccaskey, line 165: is the use of xtensor intended here?

schweitzpgi: Yes. This is using the old-school double dispatch pattern to determine the two input tensors have the same implementation. This is needed so that our matrix multiply code can efficiently (and correctly) access both of the input matrices to compute the product.

I'm thinking this implementation hierarchy ought to get rid of the virtual methods and be redone using the curiously recurring template pattern. However, I left it as it was to this point.

schweitzpgi:
Closing for now as an alternative solution was merged.

PR #2308, opened 2024-10-21
Enable nightly integration tests for `anyon` target

khalatepradnya:


schweitzpgi:
Have we determined why anyon is dropping output, etc.?

bmhowe23:
> Have we determined why anyon is dropping output, etc.?

Yes, that is being worked with Anyon right now.

bmhowe23 (approved):
This test passes now (manually run here: https://github.com/NVIDIA/cuda-quantum/actions/runs/12285298737), so this PR LGTM. Thanks, Pradnya!

Issue #2250, opened 2024-10-03
[Python] Returning from non-top kernels fails during compilation


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Returning a boolean value from a kernel that is called by and entry point kernel results in an error during compilation

### Steps to reproduce the bug

```python

import cudaq

@cudaq.kernel
def bar(qubits: cudaq.qview) -> bool:
    x(qubits)
    return False

@cudaq.kernel
def foo(n: int):
    qubits = cudaq.qvector(n)
    bar(qubits)

counts = cudaq.sample(foo, 3)
counts.dump()
```

### Expected behavior

Expected:

{ 000: 1000 }

Actual:
```
TypeError: isinstance(): incompatible function arguments. The following argument types are supported:
    1. (other: cudaq.mlir._mlir_libs._mlir.ir.Type) -> bool
 ```

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #2319, opened 2024-10-25
Fix incorrect call return for python device kernels

annagrin:
Detect correct return type of the callees in python when generating return statements in ast_bridge.py

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2250


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2162,6 +2174,11 @@ def bodyBuilder(iterVal):
                         node)
 
                 if node.func.attr == 'qvector':
+                    if len(self.valueStack) == 0:
+                        self.emitFatalError(
+                            'qvector does not have nullary constructor. init from size or existing state.',
```
khalatepradnya, line 2179: nit: Wording of the error message, is 'default' constructor Pythonic than 'nullary' constructor?

python/tests/kernel/test_call_return.py
```diff
@@ -0,0 +1,70 @@
...
+def test_call_with_return_bool():
+
+    @cudaq.kernel()
+    def callee(q: cudaq.qubit) -> bool:
+        h(q)
+        m = mz(q)
+        return m
+
+    @cudaq.kernel()
+    def caller() -> bool:
```
khalatepradnya, line 21: Q: I could be mistaken, or it could be a typo, but the example in issue #2250, makes it look like the caller function does not need to specify the return type hint?

khalatepradnya (approved):
Changes look good. Thanks, Anna! 
One clarifying Q and one nit.

amccaskey (approved):

python/tests/kernel/test_call_return.py
annagrin: The simpler test in the issue just checks for crashes, but this one should verify that the execution is also correct. Added both, thanks!

python/cudaq/kernel/ast_bridge.py
annagrin: done, thanks!

PR #2324, opened 2024-10-28
[runtime] Add a new matrix class.

schweitzpgi:
This PR supercedes and greatly simplifies the work in PR #2304 . The implementation is streamlined for the upcoming release.

### DIFF
--- a/runtime/cudaq/utils/tensor.cpp
+++ b/runtime/cudaq/utils/tensor.cpp
@@ -7,19 +7,23 @@
  ******************************************************************************/
 
 #include "cudaq/utils/tensor.h"
+#include <strstream>
 
 inline std::complex<double> &access(std::complex<double> *p,
                                     cudaq::matrix_2::Dimensions sizes,
-                                    unsigned row, unsigned col) {
+                                    std::size_t row, std::size_t col) {
   return p[row * sizes.second + col];
 }
 
 cudaq::matrix_2 &cudaq::matrix_2::operator*=(const cudaq::matrix_2 &right) {
-  auto new_data =
-      new std::complex<double>[dimensions.first * right.dimensions.second];
-  for (unsigned i = 0; i < dimensions.first; i++)
-    for (unsigned j = 0; j < right.dimensions.second; j++)
-      for (unsigned k = 0; k < dimensions.second; k++)
+  if (get_columns() != right.get_rows()) {
+    clear();
+    return *this;
+  }
+  auto new_data = new std::complex<double>[get_rows() * right.get_columns()];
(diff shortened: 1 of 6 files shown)


runtime/cudaq/utils/tensor.cpp
```diff
@@ -0,0 +1,102 @@
...
+  return p[row * sizes.second + col];
+}
+
+cudaq::matrix_2 &cudaq::matrix_2::operator*=(const cudaq::matrix_2 &right) {
+  if (get_columns() != right.get_rows()) {
+    clear();
+    return *this;
+  }
+  auto new_data = new std::complex<double>[get_rows() * right.get_columns()];
+  for (std::size_t i = 0; i < get_rows(); i++)
```
annagrin, line 24: do we need to check for the dimensions to match?

schweitzpgi: Yes. That's done on line 19.

BTW, errors like that are "reported" by returning an empty matrix. Feel free to suggest something more exciting.

annagrin: I suggest throwing an error

runtime/cudaq/utils/tensor.h
```diff
@@ -0,0 +1,174 @@
...
+    other.data = nullptr;
+  }
+  matrix_2(const std::vector<std::complex<double>> &v,
+           const Dimensions &dim = {2, 2})
+      : dimensions{dim}, data{new std::complex<double>[get_size(dim)]} {
+    assert(v.size() >= get_size(dimensions) &&
+           "vector must have enough elements");
+    std::copy(v.begin(), v.begin() + get_size(dimensions), data);
+  }
+  matrix_2(const std::complex<double> *v, const Dimensions &dim = {2, 2})
```
annagrin, line 55: Do you think we could use a `std::span<std::complex<double>>` instead of a pointer, so we can have boundary checks as well?

schweitzpgi: span doesn't own data and I think we want ownership semantics from the meeting.

schweitzpgi: The one case that one might get in trouble is building a matrix with a raw pointer to the data. I'm ok with that since I figure if one's using raw pointers, they need to know what they are doing.

schweitzpgi: The normal case, using a vector, has an assertion.

annagrin (approved):
LGTM!

annagrin (approved):
LGTM!

PR #2333, opened 2024-10-30
CUDA 11 and CUDA 12 tags for Docker image and cuQuantum version update

bettinaheim:
Also updates the cuquantum version.


.github/workflows/deployments.yml
```diff
@@ -263,6 +263,7 @@ jobs:
     strategy:
       matrix:
         platform: ${{ fromJson(needs.metadata.outputs.multi_platform || needs.metadata.outputs.platforms).ids }}
+        cuda_version: ["11.8", "12.0"]
```
bmhowe23, line 266: Do we need similar changes in validation_config.json? They are currently calling out only 11.8.

bettinaheim: The configuration in validation.json are only relevant for the wheels and the installer, i.e. not technically related to the docker image. The configurations for the wheels don't specify the cuda version, since it will be strictly tested based on what is in the readme (i.e. wheels are tested for both, *if* the runners actually have GPUs - in reality it is only validated during publishing/QA). The configurations you saw I think are the ones for the C++ installer, where the only system with CUDA 12 is debian. I need to check which images support newer cuda versions, and can add them - though again, this is not technically relevant since the runners that use this configuration do not have gpus...

docker/build/devdeps.ext.Dockerfile
```diff
@@ -125,12 +134,13 @@ ENV CUQUANTUM_PATH="$CUQUANTUM_INSTALL_PREFIX"
 ENV LD_LIBRARY_PATH="$CUQUANTUM_INSTALL_PREFIX/lib:$LD_LIBRARY_PATH"
 ENV CPATH="$CUQUANTUM_INSTALL_PREFIX/include:$CPATH"
 
-ENV CUQUANTUM_VERSION=24.03.0.4_cuda11
+ENV CUQUANTUM_VERSION=24.08.0.5
```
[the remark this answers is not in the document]
bettinaheim: The same update was made in configure_build.sh, which is what is (as of my previous PR) used by the cudaq.wheel.Dockerfile, as well as for the installer. I did forget the matching update in the pyproject dependencies, and added it to this PR.

bmhowe23 (approved):
LGTM

Issue #2330, opened 2024-10-30
Unexpected behavior with list[bool] for argument of kernel


ikkoham:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Unexpected behavior when using enumerate to create kernel.
Detail is below.

### Steps to reproduce the bug

```python
import cudaq
import itertools

@cudaq.kernel
def test(num_qubits: int, mask: list[bool]):
    q = cudaq.qvector(num_qubits)
    for i, m in enumerate(mask):
        if m:
            x(q[i])
    mz(q)
    
num_qubits = 3
for mask in itertools.product((True, False), repeat=num_qubits):
    print(mask)
    print(cudaq.sample(test, num_qubits, list(mask)))
```
outputs
```
(True, True, True)
{ 000:1000 }

(True, True, False)
{ 000:1000 }

(True, False, True)
{ 000:1000 }

(True, False, False)
{ 100:1000 }

(False, True, True)
{ 000:1000 }

(False, True, False)
{ 000:1000 }

(False, False, True)
{ 000:1000 }

(False, False, False)
{ 000:1000 }
```

Only the case (True, False, False) is correct.

### Expected behavior

Should return the correct bit string

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 0.8.0
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

ikkoham:
hmm, this may not be the problem of `enumerate`.

```python
import cudaq
import itertools

@cudaq.kernel
def test(num_qubits: int, mask: list[bool]):
    q = cudaq.qvector(num_qubits)
    for i in range(num_qubits):
        if mask[i]:
            x(q[i])
    mz(q)
    
num_qubits = 3
for mask in itertools.product((True, False), repeat=num_qubits):
    print(list(mask))
[...]

Issue #2262, opened 2024-10-09
Unexpected behavior when kernel takes list of booleans


jezerjojo14:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

I needed to send a list of $n$ booleans as a parameter to a kernel to specify which of $n$ qubits to apply $X$ gates on. The kernel does not throw any errors, but does not seem to be interpreting this list of booleans correctly. I'm not sure but I think it might be interpreting the list of booleans as a single integer?

### Steps to reproduce the bug

This is the kernel:

```
import cudaq
@cudaq.kernel
def circ_test(n: int, bools: list[bool]):
    q = cudaq.qvector(n)
    for j in range(n):
        b=bools[j]
        if b==False:
            x(q[j])

```

This is how it behaves for different examples:

1)
```
print(cudaq.draw(circ_test,5,[True,True,True,True,True]))
          
q0 : ─────
     ╭───╮
q1 : ┤ x ├
     ├───┤
q2 : ┤ x ├
     ├───┤
q3 : ┤ x ├
     ├───┤
q4 : ┤ x ├
     ╰───╯


```

2)
```
print(cudaq.draw(circ_test,5,[False,True,True,False,True]))

q0 : ─────
     ╭───╮
q1 : ┤ x ├
     ├───┤
q2 : ┤ x ├
     ├───┤
q3 : ┤ x ├
     ├───┤
q4 : ┤ x ├
     ╰───╯
```

3)

```
print(cudaq.draw(circ_test,5,[False,False,False,False,False]))

     ╭───╮
q0 : ┤ x ├
     ├───┤
q1 : ┤ x ├
     ├───┤
q2 : ┤ x ├
     ├───┤
q3 : ┤ x ├
     ├───┤
q4 : ┤ x ├
     ╰───╯
```

Here's a modified kernel:

```
import cudaq
@cudaq.kernel
def circ_test_2(n: int, bools: list[bool]):
[...]

PR #2349, opened 2024-11-01
Fix passing lists of boolean values to kernels in python.

annagrin:
### Description
Fix passing lists of boolean values to kernels in python.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2262
Closes: https://github.com/NVIDIA/cuda-quantum/issues/2330

### DIFF
--- a/python/utils/OpaqueArguments.h
+++ b/python/utils/OpaqueArguments.h
@@ -393,11 +393,11 @@ inline void packArgs(OpaqueArguments &argData, py::args args,
               .Case([&](IntegerType type) {
                 // Handle vec<bool> and vec<int>
                 if (type.getIntOrFloatBitWidth() == 1) {
-                  genericVecAllocator.template operator()<bool>(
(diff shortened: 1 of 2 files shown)


python/utils/OpaqueArguments.h
```diff
@@ -393,11 +393,11 @@ inline void packArgs(OpaqueArguments &argData, py::args args,
               .Case([&](IntegerType type) {
                 // Handle vec<bool> and vec<int>
                 if (type.getIntOrFloatBitWidth() == 1) {
-                  genericVecAllocator.template operator()<bool>(
+                  genericVecAllocator.template operator()<char>(
```
schweitzpgi, line 396: Is this required if we merge #2338 ?

schweitzpgi: The changes to this file can be dropped AFAICT.

It would be great to add the extended tests though.

PR #2355, opened 2024-11-05
Packaging fixes

bettinaheim:
Now that I finally can test the publishing, I needed to fix a couple of oversights in the previous updates to add the cuda 12 builds.


1tnguyen (approved):
LGTM 👍

PR #2359, opened 2024-11-06
Remove the TODO_loc on functions in the cudaq namespace

amccaskey:
Allow libraries to expose kernel code in cudaq namespace


schweitzpgi:
I don't think this is needed any longer. Closing for now.

PR #2361, opened 2024-11-06
Change the test port for `anyon` target 

khalatepradnya:
Changed from '5000' to '62446' to match other targets' ports.


bmhowe23 (approved):
Thanks!

PR #2366, opened 2024-11-07
Switch the default to enable mlir.

schweitzpgi:
Update the tests and the driver to make --enable-mlir the default. Add a new option, --library-mode, to allow the user to select the old default of library mode.


annagrin (approved):
LGTM!

PR #2370, opened 2024-11-08
Ignore terms with 0 coefficients

sacpis:
Ignore terms with 0 coefficients

# Test result
```
def hamiltonian_clique_graph(penalty, nodes, weights, non_edges) -> cudaq.SpinOperator:

    spin_ham = 0.0
    for wt, node in zip(weights, nodes):
        #print(wt,node)
        spin_ham += 0.5 * wt * spin.z(node)
        spin_ham -= 0.5 * wt * spin.i(node)

    for non_edge in non_edges:
        u, v = (non_edge[0], non_edge[1])
        spin_ham += penalty / 4.0 * (spin.z(u) * spin.z(v) - spin.z(u) -
                                     spin.z(v) + spin.i(u) * spin.i(v))

    return spin_ham

print(hamiltonian_clique_graph(2., [0,1], [1., 1.], [[0,1]]))
```
```
root@ae0fb315fdde:/workspaces/cuda-quantum# python3 docs/sphinx/examples/python/test1.py 
[0.5+0j] ZZ
[-0.5+0j] II
```


runtime/cudaq/spin/spin_op.cpp
```diff
@@ -414,9 +414,11 @@ spin_op &spin_op::operator+=(const spin_op &v) noexcept {
 
   for (auto [term, coeff] : tmpv.terms) {
     auto iter = terms.find(term);
-    if (iter != terms.end())
+    if (iter != terms.end()) {
       iter->second += coeff;
-    else
+      if (std::abs(iter->second) < 1e-12)
+        terms.erase(iter);
+    } else
```
schweitzpgi, line 421: Even though the `else` block is a simple one liner, let's put braces around it to match the `if then` part. I think that'll make it evre so slightly easier to skim-read.

sacpis:
Closing this PR as it has been fixed with the Operators.

PR #2375, opened 2024-11-13
Allow to define publishing assets as repo variable, disable codeql for C++, and fix issue in nightly workflows

bettinaheim:
This will allow to configure the repo such that all publishing runs bypass rebuilding these assets.


schweitzpgi (approved):
+1

PR #2376, opened 2024-11-13
Excluding building_kernels from NVQC integration test

sacpis:
Excluding building_kernels from NVQC integration test

This will fix the NVQC integration test below
https://github.com/NVIDIA/cuda-quantum/actions/runs/11789940366/job/32839604580#step:4:1667


bmhowe23 (approved):
Thanks, Sachin.

Issue #2325, opened 2024-10-28
Some target tests don't work with translation to OpenQASM 2.0


kshyatt-aws:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Creating a new backend which uses OpenQASM 2.0 as the IR breaks several of the `targettests/execution` tests, including:

- `custom_operation_basic.cpp` - custom ops aren't supported in OQ2
- `custom_operation_adj.cpp`
- `custom_operation_ctrl.cpp`
These 3 fail with:
```bash
+ /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/custom_operation_adj.cpp
loc("-":30:1): error: 'cc.global' op unable to translate op to OpenQASM 2.0
terminate called after throwing an instance of 'std::runtime_error'
  what():  Could not successfully translate to qasm2.
FileCheck error: '<stdin>' is empty.
FileCheck command line:  /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/custom_operation_adj.cpp
```

Also failing are:
- `graph_coloring-1.cpp`
- `graph_coloring.cpp`
with:
```bash
+ /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/graph_coloring-1.cpp
loc(callsite("graph_coloring-1.cpp":42:3 at "graph_coloring-1.cpp":86:3)): error: cannot convert operation to OpenQASM 2.0.
terminate called after throwing an instance of 'std::runtime_error'
  what():  Could not successfully translate to qasm2.
FileCheck error: '<stdin>' is empty.
FileCheck command line:  /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/graph_coloring-1.cpp
```
and
```bash
[...]

PR #2379, opened 2024-11-15
Fix failing tests that use TranslateToOpenQASM pipeline

annagrin:
### Description
Fix failing tests that use `TranslateToOpenQASM` pipeline

- Update `cc.GlobalOp` with symbol visibility so it can be removed by `symbol-dce` pass if not used
  - set private visibility on various auxiliary global ops created during other passes.
- Run additional passes as part of pipelines that translate to OpenQasm2:
  - all passes related to unitary synthesis
  - decomposition patterns "CCzToCx"
  - add new decomposition patterns  "RxAdjToRx", "RyAdjToRy","RyAdjToRy",
  - symbol-dce
  - and more
 - Update bracket pipeline with missing passes as well
 - Add tests

Towards: https://github.com/NVIDIA/cuda-quantum/issues/2325
Note: a few tests from that issue are still failing:
- custom_operation_ctrl.cpp
- qspan_slices.cpp
- phase_estimation.cpp


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -407,7 +407,7 @@ def GetConcreteMatrix : Pass<"get-concrete-matrix", "mlir::func::FuncOp"> {
           return
         }
 
-        cc.global constant @__nvqpp__mlirgen__function_foo_generator_1.bar.rodata_0 ((dense<[(0.000000e+00,0.000000e+00), (1.000000e+00,0.000000e+00), (1.000000e+00,0.000000e+00), (0.000000e+00,0.000000e+00)]> : tensor<4xcomplex<f64>>) : !cc.array<complex<f64> x 4>
+        cc.global constant @__nvqpp__mlirgen__function_foo_generator_1.bar.rodata_0 ((dense<[(0.000000e+00,0.000000e+00), (1.000000e+00,0.000000e+00), (1.000000e+00,0.000000e+00), (0.000000e+00,0.000000e+00)]> : tensor<4xcomplex<f64>>) : !cc.array<complex<f64> x 4> {sym_visibility = "private"}
```
schweitzpgi, line 410: Can we make this pretty-print nicer?
```
cc.global constant private @__nvqpp__mlirgen_function....
```

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -1181,6 +1181,41 @@ struct RxToPhasedRx : public OpRewritePattern<quake::RxOp> {
   }
 };
 
+// quake.rx<adj> (θ) target
+// ─────────────────────────────────
+// quake.rx(-θ) target
```
schweitzpgi, line 1186: These adjoint forms are a bit surprising. Do you know what is generating them? The code that generates generalized adjoints doesn't use these. It just negates the angle argument.

I think we should probably get rid of these `r?<adj>` as a valid operation. We did that with `x<adj>` and such.

targettests/execution/custom_operation_adj.cpp
```diff
@@ -6,12 +6,15 @@
...
  ******************************************************************************/
 
-// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target anyon --emulate %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target ionq --emulate %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target iqm --emulate --iqm-machine Apollo %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target oqc --emulate %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t | FileCheck %s
+// clang-format off
+// RUN: nvq++ -std=c++17 --enable-mlir                             %s -o %t && %t | FileCheck %s
```
schweitzpgi, line 10: I find all this whitespace sort of annoying for cut-n-paste reasons.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
annagrin: I also found that some decomposition patterns create them, "CRxToCX" for example

annagrin: chatted offline, will postpone the `adjoint` removal for later

schweitzpgi (approved):
Thanks, Anna.

PR #2380, opened 2024-11-15
Cudart installation for mpi tests

sacpis:
Cudart installation for mpi tests


bettinaheim:
Looks like the conda packages are fixed.

PR #2385, opened 2024-11-15
Make remote-mqpu auto-launcher aware of CUDAQ_DYNLIBS env var

bmhowe23:
If auto-launch is used from a Python environment, it is aware of where the custatevec and cutensornet libs are through the CUDAQ_DYNLIBS environment variable, but it was counting on the LD_LIBRARY_PATH to find them instead. This updates the logic to use the CUDAQ_DYNLIBS if the environment variable is present.


bettinaheim (approved):

runtime/cudaq/platform/mqpu/helpers/MQPUUtils.cpp
```diff
@@ -79,6 +79,26 @@ cudaq::AutoLaunchRestServerProcess::AutoLaunchRestServerProcess(
...
+      preload = p;
+      if (preload.size() > 0)
+        preload += ":";
+    }
+    preload += std::string(ch);
+    for (char **env = environ; *env != nullptr; ++env) {
+      if (!std::string(*env).starts_with("LD_PRELOAD="))
+        Env->push_back(*env);
+    }
+    preload = "LD_PRELOAD=" + preload;
```
bettinaheim, line 98: The issue with using LD_PRELOAD is that it fails if the libraries can't be loaded. 
Currently, we don't have a CUDA-Q build shipped separately without GPU components. Hence, LD_PRELOAD makes it impossible to use CUDA-Q without GPU/a CUDA installation right now. We either need to ship a version without CUDA (and generally I prefer not adding more packages), or we need a non-failing alternative.

bettinaheim: What was the reason to use LD_PRELOAD instead of just adding these to the LD_LIBRARY_PATH like we do in the *qpud.py?

bettinaheim: I took the liberty of pushing two commits that should have the (almost) exact same logic as we have in the python script for qpud.

bettinaheim:
Note that testing this change fully requires a GPU runner, and hence will only happen post-merge as far as the pipelines are concerned (hence the failing codecov check).

Issue #2383, opened 2024-11-15
Unexpected JIT compilation error when `cudaq.observe` from Python


bmhowe23:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Running cudaq.observe() from a Python test program shown below produces the following error:
```bash
error: 'cc.address_of' op must reference a global
RuntimeError: Could not apply measurements to ansatz.
```

### Steps to reproduce the bug

Save the following to `test.py`
```
import cudaq
from typing import List

cudaq.set_target("oqc", emulate=True)

# Initialize a kernel/ ansatz and variational parameters.
@cudaq.kernel
def kernel(angles: List[float]):
    # Allocate a qubit that is initialized to the |0> state.
    qubit = cudaq.qubit()
    # Define gates and the qubits they act upon.
    rx(angles[0], qubit)
    ry(angles[1], qubit)


# Our Hamiltonian will be the Z expectation value of our qubit.
hamiltonian = cudaq.spin.z(0)

# Initial gate parameters which initialize the qubit in the zero state
initial_parameters = [0, 0]

cost_values = []

def cost(parameters):
    """Returns the expectation value as our cost."""
    expectation_value = cudaq.observe(kernel, hamiltonian,
                                      parameters).expectation()
    cost_values.append(expectation_value)
    return expectation_value

# We see that the initial value of our cost function is one, demonstrating that our qubit is in the zero state
initial_cost_value = cost(initial_parameters)
print(initial_cost_value)
```
And then run the test program:
```bash
$ python3 test.py
error: 'cc.address_of' op must reference a global
RuntimeError: Could not apply measurements to ansatz.
```

### Expected behavior

The program should run without errors. Note that a very similar program in C++ (targettests/execution/cudaq_observe.cpp in the repo) seems to run without errors.

### Is this a regression? If it is, put the last known working version (or commit) here.

Yes; last known to have worked in 0.8.0

### Environment

[...]

PR #2392, opened 2024-11-18
Fix for issue #2383.

schweitzpgi:
Instead of erroneously trying to copy a part of the moduleOp, clone the entire moduleOp so that any referenced artifacts are properly copied.

Fixes issue #2383.


khalatepradnya (approved):
LGTM. Thanks, Eric!

PR #2401, opened 2024-11-21
[braket] Python tests

khalatepradnya:
Support for the `braket` target with Python frontend


python/runtime/utils/PyRemoteRESTQPU.cpp
```diff
@@ -89,7 +89,9 @@ class PyRemoteRESTQPU : public cudaq::BaseRemoteRESTQPU {
     pm.addPass(mlir::createCanonicalizerPass());
     pm.addPass(cudaq::opt::createApplyOpSpecializationPass());
     pm.addPass(createInlinerPass());
-    pm.addPass(cudaq::opt::createExpandMeasurementsPass());
+    /// NOTE: This change is for the `braket` target. May require additional
+    /// pass in other pipelines.
+    // pm.addPass(cudaq::opt::createExpandMeasurementsPass());
```
schweitzpgi, line 94: Is this pipeline spec common to all targets?

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+    counts = cudaq.sample(kernel, shots_count=100)
+    assert len(counts) == 2
+    assert "00" in counts
+    assert "11" in counts
+
+
+def test_all_gates():
+
+    @cudaq.kernel
+    def single_qubit_gates():
```
bettinaheim, line 91: It would be good to also add a test for all the singly-controlled gates, and another test for all the adjoints.

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+        q0 = cudaq.qubit()
+        q1 = cudaq.qubit()
+        h(q0)
+        cx(q0, q1)
+        mz(q0)
+        mz(q1)
+
+    with pytest.raises(RuntimeError) as e:
+        cudaq.sample(kernel, shots_count=100)
+    assert "cannot declare a qubit register. Only 1 qubit register(s) is/are supported" in repr(
```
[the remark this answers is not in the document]
schweitzpgi: Yes. And agree this could be a confusing error message. "A single register of qubits" would be clearer.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -603,6 +622,34 @@ struct SToR1 : public OpRewritePattern<quake::SOp> {
   }
 };
 
+// quake.s<adj> target
+// ─────────────────────────────────
+// (quake.z * quake.s) target
```
schweitzpgi, line 627:
```suggestion
// quake.s<adj> [control] target
// ─────────────────────────────────
// quake.z [control] target
// quake.s [control] target
```

python/runtime/utils/PyRemoteRESTQPU.cpp
annagrin: it is but I think it should be in the target.yml file already for cpp code to work

python/tests/backends/test_braket.py
khalatepradnya: Done. Updated `test_control_modifier` and `test_adjoint_modifier`

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+        x(q)
+        y(q)
+        z(q)
+        # r1(np.pi, q) ## Unsupported
+        rx(np.pi, q)
+        ry(np.pi, q)
+        rz(np.pi, q)
+        s(q)
+        t(q)
+        # mx(q) ## Unsupported
```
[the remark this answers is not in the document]
khalatepradnya: Won't fix now, may address in follow-up PR.

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+    assert "11" in counts
+
+
+def test_adjoint_modifier():
+
+    @cudaq.kernel
+    def single_adjoint_test():
+        q = cudaq.qubit()
+        s(q)
+        s.adj(q)
```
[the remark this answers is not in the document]
khalatepradnya: Fixed with your patch.

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+        ## NOTE: Measure required since 'Device requires all qubits in the program to be measured.'
+        mz(qreg)
+
+    # Define its spin Hamiltonian.
+    hamiltonian = 5.907 - 2.1433 * spin.x(0) * spin.x(1) - 2.1433 * spin.y(
+        0) * spin.y(1) + .21829 * spin.z(0) - 6.125 * spin.z(1)
+
+    with pytest.raises(RuntimeError) as e:
+        cudaq.observe(ansatz, hamiltonian, .59, shots_count=100)
+    assert "observe specification violated for 'ansatz': kernels passed to observe cannot have measurements specified." in repr(
```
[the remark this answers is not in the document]
khalatepradnya: Will be handled separately.

python/tests/backends/test_braket.py
```diff
@@ -0,0 +1,292 @@
...
+@pytest.mark.parametrize("device_arn", [
+    "arn:aws:braket:::device/quantum-simulator/amazon/dm1",
+    "arn:aws:braket:::device/quantum-simulator/amazon/tn1"
+])
+def test_other_simulators(device_arn):
+    cudaq.set_target("braket", machine=device_arn)
+    test_qvector_kernel()
+    test_builder_sample()
+    cudaq.reset_target()
+
```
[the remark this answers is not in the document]
khalatepradnya: Yes, following fails -
```@cudaq.kernel
    def kernel():
        qubits = cudaq.qvector(2)
        h(qubits[0])
        mz(qubits[0])
        mz(qubits[1])
```
with `cannot declare bit register. Only 1 bit register(s) is/are supported`
Added new test `test_multiple_measurement`

python/runtime/utils/PyRemoteRESTQPU.cpp
khalatepradnya: I am leaning towards keeping this comment as reminder in case we see any failures in integration tests.

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
```diff
@@ -22,24 +23,20 @@ using namespace cudaq;
 //===----------------------------------------------------------------------===//
 // Helper functions
 //===----------------------------------------------------------------------===//
-
+/// FIXME: Re-evaluate this renaming workflow; consider using decomposition
```
khalatepradnya, line 26: I think we may want to retain this logic to map the OpenQASM2.0 gate set.

test/Translate/openqasm2_adj_rotations.cpp
```diff
@@ -49,13 +51,13 @@ int main() {
 
 // CHECK:  qreg var0[6];
 // CHECK:  x var0[5];
-// CHECK:  ch var0[4], var0[5];
+// CHECK:  h var0[4], var0[5];
```
bettinaheim, line 54: this is not the same as ch; this will instead lead to a broadcasting. 
We have two options:
- translate ch into a singly-controlled rotation (seems like the best path)
- see if AWS supports ctrl(3) @ ... for multi-controlled gates.

lib/Optimizer/CodeGen/Passes.cpp
```diff
@@ -21,6 +21,7 @@ static void addAnyonPPipeline(OpPassManager &pm) {
   };
   BasisConversionPassOptions options;
   options.basis = basis;
+  options.disabledPatterns = {"R1ToU3"};
```
bmhowe23, line 24: `disabledPatterns` is an ArrayRef, and since the temporary vector is destroyed after the statement, this will cause erroneous behavior. https://gist.github.com/bmhowe23/e84a6ea82820bdae3a5ead11f1808610 is one possible alternative that seems to work when I did local testing.

annagrin (approved):
LGTM, Thanks Pradnya!

PR #2406, opened 2024-11-25
Fix `NoiseModel` for single-precision backends

1tnguyen:
### Description

`NoiseModel.cpp` is having some dangling dependency on `host_config.h` (defining target-dependent data type).
This may cause data corruption when a `fp32` backend links to `libcudaq-common`, which is always compiled with fp64 config. 

In particular,

- A single instance of `validateCompleteness` is called in `.cpp` file -> the fix is to move it to the header.

- The `kraus_channel::parameters` is declared as a `host_config.h`'s `real` type. As we don't expect channels with a long list of parameters, I think it makes sense to just use `double` type instead. This prevents subtle issues with copy operators being implemented in `NoiseModel.cpp`.

### DIFF
--- a/runtime/common/NoiseModel.cpp
+++ b/runtime/common/NoiseModel.cpp
@@ -78,10 +78,6 @@ void validateCompletenessRelation_fp64(const std::vector<kraus_op> &ops) {
         "Provided kraus_ops are not completely positive and trace preserving.");
 }
 
-kraus_channel::kraus_channel(std::vector<kraus_op> &_ops) : ops(_ops) {
-  validateCompleteness();
-}
-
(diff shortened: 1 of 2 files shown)


bmhowe23 (commented):
This LGTM, but since NoiseModel.h is a public header file, it would be good for others (@amccaskey and @boschmitt) to review as well.

amccaskey (approved):
This looks good to me. We hit a similar issue in cuda-qx. The NoiseModel is acts like a template type (in some sense) due to the preprocessor dependence. Therefore the best workarounds are to keep `cudaq::complex` / `cudaq::real` things in the header.

PR #2407, opened 2024-11-25
[WIP] Operators and Integration with cudensitymat

sacpis:
Adding missing operations in operators

Co-authored with @anthony-santana.


runtime/cudaq/dynamics/cudm_helpers.cpp
```diff
@@ -0,0 +1,209 @@
...
+          std::vector<std::complex<double>> flat_matrix;
+          for (size_t i = 0; i < matrix.get_rows(); i++) {
+            for (size_t j = 0; j < matrix.get_columns(); j++) {
+              flat_matrix.push_back(matrix[{i, j}]);
+            }
+          }
+
+          // Create a cudensitymat elementary operator
+          HANDLE_CUDM_ERROR(cudensitymatCreateElementaryOperator(
+              handle, 1, mode_extents.data(),
```
1tnguyen, line 133: Q: is the `mode_extents` here all the extents? 
If so, we need to retrieve the extent of the sub-space this elementary operation acts on.

runtime/cudaq/dynamics/cudm_helpers.cpp
```diff
@@ -0,0 +1,209 @@
...
+              CUDENSITYMAT_OPERATOR_SPARSITY_NONE, 0, nullptr, CUDA_C_64F,
+              flat_matrix.data(), {nullptr, nullptr}, &cudm_elem_op));
+
+          // Append the elementary operator to the term
+          HANDLE_CUDM_ERROR(cudensitymatOperatorTermAppendElementaryProduct(
+              handle, term, 1, &cudm_elem_op, &elem_op.degrees[0], nullptr,
```
1tnguyen, line 139: It might be illegal to pass `nullptr` to `modeActionDuality`.

runtime/cudaq/dynamics/cudm_helpers.cpp
```diff
@@ -0,0 +1,209 @@
...
+              make_cuDoubleComplex(1.0, 0.0), {nullptr, nullptr}));
+
+          // Destroy the elementary operator after appending
+          HANDLE_CUDM_ERROR(
+              cudensitymatDestroyElementaryOperator(cudm_elem_op));
```
1tnguyen, line 144: We'll need to be careful about this. Perhaps we cannot destroy it yet as we passed a pointer to `cudensitymatOperatorTermAppendElementaryProduct`

runtime/cudaq/dynamics/cudm_helpers.cpp
```diff
@@ -0,0 +1,209 @@
...
+        } else if (std::holds_alternative<cudaq::scalar_operator>(component)) {
+          const auto &scalar_op = std::get<cudaq::scalar_operator>(component);
+
+          // Use the scalar coefficient
+          auto coeff = scalar_op.evaluate(parameters);
+          HANDLE_CUDM_ERROR(cudensitymatOperatorTermAppendElementaryProduct(
+              handle, term, 0, nullptr, nullptr, nullptr,
+              {make_cuDoubleComplex(coeff.real(), coeff.imag())},
+              {nullptr, nullptr}));
```
1tnguyen, line 153: We'll need to add a `TODO` here to handle callback scalars (time-dependent) via `cudensitymatScalarCallback_t`.

unittests/CMakeLists.txt
```diff
@@ -44,11 +44,25 @@ set(CUDAQ_RUNTIME_TEST_SOURCES
...
+  dynamics/elementary_ops_simple.cpp
+  dynamics/elementary_ops_arithmetic.cpp
+  dynamics/product_operators_arithmetic.cpp
+  dynamics/test_runge_kutta_time_stepper.cpp
+  dynamics/test_runge_kutta_integrator.cpp
+  dynamics/test_helpers.cpp
+  dynamics/rydberg_hamiltonian.cpp
+  dynamics/test_cudm_helpers.cpp
+  dynamics/test_cudm_state.cpp
+  dynamics/test_cudm_time_stepper.cpp
```
1tnguyen, line 58: These `dynamics` tests are very different from the usual `CUDAQ_RUNTIME_TEST_SOURCES` so we should move them into a separate target. Adding them here will cause these tests to be repeated for all simulator backends.

unittests/CMakeLists.txt
```diff
@@ -44,11 +44,25 @@ set(CUDAQ_RUNTIME_TEST_SOURCES
...
 )
 
 # Make it so we can get function symbols
 set (CMAKE_ENABLE_EXPORTS TRUE)
 
+include_directories(/usr/local/cuda/targets/x86_64-linux/include)
```
1tnguyen, line 64: The CUDA `include` and `lib` paths should be detected using CMake. We probably want to wrap it in a `CUDA_FOUND` block as well since not all build environment has CUDA.

unittests/CMakeLists.txt
```diff
@@ -375,7 +415,8 @@ target_link_libraries(${TEST_EXE_NAME}
   cudaq-platform-default
   cudaq-rest-qpu
   cudaq-builder
-  gtest_main)
+  gtest_main
+  $ENV{CUQUANTUM_INSTALL_PREFIX}/lib/libcudensitymat.so.0)
```
1tnguyen, line 419: We shouldn't need to add this to existing unit tests.

runtime/cudaq/dynamics/cudm_state.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include <numeric>
+#include <sstream>
+#include <stdexcept>
+
+namespace cudaq {
+
+cudm_state::cudm_state(std::vector<std::complex<double>> rawData)
+    : rawData_(rawData), state_(nullptr), handle_(nullptr),
+      hilbertSpaceDims_() {
+  HANDLE_CUDM_ERROR(cudensitymatCreate(&handle_));
```
1tnguyen, line 21: We should only have a single `cudensitymatHandle_t` throughout the code (i.e., a single `cudensitymatCreate` and `cudensitymatDestroy`).

For example, we already create a handle in `test_cudm_time_stepper.cpp` (`SetUp`); thus creating another handle here is problematic.

I'd suggest passing the handle here, e.g. 

```
cudm_state::cudm_state(cudensitymatHandle_t handle,
                       const std::vector<std::complex<double>> &rawData)
    : rawData_(rawData), state_(nullptr), handle_(handle), hilbertSpaceDims_() 
 ```

File: runtime/cudaq/dynamics/cudm_state.cpp [hunk 9]
```diff
@@ -0,0 +1,252 @@
...
+}
+
+cudm_state::~cudm_state() {
+  if (state_) {
+    cudensitymatDestroyState(state_);
+  }
+  if (handle_) {
+    cudensitymatDestroy(handle_);
+  }
```
1tnguyen, line 30: Like above, we shouldn't be creating and destroying the handle during state object life cycle.

runtime/cudaq/dynamics/cudm_state.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Attach storage for each component
+  std::vector<void *> componentBuffers(numStateComponents);
+  std::vector<std::complex<double> *> rawComponentData(numStateComponents);
+
+  size_t offset = 0;
+  for (int32_t i = 0; i < numStateComponents; i++) {
+    rawComponentData[i] =
+        reinterpret_cast<std::complex<double> *>(rawData_.data()) + offset;
+    componentBuffers[i] = static_cast<void *>(rawComponentData[i]);
```
1tnguyen, line 171: We're supposed to attached device memory here (`rawData_` is a host array).
I think we need to add a `gpuData_` member var and perform `cudaMalloc` + `cudaMemcpy`.

runtime/cudaq/dynamics/cudm_time_stepper.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (!liouvillian_) {
+        throw std::runtime_error("Liouvillian is not initialized.");
+    }
+
+    std::cout << "cudensitymatOperatorComputeAction ..." << std::endl;
+    HANDLE_CUDM_ERROR(cudensitymatOperatorComputeAction(handle_, liouvillian_, t, 0, nullptr, state.get_impl(), next_state.get_impl(), workspace, 0));
+
+    std::cout << "Update the state ..." << std::endl;
+    // Update the state
+    state = std::move(next_state);
```
1tnguyen, line 51: Should we update the input state or return the action result (RHS) in this function?
We might need the latter since the RHS might be added to the input state with some scaling depending on the integration algorithm.

unittests/dynamics/test_cudm_time_stepper.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    void SetUp() override {
+        // Create library handle
+        HANDLE_CUDM_ERROR(cudensitymatCreate(&handle_));
+
+        // Create a mock Liouvillian
+        liouvillian_ = mock_liouvillian(handle_);
+
+        // Initialize the time stepper
+        time_stepper_ = new cudm_time_stepper(liouvillian_, handle_);
```
1tnguyen, line 55: This is a subtle but serious bug: the two `void*` pointers are swapped! :-)
The function signature requires a handle first then an operator.

PR #2414, opened 2024-11-25
[nvq++][rt] Improve binary sizes by not globing MLIR dialects/passes.

boschmitt:
### Description
By not globing all MLIR dialects/passes, many of which CUDAQ does not use, we can significantly improve binary sizes:

```
Filename                    Size (old)      Size (new)      Improvement
--------                    -----------     -----------     -----------
cudaq-opt                   161.9 MB        39.6 MB         -122.2 MB
cudaq-quake                 187.6 MB        108.6 MB        -79.0 MB
cudaq-translate             138.2 MB        76.1 MB         -62.0 MB
libcudaq-mlir-runtime.so    179.9 MB        111.8 MB        -68.1 MB
```

This change also improves compilation time when building `llvm` along with CUDAQ, instead of using a pre-built version for llvm. It also improves the `--help` output for the `cudaq-*` tools, as they won't be "polluted" with many passes/options that are not relevant in the context of `nvq++`.

Improving the python bindings situation, which can possibly profit from a similar change, will be left to a later commit.

This PR is part of a series that address the long standing need to refactor the build system.


include/cudaq/Optimizer/InitAllDialects.h
```diff
@@ -10,13 +10,34 @@
...
+    // MLIR dialects
+    mlir::AffineDialect,
+    mlir::arith::ArithDialect,
+    mlir::cf::ControlFlowDialect,
+    mlir::complex::ComplexDialect,
+    mlir::func::FuncDialect,
+    mlir::LLVM::LLVMDialect,
+    mlir::math::MathDialect,
+    mlir::memref::MemRefDialect,
+    mlir::scf::SCFDialect,
```
schweitzpgi, line 38: We shouldn't be using Affine, MemRef,, or SCF. There may be some dusty deck tests that use those, but we don't need or want to use them.

boschmitt: I agree, I think there are some really old test that actually use those. I just didn't wanted to go ahead removing them before getting your eyes on this (:

bmhowe23: This PR is a nice find, @boschmitt! So is the plan to make the Affine, MemRef, and SCF changes as part of this PR or to leave them for a future PR?

boschmitt: I spoke with Eric offline (should have posted here). I think it would be best to leave this PR as-is, and work on removing those in a different one (I can do that too). I wanted to keep code/test changes to a minimum to facilitate review and making sure the different steps don't break other things. (I could also create a PR train base on this one.)

schweitzpgi (approved):
+1

PR #2416, opened 2024-11-25
Support measurements of subveqs for bracket 

annagrin:
### Description
Support measurements of subveqs for braket

- Add support for setting and using output names in braket executor and server helper
- Add `CombineMeasurements` pass
  - extend measurements from `subveq` to the full vector and add `output_names` attributes
  - combine separate measurements into one
- Remove passes in `translateToOpenQasm` pipeline that would be normally specified in target config files.
- Add tests


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -141,6 +141,32 @@ def CheckKernelCalls : Pass<"check-kernel-calls", "mlir::func::FuncOp"> {
...
+      %1 = ... : !quake.veq<4>
+      %2 = quake.subveq %1, %c2, %c3 : (!quake.veq<4>, i32, i32) ->
+            !quake.veq<2>
+      %measOut = quake.mz %2 : (!quake.veq<2>) -> !cc.stdvec<!quake.measure>
+    }
+    ```
+    with:
+    ```
+    func.func @kernel() attributes {"cudaq-entrypoint", ["output_names",
+    "[[[0,[1,\22q0\22]],[1,[2,\22q1\22]]]]"]} {
```
schweitzpgi, line 160: Why the escaped characters?

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -45,20 +45,11 @@ void cudaq::opt::addPipelineTranslateToOpenQASM(PassManager &pm) {
...
-  pm.addPass(createLoopUnroll());
-  pm.addPass(createCanonicalizerPass());
-  pm.addNestedPass<func::FuncOp>(createLiftArrayAlloc());
-  pm.addPass(createGlobalizeArrayValues());
-  pm.addPass(createStatePreparation());
-  pm.addNestedPass<func::FuncOp>(createGetConcreteMatrix());
-  pm.addPass(createUnitarySynthesis());
+  pm.addNestedPass<func::FuncOp>(createExpandControlVeqs());
+  pm.addNestedPass<func::FuncOp>(createCombineQuantumAllocations());
+  pm.addNestedPass<func::FuncOp>(createCombineMeasurements());
```
schweitzpgi, line 50: Does using OpenQASM2 require measurements to be replaced with "measure everything and sort it out later"?

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -460,6 +460,25 @@ struct R1ToRz : public OpRewritePattern<quake::R1Op> {
   }
 };
 
+// Naive mapping of R1 to U3
+// This is only expected to work with full inlining and
+// quake apply specialization.
```
schweitzpgi, line 465: I think this is universal wrt decompositions. Comment isn't needed.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -562,7 +562,7 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createLoopNormalize());
-  pm.addNestedPass<func::FuncOp>(cudaq::opt::createLoopUnroll());
+  pm.addPass(cudaq::opt::createLoopUnroll());
```
schweitzpgi, line 565: Why would we want to run normalization and unrolling at different levels? Seems better if we do not.

python/runtime/utils/PyRemoteRESTQPU.cpp
```diff
@@ -89,7 +89,6 @@ class PyRemoteRESTQPU : public cudaq::BaseRemoteRESTQPU {
     pm.addPass(mlir::createCanonicalizerPass());
     pm.addPass(cudaq::opt::createApplyOpSpecializationPass());
     pm.addPass(createInlinerPass());
-    pm.addPass(cudaq::opt::createExpandMeasurementsPass());
```
schweitzpgi, line 92: Is this pipeline used by targets other than Braket?

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: that's how quotes appear in the following mlir attirbutes`[[[0,[1,"q0"]],[1,[2,"q1"]]]]"]`. Note that it is a pre-existing condition:)

lib/Optimizer/CodeGen/Pipelines.cpp
annagrin: Yes, I am following a similar path to what we do in Ionq: measure everything, keep a mapping for the qubits we want to report, and filter out counts after the execution.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
annagrin: I am relying on Preadnya's PR here

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: it only works as a non-nested pass, this was giving me an errror

python/runtime/utils/PyRemoteRESTQPU.cpp
annagrin: Also comes from Pradnya's PR but it shouldn't be used in common pipelnes, we should add it to the yml files if needed instead. Since it works for c++ already I presume we can just remove it here, but I will fix the yml files if the tests fail

lib/Optimizer/Transforms/DecompositionPatterns.cpp
khalatepradnya: I have now fixed it in PR#2401 with 09011e866360f1bedfe0070c762ab3cc393b2503

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: updated both to run at module level, thanks!

lib/Optimizer/CodeGen/Pipelines.cpp
annagrin: moved the extra passes to python's code.

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -0,0 +1,258 @@
...
+      }
+    }
+
+    return success();
+  }
+};
+
+class ExtendQubitMeasurePattern : public OpRewritePattern<quake::MzOp> {
+  using OutputNamesType =
+      std::map<std::size_t, std::pair<std::size_t, std::string>>;
```
bmhowe23, line 91: This looks the same as above. Should it be pulled out separately and used as a common definition?

runtime/cudaq/platform/default/rest/helpers/braket/braket.yml
```diff
@@ -17,9 +17,11 @@ config:
...
+  platform-lowering-config: "func.func(const-prop-complex,canonicalize,cse,lift-array-alloc),globalize-array-values,state-prep,unitary-synthesis,canonicalize,apply-op-specialization,aggressive-early-inlining,unrolling-pipeline,func.func(lower-to-cfg),canonicalize,func.func(multicontrol-decomposition),decomposition{enable-patterns=SToR1,TToR1,R1ToU3,U3ToRotations,CHToCX,CCZToCX,CRzToCX,CRyToCX,CRxToCX,CR1ToCX,RxAdjToRx,RyAdjToRy,RzAdjToRz},func.func(expand-control-veqs,combine-quantum-alloc,canonicalize,combine-measurements)"
   # Tell the rest-qpu that we are generating OpenQASM 2.0.
   codegen-emission: qasm2
+  # Additional passes to run after lowering to QUAKE
+  #post-codegen-passes: "remove-measurements"
```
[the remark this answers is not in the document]
annagrin: We don't convert to QIR and llvm in this case, so the pass was not doing anything:)

lib/Optimizer/Transforms/CombineMeasurements.cpp
annagrin: Done, thanks!

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+#define DEBUG_TYPE "combine-measurements"
+
+using namespace mlir;
+
+namespace {
+
+struct Analysis {
+  using OutputNamesType =
+      std::map<std::size_t, std::pair<std::size_t, std::string>>;
```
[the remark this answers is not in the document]
annagrin: Added comments

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
```diff
@@ -23,13 +23,12 @@ using namespace cudaq;
 //===----------------------------------------------------------------------===//
 // Helper functions
 //===----------------------------------------------------------------------===//
-
```
schweitzpgi, line 26: nit: Leave the blank line since the banner comment and the doxygen comment are really doing different things here.

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -0,0 +1,255 @@
...
+  // %0 = ...: !quake.veq<2>
+  // %1 = quake.extract_ref %0[0] : (!quake.veq<2>) -> !quake.ref
+  // %measOut = quake.mz %1 : (!quake.ref) -> !quake.measure
+  // ```
+  // with:
+  // ```
+  //   %1 = ... : !quake.veq<4>
+  //   %measOut = quake.mz %1 : (!quake.veq<4>) -> !cc.stdvec<!quake.measure>
+  // ```
+  // And collect output names information:  `"[[[0,[1,"q0"]],[1,[2,"q1"]]]]"`
```
schweitzpgi, line 113: nice quotes! :)

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -623,6 +623,34 @@ struct SToR1 : public OpRewritePattern<quake::SOp> {
   }
 };
 
+// quake.s<adj> target
+// ─────────────────────────────────
+// (quake.z * quake.s) target
```
schweitzpgi, line 628: Isn't this from Pradnya's PR? I changed this comment there. Might need to rebase?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -561,8 +561,8 @@ MlirModule synthesizeKernel(const std::string &name, MlirModule module,
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createExpandMeasurementsPass());
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-  pm.addNestedPass<func::FuncOp>(cudaq::opt::createLoopNormalize());
-  pm.addNestedPass<func::FuncOp>(cudaq::opt::createLoopUnroll());
+  pm.addPass(cudaq::opt::createLoopNormalize());
+  pm.addPass(cudaq::opt::createLoopUnroll());
```
schweitzpgi, line 565: Sorry, I wasn't precise enough. Let's leave these as Func level passes so we can utilize multithreading. There is no need to run them at the module level AFAICT.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -634,10 +634,46 @@ std::string getASM(const std::string &name, MlirModule module,
...
+
   PassManager pm(context);
   pm.addPass(cudaq::opt::createLambdaLiftingPass());
+  // Run most of the passes from hardware pipelines.
+
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
+  pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
+  pm.addPass(cudaq::opt::createLoopUnroll());
+  pm.addPass(createCanonicalizerPass());
```
schweitzpgi, line 650: Make these two Func level so as to get better locality and performance here.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -634,10 +634,46 @@ std::string getASM(const std::string &name, MlirModule module,
...
+  pm.addNestedPass<func::FuncOp>(cudaq::opt::createLiftArrayAlloc());
+  pm.addPass(cudaq::opt::createGlobalizeArrayValues());
+  pm.addPass(cudaq::opt::createStatePreparation());
+  pm.addNestedPass<func::FuncOp>(cudaq::opt::createGetConcreteMatrix());
+  pm.addPass(cudaq::opt::createUnitarySynthesis());
+  pm.addPass(cudaq::opt::createApplyOpSpecializationPass());
+  cudaq::opt::addAggressiveEarlyInlining(pm);
+  pm.addPass(createSymbolDCEPass());
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
```
schweitzpgi, line 660: These 2 as well?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -634,10 +634,46 @@ std::string getASM(const std::string &name, MlirModule module,
...
+  pm.addNestedPass<func::FuncOp>(
+      cudaq::opt::createMultiControlDecompositionPass());
+  pm.addPass(cudaq::opt::createDecompositionPass(
+      {.enabledPatterns = {"SToR1", "TToR1", "R1ToU3", "U3ToRotations",
+                           "CHToCX", "CCZToCX", "CRzToCX", "CRyToCX", "CRxToCX",
+                           "CR1ToCX", "CCZToCX", "RxAdjToRx", "RyAdjToRy",
+                           "RzAdjToRz"}}));
+  pm.addPass(createCanonicalizerPass());
```
schweitzpgi, line 668: Yet another?

lib/Optimizer/Transforms/CombineMeasurements.cpp
annagrin: :)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: The loop unroll complains if I make it a func pass: `error: can't run 'builtin.module' pass manager on 'func.func' op`

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
```diff
@@ -23,13 +23,12 @@ using namespace cudaq;
...
 /// Translates operation names into OpenQASM gate names
 static LogicalResult translateOperatorName(quake::OperatorInterface optor,
                                            StringRef &name) {
   StringRef qkeName = optor->getName().stripDialect();
   if (optor.getControls().size() == 0) {
-    name = StringSwitch<StringRef>(qkeName).Case("r1", "u1").Default(qkeName);
+    name = StringSwitch<StringRef>(qkeName).Default(qkeName);
```
boschmitt, line 31: You don't need a `StringSwitch` if you are only using the default case.
```suggestion
    name = qkeName;
```

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
khalatepradnya: This was removed at one point in PR#2401 and restored later. So, aligning with the latest mainline should resolve it.

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
khalatepradnya, line 31: This was also restored in PR#2401. Making a note here to confirm that this change is not required.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
khalatepradnya: This decomposition pattern was removed since the existing `SToR1` was sufficient and this added additional gates in basis conversion. Again, probably will go away with rebase, making a note here to confirm that this change is not required.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4032,4 +4032,5 @@ def compile_to_mlir(astModule, metadata,
     if len(bridge.dependentCaptureVars):
         extraMetaData['dependent_captures'] = bridge.dependentCaptureVars
 
+    bridge.module.dump()
```
khalatepradnya, line 4035: Q: Is this a debugging artifact?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -634,10 +634,46 @@ std::string getASM(const std::string &name, MlirModule module,
...
+  pm.addPass(cudaq::opt::createUnitarySynthesis());
+  pm.addPass(cudaq::opt::createApplyOpSpecializationPass());
+  cudaq::opt::addAggressiveEarlyInlining(pm);
+  pm.addPass(createSymbolDCEPass());
+  pm.addPass(createCanonicalizerPass());
+  pm.addPass(createCSEPass());
+  pm.addNestedPass<func::FuncOp>(
+      cudaq::opt::createMultiControlDecompositionPass());
+  pm.addPass(cudaq::opt::createDecompositionPass(
+      {.enabledPatterns = {"SToR1", "TToR1", "R1ToU3", "U3ToRotations",
```
khalatepradnya, line 664: `SToR1` and `TToR1` may not be required on all OpenQASM2.0 targets, only those which don't support `sdg` and `tdg` out of the box. Similarly, we should remove `R1toU3` and `U3ToRotations` from generic workflow.

python/tests/backends/test_braket.py
```diff
@@ -289,7 +289,51 @@ def kernel(qubit_count: int):
...
+        qreg = cudaq.qvector(4)
+        x(qreg[1])
+        x(qreg[2])
+        v = qreg[0:2]
+        v = qreg[2:3]
+        mz(v)
+
+    counts = cudaq.sample(kernel, shots_count=100)
+    assert len(counts) == 1
+    assert "011" in counts
```
khalatepradnya, line 319: Q: I don't understand what this test is checking since `v` was overwritten to be of size 2...?

python/tests/backends/test_braket.py
```diff
@@ -289,7 +289,51 @@ def kernel(qubit_count: int):
...
+
+def test_kernel_qubit_subveq():
+
+    @cudaq.kernel
+    def kernel():
+        qreg = cudaq.qvector(4)
+        x(qreg[1])
+        x(qreg[2])
+        v = qreg[0:2]
+        v = qreg[2]
```
khalatepradnya, line 329: Same a above.

python/tests/backends/test_braket.py
```diff
@@ -299,10 +343,24 @@ def kernel():
...
         mz(qubits[1])
 
-    with pytest.raises(RuntimeError) as e:
-        cudaq.sample(kernel, shots_count=100).dump()
-    assert "cannot declare bit register. Only 1 bit register(s) is/are supported" in repr(
-        e)
+    counts = cudaq.sample(kernel, shots_count=100)
+    assert len(counts) == 2
+    assert "00" in counts
+    assert "10" in counts
```
khalatepradnya, line 349: 👍🏽

runtime/cudaq/platform/default/rest/helpers/braket/BraketExecutor.cpp
```diff
@@ -240,14 +257,20 @@ BraketExecutor::execute(std::vector<KernelExecution> &codesToExecute) {
           }
           auto resultsJson = nlohmann::json::parse(
               s3Response.GetResultWithOwnership().GetBody());
+
           auto c = serverHelper->processResults(resultsJson, taskArn);
 
           for (auto &regName : c.register_names()) {
+            std::cout << "Register name: " << regName << std::endl;
```
khalatepradnya, line 264: Q: Is this expected to be printed to console? I'm not sure if the user will see it with remote job sunbmission...

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
schweitzpgi: OK, interesting. I'll take a closer look at it.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
schweitzpgi: There is no reason canonicalize can't be run with a nested pass. Something strange is going on.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
schweitzpgi: There's a bug in loop-unroll that prevents the pass from being run at a FuncOp level. I'll create a PR to fix it.

schweitzpgi: See #2422

python/tests/backends/test_braket.py
bmhowe23: I agree w/ Pradnya. As written, I would expect this to be `assert "1" in counts` because I would expect it to be showing the value of `qreg[2]`.

runtime/cudaq/platform/default/rest/helpers/braket/BraketExecutor.cpp
```diff
@@ -20,6 +20,8 @@
 
 #include <aws/core/utils/ARN.h>
 
+#include <iostream>
+
```
bmhowe23, line 24: If the answer to https://github.com/NVIDIA/cuda-quantum/pull/2416/files#r1859223760 is that it is leftover debug code, then perhaps this new `#include` is no longer necessary?

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
```diff
@@ -136,7 +146,70 @@ sample_result BraketServerHelper::processResults(ServerMessage &resultsJson,
...
+  //              {    1111:(62+12)     x1110:(12+12)   }
+  //              {    1111:74           1110:24        }
+  // -----------------------------------------------------
+  // clang-format on
+
+  std::vector<ExecutionResult> execResults;
+
+  // Get a reduced list of qubit numbers that were in the original program
+  // so that we can slice the output data and extract the bits that the user
+  // was interested in. Sort by QIR qubit number.
```
bmhowe23, line 175: ?
```suggestion
  // was interested in. Sort by OpenQASM qubit number.
```

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: would it be possible to submit it after this PR to speed things up?

runtime/cudaq/platform/default/rest/helpers/braket/braket.yml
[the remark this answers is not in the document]
bmhowe23: OK, that makes sense. Should lines 23 and 24 be removed then?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: we need to rethink python's cudaq::translate implementation so it can use the target's pipeline instead. Will leave to the later time.

python/tests/backends/test_braket.py
annagrin: thanks, it is missing a meazurement!

python/tests/backends/test_braket.py
annagrin: :)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: it can, I meant the loop unroll. I moved everything that can run as a func pass to func level.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: This can wait but if I understood correctly, these changes to the `getASM` function do not affect the `braket` target, right?
And this function will be invoked only when users use the API `cudaq.translate(<kernel>, format="openqasm2")`, correct?
If yes to both, only the generic passes should be enabled here for translation to OpenQASM2.0, and `"SToR1", "TToR1", "R1ToU3", "U3ToRotations"` should be removed from here.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
schweitzpgi: If this PR is merged first, please open an issue for this discussion so someone can follow up on the technical debt that isn't addressed here.

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
annagrin: done, thanks!

runtime/cudaq/platform/default/rest/helpers/braket/BraketExecutor.cpp
annagrin: removed, thanks!

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
annagrin: done!

lib/Optimizer/Transforms/DecompositionPatterns.cpp
annagrin: updated

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: nevermind:)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: done

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Updated - I made all the passes I could func passes, thanks!

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
annagrin: done

python/cudaq/kernel/ast_bridge.py
annagrin: yes, thanks, removed

lib/Optimizer/CodeGen/TranslateToOpenQASM.cpp
annagrin: restored the original version

python/tests/backends/test_braket.py
annagrin: updated

python/tests/backends/test_braket.py
annagrin: done

runtime/cudaq/platform/default/rest/helpers/braket/BraketExecutor.cpp
annagrin: leftover printing, removed!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: will update after the #2422 lands

runtime/cudaq/platform/default/rest/helpers/braket/braket.yml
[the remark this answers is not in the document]
annagrin: done!

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
```diff
@@ -66,13 +66,13 @@ void BraketServerHelper::initialize(BackendConfig config) {
...
-    backendConfig = std::move(config);
-    return;
-  }
+  // const auto emulate_it = config.find("emulate");
+  //  if (emulate_it != config.end() && emulate_it->second == "true") {
+  //    cudaq::info("Emulation is enabled, ignore all Amazon Braket connection "
+  //                "specific information.");
+  //    backendConfig = std::move(config);
+  //    return;
+  //  }
```
bmhowe23, line 75: Was this intentional? Why do these need to be commented out?

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
```diff
@@ -136,7 +146,70 @@ sample_result BraketServerHelper::processResults(ServerMessage &resultsJson,
...
+  if (qubitNumbers.empty()) {
+    execResults.emplace_back(ExecutionResult{fullSampleResults.to_map()});
+  } else {
+    auto subset = fullSampleResults.get_marginal(qubitNumbers);
+    execResults.emplace_back(ExecutionResult{subset.to_map()});
+  }
+
+  // Return a sample result including the global register and all individual
+  // registers.
+  auto ret = cudaq::sample_result(execResults);
```
bmhowe23, line 194: If any `braket` backends have run the mapping pass, you may need changes similar to https://github.com/NVIDIA/cuda-quantum/pull/2328 right here. If we are unsure and under a time crunch, we could perhaps just create an issue and follow up on this later?

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
annagrin: that's me being tired:) will revert

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
bmhowe23: Do we need to open an issue for this comment thread?

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
annagrin: Created an issue to be addressed later: https://github.com/NVIDIA/cuda-quantum/issues/2425
Thanks for the info!

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -457,6 +470,17 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
         throw std::runtime_error("Could not successfully apply quake-synth.");
     }
 
+    // Delay combining measurements for backends that cannot handle
+    // subveqs and multiple measurements until we created the emulation code.
+    auto combineMeasurements =
+        passPipelineConfig.find("combine-measurements") != std::string::npos;
+    if (emulate && combineMeasurements) {
+      std::regex combine("(.*),([ ]*)combine-measurements(.*)");
```
bmhowe23, line 478: This is complicated and subtle enough that I think it merits a `cudaq::info` message when this is happening (even as part of this PR). Do we also plan to circle back to this? Ideally the emulation code path and true REST server API would not deviate behaviors in this way (at least not for the long term).

annagrin: done, thanks!

bmhowe23: nit: this will also not work if we ever have a future pass named `combine-measurements-with-a-random-suffix` or `this-is-a-random-prefix-to-combine-measurements`. If this is only temporary code, then it probably doesn't matter, but that's just more motivation to circle back to this at some point soon.

PR #2417, opened 2024-11-26
Follow-up to #2394 - handle remote REST QPU also

### LINKED ISSUE #2382: [Hardware path] `cudaq.observe()` incorrectly returns 0.0 when there is only a single spin op term
### Required prerequisites

### Describe the bug

When using a single spin-op term, running `cudaq::observe()` from C++ (or `cudaq.observe()` from Python) returns incorrect results when running with `--emulate` for C++ (or `emulate=True` for  Python). It returns `0.0` instead of a true expectation calculation.

Starting with the targettests/execution/cudaq_observe.cpp test already in the repo, if one were to modify `h` to a single term (e.g. `cudaq::spin_op h = i(0) * z(1);`), the resulting energy is 0.0 when running with `--emulate` and it is non-zero when running without `--emulate`.

For example, generate this test file (test.cpp):
```cpp
#include <cudaq.h>
#include <cudaq/algorithm.h>

// The example here shows a simple use case for the `cudaq::observe`
// function in computing expected values of provided spin_ops.

struct ansatz {
  auto operator()(double theta) __qpu__ {
    cudaq::qvector q(2);
    x(q[0]);
    ry(theta, q[1]);
    x<cudaq::ctrl>(q[1], q[0]);
  }
};

int main() {

  // Build up your spin op algebraically
  using namespace cudaq::spin;
  cudaq::spin_op h = i(0) * z(1); // <-- MAKE THIS CHANGE HERE
  // cudaq::spin_op h = 5.907 - 2.1433 * x(0) * x(1) - 2.1433 * y(0) * y(1) +
  //                    .21829 * z(0) - 6.125 * z(1);

  // Make repeatable for shots-based emulation
  cudaq::set_random_seed(13);

  // Observe takes the kernel, the spin_op, and the concrete
  // parameters for the kernel
  double energy = cudaq::observe(ansatz{}, h, .59);
  printf("Energy is %.16lf\n", energy);
  return 0;
}
```
And then run the following two tests, you will see different answers. The answers should be the same.
```bash
$ nvq++ --target oqc --emulate test.cpp
$ ./a.out
Energy is 0.0000000000000000
$ nvq++ --target qpp-cpu test.cpp
$ ./a.out
Energy is 0.8309406791001634
```

### Expected behavior

The answers should match.

Note that the example provided in this bug report only deals with the `--emulate` path, but based on partner feedback, I believe the bug exists when using real hardware as well.

Not a regression

### Environment

- **CUDA-Q version**: 0.8.0
- **Python version**: N/A
- **C++ compiler**: N/A
- **Operating system**: N/A

### Suggestions

I have tracked down at least one of the problems to this section of code: https://github.com/NVIDIA/cuda-quantum/blob/3300b122d5ea17c2f726358a45c37c23c6e1d88f/runtime/common/BaseRemoteRESTQPU.h#L692

That section of code is assuming that if there is a single circuit that was run (like for a single spin op), that it was not actually an observe calculation. This needs to be updated to be correctly ascertain whether the parent request originated from an observe call rather than just inferring it.

bmhowe23:
Resolves #2382


khalatepradnya (approved):
LGTM.

schweitzpgi (approved):

PR #2418, opened 2024-11-26
[docs] Support for `braket` target 

khalatepradnya:
Documentation for `braket` target


annagrin (approved):
LGTM, thanks Pradnya!

PR #2422, opened 2024-11-26
[core] Let loop unrolling be added as a nested func pass manager.

schweitzpgi:
Fixes some technical debt wrt loop unrolling to allow it to be used as it was intended. That is, as a nested pass on func::FuncOp ops. Tunes the loop unrolling pipeline to use strictly nested passes for better locality and performance.


annagrin (approved):
LGTM!

PR #2423, opened 2024-11-26
`Cudaq <> Superstaq` integration

vtomole:
Checklist:
- [x] I have added tests to cover my changes.
- [x] I have updated the documentation accordingly.

### Description
This PR allows Cudaq to talk to [Superstaq ](https://superstaq.readthedocs.io/en/latest/). With this PR, the following code should work.

```python
import cudaq

@cudaq.kernel
def kernel():
    qubit = cudaq.qubit()
    x(qubit)
    mz(qubit)

cudaq.set_target("infleqtion", target="cq_sqorpius_simulator")
result = cudaq.sample(kernel)
print(f"Counts from cudaq-superstaq submission: {result}")
# prints
# Counts from cudaq-superstaq submission: { 1:1000 }
```


utils/mock_qpu/infleqtion/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    print("Getting Job")
+    global countJobGetRequests, createdJobs, numQubitsRequired
+
+    # Simulate asynchronous execution
+    if countJobGetRequests < 3:
+        countJobGetRequests += 1
+        return {"status": "running"}
+
+    countJobGetRequests = 0
+    res = {'status': 'Done', "samples": {
```
vtomole, line 80: This is hardcoded because I don't know how to run OpenQASM in Cudaq. The other mock servers[ assumes that the input is QIR](https://github.com/bharat-thotakura/cuda-quantum/blob/4e5a5138916adcdb97f41bc4b0c19e71160e3476/utils/mock_qpu/braket/__init__.py#L83) so we can't learn from them.

khalatepradnya: @bmhowe23 : Can you help here?

bmhowe23: We cannot currently run OpenQASM in our CUDA-Q runtime. One could, in principle, write conversion logic to convert OpenQASM to QIR and then run the QIR through our CUDA-Q runtime.

targettests/infleqtion/bug_qubit.cpp
```diff
@@ -0,0 +1 @@
+../execution/bug_qubit.cpp
```
khalatepradnya, line 1: Accompanying change `// RUN: nvq++ %cpp_std --target infleqtion --emulate %s -o %t && %t | FileCheck %s` in the corresponding file needed.
(Note that `emulate` flag will simply run the QIR so this test may not be as valuable as expected)

runtime/cudaq/platform/default/rest/helpers/infleqtion/infleqtion.yml
```diff
@@ -0,0 +1,36 @@
...
+
+config:
+  # Tell DefaultQuantumPlatform what QPU subtype to use
+  platform-qpu: remote_rest
+  # Add the rest-qpu library to the link list
+  link-libs: ["-lcudaq-rest-qpu"]
+  # Tell NVQ++ to generate glue code to set the target backend name
+  gen-target-backend: true
+  # Define the lowering pipeline; # ADJUST DECOMPOSTION
+  platform-lowering-config: "func.func(const-prop-complex,canonicalize,cse,lift-array-alloc),globalize-array-values,state-prep,unitary-synthesis,canonicalize,apply-op-specialization,aggressive-early-inlining,unrolling-pipeline,func.func(lower-to-cfg),canonicalize,func.func(multicontrol-decomposition),decomposition{enable-patterns=SToR1,TToR1,R1ToU3,U3ToRotations,CHToCX,CCZToCX,CRzToCX,CRyToCX,CRxToCX,CR1ToCX,RxAdjToRx,RyAdjToRy,RzAdjToRz},func.func(memtoreg{quantum=0}),symbol-dce"
```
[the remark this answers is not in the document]
vtomole: I don't think we need a mapping pass. We handle all the mapping server-side.

utils/mock_qpu/infleqtion/__init__.py
vtomole: How is this issue handled in braket? I see that it takes qasm as well https://github.com/NVIDIA/cuda-quantum/blob/f433d74e8e23a65fe5199062b7fdd7527c218740/runtime/cudaq/platform/default/rest/helpers/braket/braket.yml#L22

targettests/infleqtion/bug_qubit.cpp
vtomole: Say more? `// RUN: nvq++ %cpp_std --target infleqtion --emulate %s -o %t && %t | FileCheck` dosn't exist in other `bug_qubit.cpp` symlink files.

utils/mock_qpu/infleqtion/__init__.py
khalatepradnya: The [Python tests here](https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/backends/test_braket.py#L18) are run against Amazon Braket server, not the mock server. (Hence skipped by the CI.) Any test(s) that uses `emulate` flag run QIR.

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
```diff
@@ -0,0 +1,315 @@
...
+  if (circuitCodes.size() != 1) {
+    throw std::runtime_error("Only single circuit submission is supported.");
+  }
+
+  auto &circuitCode = circuitCodes[0];
+
+  // Construct the job message
+  ServerMessage job;
+  job["qasm_strs"] =
+      formatOpenQasm(circuitCode.code); // Assuming code is in OpenQASM format
```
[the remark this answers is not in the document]
khalatepradnya: It would be nice to clarify in the comment on line#185.

targettests/infleqtion/bug_qubit.cpp
khalatepradnya: ah, I should have been clearer. I mean the file that this links to - https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/execution/bug_qubit.cpp

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
```diff
@@ -0,0 +1,315 @@
...
+  // Move the passed config into the member variable backendConfig
+  backendConfig = config;
+
+  // Set default URL and version if not provided
+  backendConfig["url"] = getValueOrDefault(config, "url", DEFAULT_URL);
+  backendConfig["version"] =
+      getValueOrDefault(config, "version", DEFAULT_VERSION);
+  backendConfig["user_agent"] = getValueOrDefault(
+      config, "user_agent", "cudaq/" + std::string(cudaq::getVersion()));
+  backendConfig["target"] =
```
khalatepradnya, line 121: Q on using `target` keyword: I wonder if it would be cleaner to use a field / key like `qpu`, `machine`, or `device` so as to avoid confusion with CUDA-Q `target`?
Currently Python code would look like - 
```
cudaq.set_target("infleqtion", target="cq_sqorpius_simulator")
```

runtime/cudaq/platform/default/rest/helpers/infleqtion/infleqtion.yml
```diff
@@ -0,0 +1,36 @@
...
+  # Tell the rest-qpu that we are generating OpenQASM 2.0.
+  codegen-emission: qasm2
+  # Library mode is only for simulators, physical backends must turn this off
+  library-mode: false
+
+target-arguments:
+  - key: qpu
```
khalatepradnya, line 27: I see this defines `qpu` field, but it's not being used in the `InfleqtionServerHelper` class...

utils/mock_qpu/infleqtion/__init__.py
vtomole: We'll do what braket does. Thanks.

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include <sstream>
+#include <thread>
+using json = nlohmann::json;
+
+std::string formatOpenQasm(const std::string &input_string) {
+  std::string escaped_string = input_string;
+  // Escape backslashes, double quotes, and newlines
+  std::string::size_type pos = 0;
+  while ((pos = escaped_string.find("\\", pos)) != std::string::npos) {
+    escaped_string.replace(pos, 1, "\\\\");
```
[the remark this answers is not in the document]
bharat-thotakura: Looks like it probably can, pushed an update using it

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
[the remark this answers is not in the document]
bharat-thotakura: Added

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
```diff
@@ -0,0 +1,315 @@
...
+  job["shots"] = shots;
+
+  if (backendConfig.count("options"))
+    job["options"] = backendConfig.at("options");
+
+  if (backendConfig.count("method"))
+    job["method"] = backendConfig.at("method");
+
+  // Store output names and reorder indices if necessary
+  OutputNamesType outputNamesMap;
```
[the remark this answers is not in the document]
bharat-thotakura: Could you clarify your question? That is, the `infleqtion` target is able to handle both single measurements and multi-qubit measurements in the sense of a user calling:
```python
mz(qvector)
# or
mz(qvector[0])
mz(qvector[1])
...
```
We should also be able to handle multiple classical registers, for example, something like:
```python
qubits = cudaq.qvector(4)
ancillas = cudaq.qvector(2)
```
however, I think the discrepancy might arise when measuring both registers and retrieving results from our API. Our QPU/Simulator will return counts combining the classical registers, whereas, I think `cudaq.sample()` will return separate counts dictionaries for each vector allocation

utils/mock_qpu/infleqtion/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    print("Getting Job")
+    global countJobGetRequests, createdJobs, numQubitsRequired
+
+    # Simulate asynchronous execution
+    if countJobGetRequests < 3:
+        countJobGetRequests += 1
+        return {"status": "running"}
+
+    countJobGetRequests = 0
+    res = {
```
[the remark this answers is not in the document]
bharat-thotakura: Our API actually accepts single circuits or a list (batch) of circuits in one-go. I wasn't sure how multi-circuit sampling/submission is done in CUDA-Q, so I threw in that check for the initial implementation as a debug validation. I wasn't aware of its impact on `observe`, so I've removed it now

runtime/cudaq/platform/default/rest/helpers/infleqtion/InfleqtionServerHelper.cpp
bharat-thotakura: The motivation for using `target=` was that it mirrored the format expected by our API. But I did want to consult on the naming convention to best maintain parity with the other CUDA-Q hardware implementations, and I think going with `machine` seems to be the most appropriate for us -- updated now.

runtime/cudaq/platform/default/rest/helpers/infleqtion/infleqtion.yml
bharat-thotakura: Related to: https://github.com/NVIDIA/cuda-quantum/pull/2423#discussion_r1870494667. It was kept as a placeholder/template till we settled on the naming convention to use -- updated now

utils/mock_qpu/infleqtion/__init__.py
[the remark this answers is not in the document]
khalatepradnya: I see. Now the 'observe' test(s) are failing because of the hardcoded values in mock server.
You may consider having two sets of tests - `checkSampleSync` - this would run on your test server and will be skipped in CI, and `checkSampleSyncEmulate` - this will use the QIR path. I am okay to have only the former as part of this PR and we can revisit automated testing in as follow-up.

python/tests/backends/test_Infleqtion.py
```diff
@@ -0,0 +1,63 @@
...
+    y(qubit)
+    z(qubit)
+    t(qubit)
+    s(qubit)
+    mz(qubit)
+
+
+def test_infleqtion_sample():
+    counts = cudaq.sample(kernel)
+    assert len(counts) == 2
```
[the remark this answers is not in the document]
khalatepradnya: Fixed by testing against Infleqtion server.

utils/mock_qpu/infleqtion/__init__.py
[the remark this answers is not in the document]
khalatepradnya: Removed tests for `observe` here, can be addressed in follow-up PR.

targettests/infleqtion/bug_qubit.cpp
khalatepradnya: Fixed in 511d502f12cf6e8cb134ac94a9066b285acf8c61

bmhowe23 (commented):
Overall the changes look good to me. Leaving it to @bettinaheim to confirm she is ok with the path forward for nightly testing.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -480,6 +480,40 @@ struct R1ToU3 : public OpRewritePattern<quake::R1Op> {
...
+// quake.r1(-θ) target
+struct R1AdjToR1 : public OpRewritePattern<quake::R1Op> {
+  using OpRewritePattern<quake::R1Op>::OpRewritePattern;
+
+  void initialize() { setDebugName("R1AdjToR1"); }
+
+  LogicalResult matchAndRewrite(quake::R1Op op,
+                                PatternRewriter &rewriter) const override {
+    if (!op.getControls().empty())
+      return failure();
```
bettinaheim, line 494: nit: seems unnecessary to not allow controls. The same is perfectly valid with controls.

bettinaheim (approved):

vtomole:
Thank you so much for the reviews @khalatepradnya , @bmhowe23 , @bettinaheim !!

PR #2431, opened 2024-11-27
Improve performance of some opt passes

annagrin:
### Improve performance of some of the compiler passes 

This change runs some passes on `FuncOp` instead of `ModuleOp` so the optimizations can run in parallel on multiple functions.

- Use `cc-loop-unroll` and `cc-loop-normalize` as func passes instead of module passes
- Make `state-prep` a func pass as well
- Update hardware pipelines
- Update tests

Related: 
https://github.com/NVIDIA/cuda-quantum/pull/2422
https://github.com/NVIDIA/cuda-quantum/pull/2416


runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -955,8 +955,8 @@ jitCode(ImplicitLocOpBuilder &builder, ExecutionEngine *jit,
     pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
     pm.addPass(createCanonicalizerPass());
```
annagrin, line 956: missed two canonicalizer passes here

annagrin: done

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -325,90 +316,81 @@ class StateDecomposer {
...
+
+          // Prepare state from vector data.
+          auto gateBuilder = StateGateBuilder(rewriter, loc, qubits);
+          auto decomposer = StateDecomposer(gateBuilder, vec, phaseThreshold);
+          decomposer.decompose();
+
+          // Use prepared qubits instead of the init state.
+          init.replaceAllUsesWith(qubits);
+          rewriter.eraseOp(init);
+          if (cast)
```
annagrin: only erasing the init op seems to be necessary (so we don't go into an infinite recursion)

annagrin: done

targettests/execution/state_preparation.cpp
```diff
@@ -10,6 +10,7 @@
 // RUN: nvq++ %cpp_std --enable-mlir  %s                             -o %t && %t | FileCheck %s
 
 // Quantum emulators
+// RUN: nvq++ %cpp_std --target braket                   --emulate %s -o %t && %t | FileCheck %s
```
schweitzpgi, line 13: This will need the `if $braket_avail;` wrapper when Bettina is ready to merge her PR. Can you flag that in her PR? Thanks.

schweitzpgi (approved):

targettests/execution/state_preparation.cpp
annagrin: this one is running with the `--emulate` flag, do we still need the `if $braket_avail` test? It runs on my machine without any credentials.

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -325,90 +316,80 @@ class StateDecomposer {
...
+  using OpRewritePattern::OpRewritePattern;
+
+  explicit StatePrepPattern(MLIRContext *ctx, double phaseThreshold)
+      : OpRewritePattern(ctx), phaseThreshold(phaseThreshold) {}
+
+  LogicalResult matchAndRewrite(quake::InitializeStateOp init,
+                                PatternRewriter &rewriter) const override {
+    auto loc = init.getLoc();
+    auto qubits = init.getOperand(0);
+    if (auto alloc = dyn_cast<quake::AllocaOp>(qubits.getDefiningOp())) {
```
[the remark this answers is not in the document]
annagrin: updated, thanks!

targettests/execution/state_preparation.cpp
annagrin: looked ad Bettina's PR, seems like it is still needed. Will flag.

PR #2433, opened 2024-11-28
Implement a `FIXME` in GitLab pipeline

1tnguyen:
### Description

This depends on https://github.com/NVIDIA/cuda-quantum/pull/2424.

**Note**: the "update the mgpu build to properly work with alternative libs" FIXME is included in the mgpu sha in that PR2424.


schweitzpgi (approved):
Nice code deletion. :)

PR #2435, opened 2024-11-28
[nvq++] Removes MLIR's `scf` dialect

boschmitt:
## PR chain

* **\>\>\>** #2435  **<<<**
* #2436
* #2438

### Description

We don't use this dialect and its presence is a historical artifact. This change triggered the removal of two tests:

* `test/Quake/ghz.qke`
* `test/Quake/iqft.qke`

Both tests are a reminder of a past when we had to write quantum kernels directly in MLIR because of a lack of frontend. Both no longer test aything useful.

The commit modifies `test/Quake/canonical-2.qke`, which was only testing the canonicalization of `cc.scope` operations. The new form is removes the clutter, making the test more precise.

`test/Translate/ghz.qke` had to be modified because it uses MLIR's `affined.for` and its conversion to LLVMDialect requires `scf.for`.

_Note: I'm unsure what `test/Translate/ghz.qke` is actually testing, I rewrote the test using `cc.loop`, but might have just deleted as well._ (@schweitzpgi You eyes on this would be greatly appreciated, the diff here is mostly variable names)


schweitzpgi (approved):
Looks good to me.

Tests that are being removed are historical and look to mostly (only?) be testing canonicalization.

PR #2440, opened 2024-11-28
[cmake] Remove unnecessary install commands.

boschmitt:
### Description
In the case of the libraries, there is no need to install them twice. In most cases, I opted to remove the install command that contains `EXPORT` because we don't use/install these export files, and so far they don't seem to be missed. (Even if we want to install these exports, the current way of doing this seems overly complicated, we can generate a single runtime export with everything that was built---something to be done on a later commit.)

In the case of headers files: their installation is already covered by the installation of the `cudaq` directory and all `*.h` within.


amccaskey (approved):
nice cleanup

PR #2441, opened 2024-11-28
[cmake] Fix double installation of nvq++ libs.

boschmitt:
### Description
This PR disables MLIR's installation of our optimizer's libs. We do it ourselves. Fixing this problem is particularly relevant our python wheels: The issue is aggravated there because the manylinux image has an enviroment that favors libraries to be installed in `lib64` prefix. MLIR's build system honors that; our's don't, and thus we ship wheels that have these libs installed twice, one in `lib` and another in `lib64`.

The issue that these libs don't need to be part of the wheels will be addressed in a later commit.


schweitzpgi:
Do these have to be installed in `lib/`? It sounds like we're not installing them to the correct directory from the description.

boschmitt:
> Do these have to be installed in `lib/`? It sounds like we're not installing them to the correct directory from the description.

Currently yes. The `lib/` path is hardcoded in various places. If that were not the case, then we could honor the system's preference. (I run into this similar issue in #1840)

schweitzpgi (approved):

PR #2443, opened 2024-11-29
[python] Remove unused MLIR components

boschmitt:
### Description
We don't need to take everything from MLIR for our python bindings. This change cherry picks the upstream components our compiler depends on.

The commit also cleans up some unnecessary code that ends up registering dialects more than once.


python/runtime/mlir/py_register_dialects.cpp
```diff
@@ -28,32 +22,10 @@ using namespace mlir::python::adaptors;
 using namespace mlir;
 
 namespace cudaq {
-static bool registered = false;
 
-void registerQuakeDialectAndTypes(py::module &m) {
+void registerQuakeTypes(py::module &m) {
```
schweitzpgi, line 26:
```suggestion
static void registerQuakeTypes(py::module &m) {
```

python/runtime/mlir/py_register_dialects.cpp
```diff
@@ -143,21 +115,10 @@ void registerQuakeDialectAndTypes(py::module &m) {
       });
 }
 
-void registerCCDialectAndTypes(py::module &m) {
+void registerCCTypes(py::module &m) {
```
schweitzpgi, line 118:
```suggestion
static void registerCCTypes(py::module &m) {
```
Assuming that's ok, should move these 2 out of the namespace as well.

python/runtime/mlir/py_register_dialects.cpp
```diff
@@ -310,14 +268,17 @@ void bindRegisterDialects(py::module &mod) {
...
   mod.def("register_all_dialects", [](MlirContext context) {
     DialectRegistry registry;
-    registry.insert<quake::QuakeDialect, cudaq::cc::CCDialect>();
-    cudaq::opt::registerCodeGenDialect(registry);
-    registerAllDialects(registry);
-    auto *mlirContext = unwrap(context);
+    cudaq::registerAllDialects(registry);
+    MLIRContext *mlirContext = unwrap(context);
     mlirContext->appendDialectRegistry(registry);
     mlirContext->loadAllAvailableDialects();
```
schweitzpgi, line 274: Does loadAllAvailableDialects actually load Quake, CC, and CodeGen?

Do we need an explicit test to verify that all the dialects we expect to be loaded are loaded?

python/runtime/mlir/py_register_dialects.cpp
boschmitt: Yeah, it makes sense to remove the namespace entirely in this `.cpp` file.

python/runtime/mlir/py_register_dialects.cpp
schweitzpgi: The codegen dialect was intended to be sort of second-class in that it would only be applicable at codegen time. The point being we don't want folks to start using it in ad hoc ways in random passes, etc. Adding a few landmines was sort of the idea.

OTOH, we could clean it up a bit and remove some of those hurdles. We still don't want that dialect used "in the wild" though...

python/runtime/mlir/py_register_dialects.cpp
```diff
@@ -292,10 +259,9 @@ void registerCCDialectAndTypes(py::module &m) {
           });
 }
 
-void bindRegisterDialects(py::module &mod) {
-  registerQuakeDialectAndTypes(mod);
-  registerCCDialectAndTypes(mod);
+static bool registered = false;
```
schweitzpgi, line 262: This looks like it was pre-existing, but I'm not a fan of a global like this. Is there no good way to query whether things have been registered already by asking the dictionary of registered stuff?

boschmitt: Yes, there are actually two options: (1) makes so the dialects get registered upon Python's module loading or (2) make MLIR python automagically register things.

Option (2) requires a slightly different sort of change that doesn't fit well with the rest of this PR. I had (2) implemented before, but I reverted the change because the CI is failing and I can't reproduce the issue locally, even when using the same docker image :( 

After various experiments of changing the code, adding global state back, I still can't make it pass CI. (So this is the state of my last try, I will update the branch and revisit the change.)

PR #2446, opened 2024-12-03
[Tensornet backends] Make controlled rank a configurable setting

1tnguyen:
### Description

- For small controlled ops, i.e., singly controlled, expand the gate matrix and use `cutensornetStateApplyTensorOperator`.

- Add `CUDAQ_TENSORNET_CONTROLLED_RANK` threshold to determine when `cutensornetStateApplyControlledTensorOperator` should be used. For MPS, this is fixed at 1 as it cannot handle gate ops with more than 2 qubits.

- Add doc for the new setting and also remove some stale notes about random seeds (fixed in https://github.com/NVIDIA/cuda-quantum/pull/2398) in the docs.


runtime/nvqir/cutensornet/simulator_cutensornet.cpp
```diff
@@ -90,17 +90,45 @@ void SimulatorTensorNetBase::applyGate(const GateApplicationTask &task) {
     }
     return paramsSs.str() + "__" + std::to_string(vecComplexHash(task.matrix));
   }();
-  const auto iter = m_gateDeviceMemCache.find(gateKey);
 
-  // This is the first time we see this gate, allocate device mem and cache it.
-  if (iter == m_gateDeviceMemCache.end()) {
-    void *dMem = allocateGateMatrix(task.matrix);
-    m_gateDeviceMemCache[gateKey] = dMem;
+  if (controls.size() <= m_maxControlledRankForFullTensorExpansion) {
```
khalatepradnya, line 94: Can we add a test to cover this scenario?

khalatepradnya (commented):
Changes look good to me. Thanks, Thien!

runtime/nvqir/cutensornet/simulator_cutensornet.cpp
1tnguyen: Somehow, I missed the GitHub notification for this comment!
Our backend test suite should cover this case. This is the primary use case (e.g., CNOT/CZ gates) that we want to speed up using the full matrix representation of the controlled gate and these gates are commonly used in many tests.

PR #2447, opened 2024-12-03
Allow setting noise on measurement operation and apply it accordingly

1tnguyen:
### Description

- Add "mz" to the list of allowed ops (for validation purposes) in `NoiseModel`.

- Query noise channels for `mz` and apply them accordingly similar to the way we retrieve and apply gate noises.

- Add a unit test.


bmhowe23 (approved):
This LGTM. Thanks, @1tnguyen!

PR #2448, opened 2024-12-04
[docs] [testing] Follow-up for `braket` target and minor clean-up

khalatepradnya:
* Update docs to show the `braket` examples
* Remove the unsupported tests (tracked in issue #2325)
* Emacs compatible banner


unittests/backends/braket/BraketTester.cpp
```diff
@@ -164,6 +160,7 @@ CUDAQ_TEST(BraketTester, checkObserveSync) {
   kernel.x(qubit[0]);
   kernel.ry(theta, qubit[1]);
   kernel.x<cudaq::ctrl>(qubit[1], qubit[0]);
+  kernel.mz(qubit);
```
bmhowe23, line 163: Perhaps ok to save this discussion for later, but I'm surprised this works because I thought that from a language perspective we prohibited measurements in `observe` kernels?

khalatepradnya: You are right, and it doesn't work. Updated message on line#149. There's an [existing test here](https://github.com/NVIDIA/cuda-quantum/blob/main/unittests/backends/braket/BraketTester.cpp#L206) which gives error because the `braket` target requires qubits to be measured. The two tests together are to showcase the current behavior of `observe` on this target (and contradicting requirements). We have similar tests in Python. 
I removed the other C++ observe tests because they fail with the same message.

bmhowe23 (approved):
LGTM

PR #2450, opened 2024-12-04
[testing] Skip Python tests if dependency package missing

khalatepradnya:
Skip the 'test_evolve_dynamics_torch_integrators.py' tests if `torch` package not found.


1tnguyen (approved):
Thanks, @khalatepradnya!

PR #2456, opened 2024-12-09
Add number of trajectories to `observe_options` struct

1tnguyen:
### Description

For trajectory simulation, when computing the expectation value, we have the option to specify the number of trajectories to average the expectation value results. Hence, adding it as part of the existing `observe_options` struct.

**Note**: as the noise model is a compulsory member of `observe_options`, we don't need to validate that the number of trajectories is only provided for noisy simulation.


runtime/common/ExecutionContext.h
```diff
@@ -109,6 +109,10 @@ class ExecutionContext {
   /// `sample_result`.
   std::vector<char> invocationResultBuffer;
 
+  /// @brief The number of trajectories to be used for expectation calculation
```
schweitzpgi, line 112:
```suggestion
  /// @brief The number of trajectories to be used for an expectation calculation
```

runtime/cudaq/algorithms/observe.h
```diff
@@ -60,9 +60,14 @@ concept ObserveCallValid =
 /// \p shots is the number of shots to run for the given kernel. The default of
 /// -1 means direct calculations for simulation backends. \p noise is the noise
 /// model to use for the observe operation.
+/// \p num_trajectories is the optional number of trajectories to be used when
+/// computing the expectation values in the presence of noise. This parameter is
+/// only applied to simulation backends that support trajectory noisy
+/// simulation.
```
schweitzpgi, line 66:
```suggestion
/// only applied to simulation backends that support noisy
/// simulation of trajectories.
```
This sounds better to my ear, but I admit the original may be expected in the domain.

runtime/cudaq/algorithms/observe.h
```diff
@@ -75,14 +80,17 @@ std::optional<observe_result>
...
+               std::size_t batchIteration = 0, std::size_t totalBatchIters = 0,
+               std::optional<std::size_t> numTrajectories = {}) {
   auto ctx = std::make_unique<ExecutionContext>("observe", shots);
   ctx->kernelName = kernelName;
   ctx->spin = &h;
   if (shots > 0)
     ctx->shots = shots;
 
+  if (numTrajectories.has_value())
+    ctx->numberTrajectories = numTrajectories.value();
```
schweitzpgi, line 92:
```suggestion
    ctx->numberTrajectories = *numTrajectories;
```
Should be safe (checked in previous line) and foregoes a second check and pessimistically thinking an exception could be thrown.

schweitzpgi (approved):
Nice! Thanks, Thien.

PR #2458, opened 2024-12-09
Make sure uccsd from cudaqx compiles for quantum devices

annagrin:
### Description
Fix a number of issues with compiler passes discovered in `uccsd` compilation for quantum devices.

####Details
- Fix crashes in LiftArrayAlloc
- Add const prop for `if(%true)` and `if(%false)` statements
- Unroll loops twice in the unrolling pipeline, with the LiftArrayAlloc pass in between
  - helps const prop loop boundaries stored in arrays
- Use `solvers.stateprep.uccsd` code 
- Add tests


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -2038,6 +2038,63 @@ MutableOperandRange cudaq::cc::ConditionOp::getMutableSuccessorOperands(
...
+    auto cond = ifOp.getCondition();
+    if (auto constCond = cond.getDefiningOp<arith::ConstantOp>()) {
+      if (auto attr = dyn_cast<BoolAttr>(constCond.getValue())) {
+        auto value = attr.getValue();
+        auto block =
+            value ? ifOp.getThenEntryBlock() : ifOp.getElseEntryBlock();
+
+        // Prevent quantum allocations from moving to an outer scope.
+        if (block && hasQuantumAllocations(*block))
+          return failure();
```
schweitzpgi, line 2074: Was there an example where this happened? If so, the IR has bugs.

*Any* allocations must be either in a basic block owned by the func::FuncOp, cc::LambdaCreateOp, or a cc::ScopeOp. So this check isn't needed for valid IR.

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -2038,6 +2038,63 @@ MutableOperandRange cudaq::cc::ConditionOp::getMutableSuccessorOperands(
...
+
+        if (block) {
+          if (auto cont = dyn_cast<cudaq::cc::ContinueOp>(block->back())) {
+            for (std::size_t i = 0; auto opnd : ifOp.getResults())
+              rewriter.replaceAllUsesWith(opnd, cont.getOperand(i++));
+            rewriter.eraseOp(cont);
```
schweitzpgi, line 2080: There is code in LowerToCFG.cpp that folds away `cc::IfOp`s. I think you can use the `RewriteIf` pattern directly from there (share it in an `.inc` file). Basically, we would use that pattern on any `IfOp` with a constant argument.  Using the pattern will get the linear arguments propagated correctly. Also, since the argument is a constant, DCE will simply clean things up and fuse the blocks, etc.

lib/Optimizer/Transforms/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ add_cudaq_library(OptTransforms
   ApplyOpSpecialization.cpp
   ArgumentSynthesis.cpp
   BasisConversion.cpp
+  CollapseStores.cpp
+  CollapseStoresPatterns.cpp
```
schweitzpgi, line 21: It looks like these ought to be combined into 1 file. We put the patterns in the same source file with the pass.

lib/Optimizer/Transforms/CMakeLists.txt
```diff
@@ -34,9 +36,11 @@ add_cudaq_library(OptTransforms
   GlobalizeArrayValues.cpp
   LambdaLifting.cpp
   LiftArrayAlloc.cpp
+  LiftArrayAllocPatterns.cpp
```
schweitzpgi, line 39: I'm not sure why these are being split into two files in this PR?

schweitzpgi (changes requested):
I don't like the idea of making the patterns reside in files other than the pass where they are applied. That's not been the practice. So I think we should merge them back.

The exception to this general rule is when patterns are shared in multiple passes, which these are not AFAICT. Instead of block copying the same pattern in 2 or more different passes and where they will become hopelessly out of synch with one another, we want to put the pattern in an include file (`foo.inc`) and include that file in the _n_ passes where it is needed. The patterns in LowerToCFG for IfOp is such a case, since it will be the same rewrite in LowerToCFG (indiscriminate) and in the canonicalization pattern you want to add (selected only when the condition is constant).

lib/Optimizer/Dialect/CC/CCOps.cpp
annagrin: Replaced this pattern with RewriteIf from lower-to-cfg

lib/Optimizer/Dialect/CC/CCOps.cpp
annagrin: done, thanks!

lib/Optimizer/Transforms/CMakeLists.txt
annagrin: Renamed to WriteAfterWriteEliminationcpp and WriteAfterWriteEliminationPatterns.inc (used in loop unroll as well)

lib/Optimizer/Transforms/CMakeLists.txt
annagrin: I replaced this by using a LiftArrayAllocPatterns.inc, used in LoopUnroll.cpp as well

schweitzpgi (approved):
LGTM.

You probably want to get someone that understands UCCSD to look at that code however.

python/cudaq/kernels/uccsd.py
```diff
@@ -350,29 +350,37 @@ def uccsd_odd_electrons(qubits: cudaq.qview, thetas: list[float],
     lenVirtA = len(virtual_alpha_indices)
     lenVirtB = len(virtual_beta_indices)
 
-    singles_a = [[0, 0] for k in range(lenOccA * lenVirtA)]
+    singles_a0 = [0 for k in range(lenOccA * lenVirtA)]
+    singles_a1 = [0 for k in range(lenOccA * lenVirtA)]
```
marwafar, line 354: singles_a0 length should be lenOccA and singles_a1 length should be only lenVirtA. Are there a reason why you have the length is lenOccA*lenVirtA in both singles_a0, singles_a1.

python/cudaq/kernels/uccsd.py
```diff
@@ -350,29 +350,37 @@ def uccsd_odd_electrons(qubits: cudaq.qview, thetas: list[float],
...
         for q in virtual_alpha_indices:
-            singles_a[counter] = [p, q]
+            singles_a0[counter] = p
+            singles_a1[counter] = q
             counter = counter + 1
 
     counter = 0
-    singles_b = [[0, 0] for k in range(lenOccB * lenVirtB)]
+    singles_b0 = [0 for k in range(lenOccB * lenVirtB)]
+    singles_b1 = [0 for k in range(lenOccB * lenVirtB)]
```
marwafar, line 364: Same as above. singles_b0 size is lenOccB and singles_b size is lenVirtB

python/cudaq/kernels/uccsd.py
```diff
@@ -350,29 +350,37 @@ def uccsd_odd_electrons(qubits: cudaq.qview, thetas: list[float],
...
             counter = counter + 1
 
     counter = 0
-    doubles_m = [
-        [0, 0, 0, 0] for k in range(lenOccB * lenVirtB * lenOccA * lenVirtA)
-    ]
+    doubles_m0 = [0 for k in range(lenOccB * lenVirtB * lenOccA * lenVirtA)]
+    doubles_m1 = [0 for k in range(lenOccB * lenVirtB * lenOccA * lenVirtA)]
+    doubles_m2 = [0 for k in range(lenOccB * lenVirtB * lenOccA * lenVirtA)]
+    doubles_m3 = [0 for k in range(lenOccB * lenVirtB * lenOccA * lenVirtA)]
```
marwafar, line 375: same as singles

python/cudaq/kernels/uccsd.py
```diff
@@ -384,13 +392,18 @@ def uccsd_odd_electrons(qubits: cudaq.qview, thetas: list[float],
                     nEle = nEle + 1
 
     counter = 0
-    doubles_a = [[0, 0, 0, 0] for k in range(nEle)]
+    doubles_a0 = [0 for k in range(nEle)]
+    doubles_a1 = [0 for k in range(nEle)]
+    doubles_a2 = [0 for k in range(nEle)]
+    doubles_a3 = [0 for k in range(nEle)]
```
marwafar, line 398: same as before

marwafar (commented):
It looks correct to me. 

Before approve, it would be good if you can run a test for LiH molecule with this geometry = [('Li', (0.3925, 0., 0.)), ('H', (-1.1774, 0., 0.))] as in the H2 example in the cudaq doc page https://nvidia.github.io/cuda-quantum/latest/applications/python/vqe.html

Energy obtained should be -7.8826698. Calculation should take approx. 10 minutes.

annagrin:
> I don't like the idea of making the patterns reside in files other than the pass where they are applied. That's not been the practice. So I think we should merge them back.
> 
> The exception to this general rule is when patterns are shared in multiple passes, which these are not AFAICT. Instead of block copying the same pattern in 2 or more different passes and where they will become hopelessly out of synch with one another, we want to put the pattern in an include file (`foo.inc`) and include that file in the _n_ passes where it is needed. The patterns in LowerToCFG for IfOp is such a case, since it will be the same rewrite in LowerToCFG (indiscriminate) and in the canonicalization pattern you want to add (selected only when the condition is constant).

Addressed in previous PRs, thanks!

annagrin:
> It looks correct to me.
> 
> Before approve, it would be good if you can run a test for LiH molecule with this geometry = [('Li', (0.3925, 0., 0.)), ('H', (-1.1774, 0., 0.))] as in the H2 example in the cudaq doc page https://nvidia.github.io/cuda-quantum/latest/applications/python/vqe.html
> 
> Energy obtained should be -7.8826698. Calculation should take approx. 10 minutes.

Thank you Marwa, I discovered and unrelated bug while running this example. After the issue is fixed I will try running this.

PR #2468, opened 2024-12-11
Explicit registration of codegen dialect.

schweitzpgi:
Modify the cudaq-opt tool so that it explicit registers the codegen dialect. This change enables easier experimentation and has no effect on the existing compiler flow.


1tnguyen (approved):
LGTM 👍

PR #2470, opened 2024-12-11
Updating docs version index

bettinaheim:
Forgot to do that before...
Links to docs in notebooks and python files are generally suboptimal; there is no way for them to link to the correct docs version, so we need to necessarily direct them to the latest. The links are also not validated, and will break without anything failing if the page is removed in future versions on the docs. I don't have a better idea than to try and keep them somewhat general and unlikely to break.


1tnguyen (approved):
LGTM 👍

PR #2471, opened 2024-12-11
Convert static variable to a constexpr.

schweitzpgi:
Make this static data structure constant for efficiency and to eliminate an initialization ctor.


khalatepradnya (approved):
👍🏽

PR #2474, opened 2024-12-11
[test] Fix targettests for braket backend.

schweitzpgi:


khalatepradnya (approved):
LGTM! Thanks, Eric.
(I had missed the substitution logic in the first round)

PR #2475, opened 2024-12-12
Fix uninitialized memory issue for the result buffer in kron call

1tnguyen:
### Description

The kron kernel performs an accumulation-type operation (not an assignment) on the result buffer. Hence, this buffer must be initialized.

https://github.com/NVIDIA/cuda-quantum/blob/5785e44256b757263879580c82cb84adc85bcf5a/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu#L81

This issue only popped up intermittently if `cudaMalloc` returns a dirty buffer but can be observed reliably with `compute-sanitizer` (thanks @bmhowe23)


bmhowe23 (approved):
+1

PR #2477, opened 2024-12-12
Add a yapf style config file.

### LINKED ISSUE #2428: cuda-quantum needs a .style.yapf file added at the top-level directory
Generating this file should be easy with yapf itself. Adding the file to the source tree will prevent having to search for the "proper" yaof command line settings and prevent accidentally reformatting python code to "something weird".

schweitzpgi:
This makes the default "google" style and will help prevent people from accidentally using the yapf default style.

Fix #2428


bmhowe23 (commented):
FWIW, we put this into CUDA-QX. Hopefully it is equivalent to what you have here?

https://github.com/NVIDIA/cudaqx/blob/main/.style.yapf

schweitzpgi:
> FWIW, we put this into CUDA-QX. Hopefully it is equivalent to what you have here?
> 
> https://github.com/NVIDIA/cudaqx/blob/main/.style.yapf

Yes. The file being added is the output from the tool itself for the "google" style. I guess the verbose format is intended to make sure any changes to default settings in future releases of yapf don't cause all the source code to get reformatted? That's a "known" issue when using clang-format, for example.

bmhowe23:
> I guess the verbose format is intended to make sure any changes to default settings in future releases of yapf don't cause all the source code to get reformatted? That's a "known" issue when using clang-format, for example.

I suspect we should consider updating .github/workflows/repo_checks.yml, too, because it is calling `yapf --style google`, which would override this style file. That is - if a future release of yapf changed the default settings, that part of our CI would be unhappy with people attempting to use the yapf style file in the PR.

khalatepradnya (approved):
LGTM

.style.yapf
```diff
@@ -0,0 +1,2 @@
+[style]
+based_on_style = google
```
khalatepradnya, line 2: Seeing the nature of this file, I assume it does not need a copyright banner?

schweitzpgi: I think the IP is firmly at 0%. heh

bmhowe23 (approved):

PR #2478, opened 2024-12-13
Always clear `threadToQpuId` when we clear `platformQPUs`

1tnguyen:
### Description

We should clear `threadToQpuId` when we initialize the `platformQPUs` list (clear then repopulate).
The runtime has a static quantum platform instance, hence there could be stray info between run.

For example, a previous run with `nvidia-mqpu` may populate `threadToQpuId` based on the number of GPUs. Then, a subsequent `remote-mqpu` run may have a collision when looking up `threadToQpuId` with thread id => out of bound access to the `platformQPUs` map.  

This is primarily a Python specific bug as we cannot switch targets with C++.


bmhowe23 (commented):
This is a good find, Thien. Do we also need similar changes here?

https://github.com/NVIDIA/cuda-quantum/blob/2cf6b04d068aa40071a36f169795bebe28c61f3e/runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp#L84

https://github.com/NVIDIA/cuda-quantum/blob/2cf6b04d068aa40071a36f169795bebe28c61f3e/runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp#L123

(Those changes also do `platformQPUs.clear()` and would therefore seem to invalidate `threadToQpuId`, if it existed.

1tnguyen:
> This is a good find, Thien. Do we also need similar changes here?
> 
> https://github.com/NVIDIA/cuda-quantum/blob/2cf6b04d068aa40071a36f169795bebe28c61f3e/runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp#L84
> 
> https://github.com/NVIDIA/cuda-quantum/blob/2cf6b04d068aa40071a36f169795bebe28c61f3e/runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp#L123
> 
> (Those changes also do `platformQPUs.clear()` and would therefore seem to invalidate `threadToQpuId`, if it existed.

Good point. Added in https://github.com/NVIDIA/cuda-quantum/pull/2478/commits/b0b5f4ae3640e17e5ab52d53f6fac9a71d875c19.

I think for `DefaultQuantumPlatform`, this will not cause a problem as we only expect a single qpu (`qpu_id = 0`) but it's good to have both the vector and the map cleared at the same time throughout the code.

bmhowe23 (approved):
👍

PR #2479, opened 2024-12-13
Fixing wheel validation in publishing

bettinaheim:
The wheel validation as written is not configured properly to run in our publishing pipeline. 
- The applications folder should not be copied. Notebooks are validated by a separate script, and currently only during QA for notebooks that explicitly require GPU targets. While it may be nice to also run notebook validation during publishing at some point, copying the applications folder won't achieve that. 
- Running target tests requires credentials, and doesn't require the GPU runner. Hence, also running target tests should be set up separately. 
- I currently configured the validation script to silently skip nvqc snippets if NVQC_API_KEY is not set. I did set this key during publishing to validate nvqc snippets.
- A couple of other adjustments are needed, like installing matplotlib and fixing the torch download url.


1tnguyen (approved):
LGTM 👍

PR #2480, opened 2024-12-13
[core] Move the QIR peephole patterns out of tablegen.

### LINKED ISSUE #476: Rewrite tablegen rewrite patterns as C++ code

schweitzpgi:
Moves all the peephole patterns to explicit patterns and out of tablegen.

Fix #476.


lib/Optimizer/CodeGen/VerifyQIRProfile.cpp
```diff
@@ -9,7 +9,7 @@
 #include "PassDetails.h"
 #include "cudaq/Optimizer/Builder/Intrinsics.h"
 #include "cudaq/Optimizer/CodeGen/Passes.h"
-#include "cudaq/Optimizer/CodeGen/Peephole.h"
+#include "cudaq/Optimizer/CodeGen/QIRFunctionNames.h"
```
khalatepradnya, line 12: Clarifying question - why this change?

khalatepradnya (approved):
👍🏽

lib/Optimizer/CodeGen/VerifyQIRProfile.cpp
schweitzpgi: The verifying doesn't need the peephole patterns (it doesn't rewrite the code), but it was including the peephole header to get some function names. It should've included the names directly.

PR #2482, opened 2024-12-16
Extend the lifetime the constructed `LD_LIBRARY_PATH` string in `AutoLaunchRestServerProcess`

1tnguyen:
### Description

As we pass this constructed string as a `StringRef`, we need to keep it alive. Hence, use a member var for this purpose.


bettinaheim (approved):
Thank you for the fix!

PR #2489, opened 2024-12-31
Support for float comparison in ASTBridge

### LINKED ISSUE #2453: Float comparison operator throws throws error in cudaq.kernel
### Required prerequisites

### Describe the bug

Attempting to use comparison operators like `<` throws an error in a cudaq.kernel.

```
@cudaq.kernel
def test_kernel():
    a = 0.5 < 1.0
```
Will throw:
```
error: 'arith.cmpi' op operand #0 must be signless-integer-like, but got 'f64'
RuntimeError: Failure while executing pass pipeline.

During handling of the above exception, another exception occurred:

RuntimeError: could not compile code for 'test_kernel'.
```

### Expected behavior

Comparison operators should function for floats.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

sacpis:
Fixes #2453 

# Test result

```
@cudaq.kernel
def test_kernel():
    a = 0.5 < 1.0

module attributes {quake.mangled_name_map = {__nvqpp__mlirgen__test_kernel = "__nvqpp__mlirgen__test_kernel_PyKernelEntryPointRewrite"}} {
  func.func @__nvqpp__mlirgen__test_kernel() attributes {"cudaq-entrypoint"} {
    %true = arith.constant true
    %0 = cc.alloca i1
    cc.store %true, %0 : !cc.ptr<i1>
    return
  }
}
```

```
root@0a258c0dfa5e:/workspaces/cuda-quantum# pytest python/tests/kernel/test_ast_compare.py 
============================================================== test session starts ==============================================================
platform linux -- Python 3.10.12, pytest-8.2.0, pluggy-1.5.0
rootdir: /workspaces/cuda-quantum
configfile: pyproject.toml
plugins: anyio-4.6.2.post1
collected 20 items                                                                                                                              

python/tests/kernel/test_ast_compare.py ....................                                                                              [100%]

============================================================== 20 passed in 0.02s ===============================================================
```

```
100% tests passed, 0 tests failed out of 810

Label Time Summary:
gpu_required    = 241.29 sec*proc (381 tests)

Total Test time (real) = 1375.54 sec
```


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3377,73 +3377,96 @@ def visit_Compare(self, node):
         comparator = self.popValue()
```
schweitzpgi, line 3377: `comparator` should be renamed `right` to improve readability of the rest of this.

schweitzpgi (commented):
LGTM. @khalatepradnya ought to double check and approve the Python changes.

Thanks for the fixes, Sachin.

python/tests/kernel/test_ast_compare.py
```diff
@@ -0,0 +1,97 @@
...
+        # Mixed comparisons
+        (3, 4.5, "Lt", True),
+        (4.5, 3, "Gt", True),
+        (3, 3.0, "Eq", True),
+        (3, 4.5, "LtE", True),
+        (4.5, 4, "GtE", True),
+        (3, 4.5, "NotEq", True),
+    ],
+)
+def test_visit_compare(left, right, operation, expected):
```
khalatepradnya, line 67: Nice test!
For completeness, can you add another test that calls the operator from a CUDA-Q kernel? The example from the issue should suffice...
```
@cudaq.kernel
def test_kernel():
    a = 0.5 < 1.0
```

khalatepradnya: Ref: Ref: https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/mlir/ast_conditionals.py

python/tests/kernel/test_ast_compare.py
```diff
@@ -0,0 +1,109 @@
...
+
+def test_comparison_operators():
+
+    @cudaq.kernel
+    def test1():
+        a = 0.5 < 1.0
+
+    print(test1)
+
+    # CHECK-LABEL:    %true = arith.constant true
```
khalatepradnya, line 104: Thanks for the update! 
Note that the `CHECK-LABEL` needs `FileCheck` call on the output. For example, `# RUN: PYTHONPATH=../../ pytest -rP  %s | FileCheck %s`. So, this test will check that the kernel compiles but won't check output. The tests under `python/tests/mlir/` are set up to invoke `FileCheck`. You can consider moving this test case there. Or perhaps add a comment here to say - `Test here is that this kernel compiles`.

schweitzpgi (approved):

PR #1602, opened 2024-05-02
Use arch-specific pip requirements and update docs

bmhowe23:
Resolves #1570.
Resolves #1601. 

With these changes, x86 installation doesn't require any conda packages for most use cases, so it installs on Linux with a simple, standalone, `pip install cuda-quantum` now; however, aarch64 and multi-gpu x86 configurations still requires conda packages to be installed and LD_LIBRARY_PATH to be set.

Run demonstrating a failure:
https://github.com/NVIDIA/cuda-quantum/actions/runs/9035103824/job/24829689361

Run demonstrating success:
https://github.com/NVIDIA/cuda-quantum/actions/runs/9035426262/job/24830644405

PR #2501, opened 2025-01-10
Make Python CUDA dependencies resolve for arm64, too

bmhowe23:
#1602 did this for x86_64. This is an attempt to do the same for arm64 now. It seems to work on one machine I'm working on, so let's see if this makes it through the CI gauntlet.

~~Also, try to bump h5py to 3.12 since that version now has binary arm64 wheels provided again.~~ Edit: save this for a separate PR


schweitzpgi (commented):
LGTM.

pyproject.toml
```diff
@@ -22,10 +22,10 @@ dependencies = [
   'numpy >= 1.24',
   'scipy >= 1.10.1',
   'requests >= 2.31',
-  'nvidia-cublas-cu12 ~= 12.0; platform_machine == "x86_64"',
-  'nvidia-cuda-runtime-cu12 ~= 12.0; platform_machine == "x86_64"',
-  'nvidia-cusolver-cu12 ~= 11.4; platform_machine == "x86_64"',
-  'nvidia-cuda-nvrtc-cu12 ~= 12.0; platform_machine == "x86_64"'
+  'nvidia-cublas-cu12 ~= 12.0',
```
bmhowe23, line 25: @bettinaheim I'm not sure how the new Python metapackage infrastructure handles this for the different CUDA versions, but for CUDA 11, this needs to be
* 'nvidia-cublas-cu11 ~= 11.11'
* 'nvidia-cuda-runtime-cu11 ~= 11.8'
* 'nvidia-cusolver-cu11 ~= 11.4'
* 'nvidia-cuda-nvrtc-cu11 ~= 11.8'

Does his need to be handled somewhere else in the build scripts, and if so, where is that?

bmhowe23: > @bettinaheim I'm not sure how the new Python metapackage infrastructure handles this for the different CUDA versions, but for CUDA 11, this needs to be
> 
> * 'nvidia-cublas-cu11 ~= 11.11'
> * 'nvidia-cuda-runtime-cu11 ~= 11.8'
> * 'nvidia-cusolver-cu11 ~= 11.4'
> * 'nvidia-cuda-nvrtc-cu11 ~= 11.8'
> 
> Does his need to be handled somewhere else in the build scripts, and if so, where is that?

Perhaps this is already taken care of in these two places:
* https://github.com/NVIDIA/cuda-quantum/blob/da44eda8c3798018f90ea300f3a8e1042b939f44/docker/build/assets.Dockerfile#L190-L195
* https://github.com/NVIDIA/cuda-quantum/blob/da44eda8c3798018f90ea300f3a8e1042b939f44/docker/release/cudaq.wheel.Dockerfile#L42-L47

bettinaheim (approved):

PR #2505, opened 2025-01-14
[ci] Update docker_images jobs to use more CPUs for build

bmhowe23:
Many of our deployment jobs have been failing, presumably due to running out of memory during the build process, e.g. https://github.com/NVIDIA/cuda-quantum/actions/runs/12759701909/job/35565063041


1tnguyen (approved):
LGTM 👍 I think this would also speed up our pipelines!

PR #2515, opened 2025-01-16
Test seg fault

sacpis:


scripts/build_cudaq.sh
```diff
@@ -194,7 +194,7 @@ fi
 echo "Building CUDA-Q with configuration $build_configuration..."
 logs_dir=`pwd`/logs
 if $verbose; then 
-  ninja install
+  ninja -j 4 install
```
bmhowe23, line 197: Is this needed in line 201 also? (Also, I'm assuming this is just an experiment for now.)

bmhowe23: OK, that makes sense 👍

bmhowe23 (approved):

PR #2516, opened 2025-01-17
Fix a bug in tensornet backend scratch pad allocation in multi-GPU mode

1tnguyen:
### Description

`ScratchDeviceMem` allocates memory based on memory availability on construction. This mechanism is not compatible with multi-GPU code path (MPI execution), whereby the CUDA device is selected in the simulator constructor; hence we need to defer the allocation until the device is selected.

Fixed by having a separate `allocate` method to be called once during the simulator backend constructor after device selection.  

This bug was introduced in https://github.com/NVIDIA/cuda-quantum/pull/1865, where the scratch pad is allocated once (scratch pad is a member variable of the simulator class) rather than on-demand to improve performance.

Add a unit test for this case, to be executed when there are multiple GPUs.


bmhowe23 (approved):
LGTM

PR #2524, opened 2025-01-21
Upgrading notebook and jupyterlab version

sacpis:
Fix for exception id EXC-VZVQ-PUQE which consists of the following vulnerability

[GHSA-9q39-rmj3-p4r2](https://nspect.nvidia.com/vulnerability?id=GHSA-9q39-rmj3-p4r2)

Affected libraries are

jupyterlab
notebook

Fix given in: https://github.com/jupyterlab/jupyterlab/security/advisories/GHSA-9q39-rmj3-p4r2


1tnguyen (approved):
LGTM

bmhowe23:
> Seems like the deployment will most probably work. https://github.com/NVIDIA/cuda-quantum/actions/runs/12932443711
> 
> So will merge this PR now.

I think we may have (partially) identified the root cause for why the deployment isn't working (Ubuntu base VM image updates for our runner independent of the GitHub runners application that sits on top of that image), but just FYI to all, these were done with temporary runners that the main branch is not currently using, so main branches will still fail until we identify the best path forward.

PR #2528, opened 2025-01-21
[quera] Fix C++ frontend behavior

khalatepradnya:
As of CUDA-Q release version 0.9.1, the `quera` backend is supported only on Python frontend.
Prior to the changes proposed in this PR, if `quera` target was set from C++, it proceeded to use the default simulator and executed the circuit. With this change, an error will be thrown.

### DIFF
--- a/runtime/cudaq/platform/quera/CMakeLists.txt
+++ b/runtime/cudaq/platform/quera/CMakeLists.txt
@@ -6,21 +6,10 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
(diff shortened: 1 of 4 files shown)


1tnguyen (approved):
LGTM 👍

schweitzpgi:
The test for this needs to be run only if quera has been configured and successfully built in the cmake build configuration, etc.  Please see how braket tests use the `$braket_avail` variable.

As merged, the test fails every time.

khalatepradnya:
> The test for this needs to be run only if quera has been configured and successfully built in the cmake build configuration, etc. Please see how braket tests use the `$braket_avail` variable.
> 
> As merged, the test fails every time.

https://github.com/NVIDIA/cuda-quantum/pull/2532
Reusing `braket_avail` since `quera` has same dependency.

PR #2532, opened 2025-01-23
Follow-up to PR#2528: Make the newly added target test run conditionally

khalatepradnya:
Check if `quera` target is available - which in turn depends on `braket`


schweitzpgi (approved):
Thank you, Pradnya!

PR #2542, opened 2025-01-24
[core] This is a new implementation of QIR codegen.

schweitzpgi:
The QIR codegen is one of the oldest pieces of the project and really shows its age at this point. It was written before much of the rest of the core was implemented or even designed.

These changes replace the existing QIR codegen for _the C++ compiler_. Unfortunately, there are a number of semantics violations from Python that need to be fixed before it can be switched over. Therefore, this PR leaves Python intact and using the old codegen passes.

The purpose of these changes is two-fold.

1. Instead of converting all of the IR to the LLVM-IR dialect _at the same time_ as converting to QIR, these two steps are bifurcated. This allows us to convert the quake dialect to other higher level operations first and before LLVM-IR dialect is introduced. This will be beneficial in short order...
2. Instead of piecemealing different flavors of QIR in completely ad hoc spaghetti plate passes, the flavor of QIR is specified as a mixin modifier class for a singular set of steps to convert to any flavor of QIR. This does mean that one will no longer be able to convert to the LLVM-IR dialect with QIR calls and then change their mind from chocolate QIR to strawberry QIR much later.


lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1407,8 +1375,9 @@ class CustomUnitaryOpRewrite
     auto unitaryData =
         rewriter.create<LLVM::BitcastOp>(loc, complex64PtrTy, addrOp);
 
-    auto qirFunctionName =
-        std::string{cudaq::opt::QIRCustomOp} + (op.isAdj() ? "__adj" : "");
+    std::string qirFunctionName = cudaq::opt::QIRCustomOp;
+    if (op.isAdj())
+      qirFunctionName += "__adj";
```
khalatepradnya, line 1380: Can we use the newly added `QIRCustomAdjOp` directly?

schweitzpgi: Thanks. I'll make that change.

khalatepradnya (approved):
👏🏼 for the humungous effort!

I reviewed the changes under `python` and updates to the tests. I didn't closely review the new codegen pass.

PR #2546, opened 2025-01-27
Add support for -j option for build_cudaq.sh script

bmhowe23:
This is needed for some build environments that have a high CPU-to-memory ratio because without this change, ninja spawns too many jobs, and the machine can run out of memory.


khalatepradnya (approved):
LGTM!

khalatepradnya:
Q: Are we planning to use this in the CI?

bmhowe23:
> Q: Are we planning to use this in the CI?

We won't be using this in the CUDA-Q CI, but we will be using it in some steps of the CUDA-QX CI.

PR #2554, opened 2025-01-28
Python build updates for libraries

bmhowe23:
Trying to eliminate CUDA-Q customizations required to build CUDA-QX libraries.

Original author: @boschmitt 

Ref: https://github.com/NVIDIA/cudaqx/blob/c30376311bd15c1ec70933f2ad591ce5898da00e/.github/workflows/scripts/build_cudaq.sh#L96


python/runtime/cudaq/domains/plugins/CMakeLists.txt
```diff
@@ -17,6 +17,6 @@ if (SKBUILD)
     if (NOT Python_FOUND)
       message(FATAL_ERROR "find_package(Python) not run?")
     endif()
-    target_link_libraries(cudaq-pyscf PRIVATE Python::Python pybind11::pybind11 cudaq-chemistry cudaq-spin cudaq cudaq-py-utils)
+    target_link_libraries(cudaq-pyscf PRIVATE Python::Module pybind11::pybind11 cudaq-chemistry cudaq-spin cudaq cudaq-py-utils)
```
amccaskey, line 20: we should probably take this opportunity to mark the cudaq pyscf create_molecule stuff as deprecated in favor of the cuda-qx implementation.

amccaskey: Correct - I don't think this change is needed beyond this library, which we need to deprecate anyway.

boschmitt: > we should probably take this opportunity to mark the cudaq pyscf create_molecule stuff as deprecated in favor of the cuda-qx implementation.

It is most likely the necessary thing. My solution on CUDA-QX is a hack that breaks other CUDA-Q build scenarios. (As can be seen in the pipelines here.)

PR #2555, opened 2025-01-29
[fix] [quera] Follow-up to PR#2528 which unintentionally broke the Python support

khalatepradnya:
* Restore the server helper library target.
* Also, remove the redundant code for mock qpu which was not being used.

Follow-up for PR https://github.com/NVIDIA/cuda-quantum/pull/2528


bmhowe23 (approved):
LGTM

PR #2567, opened 2025-01-31
New `cudaq::sample` option: explicit measurements

bmhowe23:
~~This needs spec review/approval before being merged.~~ (Edit: spec has now been approved and merged.)

This PR introduces a new `sample_options` parameter called `explicit_measurements`. When this new option is selected, the behavior of the global register in the `sample_result` is changed. Rather than retaining a single measurement of the final "state" of each qubit in the system, this new option instructs the runtime to record all qubit measurements into the global register ... effectively "stacking" all measurements together.

While this change is supported by all simulators, the motivating factor for the change is that this will enable more efficient simulation when using Stim simulator for circuits containing mid-circuit measurements (as long as those mid-circuit measurements do not influence later gate sequences). Prior to this change, if one wanted to generically extract mid-circuit measurements out of the kernel, they would have to avoid using the `cudaq::sample` API.

### TODO
- [x] Finish spec, get approval
- [x] Error handling (precisely what an error is needs to be called out in the spec)
- [x] Add error-handling tests
- [x] Create Python tests
- [x] If time, update hardware backends; if no time, throw "unsupported" error message
- [x] Update docs


runtime/common/MeasureCounts.h
```diff
@@ -283,6 +283,9 @@ class sample_result {
   /// @return
   CountsDictionary::const_iterator end() const { return cend(); }
 
+  /// @brief Get the total number of shots in the result
+  std::size_t getTotalShots() const { return totalShots; }
```
amccaskey, line 287: snake case

bmhowe23: Updated in https://github.com/NVIDIA/cuda-quantum/pull/2567/commits/45d6c21fa34881a8710d5e8862e9c0fab77fea4a

runtime/cudaq/algorithms/sample.h
```diff
@@ -105,8 +105,18 @@ runSampling(KernelFunctor &&wrappedKernel, quantum_platform &platform,
       return std::nullopt;
     }
     platform.reset_exec_ctx(qpu_id);
-    counts += ctx->result;
+    if (counts.get_total_shots() == 0)
+      counts = std::move(ctx->result); // optimize for first iteration
```
khalatepradnya, line 109: Q: Do we need this optimization for Python as well? (in `sample.py`)

runtime/nvqir/CircuitSimulator.h
```diff
@@ -493,9 +510,23 @@ class CircuitSimulatorBase : public CircuitSimulator {
                            const std::string &regName) {
     if (executionContext && executionContext->name == "sample" &&
         !executionContext->hasConditionalsOnMeasureResults) {
+
+      // Handle duplicate measurements
+      if (executionContext->explicitMeasurements) {
+        /// TODO: Do this efficiently
```
bmhowe23, line 516: Is there still a TODO here? This looks reasonably efficient to me unless we add a new data structure like a hashmap to mirror `sampleQubits`. I'm not sure that's worth it.

runtime/cudaq/algorithms/sample.h
bmhowe23: Lol, I didn't see this comment until just now, but yeah, I posted the Python side of this in eb2d24df6.

runtime/nvqir/CircuitSimulator.h
khalatepradnya: I was considering these over the weekend, but it was too much work changing the data structure, and probably not worth it.

python/cudaq/runtime/sample.py
```diff
@@ -49,42 +57,67 @@ def sample(kernel, *args, shots_count=1000, noise_model=None):
...
 
+    has_conditionals_on_measure_result = False
+    if hasattr(kernel, 'metadata') and kernel.metadata.get(
+            'conditionalOnMeasure', False):
+        has_conditionals_on_measure_result = True
+
+    if explicit_measurements:
+        if not cudaq_runtime.supportsExplicitMeasurements():
+            raise RuntimeError(
+                "Explicit measurement option is not supported on this target.")
```
khalatepradnya, line 76: TODO: Make the error messages consistent - `Explicit measurement option` v/s `Explicit measurements option`

python/cudaq/runtime/sample.py
```diff
@@ -106,6 +106,11 @@ def sample(kernel,
             cudaq_runtime.unset_noise()
             return ctx.result
         counts += ctx.result
+        if explicit_measurements is True and (not counts.register_names or
+                                              not counts.most_probable()):
+            raise RuntimeError(
+                "Kernels executed in explicit measurements mode must contain measurements."
+            )
```
bmhowe23: Calling `register_names` and `most_probable` every time through the loop my impact throughput. Are they really necessary? Can this error message instead be raised in the `if counts.get_total_shots() == 0` section below?

I have the same comment/question for the change to `sample.h`.

khalatepradnya: Addressed in https://github.com/NVIDIA/cuda-quantum/pull/2567/commits/3b560aad9c37531c5a46ee5d6d4274f32bd3e8f9

khalatepradnya (approved):

PR #2569, opened 2025-01-31
[rt] Add requirement to find `cudensitymat.h`

boschmitt:
### Description
When bulding the `cudensitymat` backend, finding this header is a requirment. The build will fail otherwise.


1tnguyen (approved):
👍 Thanks, @boschmitt!

PR #2570, opened 2025-01-31
Fix issues in lift-array-alloc

annagrin:
### Description
Fix crashes and over-optimizations in lift-array-alloc:

Example:
```python
  qubits = cudaq.qvector(2)
  arr1 = [1]
  t = arr1[0]
  arr2 = [0]
  arr2[0] = t
  b = arr2[0]
  x(qubits[b])
```

This example generates the following IR:
```quake
func.func @test3() {
  %c1_i64 = arith.constant 1 : i64

  // qubits = cudaq.qvector(2)

  // arr1 = [1]
  %1 = cc.alloca !cc.array<i64 x 1>
  %2 = cc.cast %1 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
  cc.store %c1_i64, %2 : !cc.ptr<i64>

  // t = arr1[0]
  %3 = cc.load %2 : !cc.ptr<i64>

  // arr2 = [0]
  %4 = cc.alloca !cc.array<i64 x 1>
  %5 = cc.cast %4 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
  cc.store %c0_i64, %5 : !cc.ptr<i64> // Dominates the next store, don't lift

  // arr2[0] = t
  cc.store %3, %5 : !cc.ptr<i64>

  // b = arr2[0]
  %6 = cc.load %5 : !cc.ptr<i64>

  // x(qubits[b])
  %7 = quake.extract_ref %0[%6] : (!quake.veq<2>, i64) -> !quake.ref
  quake.x %7 : (!quake.ref) -> ()
  return
}
```

And lift-array-alloc misses the second store to `arr2` below and incorrectly assumes that `arr2[0]==0` instead of `arr2[0]==1`:

```
  cc.store %c0_i64, %5 : !cc.ptr<i64> // Dominates the next store, don't lift
  cc.store %3, %5 : !cc.ptr<i64>
```

The fix detects that the first store dominates the second and skips the optimization.


lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -89,12 +91,16 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
...
           if (auto load = dyn_cast<cudaq::cc::LoadOp>(useuser)) {
             rewriter.setInsertionPointAfter(useuser);
             LLVM_DEBUG(llvm::dbgs() << "replaced load\n");
-            rewriter.replaceOpWithNewOp<cudaq::cc::ExtractValueOp>(
-                load, eleTy, conArr,
+            auto extractValue = rewriter.create<cudaq::cc::ExtractValueOp>(
+                loc, eleTy, conArr,
                 ArrayRef<cudaq::cc::ExtractValueArg>{offset});
+            rewriter.replaceAllUsesWith(load, extractValue);
+            insertOpToErase(load);
```
schweitzpgi, line 103: I don't see why the `replaceOpWithNewOp` needs to be changed here. It will

  - create the new ExtractValueOp
  - replace all the users of the original LoadOp to use the new ExtractValueOp
 
 which is exactly what your change does. Additionally, we don't need or want to aggressively delete Ops. For example, passes may not be successful, and the MLIR infrastructure is designed to allow it to cleanly back out of the changes without leaving the IR broken. We should just let MLIR do the cleanups anywhere that it can. In this case, a LoadOp with 0 users will simply be removed by the next DCE. We don't have to do anything special.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
```diff
@@ -160,6 +167,7 @@ class AllocaPattern : public OpRewritePattern<cudaq::cc::AllocaOp> {
         if (!u)
           return nullptr;
         if (auto store = dyn_cast<cudaq::cc::StoreOp>(u)) {
+          storeSets[index].insert(u);
           if (op.getOperation() == store.getPtrvalue().getDefiningOp() &&
               isa_and_present<arith::ConstantOp, complex::ConstantOp>(
                   store.getValue().getDefiningOp())) {
```
schweitzpgi, line 173: How does any of this get past the check here? This is intended to check that we do not have multiple stores to the same offset of the array.

test/Quake/lift_array.qke
```diff
@@ -125,3 +125,50 @@ func.func @test2() -> !quake.veq<2> {
...
+  // t = arr1[0]
+  %3 = cc.load %2 : !cc.ptr<i64>
+
+  // arr2 = [0]
+  %4 = cc.alloca !cc.array<i64 x 1>
+  %5 = cc.cast %4 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
+  cc.store %c0_i64, %5 : !cc.ptr<i64> // Dominates the next store, don't lift
+
+  // arr2[0] = t
+  cc.store %3, %5 : !cc.ptr<i64>
```
schweitzpgi, line 150: The pass should refuse to convert this code because there are 2 stores to the same offset. But I don't see why it wasn't already.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
schweitzpgi: I ran a little test, and this check correctly blocks the transformation.

```mlir
func.func @f() {
    %0 = cc.alloca !cc.array<i64 x 1>
    %1 = arith.constant 1 : i64
    %2 = cc.compute_ptr %0[0] : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
    cc.store %1, %2 : !cc.ptr<i64>
    %3 = arith.constant 3 : i64
    cc.store %3, %2 : !cc.ptr<i64>
    return
}
```

Running lift-array-alloc does not change this code, which is the correct thing to happen.

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Not sure why but the original code crashes on uccsd example

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I think the failing example differs on the second store not using a constant to store to a pointer

test/Quake/lift_array.qke
annagrin: second store is not a constant. I will update the fix

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: I removed this change for now, let's see if it happens again when all the uccsd changes come together and I might get a repro

lib/Optimizer/Transforms/LiftArrayAlloc.cpp
annagrin: Found the reasons why this check does not detect two stores to the same pointer: 1) the second store is appearing first in the list of users of the ptr that we are processing 2) it does not store a constant to the pointer. Updated the check.

test/Quake/lift_array.qke
annagrin: Updated the fix

schweitzpgi:
My testing shows proper behavior with multiple stores to a single location in an stack allocated array object. Specifically, lift-array-alloc correctly detects this condition already and refuses to promote the array from a stack-based object to a first-class value. That is the correct behavior.

A pass that eliminates useless stores could be run prior to lift-array-alloc, which should only under the correct conditions, throw away stores to locations in the stack which are never read but immediately clobbered by another store (WAW) to the same location.

schweitzpgi (approved):
Thanks for fixing these bugs and the new regression tests, Anna. Nice work.

PR #2571, opened 2025-01-31
Fix issues with python range boundaries

annagrin:
### Description
Fix issues with python range boundaries

Example:
```python
range(-1)
```
generated 1 element instead of none.


khalatepradnya (approved):
Thanks, Anna! Changes LGTM!

python/tests/kernel/test_kernel_features.py
```diff
@@ -557,6 +557,74 @@ def kernel(myList: list[int]):
...
+        qubits = cudaq.qvector(4)
+        r = [i * 2 + 1 for i in range(2)]
+        for i in r:
+            x(qubits[i])
+
+    counts = cudaq.sample(kernel6)
+    assert len(counts) == 1
+    assert '0101' in counts
+
+
```
khalatepradnya, line 627: Regarding tests.. I see that CI is failing because the following two tests which use `range` now generate a different IR:
```
Failed Tests (2):
  PYCUDAQMLIR :: test_output_qir.py
  PYCUDAQMLIR :: test_output_translate_qir.py
```

annagrin: Updated those, thanks!

PR #2572, opened 2025-02-01
Add WriteAfterWriteElimination pass

annagrin:
### Description
Add a pass for removing stores that are overridden by other stores

Example:

```
func.func @test_two_stores_same_pointer() {
  %1 = cc.const_array [1] : !cc.array<i64 x 1>
  %2 = cc.extract_value %1[0] : (!cc.array<i64 x 1>) -> i64
  %3 = cc.alloca !cc.array<i64 x 1>
  %4 = cc.cast %3 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
  cc.store %c0_i64, %4 : !cc.ptr<i64>
  cc.store %2, %4 : !cc.ptr<i64>
  %5 = cc.load %4 : !cc.ptr<i64>
  %6 = quake.extract_ref %0[%5] : (!quake.veq<2>, i64) -> !quake.ref
  quake.x %6 : (!quake.ref) -> ()
  return
}
```
The in the last two stores, the first one can be removed because it is overridden by the next.
```
  cc.store %c0_i64, %4 : !cc.ptr<i64>
  cc.store %2, %4 : !cc.ptr<i64>
```


iqbalbhatti49 (approved):
### Review Summary:

The code looks great overall! You've followed best practices, and the implementation is clean and efficient. The logic is sound, and I appreciate the attention to detail in the structure.

### Highlights:
- Code is well-organized and easy to follow.
- Proper handling of edge cases.
- Clear variable naming makes the code readable.

Overall, I approve this change and recommend merging it. Great work!

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -975,6 +975,29 @@ def RegToMem : Pass<"regtomem", "mlir::func::FuncOp"> {
   let dependentDialects = ["cudaq::cc::CCDialect", "quake::QuakeDialect"];
 }
 
+def RemoveUselessStores : Pass<"remove-useless-stores", "mlir::func::FuncOp"> {
```
schweitzpgi, line 978: I think we should call this "write-after-write-elimination`.

We can also make it a generic pass, since we may want to run it on more than `func.func`.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -975,6 +975,29 @@ def RegToMem : Pass<"regtomem", "mlir::func::FuncOp"> {
...
+  let summary = "Remove stores that are overriden by other stores.";
+  let description = [{
+
```
schweitzpgi, line 981: Can we add more words here? I think the example is good, but examples aren't the best documentation as they can leave some unanswered questions.

lib/Optimizer/Transforms/RemoveUselessStores.cpp
```diff
@@ -0,0 +1,174 @@
...
+  //       return true;
+
+  //   for (auto &region : op->getRegions())
+  //     for (auto &b : region)
+  //       for (auto &innerOp : b)
+  //         if (isUsed(v, &innerOp))
+  //           return true;
+
+  //   return false;
+  // }
```
schweitzpgi, line 146: Remove?

lib/Optimizer/Transforms/RemoveUselessStores.cpp
```diff
@@ -0,0 +1,174 @@
...
+          // // Found a use for a current ptr before the overriding store
+          // if (currentPtr && isUsed(currentPtr, &op))
+          //   return false;
+        }
+      }
+      // } else {
+      //   // Found a use for a current ptr before the overriding store
+      //   if (currentPtr && isUsed(currentPtr, &op))
+      //     return false;
+      // }
```
schweitzpgi, line 86: Remove?

lib/Optimizer/Transforms/RemoveUselessStores.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (auto cast = ptrOp.getDefiningOp<cudaq::cc::CastOp>())
+      ptrOp = cast.getOperand();
+
+    if (auto computePtr = ptrOp.getDefiningOp<cudaq::cc::ComputePtrOp>())
+      ptrOp = computePtr.getBase();
+
+    if (auto alloca = ptrOp.getDefiningOp<cudaq::cc::AllocaOp>())
+      return true;
+
+    return false;
```
schweitzpgi, line 113:
```suggestion
    return isa_and_present<cudaq::cc::AllocaOp>(ptrOp.getDefiningOp());
```

lib/Optimizer/Transforms/RemoveUselessStores.cpp
```diff
@@ -0,0 +1,174 @@
...
+  /// block.
+  bool isUselessStore(cudaq::cc::StoreOp store) const {
+    Value currentPtr;
+
+    if (!isStoreToStack(store))
+      return false;
+
+    auto block = store.getOperation()->getBlock();
+    for (auto &op : *block) {
+      if (auto currentStore = dyn_cast<cudaq::cc::StoreOp>(&op)) {
```
schweitzpgi, line 67: I was thinking of doing this with an analysis step, first.

Have a `DenseMap<Block*, DenseMap<Value, SmallVector<cudaq::cc::StoreOp*>>>`
Scan the current operation's regions for `cc::StoreOp`s. For each one found, add it to the map object by its block and ptr value argument.

After that go back over the map of maps, see if any of the vectors have `> 1` element, and use dominance info to make sure no uses of the ptr value are dominated by the dominant store and dominate the post-dominant store other than any other stores in the block between the two (most dominant, most postdominant). Dominance here will catch uses into, e.g., cc::IfOp, etc.

If there is nothing but stores in the block between the two endpoint stores, go ahead and remove all but the postdominant one.

A more aggressive check would be to eliminate the dominant stores in a pair-wise, incremental approach. So if there is a `store, if(with a load), store[*], store `, the dead `store[*]` would be eliminated even though the first one cannot be removed.

schweitzpgi (commented):
"Useless" sounds a little judgmental or something to me. Maybe WAW elimination or WAW removal sounds less aggressive? heh

lib/Optimizer/Transforms/RemoveUselessStores.cpp
annagrin: I'll try that, thanks! I do need the most aggressive approach it seem from the uccsd examples

annagrin: done!

schweitzpgi (approved):

PR #2573, opened 2025-02-02
Normalize loops with negative upper bounds

annagrin:
### Description
Normalize loops with negative upper boundaries

Example of a loop that was not previously normalized and subsequently not unrolled
```
%1 = cc.loop while ((%arg0 = %c0_i32) -> (i32)) {
%2 = arith.cmpi slt, %arg0, %c-1_i32 : i32
      cc.condition %2(%arg0 : i32)
} do {
...
```

Towards: https://github.com/NVIDIA/cuda-quantum/pull/2458


schweitzpgi:
The description is missing the step expression, which is a critical bit of information here.

schweitzpgi (approved):
Thanks for the fix, Anna.

PR #2580, opened 2025-02-03
Canonicalize const ifs

annagrin:
### Description
Canonicalize const ifs

Example:

```
    cc.if(%true) {
      %6 = quake.extract_ref %0[0] : (!quake.veq<2>) -> !quake.ref
      quake.x %6 : (!quake.ref) -> ()
    }
```

Should be converted to

```
      %6 = quake.extract_ref %0[0] : (!quake.veq<2>) -> !quake.ref
      quake.x %6 : (!quake.ref) -> ()
```

Update: this change affects too many tests. I am planning to just add the pattern in question to LoopUnroll for now in a next PR


annagrin:

PR #2590, opened 2025-02-05
[py][nfc] `import`s clean up.

boschmitt:
### Description
Cleans up imports throughout the python runtime. I trie following the principles in [PEP8](https://peps.python.org/pep-0008/#imports).

The changes include:
  * Grouping imports two blocks: external and internal, w.r.t cudaq.
  * Not using relative paths when the module is in a parent dir (`..`)
  * Not using wildcard imports
  * Remove unused ones

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -7,20 +7,36 @@
 # ============================================================================ #
 import ast
 import importlib
-import hashlib
 import graphlib
-import sys, os
-from typing import Callable
-from collections import deque
 import numpy as np
+import os
+import sys
+from collections import deque
+
+from cudaq.mlir._mlir_libs._quakeDialects import (
+    cudaq_runtime, load_intrinsic, gen_vector_of_complex_constant,
+    register_all_dialects)
+from cudaq.mlir.dialects import arith, cc, complex, func, math, quake
+from cudaq.mlir.ir import (BoolAttr, Block, BlockArgument, Context, ComplexType,
+                           DenseBoolArrayAttr, DenseI32ArrayAttr,
+                           DenseI64ArrayAttr, DictAttr, F32Type, F64Type,
+                           FlatSymbolRefAttr, FloatAttr, FunctionType,
+                           InsertionPoint, IntegerAttr, IntegerType, Location,
+                           Module, StringAttr, SymbolTable, TypeAttr, UnitAttr)
+from cudaq.mlir.passmanager import PassManager
 from .analysis import FindDepKernelsVisitor
-from .utils import globalAstRegistry, globalKernelRegistry, globalRegisteredOperations, nvqppPrefix, mlirTypeFromAnnotation, mlirTypeFromPyType, Color, mlirTypeToPyType, globalRegisteredTypes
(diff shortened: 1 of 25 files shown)


iqbalbhatti49 (commented):
Good cleanup of imports! Well done following PEP8 principles.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -6,20 +6,36 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 import ast
```
khalatepradnya, line 8: nit: I think we are inconsistent with a blank line after the copyright banner and fist import.

python/cudaq/kernel/utils.py
```diff
@@ -6,16 +6,16 @@
...
-from ..mlir.passmanager import *
-import numpy as np
-from typing import Callable, List
-import ast, sys, traceback
+import ast
 import re
-from typing import get_origin
+import sys
+import traceback
+import numpy as np
```
khalatepradnya, line 13: Q: Are the external imports sorted alphabetically?

python/cudaq/operator/cudm_solver.py
```diff
@@ -10,10 +10,9 @@
 from typing import Sequence, Mapping, List, Optional
 
 from .cudm_helpers import CuDensityMatOpConversion, constructLiouvillian
-from ..runtime.observe import observe
 from .schedule import Schedule
 from .expressions import Operator
-from ..mlir._mlir_libs._quakeDialects import cudaq_runtime
+from cudaq.mlir._mlir_libs._quakeDialects import cudaq_runtime
```
khalatepradnya, line 15: Q: Is there relative ordering for the `.` imports and named internal imports?

khalatepradnya (approved):
Minor comments. LGTM.
Thanks, Bruno! 👍🏽

khalatepradnya:
@boschmitt : In case you used a tool / script for this, can you share. Perhaps we can add that as part of CI ?

python/cudaq/kernel/ast_bridge.py
boschmitt: Indeed, we don't seem to have any guideline regarding these things.

python/cudaq/kernel/utils.py
boschmitt: I was going for that but it's really an arbitrary decision. The only exception being `from __future__` which is required to be the first.

python/cudaq/operator/cudm_solver.py
boschmitt: It should have. The idea was to have project imports followed by relative imports. Also, I did change things in a way that relative imports are only for modules within the same directory (so moving the whole directory wouldn't require changing the imports)

boschmitt:
> @boschmitt : In case you used a tool / script for this, can you share. Perhaps we can add that as part of CI ?

I did it by hand while studying the code, so it is really a best-effort sort of thing and is not really free of inconsistencies. 

Thanks for the review (:

iqbalbhatti49 (approved):

PR #2591, opened 2025-02-05
Fix failures in write-after-write-elimination

annagrin:
### Description
Fix failures in write-after-write-elimination


schweitzpgi (approved):
Thanks for the bug fixes.

Issue #2579, opened 2025-02-03
cupy MemoryPointer yields error with kernel


zohimchandani:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The `cupy` import below throws an error which disappears when the import is commented out: 

```
import cudaq
from cupy.cuda.memory import MemoryPointer, UnownedMemory

cudaq.set_target('nvidia')

num_qubits = 5

@cudaq.kernel
def kernel():
    qubits = cudaq.qvector(num_qubits)
    x(qubits)

initial_state = cudaq.get_state(kernel)

```

```
AttributeError: type object 'cupy.cuda.memory.MemoryPointer' has no attribute '__annotations__'

```

### Steps to reproduce the bug

NA

### Expected behavior

NA

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

NA

### Suggestions

NA

Issue #2531, opened 2025-01-23
Conflict between cudaq and collections.OrderedDict causing AttributeError


anthonysmaldone:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When importing `OrderedDict` from the `collections` module in Python, any attempt to define or use a `@cudaq.kernel` results in an `AttributeError`. This appears to be a conflict between `cudaq` and the standard Python library, specifically the `collections` module. The issue makes it impossible to use `OrderedDict` alongside `cudaq`, which could affect projects relying on both.

### Steps to reproduce the bug

```
import cudaq
from collections import OrderedDict

# Define our kernel.
@cudaq.kernel
def kernel(qubit_count: int):
    # Allocate our qubits.
    qvector = cudaq.qvector(qubit_count)
    # Place the first qubit in the superposition state.
    h(qvector[0])
    # Loop through the allocated qubits and apply controlled-X,
    # or CNOT, operations between them.
    for qubit in range(qubit_count - 1):
        x.ctrl(qvector[qubit], qvector[qubit + 1])
    # Measure the qubits.
    mz(qvector)
```

### Expected behavior

The script should execute without errors, as importing `OrderedDict` from `collections` is a standard Python operation.

Instead, it fails with the following error:

`AttributeError: type object 'collections.OrderedDict' has no attribute '__annotations__'`

### Environment

- **CUDA-Q version**:  0.9.1 (commit hash a3ab722eab)
- **Python version**:  3.10.16
- **Operating system**: HPE Cray OS (SUSE Linux Enterprise Server, SLES)

PR #2592, opened 2025-02-06
[py] Fixes #2586 and #2531 and #2579

### LINKED ISSUE #2586: CUDA-Q kernel cannot use with itertools.product
### Required prerequisites

### Describe the bug

```python
from itertools import product
import cudaq

@cudaq.kernel
def test():
    q = cudaq.qubit()
    x(q)
```
does not work...

Error message was
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 5
      1 from itertools import product as product_
      2 import cudaq
      4 @cudaq.kernel
----> 5 def test():
      6     q = cudaq.qubit()
      7     x(q)

File [/opt/nvidia/cudaq/cudaq/kernel/kernel_decorator.py:526](http://10.20.100.80:1519/opt/nvidia/cudaq/cudaq/kernel/kernel_decorator.py#line=525), in kernel(function, **kwargs)
    517 """
    518 The `cudaq.kernel` represents the CUDA-Q language function 
    519 attribute that programmers leverage to indicate the following function 
   (...)
    523 Verbose logging can be enabled via `verbose=True`. 
    524 """
    525 if function:
--> 526     return PyKernelDecorator(function)
    527 else:
    529     def wrapper(function):

File [/opt/nvidia/cudaq/cudaq/kernel/kernel_decorator.py:96](http://10.20.100.80:1519/opt/nvidia/cudaq/cudaq/kernel/kernel_decorator.py#line=95), in PyKernelDecorator.__init__(self, function, verbose, module, kernelName, funcSrc, signature, location, overrideGlobalScopedVars)
     94 for name, var in self.globalScopedVars.items():
     95     if isinstance(var, type):
---> 96         globalRegisteredTypes[name] = (var, var.__annotations__)
     98 # Once the kernel is compiled to MLIR, we
     99 # want to know what capture variables, if any, were
    100 # used in the kernel. We need to track these.
    101 self.dependentCaptures = None

AttributeError: type object 'itertools.product' has no attribute '__annotations__'
```

### Expected behavior

no error

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

I'm using CUDA-Q container for 0.9.1

### Suggestions

_No response_

boschmitt:
### Description

Not all objects that pass the `isinstance(var, type)` check have the `__annotations__` attribute.

This is a quick fix, i.e., the change does not address the much deeper underlying issue: this code should not be globing all globals within a frame, and perhaps using `__annotations__` is undesirable. (Also, the best practices regarding the use of annotations have changed in python 3.10, so there are still unaddressed landmines here.)

Fixes #2586 #2531 #2579

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -92,7 +92,7 @@ class PyKernelDecorator(object):
         # Register any external class types that may be used
         # in the kernel definition
         for name, var in self.globalScopedVars.items():
-            if isinstance(var, type):
+            if isinstance(var, type) and hasattr(var, '__annotations__'):
                 globalRegisteredTypes[name] = (var, var.__annotations__)
 
         # Once the kernel is compiled to MLIR, we


bmhowe23 (approved):
+2

boschmitt:
Also fixes #2579

PR #2594, opened 2025-02-06
[features/run] Add function to convert data buffer to a std::vector.

schweitzpgi:
Once the decoder produces a buffer of values, as interpreted from the output log, one can use this function to convert to buffer into a std::vector<T> object. This function transfers the ownership of the data buffer to the vector instead of making another copy of the data.

### DIFF
--- a/runtime/cudaq/algorithms/run.h
+++ b/runtime/cudaq/algorithms/run.h
@@ -16,6 +16,11 @@
 #include "cudaq/host_config.h"
 #include <cstdint>
 
+extern "C" {
+void __nvqpp_initializer_list_to_vector_bool(std::vector<bool> &, char *,
+                                             std::size_t);
+}
+
 namespace cudaq {
 
 namespace details {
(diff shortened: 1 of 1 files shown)


runtime/cudaq/algorithms/run.h
```diff
@@ -37,6 +43,47 @@ struct RunResultSpan {
...
+    T *end0;
+    T *end1;
+  };
+  static_assert(sizeof(std::vector<T>) == sizeof(raw_vector) &&
+                "std::vector must use the nominal 3 pointer implementation");
+
+  // Swap vec into a local variable. vec's original content, if any will be
+  // reclaimed at the end of this function.
+  std::vector<T> deadEnder;
+  std::swap(deadEnder, result);
```
annagrin, line 66: Would returning the local variable have the same effect (instead of passing a reference as an argument and swapping it with a local? I mean after the RVO?:)

schweitzpgi: Yes. What I have here is manual RVO.

annagrin (approved):
LGTM

PR #2596, opened 2025-02-06
Add more patterns to loop unroll pass

annagrin:
### Description
Add more patterns to loop unroll pass

To be able to optimize away loops with boundaries that are computed from array storage, other loops, conditions, and more, we need to run a bunch of optimizations together in the loop unroll pass.

Added following optimizations to `cc-loop-unroll`:
- `LoopOpt` pattern from `cc-loop-normalize` (to keep normalizing loops in loop unrolling iterations)
- `RewriteIf` from `lower-to-cfg` (to const-prop if statements)
- `SimplifyWritesAnalysis` from `write-after-write-elimination` (remove multiple writes to the same location on the stack)
- `simplifyRegions` (to remove dead code)
- `AllocaPattern` from `lift-array-alloc` (to lift constant arrays, helps const prop)

Added tests.


lib/Optimizer/Transforms/LoopUnroll.cpp
```diff
@@ -343,12 +370,11 @@ static void createUnrollingPipeline(OpPassManager &pm, unsigned threshold,
                                     bool signalFailure, bool allowBreak,
                                     bool allowClosedInterval) {
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  pm.addNestedPass<func::FuncOp>(createCSEPass());
   pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
-  pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-  cudaq::opt::LoopNormalizeOptions lno{allowClosedInterval, allowBreak};
-  pm.addNestedPass<func::FuncOp>(cudaq::opt::createLoopNormalize(lno));
```
schweitzpgi, line 349: Removing loop normalization doesn't look correct.

annagrin: I added loop normalization pattern into the LoopUnroll pass itself so it is no longer needed here

annagrin: I am thinking - given that I also need something similar to the lift-array-alloc pass - would it be better to create classical-optimization-pipeline that would run a bunch of patterns for the same fix point, on all code (not only when loops are present), including the following:

classical-optimization-pipeline:
- memtoreg=quantum=0
- classical-optimizations
- cse
- classical-optimizations

where classical-optimization pass runs the following patterns as a part of single fix point (same as a LoopUnroll pass in current PR, but run them on code that does not contain loops as well):
- aggressive-early-inlining
- write-after-write-elimination
- lift-array-alloc,
- canonicalize,
- loop-normalize patterns,
- original loop unrolling patterns

We would use `classical-optimization-pipeline` instead of the initial set of opts in quantum config files: 
`func.func(const-prop-complex,canonicalize,cse,lift-array-alloc)` 

And later again after the following passes:
`apply-op-specialization,aggressive-early-inlining,expand-measurements`

annagrin: So, as an example, the new pipeline for Ionq would look like:

`"func.func(classical-optimization-pipeline),
   globalize-array-values,func.func(state-prep),unitary-synthesis,canonicalize,
   apply-op-specialization,aggressive-early-inlining,expand-measurements,
   func.func(classical-optimization-pipeline),
   decomposition{enable-patterns=U3ToRotations},
   func.func(lower-to-cfg,canonicalize,multicontrol-decomposition),
   ionq-gate-set-mapping"`

annagrin:

Issue #2597, opened 2025-02-07
Can no longer build CUDA-QX with latest CUDA-Q


bmhowe23:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

nvq++ compilation of a file that used to be buildable is now failing. I used `git bisect` to narrow this down to a53c378dea4f46e65744342945d61505b7b47f7a.

### Steps to reproduce the bug

See https://github.com/NVIDIA/cudaqx/actions/runs/13190031945/job/36821231437?pr=70#step:10:196

Snippet:
```bash
[quake-translate] Dumping Module after error.
[quake-translate] Reported Error: see current operation: %5 = "llvm.mlir.undef"() : () -> !cc.struct<{!cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>}>
loc("memory_circuit.qke.m55agh":9:10): error: 'llvm.mlir.undef' op result #0 must be LLVM dialect-compatible type, but got '!cc.struct<{!cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>}>'
[quake-translate] Dumping Module after error.
[quake-translate] Reported Error: see current operation: %5 = "llvm.mlir.undef"() : () -> !cc.struct<{!cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>, !cc.ptr<!llvm.struct<"Array", opaque>>}>
[...]

PR #2598, opened 2025-02-07
A couple of bug fixes.

schweitzpgi:
Fix for #2597


bmhowe23 (approved):
Thanks for the quick fixes, Eric.
These were also tested here: https://github.com/NVIDIA/cudaqx/actions/runs/13191750273

PR #2599, opened 2025-02-07
[Follow up #PR 2488] Removing the Italics header

sacpis:
Fixes https://github.com/NVIDIA/cuda-quantum/pull/2488#issuecomment-2640658066

# Output

image


khalatepradnya (approved):
Thanks, Sachin!

PR #2601, opened 2025-02-09
Run canonical patterns alongside lift-array-alloc

annagrin:
### Description
Run canonical patterns alongside lift-array-alloc to make sure all constant arrays are lifted in case of array copy:

```python
arr = [0]
arr1 = [arr[0]]
return arr[0] == arr1[0]
```

Previously, we would need to run two iterations of `lift-array-alloc,canonicalize` to lift both arrays, as the first one needs to be lifted and then const-propped before the second one could be lifted as well.


annagrin:

PR #2603, opened 2025-02-10
[core] Support for apply_noise.

schweitzpgi:
Add an apply_noise operator.
Teach the bridge to recognize cudaq::apply_noise and lower it.
Add a pass to erase apply_noise operations.
Modify QIR codegen to generate the C++ callback.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -493,7 +493,7 @@ def quake_ApplyNoiseOp : QuakeOp<"apply_noise", [AttrSizedOperandSegments]> {
 
   let arguments = (ins
     FlatSymbolRefAttr:$noise_func,
-    Variadic<AnyFloat>:$parameters,
+    Variadic<PointerOf<[AnyFloat]>>:$parameters,
     Variadic<NonStruqRefType>:$qubits
   );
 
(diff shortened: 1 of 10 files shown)


amccaskey:
Probably coming in a bit too early, but pulled this to kick the tires. 

Couple of bugs for this kraus_channel 
```cpp
#include "cudaq.h"

namespace test {

struct my_channel : public ::cudaq::kraus_channel {
  my_channel(const std::vector<double> &params) {
    std::vector<cudaq::complex> k0v{std::sqrt(1 - params[0]), 0, 0,
                                    std::sqrt(1 - params[0])},
        k1v{0, std::sqrt(params[0]), std::sqrt(params[0]), 0};
    push_back(cudaq::kraus_op(k0v));
    push_back(cudaq::kraus_op(k1v));
  }
};
} // namespace test

__qpu__ void test2(double d) {
  cudaq::qubit q;
  x(q);
  cudaq::apply_noise<test::my_channel>(d, q);
}

```

First one is that we do want apply_noise implemented such that the kraus_channel is returned from the noise_model. As of now we get an error for KrausChannelT not be default constructible.
```bash 
/workspaces/cuda-quantum/runtime/cudaq/qis/qubit_qis.h:49:17: error: no matching constructor for initialization of 'test::my_channel'
  KrausChannelT channel;
```
Next issue is that the double argument here is passed by reference, so the type that the bridge sees is `cc.ptr<f64>` which does not pass the error checking logic in ConvertExpr.
```bash 
ericNoise.cpp:23:3: error: apply_noise argument types not supported.
  cudaq::apply_noise<test::my_channel>(d, q);
  ^
```

PR #2605, opened 2025-02-11
Add a new classical optimization pipeline to group a bunch of existing patterns

annagrin:
### Description
Add a new classical optimization pipeline to group runs a bunch of existing patterns.

Group a bunch of classical optimizations together so we can optimize away more code:
- loops with boundaries that are computed from array storage, other loops, or conditions
- arrays that are computed in loops, from other arrays, etc.

Implemented a new `classical-optimization` pass that runs the following patterns until a fix point is reached:
- `LoopOpt` pattern from `cc-loop-normalize` (to keep normalizing loops in loop unrolling iterations)
- `RewriteIf` from `lower-to-cfg` (to const-prop if statements)
- `SimplifyWritesAnalysis` from `write-after-write-elimination` (remove multiple writes to the same location on the stack)
- `simplifyRegions` (to remove dead code)
- `AllocaPattern` from `lift-array-alloc` (to lift constant arrays, helps const prop)

Implemented a `classical-optimization-pipeline` pass that invokes the `classical-optimization` pass and its prerequisites.

Note: currently we need to run `classical-optimization` pass twice in `classical-optimization-pipeline` above to optimize complex code such as `uccsd` state prep function. This is due to not being able to run a `cse` optimization as part of the fix point. A newer version of llvm should allow us to do that in the future.

Added tests.


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -141,6 +141,40 @@ def CheckKernelCalls : Pass<"check-kernel-calls", "mlir::func::FuncOp"> {
...
+
+    The signal-failure-if-any-loop-cannot-be-completely-unrolled option controls
+    whether to signal a failure if all loops cannot be fully unrolled. This is
+    necessary when synthesizing quantum circuits from CUDA-Q kernels, such
+    as when generating a QIR base profile. A quantum circuit requires all loops
+    be completely unrolled.
+  }];
+
+  let dependentDialects = ["mlir::arith::ArithDialect",
+                           "mlir::cf::ControlFlowDialect"];
```
schweitzpgi, line 163: Consider maybe adding CCDialect for completeness.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -141,6 +141,40 @@ def CheckKernelCalls : Pass<"check-kernel-calls", "mlir::func::FuncOp"> {
...
+
+  let options = [
+    Option<"threshold", "maximum-iterations", "unsigned", /*default=*/"50",
+      "Maximum iterations to unroll.">,
+    Option<"signalFailure",
+      "signal-failure-if-any-loop-cannot-be-completely-unrolled", "bool",
+      /*default=*/"false", "Signal failure if pass can't unroll all loops.">,
```
schweitzpgi, line 170: Do we need this? The fix-point may be before every single cc.loop is unrolled. (This option isn't composable.)

lib/Optimizer/Transforms/ClassicalOptimization.cpp
```diff
@@ -0,0 +1,168 @@
...
+    : public PassPipelineOptions<ClassicalOptimizationPipelineOptions> {
+  PassOptions::Option<unsigned> threshold{
+      *this, "threshold",
+      llvm::cl::desc("Maximum iterations to unroll. (default: 1024)"),
+      llvm::cl::init(1024)};
+  PassOptions::Option<bool> signalFailure{
+      *this, "signal-failure-if-any-loop-cannot-be-completely-unrolled",
+      llvm::cl::desc(
+          "Signal failure if pass can't unroll all loops. (default: true)"),
+      llvm::cl::init(true)};
```
schweitzpgi, line 117: I don't think this option is well-placed. If loops can't be unrolled, they can't be unrolled. That's not an error.

This error condition should be checked in codegen when the assertion "loops are not allowed in this code" is known to be True or False.

Let's not repeat this "feature" in the new pass.

lib/Optimizer/Transforms/LoopUnroll.cpp
```diff
@@ -9,216 +9,24 @@
...
 #include "cudaq/Optimizer/Transforms/Passes.h"
+#include "mlir/IR/Dominance.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
+#include "mlir/Transforms/RegionUtils.h"
 
 namespace cudaq::opt {
 #define GEN_PASS_DEF_LOOPUNROLL
-#define GEN_PASS_DEF_UPDATEREGISTERNAMES
```
schweitzpgi, line 18: +1 for splitting this pass out to its own file.

lib/Optimizer/Transforms/LowerToCFGPatterns.inc
```diff
@@ -0,0 +1,104 @@
...
+        LLVM_DEBUG(llvm::dbgs() << "replacing " << *terminator << '\n');
+        rewriter.replaceOpWithNewOp<cf::BranchOp>(cont, continueBlock,
+                                                  cont.getOperands());
+      }
+      // Other ad-hoc control flow in the region need not be rewritten.
+    }
+  }
+
+private:
+  bool rewriteOnlyIfConst;
```
schweitzpgi, line 102:
```suggestion
  bool rewriteOnlyIfConst = false;
```

lib/Optimizer/Transforms/LowerToCFGPatterns.inc
```diff
@@ -0,0 +1,104 @@
...
+// This file must be included after a `using namespace mlir;` as it uses bare
+// identifiers from that namespace.
+
+namespace {
+class RewriteIf : public OpRewritePattern<cudaq::cc::IfOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  explicit RewriteIf(MLIRContext *ctx)
+      : OpRewritePattern(ctx), rewriteOnlyIfConst(false) {}
```
schweitzpgi, line 19: I don't think you need this if you add a default initializer as suggested below?

test/Quake/classical_optimization.qke
```diff
@@ -0,0 +1,255 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -classical-optimization -cse -classical-optimization %s | FileCheck %s
+// RUN: cudaq-opt -classical-optimization-pipeline %s | FileCheck %s
```
schweitzpgi, line 10: Are we expecting these two pass pipelines to always yield identical results?

schweitzpgi (commented):
This looks great. Just a few minor things...

include/cudaq/Optimizer/Transforms/Passes.td
annagrin: sounds good, will do

schweitzpgi (approved):

PR #2607, opened 2025-02-11
Enable serialized python code execution

sacpis:
Enable serialized python code execution (optimizer.optimize) for the paid-tier.

Tested it locally using remote-mqpu target.

Tests include
1. One level kernel
2. Nested kernels
3. With/without positional arguments to the optimizer.optimize function
4. Programs without optimizer.optimize calls


bmhowe23:
> Need to figure out a way to start nvqc_prozy.py script on the server (like we are doing for cudaq-qpud process).

I believe that is automatically handled when you use `docker/release/cudaq.nvqc.Dockerfile`.

tools/cudaq-qpud/nvqc_proxy.py
```diff
@@ -32,7 +32,7 @@
 WATCHDOG_TIMEOUT_SEC = 0
 RUN_AS_NOBODY = False  # Expect this to be overridden to true for NVQC deployment
 SUDO_FOUND = False
-CUDAQ_SER_CODE_EXEC = False
+CUDAQ_SER_CODE_EXEC = True
```
bmhowe23, line 35: Do we want to make this the default? It is currently "opt in", where you must set an environment variable in the deployment in order to activate this (see line 270). If we make this change, then it will become "opt out".

sacpis:
Not needed as we are not supporting NVQC anymore.

Issue #2000, opened 2024-07-24
[refactoring] [Python] Replace `MidCircuitMeasurementAnalyzer` with existing `quake-add-metadata` pass


khalatepradnya:
>  This analysis pass may actually be a relic from the original implementation - originally we had a library mode for Python and this pass was necessary to inform the runtime that the kernel had mid circuit measurement. We can probably drop it now for the MLIR pass. But you will have to update `sample.py` to check the Module attribute instead of the python kernel metadata attribute.

_Originally posted by @amccaskey in https://github.com/NVIDIA/cuda-quantum/pull/1994#discussion_r1688797509_

khalatepradnya:
Started implementation here - https://github.com/NVIDIA/cuda-quantum/pull/1994/commits/59ec10d33126069a80c3f214ac73fadc32c79547

(Remaining work - analyze failing tests related to `State`)

PR #2610, opened 2025-02-12
 [Python] [refactoring] Replace `MidCircuitMeasurementAnalyzer` with existing `quake-add-metadata` pass

khalatepradnya:
TODO in a follow-up: Function inlining.

Addresses issue https://github.com/NVIDIA/cuda-quantum/issues/2000


python/tests/kernel/test_explicit_measurements.py
```diff
@@ -280,7 +280,7 @@ def kernel_with_conditional_on_measure():
         e)
 
     ## NOTE: The following does not fail.
-    ## See: https://github.com/NVIDIA/cuda-quantum/issues/2000
+    ## Needs inlining of the function calls.
```
khalatepradnya, line 283: Note that adding inlining pass -
` "builtin.module(canonicalize,cse,inline,func.func(quake-add-metadata))"` will work for this case, but, causes `python/tests/kernel/test_sample_kernel.py::test_simple_sampling_qpe` to crash. Hence, updating the reason for failure here and propose to work on it in a follow-up PR.

python/cudaq/runtime/sample.py
```diff
@@ -66,8 +69,17 @@ def sample(kernel,
...
+
+    if isinstance(kernel, PyKernelDecorator):
+        kernel.compile()
+        if kernel.module is not None:
+            for operation in kernel.module.body.operations:
+                if not hasattr(operation, 'name'):
+                    continue
+                if nvqppPrefix + kernel.name == operation.name.value and 'qubitMeasurementFeedback' in operation.attributes:
+                    has_conditionals_on_measure_result = True
+                    break
```
1tnguyen, line 81: nit: I think we can always break here regardless of the `and` condition, e.g.,
```suggestion
                if nvqppPrefix + kernel.name == operation.name.value:
                    has_conditionals_on_measure_result = 'qubitMeasurementFeedback' in operation.attributes
                    break
```

1tnguyen (approved):
LGTM 👍

PR #2612, opened 2025-02-12
[core] Add call with vararg to CC dialect.

schweitzpgi:
This op allows CC code to call a variadic function without having to prematurely convert all the arguments to LLVM dialect and clutter the code with casts. The LLVMIR dialect's LLVMFuncOp does allow us to declare a variadic function, however.

Use the new Op in the QIR codegen to prevent the LLVMIR verifier from whining about the use of LLVM types.


1tnguyen (approved):
👍

PR #2613, opened 2025-02-13
cuda_torchdiffeq_integrator.py: check if pytorch is built with cuda capabilities

mitchdz:
Check if the installed pytorch is cuda aware.

This seems to only be an issue in the field on `aarch64` systems where the pytorch in the default pip repository is not cuda aware.

This is a code check of the warning described in https://nvidia.github.io/cuda-quantum/latest/using/backends/dynamics.html#numerical-integrators.


1tnguyen (approved):
LGTM 👍

mitchdz:
Small change to fix formatting issue :)

mitchdz:
\ok to test

PR #2615, opened 2025-02-13
[python] Implement ast.[In, NotIn] comparator, fix 2611

### LINKED ISSUE #2611: Python bridge error "IndexError: pop from an empty deque"
This code results in the above error

```
import cudaq

@cudaq.kernel
def kernel(ind: int):
    q = cudaq.qubit()
    if ind in [6, 13, 20, 27, 34]:
        x(q)    

counts = cudaq.sample(kernel, 1)
```

amccaskey:
Resolves #2611

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -2014,6 +2014,41 @@ def test_numpy_functions():
     with pytest.raises(RuntimeError):
         cudaq.sample(invalid_unsupported)
 
+def test_in_comparator():
+
+    @cudaq.kernel
+    def kernel(ind: int):
+        q = cudaq.qubit()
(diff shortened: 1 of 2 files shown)


python/tests/kernel/test_kernel_features.py
```diff
@@ -1995,6 +1995,31 @@ def invalid_unsupported():
     with pytest.raises(RuntimeError):
         cudaq.sample(invalid_unsupported)
 
+def test_in_comparator():
```
1tnguyen, line 1998: nit: Could we also throw in a simple `not in`  test here?

amccaskey: Thank you for catching that, I'll add it right now

1tnguyen (approved):
LGTM 👍 Thanks, Alex!

PR #2618, opened 2025-02-13
[core] fix a couple of bugs with loop normalization and unrolling

schweitzpgi:
This fixes some possible bugs with input code that uses extreme values, tests that are always true or always false, etc.

Generally speaking, the compiler isn't expecting infinite loops in CUDA-Q kernels. Infinite loops do not describe a circuit. Such language ought to be added to the CUDA-Q language specification.

Loops that are trivially dead are uninteresting but do not present a problem to the compiler since they can be trivially erased. This patch detects such trivially dead loops and short circuits any further analysis or optimization. These loops are dead and will be erased by existing passes.


targettests/execution/uccsd.cpp
```diff
@@ -8,7 +8,7 @@
 
 // clang-format off
 // RUN: nvq++ %cpp_std --target anyon --emulate %s -o %t && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target braket --emulate %s -o %t && %t | FileCheck %s
+// XUN: if %braket_avail; then nvq++ %cpp_std --target braket --emulate %s -o %t && %t | FileCheck %s ; fi
```
annagrin, line 11: It this test failing?

annagrin (approved):

targettests/execution/uccsd.cpp
schweitzpgi: Not sure. There is some tensornet test failing.

PR #2624, opened 2025-02-16
Add Stim support for sdg gate (required for `my` measurements)

### LINKED ISSUE #2265: Need better error message: stim target cannot handle toffoli gate
### Required prerequisites

### Describe the bug

stim target cannot handle toffoli gate

main.cpp
```cpp
#include <iostream>
#include <cudaq.h>


__qpu__ auto kernel() {
  cudaq::qvector q(3);
  x<cudaq::ctrl>(q[0], q[1], q[2]);
    
  auto result = mz(q);
  return result;
}


int main() {
    auto result = kernel();
    for (bool r: result) std::cout << r;
}
```

`nvq++ main.cpp && ./a.out`  works well, but `nvq++ main.cpp --target stim  && ./a.out` raises an error 
```
terminate called after throwing an instance of 'std::exception'
  what():  std::exception
Aborted (core dumped)
```

### Expected behavior

Output 000

Not a regression

### Environment

- **CUDA Quantum version**: latest (https://github.com/NVIDIA/cuda-quantum b0396b6b75cf3dfcf72515b6d0c249a8085f44c0)
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

bmhowe23:
We convert `my` measurements to a sequence of `sdg` + `h` + `mz` operations, so we need to support `sdg` gates. (We actually need to support `sdg` gates independent of `my` measurements; this is just a motivating use case to need to support them.)

* Also produce better error messages if gates are unsupported (fixes #2265)

### DIFF
--- a/python/tests/backends/test_stim.py
+++ b/python/tests/backends/test_stim.py
@@ -32,6 +32,7 @@ def test_stim_non_clifford():
     with pytest.raises(RuntimeError) as e:
         # Cannot perform non-Clifford gates in Stim simulator
         cudaq.sample(kernel)
+    assert 'Gate not supported by Stim simulator' in repr(e)
 
 
 def test_stim_toffoli_gates():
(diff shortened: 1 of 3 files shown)


1tnguyen (approved):
LGTM 👍

PR #2640, opened 2025-02-19
Add pointer element type constraints.

schweitzpgi:
This is a fragment from another PR.

Adding this separately so it can be used by more PRs.


annagrin (approved):
Thank Eric!

PR #2644, opened 2025-02-20
Make synthesis transitive

annagrin:
### Description
Make synthesis transitive

Allow substitutions in subst modules:
```
func.func @testy5(%arg0: i32) -> i32 {
  return %arg0: i32
}
```
Substitutions:

testy5:
```
cc.arg_subst[0] {
  %0 = arith.constant 2 : i32
}
func.func private @callee5(%arg0: i32) -> (i32) {
  return %arg0: i32
}
```

callee5:
```
cc.arg_subst[0] {
  %c4_i64 = arith.constant 4 : i32
}
```

Result:
```
func.func @testy5() -> i32 {
  %c2_i32 = arith.constant 2 : i32
  return %c2_i32 : i32
}
func.func private @callee5() -> i32 {
  %c4_i32 = arith.constant 4 : i32
  return %c4_i32 : i32
}
```


lib/Optimizer/Transforms/ArgumentSynthesis.cpp
```diff
@@ -14,6 +14,7 @@
 #include "mlir/Parser/Parser.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
+#include <list>
```
schweitzpgi, line 17: nit: Let's use `SmallVector`

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
```diff
@@ -147,6 +114,61 @@ class ArgumentSynthesisPass
...
+        return parseSourceFile<ModuleOp>(text, ctx);
+      }();
+      assert(*substModule && "module must have been created");
+
+      auto &name = funcNames.emplace_back(funcName);
+      substModules.try_emplace(name, std::move(substModule));
+    }
+
+    // 2. Merge symbols from substitution modules into the source module.
+    for (auto &[funcName, substMod] : substModules) {
```
schweitzpgi, line 157: The substitution module should be applied to the source module atomically as a single unit.

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
```diff
@@ -147,6 +114,61 @@ class ArgumentSynthesisPass
...
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
+    MLIRContext *ctx = mod.getContext();
+
+    // 1. Collect all substitution modules.
+    std::list<std::string> funcNames;
+    DenseMap<StringRef, OwningOpRef<ModuleOp>> substModules;
+
+    for (auto &item : funcList) {
+      auto pos = item.find(':');
```
schweitzpgi, line 127: I think the original code with this loop should be sufficient. We already have a list of function names and substitutions provided in `funcList`, so we don't have to build copies. It should be entirely sufficient to process the `funcList` from front to back and apply the substitution text to the source iteratively. Also, we don't want to change the algorithm's fail-safes. Specifically, if a susbt is for some function that does not exist, the subst should not be applied in whole or in part.

In other words, we walk the `funcList`, lookup the function it names, if it is found, instantiate the substitution module, walk that module and apply the argument substitutions and add sundry globals. Step to the next item in `funcList` and do it again until the list is exhausted.

If there is a problem with that, then I believe the substitutions are improperly formed. That is, when a substitution module is applied (in an atomic manner)  the new (intermediate) source module ought to be legal (but not 100% optimized) IR. So if we have two substitutions and the first one has references to functions that are added in the second one, then the substitutions themselves are incorrect. On the other hand, if the first subst is applied and adds a new function, then the second subst may refer to the newly added function on the next loop iteration. Looking up the new function in the source module must pass since it was just added (or there are catastrophic problems).

schweitzpgi (commented):
Thanks for splitting this out.

test/Quake/arg_subst_func.qke
```diff
@@ -146,3 +145,16 @@ func.func @testy4(%arg0: !cc.stdvec<!cc.struct<{i32, f64, i8, i16}>>) {
...
+
+func.func @testy5(%arg0: i32) -> i32 {
+  return %arg0: i32
+}
+
+// CHECK-LABEL:   func.func @testy5() -> i32 {
+// CHECK:           %[[VAL_0:.*]] = arith.constant 2 : i32
+// CHECK:           return %[[VAL_0]] : i32
+// CHECK:         }
+// CHECK:         func.func private @callee5() -> i32 {
```
schweitzpgi, line 157:
```suggestion
// CHECK-LABEL:         func.func private @callee5() -> i32 {
```

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
```diff
@@ -31,121 +31,120 @@ class ArgumentSynthesisPass
...
+      if (pos == std::string::npos)
+        continue;
 
-    // If there are no substitutions, we're done.
-    if (text.empty()) {
-      LLVM_DEBUG(llvm::dbgs() << funcName << " has no substitutions.");
-      return;
-    }
+      std::string funcName = item.substr(0, pos);
```
schweitzpgi, line 40:
```suggestion
      StringRef funcName = item.substr(0, pos);
```
Don't think we need to make a copy of this string?

schweitzpgi (commented):
Looking good. A couple of minor nits.

schweitzpgi (approved):

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
annagrin: removed

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
annagrin: removed

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
annagrin: done. thanks!

lib/Optimizer/Transforms/ArgumentSynthesis.cpp
annagrin: This does not compile - a temp object is destroyed.

PR #2481, opened 2024-12-16
Reorganized Backends Docs Page

mawolf2023:
### Description
This PR cleans up the backends section of the docs and makes it easier to follow.  This includes new figures and flowcharts to guide users to selecting the proper backend for their purposes.

PR #2466, opened 2024-12-11
Enable trajectory simulation for the `nvidia` target

1tnguyen:
### Description

- Bump mgpu sha to include:

    - Fixes for https://github.com/NVIDIA/cuda-quantum/issues/2434
    
    - Trajectory noisy simulation.
    
 - Follow-up changes of https://github.com/NVIDIA/cuda-quantum/pull/2456 for Python: allow passing number of trajectories to Python `observe` as well.
 
 - Adding docs for trajectory simulation.

PR #2648, opened 2025-02-21
[docs] Restore the docs for trajectory noisy simulators 

khalatepradnya:
Ported from PR #2466

Follow-up to PR #2481


1tnguyen (approved):
Thanks @khalatepradnya!

PR #2649, opened 2025-02-21
Add cudaq-stim-target to CUDAQConfig.cmake

bmhowe23:
This is useful for downstream C++ library developers. All other targets had targets defined except Stim. This adds a target for Stim, too.

This will allow us to fix a FIXME in CUDA-QX: https://github.com/NVIDIA/cudaqx/blob/fe9d088e8987282380fe8049adc416e5bae43ff0/libs/qec/unittests/backend-specific/stim/CMakeLists.txt#L9-L14


khalatepradnya (approved):
LGTM 👍🏽

PR #2650, opened 2025-02-22
Fix for nightly tests.

schweitzpgi:
Add output_name map when generating adaptive profile.


khalatepradnya (approved):
Manually tested with a handful of anyon integration tests. 
LGTM. 👍🏽

PR #2652, opened 2025-02-22
Refactor noise model for unitary mixtures

bmhowe23:
In support of the upcoming simulator work that is related to https://github.com/NVIDIA/cuda-quantum/pull/2635, this PR performs some refactoring that will enable easier support of the additional incoming noise models.

The basic idea of this PR is to move some logic that was only in the `tensornet` simulators and move it into the common `NoiseModel` section of code. There were a few minor changes as a result of those moves. For review purposes, those important changes are isolated to 06c8cf203d39948efbbb3fa1bac81589a2244936. (It is still worth reviewing the other commits, too.)

Note: despite this being related to #2635, this can go in independently of that PR.

Note, since the per-PR CI does not test tensornet, I ran these local tests to verify correctness:
```bash
$ ctest -R NoiseTest
Test project /workspaces/cuda-quantum/build
      Start 183: dm_NoiseTest.checkSimple
 1/63 Test #183: dm_NoiseTest.checkSimple ........................................   Passed    0.05 sec
      Start 184: dm_NoiseTest.checkAmplitudeDamping
 2/63 Test #184: dm_NoiseTest.checkAmplitudeDamping ..............................   Passed    0.04 sec
      Start 185: dm_NoiseTest.checkCNOT
 3/63 Test #185: dm_NoiseTest.checkCNOT ..........................................   Passed    0.04 sec
      Start 186: dm_NoiseTest.checkExceptions
 4/63 Test #186: dm_NoiseTest.checkExceptions ....................................   Passed    0.03 sec
      Start 187: dm_NoiseTest.checkDepolType
 5/63 Test #187: dm_NoiseTest.checkDepolType .....................................   Passed    0.05 sec
      Start 188: dm_NoiseTest.checkDepolTypeSimple
 6/63 Test #188: dm_NoiseTest.checkDepolTypeSimple ...............................   Passed    0.05 sec
      Start 189: dm_NoiseTest.checkAmpDampType
 7/63 Test #189: dm_NoiseTest.checkAmpDampType ...................................   Passed    0.04 sec
      Start 190: dm_NoiseTest.checkAmpDampTypeSimple
 8/63 Test #190: dm_NoiseTest.checkAmpDampTypeSimple .............................   Passed    0.04 sec
      Start 191: dm_NoiseTest.checkBitFlipType
 9/63 Test #191: dm_NoiseTest.checkBitFlipType ...................................   Passed    0.03 sec
      Start 192: dm_NoiseTest.checkBitFlipTypeSimple
[...]

PR #2635, opened 2025-02-18
Implementation of the cudaq::apply_noise feature

schweitzpgi:
This is the initial implementation of the apply_noise feature. Apply noise is a simulation-only feature and has no semantics for running on hardware.

This includes:

- C++ headers and frontend support
- Python support
- Core support
- Ability to erase "noise" from the IR (for hardware targets)
- Code generation as callbacks to the target simulator

TODOs:

 - Add more regression tests.
 - Hook this implementation up to the simulators.

PR #2653, opened 2025-02-23
[DRAFT] Implement new noise models

bmhowe23:
This depends on #2635 and #2652, so it currently includes all of those changes in this PR. Those changes will be removed once those other PRs are merged to main. This is a draft now, so I can start exercising the CI on it.

Most of the relevant changes are the implementations of the new noise models in `runtime/common/NoiseModel.h`. While this PR contains Stim-specific updates, most of the other simulators will not need to be updated once #2652 is merged.

### DIFF
--- a/python/runtime/common/py_NoiseModel.cpp
+++ b/python/runtime/common/py_NoiseModel.cpp
@@ -25,7 +25,8 @@ void extractKrausData(py::buffer_info &info, complex *data) {
         "Incompatible buffer format, must be np.complex128.");
 
   if (info.ndim != 2)
-    throw std::runtime_error("Incompatible buffer shape.");
+    throw std::runtime_error("Incompatible buffer shape " +
+                             std::to_string(info.ndim) + ".");
 
   constexpr bool rowMajor = true;
   typedef Eigen::MatrixXcd::Scalar Scalar;
@@ -56,6 +57,17 @@ void bindNoiseModel(py::module &mod) {
       "The `NoiseModel` defines a set of :class:`KrausChannel`'s applied to "
       "specific qubits after the invocation of specified quantum operations.")
       .def(py::init<>(), "Construct an empty noise model.")
+      .def(
+          "register_channel",
+          [](noise_model &self, const py::type krausT) {
+            auto key = py::hash(krausT);
+            self.register_channel(
+                key,
+                [krausT](const std::vector<double> &p) -> kraus_channel {
+                  return krausT(p).cast<kraus_channel>();
+                });
+          },
+          "")
       .def(
           "add_channel",
           [](noise_model &self, std::string &opName,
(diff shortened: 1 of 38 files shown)


python/runtime/common/py_NoiseModel.cpp
```diff
@@ -56,7 +56,42 @@ void bindNoiseModel(py::module &mod) {
...
+               if (py::hasattr(mod, name.c_str())) {
+                 py::type channelType = py::getattr(mod, name.c_str());
+                 auto key = py::hash(channelType);
+                 std::function<kraus_channel(const std::vector<double> &)>
+                     lambda =
+                         [channelType](
+                             const std::vector<double> &p) -> kraus_channel {
+                   return channelType(p).cast<kraus_channel>();
+                 };
+                 model->register_channel(key, lambda);
```
bmhowe23, line 88: @1tnguyen - this line (which pre-registers built-in noise channels) seems to make `cudaq.evolve` break ... or more specifically, this single test: `python/tests/operator/test_evolve_simulators.py::test_evolve_async`. E.g. https://github.com/NVIDIA/cuda-quantum/actions/runs/13552890652/job/37882046815?pr=2653#step:7:970. Does this ring any bells for you?

1tnguyen: I guess that this has broken something here this function https://github.com/NVIDIA/cuda-quantum/blob/7775089d4b1f8d3a08dfc942cb7eca9a6371da66/python/runtime/cudaq/algorithms/py_evolve.cpp#L127 
Looking at the comment there, perhaps we've dealt with similar issues before.

bmhowe23: OK, thanks. Working around this with ba2ad858e74e1c11241d292be11accdad0b30185

bmhowe23: @amccaskey - let me know your thoughts on this workaround.

python/runtime/common/py_NoiseModel.cpp
```diff
@@ -57,6 +57,39 @@ void bindNoiseModel(py::module &mod) {
       "The `NoiseModel` defines a set of :class:`KrausChannel`'s applied to "
       "specific qubits after the invocation of specified quantum operations.")
       .def(py::init<>(), "Construct an empty noise model.")
+      .def(
+          "register_builtins",
+          [mod](noise_model &self) {
+            // Get the built-in channel types from the module
+            std::vector<std::string> channelNames = {"DepolarizationChannel",
```
1tnguyen, line 64: nit: perhaps adding `static const`?

python/runtime/common/py_NoiseModel.cpp
```diff
@@ -57,6 +57,39 @@ void bindNoiseModel(py::module &mod) {
       "The `NoiseModel` defines a set of :class:`KrausChannel`'s applied to "
       "specific qubits after the invocation of specified quantum operations.")
       .def(py::init<>(), "Construct an empty noise model.")
+      .def(
+          "register_builtins",
```
amccaskey, line 61: we don't want to require that anyone has to manually register the builtin channels.

python/runtime/common/py_NoiseModel.cpp
```diff
@@ -289,9 +334,92 @@ void bindNoiseChannels(py::module &mod) {
...
+      .def(py::init<double>(), py::arg("probability"),
+           "Initialize the `PhaseFlipChannel` with the provided `probability`.")
+      .def_readonly_static(
+          "num_parameters", &phase_flip_channel::num_parameters,
+          "The number of parameters this channel requires at construction.");
+
+  py::class_<phase_damping, kraus_channel>(
+      mod, "PhaseDamping",
+      R"#(A Kraus channel that models the single-qubit phase damping error. This
+      is simulator to AmplitudeDamping, but for phase.)#")
```
1tnguyen, line 347:
```suggestion
      is similar to AmplitudeDamping, but for phase.)#")
```

python/runtime/common/py_NoiseModel.cpp
bmhowe23: > we don't want to require that anyone has to manually register the builtin channels.

I'm concerned that we may not be able to quickly resolve the `evolve` issue that Thien pointed out. (Maybe I'm missing something, though?) Would another viable workaround be pulling the the `noise.register_builtins()` call into all the Python `sample`, `sample_async`, `observe`, `observe_async`  calls (if noise was enabled)?

python/runtime/common/py_NoiseModel.cpp
```diff
@@ -289,9 +334,92 @@ void bindNoiseChannels(py::module &mod) {
...
+      occurs. It is the same as BitFlipChannel.)#")
+      .def(py::init<std::vector<double>>())
+      .def(py::init<double>())
+      .def_readonly_static(
+          "num_parameters", &x_error::num_parameters,
+          "The number of parameters this channel requires at construction.");
+
+  py::class_<y_error, kraus_channel>(
+      mod, "YError",
+      R"#(A Pauli error that applies the X operator when an error
```
1tnguyen, line 376:
```suggestion
      R"#(A Pauli error that applies the Y operator when an error
```

runtime/common/NoiseModel.h
```diff
@@ -597,4 +633,268 @@ class phase_flip_channel : public kraus_channel {
...
+
+  /// @brief Construct a single-qubit Pauli error Kraus channel
+  /// @param p Error probabilities for X, Y, and Z errors, respectively
+  pauli1(const std::vector<cudaq::real> &p) {
+    if (p.size() != num_parameters)
+      throw std::runtime_error(
+          "Invalid number of elements to pauli1 constructor. Must be 3.");
+    cudaq::real sum = 0;
+    for (auto pp : p)
+      sum += pp;
```
1tnguyen, line 740: nit: we could add a check for non-negative here as well.

1tnguyen (approved):
Only a few minor comments, otherwise, LGTM 👍

python/runtime/common/py_NoiseModel.cpp
bmhowe23: Updated in 958ebfd87

python/runtime/common/py_NoiseModel.cpp
bmhowe23: Updated in 958ebfd87

python/runtime/common/py_NoiseModel.cpp
bmhowe23: Updated in 958ebfd87

runtime/common/NoiseModel.h
bmhowe23: Updated in 958ebfd87

python/runtime/common/py_NoiseModel.cpp
amccaskey: yes that would be better and then revisit after march release

python/runtime/common/py_NoiseModel.cpp
bmhowe23: Resolved in 7b0053305. Thanks, Alex.

python/runtime/common/py_NoiseModel.cpp
bmhowe23: Resolved in 7b0053305. Thanks, Alex.

PR #2656, opened 2025-02-24
Fix regression in nightly tests for IONQ

schweitzpgi:


khalatepradnya (approved):
LGTM. Thanks, Eric! 👍🏽
I manually tested this patch with the failing nightly integration tests for IonQ and they work as expected.

PR #2657, opened 2025-02-24
[build] Bump up CUDA version to 12.8 in build

khalatepradnya:
* Also use `cudart` v12.8 wherever applicable

### DIFF
--- a/docker/test/installer/runtime_dependencies.sh
+++ b/docker/test/installer/runtime_dependencies.sh
@@ -33,7 +33,7 @@ esac
 
 CUDA_DOWNLOAD_URL=https://developer.download.nvidia.com/compute/cuda/repos
 CUDA_ARCH_FOLDER=$([ "$(uname -m)" == "aarch64" ] && echo sbsa || echo x86_64)
-CUDA_VERSION_SUFFIX=$(echo ${CUDART_VERSION:-'12.0'} | tr . -)
+CUDA_VERSION_SUFFIX=$(echo ${CUDART_VERSION:-'12.8'} | tr . -)
 CUDA_PACKAGES=$(echo "cuda-cudart cuda-nvrtc libcusolver libcublas" | sed "s/[^ ]*/&-${CUDA_VERSION_SUFFIX} /g")
 if [ $(echo ${CUDART_VERSION} | cut -d . -f1) -gt 11 ]; then 
     CUDA_PACKAGES+=" libnvjitlink-${CUDA_VERSION_SUFFIX}"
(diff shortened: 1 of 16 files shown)


docker/test/installer/runtime_dependencies.sh
```diff
@@ -33,7 +33,7 @@ esac
 
 CUDA_DOWNLOAD_URL=https://developer.download.nvidia.com/compute/cuda/repos
 CUDA_ARCH_FOLDER=$([ "$(uname -m)" == "aarch64" ] && echo sbsa || echo x86_64)
-CUDA_VERSION_SUFFIX=$(echo ${CUDART_VERSION:-'12.0'} | tr . -)
+CUDA_VERSION_SUFFIX=$(echo ${CUDART_VERSION:-'12.8'} | tr . -)
```
bmhowe23, line 36: Others can correct me if I'm wrong because I'm definitely not sure about this, but we might want to keep this one at 12.0 to make sure that we can run on 12.0 despite building on 12.8.

pyproject.toml
```diff
@@ -22,10 +22,10 @@ dependencies = [
...
   'scipy >= 1.10.1',
   'requests >= 2.31',
-  'nvidia-cublas-cu12 ~= 12.0',
-  'nvidia-cuda-runtime-cu12 ~= 12.0',
-  'nvidia-cusolver-cu12 ~= 11.4',
-  'nvidia-cuda-nvrtc-cu12 ~= 12.0'
+  'nvidia-cublas-cu12 ~= 12.8',
+  'nvidia-cuda-runtime-cu12 ~= 12.8',
+  'nvidia-cusolver-cu12 ~= 11.7',
+  'nvidia-cuda-nvrtc-cu12 ~= 12.8'
```
bmhowe23, line 28: I think we want to keep these wheel dependencies the same as before so that users who are just using CUDA 12.0 runtime can still use our wheels.

khalatepradnya: Q: Should I make these `>=`?
I'm getting following error in Publishing -
```
2025-02-28T04:28:57.5222378Z Collecting nvidia-cublas-cu12~=12.0
2025-02-28T04:28:57.5279353Z   Downloading nvidia_cublas_cu12-12.8.3.14-py3-none-manylinux_2_27_x86_64.whl (609.6 MB)
2025-02-28T04:28:59.2682292Z      ━━━━━━━━━╸                            163.8/609.6 MB 163.7 MB/s eta 0:00:03
2025-02-28T04:28:59.8344767Z ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
```

PR #2658, opened 2025-02-25
Make use of apply_noise with client/server targets emit an error.

schweitzpgi:


bmhowe23 (approved):
This looks good for C++, but we should leave the item on the associated issue list to remind us we still need it for Python

PR #2660, opened 2025-02-25
Update docs for `nvidia` target fuse settings

1tnguyen:
### Description

- Adding docs for the dynamic fuse size selection based on Compute Capability feature of the `nvidia` target.


.github/workflows/config/gitlab_commits.txt
```diff
@@ -1,2 +1,2 @@
 nvidia-mgpu-repo: cuda-quantum/cuquantum-mgpu.git
-nvidia-mgpu-commit: 5a4c69fa1d5bc70fecccd758e5d936ff4b8115e3
+nvidia-mgpu-commit: 403d944a745a5b4e4faab49ed31c0963b9c07b27
```
bmhowe23, line 2: @1tnguyen please double check me here, but depending on the timing of merges, #2653 may make this SHA change OBE.

1tnguyen: Done. This is now a docs-only PR.

Issue #2545, opened 2025-01-25
cudaqx vqe/observe behavior is different on simulators and quantum device emulators


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following code behaves differently on simulators and quantum device emulators:

```python

import numpy as np
import cudaq, cudaq_solvers as solvers

from scipy.optimize import minimize


def test_solvers_uccsd():

    # fail
    cudaq.set_target('ionq', emulate="true")
    #cudaq.set_target('quantinuum', emulate="true")

    # pass
    #cudaq.set_target('nvidia')
    #cudaq.set_target('nvidia-fp64')
    #cudaq.set_target('qpp-cpu')

    geometry = [('H', (0., 0., 0.)), ('H', (0., 0., .7474))]
    molecule = solvers.create_molecule(geometry, 'sto-3g', 0, 0, casci=True)

    numQubits = molecule.n_orbitals * 2
    numElectrons = molecule.n_electrons
    spin = 0
    print(numQubits)
    print(numElectrons)

    @cudaq.kernel
    def ansatz(thetas: list[float]):
        q = cudaq.qvector(numQubits)
        for i in range(numElectrons):
            x(q[i])
        #solvers.stateprep.uccsd(q, thetas, numElectrons, spin)

    ansatz.compile()

    energy, params, all_data = solvers.vqe(ansatz,
                                           molecule.hamiltonian,
                                           [-.2, -.2, -.2],
                                           optimizer=minimize,
                                           method='L-BFGS-B',
                                           jac='3-point',
                                           tol=1e-4,
                                           options={'disp': True},
                                           shots=1000,
                                           verbose=True)
    print(energy)
    
    assert np.isclose(energy, -1.04, 1e-2)

test_solvers_uccsd()
```

### Steps to reproduce the bug

```python

import numpy as np
import cudaq, cudaq_solvers as solvers

from scipy.optimize import minimize


[...]

PR #2661, opened 2025-02-26
Handle missing cases in ObserveAnsatz

annagrin:
### Description
Handle missing cases in ObserveAnsatz

- Handles veqs without ExtractRefOps in ObserveAnsatz` pass
- Add tests

Towards: https://github.com/NVIDIA/cuda-quantum/issues/2545
Closes: https://github.com/NVIDIA/cuda-quantum/issues/2662

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -185,6 +185,17 @@ def test_2grover_compute_action():
     assert '011' in counts
 
 
+def test_observe():
+
+    @cudaq.kernel
+    def ansatz():
+        q = cudaq.qvector(1)
+
+    molecule = 5.0 - 1.0 * spin.x(0)
+    res = cudaq.observe(ansatz, molecule, shots_count=10000)
+    assert np.isclose(res.expectation(), 5.0, atol=1e-1)
+
+
 def test_pauli_word_input():
 
     h2_data = [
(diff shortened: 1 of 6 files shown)


schweitzpgi (approved):
LGTM. Thanks for the bug fix.

Adding @amccaskey as well in case he has any suggestions.

targettests/execution/cudaq_observe_qvector.cpp
bmhowe23, line 1: I'm curious what was different about this test vs the other test in the same directory (https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/execution/cudaq_observe.cpp)? Is it just the fact that it is a single qubit vs 2 qubits?

python/tests/kernel/test_kernel_features.py
```diff
@@ -185,6 +185,17 @@ def grover(N: int, M: int, oracle: Callable[[cudaq.qview], None]):
     assert '011' in counts
```
khalatepradnya, line 185: Please confirm that these changes work with `braket` target by manually running these two tests:
1. https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/backends/test_braket.py#L242
2. https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/backends/test_braket.py#L261

targettests/execution/cudaq_observe_qvector.cpp
annagrin: This test is different - there are no gates invoked on any qubits. The test also compares the result against the expectation - the incorrect result was ~4, the correct one is ~5 so it is easy to catch a regression.

python/tests/kernel/test_kernel_features.py
annagrin: done! it works.

Issue #2623, opened 2025-02-15
Deployment is failing on main


sacpis:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Deployment is failing with Segmentation fault on the main branch due to failure in test_mpi_api.py

### Steps to reproduce the bug

Here is the job
https://github.com/NVIDIA/cuda-quantum/actions/runs/13335693612/job/37254100357#step:8:6360

```
Fatal Python error: Segmentation fault

Current thread 0x0000e3b78e679800 (most recent call first):
  File "/tmp/tests/parallel/test_mpi_api.py", line 19 in testMPI
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/_pytest/python.py", line 1627 in runtest
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/_pytest/runner.py", line 174 in pytest_runtest_call
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/root/miniconda3/envs/cudaq-env/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
[...]

PR #2663, opened 2025-02-26
.github/workflows/python_wheels.yml: bump openmpi version for MPI test

mitchdz:
An older version (4.0.5) of openmpi was being installed in the conda environment for redhat. This change bumps the minimum version to 5.0.7 which is the current latest version.

See: https://github.com/NVIDIA/cuda-quantum/issues/2623

### DIFF
--- a/.github/workflows/python_wheels.yml
+++ b/.github/workflows/python_wheels.yml
@@ -296,6 +296,9 @@ jobs:
             while IFS= read -r line; do
               # Replace Python version
               line=$(echo $line | sed -E "s/python(=)?3.[0-9]{1,}/python\1${{ inputs.python_version }}/g")
+              # Replace openmpi version requirement - version 4.0.5 was being installed which was not built with
+              # cuda support, thus causing a seg fault. See https://github.com/NVIDIA/cuda-quantum/issues/2623
+              line=$(echo "$line" | sed -E 's/\bopenmpi\b/openmpi\\>=5.0.7/g')
               # Install the wheel file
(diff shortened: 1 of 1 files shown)


khalatepradnya (approved):
Thanks, @mitchdz !

bettinaheim (approved):

.github/workflows/python_wheels.yml
```diff
@@ -296,6 +296,9 @@ jobs:
             while IFS= read -r line; do
               # Replace Python version
               line=$(echo $line | sed -E "s/python(=)?3.[0-9]{1,}/python\1${{ inputs.python_version }}/g")
+              # Replace openmpi version requirement - version 4.0.5 was being installed which was not built with
+              # cuda support, thus causing a seg fault. See https://github.com/NVIDIA/cuda-quantum/issues/2623
+              line=$(echo "$line" | sed -E 's/\bopenmpi\b/openmpi\\>=5.0.7/g')
```
bmhowe23, line 301: I may be too late to this party since it was merged, but shouldn't we update python/README.md.in instead?

mitchdz: That is a great question.

I chose to do it this way, since the MPI test _only_ runs on redhat at the moment, and I was able to confirm the issue+fix it for redhat. It might be different for other distributions, so I don't think we should hardcode it for all distributions yet. Perhaps if we get a issue report from users about the same thing then we can instead update the README.

PR #2665, opened 2025-02-27
[build] [testing] Run the Quantinuum integration tests unconditionally

khalatepradnya:


.github/workflows/integration_tests.yml
```diff
@@ -387,7 +387,7 @@ jobs:
         shell: bash
 
       - name: Setup quantinum account
-        if: github.event_name == 'schedule' || inputs.target == 'nightly' || inputs.target == 'quantinuum'
+        if: (success() || failure()) && (github.event_name == 'schedule' || inputs.target == 'nightly' || inputs.target == 'quantinuum')
```
schweitzpgi, line 390: Is `success() || failure()` simply `true`?

bmhowe23: You'd think so, but GitHub actions always downstream items after a failure in a prior step. This trick allows it to run it even if prior steps failed. We do the same thing in other steps in this file.

bmhowe23 (approved):

khalatepradnya:
Manually setting DCO to pass since I created this PR from GH web UI and don't have means to sign off.

PR #2666, opened 2025-02-27
Update polling interval for Braket backend

rmshaffer:
### Description
Increase default polling interval for Braket to 2000ms, and introduce a `polling_interval_ms` argument to make it user-configurable.


rmshaffer:
Will recreate on a new branch.

PR #2670, opened 2025-02-28
[CI] Follow-up to PR#2665

khalatepradnya:
* Follow-up to PR https://github.com/NVIDIA/cuda-quantum/pull/2665
* The syntax check tests are still getting skipped if one of the previous steps failed. See [here](https://github.com/NVIDIA/cuda-quantum/actions/runs/13578480396/job/37960023854) 
* Run it unconditionally.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -386,15 +386,18 @@ jobs:
           fi
         shell: bash
 
-      - name: QIR syntax check (Quantinuum)
+      - name: Setup quantinum account
         if: (success() || failure()) && (github.event_name == 'schedule' || inputs.target == 'nightly' || inputs.target == 'quantinuum')
         run: |
-          echo '### Setting up Quantinum account'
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -386,18 +386,15 @@ jobs:
           fi
         shell: bash
 
-      - name: Setup quantinum account
+      - name: QIR syntax check (Quantinuum)
```
bmhowe23, line 389: I think the reason these were kept as seperate is so that we could - with manual invocation - run on more than just the syntax checker. I.e. run on their remote emulator or their actual QPU. That is invoked using this: https://github.com/NVIDIA/cuda-quantum/blob/d9d63291201cae2f2194e893a65a156c3ecc086e/.github/workflows/integration_tests.yml#L700

khalatepradnya: How about https://github.com/NVIDIA/cuda-quantum/pull/2670/commits/2ca0dbf59f48a5361d0ff3efd626d73d0cddc6f3?

bmhowe23 (approved):

PR #2672, opened 2025-02-28
Fixes for `NoiseModel` when using with `fp32` targets

1tnguyen:
### Description

- `parameters` was defined as a `vector<double>`; hence we need to cast if `cudaq::real` is `float`.

- `get_channel` using `params` type to deduct the variant type is not reliable; could lead to bad variant access. Hence, perform a `holds_alternative` check before calling `std::get`. If needed, perform a type casting to retrieve the correct variant type.


amccaskey (approved):

bmhowe23 (approved):
LGTM, although I'm a bit curious why the cusvsim fp32 target that I used for testing didn't show any problems in the local testing I did. If there is a set of tests I missed, please let me know for next time. Thanks!

1tnguyen:
> LGTM, although I'm a bit curious why the cusvsim fp32 target that I used for testing didn't show any problems in the local testing I did. If there is a set of tests I missed, please let me know for next time. Thanks!

Actually, it was the 'enhanced' `noise_tester.cpp` that caught this error (`checkFineGrainArg` and `checkFineGrainVec`). 

In the latest `mgpu` update, I added a CMake change to force all the test cases in that file to be executed on all downstream `mgpu` backends, so we'll have better coverage.

PR #2674, opened 2025-02-28
[orca] Skip unnecessary call and avoid error condition

khalatepradnya:
The `resetExecutionContext` was previously a no-op, except for resetting the context, since no 'qubits' are involved in this job.
  Recent changes check for presence of qubits and flag errors accordingly.
  This change removes the invocation of that function. We reset the context in next line anyway.

Should fix the nightly integration test failures for `orca` seen [here](https://github.com/NVIDIA/cuda-quantum/actions/runs/13582771746/job/37971715501).

Manually tested by locally running all the nightly integration tests.


bmhowe23 (approved):
Thanks, @khalatepradnya.

PR #2676, opened 2025-02-28
Fix a bug in ObserveAnsatz when handling spin.y term

annagrin:
### Description
Use rx(pi/2) to change the basis for measurement of `spin.y`

This issue was causing different results when observing `spin.y` on simulators vs quantum devices.
Added tests.
  
Closes: https://github.com/NVIDIA/cuda-quantum/issues/2545


schweitzpgi (commented):
LGTM

bmhowe23 (approved):
Nice find, @annagrin. It looks like this bug goes all the way back to the original GitHub commit: https://github.com/NVIDIA/cuda-quantum/blob/8fa97f59b576ebf4ebca732850d98ddb6fe0a384/lib/Optimizer/Transforms/QuakeObserveAnsatz.cpp#L40

targettests/execution/cudaq_observe.cpp
```diff
@@ -45,7 +45,7 @@ int main() {
 
   // Observe takes the kernel, the spin_op, and the concrete
   // parameters for the kernel
-  double energy = cudaq::observe(ansatz{}, h, .59);
+  double energy = cudaq::observe(10000, ansatz{}, h, .59);
```
khalatepradnya, line 48: This is causing IQM integration test to fail since it supports max 1024 shots.
```2025-03-03T03:48:14.7274321Z terminate called after throwing an instance of 'std::runtime_error'
2025-03-03T03:48:14.7275485Z   what():  HTTP POST Error - status code 403: : {"message":"Cannot submit job with 10000 shots, the maximum is 1024","request_id":"00bbd640-9b4b-4079-95f2-057f18de5f84"}
2025-03-03T03:48:14.9499642Z /__w/_temp/e4591dac-7f4e-4137-bd0e-8e1b2726123a.sh: line 13: 12154 Aborted                 (core dumped) ./a.out
```

annagrin: done! https://github.com/NVIDIA/cuda-quantum/pull/2685

PR #2679, opened 2025-03-01
[core] Add checks to the CC dialect types.

schweitzpgi:
We do not allow CC dialect types to "contain" quantum types.

Fix some issues in the C++ bridge that were resulting in these invalid hybrid types being created.


annagrin (approved):
Thanks Eric!

PR #2683, opened 2025-03-03
Dynamics cpp integration

sacpis:
Dynamics cpp integration

co-authored by: @1tnguyen


bettinaheim (commented):
still going through it, but some comments while reading

runtime/cudaq/algorithms/evolve.h
```diff
@@ -10,146 +10,246 @@
...
-  for (auto observable : observables) {
-    shots_count <= 0
-        ? final_expectations.push_back(observe(prepare_state, observable))
-        : final_expectations.push_back(
-              observe(shots_count, prepare_state, observable));
+template <typename OpTy>
+cudaq::operator_sum<cudaq::matrix_operator> convertOp(const OpTy &op) {
+  if constexpr (std::is_convertible_v<
+                    OpTy, cudaq::product_operator<cudaq::matrix_operator>>) {
+    cudaq::operator_sum<cudaq::matrix_operator> convertedOp(op);
```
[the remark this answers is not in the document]
1tnguyen: Yea, in one of our test cases, we called this with `cudaq::spin_operator::z(0)` (type `product_operator<spin_operator>`), and it worked.

runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
```diff
@@ -0,0 +1,90 @@
...
+  return m_scratchSpace;
+}
+
+/// @brief Get the recommended workspace limit based on available memory.
+/// @return std::size_t Recommended workspace limit in bytes.
+std::size_t Context::getRecommendedWorkSpaceLimit() {
+  std::size_t freeMem = 0, totalMem = 0;
+  HANDLE_CUDA_ERROR(cudaMemGetInfo(&freeMem, &totalMem));
+  // Take 80% of free memory
+  freeMem = static_cast<std::size_t>(static_cast<double>(freeMem) * 0.80);
```
bmhowe23, line 68: Is this "capture 80% of free memory" described in the docs anywhere? That could be confusing for users.

runtime/cudaq/utils/tensor.h
```diff
@@ -105,6 +106,9 @@ class matrix_2 {
   /// Matrix power.
   matrix_2 power(int powers);
 
+  /// Returns the conjugate transpose of a matrix.
+  static matrix_2 adjoint(const matrix_2 &matrix);
```
bmhowe23, line 110: Why have this as a static method since it is clearly operating on an existing `matrix_2`?

unittests/CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ set(CUDAQ_RUNTIME_TEST_SOURCES
 # Make it so we can get function symbols
 set (CMAKE_ENABLE_EXPORTS TRUE)
 
+include_directories(/usr/local/cuda/targets/x86_64-linux/include)
```
bmhowe23, line 52: I'm curious how this works on arm machines?

runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
```diff
@@ -0,0 +1,90 @@
...
+  return freeMem;
+}
+
+/// @brief Construct a new Context object for a specific device.
+/// @arg deviceId ID of the CUDA device.
+Context::Context(int deviceId) : m_deviceId(deviceId) {
+  HANDLE_CUDA_ERROR(cudaSetDevice(deviceId));
+  HANDLE_CUDM_ERROR(cudensitymatCreate(&m_cudmHandle));
+  HANDLE_CUBLAS_ERROR(cublasCreate(&m_cublasHandle));
```
bmhowe23, line 77: This means that CUDA-Q is now introducing a direct dependency on cuBLAS now. This is a new dependency, so I just wanted to make sure I have that right. Is the assumption that our cublas dependency always going to be the same as custatevec's dependency?

bmhowe23 (commented):
I didn't have time to do a thorough review, but I have just a few comments.

runtime/cudaq/algorithms/evolve.h
```diff
@@ -64,6 +65,8 @@ convertOps(const std::initializer_list<OpTy> &ops) {
 template <typename HamTy,
           typename CollapseOpTy = cudaq::operator_sum<cudaq::matrix_operator>,
           typename ObserveOpTy = cudaq::operator_sum<cudaq::matrix_operator>>
+  requires cudaq::operator_type<HamTy> && cudaq::operator_type<CollapseOpTy> &&
```
amccaskey, line 68: since we have to work with C++17 and C++20, this will need to have an equivalent SFINAE version instead of constraints (which only work for C++20)

docs/sphinx/examples/cpp/dynamics/qubit_control.cpp
```diff
@@ -0,0 +1,134 @@
...
+  double omega_x = 2 * M_PI;
+  // Driving frequency, chosen to be slightly off-resonance (0.99 of omega_z).
+  double omega_drive = 0.99 * omega_z;
+
+  // The lambda function acts as a callback that returns a modulation factor for
+  // the drive. It extracts the time `t` from the provided parameters and
+  // computes cos(omega_drive * t).
+  auto mod_func =
+      [omega_drive](
+          const std::unordered_map<std::string, std::complex<double>> &params)
```
amccaskey, line 33: we'll want a `parameter_map` typedef per the operators specification

runtime/common/EvolveResult.h
```diff
@@ -100,20 +100,40 @@ class evolve_result {
...
+          "no intermediate states have been computed in the call to evolve");
+    }
   }
 
-  std::optional<std::vector<observe_result>> get_final_expectation_values() {
-    return final_expectation_values;
+  std::vector<observe_result> get_final_expectation_values() {
+    if (final_expectation_values.has_value()) {
+      return final_expectation_values.value();
+    } else {
```
amccaskey, line 115: else after return. this might go away though since the spec is moving toward simple struct with public std::optional values.

runtime/cudaq/BaseIntegrator.h
```diff
@@ -0,0 +1,76 @@
+/****************************************************************-*- C++ -*-****
```
amccaskey, line 1: with type name changed to base_integrator, would be best to have this file renamed to base_integrator.h

runtime/cudaq/algorithms/evolve_internal.h
```diff
@@ -0,0 +1,168 @@
...
+              observe(shots_count, prepare_state, observable));
+  }
+  return evolve_result(final_state, final_expectations);
+}
+
+/// @brief Evolve from an initial state to the final state and gather
+/// intermediate states.
+// Step evolution is provided as `kernels`.
+template <typename QuantumKernel>
+evolve_result evolve(state initial_state, std::vector<QuantumKernel> kernels,
```
amccaskey, line 55: is it intended that you only allow a single type of QuantumKernel for each call here to evolve? If so, should this be `const std::vector<QuantumKernel>& kernels`?

runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
1tnguyen: This assumes that we already have cuBLAS as it's a dependency of cudensitymat. 
It's working now, so I'm also not sure whether we need to list it explicitly in the link list.

bmhowe23: > It's working now

It seems like the installer tests may be failing ... I wonder if it's related. The installer environment is pretty sensitive to what is dynamically linked and what is statically linked.

1tnguyen: We added the missing test label in https://github.com/NVIDIA/cuda-quantum/pull/2683/commits/8367d9f89746ea831f763fba4a1aa0769f735023 to address the installer tests. 
I guess we need the Publishing validation pipeline to verify if it works in different environments. 
Or should we just pre-emptively add the direct linkage?

bmhowe23: OK, no need to add the direct linkage.

runtime/cudaq/algorithms/evolve.h
```diff
@@ -11,145 +11,242 @@
...
-             std::size_t qpu_id = 0,
-             std::optional<cudaq::noise_model> noise_model = std::nullopt,
-             int shots_count = -1) {
-  auto &platform = cudaq::get_platform();
-  std::promise<evolve_result> promise;
-  auto f = promise.get_future();
+template <typename HamTy,
+          typename CollapseOpTy = cudaq::operator_sum<cudaq::matrix_operator>,
+          typename ObserveOpTy = cudaq::operator_sum<cudaq::matrix_operator>,
+          typename = std::enable_if_t<cudaq::operator_type<HamTy> &&
```
amccaskey, line 68: might be reading this wrong - but make sure you `#ifdef` here on the C++ standard. You want this new enable_if_t if it is C++17, and the requires clause if it is C++20, but you do not want both

runtime/common/EvolveResult.h
```diff
@@ -100,20 +100,23 @@ class evolve_result {
     return final_state;
   }
 
-  std::optional<std::vector<state>> get_intermediate_states() {
+  const std::optional<std::vector<state>> &get_intermediate_states() {
```
amccaskey, line 103: The spec proposal is dropping these getters in favor of public std::optional members

docs/sphinx/examples/cpp/dynamics/qubit_control.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  // The Hamiltonian consists of two terms:
+  // 1. A static term: 0.5 * omega_z * `Sz`_0, representing the `qubit's`
+  // intrinsic energy splitting.
+  // 2. A time-dependent driving term: omega_x * cos(omega_drive * t) * `Sx`_0,
+  // which induces rotations about the X-axis. The scalar_operator(mod_`func`)
+  // allows the drive term to vary in time according to mod_`func`.
+  auto hamiltonian =
+      0.5 * omega_z * cudaq::spin_operator::z(0) +
+      omega_x * cudaq::scalar_operator(mod_func) * cudaq::spin_operator::x(0);
```
amccaskey, line 51: we should never see the typename `scalar_operator` in user code

python/extension/CMakeLists.txt
```diff
@@ -39,6 +39,9 @@ declare_mlir_dialect_python_bindings(
     dialects/cc.py
   DIALECT_NAME cc)
 
+# FIXME: remove this flag
```
bettinaheim, line 42: missed fixme? should be easy enough to fix the warnings, probably.

runtime/nvqir/cudensitymat/CuDensityMatState.cpp
```diff
@@ -0,0 +1,610 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include "CuDensityMatState.h"
+#include "CuDensityMatContext.h"
+#include "CuDensityMatErrorHandling.h"
```
schweitzpgi, line 10: Are the includes consistent? I think some used angle brackets, but I suppose those may have been part of a simulator package...

runtime/nvqir/cudensitymat/CuDensityMatSim.cpp
```diff
@@ -80,16 +81,18 @@ class CuDensityMatSim : public nvqir::CircuitSimulatorBase<double> {
   CuDensityMatSim() {
     int numDevices{0};
     HANDLE_CUDA_ERROR(cudaGetDeviceCount(&numDevices));
-    const int deviceId =
-        cudaq::mpi::is_initialized() ? cudaq::mpi::rank() % numDevices : 0;
+    int currentDevice = -1;
```
schweitzpgi, line 84: Is this magic number well-defined? Should really use an identifier so people can find it uses.

runtime/nvqir/cudensitymat/CuDensityMatOpConverter.cpp
```diff
@@ -0,0 +1,706 @@
...
+        parameters.begin(), parameters.end());
+    auto ks = std::views::keys(sortedParameters);
+    const std::vector<std::string> keys{ks.begin(), ks.end()};
+    wrappedTensorCallback = wrapTensorCallback(elemOp, keys);
+  }
+
+  auto flatMatrix =
+      flattenMatrixColumnMajor(elemOp.to_matrix(dimensions, parameters));
+
+  if (flatMatrix.empty()) {
```
schweitzpgi, line 277: unneeded braces

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "CuDensityMatExpectation.h"
+#include "CuDensityMatState.h"
+#include "CuDensityMatTimeStepper.h"
+#include "cudaq/algorithms/evolve_internal.h"
+#include "cudaq/dynamics_integrators.h"
+#include <iterator>
+#include <random>
+#include <stdexcept>
+namespace cudaq {
+namespace __internal__ {
```
schweitzpgi, line 20:
```suggestion
namespace cudaq::__internal__ {
```

runtime/nvqir/cudensitymat/CuDensityMatState.cpp
```diff
@@ -0,0 +1,610 @@
...
+  std::size_t storageSize{0}; // only one storage component (tensor) is needed
+  HANDLE_CUDM_ERROR(cudensitymatStateGetComponentStorageSize(
+      state.cudmHandle, state.cudmState,
+      1,              // only one storage component
+      &storageSize)); // storage size in bytes
+  const std::size_t stateVolume =
+      storageSize / sizeof(std::complex<double>); // quantum state tensor volume
+                                                  // (number of elements)
+  assert(stateVolume == state.dimension);
+  // std::cout << "Quantum state storage size (bytes) = " << storageSize
```
bettinaheim, line 400: Would be nice to remove commented out code left over from debugging

runtime/nvqir/cudensitymat/CuDensityMatContext.h
```diff
@@ -0,0 +1,63 @@
...
+ ******************************************************************************/
+
+#pragma once
+#include "CuDensityMatOpConverter.h"
+#include <cublas_v2.h>
+#include <cuda_runtime.h>
+#include <cudensitymat.h>
+
+namespace cudaq {
+namespace dynamics {
```
schweitzpgi, line 16:
```suggestion
namespace cudaq::dynamics {
```
reduce clutter and valid for all C++ standards we support.

runtime/nvqir/cudensitymat/CuDensityMatState.cpp
```diff
@@ -0,0 +1,610 @@
...
+
+CuDensityMatState &
+cudaq::CuDensityMatState::operator+=(const CuDensityMatState &other) {
+  if (dimension != other.dimension) {
+    throw std::invalid_argument(
+        fmt::format("State size mismatch for addition ({} vs {}).", dimension,
+                    other.dimension));
+  }
+
+  // double scalingFactor = 1.0;
```
bettinaheim, line 567: Please remove commented out code. It has no purpose and adds no value.

scripts/validate_container.sh
```diff
@@ -137,7 +137,7 @@ echo "============================="
 
 # Note: piping the `find` results through `sort` guarantees repeatable ordering.
 tmpFile=$(mktemp)
-for ex in `find examples/ applications/ targets/ -name '*.cpp' | sort`;
+for ex in `find examples/ applications/ targets/ -name '*.cpp' -not -path "*/dynamics/*" | sort`;
```
bettinaheim, line 140: Why are we excluding the dynamics examples etc. from the validation? 
Where are these validated?

runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
```diff
@@ -0,0 +1,90 @@
...
+    m_scratchSpaceSizeBytes = minSizeBytes;
+  }
+
+  return m_scratchSpace;
+}
+
+/// @brief Get the recommended workspace limit based on available memory.
+/// @return std::size_t Recommended workspace limit in bytes.
+std::size_t Context::getRecommendedWorkSpaceLimit() {
+  std::size_t freeMem = 0, totalMem = 0;
```
schweitzpgi, line 65:
```suggestion
  std::size_t freeMem = 0;
  std::size_t totalMem = 0;
```
Avoid K&R style declarations. :)

schweitzpgi (commented):
This is truly massive. It would've been better to split this up into 5, 10, or maybe 20 PRs.

runtime/cudaq/dynamics/operator_type.h
```diff
@@ -0,0 +1,31 @@
...
+template <typename HandlerTy>
+struct is_operator_type<product_operator<HandlerTy>> : std::true_type {};
+
+// operator_sum
+template <typename HandlerTy>
+struct is_operator_type<operator_sum<HandlerTy>> : std::true_type {};
+
+// Satisfied if T (after decay) is one of operator types.
+template <typename T>
+concept operator_type = is_operator_type<std::decay_t<T>>::value;
```
amccaskey, line 29: The concept ultimately used should be "boolean" in its name, by that I mean, the requires clause shouldn't call `operator_type` but instead call `is_operator_type`. 

Actually I think for the C++20 requires clause you could actually drop the `requires` and `typename` and directly use `operator_type` in the template parameter specification 
```cpp
template<operator_type HamTy, ...>
```
would look cleaner

scripts/validate_container.sh
bettinaheim: 1) anything that requires gpus is not validated during CI or deployment, only during publishing
2) publishing (Validate cuda-quantum image) is exactly calling into this script to do the validation. Hence, disabling it in this script disables that validation. I am not seeing these being validated anywhere. What am I missing?

bettinaheim (approved):
Looks good enough to pull in, so that we can parallelize work. Follow up items are 
- a separate PR for API adjustments
- updating the container validation script so that it properly runs the dynamics examples but only for the dynamics target

PR #2687, opened 2025-03-03
[ci] Fix tags

khalatepradnya:
Removed the incorrect use of `include` (which is used in GitHub actions, not GitLab)

Following description here: https://docs.gitlab.com/ci/jobs/job_control/#select-different-runner-tags-for-each-parallel-matrix-job


bmhowe23 (approved):
LGTM

PR #2688, opened 2025-03-04
Fix publishing error

sacpis:
Fix publishing error

# Error
```
RuntimeError: from_dlpack received an invalid capsule. Note that DLTensor capsules can be consumed only once, so you might have already constructed a tensor from it once.
```

Making from_dlpack compatible with both CuPy versions v13.4+ and v13.3 (and older).

### DIFF
--- a/python/cudaq/operator/integrators/cuda_torchdiffeq_integrator.py
+++ b/python/cudaq/operator/integrators/cuda_torchdiffeq_integrator.py
@@ -71,22 +71,13 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[CudmStateType]):
         t_scalar = t.item()
         # `vec` is a torch tensor on GPU
         # convert torch tensor to CuPy array without copying data
-        # CuPy v13.4+ (from_`dlpack` as a method)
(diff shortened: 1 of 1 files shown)


python/cudaq/operator/integrators/cuda_torchdiffeq_integrator.py
```diff
@@ -71,14 +71,13 @@ def compute_rhs(self, t, vec):
         t_scalar = t.item()
         # `vec` is a torch tensor on GPU
         # convert torch tensor to CuPy array without copying data
-        vec_cupy = cp.from_dlpack(torch.utils.dlpack.to_dlpack(vec))
+        vec_cupy = cp.from_dlpack(vec)
```
khalatepradnya, line 74: Q: This syntax works with both cuPy v13.4 and older versions?

1tnguyen: Yes, I've tested it with both versions.

khalatepradnya (approved):
LGTM. 👍🏽

PR #2699, opened 2025-03-06
[CI] [Publishing] Skip NVQC tests

khalatepradnya:
The NVQC tests in Publishing pipeline may fail due to version mismatch. Hence, don't run them in the `Publishing` pipeline.


khalatepradnya:
> Should we add a FIXME here? So that once we deploy a new image on the server, this test should pass (after updating the whitelist as Ben mentioned in the meeting).

Sure. (It wasn't clear to me whether we are going to do that update or not.)
Is there an issue tracking this? I can link it as well.

bmhowe23:
> Should we add a FIXME here? So that once we deploy a new image on the server, this test should pass (after updating the whitelist as Ben mentioned in the meeting).

From what I heard, I didn't think we were going to update the public server version ever again.

bmhowe23:
> > > Should we add a FIXME here? So that once we deploy a new image on the server, this test should pass (after updating the whitelist as Ben mentioned in the meeting).
> > 
> > 
> > From what I heard, I didn't think we were going to update the public server version ever again.
> 
> oh so, when you said new image update, is it on the internal server (nv-quantum)?

Yes.

bmhowe23:
> > > > > Should we add a FIXME here? So that once we deploy a new image on the server, this test should pass (after updating the whitelist as Ben mentioned in the meeting).
> > > > 
> > > > 
> > > > From what I heard, I didn't think we were going to update the public server version ever again.
> > > 
> > > 
> > > oh so, when you said new image update, is it on the internal server (nv-quantum)?
> > 
> > 
> > Yes.
> 
> I think we are good here then. Should we start deploying the latest image on the internal server?

Yes, the nightly integration test is attempting to do that on a regular basis, but it only leaves it deployed for the duration of the test.

.github/workflows/publishing.yml
```diff
@@ -1153,9 +1153,6 @@ jobs:
           cp    $GITHUB_WORKSPACE/$readme /tmp/README.md
           # Target tests should not be run here either, since that requires credentials but doesn't require a GPU runner.
 
-          # The NVQC API key is needed to validate NVQC related snippets and examples.
-          export NVQC_API_KEY="${{ secrets.NVQC_PROD_SERVICE_KEY }}"
```
mitchdz, line 1157: Do we want to also delete the code in `validate_pycudaq.sh` that tests nvqc? Or do we want to keep it for manual testing?

khalatepradnya: Yes, to latter. I think we should retain it to have a way to test it when needed.

mitchdz (approved):
LGTM. I asked about removing the code in `validate_pycudaq.sh` and we agreed upon keeping it in so it can be manually tested if needed.

bmhowe23 (approved):

PR #2701, opened 2025-03-06
[Python] Handle more conditions in `MidCircuitMeasurementAnalyzer`

khalatepradnya:
Ideally this logic should be replaced with `quake-add-metadata` pass, see issue [# 2000](https://github.com/NVIDIA/cuda-quantum/issues/2000).

This PR enhances the checks in the current logic to account for cases when a measurement result is assigned to new value, or is used with boolean operators.


bmhowe23 (approved):
This LGTM ... thanks, @khalatepradnya!

PR #2704, opened 2025-03-06
Support Rydberg Hamiltonian in C++ Evolve API

nvidia-dobri:
This enables Rydberg Hamiltonian evolution through analog quantum backends, currently QuEra and Pasqal. Authentication unique to each platform is required to run the examples.

QuEra requires valid AWS credentials with Braket access:
```
nvq++ --target quera 'quera_basic.cpp' -o out.x && CUDAQ_LOG_LEVEL=info ./out.x
nvq++ --target quera 'quera_intro.cpp' -o out.x && CUDAQ_LOG_LEVEL=info ./out.x
```

Pasqal requires env variables:
* `PASQAL_PROJECT_ID` from [Pasqal Cloud](https://portal.pasqal.cloud/)
* `PASQAL_AUTH_TOKEN` obtained with `pasqal_cloud` python library.
```
nvq++ --target pasqal 'pasqal.cpp' -o out.x && CUDAQ_LOG_LEVEL=info ./out.x
```


runtime/cudaq/dynamics/evolution.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    auto ph_res = ph.evaluate({{"t", step}});
+    ph_ts.push_back(std::make_pair(ph_res.real(), step.real()));
+
+    auto dg_res = dg.evaluate({{"t", step}});
+    dg_ts.push_back(std::make_pair(dg_res.real(), step.real()));
+  }
+
+  auto atoms = cudaq::ahs::AtomArrangement();
+  for (auto pair : hamiltonian.get_atom_sites()) {
```
schweitzpgi, line 42: nit: braces not needed

runtime/cudaq/dynamics/evolution.cpp
```diff
@@ -0,0 +1,101 @@
...
+  drive.detuning = delta;
+
+  auto program = cudaq::ahs::Program();
+  program.setup.ahs_register = atoms;
+  program.hamiltonian.drivingFields = {drive};
+  program.hamiltonian.localDetuning = {};
+
+  std::ostringstream programName;
+  programName << "__analog_hamiltonian_kernel__" << []() {
+    const std::string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
schweitzpgi, line 68: Do you really need to create a string object? Or can this be
```c++
  constexpr const char chars[] = "....";
```

runtime/cudaq/dynamics/evolution.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "common/EvolveResult.h"
+#include "common/Logger.h"
+#include "cudaq/operators.h"
+#include "cudaq/schedule.h"
+#include <random>
+#include <sstream>
+#include <string>
+
+namespace cudaq {
+namespace __internal__ {
```
schweitzpgi, line 19:
```suggestion
namespace cudaq::__internal__ {
```
No need to use Bronze Age C++. :)

schweitzpgi (commented):
Thanks, Dobri. LGTM with a quick skim. Just a couple of nits.

docs/sphinx/targets/cpp/pasqal.cpp
```diff
@@ -0,0 +1,69 @@
...
+// nvq++ --target pasqal pasqal.cpp -o out.x
+// ./out.x
+// ```
+// Assumes a valid set of credentials (PASQAL_AUTH_TOKEN, PASQAL_PROJECT_ID)
+// have been set.
+
+#include "cudaq/algorithms/evolve.h"
+#include "cudaq/dynamics_integrators.h"
+#include "cudaq/operators.h"
+#include "cudaq/schedule.h"
```
khalatepradnya, line 12: Q: Do we need to include all of these individually?
Wouldn't `#include <cudaq.h>` suffice?

khalatepradnya: In other words, shouldn't users get `evolve` by default? (like `sample` and `observe`)

docs/sphinx/targets/cpp/pasqal.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+  auto hamiltonian =
+      cudaq::rydberg_hamiltonian(register_sites, omega, phi, delta);
+
+  // Evolve the system
+  auto result = cudaq::evolve(hamiltonian, schedule, 100);
+  result.get_sampling_result()->dump();
+
+  return 0;
+}
```
khalatepradnya, line 69: This example can be added to the nightly integration test suite for Pasqal - [here](https://github.com/NVIDIA/cuda-quantum/blob/main/.github/workflows/integration_tests.yml#L708).
(See how it is done for Infleqtion - one C++ test and one Python test so that both frontends are covered.)
This change can go as a follow-up.

docs/sphinx/targets/cpp/quera_basic.cpp
```diff
@@ -0,0 +1,83 @@
...
+        double t = std::real(parameters.at("t"));
+        return std::complex<double>(
+            (t > time_ramp && t < time_max) ? delta_end : delta_start, 0.0);
+      });
+
+  auto hamiltonian =
+      cudaq::rydberg_hamiltonian(register_sites, omega, phi, delta);
+
+  // Evolve the system
+  auto result = cudaq::evolve(hamiltonian, schedule, 10);
```
khalatepradnya, line 79: Can you update one of these examples to use `evolve_async` API?

runtime/cudaq/algorithms/evolve.h
```diff
@@ -277,4 +277,25 @@ evolve_async(const HamTy &hamiltonian, const cudaq::dimension_map &dimensions,
...
+                                 const cudaq::schedule &schedule,
+                                 std::optional<int> shots_count = std::nullopt,
+                                 int qpu_id = 0) {
+  return cudaq::__internal__::evolve_async(
+      [=]() {
+        ExecutionContext context("evolve");
+        cudaq::get_platform().set_exec_ctx(&context, qpu_id);
+        return evolve(hamiltonian, schedule, shots_count);
+      },
+      qpu_id);
```
khalatepradnya, line 298: Can you confirm that this call is non-blocking? I think we need to set `asyncExec` on `ExecutionContext` to `true` somewhere... perhaps this is handled by the `cudaq::__internal__::evolve_async`?

runtime/cudaq/dynamics/rydberg_hamiltonian.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "cudaq/operators.h"
+#include <sstream>
+#include <stdexcept>
+
+namespace cudaq {
+rydberg_hamiltonian::rydberg_hamiltonian(
+    const std::vector<Coordinate> &atom_sites, const scalar_operator &amplitude,
+    const scalar_operator &phase, const scalar_operator &delta_global,
+    const std::vector<int> &atom_filling,
```
khalatepradnya, line 17: Should we make `atom_filling` optional as well?

unittests/dynamics/rydberg_hamiltonian.cpp
```diff
@@ -0,0 +1,124 @@
...
+  // Valid operators
+  scalar_operator amplitude(1.0);
+  scalar_operator phase(0.0);
+  scalar_operator delta_global(-0.5);
+
+  rydberg_hamiltonian hamiltonian(atom_sites, amplitude, phase, delta_global);
+
+  std::vector<int> expected_filling(atom_sites.size(), 1);
+  EXPECT_EQ(hamiltonian.get_atom_filling(), expected_filling);
+}
```
khalatepradnya, line 124: Nice set of tests!

runtime/cudaq/dynamics/rydberg_hamiltonian.cpp
khalatepradnya: Saw the tests cover this scenario.

khalatepradnya:
We need a follow-up docs update PR to add the new examples for [Pasqal](https://nvidia.github.io/cuda-quantum/latest/using/backends/hardware/neutralatom.html#pasqal) and [QuEra](https://nvidia.github.io/cuda-quantum/latest/using/backends/hardware/neutralatom.html#quera-computing).

khalatepradnya:
Manually tested examples for both Pasqal and QuEra.
(QuEra system was unavailable at the time of testing so checked payload and canceled job.)
Also tried Pasqal with `nvq++ --target pasqal --pasqal-machine EMU_FREE source.cpp`

runtime/cudaq/dynamics/evolution.cpp
```diff
@@ -0,0 +1,101 @@
...
+  auto programJson = nlohmann::json(program);
+
+  auto &platform = cudaq::get_platform();
+  auto ctx =
+      std::make_unique<ExecutionContext>("sample", shots_count.value_or(100));
+  ctx->asyncExec = false;
+  platform.set_exec_ctx(ctx.get());
+
+  auto programString = programJson.dump();
+  cudaq::info("Program JSON: {}", programString);
```
khalatepradnya, line 88: Observation - In my manual testing I found this resulted in a verbose output (while using logger of course), specifically when targeting Pasqal.

docs/sphinx/targets/cpp/pasqal.cpp
nvidia-dobri: I'll need to get back to this. Looks like all the dynamics examples also do this. Additionally, looks like we have a forward declaration in evolve that requires `dynamics_integrators` to be included even when unused, and that's pretty nasty. I already started working on the spec refactor, so let me address those there, it will be easier.

runtime/cudaq/algorithms/evolve.h
nvidia-dobri: Yeah, that's what cudaq::__internal__::evolve_async does, essentially. I can't say I fully understand that this is the canonical correct way to do this, but it does work and it is async 🙂  I "borrowed" it from the existing dynamics implementation.

runtime/cudaq/operators.h
```diff
@@ -589,6 +589,52 @@ class product_operator {
   friend product_operator<T> operator_handler::identity(int target);
 };
 
+/// @brief Representation of a time-dependent Hamiltonian for Rydberg system
+class rydberg_hamiltonian {
+public:
+  using Coordinate = std::pair<double, double>;
```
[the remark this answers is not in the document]
nvidia-dobri: Fixed.

unittests/dynamics/rydberg_hamiltonian.cpp
nvidia-dobri: Would love to take credit, but these were already in the massive operators / dynamics PR, they just got dropped along the way 😅

runtime/cudaq/dynamics/evolution.cpp
nvidia-dobri: Yeah, I'll switch this over to debug logging. I agree. I think at debug level, verbose/noisy is expected and I still think this is useful to be able to get easily.

.github/workflows/integration_tests.yml
```diff
@@ -706,6 +706,25 @@ jobs:
           export PASQAL_MACHINE_TARGET="EMU_FREE" # Speedup queuing time
           set +e # Allow script to keep going through errors
           test_err_sum=0
+          cpp_tests="docs/sphinx/targets/cpp/pasqal.cpp"
+          for filename in $cpp_tests; do
+            [ -e "$filename" ] || echo "::error::Couldn't find file ($filename)"
+            nvq++ --target pasqal $filename
```
khalatepradnya, line 712:
```suggestion
            nvq++ --target pasqal --pasqal-machine EMU_FREE $filename
```
to speed up queuing

khalatepradnya (approved):
LGTM. Thanks, @nvidia-dobri .

nvidia-dobri:
>/ok to test

bmhowe23 (approved):
I just reviewed the changes in the .github directory ... they LGTM.

.github/workflows/integration_tests.yml
```diff
@@ -702,6 +702,7 @@ jobs:
           python3 -m pip install pasqal-cloud
           export PASQAL_USERNAME='${{ secrets.PASQAL_USERNAME }}'
           export PASQAL_PASSWORD='${{ secrets.PASQAL_PASSWORD }}'
+          export PASQAL_AUTH_TOKEN=$(python3 docs/sphinx/targets/cpp/pasqal_auth.py)
```
khalatepradnya, line 705: Need to check that this new file will be available to this job, as in we don't have code that copies only `docs/sphinx/targets/cpp/*.cpp`... Checking that now.

khalatepradnya: The new file `docs/sphinx/targets/cpp/pasqal_auth.py` is incorrectly being treated as Python example by validation script and causing following failure in `Validate cuda-quantum image` stage:

```2025-03-07T22:48:28.8894236Z   File "/home/cudaq/targets/cpp/pasqal_auth.py", line 2, in <module>
2025-03-07T22:48:28.8894990Z     from pasqal_cloud import SDK
2025-03-07T22:48:28.8895320Z ModuleNotFoundError: No module named 'pasqal_cloud'
```
It needs to be skipped.

PR #2705, opened 2025-03-06
PR #2683 Follow up

sacpis:
PR #2683 Follow up

- Removing dynamics from the validation's script skip list
- Cleaning up if loop with a single statement enclosed in brackets
- Removing getters in EvolveResult and making optioanl members public, as per the spec
- Updated validation_script to skip dynamics examples in Create Docker image job and only enabled it in publishing workflow
- Updated EvolveResult API (as per the spec) to keep the states, expectation values, and sampleing_result


runtime/common/EvolveResult.h
```diff
@@ -24,20 +24,19 @@ class evolve_result {
   // The final state after the time evolution.
   state final_state;
 
+public:
```
nvidia-dobri, line 27: Can we match the spec here? Basically remove the intermediate / final distinction for state and expectation values and initialize everything to `std::nullopt`.
https://gitlab-master.nvidia.com/cuda-quantum/cudaq-spec/-/merge_requests/20/diffs

nvidia-dobri: We should be able to keep the Python API as is, juggling the binding methods. If it's more work than that, I agree, we shouldn't do it now.

This changes the public-facing api for evolve, so it's harder to change in next release.

runtime/common/EvolveResult.h
```diff
@@ -20,61 +20,62 @@ namespace cudaq {
...
-  evolve_result(state state) : final_state(state) {}
+  // Construct from single final state.
+  evolve_result(state state) {}
 
-  evolve_result(state state, const std::vector<observe_result> &expectations)
-      : final_state(state),
-        final_expectation_values(
-            std::make_optional<std::vector<observe_result>>(expectations)) {}
+  // Construct from single final observe result.
+  evolve_result(state state, const std::vector<observe_result> &expectations) {
```
bettinaheim, line 40: Did we land on having the return types for expectation be std::optional then? 
Works for me, just double checking.

bettinaheim (approved):

bettinaheim:
Pulling this in - the remaining failures are just because I removed python 3.10 from the wheel build and forgot to remove it from the metapackage validation (fixed now). I'll come back to the build changes I made here in a separate PR, but first we need the build in a working state again.

PR #2706, opened 2025-03-06
New launch kernel for python

annagrin:
### Description
Use new launch kernel from Python code for quantum devices.

This enables new synthesis (including state synthesis) for quantum devices in Python.

<strike>Required: https://github.com/NVIDIA/cuda-quantum/pull/2644</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2291</strike>

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2299
Towards: https://github.com/NVIDIA/cuda-quantum/issues/1931


python/runtime/utils/PyRemoteRESTQPU.cpp
```diff
@@ -57,16 +58,15 @@ TranslateFromMLIRRegistration::TranslateFromMLIRRegistration(
...
-  std::tuple<ModuleOp, MLIRContext *, void *>
-  extractQuakeCodeAndContext(const std::string &kernelName,
-                             void *data) override {
-
-    auto *wrapper = reinterpret_cast<cudaq::ArgWrapper *>(data);
-    auto m_module = wrapper->mod;
-    auto callableNames = wrapper->callableNames;
+private:
+  /// Creates new context without mlir initialization.
+  std::unique_ptr<MLIRContext> createContext() {
```
annagrin, line 63: @amccaskey @schweitzpgi I found this way of creating a context instead of passing it as an argument in ArgWrapper - do you think this makes sense? If yes, I can do the same for the remote simulator.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -267,9 +268,14 @@ pyAltLaunchKernelBase(const std::string &name, MlirModule module,
 
   std::string properName = name;
 
+  auto &platform = cudaq::get_platform();
   // If we have any state vector data, we need to extract the function pointer
   // to set that data, and then set it.
   for (auto &[stateHash, stateData] : *stateStorage) {
+    if (platform.is_remote() || platform.is_emulated())
```
annagrin, line 275: @amccaskey @schweitzpgi here is where I no longer pass module around

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -335,16 +345,15 @@ pyAltLaunchKernelBase(const std::string &name, MlirModule module,
...
+    auto args = runtimeArgs.getArgs();
+    if (platform.get_remote_capabilities().isRemoteSimulator)
+      args.insert(args.begin(), reinterpret_cast<void *>(&mod));
+
     if (platform.is_remote() || platform.is_emulated()) {
-      auto *wrapper = new cudaq::ArgWrapper{mod, names, rawArgs};
-      auto dynamicResult = cudaq::altLaunchKernel(
-          name.c_str(), thunk, reinterpret_cast<void *>(wrapper), size,
-          uReturnOffset);
+      auto dynamicResult = cudaq::streamlinedLaunchKernel(name.c_str(), args);
```
annagrin, line 354: @amccaskey @schweitzpgi here is where I no longer pass the module to PyRemoteRESTPU.cpp

python/runtime/utils/PyRemoteRESTQPU.cpp
amccaskey: from the test failure, looks like you'll need to add 
```cpp
  registerLLVMDialectTranslation(*context);
```
here

amccaskey: If Eric is ok with this then I think its a fine way to remove the need for passing ModuleOp+Context all the way through launching the kernel.

schweitzpgi: I feel like we're trying to coerce the Module and related MLIRContext to stick around even though they are going out of scope someplace. After spending a bit of time reading the code, it's not clear at all still.

`pyAltLaunchKernel` is getting a raw `mlir::ModuleOp` passed to it, which it uses. `extractQuakeCodeAndContext` (in REST QPU) is constructing its own `MLIRContext`, which it appears is trying to be a singleton instance. It's not entirely clear how this singleton would be the same one the `ModuleOp` used. I'm also not clear why we need a singleton `MLIRContext`. Since Python is itself multithreaded, that seems like a bottleneck and/or potential problem.

If we created `mlir::ModuleOp` instances and saved them in `std::shared_ptr`, I believe pybind11 would autogenerate the "right" code for us such that Python's reference counting would work with C++ shared ptr reference counting so that the Module and its context don't get deconstructed along the way.

boschmitt: > I feel like we're trying to coerce the Module and related MLIRContext to stick around even though they are going out of scope someplace. After spending a bit of time reading the code, it's not clear at all still.

I'm interested in learning more about the issues you are seen that are (might be) related to this. I've been trying to reason about the lifetime of these objects but it is quite hard.

The lifetime of `MLIRContext` and MLIR operations (e.g., `ModuleOp`) that cross the boundary between Python and C++ is weird. When we create `MLIRContext` on the Python side, we get a wrapper that keeps it alive for as long as other Python-side-created structures depend on it:

```python
def create_mod():
    ctx = Context()
    loc = Location.unknown(context=ctx)
    module = Module.create(loc=loc)
    return module
    # ctx goes out of scope but the object will be kept alive because of module
```

The issue starts becoming complex when we pass these objects to the C++ side. We unwrap, clone, store them in cache, but AFAIK none of these operations take ownership. Hence, for example, if we cache a cloned module in the C++ side, its related context will die if the original module dies on the Python side.

Note: on the python side, there are some incantations on `Context` that might help debug this sort of issues: `Context._get_live_count()`, `Context._get_live_module_count(...)`, and `Context._get_live_operation_count(...)`.

Note 2: An interesting thing to keep in mind is that operations can keep the context alive, but they don't keep other operations alive, i.e., if I create a `ModuleOp` with a `arith.ConstantOp` inside it, the module can die by going out of scope and the `ConstantOp` won't keep it alive:

```python
from cudaq.mlir.ir import Context, Location, Module, InsertionPoint, IntegerType
from cudaq.mlir.dialects import arith

def create_op():
    context = Context()
    loc = Location.unknown(context=context)
    module = Module.create(loc=loc)
    with InsertionPoint(module.body), loc:
        op = arith.ConstantOp(IntegerType.get_signless(32), 42)
    print("create_op:")
    print("  Live context: ", Context._get_live_count())
    print("  Live module: ", Context._get_live_module_count(context))
    print("  Live ops: ", Context._get_live_operation_count(context))
    parent = op.operation.parent
    print("  Parent : ", parent)
    return op

op = create_op()
parent = op.operation.parent
print("Parent : ", parent)
print("Live context: ", Context._get_live_count())
print("Live module: ", Context._get_live_module_count(op.context))
print("Live ops: ", Context._get_live_operation_count(op.context))
```

Outputs:
```
create_op:
  Live context:  2
  Live module:  1
  Live ops:  1
  Parent :  module {
  %c42_i32 = arith.constant 42 : i32
}

Parent :  None
Live context:  2
Live module:  0
Live ops:  1
```

schweitzpgi: Thanks, @boschmitt . This is indeed quite interesting. I'm assuming the pybind11 wrappers on the python side are keeping reference counts and that is what these `_get_live_<foo>count` calls are reporting. It's not at all clear to me that those Python objects are, in fact, keeping the C++ object itself alive though.

It should be noted that we need to port everything to nanobind anyway (in order to move to modern MLIR), so deep study of what pybind11 is doing is probably not time well spent.

Part of my own confusion in my earlier reply is that the C++ remote REST and the Python remote REST are entirely distinct. @annagrin changes here make them more aligned though, so that's a good thing.

It does seem like, barring performance concerns, that "passing MLIR code" ought to be done by using the string representation and constructing the module+context where it is needed vs. trying to keep the entire collection of MLIR objects (referenced by the MLIRContext) alive. That approach may help clarify the code on the Python side further.

boschmitt: > I'm assuming the pybind11 wrappers on the python side are keeping reference counts and that is what these `_get_live_<foo>count` calls are reporting.

The `Context` type on the Python side is a subclass of `_BaseContext`, which, in turn, is the wrapper I mentioned, `PyMlirContext`.

https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/mlir/lib/Bindings/Python/IRCore.cpp#L2317-L2323

It is this machinery that keeps the live counts and all.

> It should be noted that we need to port everything to nanobind anyway (in order to move to modern MLIR), so deep study of what pybind11 is doing is probably not time well spent.

Indeed, although some of what seems to be biting us is orthogonal to the change.

> Part of my own confusion in my earlier reply is that the C++ remote REST and the Python remote REST are entirely distinct. @annagrin changes here make them more aligned though, so that's a good thing.

Agreed. The behavior aligns and that is good; but I do wonder if it wouldn't be better to implement the Python remote REST in Python---trying minimize this complex dance of getting objects that are bound and wrapped by one part of the C++ runtime (the MLIR part) to the non-MLIR side of the C++ runtime, while passing through Python.

annagrin: @boschmitt I actually tried to break passing the module and context references to c++ side - I instead created a new context and loaded and parsed a module string from cache using the context. Would this free us from keeping the python's context and module alive? I am trying to run the tests and see what happens.

boschmitt: It feels like a workaround. I suppose it is fine for now, given that most kernels I have seen are small, but I wonder about the scalability. We take time build the IR just to dump it to a string, and then have this string parsed to rebuild the IR on the other side.

bmhowe23: > We take time build the IR just to dump it to a string, and then have this string parsed to rebuild the IR on the other side.

Is this avoided when the kernel can be cached? There are some use cases where we repeatably invoke kernels in a loop, like with `optimize` or `minimize`. Would the "dump it to a string and rebuild on the other side" be avoided in those cases?

annagrin: @boschmitt @bmhowe23 @amccaskey @schweitzpgi I updated the PR, appreciate a look! Currently I made the python side behave similar to the c++ side (ie we cash the module as a string and parse it in BaseRESTXXX.cpp. I will check if we cache it for the case Ben mentions, if not I can try adding a cache in BaseRESTXXX.cpp in a separate PR.

annagrin: @bmhowe23 Re running kernels in a loop, I think a general after-synthesis caching would be helpful (but of course would only work if we run a kernel with the same parameters). I can also try adding caching for before-synthesis quake, so we don't need to parse the module every time, but I would like to make it work of all scenarios, i.e. (python, c++)x(remote sim, quantum device) combinations so I would like to do it in a separate PR.

bmhowe23: > @boschmitt @bmhowe23 @amccaskey @schweitzpgi I updated the PR, appreciate a look! Currently I made the python side behave similar to the c++ side (ie we cash the module as a string and parse it in BaseRESTXXX.cpp. I will check if we cache it for the case Ben mentions, if not I can try adding a cache in BaseRESTXXX.cpp in a separate PR.

I probably should've been more clear in my question as I think there's actually 2 questions in play.

1. Does this PR change how non-state-synthesis kernels work with respect to caching? I think the answer to this is 'no' because https://github.com/NVIDIA/cuda-quantum/blob/615937c49c62534a1d465667b75304136b2c03e9/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp#L86 was unmodified, but it would be helpful if you could confirm that. If the answer is 'yes', then note that `mod.walk` + `llvm::hash_combine(hash, OperationEquivalence::computeHash(op)` is *significantly* faster than converting the IR to a string and then hashing the string. Also, if the answer is 'yes', I'd consider that a blocker for this PR as that would negatively impact our current performance.
2. Does this PR cache the state synthesis runs for kernels that require state synthesis to be called? This is more of a curiosity, and if the answer is 'no', then it could always be handled separately. That is - I don't consider this one a blocker for this PR.

bmhowe23: > For simulators (that don't require synthesis) this PR does not change anything.

That sounds good. Thanks for confirming.

> For quantum hardware, the llvm caching and storing of a quake module as a string has not changed and is still happening, like before. What changed is that we now use the quake cache - we load the module from a string in PyRemoteRestQPU.cpp instead of using a module argument, which is consistent with the c++ behavior.

I'm less concerned about string hashing/parsing for cases that use quantum hardware. For those use cases, I think the "over the wire" I/O is going to dominate any string parsing time, so unless you have some data that suggests otherwise, I don't think any changes would be required for that.

> Note that we always did mod.walk llvm caching and storing of quake module as a string in a hash table, but we don't hash the module string, the modules are stored by names. It was done for c++ and python as well.

Re: "storing by names" ... what if the kernel changes between sample calls? I am not sure if this can happen in c++, but I can imagine something like the program below for Python:

```python
# First definition
@cudaq.kernel
def my_kernel():
  q = cudaq.qvector(3)
  mz(q)
print(cudaq.sample(my_kernel))

# Second definition
@cudaq.kernel
def my_kernel():
  q = cudaq.qvector(3)
  x(q)
  mz(q)
print(cudaq.sample(my_kernel))
```

This currently works correctly even with quantum hardware (i.e. `python3 test.py --target quantinuum --emulate`). Does it still work with your PR?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -99,10 +99,10 @@ jitAndCreateArgs(const std::string &name, MlirModule module,
     PassManager pm(context);
     pm.addNestedPass<func::FuncOp>(cudaq::opt::createPySynthCallableBlockArgs(
         SmallVector<StringRef>(names.begin(), names.end())));
+    pm.addPass(cudaq::opt::createLambdaLiftingPass());
```
schweitzpgi, line 102: The reason this was done after GKE was to keep the size of the embedded MLIR string small. I think that still holds true, since lambda lifting can cause code growth.

We probably need to make sure lambda lifting is run in all the JIT pipelines.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -335,17 +344,25 @@ pyAltLaunchKernelBase(const std::string &name, MlirModule module,
...
+      // Remote simulator - use altLaunchKernel to support returning values.
       auto *wrapper = new cudaq::ArgWrapper{mod, names, rawArgs};
       auto dynamicResult = cudaq::altLaunchKernel(
           name.c_str(), thunk, reinterpret_cast<void *>(wrapper), size,
           uReturnOffset);
       if (dynamicResult.data_buffer || dynamicResult.size)
         throw std::runtime_error("not implemented: support dynamic results");
       delete wrapper;
+    } else if (platform.is_remote() || platform.is_emulated()) {
+      // Quantum devices or their emulation - we can use streamlinedLaunchKernel
```
schweitzpgi, line 358: Does `is_remote` mean "quantum hardware device"? That's not very intuitive, if so.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -390,6 +412,30 @@ pyCreateNativeKernel(const std::string &name, MlirModule module,
...
+
+    if (stateData.kernelName != name)
+      continue;
+
+    auto setStateFPtr = jit->lookup("nvqpp.set.cudaq.state." + stateHash);
+    if (auto error = setStateFPtr.takeError()) {
+      auto message = "python alt_launch_kernel failed to get set cudaq state "
+                     "function for kernel: " +
+                     name;
+      llvm::logAllUnhandledErrors(std::move(error), llvm::errs(), message);
```
schweitzpgi, line 430: Why is this one using the `logAllUnhandledErrors` function but other cases throw a runtime_exception?

python/runtime/utils/PyRemoteRESTQPU.cpp
```diff
@@ -57,40 +58,64 @@ TranslateFromMLIRRegistration::TranslateFromMLIRRegistration(
 // implement some core functionality here in PyRemoteRESTQPU so we don't load
 // twice
 class PyRemoteRESTQPU : public cudaq::BaseRemoteRESTQPU {
+private:
+  /// Creates new context without mlir initialization.
+  MLIRContext *createContext() {
```
schweitzpgi, line 63: Should this be guarded by a mutex? (run_once)

python/runtime/utils/PyRemoteRESTQPU.cpp
```diff
@@ -57,40 +58,64 @@ TranslateFromMLIRRegistration::TranslateFromMLIRRegistration(
...
+    pm.addNestedPass<func::FuncOp>(cudaq::opt::createClassicalMemToReg());
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+    pm.addNestedPass<func::FuncOp>(cudaq::opt::createUnwindLoweringPass());
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addPass(cudaq::opt::createApplyOpSpecializationPass());
     pm.addPass(createInlinerPass());
-    pm.addPass(createCanonicalizerPass());
-    pm.addPass(createCSEPass());
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+    pm.addNestedPass<func::FuncOp>(createCSEPass());
```
schweitzpgi, line 118: Better locality. Nice. :)

schweitzpgi (approved):
LGTM. Thanks, Anna.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: I can move it back and make PyRemoteRestQPU.cpp run it as it done before.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: Yeah this is very non-intuitive, `is_remote() || is_emulated()` is true for remote simulators, quantum devices, or their emulation. `platform.get_remote_capabilities().isRemoteSimulator` is only true for the remote simulator.

I will add some clarification.

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -140,6 +140,44 @@ state pyGetStateRemote(py::object kernel, py::args args) {
                                            size, returnOffset));
 }
 
+/// @brief Python implementation of the `QPUState`.
+// Note: Python kernel arguments are wrapped hence need to be unwrapped
+// accordingly.
+class PyQPUState : public QPUState {
```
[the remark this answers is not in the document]
schweitzpgi: Just in general, we have a raft of classes that are cloned from the C++ code and specialized for Python. That is suboptimal at best. We should probably have an issue to go in and convert them to CRTP classes at minimum so that only they share the bulk of their implementations.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -38,6 +38,8 @@
 #include <pybind11/numpy.h>
 #include <pybind11/stl.h>
 
+#include <iostream>
```
amccaskey, line 41: leftover from debugging?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: I moved the changes not directly related to enabling new launch kernel in python to: https://github.com/NVIDIA/cuda-quantum/pull/2780

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -121,7 +117,7 @@ class PyRemoteSimulationState : public RemoteSimulationState {
     return context.overlapResult.value();
   }
 
-  ~PyRemoteSimulationState() { delete argsData; }
+  virtual ~PyRemoteSimulationState() override { delete argsData; }
```
annagrin, line 120: @1tnguyen this was never called before:)

python/runtime/utils/PyRemoteRESTQPU.cpp
annagrin: It works with this PR!

python/runtime/cudaq/algorithms/py_state.cpp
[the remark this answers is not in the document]
annagrin: Created an issue: https://github.com/NVIDIA/cuda-quantum/issues/2781

PR #2631, opened 2025-02-18
Operator classes

bettinaheim:
Still needs a couple more tests and other minor clean up (files that I checked in that should not be committed). 

But it is starting to shape up decently. The general operator infrastructure here is intended to fully replace today's spin-op classes. To help parallelize the work building on this, I'll defer the removement + replacement of the existing spin ops to a separate PR. I also need to implement the in-place multiplication for boson operators + tests for those. Structurally and implementation wise, this should be in good shape otherwise. 

I chose to 1) make header files proper header files that do not contain implementation and 2) instantiate all templates such that they do not (and cannot) get instantiated in other libraries. I think this is a decent choice here, since with this the different operator types very much act like individual non-template classes. The downside is that there is a bit much of boilerplate code, since each template needs to be declared, defined, and instantiated separately from the header definition of the class. That said, I do not expect we will ever really need to make any changes to that infrastructure (i.e. to operator_sum, product_operator, operators.h, templates.h. etc.). That setup is as generic as it gets. Adding a new operator kind would require defining the handler (see operator_handler in operator_leafs.h), and instantiating the templates for the operators on this handler.

Perf wise, I think this is a good path too. Numbers for the extreme cases are posted below. The most representative number to me is to have a sum of random product terms. On that, we are about a factor 20+ faster than the existing spin ops. The worst case scenario for the new implementation is the repeated multiplication of sum terms where each sum acts on new consecutive degrees of freedom. In that case, we loose about a factor 1.5-2 on the existing spin ops. However, I think this is an acceptable tradeoff; even with 2^20 terms, we are still at about 1-2s of runtime to build the op, and the only scenario this would be relevant for is when someone does a rather naive (suboptimal) implementation of a jordan wigner transformation (which I think would usually have a less terms). 
The new implementation performs a lot better than the existing encoding when terms do not act on consecutive degrees of freedom, and it is also quite a bit less cryptic (no encoding really, hence the same implementation for all operator kinds). The handlers therefore become pretty much trivial, where each handler just stores essentially how to get the matrix for a single operand. The product multiplication performance is somewhat order dependent (we search a vector to do in-place multiplication, and necessarily either need to start searching from the back or from the front). I can assume that people likely write code with increasing index for the degree and optimize internal storage for that, but that means we would do an additional reshuffle when computing a matrix to match what I think is the user level convention (we should really align that convention with the performant internal one....)

```
multiplication inplace with itself
nr ops:  100
Old setup took 0.000232378 seconds.
[...]

PR #2710, opened 2025-03-07
Replacing existing spin_ops with the general operators

bettinaheim:
This PR contains the full replacement of the existing spin_ops with the new general operators. 
It essentially keeps the python bindings as is with the trampolining between the C++ spin_ops and the python operators. That really needs to be replaced, but is a somewhat extensive change that I suggest to make in a separate PR. I did test that just replacing spin/spin ops in Python directly with the bound data structures works throughout all tests except for dynamics (which require that all other operators also are bound). 

Some perf numbers can be found here: https://github.com/NVIDIA/cuda-quantum/pull/2631
These largely remain the same - there is a tiny bit of overhead added due to the switch from int to size_t and the introduced default value, but that is < 10% (last benchmark is unchanged).

Full list of breaking vs non-breaking changes:

Breaking changes:

- the new operator framework introduces a proper type distinction between product operators and operators that are a sums of products; this is now reflected throughout the API
  - `x`, `y`, `z`, and `i` in `cudaq::spin` create a `spin_op_term` (a product operator)
  - the meaning of `spin_op::empty` changed - it now returns an empty sum
  - behavior change in `spin_op()` constructor - it creates an uninitialized sum rather than the product identity
  - behavior change in `spin_op(std::size_t)` - it creates an empty sum reserving space for the given number of terms
- changes related to `spin_op_term`:
  - it is no longer a value of type `std::vector<bool>` but instead it's own class type (`product_op<spin_handler>`)
  - the new `spin_op_term` includes all information about the product, including its coefficient
  - a range of utility functions exist on `spin_op_term`
  - `spin_op::key_type` and `spin_op::mapped_type` no longer exist
  - the value type of the `spin_op` iterator is a `spin_op_term`
- iterators:
  - neither the `spin_op` iterator nor the `spin_op_term` iterator allow to modify the operator itself
- equality comparison:
  - two operators are only equal if they act on the same degrees (qubits)
  - two operators are only equal if all terms *and* coefficients match
  - [same as before] equality comparison accounts for commutation relations
- properties:
  - `num_qubits` reflects only the degrees the operator explicitly acts upon; e.g. `spin_op::i(2)` act on a single degree
  - `to_matrix` reflects only the degrees the operator explicitly acts upon; e.g. `spin_op::x(1)` returns a 2x2 matrix (previously a 4x4 matrix)
  - `get_coefficient` returns a scalar callback function that in general may require a parameter_map for evaluation;
    `evaluate_coefficient` replaces the previous get_coefficient implementation and takes an optional parameter_map argument
  - `to_string()` produces a different string than before - see also deprecation of `to_string(bool)`
- changes in usage:
  - [internal only change] the results for terms are stored under the value returned by `get_term_id()` and values for the entire sum are stored under the key obtained by calling `to_string()`
  - the Pauli word representation for a `spin_op_term` can be obtained by calling `get_pauli_word(size_t = 0)`;
    by default that representation does not include identities on qubits the operator does not explicitly act on,
    but if the optional integral argument is provided, the string includes identities for the degrees in `[0, arg)` that the operator does not act on
  - `exp_pauli` expects the number of qubits to match the length of the Pauli word
  - `expectation` and `count` will fail instead of return `0.0` if the requested value has not been computed/found
  - the order of iteration over term for `spin_op` may be different than with the previous implementation;
    this should largely be irrelevant but may lead to slightly different results in some cases due to slightly different finite precision effects
- build system changes:
  - the library `cudaq-spin` is replaced by `cudaq-operator`
  - `CUDAQSpinConfig.cmake` and `CUDAQSpinTargets.cmake` no longer exist - the replacements are the corresponding `CUDAQOperator*.cmake` files
  - [mostly internal only change] various minor changes in the files included by headers and in signatures (additional const's) within the CUDA-Q runtime infrastructure (all of that should be mostly internal changes and not have any impact on existing application code)

Deprecated:

- deprecated constructors:
  - `spin_op(const std::vector<bool> &term, const std::complex<double> &coeff)` (FIXME...)
  - `spin_op(std::pair<const std::vector<bool>, std::complex<double>> &termData)` (FIXME...)
  - `spin_op(pauli, const std::size_t id, std::complex<double> coeff = 1.0)` (FIXME...)
  - `spin_op(const std::unordered_map<std::vector<bool>, std::complex<double>> &_terms)` (FIXME...)
  - `spin_op(const std::vector<std::vector<bool>> &bsf, const std::vector<std::complex<double>> &coeffs)` (suggest no substitution)
  - `sum_op(const std::vector<double> &input_vec, std::size_t nQubits)` (replaced with the matching constructor for new serialization format)
- deprecated properties:
  - `get_raw_data` (no substitution)
  - `is_identity` (to be removed on `spin_op` class, but will continue to be supported on `spin_op_term`)
  - `to_string(bool)` (replaced by either `to_string()`, `get_term_id`, or `get_pauli_word`)
  - `getDataRepresentation` (replaced by `get_data_representation` to generate the new serialization format)
  - `getDataTuple` (no longer used, no substitution, matching constructor listed above)
  - `for_each_term` (replaced by iterator)
  - `for_each_pauli` (replaced by iterator)
- deprecations related to the product vs sum operator distinction:
    -  `expectation` and `count` will require a `spin_op_term` going forward; passing a spin_op (with a single term, as previously) is still supported but deprecated
- headers:
  - dynamics_integrators.h is deprecated and move to cudaq/algorithms/integrator.h instead

Other changes:

- `csr_spmatrix` is defined in the `cudaq` namespace rather than a type definition on `spin_op`
- [internal only] the context no longer stores a pointer to a `spin_op` but rather stores the value
- the serialization format for `spin_op` changed; deprecated functions producing and loading the old format are still available, but the entire internal infrastructure switched to the new format obtained by `get_data_representation`
- the `binary_spin_op_reader::read(const std::string &)` is set up to read the new data format -
  an overload  `binary_spin_op_reader::read(const std::string &, bool)` is provided that will read the old serialization format if the optional bool argument is set to true
- changing the type of `cudaq::dimension_map` to `std::unordered_map<std::size_t, int64_t>`; this should have no impact on application code in most cases, the only exception being a type change of the dimensions vector in the definition of a custom operator
- `degrees` no longer takes a boolean parameter - any optional argument that was passed can be dropped without a change in behavior
- `to_matrix` takes an additional optional boolean parameter `invert_order` that allows to retrieve the matrix with an inverted ordering convention compared to the default order
- removes the long deprecated exp_val_z and expectation_n


runtime/cudaq/dynamics/spin_operators.h
```diff
@@ -13,75 +13,81 @@
...
 
 #include "cudaq/operators.h"
-#include "cudaq/utils/tensor.h"
+#include "cudaq/utils/matrix.h"
 
 namespace cudaq {
 
-// FIXME: rename to spin ...
-class spin_operator : public operator_handler {
+enum class pauli { I, X, Y, Z };
```
bmhowe23, line 20: This is a different order than the order called out in the comment on line 27. Does that matter? Should the comment or enum be updated?
```cpp
// I = 0, Z = 1, X = 2, Y = 3
```

bettinaheim: The order does not matter, and neither ever does the private op code - it is private for a reason. :)
That said, I can align it with this one so that the data format and the internal representation match.

runtime/cudaq/operators.h
```diff
@@ -142,10 +147,17 @@ class sum_op {
   std::vector<std::size_t> degrees(bool application_order = true) const;
 
   /// @brief Return the number of operator terms that make up this operator sum.
-  std::size_t num_terms() const;
+  int num_terms() const;
```
amccaskey, line 150: why the change to int here?

bettinaheim: I need to do a full pass over int vs size_t usage. I'll come back to it with my take after looking closer at that.

runtime/common/ObserveResult.h
```diff
@@ -31,7 +31,7 @@ class observe_result {
   sample_result data;
 
 public:
-  observe_result() = default;
+  observe_result() : spinOp(cudaq::spin_op::empty()) {}
```
bmhowe23, line 34: I'd recommend retaining the default default constructor (due to a myriad of complicated C++ rules about when overriding one changes the availability of other default operations) and instead changing line 28 to be 
```cpp
  FROM:
  spin_op spinOp;
  TO:
  spin_op spinOp = cudaq::spin_op::empty();
```

runtime/common/ExecutionContext.h
```diff
@@ -31,7 +31,7 @@ class ExecutionContext {
   std::size_t shots = 0;
 
   /// @brief An optional spin operator
-  std::optional<cudaq::spin_op *> spin;
+  std::optional<cudaq::spin_op> spin;
```
bmhowe23, line 34: This will cause additional copies to be made, and spin_ops can be fairly large (i.e. slow copies). Was this change needed for anything in particular?

bettinaheim: Yes, it was. The issue was that we get an invalid pointer access in some cases. I believe it was related to a spin_op_term being converted to a spin_op. The subsequent access to it via pointer is invalid in some cases when the temporary spin_op value goes out of scope. You can check the failure if you undo the specific commit and run the python tests, but I suspect it could be a problem in C++, too. Our code generally just is not very good with testing behavior for things going out of scope in a larger code example, since all unit tests are kind of self-contained in a single function.

I did think about perf implications, but as far as I saw, I think this is one additional copy, which compared to the copies to build the operator and execute whatever code should not be noticeable. Correct me if you think this may be more than one additional copy.

runtime/common/ObserveResult.h
bettinaheim: Left spinOp as default constructed (uninintialized) for now. We can decide what to do with it when we decide on what to do with the default constructor.

runtime/cudaq/operators.h
bettinaheim: using size_t across the board seems to work well. did that.

efratshabtai:
What will be the best way to communicate these changes to the users?
I imagine https://nvidia.github.io/cuda-quantum/latest/api/languages/cpp_api.html#operators and https://nvidia.github.io/cuda-quantum/latest/specification/cudaq/operators.html#cudaq-spin-op will be updated. Obviously release notes.

Should we add a C++ example to https://nvidia.github.io/cuda-quantum/latest/examples/python/operators.html that goes through a typical usage of the public functions of the class?

As for compilation time detection - I guess some things will be obvious like empty having a different return value that will result in a compilation error. I think it's worth going over the list of changes and think what will be the experience of someone upgrading and if the compilation / runtime errors make sense.

docs/sphinx/applications/python/hamiltonian_simulation.ipynb
```diff
@@ -253,19 +253,24 @@
...
     "    return result\n",
     "\n",
     "def extractWords(hamiltonian: cudaq.SpinOperator):\n",
+    "    # Our kernel uses these words to apply exp_pauli to the entire state.\n",
+    "    # we hence ensure that each pauli word covers the entire space.\n",
+    "    n_spins = hamiltonian.get_qubit_count()\n",
     "    result = []\n",
-    "    hamiltonian.for_each_term(lambda term: result.append(term.to_string(False)))\n",
+    "    for term in hamiltonian:\n",
+    "        result.append(term.get_pauli_word(n_spins))\n",
```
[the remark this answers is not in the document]
bettinaheim: Let me double check, but I think these changes should work with 0.10.0; I largely had this PR ready prior to the 0.10.0 release, so everything that is not mitigation for deprecation should already be in 0.10.0. It was just too risky and too tight to pull it in in a way that is as nice as possible with regards to existing code.

docs/sphinx/examples/cpp/basics/expectation_values.cpp
```diff
@@ -21,9 +21,10 @@ struct ansatz {
 int main() {
 
   // Build up your spin op algebraically
-  using namespace cudaq::spin;
-  cudaq::spin_op h = 5.907 - 2.1433 * x(0) * x(1) - 2.1433 * y(0) * y(1) +
-                     .21829 * z(0) - 6.125 * z(1);
+  cudaq::spin_op h =
+      5.907 - 2.1433 * cudaq::spin_op::x(0) * cudaq::spin_op::x(1) -
+      2.1433 * cudaq::spin_op::y(0) * cudaq::spin_op::y(1) +
+      .21829 * cudaq::spin_op::z(0) - 6.125 * cudaq::spin_op::z(1);
```
1tnguyen, line 27: nit: there is another similar code in `docs/sphinx/examples/cpp/other/remote_sim/simple.cpp` that needs to be updated.

targettests/Remote-Sim/test_trotter.cpp
bmhowe23, line 1: I think you changed all the line endings to dos instead of unix here. It would be nice to revert it back to unix to make a diff possible.

docs/sphinx/applications/cpp/qaoa_maxcut.cpp
```diff
@@ -51,13 +50,13 @@ struct ansatz {
...
 int main() {
 
-  using namespace cudaq::spin;
-
   cudaq::set_random_seed(13); // set for repeatability
 
   // Problem Hamiltonian
-  const cudaq::spin_op Hp = 0.5 * z(0) * z(1) + 0.5 * z(1) * z(2) +
-                            0.5 * z(0) * z(3) + 0.5 * z(2) * z(3);
+  const cudaq::spin_op Hp = 0.5 * cudaq::spin_op::z(0) * cudaq::spin_op::z(1) +
```
bmhowe23, line 56: This format is fairly verbose, requiring `spin_op::` in front of all of these functions. I would advocate for keeping the `cudaq::spin` namespace and allow the old trivial `x`, `y`, and `z` functions to still be used ... it's much less verbose, and that's a good thing in my opinion. (Or, at a minimum, don't deprecate the old way.)

amccaskey: I agree Ben. We should keep the convenience namespaces (`spin`, `fermion`, `boson`). This will help us cut down on the boilerplate.

python/cudaq/operator/expressions.py
bmhowe23, line 1: This looks like another file where ever line changed (likely line endings).

python/runtime/common/py_ObserveResult.cpp
bmhowe23, line 1: Line endings?

runtime/common/MeasureCounts.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/helpers.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/helpers.h
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/operator_leafs.h
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/operator_sum.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/product_operators.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/spin_operators.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/dynamics/spin_operators.h
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/cudaq/operators.h
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/nvqir/cudensitymat/CuDensityMatOpConverter.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/nvqir/cutensornet/tensornet_spin_op.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

runtime/nvqir/qpp/QppCircuitSimulator.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

unittests/dynamics/operator_conversions.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

unittests/dynamics/operator_sum.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

unittests/dynamics/product_operator.cpp
bmhowe23, line 1: Perhaps Windows/DOS line endings in this file, too.

scripts/build_cudaq.sh
```diff
@@ -74,6 +74,9 @@ repo_root=$(cd "$this_file_dir" && git rev-parse --show-toplevel)
 # Prepare the build directory
 build_dir="$working_dir/build"
 echo "Build directory: $build_dir"
+if [ "${build_configuration,,}" == "release" ]; then
```
bmhowe23, line 77: This one caught me off guard while building and testing ... was this an intentional check in? We may want to advertise this one in a team meeting if it was intentional because this changes the development workflow for some users.

bettinaheim: No, I'll undo the changes to build_cudaq.sh and the remove the checked in main.cpp. I just had this for convenience during development, sorry.

docs/sphinx/specification/cudaq/operators.rst
```diff
@@ -1,81 +1,50 @@
...
-    };
-
-    namespace spin {
-      spin_op x(const std::size_t);
-      spin_op y(const std::size_t);
-      spin_op z(const std::size_t);
-    }
-  }
-
+Quantum Operators
```
amccaskey, line 1: we'll need this change reflected in the internal language spec too until we figure out how to unify them.

main.cpp
```diff
@@ -0,0 +1,448 @@
+/*******************************************************************************
```
amccaskey, line 1: remove?

python/runtime/cudaq/spin/py_spin_op.cpp
```diff
@@ -5,11 +5,14 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
+
+#include <iostream>
```
amccaskey, line 9: leftover iostream include

runtime/cudaq/operators/sum_op.cpp
```diff
@@ -1634,7 +1880,9 @@ template sum_op<spin_handler> sum_op<spin_handler>::random(std::size_t nQubits,
 SPIN_OPS_BACKWARD_COMPATIBILITY_DEFINITION
 sum_op<HandlerTy>::sum_op(const std::vector<double> &input_vec,
                           std::size_t nQubits) {
-  auto n_terms = (int)input_vec.back();
+  if (input_vec.size() == 0)
+    throw std::runtime_error("input vector must not be empty");
+  auto n_terms = (std::size_t)input_vec.back();
```
bmhowe23, line 1885: The line right below this can cause a runtime error without any useful error messages if `n_terms` is 0. I would recommend throwing an exception with a meaningful error message if this happens.

bettinaheim: good catch

python/runtime/cudaq/spin/py_spin_op.cpp
```diff
@@ -200,22 +455,78 @@ void bindSpinOperator(py::module &mod) {
...
-      .def("to_matrix", &spin_op::to_matrix,
-           "Return `self` as a :class:`ComplexMatrix`.")
+      .def(
+          "to_matrix",
+          [](spin_op &self) {
+            return self.to_matrix(); // can't bind function ref since it has
+                                     // additional (optional) args
+          },
+          "Return `self` as a :class:`ComplexMatrix`.")
+      /*
```
bmhowe23, line 465: Just confirming it was intentional to comment out `to_sparse_matrix` here? If so, should it just be removed?

python/tests/builder/test_SpinOperator.py
```diff
@@ -59,7 +59,7 @@ def test_spin_op_operators():
     each individual operator at the moment.
     """
     # Test the empty (identity) constructor.
-    spin_a = cudaq.SpinOperator()
+    spin_a = cudaq.SpinOperator.empty()
```
bmhowe23, line 62: The above comment no longer matches what the code is doing here.

python/tests/builder/test_SpinOperator.py
```diff
@@ -230,16 +241,16 @@ def test_spin_op_members():
...
     # Only have 1 term and 1 qubit.
-    assert spin_operator.get_term_count() == 1
-    assert spin_operator.get_qubit_count() == 1
+    assert spin_operator.get_term_count() == 0
+    assert spin_operator.get_qubit_count() == 0
     spin_operator += -1.0 * spin.x(1)
     # Should now have 2 terms and 2 qubits.
-    assert spin_operator.get_term_count() == 2
-    assert spin_operator.get_qubit_count() == 2
+    assert spin_operator.get_term_count() == 1
```
bmhowe23, line 252: This is now inconsistent with the above comment.

python/tests/kernel/test_trotter.py
```diff
@@ -48,31 +48,38 @@ def run_steps(steps: int, spins: int):
         omega = 2 * np.pi
 
         def heisenbergModelHam(t: float) -> cudaq.SpinOperator:
-            tdOp = cudaq.SpinOperator(num_qubits=n_spins)
+            tdOp = cudaq.SpinOperator.empty()
```
bmhowe23, line 51: This seems like a change in behavior. Is that intentional?

runtime/cudaq/operators.h
```diff
@@ -90,19 +96,17 @@ class sum_op {
...
 
-    const_iterator(const sum_op<HandlerTy> *sum,
-                   std::unordered_map<std::string, int>::const_iterator &&it)
-        : sum(sum), iter(std::move(it)), current_val(1.) {
-      if (iter != sum->term_map.end())
-        current_val = product_op<HandlerTy>(sum->coefficients[iter->second],
-                                            sum->terms[iter->second]);
+    const_iterator(const sum_op<HandlerTy> *sum, std::size_t idx)
+        : sum(sum), current_val(1.), current_idx(idx) {
+      if (current_idx < sum->num_terms())
```
bmhowe23, line 103: This dereferences `sum` without checking whether it is `nullptr`.

runtime/cudaq/operators.h
```diff
@@ -588,15 +666,20 @@ class product_op {
 
   // read-only properties
 
+#if !defined(NDEBUG)
+  bool is_canonicalized() const;
+#endif
```
bmhowe23, line 671: Just to confirm, you want this removed from release builds? Any particular reason it shouldn't be in release builds?

runtime/cudaq/operators/spin_op.cpp
```diff
@@ -119,25 +127,54 @@ complex_matrix spin_handler::to_matrix(std::string pauli_word,
...
   auto nr_deg = pauli_word.size();
 
-  complex_matrix matrix(dim, dim);
   for (std::size_t old_state = 0; old_state < dim; ++old_state) {
     std::size_t new_state = 0;
     std::complex<double> entry = 1.;
     for (auto degree = 0; degree < nr_deg; ++degree) {
-      auto canon_degree = degree;
-      auto state = (old_state & (1 << canon_degree)) >> canon_degree;
+      auto state = (old_state & (1 << degree)) >> degree;
```
bmhowe23, line 134: If degree is >31, this may not work correctly. It should probably be `1ul` to get us op to 63.

runtime/cudaq/operators/spin_op.cpp
```diff
@@ -119,25 +127,54 @@ complex_matrix spin_handler::to_matrix(std::string pauli_word,
...
+    process_element(new_state, old_state, entry);
   }
-  return std::move(matrix);
+}
+
+cudaq::detail::EigenSparseMatrix
+spin_handler::to_sparse_matrix(const std::string &pauli_word,
+                               std::complex<double> coeff, bool invert_order) {
+  using Triplet = Eigen::Triplet<std::complex<double>>;
+  auto dim = 1 << pauli_word.size();
```
bmhowe23, line 148: This should probably be `1ul` to make sure it works up to >31 degrees.

runtime/cudaq/operators/spin_op.cpp
```diff
@@ -119,25 +127,54 @@ complex_matrix spin_handler::to_matrix(std::string pauli_word,
...
+  create_matrix(pauli_word, process_entry, invert_order);
+  cudaq::detail::EigenSparseMatrix matrix(dim, dim);
+  matrix.setFromTriplets(triplets.begin(), triplets.end());
+  return matrix;
+}
+
+complex_matrix spin_handler::to_matrix(const std::string &pauli_word,
+                                       std::complex<double> coeff,
+                                       bool invert_order) {
+  auto dim = 1 << pauli_word.size();
```
bmhowe23, line 165: Same here.

runtime/cudaq/operators/sum_op.cpp
```diff
@@ -1697,39 +1966,70 @@ sum_op<HandlerTy>::sum_op(const std::vector<std::vector<bool>> &bsf_terms,
...
-      if (pauli == pauli::Z)
+  dataVec.reserve(n_targets * padded.terms.size() + 2 * padded.terms.size() +
+                  1);
+  for (std::size_t i = 0; i < padded.terms.size(); ++i) {
+    for (std::size_t j = 0; j < padded.terms[i].size(); ++j) {
+      auto pauli = padded.terms[i][get_le_index(j)].as_pauli();
+      if (pauli == pauli::X)
         dataVec.push_back(1.);
-      else if (pauli == pauli::X)
+      else if (pauli == pauli::Z)
```
bmhowe23, line 1999: It looks like this change means that `getDataRepresentation` no longer returns the same format it used to (X = 1 and Z = 2 now). Am I reading that correctly? If so, that seems strange in a function where we are trying to maintain backwards compatibility.

bmhowe23 (commented):
This is another partial review ... I'm hoping that submitting this one and then starting another one will make my browser load the remaining diffs correctly.

bmhowe23 (commented):
OK, this is the last of my comments. Unfortunately breaking up my review didn't help my browser any ... it refuses to automatically load diffs for more than X files (where X is unknown to me), so I had to manually load them I think that many of the test file changes ended up being unnecessary (simply changing to use `cudaq::spin_op::x` instead of just `x`).

docs/sphinx/examples/cpp/basics/expectation_values.cpp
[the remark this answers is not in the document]
bettinaheim: Yes, they could be reverted indeed. It's a matter of style/preference regarding whether our examples are more explicit regarding what x, y, etc mean or not. I am fine with either.

docs/sphinx/examples/cpp/dynamics/cavity_qed.cpp
```diff
@@ -12,7 +12,7 @@
 // ```
 
 #include "cudaq/algorithms/evolve.h"
-#include "cudaq/dynamics_integrators.h"
+#include "cudaq/algorithms/integrator.h"
```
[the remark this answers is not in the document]
bettinaheim: The one that is user relevant is listed under deprecations (the header indeed is properly deprecated). The other two are headers that should only be used by internal code, I think (nobody should implement their own time stepper, and we don't have any documentation for writing your own integrator, so it is not really considered part of the public API at this point). The

python/runtime/common/py_ObserveResult.cpp
```diff
@@ -75,67 +95,47 @@ to :func:`observe`, return its measurement counts.
 
 Returns:
   :class:`SampleResult`: The measurement counts data for the individual `sub_term`.)#")
+      // FIXME: deprecate
```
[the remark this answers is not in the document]
bettinaheim: I'll tackle all deprecations for python when I make the proper python update.

python/tests/kernel/test_trotter.py
bettinaheim: I believe the previous version merely had a constant offset of 1 that has no physical consequence except shifting the energy by 1. Am I missing something?

runtime/cudaq/operators.h
bettinaheim: It's got significant overhead to check this. This is really just me adding a while bunch of debug assertions for things that the code should obey if the code is written correctly. Or put differently, if a handler says it can be canonicalize, and is_canonicalized is false for a product, then something in our code is wrong.

runtime/nvqir/cutensornet/tensornet_state.cpp
```diff
@@ -791,36 +789,62 @@ std::vector<std::complex<double>> TensorNetState::computeExpVals(
...
+    // Populate the remaining identities.
+    const std::complex<double> *pauliMatrixPtr = PauliI_h;
+    while (offset < numQubits) {
+      auto *address =
+          static_cast<char *>(pauliMats_h) + offset++ * ALIGNMENT_BYTES;
+      std::memcpy(address, pauliMatrixPtr, PAULI_ARRAY_SIZE_BYTES);
+    }
     if (allIdOps) {
-      allExpVals.emplace_back(1.0);
+      allExpVals.emplace_back(prod.evaluate_coefficient());
```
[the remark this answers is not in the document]
bettinaheim: It is a change in the signature of the function where this is called. Previously, it got a bsf without coefficients. Now it gets a vector of product terms that include coefficients. Hence, the coefficient are processed here instead of in the parent function that calls this. This is entirely within the tensornet simulator and not part of any external or overridden base API.

unittests/integration/deuteron_variational_tester.cpp
```diff
@@ -22,43 +22,51 @@ struct ansatz2 {
...
   for (auto &r : results) {
-    test += r.expectation() * r.get_spin().get_coefficient().real();
+    auto spin = r.get_spin();
+    EXPECT_EQ(spin.num_terms(), 1);
+    test += r.expectation(); // expectation should include the coefficient
   }
 
   printf("TEST: %.16lf\n", test);
+  // FIXME: preexisting bug; this test was never testing properly...
+  // it got unnoticed because .expectation silently returns 1.
```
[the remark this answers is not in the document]
bettinaheim: Look at the test: enery is computed from the initial call to observe that passes the whole spin_op rather than a container of terms. This line should say `EXPECT_NEAR(test, -1.7487, 1e-3);` 
I need to double check if it is now passing with this - if it is not, then it would still be the preexisting bug, and I'd prefer to fix it in a separate PR.

python/tests/kernel/test_trotter.py
bmhowe23: I think it's more likely that I am missing something. Wouldn't "shifting the energy by 1" mean that the energy is off by 1?

bmhowe23: Actually, looking more closely, it looks like `tdOp` is just used to perform `exp_pauli` on all the terms (that is - it isn't used as the `observe` spin op), and performing the `exp_pauli` on an identity term will have no effect. I guess we can mark this thread as resolved.

docs/sphinx/applications/python/trotter.ipynb
```diff
@@ -66,7 +66,7 @@
     "\n",
     "\n",
     "def heisenbergModelHam(t: float) -> cudaq.SpinOperator:\n",
-    "    tdOp = cudaq.SpinOperator(num_qubits=n_spins)\n",
+    "    tdOp = cudaq.SpinOperator.empty()\n",
```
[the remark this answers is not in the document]
bmhowe23: Marking this as resolved because as noted in the other thread, this spin op is not the spin op that is used in the `observe` call. This change just means one less `exp_pauli` is called, and since it was the identity term, it had no effect (other than global phase).

unittests/integration/deuteron_variational_tester.cpp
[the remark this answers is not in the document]
bmhowe23: OK, I didn't catch that the line should be using `test` instead of `energy`. It is still failing locally if I change it to `test`, so I'll mark this thread as resolved.

docs/sphinx/conf.py
```diff
@@ -210,6 +210,13 @@ def setup(app):
     ('py:class', 'type'),
     ('py:class', 'cudaq::spin_op'),
     ('py:class', 'numpy.ndarray[]'),
+    # FIXME: remove these after adding proper documentation
+    # (also reexamine why some of the ones above are ignored)
```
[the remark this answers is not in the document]
bettinaheim: To be resolved after https://github.com/NVIDIA/cuda-quantum/pull/2735 completes.

bmhowe23 (approved):
Thanks, Bettina. I know there are some unresolved threads right now, but I'm approving now anyway and will let you judge which ones need to be addressed now vs later (or never).

Also, just to capture our offline discussion, we know this is likely to break compatibility with our currently deployed NVQC servers, but there is a separate plan in place for NVQC going forward.

runtime/cudaq/operators/sum_op.cpp
bettinaheim: It does. it pushes back a 1 if we have an X, and 2 if we have a Z. Not sure what the confusion is?

bmhowe23: > It does. it pushes back a 1 if we have an X, and 2 if we have a Z. Not sure what the confusion is?

The diff with `main` is showing that we used to push back a 1 for Z and a 2 for X (reversed from what this PR is changing it to).

PR #2712, opened 2025-03-07
[core] Add a basic SROA pass.

schweitzpgi:
This adds a basic scalar replacement of aggregates pass. This pass may be useful before running memtoreg in that it converts aggregates to the scalars that memtoreg requires.


test/Quake/basic-sroa.qke
```diff
@@ -0,0 +1,73 @@
...
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -cc-sroa %s | FileCheck %s
+// RUN: cudaq-opt -canonicalize -cc-sroa %s | FileCheck %s
+
+func.func private @g(i32, f64)
+
+func.func @f() {
```
annagrin, line 14: Is it possible to add a test for a nested array as well?

annagrin (approved):
LGTM!

test/Quake/basic-sroa.qke
schweitzpgi: Will do on next PR.

PR #2713, opened 2025-03-07
Improve error handling for analog targets when local emulation is requested

khalatepradnya:
Local emulation is not yet supported on analog hardware targets like `pasqal` and `quera`. Prior to the changes proposed in this PR if user set `emulate=True`, it was ignored and proceeded as if the flag was not present. This change will throw an error.


nvidia-dobri (approved):
LGTM, thanks for getting this in!

PR #2714, opened 2025-03-10
C++ operator API renaming

bettinaheim:
Minimal changes to rename operators, including unifying complex_matrix and matrix_2. Instantiating a complex_matrix now zero-initializes it.

Full list of breaking changes:
- The data is owned by complex_matrix and no longer directly accessible. I added a to_eigen function that should serve the usual purposes for accessing the data. This matrix is in row major format.
- The `matrix.h` header file moved from cudaq/matrix.h to cudaq/utils/matrix.h.
- Eigenvectors for complex_matrix are now returned as *rows*, consistent with the row-major format of the matrix. The previous implementation returned them as columns.
-  Minor changes in the signatures of `virtual SpinMeasureResult measure(const cudaq::spin_op &op) = 0;`, where the `const` was added.


lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -225,10 +225,15 @@ struct AppendMeasurements : public OpRewritePattern<func::FuncOp> {
       return funcOp.emitOpError("Errors encountered in pass analysis");
     auto nQubits = iter->second.nQubits;
 
-    if (nQubits != termBSF.size() / 2)
+    if (nQubits < termBSF.size() / 2) {
```
bettinaheim, line 228: This is the only change that isn't strictly necessary, but I think it would be good to have - it seems correct to me to make this change since the numbers of qubits in the bsf may not match nQubits (if the observed operator doesn't span the whole space). @annagrin for advice.

bmhowe23: What would it mean to observe something like `XZ` on a kernel with >2 qubits? The user might assume it operates on the last 2 qubits (or any other combination of qubits if they are not careful) whereas we silently assume it would mean the first two qubits. It seems like this change introduces opportunities for confusion in my opinion.

bmhowe23:
8fda12fbc338a960edd2fae957fc13782f43f549 successfully ran through the CUDA-QX CI w/o any required CUDA-QX changes. https://github.com/NVIDIA/cudaqx/actions/runs/13768184503?pr=119

lib/Optimizer/Transforms/ObserveAnsatz.cpp
annagrin: This is just an internal fix related to changes in representation of a term. I saw that the first qubits are always padded if missing, we just do not pad the last ones. Here is an example: 

```
Term: Z(0)
term BSF:
01
 
Term: Z(1)
term BSF:
0001

Term: Z(2)
term BSF:
000001 
``` 

So I think it is safe to relax the condition here and just skip the last unmeasured qubits (that would be equivalent to what was done before).

bettinaheim: @annagrin @bmhowe23 This is indeed a bug fix. There is no guarantee that the observable acts on the entire space, and it is not required to. the representation ensures it acts on the right qubits but fails incorrectly when there are more qubits that are not represented because we have no non-trivial observe term on the last qubit.

runtime/cudaq/base_integrator.h
```diff
@@ -18,21 +18,20 @@ namespace cudaq {
...
+      const sum_op<cudaq::matrix_handler> &ham,
+      const std::vector<sum_op<cudaq::matrix_handler>> &cOps = {},
       const std::unordered_map<std::string, std::complex<double>> &params = {})
       : modeExtents(extents), hamiltonian(ham), collapseOps(cOps),
         parameters(params) {}
 
   SystemDynamics()
-      : hamiltonian(operator_sum<cudaq::matrix_operator>(
-            cudaq::matrix_operator::empty())){};
+      : hamiltonian(sum_op<cudaq::matrix_handler>(cudaq::matrix_op::empty())){};
```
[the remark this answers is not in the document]
bettinaheim: It's a type def for sum_op<matrix_handler> in operators.h

runtime/cudaq/utils/matrix.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+/// @brief Hash function for an cudaq::complex_matrix::EigenMatrix
+struct complex_matrix_hash {
+  std::size_t
+  operator()(const cudaq::complex_matrix::EigenMatrix &matrix) const {
+    size_t seed = 0;
+    for (Eigen::Index i = 0; i < matrix.size(); ++i) {
+      auto elem = *(matrix.data() + i);
+      seed ^= std::hash<double>()(elem.real()) +
+              std::hash<double>()(elem.imag()) + 0x9e3779b9 + (seed << 6) +
```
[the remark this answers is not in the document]
bettinaheim: This code is copy pasted from the original matrix.cpp that used to be in the spin ops folder. I don't think these should be user facing, and I don't think they are very general.

1tnguyen (approved):
LGTM 👍

bettinaheim:
Even after the cache rebuild we still have issues with "THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE". I will give it a try to use `pip cache purge`, which I hope might fix it, but I'll pull that in as is, and hopefully resolve the package issue as part of pulling in 2705.

PR #2717, opened 2025-03-11
Mitigating against network issues

bettinaheim:
Undoing some of the build changes I did in https://github.com/NVIDIA/cuda-quantum/pull/2705, and putting in what I hope should mitigate the issue we were seeing.

Keeps opensuse removed from validation (seems to be flaky regarding finding its repo currently - we can check back on that in a week)

### DIFF
--- a/.github/workflows/python_wheels.yml
+++ b/.github/workflows/python_wheels.yml
@@ -188,6 +188,11 @@ jobs:
       matrix:
         os_image: ${{ fromJSON(needs.create_test_config.outputs.json).os_images }}
         pip_install_flags: ['', '--user']
+        exclude: 
+          - os_image: ubuntu:22.04
+            pip_install_flags: '--user'
+          - os_image: ubuntu:24.04
+            pip_install_flags: '--user'
       fail-fast: false
 
     steps:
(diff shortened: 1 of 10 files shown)


.github/workflows/python_wheels.yml
```diff
@@ -193,8 +193,6 @@ jobs:
             pip_install_flags: '--user'
           - os_image: ubuntu:24.04
             pip_install_flags: '--user'
-          - os_image: fedora:39
-            pip_install_flags: ''
```
mitchdz, line 197: perhaps add FIXME to reenable fedora and opensuse?

mitchdz (commented):
overall LGTM.

Just a small nit to add a FIXME to re-enable disabled pipelines.

here is rapidsai using it themselves to help their pipeline - https://github.com/rapidsai/raft/pull/2571

.github/workflows/python_wheels.yml
bettinaheim: fedora is reenabled - maybe you meant ubuntu?
Sure.

mitchdz (approved):

1tnguyen (approved):

PR #2718, opened 2025-03-11
Updating the newly deployed function

sacpis:
Updating the newly deployed function for nightly integration test

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -67,7 +67,7 @@ on:
 env:
   NGC_QUANTUM_ORG: pnyjrcojiblh
   NGC_QUANTUM_TEAM: cuda-quantum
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -67,7 +67,7 @@ on:
 env:
   NGC_QUANTUM_ORG: pnyjrcojiblh
   NGC_QUANTUM_TEAM: cuda-quantum
-  NVQC_FUNCTION_ID: 3bfa0342-7d2a-4f1b-8e81-b6608d28ca7d
+  NVQC_FUNCTION_ID: ac857a08-12df-46e5-addc-ce9d39ae8fd9
```
bmhowe23, line 70: It might be better to click "Make new version" on the existing function rather than create a new function. That would retain the same function ID. It's possible that our service key stored as a GitHub secret is only valid for this 1 function.

PR #2719, opened 2025-03-11
Remove asserts from cudaq_observe_term test

annagrin:
### Description
Try to fix an intermittent failure in "build installers" steps in CI.

We see this test periodically fail in "build installers" steps in CI, where we build against libc++

Remote asserts in the tests and print the failure message instead.

See: https://github.com/NVIDIA/cuda-quantum/issues/1712


annagrin:
> Thanks @annagrin. If removing assert fixes the test, for now its ok. Is that the root cause though?

My suspicion is that the test gives incorrect results intermittently, which used to trigger the assert. If after this PR we see a failure again, we should see if the test just crashed or gave an imprecise result (in which case we can relax the check)

PR #2720, opened 2025-03-11
[core] Add translation of QIR Array -> cudaq::qvector.

schweitzpgi:
When calling back to a noise function from within a kernel (QPU space), the data type of a vector of qubits must be translated from QIR's Array type to CUDA-Q's cudaq::qvector type. This patch does that translation.


annagrin (approved):
LGTM!

PR #2721, opened 2025-03-12
Fixing class names

sacpis:
Fixing class names

### DIFF
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -443,7 +443,7 @@ def GenerateKernelExecution : Pass<"kernel-execution", "mlir::ModuleOp"> {
   ];
 }
 
-def GetConcreteMatrix : Pass<"get-concrete-matrix", "mlir::func::FuncOp"> {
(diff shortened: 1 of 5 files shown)


docs/sphinx/examples/cpp/dynamics/cross_resonance.cpp
```diff
@@ -40,14 +40,14 @@ int main() {
...
-            std::complex<double>(0.0, 1.0) * cudaq::spin_operator::y(degree));
+           (cudaq::spin_handler::x(degree) +
+            std::complex<double>(0.0, 1.0) * cudaq::spin_handler::y(degree));
   };
 
   auto spin_minus = [](int degree) {
     return 0.5 *
-           (cudaq::spin_operator::x(degree) -
-            std::complex<double>(0.0, 1.0) * cudaq::spin_operator::y(degree));
+           (cudaq::spin_handler::x(degree) -
```
amccaskey, line 49: spin_handler? shouldn't this be spin_op::x?

amccaskey: The user should never write code that says "spin_handler". The user facing type is `spin_op`.

PR #2722, opened 2025-03-12
Skipping non-dynamics examples for dynamics target

sacpis:
Skipping non-dynamics examples for dynamics target


1tnguyen (approved):
LGTM 👍 Thanks, @sacpis

PR #2723, opened 2025-03-12
Fixing ctor with a single state

sacpis:
Fixing ctor with a single state


1tnguyen (approved):
Thanks @sacpis for the quick fix.

PR #2725, opened 2025-03-12
Disabling afqmc notebook as ipie lib has changed

sacpis:
Disabling afqmc notebook as ipie lib has changed


docs/sphinx/conf.py
```diff
@@ -93,7 +93,7 @@
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = [
-    '**/_*', '.DS_Store', 'examples/python/building_kernels.ipynb'
+    '**/_*', '.DS_Store', 'examples/python/building_kernels.ipynb', 'applications/python/afqmc.ipynb'
```
bettinaheim, line 96: Please remove non-tested, currently non- or only partially working code from the repo. We'll add it back when it is properly working.

bettinaheim (approved):

Issue #2641, opened 2025-02-20
Testing regression introduced with python/tests/kernel/test_explicit_measurements.py


schweitzpgi:
Testing bug introduced with #2567 .

The tests do not properly skip targets that may or may not exist in custom builds. Specifically, braket and quera targets are correctly opt-in and opt-out and tests should not rely on the assumption those targets are always available.

PR #2732, opened 2025-03-13
[testing] [Python] Fix issue#2641

khalatepradnya:
Make the `braket` and `quera` Python tests conditional - run only if the target is available.
(Note: The `quera` target has a dependency on `braket`)

Addresses: https://github.com/NVIDIA/cuda-quantum/issues/2641


schweitzpgi (approved):
+1

PR #2739, opened 2025-03-17
Some fixes for complex numbers and vectors

annagrin:
### Description
Some fixes for complex numbers and vectors of complex numbers

- make sure we create and read from constant arrays the same way in different scenarios
  (QuakeSynthesis, LiftArrayAlloc, ArgumentSynthesis, ConvertToQIR/eraseConstantArrayOps, etc)
- allow const prop of a complex value from a constant array
- allow using `complex<T>::real()` and `complex<T>::imag()` in kernels in c++ (already supported in python)
- add tests

The fixes are inspired by a trotter example, where we would like to pass a vector of complex numbers to a kernel, be able to get a real part of it, and make sure it works for quantum devices.


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -980,6 +980,14 @@ struct FuseWithConstantArray
 
           return success();
         }
+        if (auto complexTy = dyn_cast<ComplexType>(extval.getType())) {
+          std::int32_t i = extval.getRawConstantIndices()[0];
+          auto cval = cast<ArrayAttr>(conarr.getConstantValues()[i]);
+          rewriter.replaceOpWithNewOp<complex::ConstantOp>(extval, complexTy,
+                                                           cval);
+
```
schweitzpgi, line 988:
```suggestion
```

targettests/execution/complex_vector_synthesis.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// clang-format off
+// RUN: nvq++ %cpp_std --enable-mlir %s -o %t  && %t | FileCheck %s
+// TODO-FIX-KERNEL-EXEC
```
schweitzpgi, line 11: What does this mean? And why not fix it here?

annagrin: Just copied from another test - I believe it was an ask from Ben to add the comment and an issue to update all the tests after we switch to new launch kernel for good. So it will be fixed after:)

annagrin: Found Ben's ask: https://github.com/NVIDIA/cuda-quantum/pull/2094#discussion_r1727102688
It is currently a cleanup item in issue https://github.com/NVIDIA/cuda-quantum/issues/1931

schweitzpgi: OK. We're tracking it somewhere which is sufficient.

schweitzpgi (approved):

docs/sphinx/using/backends/hardware/neutralatom.rst
```diff
@@ -11,8 +11,8 @@ accessed via `Superstaq <https://superstaq.infleqtion.com/>`__, a cross-platform
 that performs low-level compilation and cross-layer optimization. To get started users can create a Superstaq
 account by following `these instructions <https://superstaq.readthedocs.io/en/latest/get_started/credentials.html>`__.
 
-For access to Infleqtion's neutral atom quantum computer, Sqale,
-`pre-registration <https://www.infleqtion.com/sqale-preregistration>`__ is now open.
+For access to Infleqtion's neutral atom quantum computer, Sqale, see details about
+`pre-registration <https://infleqtion.com/infleqtion-delivers-first-quantum-material-design-application-powered-by-logical-qubits-and-nvidia-cuda-q/>`__.
```
annagrin, line 15: Try to fix squale link

annagrin: fixed in https://github.com/NVIDIA/cuda-quantum/pull/2750

PR #2743, opened 2025-03-18
Convert !cc.state to !quake.state

schweitzpgi:
Although it seems reasonable for the cudaq::state object to be classified as a classical compute data structure, it is equally reasonable to classify it as quantum compute since it is only ever used with quantum allocations.

After some discussion, we leaned towards making StateType a member of the quake quantum types.

This PR just moves StateType from the cc dialect to the quake dialect as a result.


annagrin (approved):
LGTM, thanks Eric!

PR #2751, opened 2025-03-20
Nix Flake dev environment for CUDA-Q

nvidia-dobri:
This PR adds a Nix Flake dev environment to the repository, allowing devs to use it instead of managing dependencies manually or using a dev container.

## Why not dev containers?
Dev Containers are perfectly usable. This is a parallel dev environment that provides different pros and cons. Dev containers require a docker runtime and are ephemeral, making it harder to manage e.g. VS Code plugins. I want the simplicity of dev containers, while using my WSL Linux directly.

## Why not manage deps yourself?
Because every developer needs to do that independently, on different systems, in different ways. Before using nix, I never managed to get CUDA-Q to build on a linux host without using dev containers.

## The ideal dev path
I am a user with a brand new Linux box. Nothing is installed, except git. Here's my path to a full CUDA-Q build:
```bash
# Install nix & direnv
sh <(curl -L https://nixos.org/nix/install) --daemon
sudo apt-get install direnv
echo 'eval "$(direnv hook bash)"' >> ~/.bashrc

# Pull repo
git clone git@github.com:NVIDIA/cuda-quantum.git

# Allow the dev environment
cd cuda-quantum
direnv allow

# Install prereq (for now)
bash scripts/install_prerequisites.sh

# Build it!
bash scripts/build_cudaq.sh
```

## Why Nix?
[Nix](https://nixos.org/) is a declarative, ~fully~ mostly hermetic build system. Its main goal is reproducibility, independent of the host machine. Every package and dependency is built from source and linking is managed by Nix, where possible. In an ideal nix-managed environment, the host system configuration and state is irrelevant.

## Why flakes?
Turns out nix is only mostly reproducible. [Flakes](https://nixos.wiki/wiki/Flakes) increase reproducibility by pinning all dependencies in a lock file. This also has a side effect of improving incremental build times. Overall, it makes the environment setup faster and more reliable

## Why pre-commit?
Because we have more important things to do than breaking up files at 80 characters and manually calling clang-format.

## Caveats
This is a minimally-viable dev environment using Nix flakes. It still requires the user to run `install_prerequisites.sh` before being able to build with `build_cudaq.sh`. Since the script installs things on the host outside of the nix store, it is impure and cannot be ran by the flake itself. Additionally, Python is currently unsupported due to some issues with `fixup-linkage` that are probably very fixable.

## Improvements
We should be able to make the `install_prerequisites.sh` script unnecessary when using Nix. That would also imply a fully hermetic and pure dependency build, which would be cool for reproducibility reasons. When we get to that point, we can leverage the [nix build system](https://nix.dev/manual/nix/2.18/command-ref/nix-build) as well, since all the prerequisites will be in the nix store.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
 use flake
+pre-commit install
--- a/.gitignore
+++ b/.gitignore
@@ -70,7 +70,7 @@ __pycache__/
 .cache/*
 
 # LLVM/MLIR files
-*.ll 
+*.ll
 *.bc
 
 # Build results
@@ -93,3 +93,6 @@ apps/
 
 # JetBrains IDE files
 .idea
+
+# direnv
+.direnv
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,8 @@ repos:
     rev: v1.3.0
     hooks:
       - id: nixpkgs-fmt
+  - repo: https://github.com/hukkin/mdformat
+    rev: 0.7.22
+    hooks:
+      - id: mdformat
+        args: ["--wrap", "80"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,14 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v19.1.7
+    rev: v16.0.6
     hooks:
       - id: clang-format
   - repo: https://github.com/google/yapf
     rev: v0.43.0
     hooks:
       - id: yapf-diff
+  - repo: https://github.com/nix-community/nixpkgs-fmt
+    rev: v1.3.0
+    hooks:
+      - id: nixpkgs-fmt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v19.1.7
+    hooks:
+      - id: clang-format
+  - repo: https://github.com/google/yapf
+    rev: v0.43.0
+    hooks:
+      - id: yapf-diff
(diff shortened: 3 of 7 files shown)


schweitzpgi:
It sounds like we have a problem with our cmake configuration if people can't figure out how to `git clone` the repo into a Linux or WSL2 directory, run `cmake` with the necessary options, and build pieces of the project.

nvidia-dobri:
> It sounds like we have a problem with our cmake configuration if people can't figure out how to `git clone` the repo into a Linux or WSL2 directory, run `cmake` with the necessary options, and build pieces of the project.

To clarify, this PR is more about handling all the various required dependencies on the system to even run cmake. Including specific gcc, cmake versions, etc. You can still cmake with the necessary options from the nix dev shell this sets up. There's just a higher degree of confidence that you can have with using this (as long as it remains maintained) than following written instructions on what dependencies to install and setup.

nvidia-dobri:
> Seems like I will have to downgrade Python to 3.10 (have 3.12.4).
> 
> ```
> -- Performing Test C_SUPPORTS_WERROR_IMPLICIT_FUNCTION_DECLARATION
> -- Performing Test C_SUPPORTS_WERROR_IMPLICIT_FUNCTION_DECLARATION - Success
> -- Performing Test C_SUPPORTS_WERROR_MISMATCHED_TAGS
> -- Performing Test C_SUPPORTS_WERROR_MISMATCHED_TAGS - Failed
> -- Found Python3: /usr/bin/python3.10 (found suitable version "3.10.12", minimum required is "3.6") found components: Interpreter Development Development.Module Development.Embed 
> CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
>   suitable version "3.10.12", minimum required is "3.6")
> Call Stack (most recent call first):
>   /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
>   /usr/share/cmake-3.22/Modules/FindPython/Support.cmake:3180 (find_package_handle_standard_args)
>   /usr/share/cmake-3.22/Modules/FindPython3.cmake:490 (include)
>   /root/.llvm-project/mlir/cmake/modules/MLIRDetectPythonEnv.cmake:28 (find_package)
>   /root/.llvm-project/mlir/CMakeLists.txt:159 (mlir_configure_python_dev_packages)
> ```

Hm, there are several issues with Python that I'd rather not solve right now. Did you get this from a `ctest` invocation within the nix shell? Let me try that again, but if you:
* enter the nix shell (`nix develop` or by configuring `direnv`)
* `install_prerequisites.sh`
* `build_cudaq.sh`

Then, `ctest` should pass (with Python support completely disabled).

PR #2752, opened 2025-03-21
[Fix] Add missing headers.

boschmitt:
### Description
Add a couple of missing headers


schweitzpgi (commented):
Curious. Is this producing errors somewhere?

boschmitt:
> Curious. Is this producing errors somewhere?

Quite curious indeed, I would assume is some random error in the CI. Let's see.

PR #2755, opened 2025-03-24
Fix Python MLIR test for OS module

huaweil-nv:
### Description

These tests seem not to be running on CI/CD.


1tnguyen (approved):
LGTM. Thanks @huaweil-nv

PR #2756, opened 2025-03-24
[features/run] Work on getting things working end to end.

schweitzpgi:
Add a test.

Add the pass to convert return types and get end-to-end flow working.

### DIFF
--- a/targettests/execution/cudaq_run.cpp
+++ b/targettests/execution/cudaq_run.cpp
@@ -1,0 +1,68 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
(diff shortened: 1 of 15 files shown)


targettests/execution/cudaq_run.cpp
```diff
@@ -0,0 +1,68 @@
...
+    result |= ((unsigned)w) << (count - 1 - i);
+  }
+  return result;
+}
+
+int main() {
+  int c = 0;
+  {
+    std::vector<int> results =
+        cudaq::run<int>(100, std::function<int()>{nullary_test});
```
amccaskey, line 41: shouldn't the `<int>` in `run<int>` be unnecessary (inferred by the passed function return type)?

schweitzpgi: Correct, It isn't required.

PR #2765, opened 2025-03-31
Allow Return Value Optimization instead of forcing moves

bmhowe23:
This PR changes some of the operator functions to not use `std::move` when it is preventing Return Value Optimization, as that causes additional unnecessary processing.

### DIFF
--- a/runtime/cudaq/operators.h
+++ b/runtime/cudaq/operators.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <functional>
+#include <random>
 #include <set>
 #include <type_traits>
 #include <unordered_map>
@@ -17,7 +18,6 @@
 #include "operators/evaluation.h"
 #include "operators/operator_leafs.h"
 #include "operators/templates.h"
-#include "utils/cudaq_utils.h"
 #include "utils/matrix.h"
 
 namespace cudaq {
--- a/runtime/cudaq/operators/boson_op.cpp
+++ b/runtime/cudaq/operators/boson_op.cpp
@@ -47,7 +47,7 @@ std::string boson_handler::op_code_to_string() const {
     str += "Ad";
   for (auto i = 0; i > this->additional_terms; --i)
     str += "A";
-  return std::move(str);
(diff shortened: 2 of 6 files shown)


schweitzpgi (approved):
Is this the only code that is using `std::move` with return values?

One should never do that.

bettinaheim (approved):

bmhowe23:
> Is this the only code that is using `std::move` with return values?

A few others remain, but they mostly either move function parameters or struct members, which cannot use RVO. This PR just changes ones that were moving local variables. Some `std::move(*this)`'s remain...I didn't try to change those. If necessary, those could be investigated in a follow-up PR.

schweitzpgi:
> > Is this the only code that is using `std::move` with return values?
> 
> A few others remain, but they mostly either move function parameters or struct members, which cannot use RVO. This PR just changes ones that were moving local variables. Some `std::move(*this)`'s remain...I didn't try to change those. If necessary, those could be investigated in a follow-up PR.

While there are obviously cases where the compiler may not use RVO, we still should not be writing code that uses `return std::move(e);` regardless. The only case it *might* make sense is if the move-constructor has some special side-effect that is required for correctness—but that case should probably be handled in a less obscure way too.

bmhowe23:
> While there are obviously cases where the compiler may not use RVO, we still should not be writing code that uses `return std::move(e);` regardless. The only case it _might_ make sense is if the move-constructor has some special side-effect that is required for correctness—but that case should probably be handled in a less obscure way too.
> 
> So ... I guess we need an issue to follow up on these?

It seems perfectly valid to me to use `return std::move(...)` for performance reasons in some cases. (The point of this PR was just to remove the cases where I thought it was hurting performance.) I'm not sure I follow why we would categorically say that isn't allowed, though. E.g. https://github.com/NVIDIA/cuda-quantum/blob/d54b75b24ef304a953ad509a01cfa535be2f88d8/runtime/cudaq/operators/product_op.cpp#L673 seems perfectly valid to me.

PR #2767, opened 2025-03-31
[py] Fix loop invariant attribute.

boschmitt:
### Description
Python's for-loops are not guaranteed to be invariant. The AST bridge would need to do further analysis to guarantee the property. This PR does not add such an analysis, but fixes the places where the bridge might wrongly label a for-loop invariant when, in reality, it might not be.


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3272,11 +3275,11 @@ def bodyBuilder(iterVar):
...
-                self.createInvariantForLoop(endVal,
-                                            bodyBuilder,
-                                            startVal=startVal,
-                                            stepVal=stepVal,
-                                            isDecrementing=isDecrementing)
+                self.createForLoop(endVal,
+                                   bodyBuilder,
+                                   startVal=startVal,
+                                   stepVal=stepVal,
+                                   isDecrementing=isDecrementing)
```
schweitzpgi, line 3282: Something like this must be normalized. Is the cc-loop-normalize pass being run?

boschmitt: Hard to know. The pass does get run in various places but I'm not sure if it cover all paths used to execute Python kernels. Is this pass a requirement in all cases or only when doing other loop transformations ?

schweitzpgi: Normalization transforms the iteration space of a loop to a counted or invariant form.

More precisely, if we have a loop like
```c++
for (i = start_expr; comparison(i, stop_expr); i = step_expr(i)) ....
```
this loop in normalized form is more like
```c++
for (i' = 0; i' < new_uppser_bound_expr; i' = i' + 1) {
   i = some_expr(i');
   ....
 } 
 ```
 
 Once in this form, the loop is much more amenable to loop transformations such as unrolling.

schweitzpgi: So it is not a requirement for loops that are already normalized, but if Python is creating a loop that descends or uses a step other than 1, the loop must be normalized.

Normalization ought to be part of the new "classical optimizations" pipeline by default.

PR #2770, opened 2025-03-31
Use default constructors and assignment operators

bmhowe23:
This PR reduces some boilerplate code by letting the compiler do the auto-generation of constructors and (move) assignment operators where appropriate.

It also fixes a pre-existing matrix * vector bug where the returned vector was not setting the size correctly.


runtime/cudaq/utils/matrix.cpp
```diff
@@ -147,13 +148,11 @@ cudaq::operator*(const cudaq::complex_matrix &matrix,
     throw std::runtime_error("size mismatch for vector multiplication - "
                              "expecting a vector of length " +
                              std::to_string(matrix.cols()));
-  std::vector<cudaq::complex_matrix::value_type> res;
-  res.reserve(matrix.rows());
-  for (std::size_t i = 0; i < matrix.rows(); i++) {
-    res[i] = 0.;
+  // Initialized res to the correct size and set all elements to 0
+  std::vector<cudaq::complex_matrix::value_type> res(matrix.rows());
```
boschmitt, line 152: Missing the `0` initialization.

bmhowe23: That is implied with the std::vector constructor. (As a side note, I actually which one could easily disable that, but I haven't found a good way to prevent C++ from doing that without jumping through major hoops.)

Ref: https://en.cppreference.com/w/cpp/language/value_initialization
> All standard containers ([std::vector](https://en.cppreference.com/w/cpp/container/vector), [std::list](https://en.cppreference.com/w/cpp/container/list), etc.) value-initialize their elements when constructed with a single size_type argument or when grown by a call to resize(), unless their allocator customizes the behavior of construct.

bmhowe23: Yes, that would work if you're willing to create a new type for the cases where you want to skip initialization, but it requires more work to make it really functional in full use cases. For example, I could do

```cpp
struct MyDouble {
  double data;
  MyDouble() {}
};
// std::vector<double> vec(1000000); // initialized with all 0's
std::vector<MyDouble> vec(1000000); // uninitialized (good)
```

But then I would have to change all downstream usages of `vec[i]` to `vec[i].data` (or add additional operator overloads into `MyDouble`), both of which are annoying to me. I just really wish one could do something like this (below) with primitive types, as it would allow usage with minimal other changes. All other ways I've seen to do this require creating custom types.
```cpp
std::vector<double> vec;
vec.reserve_and_set_size(1000000);
// Or this
vec.resize_uninitialized(1000000);
```

All this means is that - to my knowledge - C++ w/ `std::vector` will often do 2 passes over memory when you want to custom initialize things, whereas it's trivial to avoid that in C. I.e.

```cpp
// One pass for initialization in C
int *arr = malloc(sizeof(int) * 1000);
for (int i = 0; i < 1000; i++)
  arr[i] = 2*i; // some custom initialization
// Two passes for initialization in C++
std::vector<int> vec(1000); // one pass to initialize everything to 0's
for (int i = 0; i < 1000; i++)
  vec[i] = 2*i; // another pass
```

(I believe there is also a way to do it with custom allocators in C++, but I have also found that overly verbose when I just wish `resize_uninitialized()` existed.)

_Stepping off my soap box now._ 😄

In any case, do you agree that the code in this PR is properly initializing everything to 0's? If so, I'll mark the thread as resolved. If not, let me know.

boschmitt: > In any case, do you agree that the code in this PR is properly initializing everything to 0's? If so, I'll mark the thread as resolved. If not, let me know.

Sure, I agree, as long as we don't do anything funny with `cudaq::complex_matrix::value_type`, which I assume it wound't be the case.

1tnguyen (approved):
LGTM 👍

PR #2771, opened 2025-03-31
Update .clang-format: set InsertNewlineAtEOF to Yes

bmhowe23:
Help address/prevent the "nit: missing newline at end of file" comments.

Unfortunately, I saw no such option for yapf (Python), but IMO, that shouldn't prevent us from doing that for our C++ files.


1tnguyen (approved):
💯

PR #2772, opened 2025-04-01
Add `run` support for `remote-mqpu` platform

1tnguyen:
### Description

- Delegate the run loop to the server side.

- Return the QIR output string to the client.

- Add test.

### DIFF
--- a/runtime/cudaq/algorithms/run.h
+++ b/runtime/cudaq/algorithms/run.h
@@ -42,7 +42,8 @@ struct RunResultSpan {
 // pointer to a buffer and the size of that buffer in bytes.
 RunResultSpan runTheKernel(std::function<void()> &&kernel,
                            quantum_platform &platform,
-                           const std::string &kernel_name, std::size_t shots);
+                           const std::string &kernel_name, std::size_t shots,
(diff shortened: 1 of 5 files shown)


runtime/cudaq/algorithms/run.h
```diff
@@ -97,15 +98,19 @@ std::vector<RESULT> run(std::size_t shots,
                         ARGS &&...args) {
   if (shots == 0)
     return {};
-
+#if defined(CUDAQ_REMOTE_SIM)
```
schweitzpgi, line 101: Shouldn't this depend on the target selected (i.e. dynamic selection) instead of being baked in as a constant at runtime compile-time?

1tnguyen: Yea, it should. Thanks to @annagrin, I'm now aware of the `get_remote_capabilities().isRemoteSimulator` field :)  
Fixed in https://github.com/NVIDIA/cuda-quantum/pull/2772/commits/9c9f1ba14c08cd2b81d7531690bd8cc90792aac1

schweitzpgi (approved):

PR #2773, opened 2025-04-01
Point to CUDA-Q Academics quick start notebooks

efratshabtai:
Pointing to CUDA-Q academic quick start notebooks


efratshabtai:
We discussed a better solution of pointing to CUDA-Q academics and other cloud options from CUDA-Q quick start @marwafar

Issue #2760, opened 2025-03-26
Sporadic `RuntimeError: status 15` error on the `nvidia-mgpu` target


1tnguyen:
When using the `nvidia` target with multiple processes (GPUs), we might hit a `RuntimeError: status 15` error, e.g.,

```
File "/opt/nvidia/cudaq/cudaq/runtime/sample.py", line 102, in sample
    cudaq_runtime.resetExecutionContext()
RuntimeError: status 15
```

**Potential workaround:**

Setting the random seed explicitly, e.g.,  `cudaq.set_random_seed(123)`

Issue #2749, opened 2025-03-20
GH Unified memory error


zohimchandani:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

In the latest container and running on a GH system: 

```
import os 
os.environ["CUDAQ_MAX_CPU_MEMORY_GB"] = "NONE"
print("CUDAQ_MAX_CPU_MEMORY_GB:", os.environ.get("CUDAQ_MAX_CPU_MEMORY_GB"))

import cudaq

cudaq.set_target('nvidia', option='mgpu,fp32')

# cudaq.set_target('nvidia')

cudaq.mpi.initialize()

print('mpi initialized?', cudaq.mpi.is_initialized())

n_qubits = 35

@cudaq.kernel
def kernel(n_qubits:int):
    
    qubits = cudaq.qvector(n_qubits)
    
    x(qubits)
    h(qubits)    
    y(qubits)

print('starting exp val calculation')

expectation_value = cudaq.observe(kernel, cudaq.spin.z(0), n_qubits)
# expectation_value = cudaq.sample(kernel, n_qubits)

print("num qubits", n_qubits, 'expectation_value', expectation_value)

cudaq.mpi.finalize()
```

`cudaq.sample()` works but `observe` throws the following error: 

```

python3 hello_world.py --cudaq-full-stack-trace
CUDAQ_MAX_CPU_MEMORY_GB: NONE
[lego-cg1-qs-141:1715226] PMIX ERROR: ERROR in file gds_ds12_lock_pthread.c at line 168
mpi initialized? True
starting exp val calculation
Traceback (most recent call last):
  File "/demo/hello_world.py", line 28, in <module>
    expectation_value = cudaq.observe(kernel, cudaq.spin.z(0), n_qubits)
  File "/opt/nvidia/cudaq/cudaq/runtime/observe.py", line 136, in observe
    cudaq_runtime.resetExecutionContext()
[...]

PR #2774, opened 2025-04-02
Update mgpu sha and docs

1tnguyen:
### Description

- Add an option to change the mgpu backend transfer buffer size and the corresponding doc.

- Add the doc for the option to enable/disable math mode emulation (fp32).

- Fix a bug in multi-process random seed initialization (in 0.10) - https://github.com/NVIDIA/cuda-quantum/issues/2760

- Initial state vector allocation optimization.

- Handle certain cases (relatively sparse Pauli operators) in host-device simulation at high memory utilization: https://github.com/NVIDIA/cuda-quantum/issues/2749. Adding a note in the docs regarding this use case.


bmhowe23 (approved):
+1

PR #2780, opened 2025-04-03
Fix crashes when using captured vectors and states not supported in synthesis

annagrin:
Fix crashes when using captured vectors and states not supported in synthesis

### Description
We observed some cache clashing and crashed when using captures and kernels with the same name.

Make sure to create the capture cache correctly, add tests


schweitzpgi (approved):
LGTM. Thanks for the bug fixes.

PR #2782, opened 2025-04-04
cudaq::run support for hardware emulation

1tnguyen:
### Description

- Add output recording runtime support for hardware QPU emulation. 

- Add support for integer/float computations extensions. This determines how we verify the QIR.

- Enable integer computations for Quantinuum target.

- Add a test.


runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -227,8 +227,10 @@ mlir::LogicalResult verifyOutputRecordingFunctions(llvm::Module *llvmModule) {
         auto callInst = llvm::dyn_cast_or_null<llvm::CallBase>(&inst);
         auto func = callInst ? callInst->getCalledFunction() : nullptr;
         // All call arguments must be constants
-        if (func && failed(verifyConstArguments(callInst)))
-          return mlir::failure();
+        // FIXME: remove the below block once we can confirm that const argument
+        // is not a requirement. if (func &&
+        // failed(verifyConstArguments(callInst)))
+        //   return mlir::failure();
```
bmhowe23, line 233: The formatting did something weird here, but regardless, I'd recommend limiting keeping this check for the base profile.
```suggestion
        if (isBaseProfile && func && failed(verifyConstArguments(callInst)))
          return mlir::failure();
```

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -712,7 +763,7 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
             if (hasConditionals && isObserve)
               throw std::runtime_error("error: spin_ops not yet supported with "
                                        "kernels containing conditionals");
-            if (hasConditionals) {
+            if (!isRun && hasConditionals) {
```
bmhowe23, line 766: I'm curious why this isn't `if (isRun || hasConditionals)`.

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -328,8 +332,53 @@ mlir::LogicalResult verifyLLVMInstructions(llvm::Module *llvmModule,
           llvm::errs() << "error - invalid instruction found: " << inst << '\n';
           return mlir::failure();
         } else if (isAdaptiveProfile && !isValidAdaptiveProfileInstruction) {
-          llvm::errs() << "error - invalid instruction found: " << inst << '\n';
-          return mlir::failure();
+          // Not a valid adaptive profile instruction
```
bmhowe23, line 335: Would you please update the comments in line 321-325 and potentially remove the commented out code in lines 327-330?

bmhowe23 (approved):
Looks good. Thanks, Thien.

runtime/common/BaseRemoteRESTQPU.h
1tnguyen: Indeed, we can make it like that and move the handling of `isRun` to within the loop.
i.e., escaping the set/reset execution context and accumulating the `counts`.

When I first wrote this, I kept the existing code as-is to ease debugging, but looking back at it, I think we can indeed combine. I'll update it.

runtime/common/RuntimeMLIRCommonImpl.h
1tnguyen: Sure, done in https://github.com/NVIDIA/cuda-quantum/pull/2782/commits/f72b23a946abb8efb883022a480814815dc4075a

targettests/execution/cudaq_run_emulation.cpp
```diff
@@ -0,0 +1,51 @@
...
+    if (mz(v[i]))
+      result += 1;
+  return result;
+}
+
+int main() {
+  int c = 0;
+  {
+    std::vector<int> results =
+        cudaq::run<int>(100, std::function<int(int)>{test_kernel}, 4);
```
amccaskey, line 34: can you update this such that you don't require the template int parameter to be specified, e.g.
```cpp 
auto results = cudaq::run(100, test_kernel, 4); 
// -or-
std::vector<int> results = cudaq::run(100, test_kernel, 4); 
```
may just be that this is test code and we're being explicit, but I'd prefer the user not have to write this much boilerplate.

1tnguyen: Absolutely. It's on my radar. I'll tackle it in the following PR :)

PR #2783, opened 2025-04-04
[core] Add quake simplify pass.

schweitzpgi:
This pass adds some simple quantum optimizations.


lib/Optimizer/Transforms/QuakeSimplify.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    auto targets = qop.getTargets();
+    if (targets.size() != 2 ||
+        !quake::isQuantumValueType(targets[0].getType()) ||
+        !quake::isQuantumValueType(targets[1].getType())) {
+      LLVM_DEBUG(llvm::dbgs() << "operation must have 2 targets\n");
+      return failure();
+    }
+
+    // Check that these are the same rotation op back-to-back.
```
bmhowe23, line 122:
```suggestion
    // Check that these are the same swap op back-to-back.
```

test/Quake/quake_simplify.qke
```diff
@@ -0,0 +1,156 @@
...
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --quake-simplify %s | FileCheck %s
+
+func.func @simp1() -> (i1, i1) {
+  %0 = quake.null_wire
+  %1 = quake.null_wire
+  %2:2 = quake.x [%0] %1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
+  %3:2 = quake.x [%2#0] %2#1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
```
bmhowe23, line 15: It might be worth adding some negative tests to make sure the pass doesn't accidentally simplify X, Y, X to just Y. (I'm not saying that it's behaving in correctly right now ... just thinking it would be good to prove it via testing.)

test/Quake/quake_simplify.qke
```diff
@@ -0,0 +1,156 @@
...
+// CHECK:           quake.sink %[[VAL_3]] : !quake.wire
+// CHECK:           quake.sink %[[VAL_5]] : !quake.wire
+// CHECK:           return %[[VAL_6]], %[[VAL_7]] : i1, i1
+// CHECK:         }
+
+func.func @simp5() -> (i1, i1) {
+  %0 = quake.null_wire
+  %1 = quake.null_wire
+  %2:2 = quake.swap %0, %1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
+  %3:2 = quake.swap %2#1, %2#0 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
```
bmhowe23, line 41: Reading tests like this makes me wonder what would happen if `apply_noise` were in between those two gates. I think memtoreg prevents that from happening?

lib/Optimizer/Transforms/QuakeSimplify.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    auto targets = qop.getTargets();
+    if (targets.size() != 1 ||
+        !quake::isQuantumValueType(targets[0].getType())) {
+      LLVM_DEBUG(llvm::dbgs() << "operation must have 1 target\n");
+      return failure();
+    }
+    Value trgt = targets[0];
+
+    // Check that these are the same rotation op back-to-back.
```
bmhowe23, line 47:
```suggestion
    // Check that these are the same Hermitian op back-to-back.
```

bmhowe23 (commented):
LGTM. Might be good to have @boschmitt review as well.

test/Quake/quake_simplify.qke
boschmitt: As we are dealing with value semantics, then it would mean that the gates won't be connected. When looking at `prev` you would get the `apply_noise` operation and the patterns would fail.

test/Quake/quake_simplify.qke
boschmitt: +1

lib/Optimizer/Transforms/QuakeSimplify.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+template <typename QOP>
+class HermitianElimination : public OpRewritePattern<QOP> {
+public:
+  using Base = OpRewritePattern<QOP>;
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(QOP qop,
+                                PatternRewriter &rewriter) const override {
+    if (qop.getNegatedQubitControls())
```
boschmitt, line 36: Don't we need to check for negatedQubitControls on the `prev` operation too?

schweitzpgi: Yes. Good catch.

schweitzpgi: Added f3b3046f

test/Quake/quake_simplify.qke
schweitzpgi: I'll add some negative tests. (Perhaps ironically, the sequence XYX on a qubit can be simplified to just Y. Though I have not implemented all such identity transforms here. To be thorough, we'd need to introduce a negative op, -I, though.)

schweitzpgi:
ping for approval

bmhowe23 (approved):
LGTM

PR #2788, opened 2025-04-07
Make `cudaq::run` API template easier to use

1tnguyen:
### Description

- Adjust the API so that we don't need to construct a `std::function` explicitly.

- Add `async` API.

- Adjust existing tests and add `async` tests.

### DIFF
--- a/runtime/cudaq/algorithms/run.h
+++ b/runtime/cudaq/algorithms/run.h
@@ -14,6 +14,7 @@
 #include "cudaq/algorithms/broadcast.h"
 #include "cudaq/concepts.h"
 #include "cudaq/host_config.h"
+#include "cudaq/platform/QuantumExecutionQueue.h"
 #include <cstdint>
 
 extern "C" {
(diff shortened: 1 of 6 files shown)


runtime/cudaq/algorithms/run.h
```diff
@@ -103,14 +111,234 @@ std::vector<RESULT> run(std::size_t shots,
...
+            platform, kernelName, shots);
+        std::vector<std::invoke_result_t<std::decay_t<QuantumKernel>,
+                                         std::decay_t<ARGS>...>>
+            results{
+                reinterpret_cast<ResultTy *>(span.data),
+                reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
+        p.set_value(std::move(results));
+      });
+#else
+  QuantumTask wrapped = detail::make_copyable_function(
```
schweitzpgi, line 220: We may be able to simply not support versions of C++ prior to C++20.

schweitzpgi (approved):
Thanks for cleaning up the header file, Thien. LGTM.

Issue #2798, opened 2025-04-09
The python functioncudaq.sample_async does not work with any cudaq.kernel using conditional branches.


irfankhan10:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The python function `cudaq.sample_async` does not work with any `cudaq.kernel` using conditional branches.

`pybind11::handle::dec_ref() is being called while the GIL is either not held or invalid. Please see https://pybind11.readthedocs.io/en/stable/advanced/misc.html#common-sources-of-global-interpreter-lock-errors for debugging advice.
If you are convinced there is no bug in your code, you can #define PYBIND11_NO_ASSERT_GIL_HELD_INCREF_DECREF to disable this check. In that case you have to ensure this #define is consistently used for all translation units linked into a given pybind11 extension, otherwise there will be ODR violations. The failing pybind11::handle::dec_ref() call was triggered on a PyKernelDecorator object.`

`cudaq.sample` does work with `cudaq.kernel` using conditional branching.  

### Steps to reproduce the bug
```
import cudaq
import numpy as np

@cudaq.kernel
def verification(n_qubits: int):
    qvector = cudaq.qvector(n_qubits)
    qvector_reversed = qvector[::-1]
    for qi, qj in zip(qvector_reversed[:-1], qvector[1:]):
        x.ctrl(qi, qj)
    bz = mz(qvector[0])
    reset(qvector[0])
    if bz:
        x(qvector[0])
    for qi, qj in zip(qvector[:-1], qvector[1:]):
        x.ctrl(qi, qj)
    mz(qvector)

cudaq.set_target("quantinuum", machine="H1-1SC")

syntax_check = cudaq.sample_async(verification, shots_count=100)
```
### Expected behavior

The function should submit the kernel without error and return a `cudaq.AsyncResult` object

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
[...]

PR #2800, opened 2025-04-10
Fixing `sample_async` for dynamic kernels

1tnguyen:
### Description

The code path causing the issue is: 

https://github.com/NVIDIA/cuda-quantum/blob/91953652e36c8dd463c1d674bde76af8cc78a9f7/python/runtime/cudaq/algorithms/py_sample_async.cpp#L83-L87 

`kernel.dec_ref();` shouldn't be called after `py::gil_scoped_release`. The fact that we put this inside the `hasQubitMeasurementFeedback` condition (rather than the main code path) is probably an oversight; hence, this bug only showed up in this conditional circuit case.

When looking at this, I think the mechanism of using a static `localShots` variable to determine when to delete the `args` is error-prone. For example, when we call `sample_async` on the same kernel multiple times, the static variable may be persisted between calls.

~~Hence, fixing these 2 issues by using `unique_ptr` to hold the args and kernel python object and attaching them to the async functor. For the kernel object, use a custom deleter to decrement the ref count.~~

Hence, using `unique_ptr` to hold the args (C++ args) and attaching them to the async functor.

For the issue with kernel object lifetime, doing a `dec_ref` on a Python object (requiring GIL acquire) may sporadically cause a deadlock as the `QuantumExecutionQueue` exits (joining the execution thread).

This might be related to https://github.com/pybind/pybind11/issues/1446.
Interestingly, this only occurs to wheels built as part of `assets` build, not the regular wheel pipeline.
Perhaps the destructor ordering is somewhat different between the two.

Hence, I refactored the code to not rely on the ref count manipulation. Instead, the Python's MLIR `Context`, which is truly what needs to be alive, is attached to the Python async handle returned to user code.  

Also, fixing an issue with conditional detection, `runSampling` relies on quake code in the registry (`cudaq::kernelHasConditionalFeedback`) to determine the execution mode. Hence, make sure we populate the registry with the quake code.

Adding a test case and also making sure the test case for issue 1218 is still passing.

Fixed: https://github.com/NVIDIA/cuda-quantum/issues/2798


bmhowe23 (approved):
Nice detective work, @1tnguyen!

python/tests/builder/test_sample.py
boschmitt, line 1: This test shouldn't be put in the `builder/` directory, it doesn't use the kernel builder.

python/runtime/cudaq/algorithms/py_sample_async.cpp
```diff
@@ -50,43 +53,68 @@ for more information on this programming pattern.)#")
...
-            },
+            // Notes:
+            // (1) no Python data access is allowed in this lambda body.
+            // (2) This lambda might be executed multiple times, e.g, when the
+            // kernel contains measurement feedback.
+            cudaq::detail::make_copyable_function(
+                [argData = std::move(argData), kernel = std::move(kernelPtr),
+                 kernelName, kernelMod]() mutable {
+                  pyAltLaunchKernel(kernelName, kernelMod, *argData, {});
+                }),
```
boschmitt, line 117: I don't understand how this can work. With `argData` and `kernel` being `std::unique_ptr`s, this lambda should not be copyable.

boschmitt (commented):
Nice work Thien.

python/runtime/cudaq/algorithms/py_sample_async.cpp
```diff
@@ -50,43 +53,68 @@ for more information on this programming pattern.)#")
...
+        // this kernel has qubit measurement feedback on the first invocation.
+        if (cudaq::get_quake_by_name(kernelName, false, std::nullopt).empty()) {
+          auto moduleOp = unwrap(kernelMod);
+          std::string mlirCode;
+          llvm::raw_string_ostream outStr(mlirCode);
+          mlir::OpPrintingFlags opf;
+          moduleOp.print(outStr, opf);
+          cudaq::registry::__cudaq_deviceCodeHolderAdd(kernelName.c_str(),
+                                                       mlirCode.c_str());
+        }
```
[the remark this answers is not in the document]
1tnguyen: Python kernels executed via C++ binding (`pyAltLaunchKernel`) will eventually register themselves after JIT'ing, i.e., calling the `init_func` and hence update the C++ registry.

To handle the case where the Python kernel might change, like in Jupyter notebook workflows, I think we can either

(1) remove the `if` check and always update the registry, or

(2) pass the `ConditionalFeedback` flag (as an additional optional argument) through `runSamplingAsync` -> `runSampling`. This flag can be detected here based on the MLIR module. `runSampling` uses it if available, rather than relying `cudaq::kernelHasConditionalFeedback`


(1) may incur some runtime overhead. (2) requires a bit more plumbing work :) 

What do you think?

python/tests/builder/test_sample.py
1tnguyen: Good catch! I accidentally put it there to test the fix for this issue and `1218` at the same time. 
Fixed in 8c1fa2f28bcfbd0085d14f20b52df1a602128e42

python/runtime/cudaq/algorithms/py_sample_async.cpp
boschmitt: It feels like a misleading behavior. We don't think we need to here, but it would be good to revisit this function, at least change its name to something that clearly signals that different copies of the lambda are not being created.

python/runtime/cudaq/algorithms/py_sample_async.cpp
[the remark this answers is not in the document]
boschmitt: > What do you think?

Hard to give a good answer without understanding had bad would be the performance implications of (1) (if any). Perhaps @annagrin have better insight here.

python/runtime/cudaq/algorithms/py_sample_async.cpp
1tnguyen: Agree! I added https://github.com/NVIDIA/cuda-quantum/issues/2814 to track this discussion.

python/runtime/cudaq/algorithms/py_sample_async.cpp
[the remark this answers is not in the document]
1tnguyen: I'd propose we follow a simple approach of always registering the kernel code here for robustness.
As @annagrin mentioned, the actual code that will be registered in `pyAltLaunchKernel` may be different due to optimization, hence we should let it override as is. The need to look at the kernel code to figure out how to call the kernel is specific to `sample`; thus, keeping the change local here makes sense, too.  

> we also need to verify if that does not cause races


This is indeed a consideration for all async. APIs that involve JIT'ing. We have a mutex lock around the registry to regulate it. 


I posted https://github.com/NVIDIA/cuda-quantum/pull/2800/commits/f87c7a94ef7e1fc5b3cf8d7ca27dd81f7bf7fefc to implement the simple approach (1).

bmhowe23 (approved):
LGTM. Thanks, Thien!

PR #2801, opened 2025-04-10
workflow tweaks

bettinaheim:


mitchdz (approved):
Thanks Bettina.

I think this is good enough for what we want.

While it's on my mind - If we really want to filter more here, we could scrape the `base_ref` from `metadata.txt` that was created in `PR merged`.

bmhowe23 (approved):
LGTM

PR #2802, opened 2025-04-10
Docs: CUDA-Q Academic page/ Dynamics examples

mawolf2023:
### Description
1. Added CUDA-Q Academic page w/ links to its repo
2. Added Dynamics examples to examples section w/ a crosslink in the Dynamics section
3. Added Shor's notebook which was accidently removed in application page update.

For viz: @efratshabtai


docs/sphinx/using/cudaq_academic/cudaq_academic.rst
```diff
@@ -0,0 +1,13 @@
+.. _cuda_academic:
+
+CUDA-Q Academic
+=================
+
+`CUDA-Q Academic <https://github.com/NVIDIA/cuda-q-academic>`_  is a collection of courses designed for academic use based on the CUDA-Q platform.  Courses range in complexity from the basics of quantum computing up to implementations of algorithms from the research literature.  Users will learn quantum computing, CUDA-Q, and how to run hybrid algorithms that make the most of QPUs and AI supercomputers.  
+
+The courses can be run using a local installation of CUDA-Q or via the cloud using qBraid, CoCalc, or Google Colab.
+
+Users just learning quantum computing are encouraged to begin with the `Quick Start to Quantum <https://github.com/NVIDIA/cuda-q-academic/tree/main/quick-start-to-quantum>`_ course which provides an introduction to the basics of quantum computing and helps you implement your first algorithm.
```
efratshabtai, line 10: I think it's better to point to https://github.com/NVIDIA/cuda-q-academic/blob/main/quick-start-to-quantum/READ_ME.md that has a table with launch options

docs/sphinx/using/cudaq_academic/cudaq_academic.rst
```diff
@@ -0,0 +1,13 @@
+.. _cuda_academic:
+
+CUDA-Q Academic
```
efratshabtai, line 3: Reading the rendered page, it feels a bit disconnected to be a page of its own as it's short and just points to other resources. 
Maybe instead we can point to the QC [intro ](https://github.com/NVIDIA/cuda-q-academic/blob/main/quick-start-to-quantum/READ_ME.md ) from Quick Start before the Install CUDA-Q section (not sure if it should be a note or another section).
We can also point to other courses in relevant notebooks in the applications folder, for example, point to the max-cut course from the max-cut application

docs/sphinx/using/cudaq_academic/cudaq_academic.rst
mawolf2023: Agreed

docs/sphinx/using/cudaq_academic/cudaq_academic.rst
mawolf2023: @efratshabtai You are right, but that was sort of my goal.  Just a more prominent shoutout to draw attention, explain what it is, and point users to it.   Adding a note in the quick start is a good idea, but I also worry that the quick start section is not enough visibility on its own.  I am happy either way.  If you feel it does not justify its own page no problem removing. 

Yeah, agreed with adding links from notebooks.

efratshabtai: For awareness, we already point to CUDA-Q academics from the CUDA-Q webpage and other resources. I think 'quick start' is the first page that someone will visit if they are new to CUDA-Q so it makes sense to me to point a no install option there (together with others in the future), one such an example happens to be the QC intro course from academics.

mawolf2023: @efratshabtai Sure, I can remove the page and add something on the quick start page.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -263,6 +267,7 @@ lossy
 lvalue
 makefiles
 merchantability
+mgmn
```
schweitzpgi, line 270: Code artifacts should be quoted in backquotes (or whatever is appropriate) rather than added as English words here.

schweitzpgi (commented):
Seems like a good example to return to the fold.

.github/workflows/config/spelling_allowlist.txt
schweitzpgi: Can you revisit all of these words being added? Some others look suspiciously like code artifacts to me. Thanks.

1tnguyen (approved):

docs/sphinx/using/cudaq_academic/cudaq_academic.rst
mawolf2023: @sacpis Yes, I swapped the link.

docs/sphinx/using/cudaq_academic/cudaq_academic.rst
mawolf2023: @sacpis Yes, removed the CUDA-Q Academic page on the main TOC and instead mentioned in Quick Start

Issue #2804, opened 2025-04-10
Incorrect bool to integer cast


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

`(unsigned)(true)` inside kernels is translated to a `-1` constant instead of `1`

### Steps to reproduce the bug

```c++
struct testConstBool {
  auto operator()() __qpu__ {
    cudaq::qubit q;
    unsigned i = (unsigned)(true);
    if (i == 1) {
      x(q);
    }
  }
};

int main() {
  auto counts = cudaq::sample(testConstBool{});
  counts.dump();
}
```

Result:
`{ 0: 1000 }`

### Expected behavior

`{ 1: 1000 }`

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #2805, opened 2025-04-10
Fix constant folding of booleans casted to integers

annagrin:
### Description
Fix constant folding of casts of integers.

Using the MLIR below
```mlir
%0 = arith.constant <c> : i8
%1 = cc.cast unsigned %0: i32
```
The constant folding creates the following code without changing `<c>`:
```mlir
%1 = arith.constant <c>: i32
```

This leads to incorrect values in constants when constant folding is invoked, for example,
`(unsigned)(true)` is currently converted to a `-1` integer constant.  

See added tests in `targettests/execution/cast.cpp` for more examples.

This PR updates constant folding to cast the source of the cast instruction to its original type before folding into a constant of a new type.

Example:

```c++
struct testConstBool {
  auto operator()() __qpu__ {
    cudaq::qubit q;
    unsigned i0 = (unsigned)(true);
    if (i0 == 1) {
      x(q);
    }
  }
};

int main() {
  auto counts = cudaq::sample(testConstBool{});
  counts.dump();
}
```

Quake before the fix:
```mlir
func.func @__nvqpp__mlirgen__testConstBool () attributes {"cudaq-entrypoint", "cudaq-kernel"} {
    %c-1_i32 = arith.constant -1 : i32
    %c1_i32 = arith.constant 1 : i32
    %0 = quake.alloca !quake.ref
    %1 = cc.alloca i32
    cc.store %c-1_i32, %1 : !cc.ptr<i32>
    %2 = cc.load %1 : !cc.ptr<i32>
    %3 = arith.cmpi eq, %2, %c1_i32 : i32
    cc.if(%3) {
      quake.x %0 : (!quake.ref) -> ()
    }
    return
  }
```

Quake after the fix:
```mlir
func.func @__nvqpp__mlirgen__testConstBool () attributes {"cudaq-entrypoint", "cudaq-kernel"} {
    %c1_i32 = arith.constant 1 : i32
    %0 = quake.alloca !quake.ref
    %1 = cc.alloca i32
    cc.store %c1_i32, %1 : !cc.ptr<i32>
    %2 = cc.load %1 : !cc.ptr<i32>
    %3 = arith.cmpi eq, %2, %c1_i32 : i32
    cc.if(%3) {
      quake.x %0 : (!quake.ref) -> ()
    }
    return
  }
```

Result before the fix:
`{ 0:0000 }`

Result after the fix:
`{ 1:0000 }`

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2804


targettests/execution/cast.cpp
```diff
@@ -0,0 +1,137 @@
...
+// RUN: if %braket_avail; then nvq++ %cpp_std --target braket --emulate %s -o %t && %t | FileCheck %s; fi
+// clang-format on
+
+#include <cudaq.h>
+#include <iostream>
+#include <string>
+#include <vector>
+
+struct testConstBool {
+  auto operator()() __qpu__ {
```
nvidia-dobri, line 27: Is this consistent with what C++ does? (unsigned)(true) feels to me like one of those use-at-your-own-risk things that are maybe not consistent across compilers.

nvidia-dobri (commented):
Change looks good, might be useful to know the context of when this is important / what the downstream breakage we saw was.

I'm a bit confused by the statement:
> (unsigned)(true) is currently converted to a -1 constant in MLIR

What does -1 in `unsigned` mean? `U..._MAX`?

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -214,6 +214,12 @@ OpFoldResult cudaq::cc::CastOp::fold(FoldAdaptor adaptor) {
       auto val = attr.getInt();
       if (isa<IntegerType>(ty)) {
         auto width = ty.getIntOrFloatBitWidth();
+        auto srcTy = getValue().getType();
```
schweitzpgi, line 217: The bug here is that this is not respecting the `signed` and `unsigned` flag on the cast. This is the correct code for truncation only.

What needs to be added here are the cases as below with `getSint()` and `getZint()` selectors. I suspect the `getSint()` case works with the original code as well. For `getZInt()`, we should simply mask off higher order bits. That is,
```c++
if (fromWidth < 64)
  val &= (1 << fromWidth) - 1;
```

targettests/execution/cast.cpp
```diff
@@ -0,0 +1,137 @@
...
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// clang-format off
+// RUN: nvq++ %cpp_std --enable-mlir %s -o %t && %t | FileCheck %s
+// RUN: nvq++ %cpp_std --target infleqtion      --emulate %s -o %t && %t | FileCheck %s
```
schweitzpgi, line 11: Since this change is generic and applies to any input at all (canonicalization is always run), I think we can conserve some CPU resources here and just execute the test once. We have other ways of tests all the various targets, and this is never going to change wrt the target.
We can get rid of run lines from line 11 through 18.

targettests/execution/cast.cpp
schweitzpgi: The syntax is irrelevant in the sense that a `bool true` is guaranteed to be the integer value `1` whether it is evaluated in a signed or unsigned context. It's like writing
```c++
((unsigned)1)
((unsigned)0)
```
Neither cast above changes any bits. (Although, the expression may be a subexpression of another expression.)

annagrin: @nvidia-dobri the tests includes side-by-side output of kernels and straight c++ code for comparison, they are identical after the fixes (they were not before)

lib/Optimizer/Dialect/CC/CCOps.cpp
annagrin: I used c++ to cast the const op value to the original type, to avoid possible bugs... let me know if you prefer using bit masking instead.

targettests/execution/cast.cpp
annagrin: done!

annagrin:
> Change looks good, might be useful to know the context of when this is important / what the downstream breakage we saw was.

I included an example in the description and the reference to the bug it fixes

> 
> I'm a bit confused by the statement:
> 
> > (unsigned)(true) is currently converted to a -1 constant in MLIR
> 
> What does -1 in `unsigned` mean? `U..._MAX`?

Constant folding folds the cast, so in this particular example we end up with an integer constant with value of -1, of type i32 (see example in the description). It would behave the same as uint32 max, yes.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -604,18 +604,11 @@ Value QuakeBridgeVisitor::integerCoercion(Location loc,
...
   assert(fromTy.isa<IntegerType>() && dstTy.isa<IntegerType>());
-  if (fromTy.getIntOrFloatBitWidth() < dstTy.getIntOrFloatBitWidth()) {
-    auto mode = (clangTy->isUnsignedIntegerOrEnumerationType())
-                    ? cudaq::cc::CastOpMode::Unsigned
-                    : cudaq::cc::CastOpMode::Signed;
-    return builder.create<cudaq::cc::CastOp>(loc, dstTy, srcVal, mode);
-  }
-  assert(fromTy.getIntOrFloatBitWidth() > dstTy.getIntOrFloatBitWidth());
-  return builder.create<cudaq::cc::CastOp>(loc, dstTy, srcVal);
+  auto mode = (clangTy->isUnsignedIntegerOrEnumerationType())
```
schweitzpgi, line 608: The mode should never be used when truncating a value.

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -214,8 +251,50 @@ OpFoldResult cudaq::cc::CastOp::fold(FoldAdaptor adaptor) {
...
+          case 16:
+            return createConst.template operator()<std::int16_t>(val);
+          case 32:
+            return createConst.template operator()<std::int32_t>(val);
+          case 64:
+            return createConst.template operator()<std::int64_t>(val);
+          default:
+            return nullptr;
+          }
+        };
```
schweitzpgi, line 296: Actually, the integer types can have any number of bits $n$. So this is a lot of extra code that only works for a small number of possible cases. Since we don't handle integers longer than 64 bits, we must handle $0 < n <= 64$ cases.

I presented the correct solution somewhere along the way. An integer that starts out as 64 bits cannot be extended by definition. An integer of less than 64 bits is apparently _already_ sign-extended when we read out the value. (That should be verified.) An integer of less than 64 bits is _not_ properly zero-extended. Zero extension of any $n$ bit 2's complement value is correctly computed with the following bitmasking operations.

```c++
value = value & ((1 << width) - 1);
```
This is a simple one-liner that is correct for any width from 0 to 63. (And 64 is a nop, so we're perfect with this.)

annagrin: I had to zero-extend to the source width first to get the original value, in addition to your suggestion. Appreciate another look (it did change some of my new test outcomes)

lib/Frontend/nvqpp/ConvertExpr.cpp
annagrin: done, thanks!

lib/Optimizer/Dialect/CC/CCOps.cpp
annagrin: see the last comment.

lib/Optimizer/Dialect/CC/CCOps.cpp
schweitzpgi: I don't understand this comment. If we have a 6 bit integer, the lower order 6 bits must be well-defined. That is when we read out the int64_t, the bits `b` must be correct. The bits `?` might be all 0, all 1, or random 0s and 1s.
```
?....??bbbbbb
```
The bit masking arithmetic above, `(1ull << width) - 1`, creates an unsigned 64 bit mask of
```
00...000111111
```
So when that is ANDed with the original value,, we will necessarily have
```
00,,,000bbbbbb
```
which is the zero-extended value of the i6 value, as expected. There is no need to further zero-extend either value.

annagrin: The constant already has a 1-extended value, for example, for a one-bit max integer (i.e. `True`), the value of the constant is `(std::int64_t)(-1)`. So to convert it to a `std::uint32_t`, it is not enough to perform `value = value & ((1ull << targetWidth) - 1);`, as it will result in `(std::uint32_t)(-1)`, as it will result in `4294967295` and not in `1`, and indeed my new tests failed.

When I added `value = value & ((1ull << srcWidth) - 1);`, all tests started to agree with c++ code (see the `targettests/execution/cast.cpp`.

annagrin: Actually, it might be enough to just zero-extend from the source size... I will let it run the tests.

annagrin: Looks like it is, thanks!

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -200,9 +200,46 @@ void cudaq::cc::AllocaOp::getCanonicalizationPatterns(
...
+    if (ty.getIntOrFloatBitWidth() == 1)
+      return val;
+
+    // Cast of a known unsigned value to an unsigned value.
+    if (val.getDefiningOp<cc::StdvecSizeOp>() ||
+        val.getDefiningOp<quake::VeqSizeOp>())
+      if (getZint())
+        return val;
+
+    // Fold consecutive casts when possible.
```
schweitzpgi, line 226: Eliminating cast chains is already done someplace ...

schweitzpgi: This one belongs in FuseCastCascade.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -488,7 +488,7 @@ bool QuakeBridgeVisitor::VisitCXXBoolLiteralExpr(clang::CXXBoolLiteralExpr *x) {
   auto intTy =
       builtinTypeToType(cast<clang::BuiltinType>(x->getType().getTypePtr()));
   auto intVal = x->getValue();
-  return pushValue(getConstantInt(builder, loc, intVal ? 1 : 0, intTy));
+  return pushValue(getConstantInt(builder, loc, intVal, intTy));
```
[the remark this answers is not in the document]
annagrin: I made this change when I tried to understand why we always create a `-1` for a boolean constant `true`, and it seemed redundant to do the conversion here at the time. Later I failed to find a test that fails, for example, I thought `cast<bool>(2)`  should fail as we would zero-extend the original value of `2` it to a `0`.  I suspect we never see a `cast<int>(boolLiteral(2))` in the AST. 

The compiler probably just casts it automatically to `true` before we see it... Will revert for safety.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -604,11 +604,8 @@ Value QuakeBridgeVisitor::integerCoercion(Location loc,
                                           const clang::QualType &clangTy,
                                           Type dstTy, Value srcVal) {
   auto fromTy = getResultType(srcVal.getType());
-  if (dstTy == fromTy)
```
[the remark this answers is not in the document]
annagrin: there can be still a cast needed to unsigned to zero-extend the original value, which is 1-extended to 64bits in the constant.

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -200,9 +200,46 @@ void cudaq::cc::AllocaOp::getCanonicalizationPatterns(
...
+    if (!isa<IntegerType>(ty))
+      return val;
+
+    // Cast that does not change the type or the sign.
+    if (!getZint() && !getSint())
+      return val;
+
+    // Cast of a boolean to a boolean.
+    if (ty.getIntOrFloatBitWidth() == 1)
+      return val;
```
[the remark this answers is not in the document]
annagrin: we fold casts of const ints below, this code just eliminates redundant casts, they could be done on non-constants as well. Is there an opt for it already, or should I add one somewhere else?

lib/Optimizer/Dialect/CC/CCOps.cpp
annagrin: will do, thanks!

annagrin: done!

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -378,13 +414,14 @@ LogicalResult cudaq::cc::CastOp::verify() {
...
-// pointer cast chains.
+// pointer cast chains and some integer cast chains.
 struct FuseCastCascade : public OpRewritePattern<cudaq::cc::CastOp> {
   using OpRewritePattern::OpRewritePattern;
 
   LogicalResult matchAndRewrite(cudaq::cc::CastOp castOp,
                                 PatternRewriter &rewriter) const override {
-    if (auto castToCast = castOp.getValue().getDefiningOp<cudaq::cc::CastOp>())
+    if (auto castToCast =
+            castOp.getValue().getDefiningOp<cudaq::cc::CastOp>()) {
```
schweitzpgi, line 424: Why did clang-format change this line?

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -395,6 +432,19 @@ struct FuseCastCascade : public OpRewritePattern<cudaq::cc::CastOp> {
                                                        castToCast.getValue());
         return success();
       }
+      if (isa<IntegerType>(castOp.getType()) &&
+          (castOp.getType() == castToCast.getType())) {
+        // %4 = cc.cast %3 : (i64) -> i64
+        // %5 = cc.cast unsigned %4 : (i64) -> i64
+        // ────────────────────────────────────────────
+        // %5 = cc.cast unsigned %3 : (i64) -> i64
```
schweitzpgi, line 440: This is an example of a NOP feeding a NOP. All uses of `%4` and `%5` can be replaced with `%3`.

test/AST-Quake/reverse.cpp
```diff
@@ -26,15 +26,16 @@ __qpu__ int std_reverse_std_vector_int() {
...
-// CHECK-DAG:       %[[VAL_5:.*]] = cc.alloca !cc.array<i32 x 10>
-// CHECK:           %[[VAL_6:.*]] = cc.cast %[[VAL_5]] :
+// CHECK:           %[[VAL_5:.*]] = cc.alloca !cc.array<i32 x 10>
+// CHECK:           %[[VAL_6:.*]] = cc.cast %[[VAL_5]] : (!cc.ptr<!cc.array<i32 x 10>>) -> !cc.ptr<i32>
 // CHECK:           %[[VAL_7:.*]] = cc.compute_ptr %[[VAL_5]][10] : (!cc.ptr<!cc.array<i32 x 10>>) -> !cc.ptr<i32>
 // CHECK:           %[[VAL_8:.*]] = cc.cast %[[VAL_7]] : (!cc.ptr<i32>) -> i64
 // CHECK:           %[[VAL_9:.*]] = cc.cast %[[VAL_6]] : (!cc.ptr<i32>) -> i64
 // CHECK:           %[[VAL_10:.*]] = arith.subi %[[VAL_8]], %[[VAL_9]] : i64
 // CHECK:           %[[VAL_11:.*]] = arith.divsi %[[VAL_10]], %[[VAL_4]] : i64
+// CHECK:           %[[VAL_12:.*]] = cc.cast unsigned %[[VAL_11]] : (i64) -> i64
```
schweitzpgi, line 36: What is the purpose of adding this operation? It has no semantics.

test/Quake/combine.qke
```diff
@@ -178,10 +178,14 @@ func.func @d(%c2: i64, %c3: i64, %c1: i64, %off: i16) {
 // CHECK-LABEL:   func.func @d(
 // CHECK-SAME:      %[[VAL_0:.*]]: i64, %[[VAL_1:.*]]: i64, %[[VAL_2:.*]]: i64, %[[VAL_3:.*]]: i16) {
 // CHECK:           %[[VAL_4:.*]] = quake.alloca !quake.veq<4>
-// CHECK:           %[[VAL_5:.*]] = arith.addi %[[VAL_0]], %[[VAL_2]] : i64
+// CHECK:           %[[VAL_0_casted:.*]] = cc.cast unsigned %[[VAL_0]] : (i64) -> i64
+// CHECK:           %[[VAL_2_casted:.*]] = cc.cast unsigned %[[VAL_2]] : (i64) -> i64
```
schweitzpgi, line 182: I don't understand why your adding these NOPs.

If we have a 64 bit value of 0xDeadBeefCafeBabe and "sign extend" or "zero extend" it to a 64 bit value we will always get the exact same value, 0xDeadBeefCafeBabe. These casts should be removed as dead code.

PR #2810, opened 2025-04-14
Remove vectorized instructions from LLVM IR for `cudaq::to_integer()`

bmhowe23:
This brings in MLIR's `makeOptimizingTransformer` function under our control since the one built-in to MLIR does not allow us to disable vectorization. The reason this change is important is because no QIR-based targets support LLVM's vectorized intrinsics. Maybe they will in the future, but I haven't heard of that being on anybody's near-term roadmap, so it would be best to disable that for now.

For review purposes, see 592d632f02aa004ea67c47549e163f10d3f95ca3 for the custom changes to the MLIR `makeOptimizingTransformer` function. (612d198bbda8d093c6f583a5f71b49c0c2df0ce5 is just bringing the original files into our source tree.)

### DIFF
--- a/.github/workflows/config/spelling_allowlist.txt
+++ b/.github/workflows/config/spelling_allowlist.txt
@@ -351,6 +351,7 @@ variadic
 variational
 vazirani
 vdots
+vectorization
 verifier
 vertices
 vqe
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -38,6 +38,8 @@ header:
     - 'docs/sphinx/_templates'
     - 'docs/sphinx/_static/cuda_quantum_icon.svg'
     - 'docker/test/installer/mpi_cuda_check.cpp'
+    - 'include/cudaq/Optimizer/CodeGen/OptUtils.h'
+    - 'lib/Optimizer/CodeGen/OptUtils.cpp'
     - 'runtime/cudaq/algorithms/optimizers/nlopt/nlopt-src'
 
   comment: on-failure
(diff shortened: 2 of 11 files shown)


schweitzpgi:
On the other hand, it's arguable that backends should be able to handle modern code patterns like simple vector instructions. These were conditional moves and bitmasking operations, IIRC.

More precisely, I believe this pattern, which arises from measuring a vector of qubits and building a single classical integral value, concerns *classical* value computation and has nothing to do with the quantum calls in QIR. The situation exposes an issue with the spec: what precisely is the intended division of concerns between the quantum interface and the surrounding classical code? Is it really a worthwhile goal to kneecap classical code optimizations (and thus performance) to isolate the quantum function calls subset?

bmhowe23:
These comments/concerns are definitely valid, but both the [QIR Base Profile](https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md) and [QIR Adaptive Profile](https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Adaptive_Profile.md) disallow those instructions right now. They both go into great detail about which LLVM instructions are allowed for various configurations, presumably because QPU controllers are not always regular CPUs and therefore will always be somewhat behind latest LLVM instruction sets in what they can actually support.

annagrin:
> This brings in MLIR's `makeOptimizingTransformer` function under our control since the one built-in to MLIR does not allow us to disable vectorization. 

I was looking into this as well, thanks for the fix! @1tnguyen added a test for `cudaq.run` that exposes this issue as well: https://github.com/NVIDIA/cuda-quantum/blob/50dd8e86fd3760a1c348e2968bc23765130c6a36/targettests/execution/cudaq_run.cpp#L20

THanks!

1tnguyen (approved):
LGTM 👍

PR #2811, opened 2025-04-14
Fix for high severity tasks

sacpis:
Fix for high severity tasks which appeared in the Static Application Security Testing Coverage (SAST) tool.
- Fix for memory leaks
- Removing std::move on const reference
- Resetting jit for freeing the memory
- Closing the socket to avoid leaks

# ctest result

```
100% tests passed, 0 tests failed out of 938

Label Time Summary:
gpu_required      = 1017.79 sec*proc (503 tests)
mgpus_required    =   6.03 sec*proc (3 tests)

Total Test time (real) = 2449.62 sec
```

### DIFF
--- a/docs/sphinx/examples/cpp/dynamics/cross_resonance.cpp
+++ b/docs/sphinx/examples/cpp/dynamics/cross_resonance.cpp
@@ -39,15 +39,13 @@ int main() {
   // are defined as: S^+ = 0.5 * (X + `iY`) and S^- = 0.5 * (X - `iY`) These
   // operators allow transitions between the spin states (|0> and |1>).
   auto spin_plus = [](int degree) {
-    return 0.5 *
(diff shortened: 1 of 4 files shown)


runtime/cudaq/distributed/mpi_plugin.cpp
```diff
@@ -47,6 +48,13 @@ MPIPlugin::MPIPlugin(const std::string &distributedInterfaceLib) {
   m_libFile = distributedInterfaceLib;
 }
 
+MPIPlugin::~MPIPlugin() {
```
schweitzpgi, line 51: Can this MPIPlugin ever be loaded by multiple threads?

1tnguyen (approved):
LGTM 👍 Thanks @sacpis

PR #2813, opened 2025-04-14
Add utility to skip tests based on host architecture

### LINKED ISSUE #2276: Need a way for tests to only run when the host is a particular architecture
Some parts of LLVM have to ability to add a line such as
```
REQUIRES: x86-registered-target
```
to the test to disable the test unless x86 is present. We need a way to be able to select between x86-64 and aarch64 (and anything new that's added) in order to write tests that prevent regressions for platform differences.

See also https://github.com/NVIDIA/cuda-quantum/actions/runs/11330747838/job/31509875881 where it looks like the assets builder may not be able to handle the `if` command? At any rate, we need a portable way of omitting tests when the host is not a specific architecture.

RevanthGundala:
### Description

Fixes #2276 

This PR introduces a utility module (python/tests/utils/arch.py) that allows tests to be conditionally skipped based on the host architecture (e.g., x86_64, aarch64). This addresses the need for a portable, maintainable way to prevent platform-specific test failures.

This change does not alter any test behavior by default, but provides a clear and reusable mechanism for future architecture-specific test requirements.

### DIFF
--- a/python/tests/utils/arch.py
+++ b/python/tests/utils/arch.py
@@ -1,0 +1,25 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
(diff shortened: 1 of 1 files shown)


schweitzpgi:
Were you planning to deploy this in some test(s) in particular?

Issue #2249, opened 2024-10-03
anyon target is not stable in CI


schweitzpgi:
It appears that the anyon target is not stable in the CI. Various tests will fail to verify when this target is selected and report that the standard output of the process is empty. We believe this is an undiscovered bug somewhere in this target's implementation.

As a workaround, we're removing this target from target tests on an ad hoc basis to keep the CI working in a deterministic way.

bmhowe23:
Just leaving this here for when we tackle the underlying root cause. #2236 removed some of the tests, so refer to that PR when adding the tests back in.

bmhowe23:
@schweitzpgi - is it possible this is related to https://github.com/NVIDIA/cuda-quantum/issues/1712?

schweitzpgi:
> [@schweitzpgi](https://github.com/schweitzpgi) - is it possible this is related to [#1712](https://github.com/NVIDIA/cuda-quantum/issues/1712)?

Might be. These failures were popping up for @khalatepradnya and @sacpis and myself somewhat randomly in the CI and about 50% of the time. The failure mode was that there was nothing appearing on the stdout stream. As it was hard to reproduce locally (smaller machine, fewer threads?), it was backed out rather than forcing everyone to run the CI twice per rebase on average.

Lambdauv:
just want to note that the CI tests were done multiple times without failures during the several days trying to merge the initial Anyon PR. is it possible that PRs after Sept 10 added something that increased the parallel workload in building the installer? It could be that simultaneously building multiple components that also build the same libraries repeatedly might cause problems for OS on smaller machines [inspired by discussions on sporadic build errors seen [here](https://groups.google.com/g/emscripten-discuss/c/llX0HtHyxN8?pli=1) ].

bmhowe23:
> just want to note that the CI tests were done multiple times without failures during the several days trying to merge the initial Anyon PR. is it possible that PRs after Sept 10 added something that increased the parallel workload in building the installer? It could be that simultaneously building multiple components that also build the same libraries in the processes repeatedly might cause problems for OS on smaller machines [inspired by discussions on sporadic build errors seen [here](https://groups.google.com/g/emscripten-discuss/c/llX0HtHyxN8?pli=1) ].

[...]

PR #2815, opened 2025-04-16
Fix intermittent bug with ObserveAnsatzPass

bmhowe23:
### Summary

I was able to reproduce the intermittent crash that was occurring in some of our CI runs on an internal machine. On the internal machine, the failure was occurring about 8-10 times out of 200 attempts (when running `cudaq_observe_term.cpp` w/ OQC target). With this PR applied, the failure occurs 0 out 200 times.

### Details

After reproducing the problem on the internal machine, I wanted to get a good stack trace of the segfault, so I rebuilt CUDA-Q with `Debug`. Unfortunately, the problem disappeared. So I rebuilt with `RelWithDebInfo` and saw that the segfault was occurring in the ObserveAnsatz pass, but it wasn't clear precisely where because LLVM was deep inside the (non-debug) LLVM code inside that pass.

I scanned the code and the recent changes, and I discovered that if I commented out a change in #2288 (`op->dropAllReferences();`), then the segfault disappeared. However, despite the new `dropAllReferences` code coinciding with the segfaults, it's not clear that calling `dropAllReferences` *should've* been causing any segfaults.

Since this code is in `matchAndRewrite`, I figured it would be better to use the `rewriter` to manipulate these ops rather than calling `op->dropAllReferences()` and `op->erase()` inside the function. My understanding is that the rewriter will handle any and all references and clean them up when done, and it will do so safely. @schweitzpgi to confirm.

The following issues are related, so I'm tagging them here, but I don't know if they fix these issues or not at this time:
- #2249
- #1712

### DIFF
--- a/lib/Optimizer/Transforms/ObserveAnsatz.cpp
+++ b/lib/Optimizer/Transforms/ObserveAnsatz.cpp
@@ -257,11 +257,9 @@ struct AppendMeasurements : public OpRewritePattern<func::FuncOp> {
 
       for (auto result : op->getResults())
         if (quake::isLinearType(result.getType()))
-          result.replaceAllUsesWith(op->getOperand(0));
+          rewriter.replaceAllUsesWith(result, op->getOperand(0));
 
-      // Force remove `op`.
-      op->dropAllReferences();
-      op->erase();
+      rewriter.eraseOp(op);
     }
 
     // We want to insert after the last quantum operation. We must perform this


mitchdz:
I don't immediately notice anything wrong here. Would like someone more familiar with this part of the codebase to check it though. +1 from me.

mitchdz (approved):

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -354,6 +317,51 @@ class ObserveAnsatzPass
   ObserveAnsatzPass(const std::vector<bool> &bsfData)
       : binarySymplecticForm{bsfData.begin(), bsfData.end()} {}
 
+  LogicalResult performPreprocessing(func::FuncOp funcOp,
+                                     const AnsatzFunctionInfo &infoMap) {
+    // Use an Analysis to count the number of qubits.
+    auto iter = infoMap.find(funcOp);
+    if (iter == infoMap.end())
+      return funcOp.emitOpError("Errors encountered in pass analysis");
+    auto nQubits = iter->second.nQubits;
```
annagrin, line 326: The number of qubits changed here from the original version - is it what was causing the issue?

annagrin: Got it!

annagrin (approved):

lib/Optimizer/Transforms/ObserveAnsatz.cpp
bmhowe23: Just for posterity's sake, in this new section of code, `nQubits` is just being used for error checking and nothing else.

In the original section of code where I moved this from, it was doing the following:
```cpp
auto nQubits = iter->second.nQubits;
if (nQubits < termBSF.size() / 2) { /* throw error */ }
nQubits = termBSF.size() / 2;
// Use nQubits
```

Now it is doing this...

In preprocessing:
```cpp
auto nQubits = iter->second.nQubits;
if (nQubits < termBSF.size() / 2) { /* throw error */ }
```
In matchAndRewrite:
```cpp
auto nQubits = termBSF.size() / 2;
// Use nQubits
```

schweitzpgi:
Yes, erasing Ops out of the graph when there are multiple threads operating in the graph is a terrible idea and leads to these sorts of "mystery" crashes. (We keep doing it though, so we need to do better in code reviews.) Most Ops, if left stranded by the rewrites will _automatically_ be erased, so the effort isn't only wrong (in the sense of causing a segfault), it is pointless. However, if an Op has side-effects, then it may _not_ be erased automatically. _Only_ ops with side-effects need special treatment. In those cases, it may be necessary to record the ops in a bag and then post-process the bag to see if they can be eliminated. If they are still referenced/used in some way, willy-nilly erasure will eventually cause more crashes since it may leave the IR in a broken state.

PR #2816, opened 2025-04-16
[NFC] Make tensornet backend implementation templated

1tnguyen:
### Description

Note: no function changes yet. The template type would not affect the code, i.e., it would still only use cutensornet double precision.


1tnguyen:
Superseded by https://github.com/NVIDIA/cuda-quantum/pull/2820. Closing

PR #2817, opened 2025-04-16
Migrating the Python operators to be C++ bindings

bettinaheim:
This PR introduces a type distinction between different kinds of operators and migrates the quantum operator infrastructure in Python to be bindings instead of Python code. This fixes previous performance issues after the initial introduction of the general quantum operator API in Python with 0.9.0. 

The following types are introduced:
- BosonOperator, BosonOperatorTerm, BosonOperatorElement
- FermionOperator, FermionOperatorTerm, FermionOperatorElement
- SpinOperator, SpinOperatorTerm, SpinOperatorElement
- MatrixOperator, MatrixOperatorTerm, MatrixOperatorElement

The OperatorSum class is now a union of BosonOperator | FermionOperator | SpinOperator | MatrixOperator.
The ProductOperator class is now a union of BosonOperatorTerm | FermionOperatorTerm | SpinOperatorTerm | MatrixOperatorTerm.
The ElementaryOperator class is now a union of BosonOperatorElement | FermionOperatorElement | SpinOperatorElement | MatrixOperatorElement.

Operators for bosons, fermions and spins leverage a dedicated internal representation that allows for significant performance improvements compared to general matrix operators. When operators of different kinds (types) are combined, the resulting operator will always be a matrix operator. 

Breaking changes:

- Each operator now keeps track of exactly what degrees it is acting on (no implicit injection of identities). 
  - num_qubits reflects only the degrees the operator explicitly acts upon; e.g. spin_op::i(2) act on a single degree
  - to_matrix reflects only the degrees the operator explicitly acts upon; e.g. spin_op::x(1) returns a 2x2 matrix (previously a 4x4 matrix)
  - useful functions to replicate the previous behavior where an operator would always act on all consecutive degrees from 0 to `max_degree` are: a constructor `SpinOperatorTerm(first_degree, last_degree)`, the member function `op.canonicalize(Iterable[int])`, and of course explicit multiplication with the identity(target).
- We introduce a type distinction between a `SpinOperator` (a sum of `SpinOperatorTerm`s) and a SpinOperatorTerm (a product of Pauli operators). This requires a couple of breaking changes:
  - The default constructor (`SpinOperator()`) will no longer create a product term with value 1. Instead, it will create a default instantiated sum, which will take the value of the first operator it is multiplied with or added to. For clarity and robustness, we recommend using an explicit `cudaq.spin.empty()` to create a sum with no terms, or `cudaq.spin.identity()` to create a product with value 1 as the previous constructor did.
  - If an operator is created as a product (e.g. op = spin.idenity()), it is not possible to `+=` other terms, since this requires `op` to be a `SpinOperator` rather than a `SpinOperatorTerm`. You can ensure `op` is of the appropriate type by either starting out with an empty sum (`op = spin.empty()`), or by explicitly converting it to the correct type (`op = SpinOperator(spin.identity())`). 
  - `spin.empty()` now creates a spin operator with no terms; to determine whether a variable `op` of type `SpinOperator` is empty (has no terms), use `op.num_terms == 0`. 
  - The constructor `SpinOperator(num_qubits: int)` no longer exists; the constructor `SpinOperatorTerm(first_degree: int, last_degree: int)` can be used to create a product term that applies the identity to all targets in the open range [first_degree, last_degree). The constructor `SpinOperator(size: int)` instantiates a sum without any terms, reserving space for the given size (estimated number of terms). 
- The serialization format for `SpinOperator`s changed;
  - serialization methods (`op.serialize()` and `op.to_json()`) are supported on both `SpinOperator` and `SpinOperatorTerm`, with matching constructors that take an `Iterable[float]` and methods `SpinOperator.from_json` and `SpinOperatorTerm.from_json`. 
  - the constructor `SpinOperator(filename: str)` expects the file to contain the new serialization format; please use `SpinOperator(filename: str, legacy: bool)` with `legacy = True` to load existing files using the format/representation 
- The module `cudaq.operator` has been restructured; 
   -  all operators and classes previously available under `cudaq.operator.operators` are now available under `cudaq.operators`
   - `cudaq.operators` contains the submodules `spin`, `boson`, and `fermion` (aliased also directly under `cudaq`)
   - the evolve related functionality (e.g. integrators) has been moved into a separate `cudaq.dynamics` folder; this should be largely non-breaking since the public classes and functions were and continue to be aliased directly under `cudaq`
- `ElementaryOperator.define` no longer exists; it is replaced by `cudaq.operators.define`, which defines a `MatrixOperatorElement`
- `ElementaryOperator(op_id: str, degrees: Iterable[int])` no longer exists; it is replaced by `cudaq.operators.instantiate(op_id, degrees)`, which produces a `MatrixOperatorTerm` (product operator)
- The class methods `ElementaryOperator.zero` and `ElementaryOperator.identity` no longer exist; they are replaced by `operators.zero` and `operators.identity`, which produce a `MatrixOperatorTerm` (product operator)
- all class methods for operator instantiation (such as e.g. `operators.identity` or `boson.identity`) will always produce a product operator as well; ElementaryOperators are no longer intended to be constructed directly but merely encountered when iterator over a ProductOperator 
- ProductOperator and OperatorSum constructors that take a vector of operators/terms are no longer available; products and sums of operators should be created by using +/-/* instead
- Setting the generator of a ScalarOperator after construction is no longer supported (necessary change due to C++ binding)
- ScalarOperators who are not constant are usually not equal even if they are defined with the same Python function (necessary consequence of the migration to C++ bindings)

Deprecations:
- `operators.create` and `operators.annihilate` are deprecated; please use `boson.create/annihilate` or `fermion.create/annihilate` instead.
- While the other operators (such as, e.g., `operators.number`) continue to be supported, we highly recommend replacing them with the appropriate `boson.number` or `fermion.number` instead to greatly improve performance.
- `get_coefficient` is deprecated; use `evaluate_coefficient` on a/each `SpinOperatorTerm` instead.
- `get_term_count` is deprecated; use the property `term_count` instead. Both are only available on `SpinOperator`, not on `SpinOperatorTerm`.
- `get_qubit_count` is deprecated; use the property `qubit_count` instead.
- `get_raw_data` is deprecated; raw data access will no longer be supported in future releases. Coefficients can be obtained via the `coefficient` property on a `SpinOperatorTerm`, and the binary symplectic form via the method `get_binary_symplectic_form()` on `SpinOperatorTerm`. 
- `is_identity` will no longer be supported on `SpinOperator` in future releases, but will continue to be supported on `SpinOperatorTerm`
- the constructor `SpinOperator(data: Iterable[double], num_qubits: int)` used to load the old deserialization format will be removed in future releases; it is replaced by the constructor `SpinOperator(data: Iterable[double])` using the new format
- `op.to_string(print_coefficient: bool)` is deprecated and will be removed in future releases; please use `str(op)` instead. The Pauli word representation for a `SpinOperatorTerm` can be obtained with `op.get_pauli_word(pad_identities: int = 0)` instead. 
- `for_each_term` and `for_each_pauli` are deprecated and replaced by the standard iteration over the sum, and over each term in the sum instead.

See also the description in https://github.com/NVIDIA/cuda-quantum/pull/2710 to understand the broader changes across the CUDA-Q codebase that are now reflected also in the Python API.

Potential follow ups that are not part of this PR:
- Expose commutation_behavior in Python as well
- Should we migrate RydbergHamiltonian to C++ bindings as well?
- Implement OperatorEvaluation in C++ instead?
- Add bindings and tests for to_sparse_matrix as soon as https://github.com/NVIDIA/cuda-quantum/pull/2764 is merged
- Add detection for conflicting commutation relation when a sum is evaluated
- Allow create a new operator by defining a mapping of degrees
- Support partial evaluation of operators (fixing some parameters but not others)


python/tests/builder/test_SpinOperator.py
bettinaheim, line 1: All tests in this file were migrated to python/tests/operators/test_spin_ops.py

python/tests/kernel/test_observe_kernel.py
```diff
@@ -265,7 +265,8 @@ def circuit(theta: float):
         ry(theta, q[1])
         x.ctrl(q[1], q[0])
 
-    h = spin.z(0)
+    h = spin.empty()
```
bettinaheim, line 268: calling distribute_terms on a product operator will give a deprecation warning. The function only really makes sense on a sum.

python/cudaq/operator/manipulation.py
```diff
@@ -27,7 +24,7 @@ class OperatorArithmetics(ABC, Generic[TEval]):
 
     @abstractmethod
     def evaluate(self: OperatorArithmetics[TEval],
-                 op: ElementaryOperator | ScalarOperator) -> TEval:
+                 op: CppOperatorElement | ScalarOperator) -> TEval:
```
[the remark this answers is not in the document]
bettinaheim: Yes, I needed to call it that to do the gradual migration, but it isn't intended to stay - I need to do a general pass and sort out module and class names, now that I got rid of the original python definitions.

python/runtime/cudaq/operators/py_boson_op.cpp
```diff
@@ -0,0 +1,498 @@
...
+  .def("__rmul__", [](const boson_op &other, int self) { return self * other; }, py::is_operator())
+  .def("__radd__", [](const boson_op &other, int self) { return self + other; }, py::is_operator())
+  .def("__rsub__", [](const boson_op &other, int self) { return self - other; }, py::is_operator())
+  .def("__rmul__", [](const boson_op &other, double self) { return self * other; }, py::is_operator())
+  .def("__radd__", [](const boson_op &other, double self) { return self + other; }, py::is_operator())
+  .def("__rsub__", [](const boson_op &other, double self) { return self - other; }, py::is_operator())
+  .def("__rmul__", [](const boson_op &other, std::complex<double> self) { return self * other; }, py::is_operator())
+  .def("__radd__", [](const boson_op &other, std::complex<double> self) { return self + other; }, py::is_operator())
+  .def("__rsub__", [](const boson_op &other, std::complex<double> self) { return self - other; }, py::is_operator())
+
```
1tnguyen, line 256: It seems like we missed left-hand arithmetic overloads for `boson_op` with `scalar_operator`.

python/cudaq/operator/manipulation.py
```diff
@@ -60,244 +57,140 @@ def tensor(self: OperatorArithmetics[TEval], val1: TEval,
...
+    if isinstance(operator, CppOperator):
+        evaluated = _sum_evaluation(operator, arithmetics, False)
+    elif isinstance(operator, CppOperatorTerm):
+        evaluated = _product_evaluation(operator, arithmetics, False)
+        # FIXME: CONVERT TO SUM?
+    else:
+        raise RuntimeError("the given value is not an operator")
+    if isinstance(evaluated, CppOperator) or isinstance(evaluated, CppOperatorTerm):
+        return evaluated
+    elif isinstance(evaluated, CppOperator):
```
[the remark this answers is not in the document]
bettinaheim: This should have been `elif isinstance(operator, CppOperator):`

python/tests/operator/test_scalar_op.py
```diff
@@ -0,0 +1,371 @@
...
+
+    dims = {0: 2}
+    assert so1.to_matrix(dims, t=2.0) == 2.0
+    assert np.allclose(op1.to_matrix(dims, t=2.0), [[0, 2], [2, 0]])
+    assert np.allclose(op2.to_matrix(dims, t=2.0), [[0, 2], [2, 0]])
+    assert np.allclose(op3.to_matrix(dims, t=2.0), [[2, 1], [1, 2]])
+    assert np.allclose(op4.to_matrix(dims, t=2.0), [[2, 1], [1, 2]])
+
+
+# BREAKING CHANGE: setting the generator after construction is no longer supported
```
bettinaheim, line 312: I'll remove the commented out tests here before merging - I just thought it would be useful to see what exactly we no longer support.

python/cudaq/operators/custom/custom_op.py
```diff
@@ -0,0 +1,106 @@
...
+                          id: str,
+                          expected_dimensions: Sequence[int],
+                          create: Callable[..., NDArray[numpy.complexfloating]],
+                          override: bool = False) -> None:
+    """
+    Creates the definition of an elementary operator with the given id.
+    """
+    if len(expected_dimensions) == 0:
+        raise ValueError(
+            f"custom operators needs to act on at least one degree " +
```
1tnguyen, line 28: nit: typo

```suggestion
            f"custom operators need to act on at least one degree " +
```

python/cudaq/operators/helpers.py
```diff
@@ -0,0 +1,115 @@
...
+import inspect, numpy, os, re, sys, typing  # type: ignore
+from typing import Any, Callable, Iterable, Mapping, Optional, Sequence, Tuple
+
+if (3, 11) <= sys.version_info:
+    NumericType = typing.SupportsComplex
+elif (3, 10) <= sys.version_info:
+    # type | type syntax is only supported in 3.10 or later
+    NumericType = numpy.complexfloating | complex | float | int
+else:
+    NumericType = typing.Union[numpy.complexfloating, complex, float, int]
```
1tnguyen, line 18: Perhaps we can now remove this `else` block since we don't support Python 3.9 anymore?

python/runtime/cudaq/operators/py_handlers.cpp
```diff
@@ -0,0 +1,259 @@
...
+  auto cmat_to_numpy = [](const complex_matrix &m) {
+    std::vector<ssize_t> shape = {static_cast<ssize_t>(m.rows()),
+                                  static_cast<ssize_t>(m.cols())};
+    std::vector<ssize_t> strides = {
+        static_cast<ssize_t>(sizeof(std::complex<double>) * m.cols()),
+        static_cast<ssize_t>(sizeof(std::complex<double>))};
+
+    // Return a numpy array without copying data
+    return py::array_t<std::complex<double>>(shape, strides, m.data);
+  };
```
1tnguyen, line 43: It might be nice to make this a standalone helper, as this seems to be a helpful function.

python/runtime/cudaq/operators/py_handlers.cpp
```diff
@@ -0,0 +1,259 @@
...
+                 std::move(
+                     degrees)); // FIXME: BIND AND SUPPORT COMMUTATION BEHAVIOR
+           }),
+           py::arg("id"), py::arg("degrees"),
+           "Creates the matrix operator with the given id acting on the given "
+           "degrees of "
+           "freedom. Throws a runtime exception if no operator with that id "
+           "has been defined.")
+      .def(py::init<const matrix_handler &>(), "Copy constructor.")
+      .def("__eq__", &matrix_handler::operator==)
```
1tnguyen, line 102: Should we also annotate this with `py::is_operator()`?

python/runtime/cudaq/operators/py_handlers.cpp
```diff
@@ -0,0 +1,259 @@
...
+      .def_property_readonly(
+          "target", &boson_handler::target,
+          "Returns the degree of freedom that the operator targets.")
+      .def_property_readonly("degrees", &boson_handler::degrees,
+                             "Returns a vector that lists all degrees of "
+                             "freedom that the operator targets.")
+      .def(py::init<std::size_t>(),
+           "Creates an identity operator on the given target.")
+      .def(py::init<const boson_handler &>(), "Copy constructor.")
+      .def("__eq__", &boson_handler::operator==)
```
1tnguyen, line 162: Should we also annotate this with `py::is_operator()`?

python/runtime/cudaq/operators/py_matrix.cpp
```diff
@@ -74,13 +75,12 @@ void bindComplexMatrix(py::module &mod) {
...
-          "__str__",
-          [](complex_matrix &self) {
-            std::stringstream ss;
-            self.dump(ss);
-            return ss.str();
-          },
-          "Write this matrix to a string representation.")
+          "dump", [](const complex_matrix &self) { self.dump(); },
+          "Prints the matrix to the standard output.")
+      .def("__eq__", [](const complex_matrix &lhs,
```
1tnguyen, line 80: Should we also annotate this with `py::is_operator()`?

python/runtime/cudaq/operators/py_matrix_op.h
```diff
@@ -0,0 +1,17 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
```
1tnguyen, line 8: Some new header files don't have a `#pragma once` guard. Although we don't expect them to be included in any other header files, it might be good to add the include guard for consistency.

python/runtime/cudaq/operators/py_scalar_op.cpp
```diff
@@ -0,0 +1,104 @@
...
+          "evaluate",
+          [&kwargs_to_param_map](const scalar_operator &self,
+                                 const py::kwargs &kwargs) {
+            return self.evaluate(kwargs_to_param_map(kwargs));
+          },
+          "Evaluated value of the operator.")
+
+      // comparisons
+
+      .def("__eq__", &scalar_operator::operator==)
```
1tnguyen, line 88: Should we also annotate this with `py::is_operator()`?

python/tests/dynamics/system_models.py
1tnguyen, line 1: We might need to update Python examples in https://github.com/NVIDIA/cuda-quantum/tree/main/docs/sphinx/examples/python/dynamics as well.

python/cudaq/operators/custom/custom_op.py
```diff
@@ -0,0 +1,106 @@
...
+            "of freedom - use a ScalarOperator to define operators that " +
+            "do not act on any degrees of freedom")
+
+    forwarded_as_kwarg = [["dimensions", "dims"], ["dimension", "dim"]]
```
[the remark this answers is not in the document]
bettinaheim: sure

python/cudaq/operators/custom/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+import typing
+from cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.operators import *
+from cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime import MatrixOperator, MatrixOperatorTerm
+from .custom_op import MatrixOperatorElement
```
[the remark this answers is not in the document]
bettinaheim: No, it should not. ScalarOperator and MatrixOperatorElement are both extended in Python itself.

python/cudaq/operators/expressions.py
```diff
@@ -0,0 +1,22 @@
...
+
+for op_type in typing.get_args(OperatorSum):
+    op_type._transform = _sum_transformation
+for op_type in typing.get_args(ProductOperator):
+    op_type._transform = _product_transformation
+
+for op_type in typing.get_args(OperatorSum):
+    op_type.evaluate = _evaluate
+for op_type in typing.get_args(ProductOperator):
+    op_type.evaluate = _evaluate
```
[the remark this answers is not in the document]
bettinaheim: Underscore vs not underscore is the python way of "marking" something as private. Evaluate is public and exists for all operators, but transform I'd prefer to keep private for now.

python/runtime/cudaq/operators/py_matrix_op.h
bettinaheim: py_spin_op.h also didn't have it. I tend to only add pragma once when including the header multiple times would cause issues (e.g. the header contains more than declarations only), which is not the case here. Is it best practice to always have pragma once? I thought it might be more hassle for the compiler and cause more issues than it solves if it is fine to have it included multiple times.

python/cudaq/operators/custom/custom_op.py
```diff
@@ -0,0 +1,106 @@
...
+                            f'for {len(dimensions)} degree(s) of freedom with dimension(s) {dimensions}')
+        # If the population of `kwargs` here is changed, adjust the filtering
+        # in the `parameters` property below.
+        for forwarded in forwarded_as_kwarg[0]:
+            kwargs[forwarded] = kwargs.get(
+                forwarded, dimensions)  # add if it does not exist
+        if len(dimensions) == 1:
+            for forwarded in forwarded_as_kwarg[1]:
+                kwargs[forwarded] = kwargs.get(
+                    forwarded, dimensions[0])  # add if it does not exist
```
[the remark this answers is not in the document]
bettinaheim: I don't see the repetition here. This merely adds some convenience that dim, dims, dimension and dimensions are automatically correctly passed to the custom generator even when the dimension argument during evaluation is passed as positional argument.

python/cudaq/operators/manipulation.py
```diff
@@ -0,0 +1,212 @@
...
+        for term in operator:
+            evaluated_term = _product_transformation(padded_term(term, degrees),
+                                                     arithmetics, pad_terms)
+            evaluated = arithmetics.add(evaluated, evaluated_term)
+    else:
+        for term in operator:
+            evaluated_term = _product_transformation(term, arithmetics,
+                                                     pad_terms)
+            evaluated = arithmetics.add(evaluated, evaluated_term)
+    return evaluated
```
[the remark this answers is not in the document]
bettinaheim: Copy is needed because we don't want to modify the operator during evaluation. 
The change you did with using sets for the degrees is going to be less performant - we want to preserve ordering to make the actual computation of the operator faster.

python/cudaq/operators/scalar/scalar_op.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Helper function to avoid duplicate code in the various arithmetic 
+    operations supported on a ScalarOperator.
+    """
+    if isinstance(other, NumericType):
+        if self.is_constant():
+            return ScalarOperator.const(fct(self.evaluate(), other))
+        generator = lambda **kwargs: fct(self.evaluate(**kwargs), other)
+        return ScalarOperator(generator, self.parameters)
+    elif type(other) == ScalarOperator:
```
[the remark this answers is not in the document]
bettinaheim: Probably fine either way. Forcing an exact match means the operator is delegated to the child class of ScalarOp, using isinstance means it is processed here. That said, the code is in general not really written with the intent that someone should inherit from operators and implement their own classes. that won't work for all the other operators in any case. In that sense, the exact match here matches the behavior for other operators, but I can go either way.

python/runtime/cudaq/operators/py_matrix_op.cpp
```diff
@@ -0,0 +1,864 @@
...
+  auto kwargs_to_param_map = [](const py::kwargs &kwargs) {
+    parameter_map params;
+    for (auto &[keyPy, valuePy] : kwargs) {
+      std::string key = py::str(keyPy);
+      std::complex<double> value = valuePy.cast<std::complex<double>>();
+      params.insert(params.end(),
+                    std::pair<std::string, std::complex<double>>(key, value));
+    }
+    return params;
+  };
```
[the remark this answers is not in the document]
bettinaheim: factored this out into their own py_helpers.cpp file.

python/tests/dynamics/system_models.py
bettinaheim: I checked the examples manually.

PR #2819, opened 2025-04-17
Add parser for the legacy tagged output format

1tnguyen:
### Description

Add extension points to the `RecordLogDecoder`:

- Key to detect if this is an output line (default is `OUTPUT`)

- Pull the handling code into a virtual method for derived classes to modify.

- Add a flag to indicate whether the `END` line has shot status or not.

Add the decoder for tagged output: use `RESULT` key to detect the output line, only use label to detect the type, and filter `mz` result registers.

Add a couple of test cases with synthetic data.

### DIFF
--- a/runtime/common/RecordLogDecoder.cpp
+++ b/runtime/common/RecordLogDecoder.cpp
@@ -7,6 +7,7 @@
  ******************************************************************************/
 
 #include "RecordLogDecoder.h"
+#include "Logger.h"
 
 void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
   std::vector<std::string> lines = cudaq::split(outputLog, '\n');
@@ -22,14 +23,14 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
       currentRecord = RecordType::HEADER;
     else if ("METADATA" == entries[0])
       currentRecord = RecordType::METADATA;
-    else if ("OUTPUT" == entries[0])
+    else if (OutputFieldKey == entries[0])
       currentRecord = RecordType::OUTPUT;
     else if ("START" == entries[0])
       currentRecord = RecordType::START;
     else if ("END" == entries[0])
       currentRecord = RecordType::END;
     else
-      throw std::runtime_error("Invalid data");
+      throw std::runtime_error(fmt::format("Invalid data: {}", entries));
 
     switch (currentRecord) {
     case RecordType::HEADER: {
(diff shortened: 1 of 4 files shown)


runtime/common/RecordLogDecoder.h
```diff
@@ -73,5 +76,28 @@ class RecordLogDecoder {
...
+  /// Key name for an output record.
+  std::string OutputFieldKey = "OUTPUT";
+  bool hasShotStatus = true;
+};
+
+// Decoder for the legacy/tagged (unspec'ed) output format.
+class TaggedRecordLogDecoder : public RecordLogDecoder {
+public:
+  TaggedRecordLogDecoder(const std::vector<std::string> &measureRegLabels = {})
+      : resultRecordLabels(measureRegLabels) {
```
khalatepradnya, line 88: Q: How will the caller know about these labels at time of instantiation?

1tnguyen: I guess this is something that we need to work out in the remote REST QPU runtime during the actual integration. It's available and is part of the payload we generated; thus, we can propagate that info back, e.g., via the execution context, or we can simply prepend an extra metadata line to communicate these register names back to the cudaq runtime.

khalatepradnya (approved):
LGTM. Thanks, Thien!

runtime/common/RecordLogDecoder.cpp
```diff
@@ -22,14 +23,14 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
       currentRecord = RecordType::HEADER;
     else if ("METADATA" == entries[0])
       currentRecord = RecordType::METADATA;
-    else if ("OUTPUT" == entries[0])
+    else if (OutputFieldKey == entries[0])
```
schweitzpgi, line 26: Thanks for removing the Yoda test syntax. :)

runtime/common/RecordLogDecoder.cpp
```diff
@@ -51,41 +52,48 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
...
+  if ((schema == SchemaType::LABELED) && (entries.size() != 4))
+    throw std::runtime_error("Unexpected record size for a labeled record");
 
-      if ("RESULT" == recType)
-        throw std::runtime_error("This type is not yet supported");
-      if ("TUPLE" == recType)
-        throw std::runtime_error("This type is not yet supported");
-      if ("ARRAY" == recType)
-        throw std::runtime_error("This type is not yet supported");
+  std::string recType = entries[1];
```
schweitzpgi, line 76:
```suggestion
  std::string &recType = entries[1];
```
Can these be references? To avoid a copy?

runtime/common/RecordLogDecoder.cpp
```diff
@@ -51,41 +52,48 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
...
 
-      if ("BOOL" == recType)
-        currentOutput = OutputType::BOOL;
-      else if ("INT" == recType)
-        currentOutput = OutputType::INT;
-      else if ("DOUBLE" == recType)
-        currentOutput = OutputType::DOUBLE;
-      else
-        throw std::runtime_error("Invalid data");
+  if ("RESULT" == recType)
```
schweitzpgi, line 80: Maybe we can undo the Yoda tests here? (It doesn't follow our coding practice.)

khalatepradnya: See: https://github.com/NVIDIA/cuda-quantum/pull/2826

PR #2823, opened 2025-04-18
Change ObserveAnsatz pass to directly manipulate IR

bmhowe23:
Use of the pattern rewriter is not needed in this case because we are explicitly modifying the IR in this pass; we don't need any pattern folding capabilities here.

GitHub is showing the diffs strangely for this PR, so I will highlight what is actually changing by adding comments to the PR.

### DIFF
--- a/lib/Optimizer/Transforms/ObserveAnsatz.cpp
+++ b/lib/Optimizer/Transforms/ObserveAnsatz.cpp
@@ -6,12 +6,8 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-#include "PassDetails.h"
 #include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
 #include "cudaq/Optimizer/Transforms/Passes.h"
-#include "cudaq/Todo.h"
-#include "mlir/Transforms/DialectConversion.h"
-#include "mlir/Transforms/Passes.h"
 
 namespace cudaq::opt {
 #define GEN_PASS_DEF_OBSERVEANSATZ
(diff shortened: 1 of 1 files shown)


lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -202,30 +198,76 @@ struct AnsatzFunctionAnalysis {
   AnsatzFunctionInfo infoMap;
 };
 
-/// This OpRewritePattern will use the quake ansatz analysis info to append
-/// measurement basis change operations.
-struct AppendMeasurements : public OpRewritePattern<func::FuncOp> {
-  explicit AppendMeasurements(MLIRContext *ctx, const AnsatzFunctionInfo &info,
-                              ArrayRef<bool> bsf)
-      : OpRewritePattern(ctx), infoMap(info), termBSF(bsf) {}
+/// This pass will compute ansatz analysis meta data and use that in a custom
```
bmhowe23, line 201: Despite how GitHub is rendering this, lines 201-259 did not actually change here; they are just moved from lower in the file.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -202,30 +198,76 @@ struct AnsatzFunctionAnalysis {
...
+
+      op->dropAllReferences();
+      op->erase();
+    }
+    return success();
+  }
+
+  /// Append measurements in the correct basis.
+  LogicalResult appendMeasurements(func::FuncOp funcOp,
+                                   const AnsatzFunctionInfo &infoMap) {
```
bmhowe23, line 263: This is new `appendMeasurements` function is replacing the `matchAndRewrite` function.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -202,30 +198,76 @@ struct AnsatzFunctionAnalysis {
...
     // Use an Analysis to count the number of qubits.
     auto iter = infoMap.find(funcOp);
     if (iter == infoMap.end())
       return funcOp.emitOpError("Errors encountered in pass analysis");
 
     // Set nQubits so we only measure the requested qubits.
-    auto nQubits = termBSF.size() / 2;
+    auto nQubits = binarySymplecticForm.size() / 2;
```
bmhowe23, line 270: The old struct where this function resided had a `termBSF` member whereas that same data is called `binarySymplecticForm` in the struct where this new function resides.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -386,13 +364,8 @@ class ObserveAnsatzPass
...
 
-    patterns.insert<AppendMeasurements>(ctx, funcAnalysisInfo,
-                                        binarySymplecticForm);
-    ConversionTarget target(*ctx);
-    target.addLegalDialect<quake::QuakeDialect>();
-
-    if (failed(applyPartialConversion(getOperation(), target,
-                                      std::move(patterns))))
+    // Now append the measurements.
+    if (failed(appendMeasurements(funcOp, funcAnalysisInfo)))
```
bmhowe23, line 368: This is arguably the main change for this PR.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -202,30 +198,76 @@ struct AnsatzFunctionAnalysis {
...
-  explicit AppendMeasurements(MLIRContext *ctx, const AnsatzFunctionInfo &info,
-                              ArrayRef<bool> bsf)
-      : OpRewritePattern(ctx), infoMap(info), termBSF(bsf) {}
+/// This pass will compute ansatz analysis meta data and use that in a custom
+/// rewrite pattern to append basis changes + mz operations to the ansatz quake
+/// function.
+class ObserveAnsatzPass
+    : public cudaq::opt::impl::ObserveAnsatzBase<ObserveAnsatzPass> {
+protected:
+  SmallVector<bool> binarySymplecticForm;
```
schweitzpgi, line 207: Prefer to have this at the bottom of the class.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -373,9 +354,6 @@ class ObserveAnsatzPass
       for (auto &b : termBSF)
         binarySymplecticForm.push_back(b);
 
-    auto *ctx = funcOp.getContext();
```
schweitzpgi, line 376: This is a FuncOp pass, right? The dyn_cast on line 347 is not needed if this is a declared func::FuncOp pass.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -386,13 +364,8 @@ class ObserveAnsatzPass
       return;
     }
 
-    patterns.insert<AppendMeasurements>(ctx, funcAnalysisInfo,
```
schweitzpgi, line 389: Is it ever the case that the analysis would validly conclude not to change the FuncOp? If so, line 363 probably shouldn't signal a total failure.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
```diff
@@ -202,30 +198,76 @@ struct AnsatzFunctionAnalysis {
...
+    // kernels, we remove them here since we are adding specific measurements
+    // below. Note: each `op` in the list of measurements must be removed by the
+    // end of this loop, otherwise the end result may be incorrect.
+    for (auto *op : iter->second.measurements) {
+      bool safeToRemove = [&]() {
+        for (auto user : op->getUsers())
+          if (!isa<quake::SinkOp, quake::ReturnWireOp>(user))
+            return false;
+        return true;
+      }();
```
schweitzpgi, line 246: How does this test determine that the `discriminate` either doesn't exist or its result isn't really used?

bmhowe23: Lines 243-244 ensure that no discriminate's exist after the measurement. Only sink's and ReturnWire's are allowed after the measurement.

lib/Optimizer/Transforms/ObserveAnsatz.cpp
bmhowe23: Updated in 794f7919

lib/Optimizer/Transforms/ObserveAnsatz.cpp
bmhowe23: Updated in 794f7919

schweitzpgi (approved):
Thanks for finding this ninja bug.

Issue #2822, opened 2025-04-17
Controlled exp_pauli is not working


psm-optimizes:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

1. I want to use the controlled version of exp_pauli to apply the Hamiltonian simulation of the coefficient matrix of a linear system of equations as a part of a HHL-type algorithm. 
2. Throughout the circuit, I call the controlled exp_pauli in a for loop to apply the Hamiltonian controlled by different cubits. 
3. The error I get is as follows:
error: 'quake.extract_ref' op invalid constant index value
RuntimeError                              Traceback (most recent call last)
Cell In[17], line 3
      1 # Draw the quantum circuit
      2 print("============================== Circuit structure ============================== \n")
----> 3 print(cudaq.draw(GQLSS, b, eigs, r_choice, 2*np.pi, coefficients, words))

RuntimeError: cudaq::builder failed to JIT compile the Quake representation.

### Steps to reproduce the bug
The code is available [here](https://github.com/psm-optimizes/CudaQ-QLSA/blob/main/GQLSS.ipynb).

### Expected behavior

The circuit should be drawn. 

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 0.10.0 
- **Python version**: 3.11.8
- **C++ compiler**: 
- **Operating system**: Linux blue2 5.10.0-20-amd64

bmhowe23:
Thanks for reporting this issue, @psm-optimizes. Based on a quick look, it looks like there is bug in how CUDA-Q is handling this section of code:
```python
    # Apply y rotations on Ancilla qubit
    for i in range(len(qReg)+1):
        if i != 0:
            ry.ctrl((2*np.pi)/eigs[i-1], [qReg[i-1]], qAnc)  
```
[...]

PR #2824, opened 2025-04-18
[core] Add direct simplification of cc.if when condition is a constant.

schweitzpgi:


schweitzpgi:
Addresses one of the issues in #2822 (but not all of them).

bmhowe23 (approved):
LGTM. Thanks, Eric.

PR #2826, opened 2025-04-20
[feature/run] Support parsing of arrays

khalatepradnya:
TODO: Follow-up PR for nested vectors and tuples.


runtime/common/RecordLogDecoder.h
```diff
@@ -46,17 +45,47 @@ class RecordLogDecoder {
...
-        return OutputType::BOOL;
-      return OutputType::INT;
-    } else if ('f' == label[0]) {
-      return OutputType::DOUBLE;
-    }
-    throw std::runtime_error("Unknown datatype in label");
-  }
+  /// A helper structure that provides handlers for various operations depending
+  /// on the data types.
+  struct TypeHandler {
```
schweitzpgi, line 50: I wonder if you couldn't just skip the lambda class wrappers entirely and use plain-old class overloading instread. Something like
```c++
template <typename CRTP>
struct TypeHandlerBase {
  void addRecord(const std::string& s) { static_cast<const CRTP*>(this)->addRecord(s); }
  ...
};

struct BoolHandler : public TyoeHandlerBase<BoolHandler> {
  void addRecord(const std::string& s) { .... }
  ...
};
```

khalatepradnya: Thanks for the comment. Is one method better / preferable than the other?
I tried to modify this with a virtual base class, 'TypeHandler' and child classes for each type, but ran into errors while accessing the members and functions of the 'RecordLogDecoder' class.

runtime/common/RecordLogDecoder.cpp
```diff
@@ -13,118 +13,355 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
...
+}
+
+void cudaq::RecordLogDecoder::handleOutput(
+    const std::vector<std::string> &entries) {
+  if (entries.size() < 3)
+    throw std::runtime_error("Insufficient data in a record");
+  if ((schema == SchemaType::LABELED) && (entries.size() != 4))
+    throw std::runtime_error("Unexpected record size for a labeled record");
+  std::string recType = entries[1];
+  std::string recValue = entries[2];
```
1tnguyen, line 258: Would this change compile? If so, we can save 2 string copies :) 

```suggestion
  const std::string& recType = entries[1];
  const std::string& recValue = entries[2];
```

runtime/common/RecordLogDecoder.cpp
```diff
@@ -13,118 +13,355 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
...
+    auto it = dataTypeMap.find(ty);
+    if (it != dataTypeMap.end())
+      it->second.allocateTuple();
+    else
+      throw std::runtime_error("Unsupported array type");
+  }
+}
+
+bool cudaq::RecordLogDecoder::convertToBool(const std::string &value) {
+  if (value == "true")
```
annagrin, line 323: I encountered "0" and "1" as values here

khalatepradnya: Addressed in 08b4e76644b87ebc371ad7efafa106f38b103771

runtime/common/RecordLogDecoder.cpp
```diff
@@ -13,118 +13,355 @@ void cudaq::RecordLogDecoder::decode(const std::string &outputLog) {
...
+  if (m_size != tupleTypes.size())
+    throw std::runtime_error("Tuple size mismatch in value and label.");
+  for (auto &ty : tupleTypes)
+    ty.erase(std::remove(ty.begin(), ty.end(), ' '), ty.end());
+}
+
+std::size_t cudaq::RecordLogDecoder::ContainerHandler::extractIndex(
+    const std::string &label) {
+  if ((label[0] == '[') && (label[label.size() - 1] == ']'))
+    return std::stoi(label.substr(1, label.size() - 1));
```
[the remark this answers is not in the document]
khalatepradnya: Turns out it was working because as per the docs, `If the string is too short, use the remainder of the characters.`

runtime/common/RecordLogDecoder.cpp
```diff
@@ -8,123 +8,391 @@
...
+          [](RecordLogDecoder *self) {
+            self->buffer.resize(self->handler.offset + sizeof(char));
+            self->handler.tupleOffsets.push_back(self->handler.offset);
+            self->handler.offset += sizeof(char);
+          },
+          [](RecordLogDecoder *self, std::size_t index,
+             const std::string &value) {
+            self->insertIntoTuple<char>(
+                index, static_cast<char>(self->convertToBool(value)));
+          }}},
```
schweitzpgi, line 60: Please replace these arrays of std::functions with a class overload with named member functions. That will be better self-documenting code than trying to remember whether the function to call is the 3rd or 5th one in an array.

runtime/common/RecordLogDecoder.h
schweitzpgi: > Thanks for the comment. Is one method better / preferable than the other? I tried to modify this with a virtual base class, 'TypeHandler' and child classes for each type, but ran into errors while accessing the members and functions of the 'RecordLogDecoder' class.

note: A virtual base class is only (semi-) useful when using multiple inheritance with more than one level of inheritance as it can be used to merge/fuse repetitive data members. These classes don't fit those parameters.

To access members (of data or function sorts) in another class in C++, we simply need to have a C++ specification of the class to be accessed (a decl or defn that specifies what the data and function members are) and we need the proper access rights (a.k.a. `public` generally speaking). If you'd rather not make everything public but still want a member function in some other class to have access rights to `private` data members, then C++ provides the `friend` declaration for exactly that.

khalatepradnya: Yes, I tried using `friend` class but ran into circular dependency. I have now moved buffer handling and type conversion into separate classes, which hopefully addresses the initial concern. Appreciate the discussion and feedback. 🙏🏽

runtime/common/RecordLogDecoder.cpp
khalatepradnya: Refactored in 69034283d4da4d70fe64993f2f0a5c33cf95246b

runtime/common/RecordLogDataUtils.h
```diff
@@ -0,0 +1,280 @@
...
+    bh.insertIntoTuple<T>(offset, converter->convert(value));
+  }
+};
+
+/// Maps QIR type strings (e.g., "i32", "f64") to the appropriate data handling
+/// instance.
+class DataHandlerFactory {
+public:
+  static std::unique_ptr<DataHandlerBase>
+  createDataHandler(const std::string &type) {
```
annagrin, line 252: would it be simpler to just have a bunch of converters for each of the types below as static fields instead of creating them, storing in the cache, and then looking them up? Code that uses it can just switch on the type and pick the right static converter. Or, alternatively, switch on the type and create a converter on the fly when you use, as they are not large objects. This would save us from using costly hash lookup operations IMHO.

runtime/common/RecordLogDecoder.cpp
```diff
@@ -8,123 +8,181 @@
 
 #include "RecordLogDecoder.h"
 
+const std::unordered_map<std::string,
+                         std::function<void(cudaq::RecordLogDecoder *,
+                                            const std::vector<std::string> &)>>
+    cudaq::RecordLogDecoder::recordHandlers = {
```
annagrin, line 14: we could have a `handleRecord` function that has a switch inside to call the correct `handleXXX` function based on the type. This way the there is no costly hash table lookup involved...

khalatepradnya: Addressed in 6a7ae11d924e0fbf8d8b01b1a6a160b185a62b0c

runtime/common/RecordLogDataUtils.h
khalatepradnya: Addressed in 0244d5be38e1e5e76589b56b6a7b273c9c090dbe

PR #2828, opened 2025-04-21
Follow-up to #2809 - only set linker flags if environment supports it

bmhowe23:
Some environments (like our manylinux build environment that is based off of AlmaLinux) don't support things like `-fuse-ld=ld`. That is - even if `ld` is the system linker, the compiler doesn't support being told `-fuse-ld=ld`. This change omits setting CMake variables for cases like that, while still allowing the lld linker to be used in environments that support it.

This is needed for the CUDA-QX builds, which invoke the `build_cudaq.sh` script for the manylinux environment.

Example failure (before this PR): https://github.com/NVIDIA/cudaqx/actions/runs/14544697190/job/40808489265


schweitzpgi (approved):
+1

PR #2829, opened 2025-04-21
[core] Prep work to refactor the GKE code.

schweitzpgi:
This work is preparation for other future work.

### Description

This moves a bunch of code out of the GKE pass proper into the builder library where it can be used by other passes. There are various minor changes to rename things in this refactor, but the essential functionality is not changed.


khalatepradnya (approved):
This refactoring looks good to me.

PR #2840, opened 2025-04-24
Drop the result_record_output call when using cudaq::run.

schweitzpgi:


1tnguyen (approved):
LGTM 👍 Thanks, @schweitzpgi!

PR #2844, opened 2025-04-29
Restructure `AsyncSampleResult` binding

1tnguyen:
### Description

There is an oversight in PR #2800, where we created a wrapper for the underlying `async_sample_result` type.
This type is directly exposed in other binding code paths, hence a binding is required for the underlying type.

Hence, refactoring the implementation to make the wrapper a proper trampoline class of the `async_sample_result`.

Also, add a virtual destructor to enable trampoline inheritance.


khalatepradnya (approved):
LGTM. Thanks, Thien!

PR #2842, opened 2025-04-26
[Python] Return aggregate types in cudaq.run

annagrin:
### Description
Support returning tuples and lists in `cudaq.run` for Python.

<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2826</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2839</strike>

PR #2851, opened 2025-04-30
Decode and read bool vector

annagrin:
### Description
Respect `vector<bool>` data layout during log parsing

<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2842</strike>


dictionary.dic
khalatepradnya, line 1: This might be leftover from running spell checker.

targettests/execution/cudaq_run.cpp
```diff
@@ -148,6 +148,22 @@ int main() {
   //   }
   // }
 
+  {
```
khalatepradnya, line 151: Q: Isn't this same as the commented code above on lines 137-149?
In other words, the TODO is fixed by this PR? So, we can remove it, right?

runtime/common/RecordLogParser.h
```diff
@@ -105,6 +105,16 @@ class BufferHandler {
   template <typename T>
   size_t allocateArrayRecord(size_t arrSize) {
     size_t vectorOffset = buffer.size();
+    if constexpr (std::is_same_v<T, bool>) {
+      auto *allocation = new std::vector<bool>(arrSize);
+      if (!allocation)
+        throw std::runtime_error("Memory allocation failed");
```
schweitzpgi, line 111: Just use malloc here to avoid the C++ overhead.
```c++
  result = malloc(sizeof(std::vector<bool>) * arrSize);
  if (!result) ...
```
Specifically, `new` will throw exceptions.

runtime/common/RecordLogParser.h
```diff
@@ -125,8 +135,13 @@ class BufferHandler {
 
   template <typename T>
   void insertIntoArray(size_t offset, std::size_t index, T value) {
```
schweitzpgi, line 137: This would be much more efficient if we used memcpy to copy the data (instead of calling this function once per element) or, in the case of `std::vector<bool>`, used the `std::initializer_list` variant of the `std::vector<bool>` ctor that I detailed in slack.

targettests/execution/cudaq_run.cpp
annagrin: Merge conflict;) thank for noticing, will fix!

runtime/common/RecordLogParser.h
annagrin: Wouldn’t it not be exact memory allocate though? Size(vecvor<bool>) is a constant (40 on my machine) and I am not sure how much it will actually use for the elements and where… so won’t we introduce potential memory leaks or bad memory access this way?

schweitzpgi: In short, no.

annagrin: resolved offline

runtime/common/RecordLogParser.h
annagrin: resolved offline

targettests/execution/cudaq_run.cpp
annagrin: done, thanks!

khalatepradnya (approved):
LGTM!

PR #2852, opened 2025-04-30
[Python] Support returning data class objects

annagrin:
### Description
Support returning data class objects in Python

## Details

- Move dataclass registry to c++
- Use stored types to create python dataclass objects when returning from `cudaq.run`
- Add python bindings for dataclass registry
- Update python code to use dataclass bindings instead of a global python map

<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2842</strike>


schweitzpgi (commented):
Looks like this fixes some bug(s) that the dataclasses aren't always registered correct.ly?

annagrin:
> Looks like this fixes some bug(s) that the dataclasses aren't always registered correct.ly?

No, I don't believe it does, at least not intentionally:) I just moved the dataclass registry to c++ so I can use it to create dataclass objects on return

import.log
khalatepradnya, line 1: Does this file need to be checked in?

annagrin: no, will remove, my apologies

PR #2856, opened 2025-05-02
Maintenance: refactor canonicalization patterns for clarity.

schweitzpgi:
Adds a bit of documentation and refactors to physically partition the canonicalization patterns of two completely unrelated Ops that happen to use the same hook for convenience. (One of the Ops is defined in MLIR, so this adds a canonicalization pattern by letting it ride the coattails of another Op and evade security.)

### DIFF
--- a/lib/Optimizer/Dialect/CC/CCOps.cpp
+++ b/lib/Optimizer/Dialect/CC/CCOps.cpp
@@ -430,8 +430,12 @@ struct SimplifyIntegerCompare : public OpRewritePattern<arith::CmpIOp> {
     return failure();
   }
 };
+} // namespace
 
-// Ad hoc pattern to erase complex.create. (MLIR doesn't do this.)
+namespace {
+// Ad hoc pattern to erase complex.create. (MLIR doesn't do this.) This pattern
+// gets piggybacked into the canonicalizations, but does NOT have anything to do
+// with cc::CastOp.
 struct FuseComplexCreate : public OpRewritePattern<complex::CreateOp> {
   using OpRewritePattern::OpRewritePattern;
   LogicalResult matchAndRewrite(complex::CreateOp create,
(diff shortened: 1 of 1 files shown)


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -453,10 +457,16 @@ struct FuseComplexCreate : public OpRewritePattern<complex::CreateOp> {
...
+                                           MLIRContext *context) {
+  patterns.add<FuseComplexCreate>(context);
+}
+
 void cudaq::cc::CastOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
                                                     MLIRContext *context) {
-  patterns.add<FuseCastCascade, SimplifyIntegerCompare, FuseComplexCreate>(
-      context);
+  patterns.add<FuseCastCascade, SimplifyIntegerCompare>(context);
+  getArbitraryCustomCanonicalizationPatterns(patterns, context);
```
bmhowe23, line 469: No objections, but I'm not sure I understand the functional difference between:
```cpp
// The new better way
patterns.add<pattern1>(context);
patterns.add<pattern2>(context);
```
and
```cpp
// The old bad way
patterns.add<pattern1, pattern2>(context);
```

bmhowe23 (approved):

lib/Optimizer/Dialect/CC/CCOps.cpp
schweitzpgi: Two of the patterns deal with cc::CastOp, which is what the hook is actually for.

A thrid pattern was added to canonicalize complex::CreateOp, which has nothing to do with anything here ... it's just using the hook as a sneaky way to add to the canonicalizer instead of modifying the MLIR sources. When we upgrade MLIR, someone will probably have to remove this 3rd pattern as I assume the MLIR folks will have fixed it by then.

PR #2860, opened 2025-05-04
[CI] [testing] Updates to `pycudaq-mlir` tests

khalatepradnya:
* The `pycudaq-mlir` tests are being run twice in the dev pipeline - once with `ctest` and second time explicitly with `llvm-lit`.
* Removing the explicit invocation from `Build and Test` stage of CI (since it is slightly slower in a handful of instances I checked)
* Added `FileCheck` to the command wherever `CHECK` patterns are present in the test file.
* Removed duplicate assertions if `CHECK` already present.
* Code formatting.

Ref: https://github.com/NVIDIA/cuda-quantum/actions/runs/14803509856/job/41567993343?pr=2857
```
Internal ctest changing into directory: /workspaces/cuda-quantum/build
Test project /workspaces/cuda-quantum/build
        Start   1: pycudaq-mlir
  1/443 Test   #1: pycudaq-mlir ...................................................   Passed  215.14 sec
        Start   2: qpp_AdjointTester.checkSimple
...

-- Testing: 64 tests, 8 workers --
PASS: PYCUDAQMLIR :: target/cudaq_observe.py (1 of 64)
...
PASS: PYCUDAQMLIR :: target/conditional_sample.py (64 of 64)

Testing Time: 229.31s

Total Discovered Tests: 64
  Passed           : 63 (98.44%)
  Expectedly Failed:  1 (1.56%)
```


schweitzpgi:
I was a little surprised when these tests appeared as part of the `ctest` collected tests. I believe it is still possible to build the project with python disabled. right? I wonder if ctest is smart enough to disable these tests in that case.

khalatepradnya:
> I was a little surprised when these tests appeared as part of the `ctest` collected tests. I believe it is still possible to build the project with python disabled. right? I wonder if ctest is smart enough to disable these tests in that case.

I don't know what caused these tests to show up in `ctest`. I distinctly recall adding the extra step in the YAML file because they were not showing up... that was a year ago. I can check what happens when we disable Python.

1tnguyen:
I wonder whether we should rather add it to the `ctest` exclude argument, which is currently `-E "ctest-nvqpp|ctest-targettests"`, just for uniformity with the other two lit test suites that are also hooked to `ctest` but excluded. 

I'm not sure about the intention of explicit invocation (a bit slower as we've seen), perhaps to see the log as the test cases progress :)

khalatepradnya:
> I wonder whether we should rather add it to the `ctest` exclude argument, which is currently `-E "ctest-nvqpp|ctest-targettests"`, just for uniformity with the other two lit test suites that are also hooked to `ctest` but excluded.
> 
> I'm not sure about the intention of explicit invocation (a bit slower as we've seen), perhaps to see the log as the test cases progress :)

The `pycudaq-mlir` test explicit invocation was added because they didn't use to run with `ctest` (a year ago). I'm not sure what changed / when they started running. I kept the one which is slightly faster.

khalatepradnya:
Closing this PR as the random failures are no longer prevalent.

PR #2864, opened 2025-05-05
Fix #2863.

### LINKED ISSUE #2863: Wrong handling of `u3`'s adjoint
### Required prerequisites

### Describe the bug

@amccaskey found the issue.

The adjoint of `U3` is not only the negation of its parameters. On top of negating the parameters, on must also swap the last two. Using CUDA-Q's parameter ordering for `U3` the following should hold:

```
u3<adj>(a, b, c) != u3(-a, -b, -c)

# The adjoint is:
u3<adj>(a, b, c) == u3(-a, -c, -b)
```

The following test shouldn't be passing:

https://github.com/NVIDIA/cuda-quantum/blob/b9e780db1be6a818502b8f95bf799caefbe7d1ab/unittests/qis/QubitQISTester.cpp#L419-L436

The `u3` mimicking `rx` should be (the minus sign on the second parameter):
```c++
     u3(1.1, -M_PI_2, M_PI_2, q); 
     // rx<adj>(angle) = u3<adj>(angle, -pi/2, pi/2) 
     u3<cudaq::adj>(1.1, -M_PI_2, M_PI_2, q); 
```

Fixing the test case for the correct code makes it not pass, thus we must be handling the `u3`'s adjoint wrong in the compiler and/or in the simulators. 

These are a couple of places in the code base where the adjoint of `u3` might be mishandled (Only negating the parameters):

https://github.com/NVIDIA/cuda-quantum/blob/b9e780db1be6a818502b8f95bf799caefbe7d1ab/lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp#L1299-L1304

https://github.com/NVIDIA/cuda-quantum/blob/b9e780db1be6a818502b8f95bf799caefbe7d1ab/lib/Optimizer/Transforms/DecompositionPatterns.cpp#L1605-L1609

### Expected behavior

Correctly handling the adjoint of `u3` 

Not a regression

### Environment

- **CUDA-Q version**: main (b9e780db1be6a818502b8f95bf799caefbe7d1ab)

### Suggestions

_No response_

schweitzpgi:
Fix bugs with the u3<adj> operator.

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -1519,7 +1519,7 @@ def test_u3_adj():
         # implement Rx gate with U3
         u3(1.1, -np.pi / 2, np.pi / 2, q)
         # rx.adj(angle) = u3.adj(angle, pi/2, -pi/2)
-        u3.adj(1.1, np.pi / 2, -np.pi / 2, q)
+        u3.adj(1.1, -np.pi / 2, np.pi / 2, q)
 
         # implement Ry gate with U3
         u3(1.1, 0, 0, q)
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -1518,7 +1518,7 @@ def test_u3_adj():
 
         # implement Rx gate with U3
         u3(1.1, -np.pi / 2, np.pi / 2, q)
-        # rx.adj(angle) = u3.adj(angle, pi/2, -pi/2)
+        # rx.adj(angle) = u3.adj(angle, -pi/2, pi/2)
         u3.adj(1.1, -np.pi / 2, np.pi / 2, q)
 
         # implement Ry gate with U3
--- a/unittests/qis/QubitQISTester.cpp
+++ b/unittests/qis/QubitQISTester.cpp
@@ -422,7 +422,7 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
     // mimic Rx gate
     u3(1.1, -M_PI_2, M_PI_2, q);
     // rx<adj>(angle) = u3<adj>(angle, pi/2, -pi/2)
-    u3<cudaq::adj>(1.1, M_PI_2, -M_PI_2, q);
+    u3<cudaq::adj>(1.1, -M_PI_2, M_PI_2, q);
     // mimic Ry gate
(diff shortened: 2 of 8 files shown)


unittests/qis/QubitQISTester.cpp
```diff
@@ -422,7 +422,7 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
     // mimic Rx gate
     u3(1.1, -M_PI_2, M_PI_2, q);
     // rx<adj>(angle) = u3<adj>(angle, pi/2, -pi/2)
-    u3<cudaq::adj>(1.1, M_PI_2, -M_PI_2, q);
+    u3<cudaq::adj>(1.1, -M_PI_2, M_PI_2, q);
```
khalatepradnya, line 425: There's a Python counterpart to this test - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_kernel_features.py#L1513 which should also be updated similarly.

schweitzpgi: See 471f0b85

unittests/qis/QubitQISTester.cpp
```diff
@@ -422,7 +422,7 @@ CUDAQ_TEST(QubitQISTester, checkU3Adj) {
     // mimic Rx gate
     u3(1.1, -M_PI_2, M_PI_2, q);
     // rx<adj>(angle) = u3<adj>(angle, pi/2, -pi/2)
```
boschmitt, line 424:
```suggestion
    // rx<adj>(angle) = u3<adj>(angle, -pi/2, pi/2)
```

python/tests/kernel/test_kernel_features.py
```diff
@@ -1519,7 +1519,7 @@ def rotation_adjoint_test():
         # implement Rx gate with U3
         u3(1.1, -np.pi / 2, np.pi / 2, q)
         # rx.adj(angle) = u3.adj(angle, pi/2, -pi/2)
```
boschmitt, line 1521:
```suggestion
        # rx.adj(angle) = u3.adj(angle, -pi/2, pi/2)
```

boschmitt (approved):
Thanks for the work Eric. Couple of suggestions to fix the comments too.

khalatepradnya (approved):
LGTM! Thanks, Eric.

PR #2868, opened 2025-05-06
[Python] [testing] Restore the symbolic link to `system_models.py` from the `integrators`

khalatepradnya:
* Follow-up to PR #2817 
* Should fix the error -
  `E   ModuleNotFoundError: No module named 'system_models'`

### DIFF
--- a/python/tests/dynamics/integrators/system_models.py
+++ b/python/tests/dynamics/integrators/system_models.py
@@ -1,1 +1,1 @@
-/workspaces/cuda-quantum/python/tests/dynamics/system_models.py
+../system_models.py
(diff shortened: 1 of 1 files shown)


python/tests/dynamics/integrators/system_models.py
```diff
@@ -0,0 +1 @@
+/workspaces/cuda-quantum/python/tests/dynamics/system_models.py
```
bmhowe23, line 1: Should this be a relative path instead of a fully qualified path?

khalatepradnya: Fixed in 3ce3031eb7b0975a19e310114c41e7a1e321bb14

PR #2874, opened 2025-05-07
Refactor remote analog targets

khalatepradnya:
* Define a base class (`AnalogRemoteRESTQPU`) for common functionality
* Manually tested all examples for `quera` and `pasqal` targets in both C++ and Python frontends.


nvidia-dobri (approved):
This is great, sorry for the delay! We should also look at unifying the server helpers in a follow-up, should be a good chunk of shared code as well.

PR #2881, opened 2025-05-07
[python] Handle returning values in direct calls and cudaq::run

annagrin:
### Description
Handle returning values via additional return argument

The current implementation incorrectly passed a `NoneType` return type to `pyAltLaunchKernel`,
which caused the `thunk` function to write to offset 0 in the argument buffer, causing seg faults.

To fix the issue, this PR:
- split creating IR and launching it into separate functions
- make sure we pass correct return type to the launcher
- handle returning aggregate types in direct calls
- add tests

<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2852</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2889</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2948</strike>
<strike>Requires: https://github.com/NVIDIA/cuda-quantum/pull/2950</strike>


lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -14,7 +14,6 @@ void cudaq::opt::commonPipelineConvertToQIR(PassManager &pm,
                                             StringRef codeGenFor,
                                             StringRef passConfigAs) {
   pm.addNestedPass<func::FuncOp>(createApplyControlNegations());
-  addAggressiveEarlyInlining(pm);
```
khalatepradnya, line 17: Q: Isn't this common pipeline? Won't it affect C++?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -311,14 +331,50 @@ void storeCapturedData(ExecutionEngine *jit, const std::string &kernelName) {
...
+                    void *rawArgs, std::size_t size, std::uint32_t returnOffset,
+                    const std::vector<std::string> &names) {
+  auto &platform = cudaq::get_platform();
+  auto isRemoteSimulator = platform.get_remote_capabilities().isRemoteSimulator;
+  auto isQuantumDevice =
+      !isRemoteSimulator && (platform.is_remote() || platform.is_emulated());
+
+  if (isRemoteSimulator) {
+    // Remote simulator - use altLaunchKernel to support returning values.
+    // TODO: after cudaq::run support this should be merged with the quantum
```
khalatepradnya, line 345: Q about this `TODO` comment - isn't this the branch for `cudaq::run` support?

lib/Optimizer/CodeGen/Pipelines.cpp
annagrin: added missing inlining in cudaq-translate

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
annagrin: This is an existing comment that just got moved, will need to revisit this after the support for cudaq,run is done

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -282,6 +282,9 @@ class GenerateKernelExecution
     auto thunkTy = cudaq::opt::marshal::getThunkType(ctx);
     auto thunk =
         builder.create<func::FuncOp>(loc, classNameStr + ".thunk", thunkTy);
+
+    thunk->setAttr("noinline", builder.getUnitAttr());
+    thunk->setAttr("optnone", builder.getUnitAttr());
```
schweitzpgi, line 287: I'm not sure I follow why we're changing when inlining is done *and* changing what gets inlined.

IIRC, the issue was that a kernel was being inlined into its thunk function before it was synthesized, which meant the synthesis was happening too late. If we're changing the order there, do synthesis and then do inlining, it no longer matters if the kernel is inlined into the thunk, right?

schweitzpgi: Also, disabling *any* optimization on a thunk function seems punitive. What is the justification for doing that?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -460,6 +472,12 @@ def __call__(self, *args):
                 mlirType = self.argTypes[i]
                 continue
 
+            if cc.StructType.isinstance(mlirType):
+                if cc.StructType.isinstance(self.argTypes[i]):
+                    memberTys = cc.StructType.getTypes(mlirType)
+                    if len(memberTys) == 0:
+                        mlirType = self.argTypes[i]
```
schweitzpgi, line 479: I'm kind of confused by this loop. There are cases above here that set `mlirType` and then do a continue. But the loop will just advance to the next argument and its type. So what is going on there? Why set `mlirType` when it will be overwritten with the next argument type?

In this case, we reset the `mlirType` to whatever it was before we converted it to an MLIR type if and only if it has no members. And then we fall through to the rest of this case logic. The obvious question is why are we doing that? Is there some sort of product type that gets converted incorrectly? What does undoing the conversion to an MLIR type actually intending to do?

python/cudaq/kernel/utils.py
```diff
@@ -289,6 +289,10 @@ def pyInstanceFromName(name: str):
         return int(0)
     if name == 'float':
         return float(0.0)
+    if name in ['numpy.float32', 'np.float32']:
```
schweitzpgi, line 292: Is `'np'` an actual namespace or a typical spelling of an alias for `'numpy'` in code? If `np` is really an alias of `numpy`, we have potential bugs here. Someone could use the `np` alias to alias some other module namespace entirely, such as
```
import fake_numpy as np
```

annagrin: Yes this is a problem with our bridge being a prototype still. There are issues like this in many places - Bruno is working on a rewrite that will support proper name resolution.

python/cudaq/kernel/kernel_decorator.py
annagrin: The function is trying to cast actuals to formal types, so it sets the type of actual to the type of formal when the cast succeeds. The confusing part in struct type case is the check for the zero length; I will add a comment to clarify. This is a workaround for an issue I don't know how to solve properly yet - in a case of a direct call, actual tuple argument type is just "tuple", with no other details available. I converted it to a tuple of 0 length in `mlirTypeFromPyType` and allow it to be casted here to any tuple. I need to do some investigation to see how it is handed in `sample` and `run` cases which seem to show types with more information.

lib/Optimizer/Transforms/GenKernelExecution.cpp
annagrin: I think I can remove those as they don't seem to work anyway, and the problem is solved by reordering the passes.

Details:  It was me trying to not inline the entry point function into the trunk before the `ReturnToLogOutput` pass. AFAIK the inlining is done by the canonicalize (surprisingly, not the inline pass) so I had to try `no opt` as well. Neither of those attributes seem to help, so I moved the aggressive inlining pass out of `commonPipelineConvertToQIR` (we already have it in all the pipelines for quantum devices, and I had to add it separately in `cudaq-translate` (for QIR profiles).

annagrin: Removed, thanks!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -534,8 +559,10 @@ py::object readPyObject(mlir::Type ty, char *arg) {
   return py_ext::convert<T>(concrete);
 }
 
-py::object convertResult(mlir::func::FuncOp kernelFuncOp, mlir::Type ty,
-                         char *data, std::size_t size) {
+py::object convertResult(mlir::ModuleOp module, mlir::func::FuncOp kernelFuncOp,
```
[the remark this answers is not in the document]
annagrin: done

python/tests/kernel/test_direct_call_return_kernel.py
```diff
@@ -0,0 +1,617 @@
...
+    class MyClass:
+        x: list[int]
+        y: bool
+
+    @cudaq.kernel
+    def test_return_dataclass(n: int, t: MyClass) -> MyClass:
+        qubits = cudaq.qvector(n)
+        return t
+
+    # TODO: RuntimeError: Tuple size mismatch in value and label
```
khalatepradnya, line 534: Q: What's the cause of this error?

khalatepradnya (approved):
I reviewed all the tests.

python/tests/kernel/test_direct_call_return_kernel.py
annagrin: This is due to missing support for return of nested aggregate types. I will update the comment.

annagrin: Updated here and in other places

python/tests/kernel/test_run_kernel.py
```diff
@@ -630,23 +609,50 @@ def test_return_tuple_int_bool():
...
+    assert results[1] == (-13, True)
+
+
+def test_return_tuple_int32_bool():
+
+    @cudaq.kernel
+    def simple_tuple_int32_bool_no_args() -> tuple[np.int32, bool]:
+        return (-13, True)
+
+    # FIXME: CompilerError: test_run_kernel.py:634: error:
```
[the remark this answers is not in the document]
annagrin: It's type promotion that we don't support yet for tuples (regular python does), I updated the comment to be clear and added a similar test for direct calls tests. Will fix it in another PR.

Issue #1813, opened 2024-06-13
DepolarizationChannel not completely depolarizing


jin-sungk:
### Required prerequisites

- [X] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When I set the error probably to 1 using the cudaq.DepolarizationChannel to a given gate and measure the output, I expect the expectation value to be 0 in any basis. I observe a consistent <Z> of 0.338 after applying an x gate with an error rate of 1.  Further testing with a Rx gate with varying angles that Z expectation correlates with the rotation angle, even with a depolarization rate of 1. 

### Steps to reproduce the bug

```
cudaq.set_target('density-matrix-cpu')

def rabi_noisy(time,shots):

    kernel=cudaq.make_kernel()

## set depolarization channel to 1 for rx gate
    noise_model=cudaq.NoiseModel()
    noise_model.add_channel("rx",[0],cudaq.DepolarizationChannel(1))

## vary the rx rotation angle
    qubits=kernel.qalloc(1)
    kernel.rx(time,qubits[0])
    kernel.mz(qubits[0])
    counts=cudaq.sample(kernel,shots_count=shots,noise_model=noise_model)
    exp_z=counts.expectation()
    return exp_z


angles = [x*0.1 for x in range(0,62)]
shots=10000

exp_z=[]

for angle in angles:
    exp_z.append(rabi_noisy(angle,shots))
```
![image](https://github.com/NVIDIA/cuda-quantum/assets/127429095/977833cf-76bf-46e1-974e-0b40f8044869)

### Expected behavior

The <Z> expectation should be 0 for all Rx angles with DepolarizationChannel(1)

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 0.7.1
- **Python version**: 3.10.12
- **C++ compiler**: 
- **Operating system**: macos Sonoma 14.5

bettinaheim:
[...]

PR #2884, opened 2025-05-08
[Testing] Fix noise channel test: use `p=0.75` for complete depolarizing channel

1tnguyen:
### Description

The depolarizing noise channel test used an incorrect `p` value to create a complete depolarizing channel. The check conditions are designed for a maximally mixed state.

This leads to test instability, especially for tensornet trajectory-based simulators, where the noise channel sampling doesn't support setting the random seed.

This relates to https://github.com/NVIDIA/cuda-quantum/issues/1813, for which we have updated the documentation accordingly.

### DIFF
--- a/unittests/integration/noise_tester.cpp
+++ b/unittests/integration/noise_tester.cpp
@@ -429,13 +429,17 @@ CUDAQ_TEST(NoiseTest, checkApplySimplePauliErrors) {
 
 CUDAQ_TEST(NoiseTest, checkDepolTypeSimple) {
   cudaq::set_random_seed(13);
-  cudaq::depolarization_channel depol(1.);
+  // Complete depolarizing channel
+  // (https://en.wikipedia.org/wiki/Quantum_depolarizing_channel) to produce a
+  // maximally-mixed state: lambda = 1.0 => p/3 == lambda/4 => p = 3/4 = 0.75
+  cudaq::depolarization_channel depol(0.75);
   cudaq::noise_model noise;
   noise.add_channel<cudaq::types::x>({0}, depol);
   cudaq::set_noise(noise);
   auto counts = cudaq::sample(xOp{});
   counts.dump();
   EXPECT_EQ(2, counts.size());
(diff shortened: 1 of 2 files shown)


unittests/integration/noise_tester.cpp
```diff
@@ -429,13 +429,17 @@ CUDAQ_TEST(NoiseTest, checkApplySimplePauliErrors) {
 
 CUDAQ_TEST(NoiseTest, checkDepolTypeSimple) {
   cudaq::set_random_seed(13);
-  cudaq::depolarization_channel depol(1.);
+  // Complete depolarizing channel
+  // (https://en.wikipedia.org/wiki/Quantum_depolarizing_channel) to produce a
+  // maximally-mixed state: lambda = 1.0 => p/3 == lambda/4 => p = 3/4 = 0.75
+  cudaq::depolarization_channel depol(0.75);
```
bmhowe23, line 435: Does a similar change need to be made to https://github.com/NVIDIA/cuda-quantum/blob/8d045bab2b13c8f8f02fb806fb2e5ca38a8da795/docs/sphinx/examples/cpp/basics/noise_depolarization.cpp#L25 and https://github.com/NVIDIA/cuda-quantum/blob/8d045bab2b13c8f8f02fb806fb2e5ca38a8da795/unittests/integration/observe_result_tester.cpp#L55?

bmhowe23 (approved):
Thanks, Thien.

unittests/integration/noise_tester.cpp
1tnguyen: Good point.

Indeed, the example also needs to be updated since its comments refer to a 50/50 distribution, but the result is more like (30/70). Fixed in (https://github.com/NVIDIA/cuda-quantum/pull/2884/commits/14fecb3c497cf39b9aba2b92ef915756701f4aeb).

The observe test doesn't seem to rely on the actual result statistics, so we can keep it as-is.

PR #2885, opened 2025-05-08
Grovers algorithm added to applications

mmvandieren:
### Description

Added Grover's Algorithm notebook to the documentation page.  Also added the image files and edited the applications.rst so that the notebook gets added to the docs webpage.


mmvandieren:
> @mmvandieren Have you tried to run this notebook locally?

Yes.  It was working fine in version 0.9.1.  I just tried it in version 0.10.0 and the last cell is crashing

mmvandieren:
> LGTM. Thanks @mmvandieren.

Big thanks goes to @BorisTsv !

mmvandieren:
Just noticed the licensing code block at the top of the notebook needs removed.  I'll do this now.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -235,6 +235,7 @@ fmt
 fourier
 frontend
 frontends
+grovers
```
bettinaheim, line 238: Why did you need to add this to the spellcheck? Did you try without adding that? If it is because of the index in applications.rst, it's a bit suboptimal, since we wouldn't want this spelling to be in text (should be Grover's there). 
Please use `|:spellcheck-disable:| text |:spellcheck-enable:|`  instead and remove it from here.

PR #2889, opened 2025-05-08
[python] Support creation and field access for tuples

annagrin:
### Description
Support creating and field access for tuples:

```python
@cudaq.kernel
def kernel() -> int:
    t = (-42, True, 13.5)
    return t[0]

ret = kernel()
assert ret == -42
```


khalatepradnya (approved):
LGTM.

PR #2890, opened 2025-05-08
[WIP] Library restructure

khalatepradnya:

### DIFF
--- a/runtime/common/CMakeLists.txt
+++ b/runtime/common/CMakeLists.txt
@@ -172,3 +172,29 @@ target_link_libraries(cudaq-mlir-runtime
 )
 
 install(TARGETS cudaq-mlir-runtime EXPORT cudaq-targets DESTINATION lib)
+
+## ----- Data layout extraction Library -----
+## This library is used by runtime to get the LLVM data layout for a specific 
+## kernel. This will be used by the `run` API and hence a dependency for the
+## `cudaq` library.
+
+add_library(cudaq-layout-extractor
+  SHARED
+    LayoutExtractor.cpp
+)
+set_property(GLOBAL APPEND PROPERTY CUDAQ_RUNTIME_LIBS cudaq-layout-extractor)
+
+target_include_directories(cudaq-layout-extractor
+  PRIVATE .
+    ${CMAKE_SOURCE_DIR}/runtime)
+
+target_link_libraries(cudaq-layout-extractor
+  PRIVATE
+    cudaq-mlir-core
+)
+install(TARGETS cudaq-layout-extractor EXPORT cudaq-layout-extractor-targets DESTINATION lib)
+
+install(EXPORT cudaq-layout-extractor-targets
+        FILE CUDAQLayoutExtractorTargets.cmake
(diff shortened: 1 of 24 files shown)


runtime/common/CMakeLists.txt
```diff
@@ -86,19 +143,14 @@ endif()
 
 add_library(cudaq-mlir-runtime
   SHARED
-    ArgumentConversion.cpp
     Environment.cpp
-    JIT.cpp
     Logger.cpp
```
amccaskey, line 147: Seeing Logger.cpp and Environment.cpp here for the first time - these are part of common, I wonder why they are here too? Shouldn't be. 

With that, then you don't even have a cudaq-mlir-runtime library any more. You would just have the cudaq-mlir-core and the cudaq-mlir-compiler.

runtime/common/CMakeLists.txt
```diff
@@ -77,6 +77,63 @@ if(EXISTS "$ENV{CURL_INSTALL_PREFIX}/cacert.pem")
...
+    MLIRTransforms
+    MLIRTargetLLVMIRExport
+    MLIRLLVMCommonConversion
+    MLIRLLVMToLLVMIRTranslation
+    fmt::fmt-header-only
+)
+install(TARGETS ${LIBRARY_NAME} EXPORT cudaq-targets DESTINATION lib)
+
+## ----- MLIR-dependent all compiler components including argument converter -----
+set(LIBRARY_NAME cudaq-mlir-compiler)
```
amccaskey, line 114: all this library provides is argument synthesis via the ArgumentConversion? Maybe you should call it cudaq-arg-synth or something that isn't as general as mlir-compiler?

runtime/cudaq/platform/default/rest/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ target_link_libraries(cudaq-rest-qpu
     cudaq-operator
     cudaq-common 
   PRIVATE 
-    cudaq-mlir-runtime 
+    cudaq-mlir-compiler
+    nvqir
```
amccaskey, line 35: Do you really need nvqir here? Or do you just need it when you compile a unit test or executable? If I look in BaseRemoteRESTQPU, I don't think I see any explicit callouts to nvqir symbols. They will be called when we emulate  via simulation, but those symbols need not be here at this time I would think (goes toward my whole thing about decoupling these pieces for better modularity)

runtime/cudaq/platform/fermioniq/CMakeLists.txt
```diff
@@ -22,7 +22,8 @@ target_link_libraries(${LIBRARY_NAME}
     cudaq-common 
   PRIVATE
     pthread
-    cudaq-mlir-runtime 
+    cudaq-mlir-compiler
+    nvqir
```
amccaskey, line 26: same here as above comment

khalatepradnya:
Closing this PR since the approach proposed here didn't pan out for Python (LLVM static initialization issue).

PR #2891, opened 2025-05-09
Gracefully handle expectation of an empty `sample_result`

1tnguyen:
### Description

This is a regression since PR#2710.
Previously, `sample_result::expectation` would return 0.0 if the register name could not be found. 
In 2710, we changed it to a 'strict' version (`retrieve_result`), which throws.

This PR changes it back to a non-throw `try_retrieve_result` and returns 0.0 if not found.

This caused problems (exceptions) when observing an empty spin op, e.g., when distributing observe over more QPUs than the number of terms.

### DIFF
--- a/runtime/common/MeasureCounts.cpp
+++ b/runtime/common/MeasureCounts.cpp
@@ -366,7 +366,10 @@ bool sample_result::has_expectation(const std::string_view registerName) const {
 }
 
 double sample_result::expectation(const std::string_view registerName) const {
-  const auto &result = retrieve_result(registerName.data());
+  auto [found, result] = try_retrieve_result(registerName.data());
+  if (!found)
+    return 0.0;
+
   if (result.expectationValue.has_value())
     return result.expectationValue.value();
 


runtime/common/MeasureCounts.cpp
```diff
@@ -366,7 +366,10 @@ bool sample_result::has_expectation(const std::string_view registerName) const {
 }
 
 double sample_result::expectation(const std::string_view registerName) const {
-  const auto &result = retrieve_result(registerName.data());
+  auto [found, result] = try_retrieve_result(registerName.data());
+  if (!found)
+    return 0.0;
```
mitchdz, line 371: Is there any concern of this being ambiguous and misconceived as an actual expectation value of `0.0`?

1tnguyen: In fact, it's consistent to return 0.0 as the actual expectation. 
An empty operator meant `0.0*I`, and should have zero expectation value.
For example, with the case of `H = 1.23 * spin.x(0) + 2.345 * spin.y(0)` to be distributed over 4 QPUs, we can think of it as adding 2 zero terms, which should have zero expectation values.

mitchdz (approved):

runtime/common/MeasureCounts.cpp
bmhowe23: I think that it is valid to add 2 extra zero terms in the case you described, but it may not be valid in other cases. I know that @bettinaheim intentionally made this change in 2710 to avoid returning valid-looking zero values for `expectation("this_reg_does_not_exist")`. Shouldn't it be possible to avoid the `expectation` call on cases where one of the QPUs is processing a circuit with zero terms?

1tnguyen: Yea, the zero padding in CUDA-Q is from here:

https://github.com/NVIDIA/cuda-quantum/blob/7dbcd1a066fc8612d28ed012aac6d11082cc866c/runtime/cudaq/operators/sum_op.cpp#L1656-L1663

We could consider updating that to prevent the need for calling observe with an empty spin operator.
Since this zero padding is existing behavior (`distribute_terms` always returns a vector of `numChunks` size), we might do it as a separate follow-up.

bmhowe23: OK, saving that for a follow-up PR sounds reasonable to me.

PR #2894, opened 2025-05-09
[Python] Support passing nested lists as arguments

annagrin:
### Description
[Python] Support passing nested lists as arguments

Example:
```python
@cudaq.kernel
def kernel(qubit_num: int, words: list[list[cudaq.pauli_word]], theta: list[float]):
    
    qubits = cudaq.qvector(qubit_num)
    for i in range(len(words)):
        for term in words[i]:
            exp_pauli(theta[i], qubits, term)

words = [['YZXI', 'XZYI'], ['IYZX', 'IXZY']]
counts = cudaq.sample(kernel, 4, words, [0.5, 0.5])
assert len(counts) == 4
assert '0000' in counts
assert '0101' in counts
assert '1010' in counts
assert '1111' in counts
```


python/tests/kernel/test_kernel_features.py
```diff
@@ -537,6 +537,118 @@ def kernel4(myList: List[int]):
...
+                exp_pauli(theta[i], qubits, term)
+
+    words = [['YZXI', 'XZYI'], ['IYZX', 'IXZY']]
+    counts = cudaq.sample(kernel, 4, words, [0.5, 0.5])
+    assert len(counts) == 4
+    assert '0000' in counts
+    assert '0101' in counts
+    assert '1010' in counts
+    assert '1111' in counts
+
```
khalatepradnya, line 651: Nice set of tests! 
Can we add a couple more to cover - what's not supported and the scenario for unintentional broadcast (requirement to have an integer arg before list)?

khalatepradnya (approved):
LGTM. Thanks, Anna!
Tests can be extended as a follow-up.

python/tests/kernel/test_kernel_features.py
annagrin: will do as a follow-up, thanks!

python/cudaq/dbg/__init__.py
```diff
@@ -6,4 +6,4 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-from .ast import print_i64, print_f64
+from .ast import print_i1, print_i64, print_f64
```
schweitzpgi, line 9: Is it possible to have a `kernel_print` function that takes a descriptor argument (like C's `printf`) instead of having all these different functions with the type as part of the name?

python/cudaq/kernel/utils.py
```diff
@@ -325,7 +325,7 @@ def mlirTypeFromPyType(argType, ctx, **kwargs):
             return cc.StdvecType.get(ctx, mlirTypeFromPyType(np.complex64, ctx))
 
         if isinstance(argInstance[0], pauli_word):
-            return cc.StdvecType.get(ctx, cc.CharspanType.get(ctx))
+            return cc.StdvecType.get(ctx, mlirTypeFromPyType(pauli_word, ctx))
```
schweitzpgi, line 328: Is `mlirTypeFromPyType` only used for arguments?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -388,6 +388,17 @@ def from_json(jStr, overrideDict=None):
             location=j['location'],
             overrideGlobalScopedVars=overrideDict)
 
+    def __convertStringsToPauli__(self, arg):
+        if isinstance(arg, str):
+            # Only allow `pauli_word` as string input
+            emitErrorIfInvalidPauli(arg)
```
schweitzpgi, line 394: Is there a test that triggers this error?

python/tests/kernel/test_kernel_features.py
schweitzpgi: We should track any follow-up work with an issue, so it's not lost by accident.

python/utils/OpaqueArguments.h
```diff
@@ -274,6 +273,92 @@ inline void handleStructMemberVariable(void *data, std::size_t offset,
...
+  llvm::TypeSwitch<Type, void>(eleTy)
+      .Case([&](IntegerType ty) {
+        if (ty.isInteger(1))
+          appendValue.template operator()<bool>(
+              list, data, [](py::handle v, std::size_t i) {
+                checkListElementType<py::bool_>(v, i);
+                return v.cast<bool>();
+              });
+        else
+          appendValue.template operator()<std::size_t>(
```
schweitzpgi, line 298: What about i16, i24, i32?

python/utils/OpaqueArguments.h
```diff
@@ -274,6 +273,92 @@ inline void handleStructMemberVariable(void *data, std::size_t offset,
...
+/// value into the dynamically-constructed vector.
+inline void handleVectorElements(void *data, Type eleTy, py::list list) {
+  auto appendValue = []<typename T>(py::list list, void *data,
+                                    auto &&converter) {
+    std::vector<T> *values = new std::vector<T>(list.size());
+    for (std::size_t i = 0; auto &v : list) {
+      auto converted = converter(v, i);
+      (*values)[i++] = converted;
+    }
+    std::memcpy(((char *)data), values, sizeof(std::vector<T>));
```
schweitzpgi, line 286: This must leak memory since `values` silently falls out of scope. The `new vector` is creating a vector object (say at memory address _A_) and an associated buffer that can hold `list.size()` elements (say at address _B_) and that buffer is initialized to the default value of `T` (aka, `T{}`). So after the new `values` contains the address _A_ and the pointers at _A_ point to _B_.

The `memcpy` is bitwise copying the vector object. So the pointers at _A_ and copied to `data`. At this point there are two vector objects with identical data and assumed ownership.

When this function exits, `values` still points to _A_ and there is nothing to delete _A_ from the heap. Which is semi-good, because the `vector::~vector` would, by the rule of ownership, free everything at _B_, which would mean `data` would have pointers to freed memory. Instead of that bug though, we are leaking _A_.

annagrin: I tried to fix it. Let me know what you think.

python/tests/kernel/test_kernel_features.py
annagrin: added the tests for broadcast

python/utils/OpaqueArguments.h
annagrin: added support for those

python/cudaq/kernel/utils.py
schweitzpgi: I was asking because some cases have a contextual nuance.

Say we have `list[int]`. Is this type *always* going to be `!stdvec<i32>` in the IR? The answer is no. If it is an argument then we would likely use `!stdvec<i32>` since we don't want to pass ownership to the kernel (necessarily). However, if this is a type of a local variable then `!stdvec<i32>` is wrong since it is a span (reference) over another object. Our local variable is itself an object, so it requires storage. The required type there for the allocation is `!array<i32 x size>`. Of course once we have the storage for the object, we can always wrap it in a span as a `!stdvec<i32>`. Note that a `!charspan<n>` does not have any storage either.

schweitzpgi (approved):
Did you double check with valgrind that this isn't leaking memory? It looks ok on my skimming through the code.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -313,21 +316,42 @@ def isCastable(self, fromTy, toTy):
...
+                    return bool(value)
+                if intToTy.width == 8:
+                    return np.int8(value)
+                if intToTy.width == 16:
+                    return np.int16(value)
+                if intToTy.width == 32:
+                    return np.int32(value)
+                if intToTy.width == 64:
+                    return int(value)
+
```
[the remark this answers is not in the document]
annagrin: it returns the original value if casting didn't happen, we have a generic type check in the caller after this call. I didn't change the existing logic, just added more cases and allowed recursive types.

PR #2896, opened 2025-05-09
Update mock server test timeouts

bmhowe23:
Importing `cudaq` is taking a long time, especially in Debug builds. Since QA is doing some of their testing with Debug builds, we need to extend the timeouts for some mock server tests.

See nvbug 5268521.

### DIFF
--- a/unittests/backends/anyon/AnyonStartServerAndTest.sh.in
+++ b/unittests/backends/anyon/AnyonStartServerAndTest.sh.in
@@ -43,7 +43,7 @@ n=0
 while ! checkServerConnection; do
   sleep 1
   n=$((n+1))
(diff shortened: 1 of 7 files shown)


mitchdz:
I'm investigating the cause of the slow import time, but for now this looks fine to me.

bmhowe23:
> I'm investigating the cause of the slow import time, but for now this looks fine to me.

From what I can tell, it is approximately the same as 0.10, so I don't think we have a regression since 0.10. Let me know you disagree.

bmhowe23:
> Does that mean it was not caught in 0.10?

Yes. I think we definitely want to improve it, but since it appears to be no worse than the already-released 0.10, I am inclined not hold up the 0.11 release at this point.

PR #2899, opened 2025-05-12
[nvqir] Fix handling of `phased_rx` gate.

boschmitt:
### Description
The circuit simulator has a dedicated method for applying `phased_rx` operations. If we don't use it, then printing circuits that use this operation will incorrectly print them as "unknown op".

### DIFF
--- a/runtime/nvqir/NVQIR.cpp
+++ b/runtime/nvqir/NVQIR.cpp
@@ -495,11 +495,7 @@ void __quantum__qis__cphase(double d, Qubit *q, Qubit *r) {
 
 void __quantum__qis__phased_rx(double theta, double phi, Qubit *q) {
   auto qI = qubitToSizeT(q);
-  std::complex<double> i(0, 1.);
-  std::vector<std::complex<double>> matrix{
-      std::cos(theta / 2.), -i * std::exp(-i * phi) * std::sin(theta / 2.),
-      -i * std::exp(i * phi) * std::sin(theta / 2.), std::cos(theta / 2.)};
-  nvqir::getCircuitSimulatorInternal()->applyCustomOperation(matrix, {}, {qI});
+  nvqir::getCircuitSimulatorInternal()->phased_rx(theta, phi, qI);
 }
 
 void __quantum__qis__phased_rx__body(double theta, double phi, Qubit *q) {


runtime/nvqir/NVQIR.cpp
```diff
@@ -495,11 +495,7 @@ void __quantum__qis__cphase(double d, Qubit *q, Qubit *r) {
 
 void __quantum__qis__phased_rx(double theta, double phi, Qubit *q) {
   auto qI = qubitToSizeT(q);
-  std::complex<double> i(0, 1.);
-  std::vector<std::complex<double>> matrix{
-      std::cos(theta / 2.), -i * std::exp(-i * phi) * std::sin(theta / 2.),
-      -i * std::exp(i * phi) * std::sin(theta / 2.), std::cos(theta / 2.)};
-  nvqir::getCircuitSimulatorInternal()->applyCustomOperation(matrix, {}, {qI});
+  nvqir::getCircuitSimulatorInternal()->phased_rx(theta, phi, {}, qI);
```
nvidia-dobri, line 498: Signature from header:
```
void phased_rx(const double phi, const double lambda, const std::size_t qubitIdx)
```

Is this right? Or do we need `phased_rx(phi, theta, {}, qI);`? We really need to stop using random greek letters for random angle args 😅

nvidia-dobri (approved):
LGTM with "pls double-check order of args makes sense"

runtime/nvqir/NVQIR.cpp
nvidia-dobri: I guess, also, you can use the overload that doesn't require empty controls, whose signature is actually the one above.

boschmitt: > Is this right? Or do we need phased_rx(phi, theta, {}, qI);? We really need to stop using random greek letters for random angle args

Most likely came from looking at different docs to find out what the matrix is. Usually parameters names change but the order is consistent. 

Modulo parameters names, the matrix deleted here seems to be the same as the one in: https://github.com/NVIDIA/cuda-quantum/blob/18e0ff0ba0bd374ed3f4c0f9ae95a133646b3be8/runtime/nvqir/Gates.h#L120-L130

Issue #2892, opened 2025-05-09
MPI error when scaling out beyond 1 server


pioch-02:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Workloads crash with MPI errors when using more than 1 node.
Platform background: DGX H100 with a Slurm and Pyxis, container `nvcr.io/nvidia/quantum/cuda-quantum:cu12-0.10.0`
Job started with `srun --mpi=pmix --container-image=$HOME/nvidia+quantum+cuda-quantum+cu12-0.10.0.sqsh bash -c "(export CUDAQ_MPI_COMM_LIB=${HOME}/libcudaq_distributed_interface_mpi.so; python ${HOME}/mgpu.py)"`

When running with containers the error message looks like:
```
*** An error occurred in MPI_Allgatherv
*** reported by process [351672370,508]
*** on communicator MPI_COMM_WORLD
*** MPI_ERR_TRUNCATE: message truncated
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```

We have also compiled it as an environment module, the crash when using those have more information

```
==== backtrace (tid:3833225) ====
 0 0x0000000000003703 uct_cma_ep_tx_error()  /tmp/UCX/1.14.1/GCCcore-12.3.0/ucx-1.14.1/src/uct/sm/scopy/cma/cma_ep.c:81
 1 0x0000000000003a38 uct_cma_ep_tx()  /tmp/UCX/1.14.1/GCCcore-12.3.0/ucx-1.14.1/src/uct/sm/scopy/cma/cma_ep.c:114
 2 0x000000000001dad5 uct_scopy_ep_progress_tx()  /tmp/UCX/1.14.1/GCCcore-12.3.0/ucx-1.14.1/src/uct/sm/scopy/base/scopy_ep.c:151
[...]

PR #2905, opened 2025-05-13
activate_custom_mpi.sh: do not link against mlir

mitchdz:
See https://github.com/NVIDIA/cuda-quantum/issues/2892

It might be worth it to further trim down what the ELF file links against, but the other files seem benign for now.

```
cudaq@2570547-lcedt:/opt/nvidia/cudaq/distributed_interfaces$ readelf -d libcudaq_distributed_interface_mpi.so  | grep cudaq
 0x000000000000001d (RUNPATH)            Library runpath: [/usr/local/llvm/lib:/opt/nvidia/cudaq/lib:/opt/nvidia/cudaq/lib/plugins:/opt/nvidia/cudaq/distributed_interfaces:/usr/local/openmpi/lib64:/usr/local/openmpi/lib]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-common.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-ensmallen.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-nlopt.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-spin.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-operator.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-comm-plugin.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-pyscf.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-em-default.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudaq-platform-default.so]
 ```

### DIFF
--- a/runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
+++ b/runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
@@ -48,7 +48,9 @@ fi
 
 echo "Using $CXX to build the MPI plugin for MPI installation in $MPI_PATH."
 lib_mpi_plugin="$this_file_dir/libcudaq_distributed_interface_mpi.so"
+# --disable-mlir-links for https://github.com/NVIDIA/cuda-quantum/issues/2892
 $CXX -shared -std=c++17 -fPIC \
+    --disable-mlir-links \
     -I"${MPI_PATH}/include" \
     -I"$this_file_dir" \
     "$this_file_dir/mpi_comm_impl.cpp" \

### CODE REVIEW

File: runtime/cudaq/distributed/builtin/activate_custom_mpi.sh [hunk 1]
```diff
@@ -48,7 +48,9 @@ fi
 
 echo "Using $CXX to build the MPI plugin for MPI installation in $MPI_PATH."
 lib_mpi_plugin="$this_file_dir/libcudaq_distributed_interface_mpi.so"
+# --disable-mlir-links for https://github.com/NVIDIA/cuda-quantum/issues/2892
 $CXX -shared -std=c++17 -fPIC \
+    --disable-mlir-links \
```
schweitzpgi, line 53: I'm curious. What is this option?
```console
% clang++ -c --disable-mlir-links marwa.cpp
clang-16: error: unsupported option '--disable-mlir-links'
%
```

runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
```diff
@@ -48,7 +48,9 @@ fi
 
 echo "Using $CXX to build the MPI plugin for MPI installation in $MPI_PATH."
 lib_mpi_plugin="$this_file_dir/libcudaq_distributed_interface_mpi.so"
+# --disable-mlir-links for https://github.com/NVIDIA/cuda-quantum/issues/2892
```
schweitzpgi, line 51: It's ok to put this sort of documentation (cross referencing) in the commit message rather than in the source code.

runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
mitchdz: That is a good point - This is a `nvq++` only option. So if `$CXX` is, for example `g++` this will fail as is.

runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
mitchdz: Since this is a shell script that gets shipped with the container, I felt like it would be useful having it in the script itself so users could reference it more easily.

schweitzpgi: I don't think that is our process in general.

runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
schweitzpgi: Thanks for the explanation.

runtime/cudaq/distributed/builtin/activate_custom_mpi.sh
mitchdz: I removed the comment in the latest commit.

mitchdz:
/ok to test [591c725](https://github.com/NVIDIA/cuda-quantum/pull/2905/commits/591c725ad068f430eceb3901991efa88a3a88715)

mitchdz:
/ok to test [c5fb8df](https://github.com/NVIDIA/cuda-quantum/pull/2905/commits/c5fb8dfbdb18e23d3b2e8f84d7397c024d33d4fb)

mitchdz:
I will modify this PR shortly to instead of just dropping the mlir links, add a new flag to `nvq++.in`. Something like `--disable-cudaq-links`. This is ultimately what I'd like to do anyways, so let's just do it right the first time rather than having the `--disable-mlir-links` as a bandaid.

PR #2915, opened 2025-05-15
[features/run] Support tuple parsing

khalatepradnya:
This PR adds logic to parse output records containing tuple data by using the data layout info.
Primary changes:
* `runtime/common/RecordLogParser.cpp`: Added logic to handle data layout information during tuple parsing, including alignment and memory allocation. 
* Migrated the `runTheKernel` function from `runtime/cudaq/algorithms/run.cpp` to `runtime/common/RunTheKernel.cpp` for better modularity and added support for data layout extraction. Added a `LayoutExtractor` class to extract data layout information from MLIR modules and integrated it into the `runTheKernel` workflow.

Covered: C++ support

_Note_: Updated the reason for failure in Python tests - will be addressed in a follow-up.


runtime/common/RunTheKernel.cpp
```diff
@@ -0,0 +1,164 @@
+/*******************************************************************************
```
schweitzpgi, line 1: Is this just a copy of `run.cpp` with a name change?

python/tests/kernel/test_run_kernel.py
```diff
@@ -324,7 +333,17 @@ def simple_tuple_int_bool(n: int, t: tuple[int, bool]) -> tuple[int, bool]:
...
+    '''
+    OUTPUT  TUPLE   2       tuple<i64, i1>
+    OUTPUT  INT     13      .0
+    OUTPUT  BOOL    true    .1
+    OUTPUT  TUPLE   2       tuple<i64, i1>
+    OUTPUT  INT     13      .0
+    OUTPUT  BOOL    true    .1
+    '''
     assert len(results) == 2
+    ## FIXME: Fails on arm64 in CI
```
khalatepradnya, line 346: This is most likely because the datalayout string is missing in kernels from Python..

runtime/common/RunTheKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+namespace details {
+/// Extracts data layout information from MLIR modules
+class LayoutExtractor {
+public:
+  std::pair<std::size_t, std::vector<std::size_t>>
+  extractLayout(const std::string &, const std::string &);
+
+private:
+  mlir::MLIRContext *createContext();
+};
```
1tnguyen, line 43: As this class doesn't have any member variables, perhaps we could make these methods `static` so that the downstream code can use `LayoutExtractor::extractLayout(...)` without constructing a class object.

1tnguyen (approved):
👍

runtime/common/RunTheKernel.cpp
khalatepradnya: Done in [9c20489](https://github.com/NVIDIA/cuda-quantum/pull/2915/commits/9c20489ccec2a9a413d1225af4231daa78c3ed46)

Issue #2914, opened 2025-05-14
Invalid Target: (dynamics)


YauheniTalochkaN:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Dear developers,

Could you please help me resolve the following issue with compiling the [heisenberg_model](https://github.com/NVIDIA/cuda-quantum/blob/releases/v0.11.0/docs/sphinx/examples/cpp/dynamics/heisenberg_model.cpp) example? 

When using the command: 
`
nvq++ --target dynamics heisenberg_model.cpp -o heisenberg_model.x 
`

I encounter the error: 
`
nvq++:646: Invalid Target: (dynamics) 
`

CUDA-Q version 0.11.0 was installed from [
install_cuda_quantum_cu12.x86_64 ](https://github.com/NVIDIA/cuda-quantum/releases/tag/0.11.0).

### Steps to reproduce the bug

`
nvq++ --target dynamics heisenberg_model.cpp -o heisenberg_model.x 
`

### Expected behavior

Successful compilation of heisenberg_model.cpp without the "Invalid Target: (dynamics)" error.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 0.11.0
- **C++ compiler**: gcc 14.2.1 & nvq++ 0.11.0
- **Operating system**: Fedora 41

bebora:
I can confirm this occurs when using the pre-built binaries. The container image is not affected. If you compare `nvq++ --list-targets` between the release on Github and the corresponding container (nvcr.io/nvidia/nightly/cuda-quantum:cu12-0.11.0), dynamics is the only target missing.

bebora:
Hi @YauheniTalochkaN, I am a CUDA-Q user and I am not affiliated with NVIDIA. We will have to wait for feedback from the maintainers. New releases are typically published every few months. I suggest using the container version in the meantime if you need the dynamics target.

[...]

PR #2916, opened 2025-05-15
Fixing the dynamics build in the installer

bettinaheim:
This PR should fix the reported issue https://github.com/NVIDIA/cuda-quantum/issues/2914. 
I checked locally that the dynamics backend can run one of the example both in c++ and python in the assets build.


mitchdz (approved):
These changes look fine to me. As a future improvement we should add a C++ test for this and keep that in mind for future backends that migrate from Python.

bettinaheim:
> These changes look fine to me. As a future improvement we should add a C++ test for this and keep that in mind for future backends that migrate from Python.

Yes, I need to think a bit more about it (so it's for a separate PR). The issue with the current validation script is that it automatically tests all targets that are defined. We do explicitly check for the mgpu target, since that one is built on an external repo, but we don't explicitly check that all targets that we build on this repo are actually available. We can make that check explicit, it's just that we need to think to add to that manually every time we add a target, so chances are we may forget.

Issue #630, opened 2023-09-08
Python formatting not enforced in CI


anthony-santana:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

We run a python formatting check in the CI with yapf, but it appears that it's not actually being enforced. This can be confirmed by `pip install yapf`, then pulling down CUDA Quantum main and running `yapf cuda-quantum/python/tests/unittests -r -i`. For main to be passing the CI, yapf should produce no changes to the code. Instead, it has to reformat multiple files -- indicating the python formatting step isn't throwing an error when it should be.

### Steps to reproduce the bug

```
$ pip install yapf
$ yapf cuda-quantum/python/tests/unittests -r -i
```

### Expected behavior

An error should be thrown in the CI if the python files have not been formatted through yapf.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 0.4
- **Python version**: 3.10
- **C++ compiler**: NA
- **Operating system**: NA

bettinaheim:
Hi @anthony-santana, 
I believe the style should be enforced, but certain folders, like the tests, are currently excluded for the style check. You can see which files are excluded [here](https://github.com/NVIDIA/cuda-quantum/blob/77cbcd1ab00ef597727858abb9e248602c53c27d/.github/workflows/repo_checks.yml#L81). I don't remember why I excluded them; it might be that they can be properly formatted, or it might be that some tests make that difficult.

PR #2925, opened 2025-05-18
[nfc] Code formatting on Python test files

khalatepradnya:


khalatepradnya:
> LGTM. Thanks @khalatepradnya. Just curious to know if the CI is not complaining about the fomratting for these files, do we need this PR?

I agree that CI isn't complaining, but I get these when I re-run the formatter locally, so thought of fixing. Although now that I looked at the issues, I see this one - https://github.com/NVIDIA/cuda-quantum/issues/630 originally called it out, so the proper fix would be to address that issue. I'm okay to address that as a follow-up.

Issue #2526, opened 2025-01-21
Implement cudaq::run API


annagrin:
### Required prerequisites

- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

#### Examples (c++)

```c++
bool kernel(int n) __qpu__ {
  cudaq::qvector q(n);
  x(q[0]);
  return mz(q[0])
}


int main() {
  auto results = cudaq::run<bool>(kernel, 2, /*shots=*/1000);
  std::cout << results;
}
```

Output:
`[1,1, ... 1] // 1000 results`

Currently we would like to support returning the following types:

```c++
bool
int
float
double
vector<bool> // For measurements of qvectors (const size after opts)
```

#### Examples (python)

```python
@cudaq.kernel()
def kernel(n: int): bool
  q = cudaq.qvector(n)
  x(q[0]);
  return mz(q[0])


results = cudaq.run(kernel, 2, shots=1000)
print(results)
```

Output:
`[1,1, ... 1] // 1000 results`

Currently we would like to support returning the following types:
```python
bool
int
np.float32
np.float64
list<bool> // For measurements of qvectors (const size after opts)
```

### Tasks

- [ ] Define log parsing API in c++ (no templates)
Idea:
```c++
struct Record {
 void* ptr;
 std::size_t byteSize;
};

class OutputLogParser {
 virtual std::vector<Record> parse(std::istream& s);
};
```
- [ ] Implement log parsing API for backends that will use logs for output (quantum devices and remote simulator)
- [ ] Implement log parsing API for backends that use memory for output (local simulators)
  - will need runtime QIR functions implementation
  - make sure we handle multiple and single shot scenarios
- [ ] Implement `cudaq::run` API in c++
- [ ] Implement `cudaq.run` API in Python
- [ ] Implement retrieving logs from quantum backends (start with quantinuum)
- [ ] Implement creating and retrieving logs from remote simulator server
- [ ] Make sure quantum emulation works with new output log functionality
  - run LLVM generated from QIR on the simulator (needs runtime functions implementations)

PR #2927, opened 2025-05-19
Enable `run` API

khalatepradnya:
Towards: issue https://github.com/NVIDIA/cuda-quantum/issues/2526

This PR introduces new CUDA-Q API - `run` which provides a way to execute a specific kernel with non-void return values. 

For a kernel with return type `T`, one can use `cudaq::run` (C++), or `cudaq.run` (Python) to launch the quantum kernel, specifying the number of shots, the kernel argument values, etc. When the kernel has been executed, control will return to the user code with `cudaq::run` returning a value of `std::vector<T>`. 

Added a new pass, `ReturnToOutputLog`, to convert kernel return values into QIR output logging calls.

This PR covers: 
 - kernels that return primitive types (integers, floating point numbers, Boolean values) and vectors / lists of primitive types supported in both C++ and Python for local and remote simulators
 - kernels that return tuples / structs of primitive types supported in C++ only for local simulators

Not supported:
 - kernels that return nested aggregate data types not yet supported
 - library mode

Note: `-fenable-cudaq-run` flag required in C++ (this flag will be enabled by default in the future)


lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -0,0 +1,244 @@
...
+          Value label = makeLabel(loc, rewriter, labelStr);
+          std::int32_t sz = strTy.getNumMembers();
+          Value size = rewriter.create<arith::ConstantIntOp>(loc, sz, 64);
+          rewriter.create<func::CallOp>(loc, TypeRange{},
+                                        cudaq::opt::QIRTupleRecordOutput,
+                                        ArrayRef<Value>{size, label});
+          std::string preStr = prefix ? prefix->str() : std::string{};
+          for (std::int32_t i = 0; i < sz; ++i) {
+            std::string offset = preStr + std::string(".") + std::to_string(i);
+            Value w = rewriter.create<cudaq::cc::ExtractValueOp>(
```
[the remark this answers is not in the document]
schweitzpgi: It is defined, used on the next line, and then goes out of scope. I'm not sure how much value actual value there is in being overly concerned about the length of the name given this value. Very short names used as shorthand in a same amount of code text seems quite reasonable.

Devil's advocate: Hypothetically speaking, `theMemberOfTheStructRecordAtTheCurrentCursorPosition` is less clear because one has to scan dozens of characters visually and make sure the name isn't the same as `theMemberOfTheVectorAtTheCurrentCursorPosition` used elsewhere.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -834,6 +834,9 @@ class GenerateKernelExecution
     if (failed(irBuilder.loadIntrinsic(module, cudaq::llvmMemCopyIntrinsic)))
       return module.emitError(std::string("could not load ") +
                               cudaq::llvmMemCopyIntrinsic);
+    if (failed(irBuilder.loadIntrinsic(module, cudaq::llvmMemSetIntrinsic)))
+      return module.emitError(std::string("could not load ") +
+                              cudaq::llvmMemSetIntrinsic);
```
schweitzpgi, line 839: This error is "impossible" if the code is correctly built. Ignore the codecov complaint.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -402,6 +411,13 @@ def promoteOperandType(self, ty, operand):
             if IntegerType.isinstance(operand.type):
                 operand = arith.SIToFPOp(ty, operand).result
 
+        if IntegerType.isinstance(ty):
+            if IntegerType.isinstance(operand.type):
+                if IntegerType(ty).width < IntegerType(operand.type).width:
+                    operand = arith.TruncIOp(ty, operand).result
+                else:
+                    operand = arith.ExtSIOp(ty, operand).result
```
schweitzpgi, line 419: Are we sign extending the integer if the length is already correct?

runtime/common/RecordLogParser.h
```diff
@@ -0,0 +1,341 @@
...
+
+  /// Parse string like "array<i32 x 4>"
+  void extractArrayInfo(const std::string &label) {
+    auto isArray = label.find("array");
+    auto lessThan = label.find('<');
+    auto greaterThan = label.find('>');
+    auto x = label.find('x');
+    if ((isArray == std::string::npos) || (lessThan == std::string::npos) ||
+        (greaterThan == std::string::npos) || (x == std::string::npos))
+      throw std::runtime_error("Array label missing keyword");
```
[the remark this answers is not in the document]
khalatepradnya: Array label must be of the type `array<type x size>`. If any of these is missing in the label we throw an error.

python/tests/kernel/test_run_kernel.py
```diff
@@ -0,0 +1,539 @@
...
+    @cudaq.kernel
+    def test_return_dataclass(n: int, t: MyClass) -> MyClass:
+        qubits = cudaq.qvector(n)
+        return t
+
+    # TODO: RuntimeError: Tuple size mismatch in value and label
+    # results = cudaq.run(test_return_dataclass, 2, MyClass([0,1], 18), shots_count=2)
+    # assert len(results) == 2
+    # assert results[0] == MyClass([0,1], 18)
+    # assert results[1] == MyClass([0,1], 18)
```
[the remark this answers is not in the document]
khalatepradnya: Yes, being addressed in Anna's PR https://github.com/NVIDIA/cuda-quantum/pull/2881

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -0,0 +1,244 @@
...
+          std::int32_t sz = arrTy.getSize();
+          Value size = rewriter.create<arith::ConstantIntOp>(loc, sz, 64);
+          rewriter.create<func::CallOp>(loc, TypeRange{},
+                                        cudaq::opt::QIRArrayRecordOutput,
+                                        ArrayRef<Value>{size, label});
+          std::string preStr = prefix ? prefix->str() : std::string{};
+          for (std::int32_t i = 0; i < sz; ++i) {
+            std::string offset = preStr + std::string("[") + std::to_string(i) +
+                                 std::string("]");
+            Value w = rewriter.create<cudaq::cc::ExtractValueOp>(
```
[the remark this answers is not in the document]
khalatepradnya: Marking resolved as per discussion above.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -97,6 +101,15 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
   /// execution locally.
   bool emulate = false;
 
+  /// @brief to indicate the backend support QIR integer computation extension.
+  // Applicable to `qir-adaptive` codegenTranslation only.
+  bool qirIntegerExtension = false;
+
+  /// @brief to indicate the backend support QIR floating point computation
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 84c3d1a701011b3971e8f13ba77f5a891b62b21d

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -489,6 +483,131 @@ py::object readPyObject(mlir::Type ty, char *arg) {
...
+        if (!DataClassRegistry::isRegisteredClass(name))
+          throw std::runtime_error("Dataclass is not registered: " + name);
+
+        // Find class information.
+        auto [cls, attributes] = DataClassRegistry::getClassAttributes(name);
+
+        // Collect field names.
+        std::vector<py::str> fieldNames;
+        for (const auto &[attr_name, unused] : attributes)
+          fieldNames.push_back(py::str(attr_name));
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 84c3d1a701011b3971e8f13ba77f5a891b62b21d

python/runtime/cudaq/algorithms/py_run_async.cpp
```diff
@@ -0,0 +1,84 @@
...
+          std::size_t qpu_id) {
+        kernel.inc_ref();
+        auto &platform = cudaq::get_platform();
+        auto kernelName = kernel.attr("name").cast<std::string>();
+        auto kernelMod = kernel.attr("module").cast<MlirModule>();
+        args = simplifiedValidateInputArguments(args);
+        auto *argData = toOpaqueArgs(args, kernelMod, kernelName);
+        // Note: we only pass the underlying buffer pointer to the async.
+        // functor to load the result data. i.e., no Python object is accessed
+        // during the async. execution.
```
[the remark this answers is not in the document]
khalatepradnya: Fixed in 84c3d1a701011b3971e8f13ba77f5a891b62b21d

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -0,0 +1,104 @@
...
+    if (platform.is_remote())
+      throw std::runtime_error(
+          "Noise model is not supported on remote platforms.");
+  }
+
+  if (shots_count == 0)
+    return {};
+
+  // Launch the kernel in the appropriate context.
+  if (noise_model.has_value())
```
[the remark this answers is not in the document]
khalatepradnya: @annagrin for viz.

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -0,0 +1,104 @@
...
+  mod->setAttr(runtime::enableCudaqRun, mlir::UnitAttr::get(mod->getContext()));
+
+  auto &platform = get_platform();
+  if (noise_model.has_value()) {
+    if (platform.is_remote())
+      throw std::runtime_error(
+          "Noise model is not supported on remote platforms.");
+  }
+
+  if (shots_count == 0)
```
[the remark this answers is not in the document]
annagrin: done

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3256,6 +3280,38 @@ def visit_Subscript(self, node):
                 self.pushValue(cc.LoadOp(eleAddr).result)
                 return
 
+        if cc.StructType.isinstance(var.type):
+            # Return the pointer if someone asked for it
+            if self.subscriptPushPointerValue:
+                self.pushValue(var)
+                return
```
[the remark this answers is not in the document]
annagrin: Added a test with a reference to the bug.

python/tests/kernel/test_run_kernel.py
```diff
@@ -0,0 +1,869 @@
...
+        return MyClass(42.5, 17)
+
+    # TODO: seg fault on running any kernel with no args
+    # results = cudaq.run(simple_dataclass_float_int_no_args, shots_count=2)
+    # assert len(results) == 2
+    # assert results[0] == MyClass(42.5, 17)
+    # assert results[1] == MyClass(42.5, 17)
+
+    @cudaq.kernel
+    def simple_dataclass_float_int(n: int, t: MyClass) -> MyClass:
```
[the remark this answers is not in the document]
annagrin: Added more tests

python/cudaq/kernel/utils.py
```diff
@@ -430,6 +512,17 @@ def mlirTypeToPyType(argType):
         if cc.StateType.isinstance(valueTy):
             return State
 
+    if cc.StructType.isinstance(argType):
+        if (cc.StructType.getName(argType) == "tuple"):
+            elements = [
+                mlirTypeToPyType(v)() for v in cc.StructType.getTypes(argType)
+            ]
+            return type(tuple(elements))
```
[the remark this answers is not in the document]
annagrin: I wanted a type hint, as in `tuple[int, bool]` if the element types are [`int`, `bool`], but this it is not the right way to do that. Fixed and added a test.

python/runtime/cudaq/algorithms/py_run.cpp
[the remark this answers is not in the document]
annagrin: done

python/tests/kernel/test_run_kernel.py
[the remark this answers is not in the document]
khalatepradnya: See 9bfc069013eddad1faac829a6ed1f580b5cee46e

python/runtime/cudaq/algorithms/py_run_async.cpp
```diff
@@ -0,0 +1,84 @@
...
+        result.ready = promise.get_future();
+        QuantumTask wrapped = detail::make_copyable_function(
+            [p = std::move(promise), bufferPtr, shots, &platform, argData,
+             kernelName, kernelMod,
+             noise_model = std::move(noise_model)]() mutable {
+              if (noise_model.has_value())
+                platform.set_noise(&noise_model.value());
+              details::RunResultSpan span = details::runTheKernel(
+                  [&]() mutable {
+                    pyAltLaunchKernel(kernelName, kernelMod, *argData, {});
```
annagrin, line 69: This needs to be updated to use `pyLaunchKernel` (like in py_run.cpp) and pass the correct argument type

khalatepradnya:
> /ok to test 53bc46adffb320f937d8a080ad676813592e0d1e

include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -271,5 +271,19 @@ def QuakeToQIRAPIPrep : Pass<"quake-to-qir-api-prep", "mlir::ModuleOp"> {
   ];
 }
 
+def ReturnToOutputLog : Pass<"return-to-output-log", "mlir::ModuleOp"> {
+  let summary = "Convert a kernel to be compatible with cudaq::run().";
+  let description = [{
+    When the target supports the cudaq::run() launch function, the kernel's
+    return value(s) are translated into QIR output logging functions. This
+    conversion allows the kernel to be executed as a group of shots on the QPU
+    and a log file produced for the data produced by each kernel execution.
```
sacpis, line 280: nit: `and each kernel execution generates a log file.`

python/cudaq/kernel/ast_bridge.py
```diff
@@ -402,6 +411,15 @@ def promoteOperandType(self, ty, operand):
...
+        if IntegerType.isinstance(ty):
+            if F64Type.isinstance(operand.type):
+                operand = arith.FPToSIOp(self.getIntegerType(), operand).result
+            if IntegerType.isinstance(operand.type):
+                if IntegerType(ty).width < IntegerType(operand.type).width:
+                    operand = arith.TruncIOp(ty, operand).result
+                else:
+                    operand = arith.ExtSIOp(ty, operand).result
+
         return operand
```
sacpis, line 423: Should we replace these if statements with elif so that we can avoid evaluating every if statement?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -552,7 +570,10 @@ def getStructMemberIdx(self, memberName, structTy):
         else:
             structName = quake.StruqType.getName(structTy)
         structIdx = None
-        _, userType = globalRegisteredTypes[structName]
+        if not globalRegisteredTypes.isRegisteredClass(structName):
+            self.emitFatalError(f'Dataclass is not registered: {structName})')
```
sacpis, line 574: nit: `Dataclass {structName} is not registered.)`

python/cudaq/runtime/run.py
```diff
@@ -0,0 +1,66 @@
...
+    Defaults to 1. Key-word only.
+  noise_model (Optional[`NoiseModel`]): The optional :class:`NoiseModel`
+    to add noise to the kernel execution on the simulator. Defaults to
+    an empty noise model.
+  `qpu_id` (Optional[int]): The id of the QPU.
+    Defaults to 0. Key-word only.
+
+Returns:
+  `AsyncRunResult`: 
+  A handle, which can be waited on via a `get()` method, which returns an array of `kernel` return values. The length of the list is equal to `shots_count`.
```
sacpis, line 34: nit: A handle, which can be waited on via a `get()` method, returns an array of `kernel` return values. The length of the list is equal to `shots_count`.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -309,14 +382,50 @@ void storeCapturedData(ExecutionEngine *jit, const std::string &kernelName) {
...
+    // Remote simulator - use altLaunchKernel to support returning values.
+    // TODO: after cudaq::run support this should be merged with the quantum
+    // device case.
+    auto *wrapper = new cudaq::ArgWrapper{mod, names, rawArgs};
+    auto dynamicResult = cudaq::altLaunchKernel(
+        name.c_str(), thunk, reinterpret_cast<void *>(wrapper), size,
+        returnOffset);
+    if (dynamicResult.data_buffer || dynamicResult.size)
+      throw std::runtime_error("not implemented: support dynamic results");
+    delete wrapper;
```
sacpis, line 404: Just a question. If we throw runtime_error on the previous line, would we still delete the wrapper?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -902,7 +1099,8 @@ void bindAltLaunchKernel(py::module &mod) {
         if (attr)
           m->setAttr("quake.mangled_name_map", attr);
       },
-      "Synthesize away the callable block argument from the entrypoint in modA "
+      "Synthesize away the callable block argument from the entrypoint in "
+      "modA "
       "with the FuncOp of given name.");
```
sacpis: nit: Are these lines formatted?

python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -426,6 +426,33 @@ def kernel():
         counts = cudaq.sample(kernel)
 
 
+def test_run():
+
+    @cudaq.kernel
+    def simple(numQubits: int) -> int:
```
sacpis, line 432: Should we add tests for other primitive return types as well?

khalatepradnya: This target supports only integer computations.

runtime/common/RecordLogParser.cpp
```diff
@@ -0,0 +1,220 @@
...
+ ******************************************************************************/
+
+#include "RecordLogParser.h"
+#include "Logger.h"
+
+void cudaq::RecordLogParser::parse(const std::string &outputLog) {
+  cudaq::info("Parsing log:\n{}", outputLog);
+  std::vector<std::string> lines = cudaq::split(outputLog, '\n');
+  if (lines.empty())
+    return;
```
sacpis, line 16: Do we need this check?

runtime/common/RecordLogParser.cpp
```diff
@@ -0,0 +1,220 @@
...
+      schema = RecordSchemaType::ORDERED;
+    else
+      throw std::runtime_error("Unknown schema type");
+  }
+  /// TODO: Handle schema version if needed
+}
+
+void cudaq::RecordLogParser::handleMetadata(
+    const std::vector<std::string> &entries) {
+  // Ignore metadata for now
```
sacpis, line 59: Do we need to create a TODO item for this one?

runtime/common/RecordLogParser.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+void cudaq::RecordLogParser::handleStart(
+    const std::vector<std::string> &entries) {
+  // Ignore start of a shot for now
```
sacpis, line 64: Do we need to create a TODO item for this one?

runtime/common/RecordLogParser.cpp
```diff
@@ -0,0 +1,220 @@
...
+  if (recType == "ARRAY") {
+    containerMeta.m_type = ContainerType::ARRAY;
+    containerMeta.elementCount = std::stoul(recValue);
+    if (!recLabel.empty()) {
+      schema = RecordSchemaType::LABELED;
+      containerMeta.extractArrayInfo(recLabel);
+      preallocateArray();
+    }
+    return;
+  }
```
sacpis, line 95: nit: } else if

runtime/common/RecordLogParser.cpp
```diff
@@ -0,0 +1,220 @@
...
+  if (recType == "TUPLE") {
+    containerMeta.m_type = ContainerType::TUPLE;
+    containerMeta.elementCount = std::stoul(recValue);
+    if (!recLabel.empty()) {
+      schema = RecordSchemaType::LABELED;
+      containerMeta.extractTupleInfo(recLabel);
+      preallocateTuple();
+    }
+    return;
+  }
```
sacpis, line 105: nit: } else if

runtime/common/RecordLogParser.cpp
khalatepradnya: No, from functional perspective, this check is not strictly necessary. It can be considered as early return optimization that avoids unnecessary loop setup when there's nothing to process.

runtime/common/RecordLogParser.cpp
khalatepradnya: Need target which emits this hence future implementation.

runtime/common/RecordLogParser.cpp
khalatepradnya: Need target which emits this hence future implementation.

runtime/common/RecordLogParser.cpp
khalatepradnya: Is it really necessary when the `if` code block ends with a return statement ?

runtime/common/RecordLogParser.cpp
khalatepradnya: Is it really necessary when the `if` code block ends with a return statement?

sacpis: I am thinking of making it consistent with the other conditions below. So made a comment to replace `return; }` with `} else if`.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: Good point! No, it won't be deleted. 
Tagging @annagrin for a follow-up.

python/cudaq/kernel/ast_bridge.py
khalatepradnya: I think it is expected to go through the next condition for sign extension. @annagrin can confirm.

include/cudaq/Optimizer/CodeGen/Passes.td
khalatepradnya: I think the current wording is clear because single log file is expected to be generated for multiple shots as opposed the proposed rewording which might imply one log file per shot.

python/cudaq/runtime/run.py
khalatepradnya: I think the original wording with means that the `get` call returns the array not the `AsyncRunResult`.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: Fixed in 8f133dbd28c0c490af092df7e9781b84d75b90de

khalatepradnya:
> /ok to test 8f133dbd28c0c490af092df7e9781b84d75b90de

python/runtime/cudaq/algorithms/py_run_async.cpp
khalatepradnya: Captured in https://github.com/NVIDIA/cuda-quantum/issues/2984

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: Captured in https://github.com/NVIDIA/cuda-quantum/issues/2991

runtime/common/RecordLogParser.cpp
khalatepradnya: The `ARRAY` and `TUPLE` cases need to return early since they contain the container / aggregate type data and don't need further processing. The primitive types on the other hand can either be part of individual results or entries in aggregate type.

PR #2943, opened 2025-05-22
Fix publishing pipeline after dynamics migration to pybind11 binding

1tnguyen:
### Description

Follow up for https://github.com/NVIDIA/cuda-quantum/pull/2853

- Default integrator in case the optional integrator is not provided. This was an oversight when I re-wrote the `evolve_dynamics` body.

- Handle a constant (no operator) term in conversion. This was handled in the original Python operator conversion utility but not in the C++ implementation.

- Handle no-mpi environment for the new example.

- Remove a stale test case, as we have removed the unused `step_size` argument.


mitchdz (approved):
Thanks @1tnguyen

PR #2947, opened 2025-05-22
Fix binary output directory of dynamics bindings to be compatible with using CUDA-Q python from build directory

1tnguyen:
### Description

When invoking some Python checks hooked to cmake (ctest), we use the CUDA-Q Python package in build directory, not install directory. Hence, the path must be consistent.

Thanks, @khalatepradnya for spotting it!


khalatepradnya (approved):
Thanks, Thien for the quick fix!
I tested this patch locally, and it works as expected.

PR #2948, opened 2025-05-22
[python] Fix incorrect alignment of structure elements in python

annagrin:
### Description
Fix incorrect alignment of structure elements in python

#### Details

Set data layout attribute on module before processing arguments and return values in python (`packArgs` and `convertResults`)

### DIFF
--- a/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
+++ b/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
@@ -28,7 +28,11 @@
 #include "runtime/cudaq/algorithms/py_utils.h"
 #include "utils/OpaqueArguments.h"
 #include "utils/PyTypes.h"
+#include "llvm/MC/SubtargetFeature.h"
+#include "llvm/MC/TargetRegistry.h"
 #include "llvm/Support/Error.h"
+#include "llvm/Support/Host.h"
+#include "llvm/Target/TargetMachine.h"
 #include "mlir/Bindings/Python/PybindAdaptors.h"
 #include "mlir/CAPI/ExecutionEngine.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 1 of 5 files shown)


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -840,6 +891,7 @@ void bindAltLaunchKernel(py::module &mod) {
         auto kernelFunc = getKernelFuncOp(module, kernelName);
 
         cudaq::OpaqueArguments args;
+        setDataLayout(module);
```
khalatepradnya, line 894: If I understand this correctly, calling this setter here means the module is updated before it is saved into the `quakeRegistry`?

khalatepradnya: If so, may be some of the Python MLIR tests would fail?

annagrin: Yes, you understand it correctly. Good point about the tests, let's see if anything fails.

annagrin: Looks like everything passes except for one test, updated.

python/tests/mlir/exp_pauli.py
```diff
@@ -98,8 +98,8 @@ def kernel_noancilla_rotation(angles: list[float]):
 # CHECK:         %[[VAL_18:.*]] = load %Qubit*, %Qubit** %[[VAL_17]], align 8
 # CHECK:         tail call void @__quantum__qis__rx(double %[[VAL_16]], %Qubit* %[[VAL_18]])
 # CHECK:         %[[VAL_14]] = add nuw nsw i64 %[[VAL_13]], 1
-# CHECK:         %[[VAL_19:.*]] = icmp slt i64 %[[VAL_14]], %[[VAL_4]]
-# CHECK:         br i1 %[[VAL_19]], label %[[VAL_12]], label %[[VAL_9]]
+# CHECK:         %[[VAL_19:.*]] = icmp eq i64 %[[VAL_14]], %[[VAL_4]]
```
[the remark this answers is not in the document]
annagrin: I suspect this is coming from the main somehow. I'll double check.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -69,6 +73,53 @@ static std::unique_ptr<PyStateVectorStorage> stateStorage =
...
+  auto mod = unwrap(module);
+  if (!mod->hasAttr(opt::factory::targetDataLayoutAttrName)) {
+    auto dataLayout = createDataLayout();
+    mod->setAttr(opt::factory::targetDataLayoutAttrName,
+                 StringAttr::get(mod->getContext(), dataLayout));
+  }
+}
+
+/// @brief Create a new OpaqueArguments pointer and pack the
+/// python arguments in it. Clients must delete the memory.
```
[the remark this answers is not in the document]
annagrin: updated

PR #2951, opened 2025-05-23
Proper handling of `evolve_async` input state 

1tnguyen:
### Description

The input state to `evolve_async` may be constructed on a different device (GPU), as this state is built on the main thread before the async. functor is executed.

Hence, adding a helper function to detect and migrate the state data if needed.


1tnguyen:
> A question. Do we cover this case in tests for evolve_async?

Yes. In fact, it's how I detected this bug :) It came to light thanks to the refactor we did in `to_density_matrix`, which now operates directly on GPU data. 

The `DynamicsAsyncTester` should cover this case now with the refactored code. Previously, this was an issue, but the test couldn't detect it since the task assigned to QPU 1 requires `to_density_matrix`. This method was using device -> host transfer before, thus masking this issue.

PR #2970, opened 2025-05-28
Add logging macros to reduce logging overhead

bmhowe23:
Our logging functions currently perform argument and string evaluations even if logging at that level is disabled. One can manually avoid that overhead by calling `if (should_log(...)) do_the_log_call()`, and we even do that in some "hot spots" in the code. These macros will make that code (and any other downstream code that uses CUDA-Q, like CUDA-QX libraries) much less verbose.

While the new macros are a drop-in replacement for all cudaq::warn/info/debug messages, I elected to be a bit more surgical in this PR and only change a few specific spots in the CUDA-Q code for that. In the long term, we could migrate all the logging calls to the macros, but I don't want to unnecessarily complicate any refactoring efforts going on in parallel.

### DIFF
--- a/runtime/common/Logger.h
+++ b/runtime/common/Logger.h
@@ -290,7 +290,7 @@ public:
   } while (false)
 
 #ifdef CUDAQ_DEBUG
-#define CUDAQ_DBG(...)                                                       \
+#define CUDAQ_DBG(...)                                                         \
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
LGTM 💯 Thanks, @bmhowe23

PR #2977, opened 2025-05-30
[C++] Drop library mode support in remote simulators

khalatepradnya:
Note that the serialization logic in `KernelWrapper.h` is used to support VQE algorithm in remote simulator targets.


runtime/common/BaseRestRemoteClient.h
```diff
@@ -125,141 +125,120 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
     enablePrintMLIREachPass =
         getEnvBool("CUDAQ_MLIR_PRINT_EACH_PASS", enablePrintMLIREachPass);
 
-    if (cudaq::__internal__::isLibraryMode(name)) {
```
schweitzpgi, line 128: I just realized the renaming of these namespaces got derailed.

khalatepradnya: Yes! This PR deletes some of that code, so less refactoring later :)

Issue #710, opened 2023-09-28
Move larger code blocks in docs into their own files


bettinaheim:
Including code snippets directly as code blocks in rst files is fragile and there is a good change that the code in the docs will become incorrect over time. We hence should move code snippets into their own file and build and run that file as part of CI. Selective lines from these files can be included in the docs using [literalinclude](https://devopstutodoc.readthedocs.io/en/stable/documentation/doc_generators/sphinx/rest_sphinx/code/literalinclude/literalinclude.html). 
In particular the examples in the language specification section should not be hardcoded in the [rst file](https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/specification/cudaq/examples.rst).

ahkatlio:
![Image](https://github.com/user-attachments/assets/2b307a77-cc40-4bab-bf1a-ea0f047866be)
Hello may I know what this mean?
Thank you

cabreraam:
I created a PR for some of the refactor requested in this issue. I have more questions about some other `.rst` files that need to be refactored, but I'll ask them tomorrow.

I'm happy to not be the sole dev on this; if anyone else wants to take any of the other pieces, I'm happy to divide and conquer!

Death0004:
> I created a PR for some of the refactor requested in this issue. I have more questions about some other `.rst` files that need to be refactored, but I'll ask them tomorrow.
> 
> I'm happy to not be the sole dev on this; if anyone else wants to take any of the other pieces, I'm happy to divide and conquer!

Hey I'd be happy to help!! If you have a discord or smth please let me know, otherwise posting any progress updates here/overview/summary of your work would also be helpful!!

cabreraam:
@Death0004 sounds great! I actually don't have a ton of time to work on this week, but I'll provide you with the notes I had when scoping out what needs to get done and my approach. You have free reign; work on whatever you want! I'll follow your lead if I decide to pick this back up.

To make the PRs more modular -- as opposed to having a giant PR that does the complete refactoring -- I wanted to just make a PR per `.rst` file that got refactored. You can see this approach in my PR https://github.com/NVIDIA/cuda-quantum/pull/2981.

From this PR, you can see that
* [ ] A new `.cpp` and or `.py` file gets created that just takes the original code snippet from the `.rst` file and gives it its own file
* [ ] Replacing the hard-coded snippet in the `.rst` file with a pointer to the code files you created above
[...]

PR #2981, opened 2025-06-02
[docs] refactoring examples.rst with standalone CI integration

cabreraam:
### Description

This, in part, addresses #710. Refactoring `examples.rst`  is more straightforward than the other `.rst` files that need to be refactored.

> [!NOTE]
> In order to satisfy an overload error caused by using C++20 ([here](https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/qis/execution_manager.h#L59-L60)), I added a `static_cast` [here](https://github.com/cabreraam/cuda-quantum/blob/42bf2c2298138af4eecb088ab4e51371ed6bd507/docs/sphinx/applications/cpp/bell_state.cpp#L11) to explicitly expect an `int`. I don't know if this was the best approach, but I'm noting it here.


docs/sphinx/applications/cpp/bell_state.cpp
```diff
@@ -0,0 +1,21 @@
+#include <cudaq.h>
```
schweitzpgi, line 1: Copyright banner?

docs/sphinx/applications/cpp/bell_state.cpp
```diff
@@ -0,0 +1,21 @@
...
+        nCorrect++;
+
+      reset(q[0]);
+      reset(q[1]);
+    }
+    return nCorrect;
+  }
+};
+
+int main() { printf("N Correct = %d\n", bell{}(100)); }
```
schweitzpgi, line 21:
```suggestion
int main() { printf("N Correct = %d\n", bell{}(100));
return 0; }
```

docs/sphinx/applications/cpp/deuteron_binding_energy_param_sweep.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+  // Perform parameter sweep for deuteron N=2 Hamiltonian
+  const auto param_space = cudaq::linspace(-M_PI, M_PI, 25);
+  for (const auto &param : param_space) {
+    // KERNEL::observe(...) <==>
+    // E(params...) = <psi(params...) | H | psi(params...)>
+    double energy_at_param = cudaq::observe(deuteron_n2_ansatz{}, h, param);
+    printf("<H>(%lf) = %lf\n", param, energy_at_param);
+  }
+}
```
schweitzpgi, line 26:
```suggestion
return 0; }
```

docs/sphinx/applications/cpp/bell_state.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+struct bell {
+  int operator()(int num_iters) __qpu__ {
+    cudaq::qarray<2> q;
+    int nCorrect = 0;
+    for (int i = 0; i < num_iters; i++) {
+      h(q[0]);
+      x<cudaq::ctrl>(q[0], q[1]);
+      auto results = mz(q);
+      if (static_cast<int>(results[0]) == static_cast<int>(results[1]))
```
schweitzpgi, line 11: I suppose you could write this as

```suggestion
      if (results[0] && results[1] || !results[0] && !results[1])
```
if you want to get rid of the casts.

udsy19:
Thanks for picking this up — extracting these into runnable files is exactly
what makes this class of problem findable.

One thing worth catching while `deuteron_binding_energy_param_sweep.py` is
being extracted, since this PR also adds it to `docs/CMakeLists.txt` as a
CI-executed test: the sweep does not actually sweep. The loop variable `angle`
is bound but never used, and the literal `0.59` is passed to `observe` on every
iteration, so all 25 lines report the same energy while printing a different
label. The C++ file added alongside it does the right thing — it passes
`param`.

```python
for angle in np.linspace(-np.pi, np.pi, 25):
    energyAtParam = cudaq.observe(ansatz, hamiltonian, 0.59)   # <- 0.59, not angle
    print("<H>({}) = {}".format(angle, energyAtParam))
```

There is a second, smaller thing on the same two lines: `energyAtParam` is an
`ObserveResult`, so `format` renders the repr rather than a number.

Running the file exactly as it stands in this PR (first 5 of the 25 angles):

```
<H>(-3.141592653589793) = <cudaq...ObserveResult object at 0x10c559b30>
<H>(-1.5707963267948966) = <cudaq...ObserveResult object at 0x10c559bf0>
<H>(0.0) = <cudaq...ObserveResult object at 0x10c559a70>
<H>(1.5707963267948966) = <cudaq...ObserveResult object at 0x10c559b30>
<H>(3.141592653589793) = <cudaq...ObserveResult object at 0x10c559bf0>
```

With `angle` passed through and `.expectation()` called:

```
<H>(-3.141592653589793) = 12.250289999999994
<H>(-1.5707963267948966) = 10.193599999999996
<H>(0.0) = -0.4362899999999955
<H>(1.5707963267948966) = 1.6204000000000023
<H>(3.141592653589793) = 12.250289999999993
```

which is the energy curve the section is meant to show, and it matches what the
C++ file in this PR produces.

So the suggestion is just the two edits:

```suggestion
    energyAtParam = cudaq.observe(ansatz, hamiltonian, angle).expectation()
```

I have a separate documentation-corrections PR open, but I deliberately kept it
out of `examples.rst` so it does not conflict with yours — this one belongs
here.

(Verified against the `cudaq` 0.15.1 wheel on macOS arm64.)

PR #2985, opened 2025-06-02
Upgrading tornado to 6.5

sacpis:
Upgrading tornado to fix Tornado vulnerability causing excessive logging.
https://github.com/advisories/GHSA-7cx3-6m66-7c5m


mitchdz (approved):
LGTM, thanks @sacpis

PR #2988, opened 2025-06-03
Changing the base image for CUDAQ

sacpis:
Fetching the base image from nvcr as per the security requirement.


1tnguyen (approved):
LGTM 👍

Issue #865, opened 2023-11-06
Support individual qubits exclusion from circuit routing


kukushechkin:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

While IQM produces several predefined QPU architecture (e.g., Adonis, Apollo, and others in the future), the actual components availability in a concrete IQM stations can use only a subset of qubits because of, for example, manufacturing defects. This fact is represented in the station QPU description returned by the `/cocos/quantum-architecture`, for example this example, notice absence of the `QB2`, the full `Adonis` would have had:

```json
{
  "quantum_architecture": {
    "name": "Adonis Variant",
    "operations": [
      "phased_rx",
      "cz",
      "measurement",
      "barrier"
    ],
    "qubits": [
      "QB1",
      "QB3",
      "QB4",
      "QB5"
    ],
    "qubit_connectivity": [
      [
        "QB1",
        "QB3"
      ],
      [
        "QB3",
        "QB4"
      ],
      [
        "QB3",
        "QB5"
      ]
    ]
  }
}
```

CUDA Quantum addresses translate qubit by indices, making it impossible to use such an architecture with `nvq++` if the allocated quantum memory covers any of the excluded qubits.

There should be a way to exclude certain qubits from circuit routing. For example, it can be specified in the predefined IQM QPU architecture (currently there are `Adonis.txt` and `Apollo.txt`), so it is possible to create e.g., `Adonis Variant.txt`. 

For the current format of connectivity mapping it could look like this for an Apollo with `QB2`, `QB3` and `QB4` excluded:
```
# ============================================================================ #
# Copyright (c) 2022 - 2023 NVIDIA Corporation & Affiliates.                   #
# All rights reserved.                                                         #
#                                                                              #
# This source code and the accompanying materials are made available under     #
# the terms of the Apache License 2.0 which accompanies this distribution.     #
# ============================================================================ #

# The format of the file is as follows:
# 1. You must specify the number of nodes
Number of nodes: 20

# 2. Specifying the number of edges is optional
Number of edges: 60

# 3. For each node, list the connections as:
# <NodeNumber> --> { <ConnectedNode1>, <ConnectedNode2>, ... }
# Notes:
# - Node numbers are 0-based
# - Each node's info must be entirely on one line.
# - The nodes and lines do not have to be in order.
[...]

PR #2995, opened 2025-06-04
Allow excluding qubits in Device

nvidia-dobri:
Addresses https://github.com/NVIDIA/cuda-quantum/issues/865

The main changes in this PR are:

- Instead of creating an adjacency from a `Device` in `MappingPrep` just to create a `Device` from that adjacency in `MappingFunc`, both `MappingPrep` and `MappingFunc` now expect a device descriptor (ideally, the same one 🙂)
- `GraphCSR` now allows removing edges (but not nodes)
- `Device` can now keep track of excluded qubits
- The initially created `Placement` will assign to the first N usable qubits, not the first N physical qubits
- Any subsequent mapping uses `swap` to swap qubits, so no excluded qubits should be mapped accidentally. However, an additional check is added to `SabreRouter::mapOperation` to prevent foot-gunning in the future.

Remaining TODOs:
- [ ] Tests
- [x] Final linting / pretty pass

### DIFF
--- a/.github/workflows/config/gitlab_commits.txt
+++ b/.github/workflows/config/gitlab_commits.txt
@@ -1,2 +1,2 @@
 nvidia-mgpu-repo: cuda-quantum/cuquantum-mgpu.git
-nvidia-mgpu-commit: 86ea9c7c3f2ebc155514754fa2e816e642a2ad44
+nvidia-mgpu-commit: bfccb143f12b42be129ed2fbf16c39428eaba7b7
--- a/docs/sphinx/applications/cpp/random_walk_qpe.cpp
+++ b/docs/sphinx/applications/cpp/random_walk_qpe.cpp
@@ -8,7 +8,7 @@
 
 // Compile and run with:
 // ```
-// nvq++ random_walk_qpe.cpp -o qpe.x && ./qpe.x
+// nvq++ -fno-enable-cudaq-run random_walk_qpe.cpp -o qpe.x && ./qpe.x
 // ```
 
 #include <cudaq.h>
--- a/docs/sphinx/applications/python/adapt_qaoa.ipynb
+++ b/docs/sphinx/applications/python/adapt_qaoa.ipynb
@@ -28,7 +28,7 @@
     "Below is a schematic representation of the ADAPT-QAOA algorithm explained above.\n",
     "\n",
     "<div>\n",
-    "<img src=\"images/adapt-qaoa.png\" width=\"600\">\n",
+    "<img src=\"images/adapt-qaoa.png\" width=\"1000\">\n",
     "</div>\n",
     "\n",
     "\n"
@@ -36,7 +36,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 15,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -60,7 +60,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 12,
+   "execution_count": 16,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -103,7 +103,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": 17,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -158,7 +158,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 14,
+   "execution_count": 18,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -204,7 +204,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": 19,
    "metadata": {},
    "outputs": [
     {
@@ -231,7 +231,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": 20,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -252,7 +252,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": 21,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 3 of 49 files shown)


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -729,8 +729,7 @@ def quake_WireSetOp : QuakeOp<"wire_set", [IsolatedFromAbove, Symbol]> {
 
   let arguments = (ins
     StrAttr:$sym_name,
-    I32Attr:$cardinality,
-    OptionalAttr<ElementsAttr>:$adjacency
```
schweitzpgi: The adjacency set is a fundamental part of the semantics of a wire set. It cannot be dropped.

include/cudaq/Support/Device.h
```diff
@@ -75,6 +87,15 @@ class Device {
       }
     }
 
+    // We need to remove paths that include excluded qubits
+    for (auto excluded : device.excludedQubits) {
+      auto neighbors = device.topology.getNeighbours(excluded);
+      while (!neighbors.empty()) {
+        device.topology.removeEdge(excluded, neighbors.front());
+        neighbors = device.topology.getNeighbours(excluded);
```
sacpis: Instead of refetching updated neighbors after removing the edge, can we just move onto the next neighbor?
Something like

```
for (const auto &neighbor : neighbors) {
    device.topology.removeEdge(excluded, neighbor);
}
```

include/cudaq/Support/Device.h
```diff
@@ -214,6 +235,16 @@ class Device {
...
   unsigned getNumQubits() const { return topology.getNumNodes(); }
 
+  unsigned getNumUsableQubits() const {
+    return getNumQubits() - excludedQubits.size();
+  }
+
+  /// Returns true if the qubit is excluded from the device.
+  bool isQubitExcluded(Qubit qubit) const {
+    return std::find(excludedQubits.begin(), excludedQubits.end(), qubit) !=
+           excludedQubits.end();
```
sacpis: Just a question. Can we use ordered set for excludedQubits instead of a list, so that the lookup will be in constant time?

lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -44,9 +44,17 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
 // Placement
 //===----------------------------------------------------------------------===//
 
-void identityPlacement(Placement &placement) {
-  for (unsigned i = 0, end = placement.getNumVirtualQ(); i < end; ++i)
-    placement.map(Placement::VirtualQ(i), Placement::DeviceQ(i));
+bool identityPlacement(Placement &placement, Device &device) {
+  unsigned j = 0;
+  for (unsigned i = 0, end = placement.getNumVirtualQ(); i < end; ++i, ++j) {
```
sacpis: I think the letter `Q` in getNumVirtualQ stands for Qubits, right? And elsewhere in this function as well.
If so, should we mention it in the name? Like `getNumVirtualQubits`.

runtime/cudaq/platform/default/rest/helpers/iqm/Adonis-variant.txt
```diff
@@ -0,0 +1,29 @@
...
+# - Each node's info must be entirely on one line.
+# - The nodes and lines do not have to be in order.
+# - Connections are assumed to be bidirectional.
+# - Any connections to/from excluded nodes will be ignored.
+# - Any duplicates will be automatically removed.
+0 --> {2}
+1 --> {2}
+2 --> {0, 1, 3, 4}
+3 --> {2}
+4 --> {2}
```
sacpis, line 31: A question. Will there be a case where the graph will have disconnected nodes? Something like two groups of nodes where there is no connection between them.

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -769,13 +764,9 @@ void quake::WireSetOp::print(OpAsmPrinter &p) {
...
   p.printSymbolName(getSymName());
   p << '[' << getCardinality() << ']';
-  if (auto adj = getAdjacency()) {
-    p << " adjacency ";
-    p.printAttribute(*adj);
-  }
   p.printOptionalAttrDictWithKeyword(
       (*this)->getAttrs(),
-      {getSymNameAttrName(), getCardinalityAttrName(), getAdjacencyAttrName()});
+      {getSymNameAttrName(), getCardinalityAttrName()});
```
schweitzpgi: These changes should not be made as well.

nvidia-dobri:
> Let's not break wire_set semantics as a side effect here. A wire_set isn't used by the mapping algorithm, actually. It is intended to capture, in the IR, the hardware topology as a function of the target instead of as ad hoc information and casual assumptions in the mapping (collection) of pass(es).

OK. How should we capture excluded qubits in MLIR? I understand your point, but this approach seems very error-prone. We consume structure A (`Device`) to make structure B (`adjacency`) which we then use to convert back into structure A. I spent some time chasing my tail on why the `Device` in `MappingFunc` does not include my excluded qubits. We **can** just add excluded qubits as an attribute of wire_set, but the next property of device topology we want to rely on in mapping passes will run into the same issue, no?

schweitzpgi:
> > Let's not break wire_set semantics as a side effect here. A wire_set isn't used by the mapping algorithm, actually. It is intended to capture, in the IR, the hardware topology as a function of the target instead of as ad hoc information and casual assumptions in the mapping (collection) of pass(es).
> 
> OK. How should we capture excluded qubits in MLIR? I understand your point, but this approach seems very error-prone. We consume structure A (`Device`) to make structure B (`adjacency`) which we then use to convert back into structure A. I spent some time chasing my tail on why the `Device` in `MappingFunc` does not include my excluded qubits. We **can** just add excluded qubits as an attribute of wire_set, but the next property of device topology we want to rely on in mapping passes will run into the same issue, no?

The mapping passes do not use `wire_set`s, so it isn't part of the issue. There is no assumption that the ad hoc structures will generate the adjacency list in `wire_set`. It should be the other way around, if anything.

As to how "dropped qubits" would work with respect to a `wire_set`, the topology of the target machine can be expressed as a set of adjacencies. A "dead" qubit simply means that qubit isn't included in any set. A "broken connection" is easily modeled by removing specific edges in the adjacency graph. Again, as these would be captured in the IR and not in some ad hoc way in a pass, the network topology is constant for the device the code is being generated for with a single source of truth, available in the IR.

nvidia-dobri:
@schweitzpgi - would you like to take another look re: design while I add some tests for the functionality?

runtime/cudaq/platform/default/rest/helpers/iqm/Adonis-variant.txt
nvidia-dobri: I guess it's entirely possible. Technically, the mapping pass should be able to handle that case just fine.

include/cudaq/Support/Device.h
nvidia-dobri: Because of the underlying data structure, that's very difficult. But not impossible. I choose the more readable and less efficient version, which may break down for hundreds or thousands of "bad" qubits.

include/cudaq/Support/Device.h
nvidia-dobri: I think that's reasonable, I'll update in a bit.

runtime/cudaq/platform/default/rest/helpers/iqm/Adonis-variant.txt
boschmitt: The computation of shortest paths assumes that the graph is connected. Also, the mapping pass assumes that it can always apply an operation between two qubits, as long as it adds enough swaps operations.

I haven't looked closely at the PR, but there needs to be modifications on those pieces to be able to handle cases where two nodes (qubits) don't have a path between them. (For example, the mapping pass needs to understand the cost value of infinity between two qubits.)

Another complication added in such cases is that mapping will no longer guaranteed to succeed if the device has enough qubits. (As it stands today, if a device has enough qubits, mapping is only a question of how many swaps it needs to add.)

Issue #2966, opened 2025-05-27
[run, python] Updating dataclass fields fails to compile


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following code does not compile (see error messages inline):

```python

def test_modify_struct():

    @dataclass
    class MyClass:
        x: int
        y: bool

    @cudaq.kernel
    def simple_struct(t: MyClass) -> MyClass:
        q = cudaq.qubit()
        t.x = 42
        return t

    # invalid assignment detected.
    # (offending source -> t.x = 42)
    results = cudaq.run(simple_struct, MyClass(-13, True), shots_count=2)
    print(results)
    assert len(results) == 2
    assert results[0] == MyClass(42, True)
    assert results[1] == MyClass(42, True)

test_modify_struct()

def test_create_and_modify_struct():

    @dataclass
    class MyClass:
        x: int
        y: bool

    @cudaq.kernel
    def simple_struct() -> MyClass:
        q = cudaq.qubit()
        t = MyClass(-13, True)
        t.x = 42
        return t

    # invalid assignment detected.
    # (offending source -> t.x = 42)
    results = cudaq.run(simple_struct, shots_count=2)
    print(results)
    assert len(results) == 2
    assert results[0] == MyClass(42, True)
    assert results[1] == MyClass(42, True)

test_create_and_modify_struct()
```

### Steps to reproduce the bug

```python

def test_modify_struct():

    @dataclass
    class MyClass:
        x: int
        y: bool

    @cudaq.kernel
    def simple_struct(t: MyClass) -> MyClass:
        q = cudaq.qubit()
        t.x = 42
        return t

    # invalid assignment detected.
    # (offending source -> t.x = 42)
    results = cudaq.run(simple_struct, MyClass(-13, True), shots_count=2)
    print(results)
    assert len(results) == 2
    assert results[0] == MyClass(42, True)
    assert results[1] == MyClass(42, True)

[...]

PR #3000, opened 2025-06-05
[Python] Allow assignment on `dataclass` members

khalatepradnya:
Fix for issue https://github.com/NVIDIA/cuda-quantum/issues/2966

### DIFF
--- a/python/tests/kernel/test_run_kernel.py
+++ b/python/tests/kernel/test_run_kernel.py
@@ -935,6 +935,26 @@ def test_modify_struct():
     assert results[0] == MyClass(42, True)
     assert results[1] == MyClass(42, True)
 
+    @dataclass
+    class Foo:
+        x: bool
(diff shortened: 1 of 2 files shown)


python/tests/kernel/test_run_kernel.py
```diff
@@ -916,6 +916,47 @@ def simple(numQubits: int) -> int:
     assert 'Invalid number of arguments passed to run:1 expected 0' in repr(e)
 
 
+def test_modify_struct():
+
+    @dataclass
+    class MyClass:
```
annagrin, line 922: Could the bool field come first so we also test for the values stored in the return object with correct alignment?

khalatepradnya: Added in dab5cba40a43fe47ed154e48bee493966900e82c

PR #3001, opened 2025-06-05
[cudaq::run] Disallow std::vector<T> return type from entry-point ker…

schweitzpgi:
…nels when called via cudaq::run.

### DIFF
--- a/runtime/cudaq/algorithms/run.h
+++ b/runtime/cudaq/algorithms/run.h
@@ -97,7 +97,13 @@ void resultSpanToVectorViaOwnership(std::vector<T> &result,
 template <typename QuantumKernel, typename... ARGS>
 #if CUDAQ_USE_STD20
   requires(!std::is_void_v<std::invoke_result_t<std::decay_t<QuantumKernel>,
-                                                std::decay_t<ARGS>...>>)
+                                                std::decay_t<ARGS>...>> &&
+           !std::is_same_v<
+               std::remove_cvref_t<std::invoke_result_t<
+                   std::decay_t<QuantumKernel>, std::decay_t<ARGS>...>>,
+               std::vector<typename std::remove_cvref_t<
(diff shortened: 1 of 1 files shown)


runtime/cudaq/algorithms/run.h
```diff
@@ -97,7 +97,13 @@ void resultSpanToVectorViaOwnership(std::vector<T> &result,
 template <typename QuantumKernel, typename... ARGS>
 #if CUDAQ_USE_STD20
   requires(!std::is_void_v<std::invoke_result_t<std::decay_t<QuantumKernel>,
-                                                std::decay_t<ARGS>...>>)
+                                                std::decay_t<ARGS>...>> &&
+           !std::is_same_v<
```
khalatepradnya, line 101: Q: What about C++17?

runtime/cudaq/algorithms/run.h
```diff
@@ -97,7 +97,13 @@ void resultSpanToVectorViaOwnership(std::vector<T> &result,
...
+               std::remove_cvref_t<std::invoke_result_t<
+                   std::decay_t<QuantumKernel>, std::decay_t<ARGS>...>>,
+               std::vector<typename std::remove_cvref_t<
+                   std::invoke_result_t<std::decay_t<QuantumKernel>,
+                                        std::decay_t<ARGS>...>>::value_type>>)
```
khalatepradnya, line 106: Can we have a test showcasing the expected failure?

runtime/cudaq/algorithms/run.h
schweitzpgi: Support for C++17 is now deprecated but not fully eliminated.

runtime/cudaq/algorithms/run.h
schweitzpgi: There was a test, but I eliminated the failure so it would pass. :)

Issue #2965, opened 2025-05-27
[run,python] Updating a tuple passed as argument in kernels does not work


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following code does not update the tuple element:

```python
def test_modify_tuple():

    @cudaq.kernel
    def simple_tuple(t: tuple[int, bool]) -> tuple[int, bool]:
        q = cudaq.qubit()
        t[0] = 15
        return t

    results = cudaq.run(simple_tuple, (13, True), shots_count=2)
    print(results)
    assert len(results) == 2
    assert results[0] == (15, True)
    assert results[1] == (15, True)

test_modify_tuple()
```

### Steps to reproduce the bug

```python
def test_modify_tuple():

    @cudaq.kernel
    def simple_tuple(t: tuple[int, bool]) -> tuple[int, bool]:
        q = cudaq.qubit()
        t[0] = 15
        return t

    results = cudaq.run(simple_tuple, (13, True), shots_count=2)
    print(results)
    assert len(results) == 2
    assert results[0] == (15, True)
    assert results[1] == (15, True)

test_modify_tuple()
```

### Expected behavior

Test passes

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

sacpis:
Isn't tuple an immutable data structure?

sacpis:
In python3, I do get TypeError: 'tuple' object does not support item assignment. Which is expected. 
Wondering why are we changing this behavior in CUDA-Q. Would you please point me to the spec?

PR #3002, opened 2025-06-05
[Python] Throw error on tuple item assignment in a kernel

khalatepradnya:
Addresses https://github.com/NVIDIA/cuda-quantum/issues/2965

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1209,6 +1209,17 @@ class PyASTBridge(ast.NodeVisitor):
             elif isinstance(node.targets[0], ast.Subscript) and isinstance(
                     node.targets[0].value,
                     ast.Name) and node.targets[0].value.id in self.symbolTable:
+
+                # Tuples are immutable, so we cannot assign to items in them
+                # Get the variable from the symbol table and check if it's a tuple
+                varType = self.symbolTable[node.targets[0].value.id].type
+                if cc.PointerType.isinstance(varType):
+                    varType = cc.PointerType.getElementType(varType)
+                if cc.StructType.isinstance(varType) and cc.StructType.getName(
+                        varType) == "tuple":
+                    self.emitFatalError(
+                        "'tuple' object does not support item assignment", node)
+
                 # Visit_Subscript will try to load any pointer and return it
                 # but here we want the pointer, so flip that flag
                 # FIXME: move loading from Visit_Subscript to the user instead.
--- a/python/tests/kernel/test_direct_call_return_kernel.py
+++ b/python/tests/kernel/test_direct_call_return_kernel.py
@@ -382,10 +382,9 @@ def test_return_tuple_int_float():
         t[1] = 11.5
         return t
 
-    # TODO: Fix incorrect IR generation for tuple element assignment
-    # https://github.com/NVIDIA/cuda-quantum/issues/2965
-    result = simple_tuple_int_float_assign(2, (-13, 42.3))
-    # assert result == (-14, 11.5)
+    with pytest.raises(RuntimeError) as e:
+        simple_tuple_int_float_assign(2, (-13, 42.3))
+    assert "'tuple' object does not support item assignment" in str(e.value)
 
 
 def test_return_tuple_float_int():
--- a/python/tests/kernel/test_run_kernel.py
+++ b/python/tests/kernel/test_run_kernel.py
@@ -560,18 +560,28 @@ def test_return_tuple_int_float():
     assert results[1] == (13, 42.3)
 
     @cudaq.kernel
-    def simple_tuple_int_float_assign(
+    def simple_tuple_int_float_update(
             n: int, t: tuple[int, float]) -> tuple[int, float]:
         qubits = cudaq.qvector(n)
         t[0] = -14
         t[1] = 11.5
         return t
 
-    # TODO: Fix incorrect IR generation for tuple element assignment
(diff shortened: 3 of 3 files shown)


sacpis:
Just a question. In which case will we need to pass an immutable object to the kernel? We can have the return type an tuple.

khalatepradnya:
> Just a question. In which case will we need to pass an immutable object to the kernel? We can have the return type an tuple.

I don't quite understand the Q. But, yes, tuple return type is allowed as shown by test here - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_run_kernel.py#L543

sacpis:
> > Just a question. In which case will we need to pass an immutable object to the kernel? We can have the return type an tuple.
> 
> I don't quite understand the Q. But, yes, tuple return type is allowed as shown by test here - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_run_kernel.py#L543

I mean why do we need to pass a tuple to the kernel?

khalatepradnya:
> > > Just a question. In which case will we need to pass an immutable object to the kernel? We can have the return type an tuple.
> > 
> > 
> > I don't quite understand the Q. But, yes, tuple return type is allowed as shown by test here - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_run_kernel.py#L543
> 
> I mean why do we need to pass a tuple to the kernel?

It depends on the user / application. Any reason to not allow it?

sacpis:
> > > > Just a question. In which case will we need to pass an immutable object to the kernel? We can have the return type an tuple.
> > > 
> > > 
> > > I don't quite understand the Q. But, yes, tuple return type is allowed as shown by test here - https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/kernel/test_run_kernel.py#L543
> > 
> > 
> > I mean why do we need to pass a tuple to the kernel?
> 
> It depends on the user / application. Any reason to not allow it?

Rather I am searching for the use cases of it.

boschmitt:
The current specification doesn't address the behavior of tuples in CUDA-Q. There is a need for a discussion about it. 

We already deviate from Python behavior is various places, thus the reasoning that this is how tuples work in Python is not enough of a justification. Note that the concept of immutability in python is quite nuanced:

```python
t1 = ([], 1)
t1[0].append(2)
print(t1) # prints: ([2], 1)

t1[1] = 2 # error
```

In the case of tuples, they are considered immutable because the tuple's structure (the references it holds) cannot be modified after creation, even if the objects it contains are mutable. This reasoning doesn't apply in CUDA-Q, we don't follow Python's "object reference" semantics, which has behavior determined by object's mutability.

The way CUDA-Q works and enforces types should already guarantee the immutability of tuples' structure. However, assignment behaves completely different. In Python `1` and `2` are different immutable objects, thus when one does:

```
var = 1
var = 2
```

The second assignment is not changing the state of a integer object from `1` to `2` (Which is somewhat how CUDA-Q behaves). It is making `var` reference a new integer object that holds `2`. Thus, in the previous example, `t1[1] = 2` is trying to change what object the tuple holds, something disallowed in python. (However, note that changing the object held by the tuple is ok, e.g., `t1[0].append(2)`.)

This is all to say that, in CUDA-Q, tuple item assignment could (and perhaps should) be legal. CUDA-Q is _not_ Python, and it already deviates from Python in non-trivial ways.

sacpis:
"This is all to say that, in CUDA-Q, tuple item assignment could (and perhaps should) be legal. CUDA-Q is not Python, and it already deviates from Python in non-trivial ways."

Wouldn't it create confusion among users when using Python to code up applications using CUDA-Q? Or rather do they have to learn new data types behavior to use CUDA-Q? 
Also, if we are ok to modify the behavior (of tuple in this case), then wondering why do we need tuple in the first place? Why can we just use list? 

"deviates from Python in non-trivial ways"

Would you please point me where there is a deviation in CUDA-Q? Is it visible to the end user?

I am not arguing here. Just trying to understand the thought process behind the inclusion of a data type (existing in a programming language) in CUDA-Q.

sacpis:
Or can we introduce a new data type (something like qtuple) which wraps tuple in Python. That way we can define customized behavior.

sacpis:
>CUDA-Q is not Python, it is a language that uses Python syntax but deviates from it semantics. This must be made clear. Users >need to learn about the differences and the tooling should aim at easing the pain points, e.g., better error messages and >explanations.

Thanks @boschmitt for the clarification. Hope everyone in the team is on the same page. Or we need to make sure we have a consensus on this. Yes, as rightly said, better error messages or even better would be to have explanations in the user docs.

>Because we also modified lists semantics. In CUDA-Q, list must be homogeneous, i.e., list cannot have items of different types-->-this is what tuples are for.

In this case, I think we shouldn't be using the same data types names as the programming language. I think then it should qlist or qtuple, like we have qvector.
If we do have these new data types for CUDA-Q, then we need to make sure that the purpose for introducing these types should be clearly mentioned or agreed upon.

>On top of not accepting lists with elements of different types, CUDA-Q's lambdas behave in a complete different way. Variables' >scope is different. Variables have immutable types, i.e., if you do foo = 1, then foo must have type int and assigning a value of >another type is an error, say foo = [1] (although this might work in some cases where coercion kicks-in) . Assignments are all >over the place, although some are likely to be bugs or temporary constraints. The lists goes on.. (but not a CUDA-Q list, because >those cannot be appended).

Seems like we have too many variations atleast using Python in CUDA-Q then.

khalatepradnya:
Superseded by https://github.com/NVIDIA/cuda-quantum/pull/3032

PR #3004, opened 2025-06-06
[Bug fix] Python `observe`: missing `unset_noise` on a return code path

1tnguyen:
### Description

This seems to be a bug since we refactored Python to support kernel mode.

We have multiple `return` statements, and one of them missed the noise model cleanup, causing the noise model to 'leak' into the global context. 

Also, adding a test case.


khalatepradnya (approved):
LGTM. Thanks, Thien!

PR #3005, opened 2025-06-06
Edit `CUDAQ_GLOBAL_INDEX_BITS` doc

1tnguyen:
### Description

Edit the documentation for `CUDAQ_GLOBAL_INDEX_BITS` to improve clarity.


ghanem-nv:
Please correct me if I'm wrong but my understanding is that this variable applies to the communication network between all MPI processes regardless of whether they are intra-node or inter-node. For example, the given list would also apply to 4 nodes and 8 GPUs/node, with NVLinks between the GPUs. The current formulation makes the impression it only applies to the network between nodes and the that the network inside a node is not relevant. Here is a suggestion that reflects my understanding:

> Specify the network structure (faster to slower). For example, assuming a 32 MPI processes simulation, whereby the network topology is divided into 4 groups of 8 processes, which have faster communication network between them. In this case, the `CUDAQ_GLOBAL_INDEX_BITS` environment variable can be set to `3,2`. The first `3` (`log2(8)`) represents **8** processes with fast communication within the group and the second `2` represents the **4** groups (8 processes each) in those total 32 processes. The sum of all elements in this list is `5`, corresponding to the total number of MPI processes (`2^5 = 32`). Default is an empty list (no customization based on network structure of the cluster).

1tnguyen:
> Please correct me if I'm wrong but my understanding is that this variable applies to the communication network between all MPI processes regardless of whether they are intra-node or inter-node. For example, the given list would also apply to 4 nodes and 8 GPUs/node, with NVLinks between the GPUs. The current formulation makes the impression it only applies to the network between nodes and the that the network inside a node is not relevant. Here is a suggestion that reflects my understanding:
> 
> > Specify the network structure (faster to slower). For example, assuming a 32 MPI processes simulation, whereby the network topology is divided into 4 groups of 8 processes, which have faster communication network between them. In this case, the `CUDAQ_GLOBAL_INDEX_BITS` environment variable can be set to `3,2`. The first `3` (`log2(8)`) represents **8** processes with fast communication within the group and the second `2` represents the **4** groups (8 processes each) in those total 32 processes. The sum of all elements in this list is `5`, corresponding to the total number of MPI processes (`2^5 = 32`). Default is an empty list (no customization based on network structure of the cluster).

That makes sense. I've incorporated your suggestions in https://github.com/NVIDIA/cuda-quantum/pull/3005/commits/26e38013c9fd3a9e0934a6627391a98048b8c916

PR #3007, opened 2025-06-06
[py] Remove wrong verbose prints in the bridge

boschmitt:
### Description
PR #2956 change the way to print verbose messages. With multiple PRs changing the bridge around the same time, these statements ended up left behind.


khalatepradnya (approved):
LGTM. Thanks, Bruno!

PR #3009, opened 2025-06-06
Fix failing to get a python kernel function pointer

annagrin:
### Description
Fix failing to get a python kernel function pointer

Cudaqx code uses `jitAndGetFunctionPointer` that passes a function name with `__nvqpp__mlirgen` prefix to `jitAndCreateArgs`, which in turn that expects the name without the suffix (the name was not used in case of getting a function pointer before `cudaq.run` implementation added a function lookup in the module).

For the purposes of getting the pointer we don't need the code creating the args from `jitAndCreateArgs`, just jit code is enough.

Moved out the jit code into `jitKernel`, updated `jitAndGetFunctionPointer` to use `jitKernel` instead.

### DIFF
--- a/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
+++ b/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
@@ -121,14 +121,12 @@ OpaqueArguments *toOpaqueArgs(py::args &args, MlirModule mod,
   return argData;
 }
 
-std::tuple<ExecutionEngine *, void *, std::size_t, std::int32_t>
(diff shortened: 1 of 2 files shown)


bmhowe23:
Thanks, @annagrin. I will run the CUDA-QX testing on this change here:
* https://github.com/NVIDIA/cudaqx/pull/180
* https://github.com/NVIDIA/cudaqx/actions/runs/15495586487?pr=180

bmhowe23 (approved):
Thanks, Anna 👍

PR #3011, opened 2025-06-06
Run python interop test in CI

annagrin:
### Description
Running python interop test in CI.

This prevents issues like https://github.com/NVIDIA/cuda-quantum/pull/3009 from breaking integration with cudaqx.


mitchdz (approved):
LGTM. Thanks @annagrin

.github/workflows/test_in_devenv.yml
```diff
@@ -223,7 +223,8 @@ jobs:
 
           DOCKER_BUILDKIT=1 docker build --platform ${{ inputs.platform }} \
             -t dev_env:local -f docker/build/cudaq.dev.Dockerfile . \
-            --build-arg base_image=$base_image 
+            --build-arg base_image=$base_image \
+            --build-arg install="CMAKE_BUILD_TYPE=Debug"
```
bmhowe23: I thought we had special tests that ran in Debug. Wouldn't this change make all the Python tests doing Debug instead of Release?

mitchdz: Yes it would. The problem is the python test needs the C built shared objects in the `build/` directory. This build directory only exists when doing a Debug build, which is currently how the C tests are ran.

Where are the special debug tests, and do they exist for Python? That would be a better way to fix this :)

bmhowe23: The Debug tests may only exist for C++ (https://github.com/NVIDIA/cuda-quantum/blob/b89969d4c2a258a01fa3cd0ba37302eccdec141d/.github/workflows/test_in_devenv.yml#L29).

On second thought, I suspect the tests are still mostly occurring in Release mode because this change isn't affecting this line (https://github.com/NVIDIA/cuda-quantum/blob/b89969d4c2a258a01fa3cd0ba37302eccdec141d/.github/workflows/test_in_devenv.yml#L239).

I guess the only downside is that it does a full Debug build and only uses a couple files from it because I'm thinking all the other files would be ignored.

annagrin: I can try moving the interop test to the c++ tests (` Dev environment (Debug)`) ? I would need to install python things there of course and I am not sure if I am missing some cases in the matrix that way. WDYT?

annagrin: Alternatively I can run only this test in debug mode (in another container)

annagrin: Moved to the c++ environment since it already builds the container in debug mode. This adds about 8 minutes to the debug mode tests.

.github/workflows/test_in_devenv.yml
```diff
@@ -140,6 +140,25 @@ jobs:
...
             fi
 
+      - name: Test CUDA Quantum interop (Python)
+        uses: ./.github/actions/run-in-docker
+        with:
+          image: cuda-quantum-dev:local
+          shell: bash
+          run: |
+            cd $CUDAQ_REPO_ROOT
+            pip install . --user -vvv
```
bmhowe23: Are you sure this `pip install` step is truly needed? I see that is still taking an additional ~9 minutes in the CI.

I am wondering if it is possible to just add the `python3 -m pytest -v build/python/tests/interop/` to right after this line:
https://github.com/NVIDIA/cuda-quantum/blob/4bc957a1ef4f1445bfebc491009b317976a1e266/.github/workflows/test_in_devenv.yml#L98

That seemed to work for me when I did a standalone test on Gorby. No `pip install` was needed. (And therefore this new GitHub action step shouldn't be needed, I think.)

annagrin: I am new to our CI files, so I am not sure if it is needed. I will try without it.

bmhowe23 (approved):
It looks like the test is running now with minimal additional CI, so LGTM. Thanks for noticing and fixing this, @annagrin.

.github/workflows/test_in_devenv.yml
annagrin: That worked out well, the interop test adds a few seconds to the debug workflow and does not require a `pip install . --user -vvv` command.

.github/workflows/test_in_devenv.yml
annagrin: Fixed by running before the c++ tests in debug environment without `pip install . -user -vvv` command.

mitchdz (approved):
The changes look good, thanks @annagrin

PR #3014, opened 2025-06-06
Making fmt private

### LINKED ISSUE #2997: Can't build with CUDA-Q when project already uses a version of `fmt`
### Required prerequisites

### Describe the bug

This is sort of related to https://github.com/NVIDIA/cuda-quantum/issues/1704. When building my project which uses CUDA-Q, I get a linking error because the version of fmt lib used in CUDA-Q is far older (3 years old commit) than the version used in my project. I don't think the version of fmt used in a downstream project should be constrained by the version used by CUDA-Q.

1. Create a cmake project that uses fmt lib and CUDA-Q
2. Use a recent version of fmt lib in the cmake project (e.g. 11.1.4)
3. Create a hello-world source file which uses:
   1. `fmt::print()` and `fmt::join()`
   2. some cuda-q functions that also use fmt
4. Try to build

You will get an error saying there are ambiguous definitions of fmt functions. I.e. cuda-q has linked against the older version of fmt and the example project has linked against the newer version.

### Expected behavior

It builds.

Not a regression

### Environment

- **CUDA-Q version**: latest (0.11.0)
- **Python version**: n/a
- **C++ compiler**: g++
- **Operating system**:  ubuntu 22.04

### Suggestions

Either

1. upgrade the version of fmt in cuda-q
2. allow me to specify the version of fmt i want to use with the build

sacpis:
Making fmt private.

Fixes #2997.


sacpis:
> Thanks for taking a look at this Sachin. Before merging, could you please check what CUDA-QX changes would be needed for this change?

Sure.

amccaskey:
The big issue here is that `Logger.h` is a common header for downstream libraries (so they can get `cudaq::info()`) and it has `FmtCore.h` as an include. The changes here do not change that. The real fix is to remove any fmtlib headers from our own headers and hide them in cpp implementation code. I'm not sure how one would do that and keep `cudaq::info()` variadic though.

boschmitt:
> Thanks for taking a look at this Sachin. Before merging, could you please check what CUDA-QX changes would be needed for this change?

CUDA-QX have a [couple of files](https://github.com/search?q=repo%3ANVIDIA%2Fcudaqx%20fmt&type=code) that depends on `fmt` that comes from CUDA-Q. Thus we would either make those files not depend on `fmt` or make CUDA-QX have its own private `fmt`.

sacpis:
> The big issue here is that `Logger.h` is a common header for downstream libraries (so they can get `cudaq::info()`) and it has `FmtCore.h` as an include. The changes here do not change that. The real fix is to remove any fmtlib headers from our own headers and hide them in cpp implementation code. I'm not sure how one would do that and keep `cudaq::info()` variadic though.

Thanks @amccaskey. I have a question about another option. Why shouldn't we upgrade fmt to the latest version (11.2.0)?

george-pippos-qb:
Thank you all for taking a look at this for me after I raised [this issue](https://github.com/NVIDIA/cuda-quantum/issues/2997). I am trying to plan some work for our company's project that uses CUDA-Q and need to know whether this is likely to be in the next release (and a rough ETA of when the next release will be). I will need to look at alternatives for working around this problem in our repo if timelines don't line up so I would greatly appreciate an update.

bmhowe23:
> Thank you all for taking a look at this for me after I raised [this issue](https://github.com/NVIDIA/cuda-quantum/issues/2997). I am trying to plan some work for our company's project that uses CUDA-Q and need to know whether this is likely to be in the next release (and a rough ETA of when the next release will be). I will need to look at alternatives for working around this problem in our repo if timelines don't line up so I would greatly appreciate an update.

It sounds like our primary near-term option is to simply update the `fmt` library. Do you agree @sacpis?

sacpis:
> > Thank you all for taking a look at this for me after I raised [this issue](https://github.com/NVIDIA/cuda-quantum/issues/2997). I am trying to plan some work for our company's project that uses CUDA-Q and need to know whether this is likely to be in the next release (and a rough ETA of when the next release will be). I will need to look at alternatives for working around this problem in our repo if timelines don't line up so I would greatly appreciate an update.
> 
> It sounds like our primary near-term option is to simply update the `fmt` library. Do you agree @sacpis?

Yes, I do. Will put up a PR with updated `fmt`.

sacpis:
@george-pippos-qb @bmhowe23 Just FYI, we have updated fmt.

sacpis:
This has been addressed in #3764 and #3770.

PR #3019, opened 2025-06-09
Fixing an error in the operator documentation

### LINKED ISSUE #2939: Potential Error in Documentation
docs/sphinx/examples/python/operators.ipynb
https://nvidia.github.io/cuda-quantum/latest/examples/python/operators.html

In the section describing how to use exponentials of pauli strings, I believe there is an error. [If I am mistaken, please explain why.]

As far as I can tell, this is actually going to create the operator $e^{i(0.432XZY)} e^{i(0.324IXX)}$ and not $e^{i(0.432XZY +0.324IXX)}$.

Could someone let me know if this is the case, and fix it if so? These should be different as the pauli operators do not commute. 

If this is the case, how would one actually get the exponential of the total?

sacpis:
Fixing the operation applied.

Fixes #2939.

### DIFF
--- a/docs/sphinx/examples/python/dynamics/ion_trap.py
+++ b/docs/sphinx/examples/python/dynamics/ion_trap.py
@@ -1,0 +1,144 @@
+"""
+Tutorial: Creating a GHZ State with Trapped Ions
+
+This example shows how to prepare a GHZ (`Greenberger-Horne-Zeilinger`) state 
+using trapped ions and CUDA-Q. We'll use the effective spin model from the 
+famous `Sørensen-Mølmer` paper.
+
+What we're doing:
+- Start with N ions all in ground state `|gg...g⟩`
+- Apply an effective Hamiltonian H = 4χJ_x²  
+- Watch the system evolve into a GHZ superposition: `(|gg...g⟩ + |ee...e⟩)/√2`
+
+This tutorial uses the simplified effective model (`Eq. 2` from the paper).
+
+Reference:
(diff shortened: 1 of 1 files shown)


docs/sphinx/examples/python/operators.ipynb
```diff
@@ -52,7 +52,7 @@
     "\n",
     "The `pauli_word` type specifies a string of Pauli operations (e.g. ‘XYXZ’) and is convenient for applying operations based on exponentiated Pauli words.  The code below demonstrates how a list of Pauli words, along with their coefficients, are provided as kernel inputs and converted into operators by the `exp_pauli` function.\n",
     "\n",
-    "The cell below applies the following operation: $e^{i(0.432XZY +0.324IXX)}$"
+    "The cell below applies the following operation: $e^{i(0.432XYZ)}.e^{i(0.324IXX)}$"
```
1tnguyen: I think we need to swap the order: with the list `words = ['XYZ', 'IXX']`, `XYZ` will be applied first, aka, `e^{i(0.432XYZ)}` will hit the state vector first.

sacpis: oh, like this one?

`$e^{i(0.324IXX)}.e^{i(0.432XYZ)}$`

1tnguyen: yea, that's right

sacpis: Done in [67dbf85](https://github.com/NVIDIA/cuda-quantum/pull/3019/commits/67dbf85060cdada74e78fcfbed3eb5ff42d5e84b).

1tnguyen (approved):
:+1:

PR #3020, opened 2025-06-09
[core] Improve handling of cc.cast of complex values.

schweitzpgi:
This adds some new tests and lines to the verifier to test complex casts and disallow invalid forms of the complex casting.


lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -263,6 +263,17 @@ OpFoldResult cudaq::cc::CastOp::fold(FoldAdaptor adaptor) {
...
+        float f = val.convertToDouble();
+        APFloat fval(f);
+        return builder.create<arith::ConstantFloatOp>(loc, fval, fltTy)
+            .getResult();
+      }
+      if (ty == dblTy) {
+        APFloat fval{val.convertToDouble()};
+        return builder.create<arith::ConstantFloatOp>(loc, fval, dblTy)
+            .getResult();
+      }
```
sacpis, line 281: Seems like similar code in if loops. Should we write like this?

```
if (ty == fltTy || ty == dblTy) {
    APFloat fval{val.convertToDouble()};
    return builder.create<arith::ConstantFloatOp>(loc, fval, ty)
            .getResult();
}
```

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -303,7 +305,8 @@ OpFoldResult cudaq::cc::CastOp::fold(FoldAdaptor adaptor) {
           auto imPart = builder.getFloatAttr(eleTy, APFloat{imVal});
           auto cv = builder.getArrayAttr({rePart, imPart});
           return builder.create<complex::ConstantOp>(loc, ty, cv).getResult();
-        } else if (eleTy == dblTy) {
+        }
+        if (eleTy == dblTy) {
```
sacpis, line 314: Wondering why we changed else if to if?

schweitzpgi: There were some `else` after `return` cases here, which is against our coding conventions.

lib/Optimizer/Dialect/CC/CCOps.cpp
schweitzpgi: The two cases are different and must be different for correctness. One is setting the APFloat from a `float` type and the other is from a `double`.

sacpis: Thanks. Missed the curly brace.

lib/Optimizer/Dialect/CC/CCOps.cpp
sacpis: Good to know the coding convention here.

Issue #2893, opened 2025-05-09
ctrl.z using list indexing worked in 0.9.1 but not in 0.10.0


mmvandieren:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The code below works in 0.9 but no longer does in 0.10.1  The problem seems to be when the ctrl.z is inside a for loop in a kernel. For some reason it works fine until an x operation is added afterwards.   (see working and minimal code examples below)

### Steps to reproduce the bug

```num_qubits = 4    
num_iterations = 1 

# Apply the Grover diffusion operation to the equal superposition state
@cudaq.kernel
def grovers(num_qubits: int, num_iterations : int):
    qubits = cudaq.qvector(num_qubits)

    for _ in range(num_iterations):
        z.ctrl(qubits[0:-1], qubits[-1])
        x(qubits)
    
# Sample
sample_result = cudaq.sample(grovers, num_qubits, num_iterations, shots_count = 5000)
print(sample_result)
```

This results in the error 
```The Kernel crashed while executing code in the current cell or a previous cell. 
Please review the code in the cell(s) to identify a possible cause of the failure. 
Click [here](https://aka.ms/vscodeJupyterKernelCrash) for more info. 
View Jupyter [log](command:jupyter.viewOutput) for further details.```

### Expected behavior

The code should run. 

Removing the `for loop` fixes the problem.  Also removing the `x(qubits)` removes the problem. ¯\_(ツ)_/¯

The code runs just fine in version 0.9.1 and only causes the problem 0.10.1

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 0.10.0
- **Python version**: 3.10.12
- **C++ compiler**: 
- **Operating system**:  Mac


### Suggestions

@sacpis for viz

mmvandieren:
[...]

PR #3022, opened 2025-06-09
Handling negative lower and upper bound of a slice

### LINKED ISSUE #2930: Negative slice does not work
### Required prerequisites

### Describe the bug

Using negative values in slices in Python allows slicing from the tail, but this feature is no longer working.

```python
import cudaq


@cudaq.kernel
def bar():
    qubits = cudaq.qvector(10)
    h(qubits[0:3])
    controls = qubits[0:-1]
    target = qubits[-1]

    x.ctrl(controls, target)


print(cudaq.draw(bar))
print(cudaq.sample(bar))
```

Image

### Expected behavior

Outputs the correct circuit diagram and sampling results.

CUDA-Q 0.8.0 works

### Environment

- **CUDA-Q version**:  0.10.0
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

I've used the container in NGC.

### Suggestions

_No response_

sacpis:
Handling negative lower and upper bound of a slice

Fixes #2930 and #2893 

```
import cudaq

@cudaq.kernel
def bar():
    qubits = cudaq.qvector(10)
    h(qubits[0:3])
    controls = qubits[0:-1]
    target = qubits[-1]

    x.ctrl(controls, target)

print(cudaq.draw(bar))
print(cudaq.sample(bar))
```

# Output
```
root@f672ccb568e7:/workspaces/cuda-quantum# python3 test.py 
     ╭───╮     
q0 : ┤ h ├──●──
     ├───┤  │  
q1 : ┤ h ├──●──
     ├───┤  │  
q2 : ┤ h ├──●──
     ╰───╯  │  
q3 : ───────●──
            │  
q4 : ───────●──
            │  
q5 : ───────●──
            │  
q6 : ───────●──
            │  
q7 : ───────●──
            │  
q8 : ───────●──
          ╭─┴─╮
q9 : ─────┤ x ├
          ╰───╯

{ 0000000000:120 0010000000:130 0100000000:129 0110000000:111 1000000000:151 1010000000:118 1100000000:131 1110000000:110 }
```

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -2238,6 +2238,39 @@ def test_in_comparator():
     assert len(c) == 1 and '0' in c
 
 
+def test_negative_indices_for_a_slice():
+
+    @cudaq.kernel
+    def kernel():
+        qubits = cudaq.qvector(6)
+        h(qubits[0:3])
+        controls = qubits[0:-1]
+        target = qubits[-1]
+
+        x.ctrl(controls, target)
+
+    # test here is that it compiles and runs
+    cudaq.sample(kernel)
+    circuit = cudaq.draw(kernel)
+    print(circuit)
+    expected_str = '''     ╭───╮     
+q0 : ┤ h ├──●──
+     ├───┤  │  
+q1 : ┤ h ├──●──
+     ├───┤  │  
+q2 : ┤ h ├──●──
+     ╰───╯  │  
+q3 : ───────●──
+            │  
+q4 : ───────●──
+          ╭─┴─╮
+q5 : ─────┤ x ├
+          ╰───╯
+'''
+
+    assert circuit == expected_str
+
+
 # leave for gdb debugging
 if __name__ == "__main__":
     loc = os.path.abspath(__file__)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3202,18 +3221,15 @@ def visit_Subscript(self, node):
                 lowerVal = self.getConstantInt(0)
             if node.slice.upper is not None:
                 self.visit(node.slice.upper)
-                upperVal = self.popValue()
+                upperVal = fix_negative(self.popValue())
```
schweitzpgi, line 3224: In the description, you have an example of a negative value in the first position, but this appears to only be adjusting for the second position of the triple. We should probably handle negative start values.

Also do we handle step values and negative step values?

What about the `slice()` generator object?

Here is a pretty good description of Python slicing for posterity. https://stackoverflow.com/questions/509211/how-slicing-in-python-works

sacpis: >In the description, you have an example of a negative value in the first position

That is just accessing the list item at negative index and it works. This issue is related only to slicing (:).

> We should probably handle negative start values.

Yes, will add a fix for negative start value.

>Also do we handle step values and negative step values?
>What about the slice() generator object?

The issue is related to the given example having negative indices while slicing the list. Should we create a separate issue to handle step value (+ve and -ve) and slice() generator object?

>Here is a pretty good description of Python slicing for posterity. https://stackoverflow.com/questions/509211/how->slicing-in-python-works

Thank you.

sacpis: Handled negative start index in [0819e5f](https://github.com/NVIDIA/cuda-quantum/pull/3022/commits/0819e5fdb0d5199a1ad96de0ba63237414d4ec25).

sacpis: Issue #3039 created to handle step value and generator object.

schweitzpgi: Thank you.

Issue #1822, opened 2024-06-16
How to get the unitary matrix of a given kernel


ACE07-Sev:
Greetings there,

Hope all are well. I would like to know how I can get the unitary matrix of a circuit in cudaq. I've seen `.sample()` and `.get_state()` for counts and statevector definitions, but haven't been able to find the unitary matrix definition. Thanks in advance!

bettinaheim:
Tentatively moving this to 0.9 for tracking, but this is not yet confirmed as planned.

zohimchandani:
+1 for this since we need it for a collaboration, thanks.

ACE07-Sev:
I would make a few suggestions for this in the interest of performance. To get the unitary we have to do tensor contraction, and it gets expensive real fast as we increase the entanglement. With that being said, I would suggest using sth like `cotengra` or `contengrust` to find optimal order of contraction. Frankly, anything by Dr. Gray is usually very good for this type of operation.

Ohh, and one thing. It would be great if you don't do this as we add gates. Some packages (I believe qiskit) do this which adds alot of overhead for deep circuits. It would best to contract it at the end if the user wants a unitary matrix of the overall circuit. This also makes it easier to use JIT, or GPU-acceleration for performing this.

ACE07-Sev:
Greetings there,

Hope all are well. May I ask if there's an expected timeline for when this feature will be done?

zohimchandani:
You can use this hack in the meantime if it helps: 

```
import cudaq 
import numpy as np
from typing import List


num_qubits = 2
input_state = [0.0] * (2**num_qubits)  #just input a zero state 
N = 2**num_qubits
U = np.zeros((N, N), dtype=np.complex128)
params = [1 ,2]

@cudaq.kernel
def kernel(params: List[float], input_state: List[complex]):

    qubits = cudaq.qvector(input_state)
    
    rx(params[0], qubits[0])
    ry(params[1], qubits[1])
    
    y.ctrl(qubits[0], qubits[1])
    

for j in range(N): 
    state_j    = np.zeros((N), dtype=np.complex128) 
    state_j[j] = 1.0
    
    U[:, j] = np.array(cudaq.get_state(kernel, params, state_j), copy=False)
    
print(U)

```

ACE07-Sev:
@zohimchandani Thank you!

ACE07-Sev:
[...]

PR #3028, opened 2025-06-10
Initial implementation of getting matrix for kernel

Randl:
Closes https://github.com/NVIDIA/cuda-quantum/issues/1822

I've added support for single-qubit gates; however, it appears that `getGateByName` doesn't support two-qubit gates:
https://github.com/NVIDIA/cuda-quantum/blob/dcb0abaf464463061df38992b00f7c3150363d18/runtime/nvqir/Gates.h#L31-L48

 It also may become more technical since a two-qubit gate unitary is not necessarily a Kronecker product of matrices.

Brief summary of changes done until now:
1. Moved `getKernelLaunchParameters` to `cudaq` namespace so it can be used outside of `draw`
2. Added `getGateNameFromString` that returns an instance of `GateName` given the string name
3. Added `get_data` to `complex_matrix` (not sure about this one, but `complex_matrix` appeared as a correct abstraction, but there is no way to access the data)
4. Added the following functions: 
    -  `expand_gate_to_system`: given gate matrix creates the unitary applying this gate to the whole system (only single-qubit gates for now)
    -  `unitary_from_trace`: gets trace, iterates over instructions in it, and multiplies the resulting unitaries
    - `get_unitary_cmat`: given the kernel generates trace and gets unitary from it
5. Added Python `get_unitary` function
6. Added Python tests that verify the correctness of the code for different single-qubit gates.

I'm not sure whether 1-3 is the correct approach, but it felt reasonable.

TODO:

- [x] Support multi-qubit gates. This probably requires `expand_gate_to_system` rewrite
- [ ] Add C++ tests and more python tests

CC @khalatepradnya @nvidia-dobri


Randl:
Probably the simplest solution of expanding non-consecutive multiqubit gate is to apply swap to expanded gate

khalatepradnya:
> /ok to test c2a61a65e53dab8050109801365aa5b64233397a

khalatepradnya:
For code formatting, please follow instructions here - https://github.com/NVIDIA/cuda-quantum/blob/main/Developing.md#code-style

python/runtime/cudaq/algorithms/py_unitary.cpp
```diff
@@ -0,0 +1,55 @@
+#include "cudaq/algorithms/draw.h"
```
khalatepradnya, line 9: Please add copyright banner -
```
 * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
 * All rights reserved.                                                        *
 *                                                                             *
```

khalatepradnya:
Note that the other CI failures like `Error: Process completed with exit code 255.` may be unrelated to this PR and are being looked at separately.

Randl:
I've added support for control gates and swaps by implementing control gates and permutations. Also fixed the formatting, added some documentation, and more tests in Python.

Main problem I think is missing support for custom gates, which would require changes in tracing that doesn't track the custom gate content.

khalatepradnya:
> /ok to test 3a2f2466d8aeee4a068a3916a3d8c82151e346ed

Randl:
Added the missing header and fixed spelling

nvidia-dobri:
> /ok to test e0db6357285e6639035839025133f901c595adab

runtime/cudaq/utils/matrix.h
```diff
@@ -197,6 +197,7 @@ class complex_matrix {
   std::size_t rows() const { return dimensions.first; }
   std::size_t cols() const { return dimensions.second; }
   std::size_t size() const { return get_size(dimensions); }
+  value_type *get_data() const { return data; }
```
nvidia-dobri: Yeah, I don't think exposing a raw pointer here is a great idea. To clean this up a bit, how about making `cmat_to_numpy` a friend, passing the entire matrix and using the rows/cols/data internal to the func? A bit unrelated, but nice cleanup.

runtime/cudaq/algorithms/unitary.h
```diff
@@ -0,0 +1,176 @@
...
+
+/// @brief Execute a quantum kernel and return its unitary as a flat
+/// std::vector.
+/// @tparam QuantumKernel The functor type of the quantum kernel.
+/// @param kernel The quantum kernel to execute.
+/// @param args Arguments to pass to the kernel.
+/// @returns A std::vector<std::complex<double>> representing the unitary in
+/// row-major order.
+template <typename QuantumKernel, typename... Args>
+std::vector<std::complex<double>> get_unitary(QuantumKernel &&kernel,
```
nvidia-dobri: What's the purpose of this overload? We should avoid APIs with flattened matrices since the row/column order is arbitrary (even if documented).

runtime/cudaq/algorithms/unitary.h
```diff
@@ -0,0 +1,176 @@
...
+    std::size_t gate_dim = static_cast<std::size_t>(std::sqrt(gate_vec.size()));
+    complex_matrix gate(gate_vec, {gate_dim, gate_dim});
+
+    // If there are control qubits, build the controlled-unitary
+    if (!inst.controls.empty()) {
+      gate = make_controlled_unitary(gate, inst.controls.size());
+      gate_dim = gate.rows();
+    }
+
+    // Get vector of all qubit indices that this gate operates on
```
nvidia-dobri, line 125: The fact that controls must come first here for everything to line up is worth a comment for posterity, probably.

runtime/cudaq/algorithms/unitary.h
```diff
@@ -0,0 +1,176 @@
...
+/// identities and permuting qubits into the specified positions.
+/// @param gate The unitary matrix acting on m qubits.
+/// @param num_qudits The total number of qubits in the full system.
+/// @param qudit_indices A vector of size m giving the target qubit indices in
+/// the full system.
+/// @returns A complex_matrix representing the expanded gate on the full n-qubit
+/// system.
+inline complex_matrix
+expand_gate_to_system(const complex_matrix &gate, std::size_t num_qudits,
+                      const std::vector<std::size_t> &qudit_indices) {
```
nvidia-dobri, line 50: This is pretty tight, good work! If we could use `qpp::expandout` or even just `qpp::kron`, that would be better than adopting the complexity of this into our own codebase, though.

nvidia-dobri (commented):
Good work so far, some comments to chew on

runtime/cudaq/algorithms/unitary.h
Randl: It was mentioned in the instructions in https://github.com/NVIDIA/cuda-quantum/issues/1822
I can remove it.

runtime/cudaq/algorithms/unitary.h
Randl: What do you mean?

runtime/cudaq/algorithms/unitary.h
Randl: I don't see how `qpp::kron` would help here, given that we already have `kronecker`. 
`expandout` could save a few lines of code, though it will require converting to `Eigen` if I understand correctly.

We could also simplify a bit via something like
```
  std::vector<complex_matrix> facs = {gate, complex_matrix::identity(1ULL << (num_qudits - m))};
  auto U0 = kronecker(facs.begin(), facs.end());
  ```

Comment (Randl) [reply runtime/cudaq/algorithms/unitary.h:50]: However, the most complexity comes from multiqubit gates that need to be permuted into correct order, and `expandout` wouldn't help with that (though I'm unfamiar with qpp so maybe there is a different function that does that)

Comment (Randl) [reply runtime/cudaq/algorithms/unitary.h:50]: Looks like `qpp::CTRL` may be helpful in this case; not 100% sure tho

File: runtime/cudaq/utils/matrix.h (hunk 2 above)
Comment (Randl) [reply runtime/cudaq/utils/matrix.h]: Making `cmat_to_numpy` requires introducing `pybind`,  which is now limited to `python`, into the `cudaq`. If you're ok with that I can try to do that

Comment (nvidia-dobri) [reply runtime/cudaq/utils/matrix.h]: Yeeeah... Yeah let's not do that. We may not have a choice but to follow the unpleasant pattern of friending users of data. Feels like the lesser of two evils (vs exposing the pointer to anyone that wants it).

File: runtime/cudaq/algorithms/unitary.h (hunk 4 above)
Comment (nvidia-dobri) [reply runtime/cudaq/algorithms/unitary.h:125]: Just document that the order things are added to `inst_qubits` matters... unless I'm missing something. Easy to break things when doing refactorings and moving things around when there's implicit ordering to "flat" containers like that.

Comment (Randl) [reply runtime/cudaq/algorithms/unitary.h:125]: Ah, thats fair

File: runtime/cudaq/algorithms/unitary.h (hunk 5 above)
Comment (nvidia-dobri) [reply runtime/cudaq/algorithms/unitary.h:50]: Yeah, I think you're right. I misread `expandout` to be an equivalent to what you had here, but it's just a subset. Hm, strange that qpp doesn't have that out of the box. Test it out, if you can get a clean(er) implementation with `qpp::CTRL` (I think doable), I'd still prefer to keep the intricate matrix manipulations out of the codebase.

File: runtime/cudaq/algorithms/unitary.h (hunk 3 above)
Comment (nvidia-dobri) [reply runtime/cudaq/algorithms/unitary.h]: Yeah, let's remove it. @khalatepradnya wrote the original instructions and we confirmed this is fine now. I'll update the issue.

**Comment by nvidia-dobri** [comment]
@Randl any updates here?

**Comment by Randl** [comment]
Hi sorry, it was a weekend and we got a bit of war here so I'm cut off my work PC as for now.
It looks like at most 1-2 hours of work when I get there:
1. Remove C++ api
2. Try qpp rewrite, which potentially takes care of other issues
3. Add docs for parameter order passed to expand
4. Don't use data pointer access

 Hopefully some time next week, but can't promise.

File: python/runtime/cudaq/algorithms/py_draw.cpp [hunk 6]
```diff
@@ -17,7 +17,6 @@
 
 namespace cudaq {
 
-namespace {
```
schweitzpgi, line 20: Let's move this function into a subnamespace. `cudaq::detail` or `cudaq::draw` would work here.

python/runtime/cudaq/algorithms/py_unitary.cpp
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include "cudaq/algorithms/draw.h"
```
schweitzpgi, line 9:
```suggestion

#include "cudaq/algorithms/draw.h"
```
nit: add whitespace

runtime/cudaq/utils/matrix.h
Randl: I've found a middle ground by making `cmat_to_numpy` a friend

runtime/cudaq/algorithms/unitary.h
Randl: @nvidia-dobri While `qpp::CTRL` indeed can replace my control implementation, `expandout` (for non-control version) doesn't support multi-qubit gates. So we still would need matrix manipulations. It seems like a simple change to qpp, so maybe we can open an issue in qpp to do that and use it when it's done?

```cpp
    std::vector<std::size_t> control_qubits, target_qubits;
    for (const auto &control : inst.controls) {
      control_qubits.push_back(control.id);
    }
    for (const auto &target : inst.targets) {
      target_qubits.push_back(target.id);
    }

    // Expand gate to full system
    auto eigen_gate = control_qubits.empty()
                         ? qpp::gt.expandout(gate.as_eigen(), target_qubits, num_qubits, 2)
                         : qpp::gt.CTRL(gate.as_eigen(), control_qubits, target_qubits,
                         num_qubits, 2);
    // expand_gate_to_system(gate, num_qubits, inst_qubits);
    
    std::vector<complex_matrix::value_type> v(eigen_gate.data(), eigen_gate.data() + eigen_gate.size());
    std::pair<std::size_t, std::size_t> dim = {static_cast<std::size_t>(eigen_gate.rows()),
                            static_cast<std::size_t>(eigen_gate.cols())};
    auto full_gate = complex_matrix(v, dim);
 ```

Randl:
I think I've adressed all the comments. Sorry for the delay

khalatepradnya:
> /ok to test 1a54a08

Randl:
The CI fail doesn't seem to be related to my changes? I've rebased on main just in case

runtime/cudaq/algorithms/unitary.h
nvidia-dobri: Yeah, seems like a good suggestion

nvidia-dobri:
> /ok to test 1a2cc0511561fce69d540293ef962151e1f4d436

nvidia-dobri (approved):
Let's just make sure all tests and CI passes - this looks much better to me, thanks for working on it!

Randl:
Looks like lint jobs are not running?

nvidia-dobri:
> /ok to test 12148ab9405d361fde896962b9c6e4a39297252d

nvidia-dobri:
> Looks like lint jobs are not running?

Yeah, we need to approve every commit from an external contributor, unfortunately. Running them now

Randl:
any updates?

khalatepradnya:
> /ok to test a3b428389b9cd3732585f95bd8b893ebbbd1629e

nvidia-dobri:
> any updates?

We had some internal conversations about this - there are some follow-up items we'd like to take care of internally, but I think they're easier as a separate PR on our side. Let's get this merged in, I think it looks great!

@bettinaheim @amccaskey FYI

nvidia-dobri:
> /ok to test 940c8a7d5c8d3a563a8fbcedf5684e63ea7f3567

Randl:
Awesome!

Here are a number of things worth mentioning that are missing as they are limitations of the tracer:

1. No support for custom gates (there is a test for that, it currently checks that there is an exception)
2. Unitary size is determined by the index of last used qubit rather than size of the kernel
3. No qudit support

nvidia-dobri:
> /ok to test d9ebdedad7c5c3baeebb9ff09c6a293250e2119b

PR #3030, opened 2025-06-10
test_cupy_integration.py: explicitly set cudaq target

mitchdz:
publishing is failing[0].

It is likely from the pytest tests running out of order and these tests inheriting the wrong target. Explicitly setting the target as best practice.

Also, fixing deprecated target calls.

[0] - https://github.com/NVIDIA/cuda-quantum/actions/runs/15504683965/job/43659398978#step:7:9653


1tnguyen (approved):
LGTM 👍

mitchdz:
> /ok to test 4145054

mitchdz:
> /ok to test b052558

Issue #3031, opened 2025-06-10
Define behavior of tuples in kernels for c++ and python


annagrin:
### Required prerequisites

- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Define behavior of tuples in kernels for c++ and python

- Do we want to support use of `tuple` and `Tuple` in python, and `std::tuple` in c++, inside kernels?
- Define behavior on 
  - assignment
  - copy
  - updating elements
  - unpacking
  - assignment to references to elements of the tuple
  - all of the above for tuples containing aggregate values (structs, tuples, vectors)

bettinaheim:
I cleaned up the behavior in https://github.com/NVIDIA/cuda-quantum/pull/3537. Here are the answers to the above questions. 

Tldr: Tuples in kernels behave like Python tuples. They are value types in python, and their items cannot be modified. 
- Assignment: formally creates a copy of the tuple, like any other value type (e.g. integer). Under the hood, we don't actually copy the value since there is nothing that would modify a tuple value once created.
- Copy: See assignment above
- Updating elements: Forbidden and results in the bridge giving the appropriate error
- Unpacking: The unpack operator (`*`) is not currently supported, but could be supported for tuples in the future. If you mean deconstruction, then deconstruction is supported. Deconstruction behaves like any other assignment of items. i.e. 
```
t = (1, 2)
a, b = t
# is equivalent to
a = t[0]
b = t[1]
```
- Assignment to tuple elements: Tuple elements can only be "assigned to" when constructing a tuple. In principle, these assignments behave like one would expect from an assignment of the corresponding data type; if the item type t is a reference type (e.g. list), then the assignment is by reference. The only deviation from Python behavior is that we do not allow modification of any tuple items via tuple item access:
```
l = [0]
t = (l, 1)
t[0][0] = 5 # gives an error within quantum kernels
l[0] = 5 # results in both t and l1 being updated 
```
- The above statements should cover all cases, including tuples containing aggregate values. The behavior of dataclasses within kernels in general matches the Python behavior, albeit with additional restrictions imposed. The same holds for lists.

Issue #3029, opened 2025-06-10
[run, python] Unpacking tuples gives an assertion error


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Unpacking tuples gives an assertion error 

### Steps to reproduce the bug

```
    @cudaq.kernel()
    def kernel(t: tuple[bool, int]) -> int:
        a, b = t 
        return b

    # assertion error
    print(kernel2)
    r = cudaq.run(kernel, (False,13), shots_count= 2)
    print(r) 
    assert r = [13, 13]
```

### Expected behavior

Test passes

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

Issue #3015, opened 2025-06-06
[Python] Tuple reassignment doesn't work in kernels


khalatepradnya:
### Describe the bug

Reassigning a tuple does not work in kernels.

### Steps to reproduce the bug

```python
    @cudaq.kernel
    def simple_tuple_int_float_reassign(n: int, t: tuple[int, float]) -> tuple[int, float]:
        qubits = cudaq.qvector(n)
        t = (-14, 11.5)
        return t

    results = cudaq.run(simple_tuple_int_float_reassign, 2, (13, 42.3))

```

Fails with 
`error: 'cc.store' op failed to verify that type of value matches element type of pointer`

### Expected behavior

Test passes

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: `main` as of commit ID 7fa5c5e42580f02610e1c2e1beb22749ed81d3a8
- **Python version**: 3.10
- **C++ compiler**: N/A
- **Operating system**: Ubuntu

PR #3032, opened 2025-06-10
Disable used of tuples in python

annagrin:
### Description
Using tuples in kernels needs a spec. Disable the current support due to multiple unresolved issues:

Related: https://github.com/NVIDIA/cuda-quantum/issues/3031
Related: https://github.com/NVIDIA/cuda-quantum/issues/2965
Related: https://github.com/NVIDIA/cuda-quantum/issues/3015
Related: https://github.com/NVIDIA/cuda-quantum/issues/3029


python/tests/kernel/test_run_kernel.py
```diff
@@ -364,7 +393,9 @@ def simple_list_int(n: int, t: list[int]) -> list[int]:
         return t
 
     results = cudaq.run(simple_list_int, 2, [-13, 5, 42], shots_count=2)
-    # FIXME: Non-const size of stdvec - ReturnToOutputLog does not create output.
+    #FIXME: Non-const size of stdvec - ReturnToOutputLog does not create output.
```
schweitzpgi, line 396: Moreover, the intent of the QIR spec is to _not_ support dynamic/variable length vectors/arrays of output.

annagrin: This is just adding an issue reference to the comment. The issue itself is not related to tuples.

 I will add the question about the QIR spec to the issue (it is a bit complicated since after synthesis this issue disappears so simulators are at disadvantage if we don’t allow non constant arrays)

schweitzpgi (approved):

PR #3034, opened 2025-06-10
[C++] Expose the `u3` gate via kernel builder

khalatepradnya:


runtime/cudaq/builder/kernel_builder.h
```diff
@@ -746,6 +750,100 @@ class kernel_builder : public details::kernel_builder_base {
...
+          QuakeValue target) {
+    std::vector<QuakeValue> empty;
+    std::vector<QuakeValue> parameters{param1, param2, param3};
+    details::u3(*opBuilder, parameters, empty, target);
+  }
+  [[deprecated("In the future, passing `ctrls` to u3 will require an explicit "
+               "`<cudaq::ctrl>` template argument. "
+               "Upon the next release, this will be interpreted as a single "
+               "qubit gate broadcast across all input qubits, per the CUDA-Q "
+               "Specification.")]] void
```
schweitzpgi: Should we just go ahead and make this a broadcast ctor now?

runtime/cudaq/builder/kernel_builder.h
```diff
@@ -746,6 +750,100 @@ class kernel_builder : public details::kernel_builder_base {
...
+        QuakeValue v1(*opBuilder, param1);
+        QuakeValue v2(*opBuilder, param2);
+        QuakeValue v3(*opBuilder, param3);
+        std::vector<QuakeValue> parameters{v1, v2, v3};
+        u3<cudaq::ctrl>(parameters, ctrls, target);
+      } else {
+        std::vector<QuakeValue> parameters{param1, param2, param3};
+        u3<cudaq::ctrl>(parameters, ctrls, target);
+      }
+      return;
```
schweitzpgi: nit: unneeded dangling `return`

runtime/cudaq/builder/kernel_builder.h
khalatepradnya: Since that would require changes across all the gates, I prefer to do that as a follow-up. Here, the intent is to align behavior with other gates.

runtime/cudaq/builder/kernel_builder.h
khalatepradnya: Fixed in 56ec3c5749bcb63af7bb98e5fee9da4456c46da6

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -742,23 +742,16 @@ CUDAQ_ONE_QUBIT_PARAM_IMPL(rz, RzOp)
 CUDAQ_ONE_QUBIT_PARAM_IMPL(r1, R1Op)
 
 void u3(ImplicitLocOpBuilder &builder, std::vector<QuakeValue> &parameters,
```
nvidia-dobri, line 744: Oh, so u3 was in the `.cpp` but not the `.h`? Is there any way it's used in the wild, before we recycle the signature?

runtime/cudaq/builder/kernel_builder.h
nvidia-dobri: We probably don't want to introduce already-deprecated API flavor just for the sake of completion, do we? Refactors aside, let's just keep the `cudaq::ctrl` flavor around, imo.

unittests/integration/builder_tester.cpp
```diff
@@ -430,7 +430,83 @@ CUDAQ_TEST(BuilderTester, checkRotations) {
...
+#ifndef CUDAQ_BACKEND_DM
+CUDAQ_TEST(BuilderTester, checkU3) {
+  cudaq::set_random_seed(13);
+
+  // Simple U3 gate.
+  {
+    auto kernel = cudaq::make_kernel();
+    auto target = kernel.qalloc();
+    kernel.u3(M_PI_2, 0, M_PI, target);
+    auto counts = cudaq::sample(kernel);
```
nvidia-dobri, line 442: Should we add explicit shot counts to these tests? The == 1000 assertion with no explicit 1000 shots makes me uncomfortable... 😅

nvidia-dobri (approved):
Looks good, thanks!

runtime/cudaq/builder/kernel_builder.cpp
khalatepradnya: Yes, and it was incorrect, so anything using it (I don't know of any) should be updated anyway.

unittests/integration/builder_tester.cpp
khalatepradnya: The default is expected to be `1000`, and if it doesn't match then we have a bug!

runtime/cudaq/builder/kernel_builder.h
khalatepradnya: I like the idea of not adding more deprecated API, so I will remove this altogether....

khalatepradnya: Fixed in 4399fb8b7960c743b7888d356edf3d478a19dbb4. `cudaq::ctrl` stays.

PR #3038, opened 2025-06-11
Fix an issue with tensornet observe when the canonicalized `spin_op` doesn't span the entire qubit space

1tnguyen:
### Description

**Symptom**: CUTENSORNET_INVALID_VALUE error when using `observe` with a `spin_op` that doesn't span the entire qubit space, e.g., `spin_op` with a single term `YIXXIZYZYYYIIY`.

This is because we're using the number of elements in the `degrees` list to create the cutensornet's network operator, while adding Pauli operators at the **values** of those elements. Hence, this caused out-of-range (invalid value) exceptions. 

The fix is that we need to find the actual max index to determine the size of the network operator to be created.

This is a regression since https://github.com/NVIDIA/cuda-quantum/pull/2710, whereby the canonicalized `spin_op` no longer tracks degrees of identity Pauli. 

Also,

- Add an example, which will be validated by the Publishing pipeline.

- Add unit tests, which will be validated by QA.


unittests/integration/observe_result_tester.cpp
```diff
@@ -124,4 +124,42 @@ CUDAQ_TEST(ObserveResult, checkExpValBug) {
...
+  const std::size_t numQubits = pauliWord.size();
+  auto pauliOp = cudaq::spin_op::from_word(pauliWord);
+  // The canonicalized degree list is less than the number of qubits
+  EXPECT_LT(cudaq::spin_op::canonicalize(pauliOp).degrees().size(), numQubits);
+  auto expVal = cudaq::observe(kernel, pauliOp);
+  std::cout << "<" << pauliWord << "> = " << expVal.expectation() << "\n";
+  EXPECT_NEAR(expVal.expectation(), -0.416147, 1e-6);
+}
+
+#ifdef CUDAQ_BACKEND_TENSORNET
```
khalatepradnya, line 145: Q: Does this change apply to `tensornet-mps` as well? (Should it?)

khalatepradnya (approved):
LGTM. Thanks, Thien!
(One Q about scope)

mitchdz (approved):
Similar question to Pradnya, otherwise LGTM!

unittests/integration/observe_result_tester.cpp
1tnguyen: Yea, it will cover both `tensornet` and `tensornet-mps`. This macro define is on for both of them: https://github.com/NVIDIA/cuda-quantum/blob/5d825b60816e1d832630391f11288c3cac31805a/unittests/CMakeLists.txt#L87-L102

`tensornet-mps` has an additional macro `CUDAQ_BACKEND_TENSORNET_MPS` that can be used to target it specifically.

Indeed, this fix to `tensornet_spin_op` should be effective for both, hence we test both backends here.

Issue #3037, opened 2025-06-10
[C++] [run] Kernel not returning expected values


khalatepradnya:
### Describe the bug

Following kernel does not return expected values

### Steps to reproduce the bug

```
#include <cudaq.h>

struct MyTuple {
  bool boolVal;
  std::int64_t i64Val;
  double f64Val;
};

auto struct_test = [](MyTuple t) __qpu__ { return t; };

int main() {
  int c = 0;
  {
    MyTuple t = {true, 654, 9.123};
    const auto results = cudaq::run(3, struct_test, t);
    if (results.size() != 3) {
      printf("FAILED! Expected 3 shots. Got %lu\n", results.size());
    } else {
      c = 0;
      for (auto i : results)
        printf("%d: {%s, %ld, %f}\n", c++, i.boolVal ? "true" : "false",
               i.i64Val, i.f64Val);
      printf("success!\n");
    }
  }
  return 0;
}
```

Build and run -
```
# bin/nvq++ -fenable-cudaq-run test.cpp && ./a.out 
FAILED! Expected 3 shots. Got 0
```

### Expected behavior
Runs successfully 
```
0: {true, 654, 9.123000}
1: {true, 654, 9.123000}
2: {true, 654, 9.123000}
success!
```

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: `main` as of commit ID 4bc957a1ef4f1445bfebc491009b317976a1e266
- **Python version**: 3.10
- **C++ compiler**: N/A
- **Operating system**: Ubuntu

schweitzpgi:
The root cause of this is the MLIR inliner. The inliner in LLVM 16 will optimistically "inline" (erase the call) any function that is effectively a NOP, which is the case in the lambda in the example.

The Inliner in our version of MLIR has some controls over inlining, but not quite the set that is needed in this case. Indeed, there is already code to prevent this exact sort of inlining, but the control does not apply to a trivial kernel that just forwards an argument like in the example.

PR #3048, opened 2025-06-12
[core] Fix a typo in inliner hook.

schweitzpgi:

### DIFF
--- a/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
+++ b/lib/Optimizer/Dialect/Quake/QuakeOps.cpp
@@ -982,9 +982,9 @@ bool cudaq::EnableInlinerInterface::isLegalToInline(Operation *call,
   if (auto applyOp = dyn_cast<quake::ApplyOp>(call))
     if (applyOp.applyToVariant())
       return false;
-  if (auto destFunc = call->getParentOfType<mlir::func::FuncOp>())
+  if (auto destFunc = call->getParentOfType<func::FuncOp>())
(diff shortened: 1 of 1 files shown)


lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -982,9 +982,9 @@ bool cudaq::EnableInlinerInterface::isLegalToInline(Operation *call,
   if (auto applyOp = dyn_cast<quake::ApplyOp>(call))
     if (applyOp.applyToVariant())
       return false;
-  if (auto destFunc = call->getParentOfType<mlir::func::FuncOp>())
+  if (auto destFunc = call->getParentOfType<func::FuncOp>())
     if (destFunc.getName().ends_with(".thunk"))
-      if (auto srcFunc = call->getParentOfType<mlir::func::FuncOp>())
+      if (auto srcFunc = dyn_cast<func::FuncOp>(callable))
```
khalatepradnya, line 987: Q: Does this fix https://github.com/NVIDIA/cuda-quantum/issues/3037?
If so, can we add that as a test?

schweitzpgi: No, unfortunately.

schweitzpgi: Since the inliner is run ubiquitously, we are already testing inlining. That we haven't noticed this before now suggests this path is rarely used, if ever.

PR #3049, opened 2025-06-12
Adding queue threshold

### LINKED ISSUE #218: [Enhancement] Optimize memory usage in gate queues
There are a couple of queues in the runtime, e.g., `BasicExecutionManager::instructionQueue` and `CircuitSimulatorBase::gateQueue`. 

Currently, we don't enforce any max queue size mechanism to safeguard these queues. This could lead to memory bloat when the circuit is very deep, especially with `CircuitSimulatorBase::gateQueue` where we store the gate matrix in the queue.

A contrived example:
```
auto operator()() __qpu__ {
    cudaq::qubit q;
    for(std::size_t i = 0; i < std::numeric_limits<std::size_t>::max(); ++i) {
      x(q); 
    }
}
```
That'd result in a large memory footprint although we only use 1 qubit :-) 
There might be more practical use cases, e.g., Trotter circuits.

sacpis:
Adding queue threshold and flushing it when the instructions in the queue reaches the threshold.

Fixes #218

Note: This can be further optimized by determining the queue threshold dynamically based on the available memory.

### DIFF
--- a/runtime/cudaq/qis/managers/BasicExecutionManager.h
+++ b/runtime/cudaq/qis/managers/BasicExecutionManager.h
@@ -245,6 +245,11 @@ public:
     // Add to the instruction queue
     instructionQueue.emplace_back(std::move(mutable_name), mutable_params,
                                   mutable_controls, mutable_targets, op);
+
+    static constexpr std::size_t kMaxQueueSize = 1024;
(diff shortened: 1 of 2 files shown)


amccaskey:
this may not be worth it since we're dropping the execution managers with the upcoming refactor.

sacpis:
Thanks @amccaskey. Closing this PR and associated issue #218.

PR #3051, opened 2025-06-12
Add simulator and replace runtime call for resource counting

atgeller:
### Description

Introduces a `ResourceCounter` simulator, which counts qubit allocations and gate usages without actually simulating any quantum state. Replaces the `estimate_resources` function with a new one that uses this simulator to perform resource counting. Also introduces an overload of `estimate_resources which accepts a choice function of type `bool()`, which is used to force the results of measurements in case control flow depends on measurement results. This allows exploring resource usage on different paths.

The new `estimate_resources` function changes the backend simulator to use the custom `ResourceCounter` simulator, but otherwise the compilation pipeline is run as specified for the designated target, allowing resource counting to be performed accurately for different targets. However, to optimize resource counting, invariant quantum operations are identified, counted, and removed during compilation.

### Breaking changes:
Return type of `estimate_resources` is now a `resource_counts` object, instead of a `Resources` object. Notably, the `resource_counts` object does not track which qubits the gate is being applied to, and tracks the number of controls via the gate string (e.g., `"ccx"` represents an `x` gate with two controls). `resource_counts` supports the following operations intended for the user to query the results of resource estimation:
```
/// @brief Return the number of times the given gate was observed
std::size_t count(const std::string &gate) const;

/// @brief Return the total number of gates observed
std::size_t count() const;

/// @brief Dump this resource_counts to standard out.
void dump() const;

/// @brief Dump this resource_counts to the given output stream
void dump(std::ostream &os) const;
```

### DIFF
--- a/python/runtime/cudaq/platform/JITExecutionCache.h
+++ b/python/runtime/cudaq/platform/JITExecutionCache.h
@@ -5,13 +5,12 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
+
 #pragma once
 
 #include "mlir/ExecutionEngine/ExecutionEngine.h"
 #include <mutex>
 #include <unordered_map>
-
-using namespace mlir;
 
 namespace cudaq {
 
@@ -27,7 +26,7 @@ protected:
   // the execution engine and to the LRU iterator that is used to track which
   // engine is the least recently used.
   struct MapItemType {
-    ExecutionEngine *execEngine = nullptr;
+    mlir::ExecutionEngine *execEngine = nullptr;
     std::list<std::size_t>::iterator lruListIt;
   };
   std::unordered_map<std::size_t, MapItemType> cacheMap;
@@ -38,8 +37,8 @@ public:
   JITExecutionCache() = default;
   ~JITExecutionCache();
 
-  void cache(std::size_t hash, ExecutionEngine *);
+  void cache(std::size_t hash, mlir::ExecutionEngine *);
   bool hasJITEngine(std::size_t hash);
-  ExecutionEngine *getJITEngine(std::size_t hash);
+  mlir::ExecutionEngine *getJITEngine(std::size_t hash);
 };
 } // namespace cudaq
--- a/runtime/nvqir/cutensornet/tensornet_state.inc
+++ b/runtime/nvqir/cutensornet/tensornet_state.inc
@@ -814,6 +814,14 @@ TensorNetState<ScalarType>::computeExpVals(
                                       /*cudaStream*/ 0));
   }
 
+  if (::cudaq::details::should_log(::cudaq::details::LogLevel::info)) {
+    double flops = 0.0;
+    HANDLE_CUTN_ERROR(cutensornetExpectationGetInfo(
+        m_cutnHandle, tensorNetworkExpectation,
+        CUTENSORNET_EXPECTATION_INFO_FLOPS, &flops, sizeof(flops)));
+    cudaq::info("Total flop count = {} GFlop.", (flops / 1e9));
+  }
+
   // Attach the workspace buffer
   int64_t worksize{0};
   HANDLE_CUTN_ERROR(cutensornetWorkspaceGetMemorySize(
@@ -947,6 +955,14 @@ std::complex<ScalarType> TensorNetState<ScalarType>::computeExpVal(
     HANDLE_CUTN_ERROR(cutensornetExpectationPrepare(
         m_cutnHandle, tensorNetworkExpectation, scratchPad.scratchSize,
         workDesc, /*cudaStream*/ 0));
+  }
+
+  if (::cudaq::details::should_log(::cudaq::details::LogLevel::info)) {
+    double flops = 0.0;
+    HANDLE_CUTN_ERROR(cutensornetExpectationGetInfo(
+        m_cutnHandle, tensorNetworkExpectation,
+        CUTENSORNET_EXPECTATION_INFO_FLOPS, &flops, sizeof(flops)));
+    cudaq::info("Total flop count = {} GFlop.", (flops / 1e9));
   }
 
   // Attach the workspace buffer


atgeller:
> /ok to test 0fd858ba8233cc0abdd0d34d3395eaa43301b955

atgeller:
> /ok to test 07851e48cfafe1f4ffff0b5f51eb2bcebdd65674

python/runtime/cudaq/algorithms/py_resource_count.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+
+#include <fmt/core.h>
+#include <pybind11/functional.h>
+#include <pybind11/stl.h>
+
+namespace py = pybind11;
+
+namespace cudaq {
+void pyAltLaunchKernel(const std::string &, MlirModule, OpaqueArguments &,
```
annagrin: I think we have that defined in "runtime/cudaq/platform/py_alt_launch_kernel.h" already

python/runtime/cudaq/algorithms/py_resource_count.cpp
```diff
@@ -0,0 +1,68 @@
...
+          kernel.attr("compile")();
+        auto &platform = cudaq::get_platform();
+        auto kernelName = kernel.attr("name").cast<std::string>();
+        auto kernelMod = kernel.attr("module").cast<MlirModule>();
+        args = simplifiedValidateInputArguments(args);
+        auto *argData = toOpaqueArgs(args, kernelMod, kernelName);
+
+        auto ctx = std::make_unique<ExecutionContext>("resourcecount", 1);
+        ctx->kernelName = kernelName;
+        ctx->choice = choice;
```
annagrin: I am wondering - can the choice function be directly communicated to the resource counting simulator, to avoid extra fields on execution context?

runtime/common/ExecutionContext.h
```diff
@@ -127,5 +127,14 @@ class ExecutionContext {
   /// @brief Whether or not to simply concatenate measurements in execution
   /// order.
   bool explicitMeasurements = false;
+
+  /// @brief Counts of resources (gates and qubits) used when in resource
+  /// counting mode
+  resource_counts resourceCounts;
```
annagrin: can `resourceCounts` and `choice` function be part of the resource count simulator? Is there a reason to save them on the execution context?

python/runtime/cudaq/algorithms/py_resource_count.cpp
atgeller: Yes, that looks like an unnecessary forward reference.

python/runtime/cudaq/algorithms/py_resource_count.cpp
atgeller: I think that would require having direct access to the simulator in the `count_resources` runtime functions for both C++ and python, which should be doable but may be a bit tricky.

runtime/common/ExecutionContext.h
atgeller: I think at least having `resourceCounts` be part of the execution context makes sense since the execution context already has fields for all of the other possible result objects, and the execution context is more readily accessible then the simulator.

atgeller: Thinking about it more, having `resourceCounts` be part of the execution context is necessary, since the preprocessing pass during JIT compilation uses it, which happens outside the context of the simulator. I think having the choice function being part of the execution context makes sense for the same reason, in case we someday want it to be accessible by preprocessing passes.

python/runtime/cudaq/algorithms/py_resource_count.cpp
atgeller: Looking into it, I'm not sure how doable this is, because there's currently a linking issue that prevents directly referencing the ResourceCounter simulator in the C++ `count_resources` function, which we would have to solve or work around.

atgeller:
> /ok to test 23dc83185f289ea37e460073132299cd85b5f5ee

runtime/common/ExecutionContext.h
atgeller: We've figured out a nice solution that negates this need.

python/runtime/cudaq/algorithms/py_resource_count.cpp
atgeller: I've replaced the extra fields on executionContext with a static ResourceCounter simulator and static functions to set the choice function and get the resource counts in nvqir.

annagrin (approved):
LGTM, thanks Adam!

annagrin:
> /ok to test

annagrin:
> /ok to test ce0074202234531af6aa381605d426c9756e63fc

python/runtime/common/py_ResourceCounts.cpp
```diff
@@ -0,0 +1,53 @@
...
+      R"#(A data-type containing the results of a call to :func:`resource_count`. 
+This includes all gate counts.)#")
+      .def(py::init<>())
+      .def(
+          "dump", [](resource_counts &self) { self.dump(); },
+          "Print a string of the raw resource counts data to the "
+          "terminal.\n")
+      .def(
+          "count",
+          [](resource_counts &self, const std::string gate) {
```
1tnguyen: nit:
```suggestion
          [](resource_counts &self, const std::string& gate) {
```

python/runtime/common/py_ResourceCounts.h
```diff
@@ -0,0 +1,17 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include <pybind11/pybind11.h>
+
+#include "utils/LinkedLibraryHolder.h"
```
1tnguyen: nit: looks like this `#include` is unnecessary.

python/runtime/cudaq/algorithms/py_resource_count.cpp
```diff
@@ -0,0 +1,74 @@
...
+  mod.def(
+      "count_resources",
+      [&](std::function<bool()> choice, py::object kernel, py::args args) {
+        if (py::hasattr(kernel, "compile"))
+          kernel.attr("compile")();
+        auto &platform = cudaq::get_platform();
+        auto kernelName = kernel.attr("name").cast<std::string>();
+        auto kernelMod = kernel.attr("module").cast<MlirModule>();
+        args = simplifiedValidateInputArguments(args);
+        auto *argData = toOpaqueArgs(args, kernelMod, kernelName);
```
1tnguyen: Looks like it's a leak here: `toOpaqueArgs` returns `OpaqueArguments*` with `new`.
```suggestion
        std::unique_ptr<OpaqueArguments> argData(toOpaqueArgs(args, kernelMod, kernelName));
```

python/tests/backends/test_resource_counter.py
```diff
@@ -0,0 +1,49 @@
+# ============================================================================ #
+# Copyright (c) 2025 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# Compile and run with:
```
1tnguyen:
```suggestion
```

python/tests/backends/test_resource_counter.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_resource_counter():
+
+    @cudaq.kernel
+    def mykernel():
+        q = cudaq.qubit()
+        p = cudaq.qubit()
+
+        # Alias them
```
1tnguyen:
```suggestion
```

python/tests/backends/test_resource_counter.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            m3 = mz(p)
+
+    def choice():
+        return True
+
+    cudaq.set_target("quantinuum", emulate=True)
+
+    counts1 = cudaq.sample(mykernel, shots_count=5)
+    counts2 = cudaq.count_resources(choice, mykernel)
```
1tnguyen: Just a suggestion: looking at this example, I wonder whether we should make the `choice` callback a `kw_only` argument.
For kernels that don't have measure gates, we don't need this argument. We could use a default one that always returns `True` (and document it) :)

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -521,7 +523,18 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
         cudaq::info("Run Quake Synth.\n");
         pm.addPass(cudaq::opt::createQuakeSynthesizer(kernelName, updatedArgs));
       }
-      pm.addPass(mlir::createCanonicalizerPass());
+      if (executionContext->name == "resourcecount") {
```
1tnguyen: Perhaps we can use `resource-count` for readability :)

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -521,7 +523,18 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
...
+        pm.addPass(mlir::createCanonicalizerPass());
+        std::function<void(std::string, size_t)> f = [&](std::string gate,
+                                                         size_t count) {
+          nvqir::getResourceCounts()->append(gate, count);
+          // executionContext->resourceCounts.append(gate, count);
```
1tnguyen: nit: we could remove this if unused.

runtime/common/ResourceCounts.h
```diff
@@ -0,0 +1,86 @@
...
+namespace cudaq {
+/// Type definition for the mapping of observed qubit measurement bit strings
+/// to the number of times they were observed.
+using CountsDictionary = std::unordered_map<std::string, std::size_t>;
+
+class resource_counts {
+private:
+  CountsDictionary gateCounts;
+
+  /// @brief Keep track of the total number of gates. We keep this
```
1tnguyen: We should also add docs comments for other member vars if possible.

runtime/common/ResourceCounts.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /// @brief Move assignment constructor
+  resource_counts &operator=(resource_counts &&counts) = default;
+
+  /// @brief The destructor
+  ~resource_counts() = default;
+
+  /// @brief Add another gate to this `resource_count`.
+  /// @param gate Encountered gate
+  void append(const GateData gate, size_t count = 1);
```
1tnguyen:
```suggestion
  void append(const GateData& gate, size_t count = 1);
```

runtime/common/ResourceCounts.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /// @brief Add another gate to this `resource_count`.
+  /// @param gate Encountered gate
+  void append(const std::string gate, size_t count = 1);
```
1tnguyen:
```suggestion
  void append(const std::string& gate, size_t count = 1);
```

runtime/common/ResourceCounts.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /// @brief Return the number of times the given gate was observed
+  /// @param gate
+  /// @return
+  std::size_t count(const GateData gate) const;
```
1tnguyen:
```suggestion
  std::size_t count(const GateData& gate) const;
```

runtime/common/ResourceCounts.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /// @brief Return the number of times the given gate was observed
+  /// @param gate
+  /// @return
+  std::size_t count(const std::string gate) const;
```
1tnguyen:
```suggestion
  std::size_t count(const std::string& gate) const;
```

runtime/cudaq/algorithms/count_resources.h
```diff
@@ -0,0 +1,89 @@
...
+#endif
+auto count_resources(std::function<bool()> choice, QuantumKernel &&kernel,
+                     Args &&...args) {
+  // Need the code to be lowered to llvm and the kernel to be registered
+  // so that we can check for conditional feedback / mid circ measurement
+  if constexpr (has_name<QuantumKernel>::value) {
+    static_cast<cudaq::details::kernel_builder_base &>(kernel).jitCode();
+  }
+
+  // Run this SHOTS times
```
1tnguyen: I think this comment needs to be updated.

runtime/nvqir/resourcecounter/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+set(LIBRARY_NAME nvqir-resource-counter)
+set(INTERFACE_POSITION_INDEPENDENT_CODE ON)
+
+add_library(${LIBRARY_NAME} SHARED ResourceCounter.cpp)
+set_property(GLOBAL APPEND PROPERTY CUDAQ_RUNTIME_LIBS ${LIBRARY_NAME})
+
+set (TRACER_DEPENDENCIES libstim fmt::fmt-header-only cudaq-common)
```
1tnguyen: I wonder why we need `libstim` dependency here.

runtime/nvqir/resourcecounter/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+target_link_libraries(${LIBRARY_NAME}
+  PRIVATE ${TRACER_DEPENDENCIES})
+
+set_target_properties(${LIBRARY_NAME}
+    PROPERTIES INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${LLVM_BINARY_DIR}/lib")
+
+install(TARGETS ${LIBRARY_NAME} DESTINATION lib)
+
+add_target_config(resource-counter)
```
1tnguyen: Q: Since we're adding `nvqir-resource-counter` as part of NVQIR lib, do we still need to make it an independent target?

runtime/nvqir/resourcecounter/ResourceCounter.cpp
```diff
@@ -0,0 +1,21 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "ResourceCounter.h"
+#include "nvqir/CircuitSimulator.h"
+
+#ifndef __NVQIR_ResourceCounter_TOGGLE_CREATE
+/// Register this Simulator with NVQIR.
+NVQIR_REGISTER_SIMULATOR(nvqir::ResourceCounter, resourcecounter)
```
1tnguyen: I'd recommend remove this registration. This will just add C-API hooks for `getCircuitSimulator` and we're only supposed to expose this `ResourceCounter` via the `__nvqir__getResourceCounterCircuitSimulator`

runtime/nvqir/resourcecounter/ResourceCounter.cpp
```diff
@@ -0,0 +1,21 @@
...
+#endif
+
+extern "C" {
+nvqir::CircuitSimulator *__nvqir__getResourceCounterCircuitSimulator() {
+  return new nvqir::ResourceCounter();
```
1tnguyen: Q: Could this be called multiple times, e.g., counting resources for multiple kernels? Do we want to make it a thread-local static instance instead?

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+namespace nvqir {
+
+class ResourceCounter : public nvqir::CircuitSimulatorBase<double> {
+protected:
+  cudaq::resource_counts resourceCounts;
+  std::function<bool()> choice;
+
+  /// @brief Grow the state vector by one qubit.
+  void addQubitToState() override {
+    // executionContext->resourceCounts.addQubit();
```
1tnguyen: nit: remove the comment if unused

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+    resourceCounts.addQubit();
+  }
+
+  void applyGate(const GateApplicationTask &task) override {
+    cudaq::info("Applying {} with {} controls", task.operationName,
```
1tnguyen: nit: we should use the more efficient `CUDAQ_INFO` macro.

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+                task.controls.size());
+    auto gate = cudaq::resource_counts::GateData{task.operationName,
+                                                 task.controls.size()};
+    // executionContext->resourceCounts.append(gate);
```
1tnguyen: nit: remove unused comment

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+    resourceCounts.append(gate);
+  }
+
+  /// @brief Measure the qubit and return the result. Collapse the
+  /// state vector.
+  bool measureQubit(const std::size_t index) override {
+    // assert(executionContext->choice);
+    // auto measure = executionContext->choice();
```
1tnguyen: nit: remove commented code

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+  /// @param index 0-based index of qubit to reset
+  void resetQubit(const std::size_t index) override {}
+
+  /// @brief Sample the multi-qubit state.
+  cudaq::ExecutionResult sample(const std::vector<std::size_t> &qubits,
+                                const int shots) override {
+    throw std::runtime_error("Can't sample from resource counter simulator!");
+  }
+
+  std::string name() const override { return "resourcecounter"; }
```
1tnguyen, line 66: I guess we can use `resource-counter` for readability

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -0,0 +1,92 @@
...
+    // deconstructor.
+    summaryData.name = name();
+  }
+  virtual ~ResourceCounter() = default;
+
+  bool canHandleObserve() override { return false; }
+
+  /// @brief Reset the qubit
+  /// @param index 0-based index of qubit to reset
+  void resetQubit(const std::size_t index) override {}
```
1tnguyen: Q: should we consider `reset` a gate instruction?

runtime/nvqir/resourcecounter/resource-counter.yml
1tnguyen, line 1: Same as above: do we need to create a target for this resource counter as it's added as part of nvqir linking?

runtime/nvqir/resourcecounter/ResourceCounter.h
atgeller: It doesn't hurt to include `reset` (and measurements for that matter, which currently aren't tracked). They may become useful later on.

runtime/nvqir/resourcecounter/ResourceCounter.cpp
atgeller: I'm wondering if we want the resource counter simulator to be a separate library at all then, and instead attach it to NVQIR, since NVQIR will rely on the thread local ResourceCounter simulator, and therefore it should always be accessible by NVQIR.

atgeller: An alternative approach would be to move `switchToResourceCounterSimulator` and `stopUsingResourceCounterSimulator` to the ResouceCounter library, which would use the NVQIR functions to switch the simulators back and forth, but that approach seems a little messier to me than the current approach of having both simulators available in NVQIR, and choosing which one to use.

1tnguyen: > attach it to NVQIR, since NVQIR will rely on the thread local ResourceCounter simulator, and therefore it should always be accessible by NVQIR.

This is a good idea. As `ResourceCounter` is a lightweight class, it makes sense to add this to NVQIR rather than having it in a separate lib.

atgeller:
> /ok to test 9d0b4a6eeeb68929c53479e0dd582e5e82cc121a

runtime/nvqir/resourcecounter/CMakeLists.txt
1tnguyen, line 1: I assume we can remove this file :)

bmhowe23:
> /ok to test 9d0b4a6eeeb68929c53479e0dd582e5e82cc121a

mitchdz:
> /ok to test 39dd3b2

mitchdz:
> /ok to test 65757f8

mitchdz:
> /ok to test 838f82e

bettinaheim:
> /ok to test 62895510dd2c1dc8a2950ffd783fc566e698770b

bettinaheim:
> /ok to test 59b6a1f

bettinaheim:
> /ok to test e3110df

anuragksv:
This is fantastic work and much needed in [cuda-quantum](https://github.com/NVIDIA/cuda-quantum), excited to see [PR #3051](https://github.com/NVIDIA/cuda-quantum/pull/3051) land in `main` soon!

I'm also curious: are there any plans to expose or integrate this resource counting functionality (without requiring full quantum simulation) into the CUDAQ Python API? Resource estimation in Python would be incredibly valuable for developing and analyzing current CUDA-Q variational workflows like VQE and QAOA, as well as future FTQC pipelines developed directly in Python.

Also, while my familiarity with CUDAQ is still growing, I’d be happy to help or contribute in any way I can if there’s a plan to bring this to the Python side and beyond.

Thanks for leading this effort, @atgeller, @annagrin, @bmhowe23, @mitchdz, @bettinaheim!

mitchdz:
> /ok to test fa09d87

bettinaheim:
> /ok to test 941dc43

atgeller:
> I'm also curious: are there any plans to expose or integrate this resource counting functionality (without requiring full quantum simulation) into the CUDAQ Python API?

@anuragksv the PR includes a python binding for the `estimate_resources` function. The documentation update is still pending, but you can find an example use in `python/tests/backends/test_resource_counter.py`.

bettinaheim:
> /ok to test f099ae5

bettinaheim:
> /ok to test d7a4870

atgeller:
> /ok to test 56c636d

atgeller:
> /ok to test ffefb72

atgeller:
> /ok to test c0940a0

boschmitt:
Why add resource counter as a simulator? Your change already goes through the trouble of analyzing the IR with a compiler pass, so why that pass isn't counting the resources and returning the estimate? Doing so will make this PR considerably simpler. Indeed, there a many moving pieces here: modification of the IR, a simulator that doesn't simulate, this idea of choice function, etc.

atgeller:
> /ok to test 1e6bf2e

bettinaheim:
> /ok to test db642ef

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
```diff
@@ -0,0 +1,104 @@
...
+    if (!isQuakeOperation(op))
+      return false;
+
+    auto opi = dyn_cast<quake::OperatorInterface>(op);
+
+    if (!opi)
+      return false;
+
+    // Measures may affect control flow, don't remove for now
+    if (isa<quake::MzOp>(op) || isa<quake::MyOp>(op) || isa<quake::MxOp>(op))
```
schweitzpgi, line 49: This should check for the measurement trait.
```c++
op->hasTrait<cudaq::QuantumMeasure>()
```

Issue #3042, opened 2025-06-11
[C++] [run] Returning Boolean value gives error


khalatepradnya:
### Describe the bug

Following kernel which returns a Boolean value does not compile

### Steps to reproduce the bug

```
#include <cudaq.h>

auto bool_test = [](bool b) __qpu__ {
  return b;
};

int main() {
  int c = 0;
{
    const auto results = cudaq::run(2, bool_test, true);
    if (results.size() != 2) {
      printf("FAILED! Expected 2 shots. Got %lu\n", results.size());
    } else {
      c = 0;
      for (auto i : results)
        printf("%d: %s\n", c++, i ? "true" : "false");
      printf("success!\n");
    }
  }
  return 0;
}
```

Compilation fails -
```
# bin/nvq++ -fenable-cudaq-run t.cpp
In file included from t.cpp:1:
In file included from /workspaces/cuda-quantum/runtime/cudaq.h:326:
/workspaces/cuda-quantum/runtime/cudaq/algorithms/run.h:133:11: error: type 'ResultTy *' (aka 'bool *') cannot be narrowed to 'bool' in initializer list [-Wc++11-narrowing]
  return {reinterpret_cast<ResultTy *>(span.data),
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
t.cpp:69:33: note: in instantiation of function template specialization 'cudaq::run<(lambda at t.cpp:9:18) &, bool>' requested here
    const auto results = cudaq::run(2, bool_test, true);
                                ^
/workspaces/cuda-quantum/runtime/cudaq/algorithms/run.h:133:11: note: insert an explicit cast to silence this issue
  return {reinterpret_cast<ResultTy *>(span.data),
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          static_cast<bool>(                     )
/workspaces/cuda-quantum/runtime/cudaq/algorithms/run.h:134:11: error: type 'ResultTy *' (aka 'bool *') cannot be narrowed to 'bool' in initializer list [-Wc++11-narrowing]
          reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #3053, opened 2025-06-12
[core] Workaround overly excitable MLIR inliner.

### LINKED ISSUE #3037: [C++] [run] Kernel not returning expected values

schweitzpgi:
Fix #3037.

We generally want the inliner to be super aggressive and inline all calls everywhere. Except in the case where the thunk layer is calling the kernel proper. In that one instance, inlining must be prevented. The existing checks do a decent job of preventing this from happening in the typical case, but there were some cases (involving trivial kernels) that leaked through.

This adds a special NoInlineCallOp to fool the Inliner into thinking it can't perform the inlining.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/CC/CCOps.td
+++ b/include/cudaq/Optimizer/Dialect/CC/CCOps.td
@@ -1759,6 +1759,46 @@ def cc_CallableClosureOp : CCOp<"callable_closure", [Pure]> {
   }];
 }
 
+def cc_NoInlineCallOp :
+    CCOp<"noinline_call", [CallOpInterface, SymbolUserOpInterface]> {
+  let summary = "Create a call that the inliner will not inline.";
+  let description = [{
+    The MLIR inliner in LLVM 16 is too smart and will inline trivial calls.
+    Moreover, there is no mechanism in this version of MLIR to stop the inliner
+    from inlining these trivial functions.
+
+    The cc.noinline_call is therefore a special op that prevents the inliner
+    from doing the inlining by misreporting what function is called.
+  }];
+
+  let arguments = (ins
+    FlatSymbolRefAttr:$callee,
+    Variadic<AnyType>:$args
+  );
+  let results = (outs Variadic<AnyType>);
+
+  let assemblyFormat = [{
+    $callee `(` $args `)` `:` functional-type(operands, results) attr-dict
+  }];
+
+  let extraClassDeclaration = [{
+    operand_range getArgOperands() {
+      return {arg_operand_begin(), arg_operand_end()};
+    }
+
+    operand_iterator arg_operand_begin() { return operand_begin(); }
+    operand_iterator arg_operand_end() { return operand_end(); }
+
+    /// DO NOT RETURN the callee of this operation. This fools the inliner into
+    /// not knowing what is actually called.
+    mlir::CallInterfaceCallable getCallableForCallee() {
+      return mlir::FlatSymbolRefAttr::get(getContext(), ".no.inline");
+    }
+
+    mlir::LogicalResult verifySymbolUses(mlir::SymbolTableCollection &);
+  }];
+}
+
 def cc_VarargCallOp :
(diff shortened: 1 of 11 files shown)


include/cudaq/Optimizer/Dialect/CC/CCOps.td
```diff
@@ -1759,6 +1759,46 @@ def cc_CallableClosureOp : CCOp<"callable_closure", [Pure]> {
   }];
 }
 
+def cc_NoInlineCallOp :
+    CCOp<"noinline_call", [CallOpInterface, SymbolUserOpInterface]> {
+  let summary = "Create a call that the inliner will not inline.";
+  let description = [{
+    The MLIR inliner in LLVM 16 is too smart and will inline trivial calls.
+    Moreover, there is no mechanism in this version of MLIR to stop the inliner
```
khalatepradnya, line 1767: Q: Is there a mechanism in later versions of LLVM (20+)?

targettests/execution/trivial_inliner.cpp
```diff
@@ -0,0 +1,42 @@
...
+      printf("success!\n");
+    }
+  }
+  return 0;
+}
+
+// CHECK: 0: {true, 654, 9.123000}
+// CHECK: 1: {true, 654, 9.123000}
+// CHECK: 2: {true, 654, 9.123000}
+// CHECK: success!
```
khalatepradnya, line 42: Yay!

targettests/execution/trivial_inliner.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cudaq.h>
+
+struct MyTuple {
+  bool boolVal;
+  std::int64_t i64Val;
+  double f64Val;
+};
+
+auto struct_test = [](MyTuple t) __qpu__ { return t; };
```
khalatepradnya, line 19: Is it worth expanding this to other trivial kernels with primitive types?
```
auto int_test = [](int i) __qpu__ { return i; };
auto float_test = [](float f) __qpu__ { return f; };
auto bool_test = [](bool b) __qpu__ { return b; };
```

(Bool one would need fix for #3042)

khalatepradnya (approved):
LGTM. Thanks, Eric!
Suggest expanding test.

include/cudaq/Optimizer/Dialect/CC/CCOps.td
schweitzpgi: I didn't try it, but there are more ways to control inlining, yes.

targettests/execution/trivial_inliner.cpp
schweitzpgi: Created #3057

Issue #843, opened 2023-10-31
sample takes in a noise model but sample_async does not


zohimchandani:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Sample accepts a noise model - see [here](https://nvidia.github.io/cuda-quantum/latest/api/languages/python_api.html#cudaq.sample_async:~:text=1000.%20Key%2Dword%20only.-,noise_model%20(Optional%5BNoiseModel%5D)%20%E2%80%93%20The%20optional%20NoiseModel%20to%20add%20noise%20to%20the%20kernel%20execution%20on%20the%20simulator.%20Defaults%20to%20an%20empty%20noise%20model.,-Returns)

Sample Async does not accept a noise model - see [here](https://nvidia.github.io/cuda-quantum/latest/api/languages/python_api.html#cudaq.sample_async:~:text=cudaq.sample_async(),%C2%B6)

Is it just missing in the documentation or a missing feature? I am not sure.

Thanks team. 

### Steps to reproduce the bug

NA

### Expected behavior

NA

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

sacpis:
Seems like this was just missing in doc. Will be fixed with PR #4383.

PR #3054, opened 2025-06-12
[Python] Expose noise model to the asynchronous sample API

khalatepradnya:
Addressing https://github.com/NVIDIA/cuda-quantum/issues/843


1tnguyen (approved):
LGTM 👍

schweitzpgi:
Is this a problem in C++ as well? We will need to track that issue, if it is an issue, as this doesn't address C++.

khalatepradnya:
> Is this a problem in C++ as well? We will need to track that issue, if it is an issue, as this doesn't address C++.

C++ uses sample options which expose noise model.

Issue #3013, opened 2025-06-06
[cudaq.run] Current cudaq.run issues


annagrin:
Issues:

- [ ] https://github.com/NVIDIA/cuda-quantum/issues/2963
  (ReturnToOutputLog only handles constant lists)
  - Anna is working on it

- [ ] https://github.com/NVIDIA/cuda-quantum/issues/2965
  (incorrect quake for updating elements of struct)
  https://github.com/NVIDIA/cuda-quantum/pull/3002

- [ ] https://github.com/NVIDIA/cuda-quantum/issues/2966
   (tuples are immutable)
    https://github.com/NVIDIA/cuda-quantum/pull/3000

- [ ] https://github.com/NVIDIA/cuda-quantum/issues/2984
  - Anna is working on it

- [ ] https://github.com/NVIDIA/cuda-quantum/issues/3012

Quantum only:
- [ ] https://github.com/NVIDIA/cuda-quantum/issues/2964
	Full solution is still in workings, for now disable returning vectors on quantum devices for current release
	- [ ] for cpp:
	https://github.com/NVIDIA/cuda-quantum/pull/3010
	- [ ] for python

PR #3059, opened 2025-06-13
[Python] [run] Disallow returing `list` from entry-point kernels only for now

khalatepradnya:
Towards https://github.com/NVIDIA/cuda-quantum/issues/3013


bmhowe23 (approved):
LGTM

Issue #3057, opened 2025-06-13
[test] expanding inlining test to trivial kernels with other primitive types


schweitzpgi:
Is it worth expanding this to other trivial kernels with primitive types?
```
auto int_test = [](int i) __qpu__ { return i; };
auto float_test = [](float f) __qpu__ { return f; };
auto bool_test = [](bool b) __qpu__ { return b; };
```

(Bool one would need fix for #3042)

_Originally posted by @khalatepradnya in https://github.com/NVIDIA/cuda-quantum/pull/3053#discussion_r2143765686_

PR #3061, opened 2025-06-13
[C++] [run] Enable returning Boolean value from kernels

khalatepradnya:
Fix bug #3042 
Also addresses #3057

### DIFF
--- a/runtime/cudaq/algorithms/run.h
+++ b/runtime/cudaq/algorithms/run.h
@@ -191,9 +191,15 @@ run(std::size_t shots, cudaq::noise_model &noise_model, QuantumKernel &&kernel,
       [&]() mutable { kernel(std::forward<ARGS>(args)...); }, platform,
       kernelName, shots);
   platform.reset_noise();
-
-  return {reinterpret_cast<ResultTy *>(span.data),
-          reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
+  if constexpr (std::is_same_v<ResultTy, bool>) {
+    // Special case for std::vector<bool> which is a specialization.
+    std::vector<bool> results;
+    __nvqpp_initializer_list_to_vector_bool(results, span.data,
+                                            span.lengthInBytes);
+    return results;
+  } else
+    return {reinterpret_cast<ResultTy *>(span.data),
+            reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
 }
 
 /// @brief Launch a run of a kernel for \p shots number of times on a specific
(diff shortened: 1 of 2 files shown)


runtime/cudaq/algorithms/run.h
```diff
@@ -184,9 +191,15 @@ run(std::size_t shots, cudaq::noise_model &noise_model, QuantumKernel &&kernel,
...
-          reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
+  if constexpr (std::is_same_v<ResultTy, bool>) {
+    // Special case for std::vector<bool> which is a specialization.
+    std::vector<bool> results;
+    __nvqpp_initializer_list_to_vector_bool(results, span.data,
+                                            span.lengthInBytes);
+    return results;
+  } else
+    return {reinterpret_cast<ResultTy *>(span.data),
+            reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
```
1tnguyen: It looks like we could potentially use `resultSpanToVectorViaOwnership` in these places to reduce code duplication.

One potential hiccup is this `static_assert`
```
static_assert(sizeof(std::vector<T>) == sizeof(raw_vector) &&
                "std::vector must use the nominal 3 pointer implementation");
```
which may not be true for `vector<bool>`, i.e., should be moved to inside the `else` block.

cc @schweitzpgi

runtime/cudaq/algorithms/run.h
```diff
@@ -130,8 +130,15 @@ run(std::size_t shots, QuantumKernel &&kernel, ARGS &&...args) {
...
```
schweitzpgi:
```suggestion
  } else {
    return {reinterpret_cast<ResultTy *>(span.data),
            reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
  }
```
nit: `if constexpr` is a bit of an exception since only 1 code block will actually exist in an instance (not both like a normal `if else`).

runtime/cudaq/algorithms/run.h
schweitzpgi:
```suggestion
  } else {
    return {reinterpret_cast<ResultTy *>(span.data),
            reinterpret_cast<ResultTy *>(span.data + span.lengthInBytes)};
  }
```

schweitzpgi: Agree with the removing of code duplication. Let's do that.

Also, you're correct that the assertion is wrong for `std::vector<bool>`. That specialization does not use the same internal layout as other vectors.

khalatepradnya: Fixed in 45f45f160167dec9afccb85139dfd0e9cf3825cb

1tnguyen (approved):
LGTM 👍

Issue #2984, opened 2025-06-02
Update py_async_run.cpp to use `pyLaunchKernel` (like in py_run.cpp) and pass the correct argument type


annagrin:
This needs to be updated to use `pyLaunchKernel` (like in py_run.cpp) and pass the correct argument type

_Originally posted by @annagrin in https://github.com/NVIDIA/cuda-quantum/pull/2927#discussion_r2122068654_

PR #3072, opened 2025-06-17
Update run_async to work with python API and add tests

annagrin:
### Description
Update `cudaq.run_async` API in python to use python-specific launch API and return python objects.

- Solve seg faults when running kernels with no arguments
- Support returning primitive types and data classes
- Will support tuples and lists when those are enabled
- Propagate errors to the python code
- Add tests

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2984


annagrin:
replaced by https://github.com/NVIDIA/cuda-quantum/pull/3074

PR #3074, opened 2025-06-17
Update run_async to work with python API and add tests

annagrin:
### Description
Update `cudaq.run_async` API in python to use python-specific launch API and return python objects.

- Solve seg faults when running kernels with no arguments
- Support returning primitive types and data classes
- Will support tuples and lists when those are enabled
- Propagate errors to the python code
- Add tests

Closes: https://github.com/NVIDIA/cuda-quantum/issues/2984


python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -122,10 +125,108 @@ std::vector<py::object> pyRun(py::object &kernel, py::args args,
...
+      [p = std::move(promise), resultPtr = result.results,
+       errorPtr = result.error, shots_count, argData, name, module, func,
+       noise_model = std::move(noise_model)]() mutable {
+        auto &platform = get_platform();
+
+        // Launch the kernel in the appropriate context.
+        if (noise_model.has_value())
+          platform.set_noise(&noise_model.value());
+
+        py::gil_scoped_acquire gil{};
```
1tnguyen: I wonder whether this GIL acquire would cause all the kernel executions (on different GPU-emulated QPUs) sequential.
We'd want these `QuantumTask` functors to run independently on threads.

annagrin: Yeah I think you are right. I have to hold the GIL for converting results to python objects, but I can try splitting out the c++ part out. Then only the conversion will need to be synchronized on…

python/cudaq/runtime/run.py
```diff
@@ -30,15 +31,16 @@ def run_async(kernel, *args, shots_count=1, noise_model=None, qpu_id=0):
...
   """
-    kernel.enable_return_to_log()
     if kernel.returnType is None:
-        raise ValueError("cudaq.run only supports kernels that return values.")
+        raise RuntimeError(
+            "`cudaq.run` only supports kernels that return a value.")
 
     if shots_count < 0:
-        raise ValueError("Invalid shots_count. Must be non-negative.")
+        raise RuntimeError("Invalid shots_count. Must be non-negative.")
```
khalatepradnya: nit:
```suggestion
        raise RuntimeError("Invalid `shots_count`. Must be non-negative number.")
```

annagrin: thanks, updated!

python/runtime/cudaq/algorithms/py_run.cpp
annagrin: Updated to only acquire the GIL for the translation of the c++ return values to python object (it is required there)

PR #3060, opened 2025-06-13
Remove deprecated backends from the validation script

sacpis:
Remove deprecated backends in order to avoid duplicate testing when validating container in the publishing task.
https://github.com/NVIDIA/cuda-quantum/actions/runs/15632777958/job/44068728736#step:5:117

# Before
```
Testing backends:
dynamics
nvidia-fp64
nvidia-mgpu
nvidia-mqpu-fp64
nvidia-mqpu-mps
nvidia-mqpu
nvidia
remote-mqpu
tensornet-mps
tensornet

=============================
==        C++ Tests        ==
=============================
Testing amplitude_estimation:
Source: applications/cpp/amplitude_estimation.cpp
Skipping dynamics target for applications/cpp/amplitude_estimation.cpp.
Testing on nvidia-fp64 target...
Warning: This target is deprecating. Please use the 'nvidia' target with option 'fp64' by adding the command line option '--target-option fp64' or passing it as cudaq.set_target('nvidia', option='fp64') in Python. Please refer to CUDA-Q 
Exited with code 0
Testing on nvidia-mgpu target...
Warning: This target is deprecating. Please use the 'nvidia' target with option 'mgpu,fp64' by adding the command line option '--target-option mgpu,fp64' or passing it as cudaq.set_target('nvidia', option='mgpu,fp64') in Python. Please refer to CUDA-Q 
Exited with code 0
Testing on nvidia-mqpu-fp64 target...
Warning: This target is deprecating. Please use the 'nvidia' target with option 'mqpu,fp64' by adding the command line option '--target-option mqpu,fp64' or passing it as cudaq.set_target('nvidia', option='mqpu,fp64') in Python. Please refer to CUDA-Q 
Exited with code 0
Skipping nvidia-mqpu-mps target.
Testing on nvidia-mqpu target...
Warning: This target is deprecating. Please use the 'nvidia' target with option 'mqpu,fp32' or 'mqpu' (fp32 is the default precision option) by adding the command line option '--target-option mqpu,fp32' or passing it as cudaq.set_target('nvidia', option='mqpu,fp32') in Python. Please refer to CUDA-Q 
Exited with code 0
Testing on nvidia target...
  Testing nvidia target option: fp32
  Exited with code 0
  Testing nvidia target option: fp64
  Exited with code 0
  Testing nvidia target option: fp32,mqpu
  Exited with code 0
  Testing nvidia target option: fp64,mqpu
  Exited with code 0
  Testing nvidia target option: fp32,mgpu
[...]

PR #3075, opened 2025-06-17
Fix publishing deprecated backend usage

mitchdz:
Fixes failing pipelines - https://github.com/NVIDIA/cuda-quantum/actions/runs/15694498377/job/44218950832

Recently, the deprecated backends were removed from the validation script https://github.com/NVIDIA/cuda-quantum/pull/3060. This has the knock-on effect that the publishing pipeline is now breaking because it was looking for those deprecated backends. 
This updates the logic to :
* use the non-deprecated backend
* check for the non-deprecated backend for `fp32,mgpu`.

If a failure occurs, the pipeline will print the block it sees that failed to find `fp32,mgpu` like so:
```
::error::Missing tests for nvidia fp32,mgpu backend.
Observed target options:
  Testing nvidia target option: fp32
  Exited with code 0
  Testing nvidia target option: fp64
  Exited with code 0
  ...
```

I tested this in a manual way by copying `validation.out` to `/tmp/` and just running the bash code that is changed.

```bash
$ head /tmp/validation.out 
Installed backends:
default
anyon
braket
density-matrix-cpu
dynamics
fermioniq
infleqtion
ionq
iqm
$ cat test.sh
          # Check that the tests included the nvidia-mgpu backend:
          # Grab the first line where "Testing on nvidia target..." starts
          # The first should be okay, since any target being tested against
          # nvidia should be using all the options for the validation script.
          nvidia_block_start=$(grep -n "Testing on nvidia target..." /tmp/validation.out | head -n1 | cut -d: -f1)

          # From that line, extract the block until the next "Testing on" or EOF
          nvidia_block=$(tail -n +$((nvidia_block_start + 1)) /tmp/validation.out | sed -n '/^Testing on /q;p')

          # Check if "fp32,mgpu" is present
          if ! echo "$nvidia_block" | grep -q "fp32,mgpu"; then
            echo "::error::Missing tests for nvidia fp32,mgpu backend."
            echo "Observed target options:"
            echo "$nvidia_block"
            exit 1
          fi

$ bash test.sh 
$ echo $?
0
```

The intro changes were easy, I just ran that in a fresh container.

```
# cat test.sh
          status_sum=0
          set +e # Allow script to keep going through errors
          # Verify that the necessary GPU targets are installed and usable
          # Note nvidia-mgpu requires MPI, so it is not available with this method.
          # List of full target options
          targets=("--target nvidia" "--target nvidia --target-option fp64" "--target tensornet")
          for tgt in "${targets[@]}"; do
              echo "Running with ${tgt}"
              python3 docs/sphinx/examples/python/intro.py $tgt
              if [ $? -ne 0 ]; then 
                  echo -e "\e[01;31mPython trivial test for ${tgt} failed.\e[0m" >&2
                  status_sum=$((status_sum+1))
              fi
          done

          set -e # Re-enable exit code error checking
          if [ "$status_sum" -ne "0" ]; then
            echo "::error::Error running validation script" 
            exit $status_sum
          fi

# file docs/sphinx/examples/python/intro.py
docs/sphinx/examples/python/intro.py: Python script, ASCII text executable
# bash test.sh
Running with --target nvidia
{ 0:487 1:513 }

Running with --target nvidia --target-option fp64
{ 0:501 1:499 }

Running with --target tensornet
{ 0:494 1:506 }

# echo $?
0
```

### DIFF
--- a/.github/workflows/publishing.yml
+++ b/.github/workflows/publishing.yml
@@ -1091,11 +1091,13 @@ jobs:
           set +e # Allow script to keep going through errors
           # Verify that the necessary GPU targets are installed and usable
           # Note nvidia-mgpu requires MPI, so it is not available with this method.
-          for tgt in nvidia nvidia-fp64 tensornet; do
-              echo "Running with target ${tgt}"
-              python3 docs/sphinx/examples/python/intro.py --target ${tgt}
+          # List of full target options
+          targets=("--target nvidia" "--target nvidia --target-option fp64" "--target tensornet")
+          for tgt in "${targets[@]}"; do
+              echo "Running with ${tgt}"
+              python3 docs/sphinx/examples/python/intro.py $tgt
               if [ $? -ne 0 ]; then 
-                  echo -e "\e[01;31mPython trivial test for target ${tgt} failed.\e[0m" >&2
+                  echo -e "\e[01;31mPython trivial test for ${tgt} failed.\e[0m" >&2
                   status_sum=$((status_sum+1))
               fi
           done
--- a/.github/workflows/publishing.yml
+++ b/.github/workflows/publishing.yml
@@ -875,10 +875,17 @@ jobs:
           bash github-repo/scripts/validate_container.sh $backends_to_test | tee /tmp/validation.out
 
           # Check that the tests included the nvidia-mgpu backend:
-          relevant_line=`grep -n "Testing backends:" /tmp/validation.out | cut -d : -f1`
-          tested_backends=`cat /tmp/validation.out | tail -n +$relevant_line | sed -e '/^$/,$d'`
-          if [ -z "$(echo $tested_backends | grep nvidia-mgpu)" ]; then 
-            echo "::error::Missing tests for nvidia-mgpu backend."
+          # Grab the line where "Testing on nvidia target..." starts
+          nvidia_block_start=$(grep -n "Testing on nvidia target..." /tmp/validation.out | cut -d: -f1)
+
+          # From that line, extract the block until the next "Testing on" or EOF
+          nvidia_block=$(tail -n +$((nvidia_block_start + 1)) /tmp/validation.out | sed -n '/^Testing on /q;p')
+
+          # Check if "fp32,mgpu" is present
+          if ! echo "$nvidia_block" | grep -q "fp32,mgpu"; then
+            echo "::error::Missing tests for nvidia fp32,mgpu backend."
+            echo "Observed target options:"
+            echo "$nvidia_block"
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


sacpis:
> /ok to test 65610c0

sacpis:
Just a question. Do we have a node with a gpu (or multiple gpus) when publshing?

mitchdz:
Yes, just a single gpu. You can see the runner here

https://github.com/NVIDIA/cuda-quantum/actions/runs/15694498377/job/44218950832#step:1:2

No multi-gpu support (yet, coming soon hopefully).

sacpis:
> Yes, just a single gpu. You can see the runner here
> 
> https://github.com/NVIDIA/cuda-quantum/actions/runs/15694498377/job/44218950832#step:1:2
> 
> No multi-gpu support (yet, coming soon hopefully).

Got it. Let's just remove this condition for multi-gpu check. I think for now that is a better fix.

.github/workflows/publishing.yml
```diff
@@ -875,10 +875,19 @@ jobs:
...
+          # The first should be okay, since any target being tested against
+          # nvidia should be using all the options for the validation script.
+          nvidia_block_start=$(grep -n "Testing on nvidia target..." /tmp/validation.out | head -n1 | cut -d: -f1)
+
+          # From that line, extract the block until the next "Testing on" or EOF
+          nvidia_block=$(tail -n +$((nvidia_block_start + 1)) /tmp/validation.out | sed -n '/^Testing on /q;p')
+
+          # Check if "fp32,mgpu" is present
+          if ! echo "$nvidia_block" | grep -q "fp32,mgpu"; then
+            echo "::error::Missing tests for nvidia fp32,mgpu backend."
```
bmhowe23, line 888: Do we need anything for fp64,mgpu? Isn't that the default mgpu configuration?

mitchdz:
I'm a little hesitant to remove it, perhaps in the past a non-gpu runner would accidentally get picked up and publishing would silently pass without doing anything. Another thing this check could be doing is ensuring that MPI can be installed alongside cudaq. This line of change came from https://github.com/NVIDIA/cuda-quantum/pull/2356 and unfortunately doesn't give insight into what was breaking or going on without this check.

@bettinaheim Do you recall what this code block was for? I'm curious if this is something we can simple remove.

.github/workflows/publishing.yml
mitchdz: The old block just checked for `nvidia-mgpu` which could be the default fp32.

I did this check simply to keep in parity of the old behavior.

To be quite honest I'm not sure if this whole check is really necessary. I asked Bettina in a comment below if she recalls why this check exists, perhaps it's so safeguard against bad runners or bad MPI installations.

mitchdz:
Per conversation, with @bettinaheim , this check was originally added to do a minimal "confirm that nvidia-mgpu is included in the installer, and can be used for very basic check."

Optimally, this should hard code check _every_ target and option.

.github/workflows/publishing.yml
mitchdz: Wrote more about it in below comment https://github.com/NVIDIA/cuda-quantum/pull/3075#issuecomment-2981388698

mitchdz:
Let's leave the check in, and make a ticket for improving it. I'd like to focus on fixing deployments, and don't want to leave publishing broken right now.

bmhowe23 (approved):

PR #3078, opened 2025-06-18
Update buildkit to use nvidia docker cache

mitchdz:
buildkit does not use the nvidia DockerHub cache by default. Pass through the configuration file only on our self-hosted runners.


mitchdz:
> /ok to test e579fec

bmhowe23 (approved):
👍

PR #3083, opened 2025-06-18
[core] Fixes a number of bugs, TODOs, etc.

schweitzpgi:
- Reorder to passes to avoid silly inefficiency.
- Fix a bug for when the lookup fails.
- Add guard to avoid crash.
- Reorganize apply specialization to prop constants before generating control, adjoint, both functions rather than after.


lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -75,20 +76,67 @@ struct ApplyOpAnalysis {
...
+            if (auto c = v.getDefiningOp<arith::ConstantOp>()) {
+              auto newConst = c.clone();
+              moveConsts.push_back(newConst);
+              mapper.map(genericFunc.getArgument(idx), newConst);
+              LLVM_DEBUG(llvm::dbgs() << "apply has constant arguments.\n");
+              arithConsts = true;
+            }
+          }
+
+          if (arithConsts) {
```
1tnguyen: I wonder if we can get rid of `arithConsts` and use `!moveConsts.empty()` instead.

PR #3094, opened 2025-06-18
[run] [parser] Throw error if missing data layout info when parsing struct / data class

khalatepradnya:
The records log parser should not assume packed data layout - it leads to erroneous behavior.


1tnguyen (approved):
👍

Issue #3095, opened 2025-06-19
[python, run_async] Async_kernel test is segfaulting and blocking all PRs


sacpis:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

test_run_async_kernel test is segfaulting when it runs in docker with CPU only. The issue is related to kernels with no args.

https://github.com/NVIDIA/cuda-quantum/actions/runs/15744558392/job/44381105194?pr=3069#step:9:966

Seems like the tests were added in PR #[3074](https://github.com/NVIDIA/cuda-quantum/pull/3074).

The error is being thrown from

```
def test_return_bool():

    @cudaq.kernel
    def simple_bool_no_args() -> bool:
        return True

    results = cudaq.run_async(simple_bool_no_args, shots_count=2).get()
    assert len(results) == 2
    assert results[0] == True
    assert results[1] == True

    @cudaq.kernel
    def simple_bool(numQubits: int) -> bool:
        qubits = cudaq.qvector(numQubits)
        return True

    results = cudaq.run_async(simple_bool, 2, shots_count=2).get()
    assert len(results) == 2
    assert results[0] == True
    assert results[1] == True
```
from this line
```
results = cudaq.run_async(simple_bool_no_args, shots_count=2).get()
```

### Steps to reproduce the bug

Run the test_run_async_kernel in CPU only docker container.

### Expected behavior

Should run sucessfully.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

khalatepradnya:
These should now work with PR #3191 ; async, noisy tests still have issues, captured them in a new task here: https://github.com/NVIDIA/cuda-quantum/issues/3193

PR #3096, opened 2025-06-19
Disabling run_async tests to unblock CI

sacpis:
Disabling run_async tests to unblock CI.

Kernels with no args are having segfault and other tests are failing for different reasons.

Please see issue #3095.


1tnguyen (approved):
Thanks, @sacpis.

PR #3106, opened 2025-06-24
[C++ bridge] Emit an error for bad storage classes.

schweitzpgi:
Kernels may not contain static and extern declared variables.


1tnguyen (approved):
👍

PR #3107, opened 2025-06-24
[C++ bridge] Fix bug with injected func.return.

schweitzpgi:
Need to support a richer set of function signatures when adding the default func.return to a function which only has unwind_return ops.


1tnguyen (approved):
LGTM 👍

PR #3109, opened 2025-06-24
[C++] Common logic to identify kernel name from kernels that are free functions

khalatepradnya:


targettests/execution/cudaq_run_free_functions.cpp
```diff
@@ -0,0 +1,82 @@
...
+// clang-format on
+
+#include <cudaq.h>
+
+struct foo_test_result {
+  int64_t a;
+  int64_t b;
+};
+
+foo_test_result foo_test(int x) __qpu__ {
```
bmhowe23, line 21: What is the point of the `int x` parameter in these tests? They look unused to me. (Same for line 34 and line 49 as well.)

khalatepradnya: No particular reason, simply used examples from the bug bash. I've updated these in 34e2886aee2030d8037bdb87e69cbcce74992e81, one test uses the value and removed from others.

runtime/cudaq/algorithms/get_kernel_name.h
```diff
@@ -0,0 +1,69 @@
+/****************************************************************-*- C++ -*-****
```
amccaskey: this is broader than algorithms, so I would suggest not putting it here (maybe in utils or qis). I also would suggest something more generic like `kernel_utils.h`.

khalatepradnya: Changed in bad6c9e9d2a5bfbcff48acfe7e5d11f16c341436

bmhowe23 (approved):
LGTM

runtime/cudaq/algorithms/run.h
```diff
@@ -125,7 +126,7 @@ run(std::size_t shots, QuantumKernel &&kernel, ARGS &&...args) {
 #endif
   // Launch the kernel in the appropriate context.
   auto &platform = cudaq::get_platform();
-  std::string kernelName{cudaq::getKernelName(kernel)};
+  std::string kernelName{details::getKernelName(kernel)};
```
schweitzpgi, line 129: Aside: weren't we going to converge on `detail` (without the `s`)? Might want to do that here?

runtime/cudaq/qis/kernel_utils.h
```diff
@@ -0,0 +1,69 @@
...
+
+#pragma once
+
+#include "common/ExecutionContext.h"
+#include "cudaq/concepts.h"
+#include "cudaq/utils/registry.h"
+#include "qkernel.h"
+
+namespace cudaq {
+namespace details {
```
schweitzpgi:
```suggestion
namespace cudaq::detail {
```

runtime/cudaq/algorithms/run.h
khalatepradnya: It was decided to wait until refactoring of runtime, hence, sticking with existing namespaces.

runtime/cudaq/qis/kernel_utils.h
khalatepradnya: Updated in 2854b20 (keeping `details`)

PR #3113, opened 2025-06-25
[readability] Update warning for applyNoise

### LINKED ISSUE #3088: [python, run] Give a warning if noise model is not supported on a target
### Required prerequisites

### Describe the bug

On targets other than `density-matrix-cpu` the noise model is ignored, which leads to unexpected results.

```python
@cudaq.kernel
def foo() -> int:
    q = cudaq.qvector(8)
    x(q)
    retval = 0
    for i in range(8):
        bit = mz(q[i])
        retval = retval * 2
        if bit == 0:
            retval = retval + 0
        else:
            retval = retval + 1
    return retval

noise = cudaq.NoiseModel()
depolar_noise = cudaq.DepolarizationChannel(0.9)
noisy_ops = ["z", "s", "x", "h", "rx", "rz"]
for op in noisy_ops:
    noise.add_all_qubit_channel(op, depolar_noise)

print(cudaq.run(foo, noise_model=noise, shots_count=20))
```

All shots give the same result.

### Expected behavior

A warning if noise model is ignored on a target. 

Setting the following target will make sure the noise model is not ingored:
```
cudaq.set_target("density-matrix-cpu")
```

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

nvidia-dobri:
Closes #3088

Updates the wording for the warning given when a circuit simulator doesn't support noise.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -262,8 +262,7 @@ public:
   /// Only supported for noise backends. By default do nothing
   virtual void applyNoise(const cudaq::kraus_channel &channel,
                           const std::vector<std::size_t> &targets) {
-    CUDAQ_WARN("kraus_channel application not supported on {} simulator.",
-               name());
+    CUDAQ_WARN("Applying noise is not supported on {} simulator.", name());
   }
 
   /// @brief Apply a custom operation described by a matrix of data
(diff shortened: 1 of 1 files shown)


1tnguyen:
We may want to add a warning to `applyNoiseChannel` as well:

https://github.com/NVIDIA/cuda-quantum/blob/736d519116a909622b574c4c0ff13eaf44aa8383/runtime/nvqir/CircuitSimulator.h#L841

PR #3121, opened 2025-06-26
Restore error reporting under non-composable option.

schweitzpgi:
This restores some of the error reporting functionality that was baked into the mapping pass. The user must explicitly set a flag to make these passes non-composable.

TODO: finish removing the non-compositional aspects of these passes in general.


1tnguyen (approved):
LGTM 👍

PR #3123, opened 2025-06-27
[python, ast] Allow AST to handle property access on calls

### LINKED ISSUE #3089: [python, run] Constructing a dataclass and accessing its field in return statement does not compile.
### Required prerequisites

### Describe the bug

Constructing a dataclass and accessing its field in return statement does not compile.

```python
@dataclass
class M:
    i:int
    j:int

@cudaq.kernel
def bugSP3() -> int:
    return M(6, 9).i

print(cudaq.run(bugSP3, shots_count = 2))
```

Output
```
error: 'func.return' op has 0 operands, but enclosing function (@__nvqpp__mlirgen__bugSP3) returns 1
```

### Expected behavior

Output:
[6, 6]

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

nvidia-dobri:
Closes #3089 

The AST Bridge doesn't currently handle accessing properties on a call result, e.g. `foobar().prop`. This allows it to handle those cases.

The debug example that did not work previously (from issue):
```
@dataclass
class M:
    i:int
    j:int

@cudaq.kernel
def bugSP3() -> int:
    return M(6, 9).i

print(cudaq.run(bugSP3, shots_count = 2))
```

This works now, and returns the expected result.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -153,6 +153,7 @@ AsyncObserveResult = cudaq_runtime.AsyncObserveResult
 AsyncStateResult = cudaq_runtime.AsyncStateResult
 vqe = cudaq_runtime.vqe
 draw = cudaq_runtime.draw
+get_unitary = cudaq_runtime.get_unitary
 run = cudaq_runtime.run
 translate = cudaq_runtime.translate
 displaySVG = display_trace.displaySVG
--- a/python/extension/CMakeLists.txt
+++ b/python/extension/CMakeLists.txt
@@ -59,6 +59,7 @@ declare_mlir_python_extension(CUDAQuantumPythonSources.Extension
     ../runtime/cudaq/algorithms/py_state.cpp
     ../runtime/cudaq/algorithms/py_evolve.cpp
     ../runtime/cudaq/algorithms/py_translate.cpp
+    ../runtime/cudaq/algorithms/py_unitary.cpp
     ../runtime/cudaq/algorithms/py_utils.cpp
     ../runtime/cudaq/algorithms/py_vqe.cpp
     ../runtime/cudaq/platform/JITExecutionCache.cpp
--- a/python/extension/CUDAQuantumExtension.cpp
+++ b/python/extension/CUDAQuantumExtension.cpp
@@ -25,6 +25,7 @@
 #include "runtime/cudaq/algorithms/py_sample_async.h"
 #include "runtime/cudaq/algorithms/py_state.h"
 #include "runtime/cudaq/algorithms/py_translate.h"
+#include "runtime/cudaq/algorithms/py_unitary.h"
 #include "runtime/cudaq/algorithms/py_utils.h"
 #include "runtime/cudaq/algorithms/py_vqe.h"
 #include "runtime/cudaq/operators/py_boson_op.h"
@@ -118,6 +119,7 @@ PYBIND11_MODULE(_quakeDialects, m) {
   cudaq::bindPyEvolve(cudaqRuntime);
   cudaq::bindEvolveResult(cudaqRuntime);
   cudaq::bindPyDraw(cudaqRuntime);
+  cudaq::bindPyUnitary(cudaqRuntime);
   cudaq::bindPyRun(cudaqRuntime);
   cudaq::bindPyRunAsync(cudaqRuntime);
   cudaq::bindPyTranslate(cudaqRuntime);
--- a/python/runtime/cudaq/algorithms/py_draw.cpp
+++ b/python/runtime/cudaq/algorithms/py_draw.cpp
@@ -17,7 +17,7 @@
 
 namespace cudaq {
 
-namespace {
+namespace details {
 std::tuple<std::string, MlirModule, OpaqueArguments *>
 getKernelLaunchParameters(py::object &kernel, py::args args) {
   if (py::len(kernel.attr("arguments")) != args.size())
@@ -35,12 +35,13 @@ getKernelLaunchParameters(py::object &kernel, py::args args) {
 
   return {kernelName, kernelMod, argData};
 }
-} // namespace
+
+} // namespace details
 
 /// @brief Run `cudaq::draw` on the provided kernel.
 std::string pyDraw(py::object &kernel, py::args args) {
   auto [kernelName, kernelMod, argData] =
-      getKernelLaunchParameters(kernel, args);
+      details::getKernelLaunchParameters(kernel, args);
 
   return details::extractTrace([&]() mutable {
     pyAltLaunchKernel(kernelName, kernelMod, *argData, {});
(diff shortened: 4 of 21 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -1434,6 +1434,52 @@ def visit_Attribute(self, node):
                     self.pushValue(self.getConstantInt(hash(channel_class)))
                     return
 
+        # Handle attribute access on call results (e.g., M(6, 9).i)
```
schweitzpgi, line 1463: I think this looks good.

Q: What about a path of properties?  e.g. `foo().a.bar().b.c.baz()`?

sacpis: Do we allow nested classes? Seems like right now, we aren't even allowing any non dunder methods in a dataclass.

nvidia-dobri: I don't think dunder methods would work either with the impl in main. Re: path of properties, this doesn't really handle that, but we can extend into a recursive version that does. Do we want to do that in this PR or just close the basic functionality gap for now?

annagrin: I think it is ok to postpone until we can test nested data classes.

annagrin (approved):
Thanks, Dobri, LGTM!

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1434,6 +1434,52 @@ def visit_Attribute(self, node):
...
+                    # Load the value
+                    eleAddr = cc.LoadOp(eleAddr).result
+                    self.pushValue(eleAddr)
+                    return
+            elif cc.StructType.isinstance(call_result.type):
+                # Handle direct struct member extraction
+                structIdx, memberTy = self.getStructMemberIdx(
+                    node.attr, call_result.type)
+                self.pushValue(
+                    quake.GetMemberOp(
```
amccaskey: This does not look right to me. We're talking about classical struct members, yet you use a Quake op meant for extracting quantum memory out of a purely quantum struct?

amccaskey (changes requested):
let's address the comment I just left

python/cudaq/kernel/ast_bridge.py
nvidia-dobri: Yeah, I understand your point. Let me trace it and see what's going on. Good catch.

nvidia-dobri: OK - I think I fell victim to my own copy/pasta from other sections of `visit_X` methods. Looks like there are some vestigial raw-struct handling snippets for things like tuples that we don't currently support. But probably did at some point? From what I can see in `ast_bridge.py` we shouldn't ever return a raw struct (not wrapped in a pointer) with the current impl, so I don't really see a way to write any test kernel or unit tests for this. I'll remove it for now.

Is that all fair, @amccaskey?

amccaskey: We introduced "quantum structs" for the CUDA-QX QEC release in November. Those structs have to be handled by value, hence the code you probably picked up for this. 

For this use case, where you require construction of a data class, you will only ever have a pointer to a cc.struct. With your latest commit you should be good now. 

Let's let @boschmitt have a chance to look at this one though.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: First, it was a question and I was expecting the answer to be "no". Thanks for confirming.

Second, it should be a general aim to design and build code for future possibilities/eventualities. That's just simply better design (open design) than the alternative of thinking only about all the current limitations and restrictions and hoping to ignore everything else as long as possible (box canyon design).

Finally, we obviously do not need to support this immediately nor in this PR. But, we definitely should be thinking of these language features and how to design the bridge to better handle them.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: For completeness, the IR fully supports values of `cc.struct` type. One would use the `cc.insert_value` and `cc.extract_value` operations to set and get the fields.

If one is using a `quake` (quantum) operation to manipulate a classic type value (`cc.struct`) then that's clearly incorrect and the verifier should given IR errors.

nvidia-dobri:
@boschmitt / @amccaskey - what's needed to get this in?

boschmitt:
> what's needed to get this in?

Ideally, the bridge would handle this more uniformly, not as an extra case in `visit_Attribute`. However, I do understand that the implementation follows the current pattern of how the bridge is being extended.

Having said that. I didn't had the time dig deeper in the PR yet, but I already found one issue. The following code does not work:
```python
@cudaq.kernel
def return_struct() -> M:
    return M(6, 9)

@cudaq.kernel
def test_attribute_on_call() -> int:
    return return_struct().i

```

If `return return_struct().i` is changed to `return return_struct()[0]` it does works.

nvidia-dobri:
> > what's needed to get this in?
> 
> Ideally, the bridge would handle this more uniformly, not as an extra case in `visit_Attribute`. However, I do understand that the implementation follows the current pattern of how the bridge is being extended.
> 
> Having said that. I didn't had the time dig deeper in the PR yet, but I already found one issue. The following code does not work:
> 
> ```python
> @cudaq.kernel
> def return_struct() -> M:
>     return M(6, 9)
> 
> @cudaq.kernel
> def test_attribute_on_call() -> int:
>     return return_struct().i
> ```
> 
> If `return return_struct().i` is changed to `return return_struct()[0]` it does works.

Isn't that just a special case of what we discussed already?
https://github.com/NVIDIA/cuda-quantum/pull/3123#discussion_r2172104646

If we want to fix recursive references, we absolutely can. Let's just agree on whether that's in scope here or not.

boschmitt:
> Isn't that just a special case of what we discussed already? [#3123 (comment)](https://github.com/NVIDIA/cuda-quantum/pull/3123#discussion_r2172104646)

No it is not. This comments seems to concern the cases of having a struct inside a struct. 

> If we want to fix recursive references, we absolutely can. Let's just agree on whether that's in scope here or not.

I assume the scope is: "Allow AST to handle property access on calls". So I fail to see why kernel calls shouldn't be included.

nvidia-dobri:
> > Isn't that just a special case of what we discussed already? [#3123 (comment)](https://github.com/NVIDIA/cuda-quantum/pull/3123#discussion_r2172104646)
> 
> No it is not. This comments seems to concern the cases of having a struct inside a struct.
> 
> > If we want to fix recursive references, we absolutely can. Let's just agree on whether that's in scope here or not.
> 
> I assume the scope is: "Allow AST to handle property access on calls". So I fail to see why kernel calls shouldn't be included.

Oh, yeah, you're right. I misread your example. OK, I think I played myself with the PR title because the original scope was "property access on `@dataclass`". Let me at least see how much work kernels working would be before I push back.

nvidia-dobri:
@boschmitt this handles kernel calls as well now.

boschmitt (approved):
Thanks for the work @nvidia-dobri.

PR #3126, opened 2025-06-27
Make ordering of matrix class configurable

amccaskey:
Looking through some other PRs, I saw pybind code in `cudaq/utils/matrix.h` (note it was just forward declared there) and Python functions were marked as `friend`. The reason for this was to allow Python runtime code to have access to the raw data pointer for the `complex_matrix`. 

This removes those declarations and the pybind forward declaration, and exposes a `get_data()` method on `complex_matrix`.


amccaskey:
> /ok to test

amccaskey:
> /ok to test 91f0aad

amccaskey:
> /ok to test

amccaskey:
> /ok to test a175366

schweitzpgi (approved):
LGTM

amccaskey:
> /ok to test  6e94c91

schweitzpgi:
> /ok to test 0e40efb

amccaskey:
> /ok to test e59b2fa

runtime/cudaq/utils/matrix.h
```diff
@@ -200,13 +196,7 @@ class complex_matrix {
...
   const EigenMatrix as_eigen() const;
 
-  friend void bindComplexMatrix(pybind11::module_ &mod);
-  friend void bindMatrixOperator(pybind11::module_ &mod);
-  friend void bindBosonOperator(pybind11::module_ &mod);
-  friend void bindFermionOperator(pybind11::module_ &mod);
-  friend void bindSpinOperator(pybind11::module_ &mod);
-  friend void bindOperatorHandlers(pybind11::module_ &mod);
-  friend void bindScalarOperator(pybind11::module_ &mod);
+  complex_matrix::value_type *get_data() const { return data; }
```
nvidia-dobri: This exposes a raw pointer to anyone and anything that wants it, though. If we change something about the internal memory layout of complex_matrix, we no longer have a list of users.

amccaskey: Thanks for the review. This is common practice for a lot of data types, especially for matrix types. Numpy, Eigen, armadillo, cupy all give you some way to extract the raw data to enable interoperability across frameworks (note one has to know the array ordering, here we document that). The C++ standard even does it (e.g. std::string). This is a much more common approach than marking certain types as `friends` to extract protected data members.

amccaskey:
> /ok to test 9da8559

runtime/cudaq/utils/matrix.h
bettinaheim: I'd really prefer that we do not expose the internal data here. The header itself should import even if you don't have pybind installed. Did it cause any issues somewhere?

schweitzpgi: From the looks of it this isn't exposing internal data so much as a way to pass a reference to an array of values to numpy, which requires a reference to a C array of data apparently to build its object(s).

<s>What is the actual suggestion here? Make a copy of the array data into a new temporary array? (At some point, numpy requires an array at any rate.)</s>

So, if I believe that the objection is to the removal of the `friend` declarations. Don't we still need the method, `get_data()` here though?

`friend` is rarely used in general and is a bit brittle in maintenance since it may require forward decls of classes, etc. It arguably should be used for tightly coupled classes/functions that require slightly different scopes for proper lookup. [Here is some debate on the subject.](https://www.reddit.com/r/cpp_questions/comments/1514o8k/whats_the_modern_c_take_on_friends/)

nvidia-dobri: The current state of things is just as performant, but there is an explicit allowlist of functions that can use the data property. Just note the `cmat.data` in the previously-friend functions turning into `cmat.get_data()`. Nothing else is changed, there are no copies added or removed. This is purely a member access change. Bettina's point is that once this is merged, `cmat.get_data()` can be called by anything - including user code outside our codebase. That's probably the main concern with replacing the explicit allowlist of trusted users of data to a method anyone can use.

bettinaheim: I agree the friend is not great. But exposing the pointer is also not great. I was wanting to keep data out of the public API because that is very brittle - there is no information whatsoever about how the data is stored. For performance, the class instead exposes the eigen representation of the matrix, which doesn't require any copy but basically tags the raw data with the information about how to interpret it as matrix. This should meet all public use cases at basically no performance cost and is much cleaner for the public API.

bettinaheim: I think we should be able to get around the use of friends by using the eigen representation in the bindings. 
The other alternative is to put get_data into a free function under a detail namespace to indicate in some form that it is not intended for public use.

amccaskey: Just a note (and confusion) from me on this "exposing the pointer is also not great" comment - exposing the pointer is standard practice across a large number of frameworks (including the C++ standard). It is used to enable interoperability across frameworks and approaches. The main requirement is that the user knows (via the class docs) that the data is laid out as a 1D row-major array (or whatever it is, maybe it is configurable). Maybe the problem here is that it is not marked as `const` (that should be added). 

The approved `cudaq::tensor` language proposal (which ultimately should replace this type) requires this raw data exposure for interoperability as well.

bettinaheim: Resolved with 5e076270ceb965388ce1a60754e76377bf3ae44c.

bettinaheim:
> /ok to test 2c5eac67e6c1ba2d5681452d0b75a79d924052c8

PR #3127, opened 2025-06-27
Update build to use native arm64 runners, remove qemu.

mitchdz:
This is an update of the deployments pipeline to build natively on arm64/amd64. This is a necessary change due to [qemu being unstable](https://github.com/tonistiigi/binfmt/issues/215).

There is currently a bit of cruft leftover from the multiarch build. I left a decent amount in so we can re-enable multiarch later. Currently the build will _not_ build multiarch without changes. If we want to get it ready to do multiarch builds I'd rather do that later and focus on getting the simple native builds done right now.

Example workflow runs are below.
**note**: If you want to test the changes, you have to rebuild the cache, or use `test_native_arm`

IN PROGRESS WORKFLOWS:
| Status               | Link|
|----------------|----------------------------------------------------------------------|
| cache update  | TBD |
| deployment     | TBD |
| publish             | TBD |

Changes:
1. removed all uses of `needs.metadata.outputs.multi_platform`

2. Use a proper matrix
```diff
      matrix:
-        platform: ${{ fromJson(needs.metadata.outputs.multi_platform || needs.metadata.outputs.platforms).ids }}
+        platform: ${{ fromJson(needs.metadata.outputs.platforms).ids }}
```

3. dispatch job is harded to build natively
```diff
-                platforms: 'linux/arm64,linux/amd64',
+                platforms: 'linux/arm64 linux/amd64',
```
This will need to be updated to dynamically build `linux/arm64,linux/amd64` as it did before if we want to enable multiarch in the future.

4. cloudposse write in `dev_images.yml` to capture the `{arch}-cu{cuda_version}-latest-base` sha256, and retrieve them later in deployment

5. use sha256 from `dev_images.yml` output to stitch together a multiarch manifest with buildx and publish to ghcr.

6. Removal of all qemu setups


mitchdz:
One known issue at the moment - in the final stages of the publishing pipeline, validating the metapackage fails to find
`libnvqir-nvidia-mgpu.so`.

[0] - https://github.com/NVIDIA/cuda-quantum/actions/runs/15939150416/job/44965750840

mitchdz:
I noticed that the `image_cu{11,12}_publishing` files were not generated. This caused an issue in the publishing stage that expects these. I just pushed a commit to generate that in deployments.yml. The logic gets a little messy because it was simply created in `docker_images.yml` prior within a matrix.

This also means that the `staging` job in `docker_images.yml` workflow is never used, so we could remove it.

I just queued another deployment off the test branch to make sure the new logic works and outputs the necessary `*_publishing` files that the publishing pipeline expects.

.github/workflows/deployments.yml
```diff
@@ -404,13 +411,202 @@ jobs:
...
+
+              digest=$(docker buildx imagetools inspect "$stitched_tag" | awk '/^Digest:/ { print $2 }')
+              stitched_output=$(echo "$stitched_output" | jq --arg key "$stitched_key" --arg val "$digest" '. + {($key): $val}')
+            done
+          done
+          echo "$stitched_output" > stitched_digests.json
+          stitched_json=$(jq -c . stitched_digests.json)
+          echo "stitched_json=$stitched_json" | tee -a $GITHUB_OUTPUT
+
+  summary:
```
mitchdz, line 564: @bettinaheim this is taking place of the `staging` job in `docker_images.yml`. I think I can completely remove the `staging` job there with this change. I don't visually see any other place in the codebase that use that.

.github/workflows/docker_images.yml
```diff
@@ -774,6 +777,7 @@ jobs:
           path: docs
           retention-days: 30
           if-no-files-found: error
+          overwrite: true
```
mitchdz: Just want to bring attention to this.

The multiplat build only runs this cache job once, but doing a proper matrix build it runs twice for amd64 now. This could be handled much more gracefully, but this is an easy fix for now, and anyways docs should be overhauled soon anyways.

.github/workflows/deployments.yml
```diff
@@ -174,15 +174,15 @@ jobs:
     if: needs.metadata.outputs.build_dependencies == 'true'
     strategy:
       matrix:
-        platform: ${{ fromJson(needs.metadata.outputs.multi_platform || needs.metadata.outputs.platforms).ids }}
+        platform: [linux/amd64, linux/arm64]
```
bettinaheim: Why not stick with `platform: ${{ fromJson(needs.metadata.outputs.platforms).ids }}`? I think hardcoding it would prevent that a manual build could build for a specific platform only.

.github/workflows/deployments.yml
```diff
@@ -174,15 +174,15 @@ jobs:
...
         toolchain: [clang16, gcc11, gcc12]
       fail-fast: false
     uses: ./.github/workflows/dev_environment.yml
     secrets:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       DOCKERHUB_READONLY_TOKEN: ${{ secrets.DOCKERHUB_READONLY_TOKEN }}
     with:
-      platforms: ${{ fromJson(needs.metadata.outputs.multi_platform || needs.metadata.outputs.platforms)[format('{0}', matrix.platform)].docker_flag }}
+      platforms: ${{ matrix.platform }}
```
bettinaheim: ... keeping the corresponding `fromJson...` here

.github/workflows/deployments.yml
```diff
@@ -404,13 +411,202 @@ jobs:
...
+        uses: cloudposse/github-action-matrix-outputs-read@1.0.0
+        with:
+          matrix-step-name: docker_images
+
+      - name: Print full digest config JSON
+        run: |
+          echo "Received matrix outputs:"
+          echo '${{ steps.read.outputs.result }}' | jq .
+
+  stitch:
```
bettinaheim, line 440: nit: It may be nice to move this into a separate reusable workflow.

.github/workflows/deployments.yml
```diff
@@ -404,13 +411,202 @@ jobs:
...
+            cuda_major="${cuda%%.*}"
+            key="arm64-cu${cuda}-ext"
+            arm64_image=$(echo "$EXT_JSON" | jq -r ".image_hash[\"$key\"]")
+            echo "CUDA $cuda arm64 image: $arm64_image"
+
+            key="amd64-cu${cuda}-ext"
+            amd64_image=$(echo "$EXT_JSON" | jq -r ".image_hash[\"$key\"]")
+            echo "CUDA $cuda amd64 image: $amd64_image"
+
+            stitched_tag="ghcr.io/nvidia/cuda-quantum-dev:cu${cuda}-ext"
```
bettinaheim: The tag here needs some work, too. It does not include any branch identifier. Rather than hardcoding it, you could get a suitable tag by looking at the tag of the image(s) you are stitching together.

.github/workflows/deployments.yml
```diff
@@ -404,13 +411,202 @@ jobs:
...
+
+            docker buildx imagetools create --tag "$stitched_tag" $refs
+            digest=$(docker buildx imagetools inspect "$stitched_tag" | awk '/^Digest:/ { print $2 }')
+            stitched_key="devdeps-cu${cuda_major}"
+            stitched_output=$(echo "$stitched_output" | jq --arg key "$stitched_key" --arg val "$digest" '. + {($key): $val}')
+          done
+
+          echo "Creating stitched manifest for cuda-quantum and cuda-quantum-dev..."
+          for image in image dev-image; do
+            for cuda in 11.8 12.0; do
```
bettinaheim, line 517: It would be nice to instead make this a matrix job as well for the CUDA versions. That way it is clearer when we update them.

.github/workflows/deployments.yml
```diff
@@ -404,13 +411,202 @@ jobs:
...
+                  registry="ghcr.io/nvidia/cuda-quantum"
+                fi
+
+                refs="$refs $registry@$digest"
+              done
+
+              cuda_major="${cuda%%.*}"
+
+              if [[ "$image" == "dev-image" ]]; then
+                stitched_tag="ghcr.io/nvidia/cuda-quantum-dev:cu${cuda_major}-latest-base"
```
bettinaheim: Similar comment here for the tags; latest is only used for the main branch, not for all other branches that have publishing support (release branches, experimental branches, ...).

.github/workflows/deployments.yml
bettinaheim: Indeed, staging is only used by the publishing (and the gitlab pipeline it triggers).

.github/workflows/docker_images.yml
bettinaheim: This is easy to clean up better; the build_docs parameter in deployments.yml can be adjusted to only be set to true for the first platform that it builds; something like
`build_docs: ${{ matrix.cuda_version == '11.8' && matrix.platform == fromJson(needs.metadata.outputs.platforms).ids[0] }}`

docker/build/assets.Dockerfile
```diff
@@ -226,9 +226,11 @@ RUN echo "Patching up wheel using auditwheel..." && \
         --exclude libcublasLt.so.11 \
         --exclude libcurand.so.10 \
         --exclude libcusolver.so.11 \
+        --exclude libcusparse.so.11 \
```
bettinaheim: good catch - both of these should be excluded. However, I think you need to add cusparse to the python wheel dependencies in the toml file then. You will also need to make a suitable update in a bunch of other places - search for curand, and you need to do probably the same thing for cusparse across the board. It would be good to make that a different PR.

docker/release/cudaq.wheel.Dockerfile
```diff
@@ -80,10 +80,11 @@ RUN echo "Building wheel for python${python_version}." \
         $python -m auditwheel -v repair dist/cuda_quantum*linux_*.whl \
             --exclude libcustatevec.so.1 \
             --exclude libcutensornet.so.2 \
+            --exclude libcudensitymat.so.0 \
             --exclude libcublas.so.$cudaq_major \
             --exclude libcublasLt.so.$cudaq_major \
             --exclude libcurand.so.10 \
-            --exclude libcusolver.so.$cudaq_major \
+            --exclude libcusolver.so.11 \
```
bettinaheim: Is libcusolver included in nvidia-cusolver-cu12 also having a .11 extension?

runtime/nvqir/cudensitymat/CMakeLists.txt
```diff
@@ -51,53 +51,73 @@ endif()
...
-  CuDensityMatTimeStepper.cpp
-  RungeKuttaIntegrator.cpp
-  CuDensityMatExpectation.cpp
-  CuDensityMatEvolution.cpp
-  CuDensityMatState.cpp
-  CuDensityMatContext.cpp
-  CuDensityMatOpConverter.cpp
-  CuDensityMatUtils.cpp
-)
+# Determine cudensitymat version
```
bettinaheim: I suggest to split this PR into two - one for the qemu elimination/pipeline changes, and a separate one for the cuquantum related changes.

.github/workflows/deployments.yml
mitchdz: Added in [73d1db4](https://github.com/NVIDIA/cuda-quantum/pull/3127/commits/73d1db4371143f227fd4761e9d7bbe43716028c9).

.github/workflows/deployments.yml
mitchdz: Added in [73d1db4](https://github.com/NVIDIA/cuda-quantum/pull/3127/commits/73d1db4371143f227fd4761e9d7bbe43716028c9).

.github/workflows/deployments.yml
mitchdz: That's a good point, I was just going off reproducing how the current build does it, but publishing should be working off of the sha256 digest anyways.

.github/workflows/deployments.yml
mitchdz: I can do that. I just wanted to avoid having to do another cloudposse usage.

.github/workflows/deployments.yml
mitchdz: why not use:
```
stitched_tag="ghcr.io/nvidia/cuda-quantum-dev:cu${cuda}-ext-${{ github.ref_name }}"
```

This point in the deployment shouldn't be used by a PR, but instead ran by you so `ref_name` shouldn't have any of the branch information.

.github/workflows/deployments.yml
mitchdz: Let's postpone this update until after it is merged. That requires a bit of code change which offers silly typos to be incurred, and when each run takes ~2hours, I'd rather not risk that time delay this close to release.

.github/workflows/deployments.yml
mitchdz: should be good in [57bc273](https://github.com/NVIDIA/cuda-quantum/pull/3127/commits/57bc2731bda9c02900e586f1782481a50c8ef80a). Will wait for it to go through CI to confirm.

docker/build/assets.Dockerfile
bettinaheim: part of a separate PR

docker/release/cudaq.wheel.Dockerfile
bettinaheim: part of a separate PR

runtime/nvqir/cudensitymat/CMakeLists.txt
bettinaheim: Separate PR here: https://github.com/NVIDIA/cuda-quantum/pull/3156

.github/workflows/deployments.yml
mitchdz: Let's postpone this as a refactor later on. I'd like to get the pipelines working sooner than later.

mitchdz:
> /ok to test c195334

mitchdz:
> /ok to test ae0eb2a

bmhowe23 (approved):
Approving per our meeting this afternoon

PR #3129, opened 2025-06-30
Dynamics operator batching

1tnguyen:
### Description

Support batching multiple dynamics simulations (different system dynamics, i.e., Hamiltonian), provided that they are compatible (similar structures). If batching is not possible, run them sequentially.

Specifically,

- Major refactor of the operator converter (from CUDA-Q to cudensitymat) infrastructure, treating batched (vector) data as a first-class citizen: single op is simply a vector of size 1.

- Split the op converter implementation into multiple source files for readability (it's getting big).

- Implementation for both C++ and Python.

- Unit tests and some examples.

Notes: documentation is not included in this PR and will be added later.


docs/sphinx/examples/python/dynamics/heisenberg_model.py
```diff
@@ -13,6 +13,11 @@
 # which exhibits the so-called quantum quench effect.
 # e.g., see `Quantum quenches in the anisotropic spin-1/2 Heisenberg chain: different approaches to many-body dynamics far from equilibrium`
 # (New J. Phys. 12 055017)
+
+# Specifically, we demonstrate the use of batched Hamiltonian operators to simulate the Heisenberg model
+# with different coupling strengths.
+# These batched Hamiltonian operators allow us to efficiently compute the dynamics
+# for multiple Hamiltonian operators in a single simulation run.
```
sacpis: nit: compute the dynamics of multiple systems in a single simulation run?

docs/sphinx/examples/python/dynamics/heisenberg_model.py
```diff
@@ -35,6 +40,7 @@
 staggered_magnetization_op /= N
 
 observe_results = []
+batched_hamiltonian = []
 for g in [0.0, 0.25, 4.0]:
```
sacpis: I think we can have this list declared outside.

docs/sphinx/examples/python/dynamics/heisenberg_model.py
```diff
@@ -48,26 +54,32 @@
...
+steps = np.linspace(0.0, 5, 1000)
+schedule = Schedule(steps, ["time"])
+
+# Prepare the initial state vector
+psi0_ = cp.zeros(2**N, dtype=cp.complex128)
+psi0_[int(spin_state, 2)] = 1.0
+psi0 = cudaq.State.from_data(psi0_)
+
+# Batched initial states (all the same in this case)
+batched_psi0 = [psi0] * len(batched_hamiltonian)
```
sacpis: Wondering if we can handle this batching of initial state internally (based on the batched_hamiltonian).

docs/sphinx/examples/python/dynamics/silicon_spin_qubit.py
```diff
@@ -16,32 +15,46 @@
...
+    H = 0.5 * resonance_frequency * spin.z(0) + amplitude * ScalarOperator(
+        lambda t: 0.5 * np.sin(resonance_frequency * t)) * spin.x(0)
+    # Append the Hamiltonian to the batched list
+    # This allows us to compute the dynamics for all amplitudes in a single simulation run
+    batched_hamiltonian.append(H)
+
+psi0 = cudaq.State.from_data(cp.array([1.0, 0.0], dtype=cp.complex128))
+# Initial state is the ground state of the spin qubit
+# We run all simulations for the same initial state, but with different Hamiltonian operators.
+initial_state = [psi0] * len(batched_hamiltonian)
```
sacpis: same here?

python/cudaq/dynamics/evolution.py
```diff
@@ -419,14 +424,28 @@ def evolve(
...
+                    raise ValueError(
+                        "If `hamiltonian` is a sequence, then `initial_state` must be a sequence of the same length."
+                    )
+                if len(collapse_operators) == 0:
+                    collapse_operators = [[] for _ in range(len(hamiltonian))]
+
+                if len(hamiltonian) != len(collapse_operators):
+                    raise ValueError(
+                        "If `hamiltonian` is a sequence, then `collapse_operators` must be a sequence of the same length."
+                    )
```
sacpis: Just a suggestion. A better way is to have a validate parameters function. This function can be called before the call to evolve_* API.

python/tests/dynamics/integrators/test_evolve_dynamics_torch_integrators.py
```diff
@@ -34,7 +34,12 @@ def do_something():
     TestCrossResonance, TestCallbackTensor, TestInitialStateEnum,
     TestCavityModelBatchedInputState, TestCavityModelSuperOperator,
     TestInitialStateEnumSuperOperator,
-    TestCavityModelBatchedInputStateSuperOperator
+    TestCavityModelBatchedInputStateSuperOperator,
+    TestCavityModelBatchedInputStateSuperOperator, TestBatchedCavityModel,
```
sacpis, line 37: nit: seems like `TestCavityModelBatchedInputStateSuperOperator` repeated.

python/tests/dynamics/system_models.py
```diff
@@ -593,3 +589,273 @@ def run_tests(self, integrator):
...
+            observables=[hamiltonian],
+            collapse_operators=collapse_operators_list,
+            store_intermediate_results=cudaq.IntermediateResultSave.
+            EXPECTATION_VALUE,
+            integrator=integrator())
+        for i, decay_rate in enumerate(decay_rates):
+            evolution_result = evolution_results[i]
+            expectation_values = []
+            for exp_vals in evolution_result.expectation_values():
+                expectation_values.append(exp_vals[0].expectation())
```
sacpis: I think here we can use a list comprehension instead.

```
expectation_values = [exp_vals[0].expectation() for exp_vals in evolution_result.expectation_values()]
```

python/tests/dynamics/system_models.py
```diff
@@ -593,3 +589,273 @@ def run_tests(self, integrator):
...
+            observables=[hamiltonian],
+            store_intermediate_results=cudaq.IntermediateResultSave.
+            EXPECTATION_VALUE,
+            integrator=integrator())
+
+        for i, decay_rate in enumerate(decay_rates):
+            evolution_result = evolution_results[i]
+            expectation_values = []
+            for exp_vals in evolution_result.expectation_values():
+                expectation_values.append(exp_vals[0].expectation())
```
sacpis: nit: list comp?

python/tests/dynamics/system_models.py
```diff
@@ -593,3 +589,273 @@ def run_tests(self, integrator):
...
+            store_intermediate_results=cudaq.IntermediateResultSave.
+            EXPECTATION_VALUE,
+            integrator=integrator(),
+            max_batch_size=1)
+
+        for i, decay_rate in enumerate(decay_rates):
+            evolution_result = evolution_results[i]
+            expectation_values = []
+            for exp_vals in evolution_result.expectation_values():
+                expectation_values.append(exp_vals[0].expectation())
```
sacpis: nit: here as well?

python/tests/dynamics/system_models.py
```diff
@@ -593,3 +589,273 @@ def run_tests(self, integrator):
...
```
sacpis: And the last one

docs/sphinx/examples/python/dynamics/heisenberg_model.py
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c43c991e7fca17056e12ed616fa890fbc6fac344

docs/sphinx/examples/python/dynamics/heisenberg_model.py
1tnguyen: Could you please clarify this a bit more? It is declared outside the `for` loop :)

docs/sphinx/examples/python/dynamics/heisenberg_model.py
1tnguyen: Good point! Implemented in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c43c991e7fca17056e12ed616fa890fbc6fac344

docs/sphinx/examples/python/dynamics/silicon_spin_qubit.py
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c43c991e7fca17056e12ed616fa890fbc6fac344

python/cudaq/dynamics/evolution.py
1tnguyen: I have moved the validation to the caller side (https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c43c991e7fca17056e12ed616fa890fbc6fac344). Indeed, it should have been validated there. 

There seem to be quite a few different validation conditions (e.g., depending on the target), so I defer making this check a separate function for the time being. We'll look into unifying all the validation checks as a follow-up.

docs/sphinx/examples/python/dynamics/heisenberg_model.py
sacpis: I mean something like 

```
anisotropy_parameters = [0.0, 0.25, 4.0]
for anisotropy_parameter in anisotropy_parameters:
   ...
```
as this `anisotropy_parameters` list can be reused on line 78.

python/tests/dynamics/integrators/test_evolve_dynamics_torch_integrators.py
1tnguyen: Good catch. Removed the duplication in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c43c991e7fca17056e12ed616fa890fbc6fac344

python/tests/dynamics/system_models.py
1tnguyen: Good point. Fixed all the usages in this test file (https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/299e4f7eca41a61e8c81b7f7cb0e85567a289be9)

python/tests/dynamics/system_models.py
1tnguyen: Fixed.

python/tests/dynamics/system_models.py
1tnguyen: Fixed :)

python/tests/dynamics/system_models.py
1tnguyen: yea, fixed.

docs/sphinx/examples/python/dynamics/heisenberg_model.py
1tnguyen: Gotcha! I was looking at the wrong line :-) 

Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/299e4f7eca41a61e8c81b7f7cb0e85567a289be9

docs/sphinx/examples/python/dynamics/heisenberg_model.py
```diff
@@ -13,6 +13,10 @@
 # which exhibits the so-called quantum quench effect.
 # e.g., see `Quantum quenches in the anisotropic spin-1/2 Heisenberg chain: different approaches to many-body dynamics far from equilibrium`
 # (New J. Phys. 12 055017)
+
+# Specifically, we demonstrate the use of batched Hamiltonian operators to simulate the Heisenberg model
```
bettinaheim, line 17: Do you also have a corresponding updated example for C++?

python/cudaq/dynamics/cudm_solver.py
```diff
@@ -51,21 +53,195 @@ def evolve_dynamics(
...
+                        f"The input super-operators are not compatible for batching. Unable to run batched simulation with the requested batch size {max_batch_size}."
+                    )
+                if max_batch_size is None:
+                    print(
+                        "Warning: The input super-operators are not compatible for batching. Running the simulation in non-batched mode."
+                    )
+                # Only run sequentially
+                max_batch_size = 1
+            else:
+                if max_batch_size is not None and max_batch_size > batch_size:
```
bettinaheim: Hm, do we really want to give an error here, or should we just silently batch all? It seems reasonable to me to set max_batch_size according to HW specs in a library function that takes the list of hamiltonians as arguments. In that case, the list may be shorter than what could reasonably fit and max_batch_size is bigger. It is a max after all, not the promised size of the batches.

python/cudaq/dynamics/cudm_solver.py
```diff
@@ -51,21 +53,195 @@ def evolve_dynamics(
...
+                    raise ValueError(
+                        f"Invalid max_batch_size {max_batch_size} for the given number of super-operators {batch_size}."
+                    )
+                if max_batch_size is not None and max_batch_size < 1:
```
bettinaheim: This seems a bit odd to me. I see that setting max_batch_size to 1 is the way to disable batching, and None is used to have CUDA-Q make an automatic choice. So it seems to me that generally, you just don't ever expect to get an argument that is not none but smaller than 1. This is regardless of whether the hamiltonian can be batched or not. Correspondingly, I'd expect this check to be at the beginning of the function outside the conditional here.

python/cudaq/dynamics/cudm_solver.py
```diff
@@ -51,21 +53,195 @@ def evolve_dynamics(
...
+                    raise ValueError(
+                        f"The input Hamiltonian and collapse operators are not compatible for batching. Unable to run batched simulation with the requested batch size {max_batch_size}."
+                    )
+                if max_batch_size is None:
+                    print(
+                        "Warning: The input Hamiltonian and collapse operators are not compatible for batching. Running the simulation in non-batched mode."
+                    )
+                # Only run sequentially
+                max_batch_size = 1
+            else:
```
bettinaheim: similar comments here as above.

runtime/nvqir/cudensitymat/CuDensityMatCallbackConverter.cpp
```diff
@@ -0,0 +1,193 @@
+/*******************************************************************************
```
bettinaheim, line 1: I haven't looked at this much - I assume this simply migrated from another file.

runtime/nvqir/cudensitymat/CuDensityMatOpConverter.cpp
```diff
@@ -291,10 +156,33 @@ cudaq::dynamics::CuDensityMatOpConverter::~CuDensityMatOpConverter() {
...
+    const auto &firstOp = elemOps[0];
+    return std::all_of(elemOps.begin(), elemOps.end(),
+                       [&firstOp](const auto &op) { return op == firstOp; });
+  }();
+  const bool isBatched = elemOps.size() > 1 && !allSame;
+  if (isBatched) {
+    // Check that the elementary operators have the same degrees.
+    const auto &firstOp = elemOps[0];
+    for (std::size_t i = 1; i < elemOps.size(); ++i) {
+      if (elemOps[i].degrees() != firstOp.degrees()) {
```
bettinaheim: nit: I'd expect to find all checks related to whether something can be batched in the same place. Small as it is, maybe move this into a separate helper where all other batching checks are?

runtime/nvqir/cudensitymat/CuDensityMatSuperOpCtor.cpp
```diff
@@ -0,0 +1,541 @@
...
+#include "CuDensityMatOpConverter.h"
+#include "CuDensityMatUtils.h"
+#include "common/Logger.h"
+#include <iostream>
+#include <map>
+#include <ranges>
+
+namespace {
+cudaq::product_op<cudaq::matrix_handler>
+computeDagger(const cudaq::matrix_handler &op) {
```
bettinaheim, line 19: Hm, I think it would make sense to move this to instead be a general .adjoint() on all operator classes. That can be done in a separate follow-up PR, however. It seems generally useful to have that.

docs/sphinx/examples/python/dynamics/heisenberg_model.py
1tnguyen: I've modified the corresponding C++ example in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/c956225d8bdea375ebbf90aed80f570d095f43e4.

python/cudaq/dynamics/cudm_solver.py
1tnguyen: That's a good point. I've removed this check and also added tests to make sure we can handle this case properly.

python/cudaq/dynamics/cudm_solver.py
1tnguyen: Agreed. I've moved this check to the top-level API, where users supply this argument.

python/cudaq/dynamics/cudm_solver.py
1tnguyen: Fixed similar to the above.

runtime/nvqir/cudensitymat/CuDensityMatCallbackConverter.cpp
1tnguyen: Yeah, I split the implementation into a couple of files for readability.

runtime/nvqir/cudensitymat/CuDensityMatOpConverter.cpp
1tnguyen: Sure. I've refactored this into a helper. Also, I converted it to an internal assertion check, as this condition should be met/validated from above (caller side).

runtime/nvqir/cudensitymat/CuDensityMatSuperOpCtor.cpp
1tnguyen: I posted https://github.com/NVIDIA/cuda-quantum/pull/3177 to implement `.adjoint()` for operator classes. We might consider pulling that in and refactoring this PR accordingly.

python/cudaq/dynamics/cudm_solver.py
```diff
@@ -51,21 +53,179 @@ def evolve_dynamics(
...
+                        f"The input super-operators are not compatible for batching. Unable to run batched simulation with the requested batch size {max_batch_size}."
+                    )
+                if max_batch_size is None:
+                    print(
+                        "Warning: The input super-operators are not compatible for batching. Running the simulation in non-batched mode."
+                    )
+                # Only run sequentially
+                max_batch_size = 1
+
+            if max_batch_size is None:
```
bettinaheim: nit: batch_size and this check can be pulled out of the if statement. You could also make a little helper function to check arguments/batch size that takes can_be_batched as argument to avoid the duplication. You could even overload the binding for superoperator and operator and have just a bindings.checkBatchingCompatibility rather than explicitly calling a different function depending on type. Generally, a lot of the code in both branches of the if could be factored out into a helper function to avoid code duplication (also the recursive call).

bettinaheim (approved):

python/cudaq/dynamics/cudm_solver.py
1tnguyen: Noted. I'll tidy up this in the upcoming doc PR that I'm working on.

1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3129/commits/ba06209eb5c85aee3bc41de834006fad4976bd27

PR #3130, opened 2025-06-30
Ties the various pieces together to allow the grover example code to compile and execute

schweitzpgi:
[core] Continue work on improving apply specialization.

We want to specialize and constant prop into apply callees. This constant prop should be used to convert the specialized callee into a DAG form. Also specialize any relaxed veq types while we're at it.

Adjust pipelines and such on Ionq and Quantinuum to allow for Grover example to JIT compile and execute.

Also tidy up the nvq++ driver a bit.

Add test.

### DIFF
--- a/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
+++ b/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
@@ -88,7 +88,9 @@ private:
           newArgs.append(apply.getArgs().begin(), apply.getArgs().end());
           IRMapping mapper;
           SmallVector<Value> preservedArgs;
+          SmallVector<Type> inputTys;
           SmallVector<arith::ConstantOp> moveConsts;
+          bool updateSignature = false;
           for (auto [idx, v] : llvm::enumerate(newArgs)) {
             if (auto c = v.getDefiningOp<arith::ConstantOp>()) {
               auto newConst = c.clone();
@@ -96,6 +98,14 @@ private:
               mapper.map(genericFunc.getArgument(idx), newConst);
               LLVM_DEBUG(llvm::dbgs() << "apply has constant arguments.\n");
             } else {
+              if (auto relax = v.getDefiningOp<quake::RelaxSizeOp>()) {
+                // Also, specialize any relaxed veq types.
+                v = relax.getInputVec();
+                updateSignature = true;
+                LLVM_DEBUG(llvm::dbgs() << "specializing apply veq argument ("
+                                        << v.getType() << ")\n");
+              }
+              inputTys.push_back(v.getType());
               preservedArgs.push_back(v);
             }
           }
@@ -107,6 +117,14 @@ private:
             func::FuncOp newFunc = genericFunc.clone(mapper);
             calleeName += std::string{"."} + std::to_string(counter++);
             newFunc.setName(calleeName);
+            auto *ctx = apply->getContext();
+            if (updateSignature) {
+              newFunc.setFunctionType(
+                  FunctionType::get(ctx, inputTys, newFunc.getResultTypes()));
+              for (auto [arg, ty] :
+                   llvm::zip(newFunc.front().getArguments(), inputTys))
+                arg.setType(ty);
+            }
             newFunc.setPrivate();
             Block &entry = newFunc.front();
             for (auto c : moveConsts)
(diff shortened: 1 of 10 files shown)


lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -88,14 +88,24 @@ struct ApplyOpAnalysis {
...
+          bool updateSignature = false;
           for (auto [idx, v] : llvm::enumerate(newArgs)) {
             if (auto c = v.getDefiningOp<arith::ConstantOp>()) {
               auto newConst = c.clone();
               moveConsts.push_back(newConst);
               mapper.map(genericFunc.getArgument(idx), newConst);
               LLVM_DEBUG(llvm::dbgs() << "apply has constant arguments.\n");
             } else {
+              if (auto relax = v.getDefiningOp<quake::RelaxSizeOp>()) {
+                // Also, specialize any relaxed veq types.
```
khalatepradnya, line 102: Q: Is this the reason for dropping `expand-measurements` from the pipeline below?

runtime/cudaq/platform/default/rest/helpers/ionq/ionq.yml
```diff
@@ -18,7 +18,7 @@ config:
   # Add the rest-qpu library to the link list
   link-libs: ["-lcudaq-rest-qpu"]
   # Define the lowering pipeline
-  platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,canonicalize,apply-op-specialization,aggressive-early-inlining,expand-measurements,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(lower-to-cfg,canonicalize,multicontrol-decomposition),ionq-gate-set-mapping"
+  platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,func.func(canonicalize),apply-op-specialization{constant-prop=1},aggressive-early-inlining,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(canonicalize,multicontrol-decomposition),ionq-gate-set-mapping"
```
khalatepradnya, line 21: Q: Should `constant-prop` option default to `true`?
(or why shouldn't it be?)

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
```diff
@@ -18,7 +18,7 @@ config:
   # Add the rest-qpu library to the link list
   link-libs: ["-lcudaq-rest-qpu"]
   # Define the lowering pipeline
-  platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,canonicalize,apply-op-specialization,aggressive-early-inlining,expand-measurements,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(lower-to-cfg,canonicalize,multicontrol-decomposition),quantinuum-gate-set-mapping"
+  platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,func.func(canonicalize),apply-op-specialization{constant-prop=1},aggressive-early-inlining,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(canonicalize,multicontrol-decomposition),quantinuum-gate-set-mapping"
```
khalatepradnya, line 21: It is not obvious to me why `lower-to-cfg` is removed?

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
schweitzpgi: No. The codegen pipeline (for every QIR target) already has an expand measurements, so this one is redundant and potentially problematic.

runtime/cudaq/platform/default/rest/helpers/ionq/ionq.yml
schweitzpgi: Being conservative for the moment. That is, not introducing sweeping change everywhere and smaller PRs are better anyway. :)

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
schweitzpgi: It is already part of the codegen pipeline and having it here is premature and blocks other optimizations. In short, it should be late in compilation, not stuck somewhere in the middle were it isn't needed and actively harmful.

targettests/execution/grover.cpp
```diff
@@ -0,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2025 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// REQUIRES: c++20
+// RUN: nvq++ --target quantinuum --emulate %s -o %t && %t | FileCheck %s
```
khalatepradnya, line 10: Also add?
```
// RUN: nvq++ --target ionq --emulate %s -o %t && %t | FileCheck %s
```

schweitzpgi: This fails because Ionq is using base profile which does not allow any control-flow, but this algorithm uses control-flow. 😞

PR #3135, opened 2025-07-01
Handle default POD constructors in Quake Bridge

nvidia-dobri:
Fixes the issue with POD default constructors hit by this gist:
https://gist.github.com/bmhowe23/44f606cac6420991daa5e2c8e853f400

Simply speaking, we were trying to make a call to a default constructor, which never gets generated. A POD "default constructor" is just an alloc, so we should, you know, do that.


lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -3154,6 +3154,13 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
     TODO_x(loc, x, mangler, "C++ constructor (non-default)");
   }
 
+  // Handle POD types with default constructors - just allocate memory
+  // No constructor call needed for POD types
+  if (parent->isPOD()) {
+    auto mem = builder.create<cc::AllocaOp>(loc, ctorTy);
```
schweitzpgi: The allocation will have already happened as it is the first argument to the ctor.

Oh, I see this is actually jumping the gun and before that happens...

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -3154,6 +3154,13 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
...
+    return pushValue(mem);
+  }
```
```suggestion
  if (parent->isPOD())
    return pushValue(mem);
```
and move it down to line 3164.

nvidia-dobri: 🤦  that's totally fair, yeah

test/AST-Quake/pod_constructor.cpp
```diff
@@ -0,0 +1,35 @@
...
+struct simple_pod {
+  long long a;
+  long long b;
+};
+
+// Test POD default constructor - should just allocate memory
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__function_test_pod_constructor._Z20test_pod_constructorv
+// CHECK:           () -> !cc.struct<"simple_pod" {i64, i64} [128,8]>
+// CHECK:           %[[ALLOC:.*]] = cc.alloca !cc.struct<"simple_pod" {i64, i64} [128,8]>
+// CHECK-NOT:       call @_ZN15foo_test_resultC1Ev
```
khalatepradnya: Q: Should this check that `simple_pod` constructor is not called, instead of `foo_test_result` which isn't defined in this test?

nvidia-dobri: Yeah, good catch, fixed.

test/AST-Quake/pod_constructor.cpp
```diff
@@ -0,0 +1,36 @@
...
+  long long a;
+  long long b;
+};
+
+// Test POD default constructor - should just allocate memory
+// CHECK-LABEL:   func.func
+// @__nvqpp__mlirgen__function_test_pod_constructor._Z20test_pod_constructorv
+// CHECK:           () -> !cc.struct<"simple_pod" {i64, i64} [128,8]>
+// CHECK:           %[[ALLOC:.*]] = cc.alloca !cc.struct<"simple_pod" {i64, i64}
+// [128,8]> CHECK-NOT:       call @_ZN10simple_podC1Ev CHECK:           return
```
khalatepradnya: Formatter probably messed this up. You may want to turn it off on these lines.

khalatepradnya (approved):
Thanks, @nvidia-dobri !

schweitzpgi (approved):

Issue #3090, opened 2025-06-18
[python, run] Adding cudaq.apply_noise in the kernel doesn't work


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Adding `cudaq.apply_noise` in the kernel doesn't work. Only passing the `noise_model` to the run API works.

### Steps to reproduce the bug

```python
import cudaq
cudaq.set_target("density-matrix-cpu")

@cudaq.kernel
def foo() -> int:
    q = cudaq.qvector(8)
    cudaq.apply_noise(cudaq.XError, 0.9, q[0])

    x(q)
    retval = 0
    for i in range(8):
        bit = mz(q[i])
        retval = retval * 2
        if bit == 0:
            retval = retval + 0
        else:
            retval = retval + 1
    return retval

results = cudaq.run(foo, shots_count=20)
print(results)
```

### Expected behavior

Same as with passing noise model

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

khalatepradnya:
Observation - invoking run with a `noise_model` works as expected.

```
results = cudaq.run(foo, shots_count=20)
print("Without noise model:")
print(results)

results = cudaq.run(foo, shots_count=20, noise_model=cudaq.NoiseModel())
print("With noise model:")
print(results)

```

gives following result:
```
# python3 bug.py 
Without noise model:
[255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255]
With noise model:
[127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127, 127]
```

PR #3137, opened 2025-07-01
[apply_noise] Give warning if noise model is not provided.

khalatepradnya:
Addresses: #3090


1tnguyen (approved):
LGTM 👍 Thanks @khalatepradnya

PR #3138, opened 2025-07-01
[Run API] Error on missing set slots option for a dataclass

### LINKED ISSUE #3085: [python, run] Creating data classes with constructors generates empty return
### Required prerequisites

### Describe the bug

Creating data classes with constructors generates empty return

```python

@dataclass
class MyClass:
    def __init__(self, i):
        self.x = i

@cudaq.kernel
def bugSP2() -> MyClass:
    return MyClass(-13)

print(cudaq.run(bugSP2, shots_count = 2))
```

Output:
[]

### Expected behavior

Error if an `__init__` is defined in a data class.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

sacpis:
As per the discussion in the Slack thread, we will be throwing an error if a defined dataclass is missing the slots=True option.

Fixes #3085.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -2177,7 +2177,7 @@ class PyASTBridge(ast.NodeVisitor):
 
                 if '__slots__' not in cls.__dict__:
                     self.emitFatalError(
-                        f"Assigning to fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True).",
+                        f"Adding new fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True).",
                         node)
 
                 # Alloca the struct
--- a/python/cudaq/kernel/utils.py
+++ b/python/cudaq/kernel/utils.py
@@ -258,7 +258,7 @@ def mlirTypeFromAnnotation(annotation, ctx, raiseError=False):
 
         if '__slots__' not in pyType.__dict__:
             localEmitFatalError(
-                f"Assigning to fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True)."
+                f"Adding new fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True)."
             )
 
         if len({
--- a/python/tests/kernel/test_run_kernel.py
+++ b/python/tests/kernel/test_run_kernel.py
@@ -1081,7 +1081,7 @@ def test_dataclass_no_slots_error():
 
     with pytest.raises(RuntimeError) as e:
         cudaq.run(kernel_with_no_slots_dataclass, shots_count=1)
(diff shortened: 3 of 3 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2173,6 +2173,14 @@ def bodyBuilder(iterVal):
                 # Handle User-Custom Struct Constructor
                 cls, annotations = globalRegisteredTypes.getClassAttributes(
                     node.func.id)
+
+                if "__dataclass_params__" in cls.__dict__:
+                    params = cls.__dict__["__dataclass_params__"]
+                    if not getattr(params, "frozen", False):
+                        self.emitFatalError(
+                            f"dataclass `{cls.__name__}` must be declared with @dataclass(frozen=True) or @dataclasses.dataclass(frozen=True).",
```
khalatepradnya: I think this error message can be made clearer by wording -
```suggestion
                            f"Assigning to fields in data classes is not yet supported. The`{cls.__name__}` must be declared with @dataclass(frozen=True) or @dataclasses.dataclass(frozen=True).",
```

sacpis: Fixed in [5d52a9f](https://github.com/NVIDIA/cuda-quantum/pull/3138/commits/5d52a9fa2a8f549f4f3eb443f1518eaf9b5c73cb).

sacpis:
> Thanks for the work @sacpis
> 
> Looking at our test cases, I see that `frozen` is a bit overreaching, it also prevents field assignments. Thus the following test shouldn't pass (but we want it to):
> 
> ```python
> @dataclass(frozen=True)
> class MyClass:
>     x: int
>     y: bool
> 
> @cudaq.kernel
> def simple_struct(t: MyClass) -> MyClass:
>     t.x = 42 # A frozen dataclass prevents this assignment
>     return t
> ```
> 
> It will be better to use `@dataclass(slots=True)`. This should prevent the addition of new fields while still allowing assignment of the "statically" declared ones.

This example works with frozen=True and fails on slots=True.

sacpis:
@boschmitt Would you like to go through this setting slots for dataclass change?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2173,6 +2174,12 @@ def bodyBuilder(iterVal):
                 # Handle User-Custom Struct Constructor
                 cls, annotations = globalRegisteredTypes.getClassAttributes(
                     node.func.id)
+
+                if '__slots__' not in cls.__dict__:
+                    self.emitFatalError(
+                        f"Assigning to fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True).",
```
boschmitt: Is this about assigning fields or adding fields ?

For example. If you have:
```python
    @dataclass(slots=True)
     class MyClass:
         x: int
         y: bool
```

Assigning values for `x` and `y` should be possible. Adding a new field, say `z`, shouldn't.

sacpis: It should be adding field. Leftover from the frozen option. Fixing the error message.

sacpis: Fixed in [7ae45a9](https://github.com/NVIDIA/cuda-quantum/pull/3138/commits/7ae45a951f8e81b7abba72f82a526afa473e8dfc).

PR #3139, opened 2025-07-02
Remove homebrew gzip decompress

nvidia-dobri:
Updating cpr seems to have broken our homebrew gzip decompression, likely because cpr handles that transparently now.
https://github.com/NVIDIA/cuda-quantum/commit/b89969d4c2a258a01fa3cd0ba37302eccdec141d

This essentially reverts https://github.com/NVIDIA/cuda-quantum/commit/ad164b9089f25928979289aab5de3bfcc8319b3d which is no longer needed.

Manually verified with the following script:
```
nvq++ -v bug_qubit.cpp --target ionq
./a.out
```
(`IONQ_API_KEY` env var is necessary to run the above)

Before this PR, it throws an error:
```
  what():  Exception during zlib decompression
Aborted (core dumped)
```

Printing the payload given to `decompress_gzip` shows it is, in fact, not gzip encoded, even though the `Content-Encoding` header is still `gzip`.

This should fix the issues related to gzip with our integration tests:
https://github.com/NVIDIA/cuda-quantum/actions/runs/15600865169/job/43940480226

### DIFF
--- a/runtime/common/RestClient.cpp
+++ b/runtime/common/RestClient.cpp
@@ -10,44 +10,9 @@
 #include "Logger.h"
 #include "cudaq/utils/cudaq_utils.h"
 #include <cpr/cpr.h>
-#include <zlib.h>
 
 namespace cudaq {
 constexpr long validHttpCode = 205;
-
-/// Decompress GZIP data. Throws an exception on error.
-std::string decompress_gzip(const std::string &data) {
-  if (data.empty())
-    return data;
-
-  std::string decompressed;
-  z_stream zs{};
-  zs.avail_in = data.size();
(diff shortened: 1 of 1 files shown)


mitchdz (approved):
I tried to track down the cpr commit that does the decompression, but wasn't able to pinpoint it on a first pass. Regardless, if it's being decoded now, I'm always a fan of removing hardcoded implementations like this. +1 from me.

schweitzpgi (approved):
Thanks, Dobri.

bmhowe23 (commented):
LGTM. Thanks, Dobri.

PR #3142, opened 2025-07-02
[core] Fix handling of empty vector in arg synth.

schweitzpgi:


runtime/test/test_argument_conversion.cpp
```diff
@@ -342,6 +342,23 @@ void test_scalars(mlir::MLIRContext *ctx) {
 }
 
 void test_vectors(mlir::MLIRContext *ctx) {
+  {
+    std::vector<std::int32_t> x;
```
sacpis, line 346: Can a vector be of other types as well?

schweitzpgi: Yes!

Issue #3087, opened 2025-06-18
[python, run] Lists nested in data classes are failing to compile


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

List nested in data classes fail to compile

### Steps to reproduce the bug

```python
@dataclass
class M:
    i:int
    j:list

@cudaq.kernel
def bug() -> M:
    return M(6, [9])

print(cudaq.run(bug, shots_count = 2))
```

Output:
```
error: 'cc.store' op failed to verify that type of value matches element type of pointer
```

### Expected behavior

Give an "not yet supported" error message

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

sacpis:
This is done. Closing it.

Output is

```
test.py:11: warning: Adding new fields in data classes is not yet supported. The dataclass must be declared with @dataclass(slots=True) or @dataclasses.dataclass(slots=True).
         (offending source -> M(6, [9]))
RuntimeError: return values with dynamically sized element types are not yet supported
```

PR #3144, opened 2025-07-02
[Run API][Python] Support only scalar types in data classes

sacpis:
Fixes issue #3087.

Check if a class has an attribute with a list type and throw an error.

# Output

```
cudaq.kernel.ast_bridge.CompilerError: test.py:11: error: `<class 'list'>` type is not yet supported in data classes. The allowed types are: [<class 'int'>, <class 'float'>, <class 'bool'>].
         (offending source -> M(6, [9]))
```

### DIFF
--- a/python/tests/kernel/test_run_kernel.py
+++ b/python/tests/kernel/test_run_kernel.py
@@ -1020,7 +1020,7 @@ def test_supported_simulators(target):
     cudaq.reset_target()
 
 
-def test_unsupported_simulator():
+def test_unsupported_targets_0():
     try:
         cudaq.set_target("dynamics")
         with pytest.raises(RuntimeError) as e:
@@ -1031,12 +1031,22 @@ def test_unsupported_simulator():
     finally:
         cudaq.reset_target()
 
+    try:
+        cudaq.set_target("orca")
+        with pytest.raises(RuntimeError) as e:
+            test_simple_run_ghz()
+        assert "No QPUs are available for this target" in repr(e)
+    except RuntimeError:
+        pytest.skip("target not available")
+    finally:
+        cudaq.reset_target()
+
 
 @pytest.mark.parametrize("target, env_var",
                          [("anyon", ""), ("infleqtion", "SUPERSTAQ_API_KEY"),
                           ("ionq", "IONQ_API_KEY"), ("quantinuum", "")])
 @pytest.mark.parametrize("emulate", [True, False])
-def test_unsupported_targets(target, env_var, emulate):
+def test_unsupported_targets_1(target, env_var, emulate):
     if env_var:
         os.environ[env_var] = "foobar"
 
@@ -1051,7 +1061,7 @@ def test_unsupported_targets(target, env_var, emulate):
 
 @skipIfBraketNotInstalled
 @pytest.mark.parametrize("target", ["braket", "quera"])
-def test_unsupported_targets2(target):
+def test_unsupported_targets_2(target):
     cudaq.set_target(target)
     with pytest.raises(RuntimeError) as e:
         test_simple_run_ghz()
--- a/python/runtime/cudaq/algorithms/py_run.cpp
+++ b/python/runtime/cudaq/algorithms/py_run.cpp
@@ -48,6 +48,9 @@ getKernelLaunchParameters(py::object &kernel, py::args args) {
     kernel.attr("compile")();
 
   auto kernelName = kernel.attr("name").cast<std::string>();
+  if (!py::hasattr(kernel, "module") || kernel.attr("module").is_none())
+    throw std::runtime_error(
+        "Unsupported target / Invalid kernel for `run`: missing module");
   auto kernelMod = kernel.attr("module").cast<MlirModule>();
   args = simplifiedValidateInputArguments(args);
   auto *argData = toOpaqueArgs(args, kernelMod, kernelName);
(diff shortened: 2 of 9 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2173,6 +2174,13 @@ def bodyBuilder(iterVal):
                 # Handle User-Custom Struct Constructor
                 cls, annotations = globalRegisteredTypes.getClassAttributes(
                     node.func.id)
+
+                for var, typ in annotations.items():
+                    origin = get_origin(typ)
+                    if origin is list or typ is list:
+                        self.emitFatalError(f'list type is not yet supported.',
```
khalatepradnya: I think this message can be made clearer by wording -
```suggestion
                        self.emitFatalError(f'`list` type is not yet supported in data classes.',
```

khalatepradnya:
Suggest updating the title as - `[Python] List is not supported in data classes`, since this restriction is outside of `run` as well.

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed in [dd3ed51](https://github.com/NVIDIA/cuda-quantum/pull/3144/commits/dd3ed5192c1141c9f30dd1d6bf4098879bb68221).

boschmitt (changes requested):
The set of types we accept inside `@dataclass` is much smaller than the set of all types a user can try to put inside it. Thus, instead of checking whether the types is not something we support, we should be checking if the type is a supported one and give an error for everything else.

sacpis:
@khalatepradnya @annagrin Would you please confirm what all types are supported in a dataclass?

khalatepradnya:
> @khalatepradnya @annagrin Would you please confirm what all types are supported in a dataclass?

Only scalar types (int, bool, float) are supported for now.

sacpis:
> > @khalatepradnya @annagrin Would you please confirm what all types are supported in a dataclass?
> 
> Only scalar types (int, bool, float) are supported for now.

Seems like cudaq.qview should be allowed as well?
https://github.com/NVIDIA/cuda-quantum/actions/runs/16092572367/job/45411413579?pr=3144#step:11:4610

python/tests/kernel/test_run_kernel.py
```diff
@@ -15,6 +15,7 @@
 import pytest
 
 list_err_msg = 'does not yet support returning `list` from entry-point kernels'
+ALLOWED_TYPES_IN_A_DATACLASS = [int, float, bool]
```
boschmitt: Import this from the kernel module. No need to define it in two different places and incur the possibility of these being different.

python/tests/kernel/test_run_kernel.py
```diff
@@ -980,6 +981,57 @@ def kernel_with_args(real: float, imag: float) -> complex:
...
+        x: list
+        y: int
+
+    @cudaq.kernel
+    def kernel() -> TestClass:
+        return TestClass([1, 2, 3], 4)
+
+    with pytest.raises(RuntimeError) as e:
+        cudaq.run(kernel, shots_count=1)
+    assert f"`<class 'list'>` type is not yet supported in data classes. The allowed types are: {ALLOWED_TYPES_IN_A_DATACLASS}." in str(
```
boschmitt: You might even not need `ALLOWED_TYPES_IN_A_DATACLASS`. Just match the first part of the exception string.

khalatepradnya:
> > > @khalatepradnya @annagrin Would you please confirm what all types are supported in a dataclass?
> > 
> > 
> > Only scalar types (int, bool, float) are supported for now.
> 
> Seems like cudaq.qview should be allowed as well? https://github.com/NVIDIA/cuda-quantum/actions/runs/16092572367/job/45411413579?pr=3144#step:11:4610

yes! 
(Scalar types only for entry-point kernels.)

python/tests/kernel/test_run_kernel.py
sacpis: In the actual error message to the user, we will have to inform as to which data types are allowed. right?

sacpis: Got it. Fixed.

sacpis: Fixed in [5362438](https://github.com/NVIDIA/cuda-quantum/pull/3144/commits/5362438a208c1aad84bcbd45968bbc005b0c4f4f).

python/tests/kernel/test_run_kernel.py
sacpis: Removed allowed types from the test in [5362438](https://github.com/NVIDIA/cuda-quantum/pull/3144/commits/5362438a208c1aad84bcbd45968bbc005b0c4f4f).

sacpis:
> > > > @khalatepradnya @annagrin Would you please confirm what all types are supported in a dataclass?
> > > 
> > > 
> > > Only scalar types (int, bool, float) are supported for now.
> > 
> > 
> > Seems like cudaq.qview should be allowed as well? https://github.com/NVIDIA/cuda-quantum/actions/runs/16092572367/job/45411413579?pr=3144#step:11:4610
> 
> yes! (Scalar types only for entry-point kernels.)

Added.

sacpis:
@boschmitt Does the `ALLOWED_TYPES_IN_A_DATACLASS` list looks good? Are we good with this PR?

boschmitt (approved):

PR #3148, opened 2025-07-03
[Run API] Handling return from loop variations in a kernel

### LINKED ISSUE #3086: [python, run] Returning from both if and else branches fails to compile
### Required prerequisites

### Describe the bug

```python
@cudaq.kernel
def bug(cond :bool) -> int:
    if cond:
        return 1
    else:
        return -1

print(cudaq.run(bug, False, shots_count = 2))
```

Output:
```
error: 'func.return' op has 0 operands, but enclosing function (@__nvqpp__mlirgen__bugMe1) returns 1
```

The following statement fails to compile:
```python
    if cond:
        return 1
    else:
        return -1
``` 

The following statement works:
```python
    if cond:
        return 1
    return -1
``` 

### Expected behavior

Output:
[ -1, -1]

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

sacpis:
Fixes #3086.

Handling return from loop variations in a kernel. 

1. The kernel with None or void return will runs as it was before.
2. The kernel with a valid return type expects return from all paths in the code.
3. An error is thrown prompting the user to add a return statement.
4. The kernel code containing a for/while loop followed by else is handled as per Python spec.

### DIFF
--- a/docs/sphinx/applications/python/quantum_transformer_src/cuda_quantum_transformer_env.yml
+++ b/docs/sphinx/applications/python/quantum_transformer_src/cuda_quantum_transformer_env.yml
@@ -105,7 +105,7 @@ dependencies:
   - jupyter-lsp=2.2.5
   - jupyter_client=8.6.3
   - jupyter_console=6.6.3
-  - jupyter_core=5.7.2
+  - jupyter_core=5.8.1
   - jupyter_events=0.11.0
   - jupyter_server=2.15.0
   - jupyter_server_terminals=0.5.3


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3969,7 +3969,13 @@ def visit_If(self, node):
             with InsertionPoint(elseBlock):
                 self.symbolTable.pushScope()
                 self.pushIfStmtBlockStack()
-                [self.visit(b) for b in node.orelse]
+                temp = []
+                for b in node.orelse:
+                    if isinstance(b, ast.Return):
+                        self.emitFatalError(
+                            f'Inner scope return is not yet supported. Please provide a return statement only in the outer (kernel) scope.',
```
khalatepradnya: Q: Does is work if there is a return statement in the outer scope as well?
For example,
```
if cond:
    return 1
else:
    return 0
return -1
```

sacpis: No, as discussed in the meeting and if I have correctly understood the discussion, we would like the user to just return from the kernel scope and not from any inner scope (especially the `if` conditional loop). If there is a return in the inner scope then we should flash a message that only return from the outer scope (kernel) is allowed.

Please correct me this is not as expected.

The following works

```
result = -1
if cond:
    result = 1
else:
    result = 0
return result
```

schweitzpgi:
> As discussed in #3086, the reported behavior is not a bug. @khalatepradnya has suggested improving the error message. This PR , however, goes way beyond that, it is modifying the behavior completely. Users should be able to return from "inner scopes".

The proper facility for doing an inner scope return is to use `cc.unwind_return` which is a higher-level operation that semantically does what you want. All the gritty details (assuming you use `cc.scope` correctly) are already handled by the compiler.

sacpis:
> As discussed in #3086, the reported behavior is not a bug. @khalatepradnya has suggested improving the error message. This PR , however, goes way beyond that, it is modifying the behavior completely. Users should be able to return from "inner scopes".

I think it should be marked as a bug. The reason being having an else block with a return statement (assuming if already has a return) should mark the end of execution in that function.

Also, as we have return from all paths (in the inner scope) we don't (should not) need to return from the outer scope (kernel).

I am finding it a bit hard to fathom what are we trying to do here, meaning returning from all inner paths as well as the outer one.

sacpis:
Ah, I can think of one scenario where we have a break statement (before return) inside the if else block which will cause it to fall to the outer scope.

boschmitt:
> The proper facility for doing an inner scope return is to use `cc.unwind_return` which is a higher-level operation that semantically does what you want. All the gritty details (assuming you use `cc.scope` correctly) are already handled by the compiler.

Yes, the question is whether the bridge should add terminator to the `func` op.

```python
@cudaq.kernel
def bug(cond :bool) -> int:
    if cond:
        return 1 # This is cc.unwind_return
    else:
        return -1 # This is cc.unwind_return

    # Should the bridge add a default `func.return` here?
    # In this case, this is unreachable, but `func` still needs a terminator
```

```python
@cudaq.kernel
def bug(cond :bool) -> int:
    if cond:
        return 1 # This is cc.unwind_return

    # Should the bridge add a default `func.return` here?
    # In this case, this is _not_ unreachable. So it is an user error
    # that the addition of a default return will mask.
```

Thus the addition of a default return operation is not always desirable, but the `func` operations still needs a terminator. Either the bridge is capable of analyzing the code to evaluate whether it is ok to add a default return, or it needs to act conservatively and complain when it is missing. (Or do we by-pass the op verifier after the bridge builds the IR?)

amccaskey:
> As discussed in #3086, the reported behavior is not a bug. @khalatepradnya has suggested improving the error message. This PR , however, goes way beyond that, it is modifying the behavior completely. Users should be able to return from "inner scopes".

+1

bettinaheim:
> > As discussed in #3086, the reported behavior is not a bug. @khalatepradnya has suggested improving the error message. This PR , however, goes way beyond that, it is modifying the behavior completely. Users should be able to return from "inner scopes".
> 
> +1

Following up in direct chat to speed things up, posting here what I mentioned there: 
Multiple returns is an optional extension of the QIR Adaptive profile. We have two options: 

Option 1: 
Allow returns from inner scope across the board. In that case, we need the following implemented for HW backends:
- The compiler has to be able to lift the return out of inner scopes and make them a single return at the end. 
- Don't require manually adding a dummy return at the end of the function. 
- Give a nice comprehensive error if an execution path exists that doesn't have a proper return. 
- Give a comprehensive error if different return statements return vectors of different length

Option 2 (if the above is not feasible for this release):
Allow returns from inner scopes only for simulator targets, don't require a dummy return at the end in that case. Give a comprehensive error for all returns from inner scopes for HW targets and their emulation.

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed in [d46c363](https://github.com/NVIDIA/cuda-quantum/pull/3148/commits/d46c363483fbc1d9bad675385c6066b8be4df6eb).

sacpis:
@boschmitt @bettinaheim @schweitzpgi @khalatepradnya  Would you be able to review this PR? So that we can merge it in if everything looks good and is done as per the discussion in Slack.

boschmitt (approved):
~LGTM~. Thanks @sacpis

python/cudaq/kernel/ast_bridge.py
```diff
@@ -265,6 +265,43 @@ def visit_FunctionDef(self, node):
 
         ValidateArgumentAnnotations(self).visit(astModule)
 
+        # Ensure that functions with a return-type annotation actually has a valid return statement
+        # in all paths, if not throw an error.
+        class ValidateReturnStatements(ast.NodeVisitor):
```
boschmitt, line 270: This should not be about requiring a `return` statement in the body of all `if`s in a kernel. For example, the following kernel is valid:

```python
@cudaq.kernel
def kernel(cond :bool) -> int:
    result = 0
    if cond:
        result = 1

    return result
```

This must be about proving whether the end of the kernel body is reachable.

```python
@cudaq.kernel
def kernel(cond :bool) -> int:
    if ... :
        for ... in ... :
            if ... :
               ...
            else:
               ...
    else:
       ...

    # The question is whether this point here is reachable or not.
    
    # The bridge has to either prove that here is unreachable, and thus
    # it is safe to silently add the `return` (in case the user didn't),
    # or complain to the user about a missing return.
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1140,7 +1177,13 @@ def isQuantumTy(ty):
                 [self.visit(n) for n in node.body[startIdx:]]
                 # Add the return operation
                 if not self.hasTerminator(self.entry):
-                    ret = func.ReturnOp([])
+                    # If the function has a known (non-None) return type, emit
+                    # an `undef` of that type and return it; else return void
+                    if self.knownResultType is not None:
+                        undef = cc.UndefOp(self.knownResultType).result
+                        ret = func.ReturnOp([undef])
```
boschmitt, line 1209: For this release, let's be conservative and just add an error message here instead of silently adding a return of an undefined value. In the next cycle, we can revisit the idea of adding an analysis to further improve user's life.

sacpis:
@boschmitt Seems like we are not handling this case in the kernel (apart from this PR's change)

```
@cudaq.kernel
def kernel(cond :bool) -> int:
    result = 0
    if cond:
        a = 1

    return a
```

it shows
```
cudaq.kernel.ast_bridge.CompilerError: test.py:9: error: Invalid variable name requested - 'a' is not defined within the quantum kernel it is used in.
         (offending source -> a)
```

Should we track this in another issue?

boschmitt:
> @boschmitt Seems like we are not handling this case in the kernel (apart from this PR's change)
> 
> ```
> @cudaq.kernel
> def kernel(cond :bool) -> int:
>     result = 0
>     if cond:
>         a = 1
> 
>     return a
> ```
> 
> it shows
> 
> ```
> cudaq.kernel.ast_bridge.CompilerError: test.py:9: error: Invalid variable name requested - 'a' is not defined within the quantum kernel it is used in.
>          (offending source -> a)
> ```
> 
> Should we track this in another issue?

I messed up this code, I started naming the variable `a` and then changed to `result` 😅 . I did fixed the comment shortly after. Clearly not fast enough :( Sorry for he churn.

sacpis:
@boschmitt @schweitzpgi Please let me know if we need to handle any more cases here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3681,6 +3744,11 @@ def visit_While(self, node):
             self.popForBodyStack()
             self.symbolTable.popScope()
 
+        # Handle the `else` branch of a while loop
```
schweitzpgi: How is this getting attached to the elseRegion of the LoopOp?

It's incorrect to just stick the code from the else after the LoopOp.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3647,11 +3706,15 @@ def bodyBuilder(iterVar):
 
         self.createInvariantForLoop(totalSize, bodyBuilder)
 
+        # Handle the `else` branch of a for loop
+        if node.orelse:
+            for stmt in node.orelse:
+                self.visit(stmt)
```
schweitzpgi: This code must go in the `elseRegion` of LoopOp...

python/cudaq/kernel/ast_bridge.py
schweitzpgi: If we make this an error, the compiler must halt here in the bridge. I don't want the bridge to create broken IR and pass it along to the rest of the compiler to crash on.

python/tests/kernel/test_run_kernel.py
```diff
@@ -1180,6 +1180,290 @@ def kernel_with_slots_dataclass() -> SlotsClass:
         e.value)
 
 
+def test_return_from_if_loop_with_true_condition():
```
schweitzpgi, line 1197: Can you add a test somewhere that dumps the kernel's Quake IR and uses FileCheck to verify it? Particularly for the `for else` and `while else` cases. Apparently it never existed. (?)

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed.

python/tests/kernel/test_run_kernel.py
sacpis: We are now rejecting the code with both these blocks. But will remember to add these tests when we enable these loops with else block.

python/cudaq/kernel/ast_bridge.py
sacpis: 1. Following case has been handled.

```
@cudaq.kernel
def kernel(cond :bool) -> int:
    result = 0
    if cond:
        result = 1

    return result
```

2. We have chosen the option of stating to the user if the return statement is missing from any of the code paths.

python/cudaq/kernel/ast_bridge.py
sacpis: Should we just throw an error message and halt the compiler here itself? Are we good with it?

sacpis:
> Looking good, Sachin. Can you add those tests too? Thanks!

Thanks @schweitzpgi. Tests have been added in de76961.

schweitzpgi (approved):

PR #3149, opened 2025-07-03
[Run API][Python] Handling integer shift operators

### LINKED ISSUE #3091: [python] Support integer shift operators
### Required prerequisites

### Describe the bug

Support integer shift operators

```python
@cudaq.kernel
def bug(n: int) -> int:
    q = cudaq.qvector(n)
    m = mz(q)
    r = 0
    for i in range(n):
        r = r & (m[i] << i)
    
    return r

print(cudaq.run(bug, 3, shots_count = 2))
```

Output:
```
# cudaq.kernel.ast_bridge.CompilerError: program.py:74: error: unhandled binary operator - <ast.LShift object at 0x7b82ed39a440>
# (offending source -> m[i] << i)
```

### Expected behavior

Runs successfully, output:
[0,0]

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

sacpis:
Handling integer shift operators

Fixes #3091.

### DIFF
--- a/runtime/common/RunTheKernel.cpp
+++ b/runtime/common/RunTheKernel.cpp
@@ -30,6 +30,8 @@
 #include <utility>
 #include <vector>
 
+using namespace mlir;
+
 namespace cudaq {
 namespace details {
 /// Extracts data layout information from MLIR modules
@@ -39,33 +41,30 @@ public:
   extractLayout(const std::string &, const std::string &);
 
 private:
-  static mlir::MLIRContext *createContext();
+  static MLIRContext *createContext();
 };
 } // namespace details
 } // namespace cudaq
 
-using namespace llvm;
-using namespace mlir;
-
-mlir::MLIRContext *cudaq::details::LayoutExtractor::createContext() {
-  mlir::DialectRegistry registry;
+MLIRContext *cudaq::details::LayoutExtractor::createContext() {
+  DialectRegistry registry;
   cudaq::opt::registerCodeGenDialect(registry);
   cudaq::registerAllDialects(registry);
-  auto context = new mlir::MLIRContext(registry);
+  auto context = new MLIRContext(registry);
   context->loadAllAvailableDialects();
-  mlir::registerLLVMDialectTranslation(*context);
+  registerLLVMDialectTranslation(*context);
   return context;
 }
 
 std::pair<std::size_t, std::vector<std::size_t>>
 cudaq::details::LayoutExtractor::extractLayout(const std::string &kernelName,
                                                const std::string &quakeCode) {
-  std::unique_ptr<mlir::MLIRContext> mlirContext(createContext());
-  auto m_module = mlir::parseSourceString<mlir::ModuleOp>(
-      llvm::StringRef(quakeCode), mlirContext.get());
+  std::unique_ptr<MLIRContext> mlirContext(createContext());
+  auto m_module =
+      parseSourceString<ModuleOp>(StringRef(quakeCode), mlirContext.get());
   if (!m_module)
     throw std::runtime_error("module cannot be parsed");
-  mlir::func::FuncOp kernelFunc = m_module->lookupSymbol<func::FuncOp>(
+  func::FuncOp kernelFunc = m_module->lookupSymbol<func::FuncOp>(
       cudaq::runtime::cudaqGenPrefixName + kernelName);
   if (!kernelFunc)
     throw std::runtime_error("Could not find " + kernelName +
@@ -74,7 +73,7 @@ cudaq::details::LayoutExtractor::extractLayout(const std::string &kernelName,
   std::size_t totalSize = 0;
   std::vector<std::size_t> fieldOffsets;
   if (kernelFunc.getNumResults() > 0) {
-    mlir::Type returnType = kernelFunc.getResultTypes()[0];
+    Type returnType = kernelFunc.getResultTypes()[0];
     auto mod = kernelFunc->getParentOfType<ModuleOp>();
     StringRef dataLayoutSpec = "";
     if (auto attr = mod->getAttr(cudaq::opt::factory::targetDataLayoutAttrName))
@@ -85,7 +84,7 @@ cudaq::details::LayoutExtractor::extractLayout(const std::string &kernelName,
     LLVMTypeConverter converter(kernelFunc.getContext());
     cudaq::opt::initializeTypeConversions(converter);
     // Handle structure types
-    if (auto structType = mlir::dyn_cast<cudaq::cc::StructType>(returnType)) {
+    if (auto structType = dyn_cast<cudaq::cc::StructType>(returnType)) {
       auto llvmDialectTy = converter.convertType(structType);
       LLVM::TypeToLLVMIRTranslator translator(context);
       auto *llvmStructTy =
@@ -95,8 +94,9 @@ cudaq::details::LayoutExtractor::extractLayout(const std::string &kernelName,
       std::size_t numElements = structType.getMembers().size();
       for (std::size_t i = 0; i < numElements; ++i)
         fieldOffsets.emplace_back(layout->getElementOffset(i));
-    } else
+    } else {
       totalSize = cudaq::opt::getDataSize(dataLayout, returnType);
+    }
   }
   return {totalSize, fieldOffsets};
 }
@@ -124,9 +124,9 @@ cudaq::details::RunResultSpan cudaq::details::runTheKernel(
     std::string remoteOutputLog(ctx->invocationResultBuffer.begin(),
                                 ctx->invocationResultBuffer.end());
     circuitSimulator->outputLog.swap(remoteOutputLog);
-  } else if (platform.is_remote() || platform.is_emulated())
+  } else if (platform.is_remote() || platform.is_emulated()) {
     throw std::runtime_error("`run` is not yet supported on this target.");
-  else {
+  } else {
     auto ctx = std::make_unique<cudaq::ExecutionContext>("run", 1);
     for (std::size_t i = 0; i < shots; ++i) {
       // Set the execution context since as noise model is attached to this


python/cudaq/kernel/ast_bridge.py
```diff
@@ -4263,8 +4275,41 @@ def visit_BinOp(self, node):
...
+
+        if isinstance(node.op, ast.LShift):
+            if IntegerType.isinstance(left.type) and IntegerType.isinstance(
+                    right.type):
+                right = self.matchIntegerWidth(left, right)
+                self.pushValue(arith.ShLIOp(left, right).result)
+                return
+            else:
+                self.emitFatalError(
+                    "unhandled BinOp.LShift types; only integers supported.",
```
khalatepradnya: I think this error message can be made simpler by - 
```suggestion
                    "unsupported operand type(s) for '<< '; only integers supported.",
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4263,8 +4275,41 @@ def visit_BinOp(self, node):
...
+
+        if isinstance(node.op, ast.RShift):
+            if IntegerType.isinstance(left.type) and IntegerType.isinstance(
+                    right.type):
+                right = self.matchIntegerWidth(left, right)
+                self.pushValue(arith.ShRSIOp(left, right).result)
+                return
+            else:
+                self.emitFatalError(
+                    "unhandled BinOp.RShift types; only integers supported.",
```
khalatepradnya: Same here. 
```suggestion
                    "unsupported operand type(s) for '>> '; only integers supported.",
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4263,8 +4275,41 @@ def visit_BinOp(self, node):
...
+                    node)
+
+        if isinstance(node.op, ast.BitAnd):
```
khalatepradnya, line 4291: This code is for bitwise-and ( `&`), right? So the heading of the PR can be `[Python] Handle bitwise operations`?
And if so, what about other bitwise operations like - `|`, `^`, `~`?

sacpis: This code is for the operators (<<, &) used in the issue #3091. Happy to extend it to other bitwise operators.

sacpis: Just FYI, `~` is a unary op and we have already handled it in UnaryOp function.

sacpis: Fixed in e64cebd.

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed in e64cebd.

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed in e64cebd.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4144,6 +4144,18 @@ def visit_Continue(self, node):
...
 
+    def matchIntegerWidth(self, target, source):
+        target_width = IntegerType(target.type).width
+        source_width = IntegerType(source.type).width
+
+        if target_width == source_width:
+            return source
+
+        if source_width < target_width:
+            return arith.ExtUIOp(target.type, source).result
```
schweitzpgi: Aren't all integers in Python signed?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4263,8 +4275,63 @@ def visit_BinOp(self, node):
...
+                return
+            else:
+                self.emitFatalError(
+                    "unsupported operand type(s) for '>>'; only integers supported.",
+                    node)
+
+        if isinstance(node.op, ast.BitAnd):
+            if IntegerType.isinstance(left.type) and IntegerType.isinstance(
+                    right.type):
+                right = self.matchIntegerWidth(left, right)
```
schweitzpgi: I'm not really sure how this should translate from Python's arbitrary precision integer values, but it seems like the longest bitwidth would win? This implementation is that the value on the left-hand side dictates, so if the lhs is shorter, we drop rhs bits on the floor (which might be incorrect for OR operator?)...

python/cudaq/kernel/ast_bridge.py
sacpis: Thanks for catching it.

sacpis: Fixed in [d930a1a](https://github.com/NVIDIA/cuda-quantum/pull/3149/commits/d930a1ae0d2fad1e9ea0ccc200a82e1ac8dee0cc).

python/cudaq/kernel/ast_bridge.py
sacpis: Have changed the implementation to make sure the longest one dictates.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4144,6 +4144,24 @@ def visit_Continue(self, node):
         else:
             cc.ContinueOp([])
 
+    def matchIntegerWidth(self, source, target):
```
annagrin: can `promoteOperandType` be used for this purpose?

python/tests/kernel/test_run_kernel.py
```diff
@@ -1060,6 +1060,96 @@ def test_unsupported_targets2(target):
     cudaq.reset_target()
 
 
+def test_run_with_integer_left_shift_operator():
```
annagrin: NIT: It feels like the tests are not really related to run functionality, maybe move them to a separate file? (even if they still will use run:)

python/cudaq/kernel/ast_bridge.py
sacpis: Sure, we can use it. Thank you.

sacpis: Fixed in [d0d4a7f](https://github.com/NVIDIA/cuda-quantum/pull/3149/commits/d0d4a7f960c40e351350bcfb3a52b615bdc55677). Thanks @annagrin.

python/tests/kernel/test_run_kernel.py
sacpis: Fixed in [d0d4a7f](https://github.com/NVIDIA/cuda-quantum/pull/3149/commits/d0d4a7f960c40e351350bcfb3a52b615bdc55677).

python/tests/kernel/test_kernel_shift_operators.py
boschmitt, line 1: The tests are weak. They are all expected to return `0` for one reason or another, and not always that related to the correctness of the shift operators. For example,

```python
        r = 0
         for i in range(n):
             r = r & (m[i] << i)
```

In this case, the shift operator might be implemented wrongly, but the compiler will still gives the correct result as long as `&` is implemented correctly.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4263,8 +4263,68 @@ def visit_BinOp(self, node):
...
             self.pushValue(arith.RemUIOp(left, right).result)
             return
-        else:
-            self.emitFatalError(f"unhandled binary operator - {node.op}", node)
+
+        if isinstance(node.op, ast.LShift):
+            if IntegerType.isinstance(left.type) and IntegerType.isinstance(
+                    right.type):
+                left = self.promoteOperandType(self.getIntegerType(), left)
+                right = self.promoteOperandType(self.getIntegerType(), right)
```
boschmitt, line 4271: Why enforce that the shift operators only work with `int64`? CUDA-Q allows the use of `np.int8`, `np.int32`, etc. Silently promoting values of these types to `int64` for no good reason is not the right thing to do.

I'm not sure that the following kernel would work:
```python
@cudaq.kernel
def kernel(r: np.int32) -> np.int32:
    r = r << 1
    return r
```

python/tests/kernel/test_kernel_shift_operators.py
sacpis: Ack, Let me add more test cases.

python/cudaq/kernel/ast_bridge.py
sacpis: In that case, would the following function work, where the number with the larger precision takes priority.

```
def matchIntegerWidth(self, source, target):
        source_width = IntegerType(source.type).width
        target_width = IntegerType(target.type).width

        if target_width == source_width:
            return source, target

        max_width = max(target_width, source_width)
        target_type = IntegerType.get_signless(max_width)

        if source_width < max_width:
            source = arith.ExtSIOp(target_type, source).result

        if target_width < max_width:
            target = arith.ExtSIOp(target_type, target).result

        return source, target
```

@annagrin for viz.

sacpis: Instead of IntegerType casting and fetching width, I will fetch the width of the original type.

python/tests/kernel/test_kernel_shift_operators.py
sacpis: Here is the PR for basic shift operators test
https://github.com/NVIDIA/cuda-quantum/pull/3165

Please let me know if you would like to have any additional test cases. To add tests for other integer types (like np.int8, np.int32), I am making changes in another PR to accept these types. Will add tests for the same there itself.

python/cudaq/kernel/ast_bridge.py
sacpis: Now, if we were to follow the actual behavior of [arith::ShLIOp](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithshli-arithshliop) then we will have to treat the original number's bit width as it is. The reason being if the second operand's value is equal or greater than the first one, we will have to throw an error.
Do we need to follow this? In that case, we would need to raise an exception after getting a poison value. 

Also, rhere are other questions about overflow flags as well.

PR #3153, opened 2025-07-03
[run] [Python] Gracefully handle unsupported targets.

khalatepradnya:
Requires: https://github.com/NVIDIA/cuda-quantum/pull/3147


schweitzpgi (approved):
LGTM

PR #3156, opened 2025-07-04
Update to cuquantum 25.06

1tnguyen:
### Description

- Remove dependency to cuquantum-**python**, just the binary distribution.

**Note**: cuquantum-python 25.06 no longer supports Python 3.10.

- Add cusparse dependency. This is a new dependency of cudensitymat.

- Fix cudensitymat API usage according to the new version, need the new gradient callback param.

- Modify the dynamics CMake to check cudensitymat version as the code is no longer compatible with the previous version.

- Workaround spurious error log (related to synchronization) from cudensitymat during shutdown. We will look into the issue later. This occurs during the final shutdown, hence it is safe to ignore for the time being.

- Fix a typo in audit-wheel exclude: cusolver doesn't follow the cuda version, e.g., it has version 11.7 in CUDA toolkit 12.x (https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html)

### DIFF
--- a/docker/build/assets.Dockerfile
+++ b/docker/build/assets.Dockerfile
@@ -226,9 +226,11 @@ RUN echo "Patching up wheel using auditwheel..." && \
         --exclude libcublasLt.so.11 \
         --exclude libcurand.so.10 \
         --exclude libcusolver.so.11 \
+        --exclude libcusparse.so.11 \
         --exclude libcutensor.so.2 \
         --exclude libcutensornet.so.2 \
         --exclude libcustatevec.so.1 \
+        --exclude libcudensitymat.so.0 \
         --exclude libcudart.so.11.0 \
         --exclude libnvToolsExt.so.1 \
         --exclude libnvidia-ml.so.1 \
--- a/docker/build/assets.Dockerfile
+++ b/docker/build/assets.Dockerfile
@@ -263,7 +263,7 @@ RUN gcc_packages=$(dnf list installed "gcc*" | sed '/Installed Packages/d' | cut
 
 ## [Python MLIR tests]
 RUN cd /cuda-quantum && source scripts/configure_build.sh && \
-    python3 -m pip install lit pytest scipy cuquantum-python-cu$(echo ${CUDA_VERSION} | cut -d . -f1)~=25.03 && \
+    python3 -m pip install lit pytest scipy && \
     "${LLVM_INSTALL_PREFIX}/bin/llvm-lit" -v _skbuild/python/tests/mlir \
         --param nvqpp_site_config=_skbuild/python/tests/mlir/lit.site.cfg.py
 # The other tests for the Python wheel are run post-installation.
@@ -324,7 +324,8 @@ RUN . /cuda-quantum/scripts/configure_build.sh install-gcc && \
         cuda-compiler-$(echo ${CUDA_VERSION} | tr . -) \
         cuda-cudart-devel-$(echo ${CUDA_VERSION} | tr . -) \
         libcublas-devel-$(echo ${CUDA_VERSION} | tr . -) \
-        libcurand-devel-$(echo ${CUDA_VERSION} | tr . -) && \
+        libcurand-devel-$(echo ${CUDA_VERSION} | tr . -) \
+        libcusparse-devel-$(echo ${CUDA_VERSION} | tr . -) && \
     if [ $(echo $CUDA_VERSION | cut -d "." -f1) -ge 12 ]; then \
         dnf install -y --nobest --setopt=install_weak_deps=False \
             libnvjitlink-$(echo ${CUDA_VERSION} | tr . -); \
--- a/docker/build/assets.Dockerfile
+++ b/docker/build/assets.Dockerfile
@@ -189,8 +189,10 @@ RUN dnf install -y --nobest --setopt=install_weak_deps=False ${PYTHON}-devel && 
 RUN cd /cuda-quantum && source scripts/configure_build.sh && \
     if [ "${CUDA_VERSION#11.}" != "${CUDA_VERSION}" ]; then \
         cublas_version=11.11 && \
+        cusparse_version=11.7 && \
         sed -i "s/-cu12/-cu11/g" pyproject.toml && \
         sed -i -E "s/(nvidia-cublas-cu[0-9]* ~= )[0-9\.]*/\1${cublas_version}/g" pyproject.toml && \
+        sed -i -E "s/(nvidia-cusparse-cu[0-9]* ~= )[0-9\.]*/\1${cusparse_version}/g" pyproject.toml && \
         sed -i -E "s/(nvidia-cuda-nvrtc-cu[0-9]* ~= )[0-9\.]*/\1${CUDA_VERSION}/g" pyproject.toml && \
         sed -i -E "s/(nvidia-cuda-runtime-cu[0-9]* ~= )[0-9\.]*/\1${CUDA_VERSION}/g" pyproject.toml; \
     fi && \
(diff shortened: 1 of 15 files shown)


runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
```diff
@@ -122,7 +122,17 @@ Context::Context(int deviceId) : m_deviceId(deviceId) {
 /// @brief Destroy the Context object and release resources.
 Context::~Context() {
   m_opConverter.reset();
-  cudensitymatDestroy(m_cudmHandle);
+  // Prevent spurious error message during shutdown when destroying the
```
bmhowe23: I'm curious what the underlying issue is here. I don't think this workaround is thread-safe (i.e. if multiple threads are calling the destructor at once, they could be trying to modify `std::cerr` at the same time.)

bmhowe23:
> **Note**: cuquantum-python 25.06 no longer supports Python 3.10.

AFAICT, this PR still maintains support for Python 3.10, and it successfully ran through the CI (https://github.com/NVIDIA/cuda-quantum/actions/runs/16081951805/job/45391676912). Is there a known problem with cuQuantum 25.06 and Python 3.10, or is this just meaning it hasn't been thoroughly tested?

1tnguyen:
> > **Note**: cuquantum-python 25.06 no longer supports Python 3.10.
> 
> AFAICT, this PR still maintains support for Python 3.10, and it successfully ran through the CI (https://github.com/NVIDIA/cuda-quantum/actions/runs/16081951805/job/45391676912). Is there a known problem with cuQuantum 25.06 and Python 3.10, or is this just meaning it hasn't been thoroughly tested?

cuquantum-python dropped support for 3.10 in 25.06 as it added support for 3.13. There shouldn't be any problem; it's just that they followed their schedule w.r.t. Python support. 

We no longer use the Python API from cuquantum, just using our pybind11 binding to the C-API library. Our CI for this PR only tested the build for now. The Publishing pipeline will test the `dynamics` examples.

runtime/nvqir/cudensitymat/CuDensityMatContext.cpp
1tnguyen: I think the underlying issue is related to the shutdown order. I've found the fix (https://github.com/NVIDIA/cuda-quantum/pull/3156/commits/5f16f3d3a9b442604a5a5fc12c2d95afedffaf35). Tested locally by running C++/Python unit tests and examples without any error messages.

docker/build/assets.Dockerfile
```diff
@@ -226,9 +226,11 @@ RUN echo "Patching up wheel using auditwheel..." && \
         --exclude libcublasLt.so.11 \
         --exclude libcurand.so.10 \
         --exclude libcusolver.so.11 \
+        --exclude libcusparse.so.11 \
```
bettinaheim, line 231: You will need to make a bunch of changes across the board to have cusparse be an external dependency (as it should be). Searching for `curand` across the codebase and making a matching change probably should be the right things to do, unless I miss something about this specific dependency. In particular, you may need to add it to the python __init__.py file to make sure the dependency is properly loaded/found.

pyproject.toml
```diff
@@ -18,7 +18,7 @@ requires-python = ">=3.10"
 license = { file="LICENSE" }
 dependencies = [
   'astpretty ~= 3.0',
-  'cuquantum-python-cu12 >= 25.03',
+  'cuquantum-cu12 >= 25.06',
```
bettinaheim, line 21: Does this pull in cusparse?

scripts/configure_build.sh
```diff
@@ -71,7 +71,7 @@ if [ "$1" == "install-cuquantum" ]; then
     CUDA_ARCH_FOLDER=$([ "$(uname -m)" == "aarch64" ] && echo sbsa || echo x86_64)
 
 # [>cuQuantumInstall]
-    CUQUANTUM_VERSION=25.03.0.11
+    CUQUANTUM_VERSION=25.06.0.10
```
bettinaheim, line 75: I don't expect cusparse is bundled with cuquantum, so you will need to add it to the CUDA libraries that are installed.

docker/build/assets.Dockerfile
1tnguyen: Thanks for the tip. I've made a matching change across the codebase, using `cusolver` as the reference.
`cusparse` should be equivalent to `cusolver` (and `cublas`): an indirect dependency of our `.so` files through `libcudensitymat.so`. Hence, I think we don't need to load it explicitly in the `init.py` file, similar to `cusolver`.

pyproject.toml
1tnguyen: Yea, I've missed `cusparse`. I've added it to the `dependencies` list accordingly.

scripts/configure_build.sh
1tnguyen: Indeed, I've missed this one. Added in the sweeping change across all our scripts using `cusolver` as the reference.

docker/build/devdeps.ext.Dockerfile
```diff
@@ -130,7 +130,7 @@ ENV UCX_TLS=rc,cuda_copy,cuda_ipc,gdr_copy,sm
 
 # Install CUDA
 
-ARG cuda_packages="cuda-cudart cuda-nvrtc cuda-compiler libcublas-dev libcurand-dev libcusolver"
+ARG cuda_packages="cuda-cudart cuda-nvrtc cuda-compiler libcublas-dev libcurand-dev libcusolver libcusparse"
```
bettinaheim, line 133: I think you are also missing a change to add cusparse to devdeps.manylinux.Dockerfile in the # Install CUDA section.

bettinaheim:
> /ok to test 1e9a66b7ce1b699b7c50bf6ab7bdaf2b8542d438

runtime/nvqir/cudensitymat/CMakeLists.txt
```diff
@@ -51,53 +51,73 @@ endif()
...
-    ${CUDENSITYMAT_INCLUDE_DIR}
-)
+string(REGEX MATCH "CUDENSITYMAT_PATCH ([0-9]*)" _ ${cudensitymat_header})
+set(CUDENSITYMAT_PATCH ${CMAKE_MATCH_1})
+
+set(CUDENSITYMAT_VERSION ${CUDENSITYMAT_MAJOR}.${CUDENSITYMAT_MINOR}.${CUDENSITYMAT_PATCH})
+message(STATUS "Found cudensitymat version: ${CUDENSITYMAT_VERSION}")
+
+# We need cudensitymat v0.2.0 
+if (${CUDENSITYMAT_VERSION} VERSION_GREATER_EQUAL "0.2")
```
mitchdz, line 69: Should we error if cudensitymat != 0.2?

docker/build/devdeps.ext.Dockerfile
1tnguyen: Thanks! Added in https://github.com/NVIDIA/cuda-quantum/pull/3156/commits/359e02840442bb3b0b12cb63f33851b2df8be8c8

runtime/nvqir/cudensitymat/CMakeLists.txt
1tnguyen: Normally, we don't expect breaking changes from these libraries in every version; hence, we don't want to be too strict.
This check is intended to help generate a more informative error message (rather than compiler errors) for the known incompatible versions. For example, future versions may be fully compatible, so this would allow building from source with different versions of the library.

PR #3162, opened 2025-07-07
Add an env variable to enable run for quantum devices or their emulation for testing

annagrin:
### Description
Add an environment variable to enable run for quantum devices or their emulation for testing

We disabled `cudaq::run` for quantum devices or their emulation for now, but would like to continue the work and add tests for features and examples.

Add an environment variable `CUDAQ_ENABLE_QUANTUM_DEVICE_RUN` for testing purposes until the `run` is enabled for quantum devices.


runtime/common/RunTheKernel.cpp
```diff
@@ -109,10 +109,16 @@ cudaq::details::RunResultSpan cudaq::details::runTheKernel(
   auto *circuitSimulator = nvqir::getCircuitSimulatorInternal();
   circuitSimulator->outputLog.clear();
 
+  static bool enableQuantumDeviceRun =
+      getEnvBool("CUDAQ_ENABLE_QUANTUM_DEVICE_RUN", false);
```
schweitzpgi: Why `static`? Are you trying to do an initialize once sort of thing here? You can use `std::call_once` as in `RuntimeMLIR.cpp` in that case.

(I think that is a good idea, actually. We don't want to read in an environment variable every time the program does a `run`.)

annagrin: will do, thanks!

annagrin: done!

bmhowe23 (approved):
Changes look good to me, and I'd like to get this propagated into the private repos, too. (@schweitzpgi - another rebase will probably be needed)

bmhowe23:
> Add an environment variable `CUDAQ_ENABLE_QUANTIUM_DEVICE_RUN` for testing purposes until the `run` is enabled for quantum devices.

I think that is supposed to be `CUDAQ_ENABLE_QUANTUM_DEVICE_RUN` (no `I`).

Issue #556, opened 2023-08-15
Improve compliance to the QIR Base Profile Specification


bmhowe23:
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

The [Base Profile](https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md) contains many requirements for a Base Profile QIR 1.0 program, and cuda-quantum is not currently enforcing, complying with, and verifying all of them. This top-level issue will be used to discuss items that affect multiple requirements and track improvements to cuda-quantum with respect to these specifications. New issues will be created to work specific items and track individual PRs.

The attached spreadsheet contains relatively priorities for each requirement, but they are tentative priorities and may change as needed.

[Base_Profile_Spec_Analysis_GH.xlsx](https://github.com/NVIDIA/cuda-quantum/files/12352633/Base_Profile_Spec_Analysis_GH.xlsx)

bmhowe23:
PR #546 resolved 29 elements in the spreadsheet attached to this Issue.

bmhowe23:
Status update: the [change](https://github.com/qir-alliance/pyqir/pull/251) to a commonly used library (PyQIR) was included in PyQIR v0.10.0 (released on Nov 20, 2023). This change updated PyQIR to support the spec compliant attribute names of `required_num_qubits` and `required_num_results`, so now our partner providers will automatically inherit this change whenever they update their PyQIR dependencies.

PR #3169, opened 2025-07-08
Update QIR to match the spec

annagrin:
### Description
Update generated QIR to match the QIR adaptive spec

Spec: https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Adaptive_Profile.md

Updates included:
- Replace use of `requiredQubits` and `requiredResults` by `required_num_qubits` and `required_num_results`
- Update module flags
  - remove outdated flags
  - add new flags according to https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Adaptive_Profile.md#module-flags-metadata
- Replace use of `__quantum__qis__read_result__body` by `__quantum__rt__read_result`
- Add tests

Towards: https://github.com/NVIDIA/cuda-quantum/issues/556


python/tests/mlir/target/qir_adaptive_format.py
```diff
@@ -0,0 +1,49 @@
...
+        if j > 5:
+            x(q[0])
+
+    cudaq.sample(kernel, 2, 134)
+
+
+test_module_flags()
+
+# CHECK: ; ModuleID = 'LLVMDialectModule'
+# CHECK: {{.*}}
```
schweitzpgi, line 40: nit: you can omit this CHECK

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -270,11 +362,11 @@ mlir::LogicalResult verifyQubitAndResultRanges(llvm::Module *llvmModule) {
   std::size_t required_num_qubits = 0;
   std::size_t required_num_results = 0;
   for (llvm::Function &func : *llvmModule) {
-    if (func.hasFnAttribute("entry_point")) {
```
schweitzpgi, line 273: +1. Thanks for removing all these hardcoded string values!

targettests/execution/qir_adaptive_format.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  // Test module flags
+  {
+    auto counts = cudaq::sample(kernel_module_flags, 6, 134);
+    counts.dump();
+  }
+
+  // clang-format off
+  // CHECK: ; ModuleID = 'LLVMDialectModule'
+  // CHECK: {{.*}}
```
schweitzpgi, line 47: nit: FileCheck will just skip lines so no need for this CHECK.

schweitzpgi (approved):
LGTM.

python/tests/mlir/target/qir_adaptive_format.py
annagrin: It is strangely failing without this line if another test is added… so if it is ok I would like to keep it.

targettests/execution/qir_adaptive_format.cpp
annagrin: Same as above

PR #3170, opened 2025-07-08
Clean up device from adjacency hack

nvidia-dobri:
Prelude to #2995 for readability / ease of review.

Instead of creating an adjacency from a `Device` in `MappingPrep` just to create a `Device` from that adjacency in `MappingFunc`, both `MappingPrep` and `MappingFunc` now expect a device descriptor (ideally, the same one 🙂)


include/cudaq/Support/Placement.h
```diff
@@ -70,6 +70,17 @@ class Placement {
     std::swap(phyToVr[phy0.index], phyToVr[phy1.index]);
   }
 
+  void dump(llvm::raw_ostream &os = llvm::errs()) const {
```
schweitzpgi, line 73: I think technically we're supposed to flag these with LLVM_DUMP_METHOD or something. We haven't done that anywhere, so just file it away for a future PR. :)

schweitzpgi (approved):

PR #3172, opened 2025-07-08
[testing] [Python] Follow-up to PR#3163

khalatepradnya:
When running the full test suite, the `nvidia-mqpu` target might be initialized without any actual QPUs being available. With the fix in PR #3163, this throws error instead of silently failing. Updating the tests to check QPU availability before proceeding. 
Note - these tests don't fail if run in isolation.

Fixes following crash when running full Python test suite:
```python/tests/parallel/test_mpi_mqpu.py::testMPI Fatal Python error: Floating point exception```


schweitzpgi (approved):
Thank you.

PR #3173, opened 2025-07-08
[run] [bug-fix] Asynchronous API requires proper handling of QPU ids

khalatepradnya:
Propagate `qpu_id` so it can be set correctly on the platform


1tnguyen (approved):
LGTM  👍 Thanks, @khalatepradnya

Issue #3084, opened 2025-06-18
[python, run] Returning True with int return type prints -1.


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Returning True with int return type prints -1.

Looks like a casting issue. Suggestion: use `cc.cast` in python AST bridge for all casts instead of extensions and truncations.

#### Example 1:
```python
@cudaq.kernel
def bug() -> int:
    return True

print(cudaq.run(bug, shots_count = 2))
```

Output:
[-1, -1]

#### Example 2:
```
@cudaq.kernel
def foo() -> int:
    q = cudaq.qubit()
    h(q)
    return mz(q)

results = cudaq.run(foo, shots_count=10)
```
Output: 
[0, 0, -1, -1, -1, 0, -1, 0, 0, -1]

### Expected behavior

Output in example 1:
[1, 1]

Output in example 2:
[0, 0, 1, 1, 1, 0, 1, 0, 0, 1]

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

PR #3176, opened 2025-07-09
[run] [bugfix] Use cc.cast in python bridge for operand type promotion

annagrin:
### Description
Use `cc.cast` in python bridge for operand type promotion.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/3084


sacpis:
@annagrin Seems like some tests related to this change are failing in CI.

python/tests/kernel/test_cast_kernel.py
```diff
@@ -0,0 +1,170 @@
+# ============================================================================ #
```
schweitzpgi, line 1: Looks like a couple old tests need a few tweaks.

annagrin: working on it!

annagrin: done!

PR #3177, opened 2025-07-09
Implement adjoint method for operator classes 

1tnguyen:

### DIFF
--- a/runtime/cudaq/operators.h
+++ b/runtime/cudaq/operators.h
@@ -296,6 +296,12 @@ public:
                 &parameters = {},
             bool invert_order = false) const;
 
+  // Adjoint
+
+  /// @brief Returns the adjoint of the operator.
+  /// The adjoint is computed by summing the adjoint of each product term.
+  sum_op<HandlerTy> adjoint() const;
+
   // comparisons
 
   /// @brief True, if the other value is an sum_op<HandlerTy> with
@@ -1194,6 +1200,15 @@ public:
             const std::unordered_map<std::string, std::complex<double>>
                 &parameters = {},
             bool invert_order = false) const;
+
+  // Adjoint
+
+  /// @brief Returns the adjoint of the operator.
+  /// The adjoint is computed by multiplying the adjoint of operators and
+  /// coefficient in reverse order.
+  // For example, adjoint(coeff * A * B * C) = conjugate(coeff) * adjoint(C) *
+  // adjoint(B) * adjoint(A) *
+  product_op<HandlerTy> adjoint() const;
 
   // comparisons
 
--- a/runtime/cudaq/operators/boson_op.cpp
+++ b/runtime/cudaq/operators/boson_op.cpp
@@ -318,6 +318,19 @@ std::string boson_handler::to_string(bool include_degrees) const {
     return this->op_code_to_string();
 }
 
+boson_handler boson_handler::adjoint() const {
+  boson_handler adjoint(this->degree);
+  adjoint.additional_terms = -this->additional_terms; // A => Ad, Ad => A
+  adjoint.number_offsets = this->number_offsets;
+  if (!adjoint.number_offsets.empty() && adjoint.additional_terms != 0) {
+    // If we have a 'paired' operator, e.g., N*A^k or N*Ad^k,
+    // its dagger will have the reverse order. Hence, we need to shift N back to
+    // the LHS using the relationship: A^kN = (N+k)A^k and Ad^kN = (N-k)Ad^k.
+    adjoint.number_offsets.back() += this->additional_terms;
+  }
+  return adjoint;
+}
+
 // comparisons
 
 bool boson_handler::operator==(const boson_handler &other) const {
--- a/runtime/cudaq/operators/matrix_op.cpp
+++ b/runtime/cudaq/operators/matrix_op.cpp
@@ -433,6 +433,29 @@ std::string matrix_handler::to_string(bool include_degrees) const {
     return this->op_code;
 }
 
+matrix_handler matrix_handler::adjoint() const {
+  const std::string op_code = this->op_code + "_autogen_dagger";
+  if (matrix_handler::defined_ops.find(op_code) ==
+      matrix_handler::defined_ops.end()) {
+    auto func =
+        [op = *this](const std::vector<std::int64_t> &dimensions,
+                     const std::unordered_map<std::string, std::complex<double>>
+                         &params) {
+          cudaq::dimension_map dims;
+          if (dimensions.size() != op.degrees().size())
+            throw std::runtime_error("Dimension mismatched");
+
+          for (std::size_t i = 0; i < dimensions.size(); ++i) {
+            dims[op.degrees()[i]] = dimensions[i];
+          }
+          auto originalMat = op.to_matrix(dims, params);
+          return originalMat.adjoint();
+        };
+    matrix_handler::define(op_code, {-1}, func);
+  }
+  return matrix_handler(op_code, this->targets);
+}
+
 // comparisons
 
 bool matrix_handler::operator==(const matrix_handler &other) const {
--- a/runtime/cudaq/operators/product_op.cpp
+++ b/runtime/cudaq/operators/product_op.cpp
@@ -708,6 +708,32 @@ INSTANTIATE_PRODUCT_EVALUATIONS(boson_handler);
 INSTANTIATE_PRODUCT_EVALUATIONS(fermion_handler);
 #endif
 
+// Adjoint
+
+template <typename HandlerTy>
+product_op<HandlerTy> product_op<HandlerTy>::adjoint() const {
+  // The adjoint of a product is the product of the adjoints in reverse order.
+  // The coefficient is conjugated.
+  product_op<HandlerTy> adjoint_op(this->coefficient.adjoint(),
+                                   std::vector<HandlerTy>());
+  for (auto it = this->operators.crbegin(); it != this->operators.crend();
+       ++it) {
+    adjoint_op.insert(it->adjoint());
+  }
+  return adjoint_op;
+}
+
+#define INSTANTIATE_PRODUCT_ADJOINT(HandlerTy)                                 \
+                                                                               \
+  template product_op<HandlerTy> product_op<HandlerTy>::adjoint() const;
+
+#if !defined(__clang__)
+INSTANTIATE_PRODUCT_ADJOINT(matrix_handler);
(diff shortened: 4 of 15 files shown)


python/runtime/cudaq/operators/py_boson_op.cpp
```diff
@@ -225,6 +225,11 @@ void bindBosonOperator(py::module &mod) {
           "`True`. "
           "See also the documentation for `degrees` for more detail.")
 
+      // Adjoint
+      .def(
+          "adjoint", [](const boson_op &self) { return self.adjoint(); },
```
bettinaheim: nit: you should be able to just do
```suggestion
          "adjoint", &boson_op::adjoint,
```

python/tests/operator/test_boson_op.py
```diff
@@ -40,6 +40,8 @@ def test_definitions():
     assert np.allclose(position(1).to_matrix(dims), position_matrix(3))
     assert np.allclose(momentum(1).to_matrix(dims), momentum_matrix(3))
 
+    assert np.allclose(boson.create(1).adjoint().to_matrix(dims), annihilate_matrix(3))
```
bettinaheim: very silly, but I would also add tests for all operators that op * op.adjoint() is the identity for all operators. Going only via to_matrix assumes that the to_matrix computation does not inject anything (e.g. a sign). Better to test more than less.

runtime/cudaq/spin_op.h
bettinaheim, line 1: I don't see any additional tests for spin_op.adjoint()

runtime/cudaq/operators/boson_op.cpp
```diff
@@ -318,6 +318,19 @@ std::string boson_handler::to_string(bool include_degrees) const {
...
 
+boson_handler boson_handler::adjoint() const {
+  boson_handler adjoint(this->degree);
+  adjoint.additional_terms = -this->additional_terms; // A => Ad, Ad => A
+  adjoint.number_offsets = this->number_offsets;
+  if (!adjoint.number_offsets.empty() && adjoint.additional_terms != 0) {
+    // If we have a 'paired' operator, e.g., N*A^k or N*Ad^k,
+    // its dagger will have the reverse order. Hence, we need to shift N back to
+    // the LHS using the relationship: A^kN = (N+k)A^k and Ad^kN = (N-k)Ad^k.
+    adjoint.number_offsets.back() += this->additional_terms;
```
bettinaheim: Why is only the last entry of number_offsets updated? I think you need to update all of them. 

A pretty good test would be to update the checkPreBuiltBosonOps test; for each checkEqual below  the comment `// basic in-place multiplication`, add a test that check that `got * got.adjoint()` is the identity. That should cover pretty much all cases, I think.

runtime/cudaq/operators/matrix_op.cpp
```diff
@@ -433,6 +433,29 @@ std::string matrix_handler::to_string(bool include_degrees) const {
     return this->op_code;
 }
 
+matrix_handler matrix_handler::adjoint() const {
```
bettinaheim, line 465: Hm, I am wondering if it would be better to just add a protected bool adjoint to each operator instance rather than registering a new matrix for the adjoint. The length of the string likely is giving a perf hit across the board for most of the API around operators and we already have a bool there, so alignment wise it also makes sense. Additionally, it is more readable as well and will not clash with any custom name. 

You would need to then adjust the following implementations: 
- canonical_form needs to include both the op_code and whether it is adjoint
- same for unique_id and the second half the if-statement in to_string
- constructors and assignment operators to propagate the information if needed
-  to_matrix to adjoint the matrix on the fly if needed (this is the perf price - we compute it every time)
- equality comparison

One thing to confirm if I am right: I think the statement that "if an operator commutes with all other operators then so does its adjoint" should be always correct; my thinking is that in that case, the operator has to be diagonal so its adjoint is as well. 

What do you think?

unittests/operators/fermion_op.cpp
bettinaheim, line 1: The same comment regarding where I would add tests to get decent coverage as I made for bosons applies here.

unittests/operators/matrix_op.cpp
```diff
@@ -16,6 +16,9 @@ TEST(OperatorExpressions, checkMatrixOpsUnary) {
   utils::checkEqual((-op).to_matrix({{0, 2}}),
                     -1.0 * utils::position_matrix(2));
   utils::checkEqual(op.to_matrix({{0, 2}}), utils::position_matrix(2));
+  auto adjoint = op.adjoint();
+  utils::checkEqual(adjoint.to_matrix({{0, 2}}),
+                    utils::position_matrix(2).adjoint());
 }
 
 TEST(OperatorExpressions, checkMatrixOpsConstruction) {
```
bettinaheim, line 26: It would be good to add the following tests: 
- for each checkEqual in checkPreBuiltMatrixOps, check that op * op.adjoint() equals the identity
- the same goes for checkCustomMatrixOps

That gives us some very basic coverage of adjoints of sums and products. Additionally, it would be good to add a test each thinking about a hard case to the product_op tests and one to the sum_op tests.

runtime/cudaq/operators/sum_op.cpp
```diff
@@ -546,6 +546,39 @@ INSTANTIATE_SUM_EVALUATIONS(boson_handler);
 INSTANTIATE_SUM_EVALUATIONS(fermion_handler);
 #endif
 
+// Adjoint
+
+template <typename HandlerTy>
+sum_op<HandlerTy> sum_op<HandlerTy>::adjoint() const {
+  sum_op<HandlerTy> result(false);
```
bettinaheim: I think the best thing to do here is to have the adjoint of a default sum op also be default initialized:
```suggestion
  sum_op<HandlerTy> result(this->is_default);
```

runtime/cudaq/operators/sum_op.cpp
```diff
@@ -546,6 +546,39 @@ INSTANTIATE_SUM_EVALUATIONS(boson_handler);
 INSTANTIATE_SUM_EVALUATIONS(fermion_handler);
 #endif
 
+// Adjoint
+
+template <typename HandlerTy>
+sum_op<HandlerTy> sum_op<HandlerTy>::adjoint() const {
```
bettinaheim, line 552: You could make this more efficient by special casing for operator classes that can be canonicalized. If they can be canonicalized, then the operator framework guarantees that they are (you can `assert(assert(!HandlerTy::can_be_canonicalized || this->is_canonicalized());)`). In the case of a canonicalized operator, each degree is acted on only once. Regardless of commutation relation, this means that adjointing preserves the order of operators (not need to product multiply, you can just create a vector and populate it with the adjoint of the original in the same order).

runtime/cudaq/operators/product_op.cpp
```diff
@@ -708,6 +708,32 @@ INSTANTIATE_PRODUCT_EVALUATIONS(boson_handler);
 INSTANTIATE_PRODUCT_EVALUATIONS(fermion_handler);
 #endif
 
+// Adjoint
+
+template <typename HandlerTy>
+product_op<HandlerTy> product_op<HandlerTy>::adjoint() const {
```
bettinaheim, line 714: same comment regarding special casing for operator classes that can be canonicalized applies here.

runtime/cudaq/operators.h
```diff
@@ -296,6 +296,12 @@ class sum_op {
                 &parameters = {},
             bool invert_order = false) const;
 
+  // Adjoint
+
+  /// @brief Returns the adjoint of the operator.
+  /// The adjoint is computed by summing the adjoint of each product term.
+  sum_op<HandlerTy> adjoint() const;
```
bettinaheim, line 310: Another comment regarding performance: It may be worth it to actually overload for const & and &&  to save the copy for rvalues.

python/runtime/cudaq/operators/py_boson_op.cpp
1tnguyen: Fixed for all bindings.

python/tests/operator/test_boson_op.py
1tnguyen: I've thought of it as well, but the "op * op.adjoint() is the identity" condition is only valid for unitary matrices.
In the C++ unit tests, I've added `op.adjoint().to_matrix() == op.to_matrix().adjoint()` check for all the existing test cases that we have. This should provide good coverage as we already have comprehensive test cases there.

I did try to inject the "op * op.adjoint() == identity" check condition in a few places that are applicable.

runtime/cudaq/spin_op.h
1tnguyen: Indeed, that was missing in the first commit. I've added tests for `spin_op`.

runtime/cudaq/operators/boson_op.cpp
1tnguyen: This is a bug. I've fixed it accordingly.

I added a check condition w.r.t. `adjoint()` for all test cases in `unittests/operators/boson_op.cpp`

runtime/cudaq/operators/matrix_op.cpp
1tnguyen: That's a great idea 💯 

I've modified the implementation accordingly.

unittests/operators/fermion_op.cpp
1tnguyen: I've added the check condition for `adjoint()` in all `fermion_op` test cases.

unittests/operators/matrix_op.cpp
1tnguyen: This `adjoint().to_matrix() == to_matrix().adjoint()` condition should be a reliable test condition (assuming our `complex_matrix` adjoint implementation is correct). 

I'll add some more test cases for `op * op.adjoint()`, especially with sums and products, to provide more coverage.

runtime/cudaq/operators/sum_op.cpp
1tnguyen: Fixed accordingly.

runtime/cudaq/operators/sum_op.cpp
1tnguyen: Thanks for the tip. I've implemented a special case for operator handlers that can be canonicalized.

runtime/cudaq/operators/product_op.cpp
1tnguyen: Definitely. I've added a special case for operator classes that can be canonicalized.

PR #3179, opened 2025-07-09
[core] Add variable coalescing optimization pass.

schweitzpgi:


test/Quake/coalesce.qke
```diff
@@ -0,0 +1,101 @@
...
+// CHECK-DAG:       %[[VAL_6:.*]] = arith.constant 4.000000e+00 : f64
+// CHECK-DAG:       %[[VAL_7:.*]] = arith.constant 2 : i32
+// CHECK-DAG:       %[[VAL_8:.*]] = arith.constant 3.000000e+00 : f64
+// CHECK-DAG:       %[[VAL_9:.*]] = arith.constant 1 : i32
+// CHECK-DAG:       %[[VAL_10:.*]] = cc.alloca f64
+// CHECK-DAG:       %[[VAL_11:.*]] = cc.alloca i32
+// CHECK-DAG:       %[[VAL_12:.*]] = cc.alloca f64
+// CHECK-DAG:       %[[VAL_13:.*]] = cc.alloca i32
+// CHECK-DAG:       %[[VAL_14:.*]] = cc.alloca i32
+// CHECK-DAG:       %[[VAL_15:.*]] = cc.alloca f64
```
khalatepradnya, line 76: Input had 10 occurrences of `cc.alloca` and output has only 6, so 4 variables got coalesced. It's not clear to me which ones ?
Shouldn't `a`, `c`, `e`, `g` combine to one and `b`,`d`, `f`, `h` combine to another so only 4 `alloca` (including two original ones) in outer scope and none in inner scope?

schweitzpgi: There are 10 variables and 5 scopes. The outer scope variables (2) aren't candidates. The "middle" scope contains the other 3 scopes, so the variables (2) there are live through the innermost 3 scopes. Those two are candidates, but are rejected by the algorithm (correctly).

Finally, there are 6 variables, 2 each, in 3 innermost scopes. Those 3 scopes are disjoint which means the variables in any one of those scopes are dead for the other 2 scopes. Since these variables are candidates and their live ranges are disjoint, the algorithm coalesces those variables by type. Therefore those 6 variables are fused into 2 variables (c/e/g and d/f/h) where the storage is reused in distinct scopes.

PR #3180, opened 2025-07-09
Part 1: merges changes to core and the C++ bridge to support device calls

schweitzpgi:


lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1574,6 +1514,76 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
             }
       }
 
+    if (isInNamespace(func, "driver")) {
```
amccaskey: I think from a language specification perspective, these in-kernel malloc/malloc_set/free functions are not quite ready for prime time. Still a bit to work out here.

schweitzpgi: Removed in afae659d

include/cudaq/Optimizer/Dialect/CC/CCOps.td
```diff
@@ -1799,6 +1803,88 @@ def cc_NoInlineCallOp :
...
+    mlir::CallInterfaceCallable getCallableForCallee() {
+      return getCalleeAttr();
+    }
+
+    mlir::LogicalResult verifySymbolUses(mlir::SymbolTableCollection &);
+  }];
+}
+
+
+def cc_ResolveDevicePtrOp : CCOp<"resolve_device_ptr", []> {
```
bmhowe23, line 1869: Should this be deferred to a later PR when it's ready for prime time?

lib/Frontend/nvqpp/ConvertDecl.cpp
```diff
@@ -358,6 +363,18 @@ bool QuakeBridgeVisitor::TraverseFunctionDecl(clang::FunctionDecl *x) {
   // If we're already generating code (this FunctionDecl is nested), we only
   // traverse the type, adding the function type to the type stack.
   if (builder.getBlock()) {
+    if (isInNamespace(x, "cudaq") && isInNamespace(x, "driver")) {
```
bmhowe23: Same for this, and same for line 172?

include/cudaq/Optimizer/Dialect/CC/CCOps.td
schweitzpgi: I think we can leave this. It primes the pump for the next round(s).

lib/Frontend/nvqpp/ConvertDecl.cpp
schweitzpgi: Removed 8e9b1baa

bmhowe23 (approved):
Changes LGTM. Just noting for others: more changes will be needed in order to lower the device_call, even for local simulators.

PR #3189, opened 2025-07-10
Fix devdeps tag, remove regular devdeps as unneeded

mitchdz:
This will produce a devdeps image at

http://ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu12.0-gcc11-main

Like it was before. The incorrect key was used previously, publishing the regular devdeps image, not ext.devdeps. This now publishes ext.devdeps with the extra dependencies such as cuQuantum, cuTensor, and the CUDA packages necessary.

Removes stitching the development dependencies image, as no one consumes that tag, and the build pipeline uses hashes instead.

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -511,7 +511,7 @@ jobs:
             amd64_image=$(echo "$EXT_JSON" | jq -r ".image_hash[\"$key\"]")
             echo "CUDA $cuda amd64 image: $amd64_image"
 
-            stitched_tag="ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu${cuda}-gcc11-main
+            stitched_tag="ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu${cuda}-gcc11-${branch}"
 
             refs="$arm64_image $amd64_image"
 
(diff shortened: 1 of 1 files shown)


bmhowe23 (approved):
LGTM

.github/workflows/deployments.yml
```diff
@@ -537,7 +511,7 @@ jobs:
             amd64_image=$(echo "$EXT_JSON" | jq -r ".image_hash[\"$key\"]")
             echo "CUDA $cuda amd64 image: $amd64_image"
 
-            stitched_tag="ghcr.io/nvidia/cuda-quantum-devdeps:cu${cuda}-ext-${branch}"
+            stitched_tag="ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu${cuda}-gcc11-${branch}"
```
bettinaheim, line 514: Please add this to the list of follow up items: The gcc11 hardcoded here is also something that would be nice to get from prior jobs instead.

bettinaheim (approved):

.github/workflows/deployments.yml
mitchdz: Noted.

PR #3190, opened 2025-07-10
Introduce `cudaq::contrib`

nvidia-dobri:
This introduces the `contrib` namespace within CUDA-Q for new, useful, **fully implemented and functional** experimental features. These features, often by external contributors, might not have the stability and performance guarantees we want out of the main `cudaq` namespace in C++ but are useful to a limited part of our users anyway.

The only functionality in it right now are `cudaq::contrib::get_unitary_cmat` and `cudaq::contrib::draw`, with the latter moved from `cudaq::draw` via deprecation warnings.


python/runtime/cudaq/algorithms/py_draw.cpp
```diff
@@ -17,7 +17,7 @@
 
 namespace cudaq {
 
-namespace details {
+namespace contrib {
```
bettinaheim: I think this change was accidental? getKernelLaunchParameters does not belong in contrib.
Edit: I guess not accidental. We should still keep it in details.

docs/sphinx/api/languages/cpp_api.rst
```diff
@@ -163,8 +163,6 @@ Common
 .. doxygenfunction:: cudaq::range(ElementType total)
 .. doxygenfunction:: cudaq::range(ElementType begin, ElementType end, ElementType step)
 
-.. doxygenfunction:: cudaq::draw(QuantumKernel &&kernel, Args&&... args)
```
bettinaheim, line 171: We should still keep the docs, just now for the appropriate namespace.

docs/sphinx/specification/cudaq/namespace.rst
```diff
@@ -3,9 +3,14 @@ Namespace and Standard
 
 **[1]** The CUDA-Q language is a library-based extension for C++. As such, all
 pertinent data structures and library functions exist within the CUDA-Q
-namespace. This namespace is denoted :code:`::cudaq`. 
+namespace. This namespace is denoted :code:`::cudaq`.
 
-**[2]** CUDA-Q is a C++ library-based language extension adherent to the `C++20 <https://en.cppreference.com/w/cpp/20>`_
+**[2]** The :code:`cudaq::contrib` namespace contains features contributed by
```
bettinaheim, line 8: My suggested phrasing: 

The :code:`cudaq::contrib` namespace contains features contributed by the larger community. While these features provide useful functionality and are maintained across releases, they may be more limited in functionality and may not cover some scenarios. If you encounter issues or would like to request more extensive features, please file a GitHub issue and we will consider them for a future release.

docs/sphinx/api/languages/cpp_api.rst
nvidia-dobri: Sure - the namespace is excluded by doxy, though, so it will make the exclusion a bit messier.

docs/sphinx/specification/cudaq/namespace.rst
nvidia-dobri: I do like this much better, too. Thanks for the suggestion!

docs/sphinx/api/languages/cpp_api.rst
nvidia-dobri: Actually, it wasn't excluded, just needed to have the namespace in the file as well. Fixed.

python/README.md
bettinaheim, line 1: This file was accidentally checked in.  The fix is to add this to .gitignore (this is generated off of the template Readme.in as part of the build) - I meant to fix that but never got to it.

bettinaheim (approved):
Approving - please remove the accidentally checked in readme before merging

python/README.md
nvidia-dobri: Yeah, it almost got me a few times, it actually got me this time.

nvidia-dobri:
> /ok to test 39f871475a33b73b2f20c097546cfe5ed9155765

PR #3191, opened 2025-07-10
[run][Python] Set default shots counts to 1000

khalatepradnya:
* Enable run_async tests

### DIFF
--- a/docker/build/assets.Dockerfile
+++ b/docker/build/assets.Dockerfile
@@ -191,6 +191,7 @@ RUN cd /cuda-quantum && source scripts/configure_build.sh && \
         cublas_version=11.11 && \
         cusparse_version=11.7 && \
         sed -i "s/-cu12/-cu11/g" pyproject.toml && \
+        sed -i "s/-cuda12/-cuda11/g" pyproject.toml && \
         sed -i -E "s/(nvidia-cublas-cu[0-9]* ~= )[0-9\.]*/\1${cublas_version}/g" pyproject.toml && \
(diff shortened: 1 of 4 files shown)


schweitzpgi (approved):
Thanks, Pradnya.

python/tests/kernel/test_run_async_kernel.py
```diff
@@ -71,7 +71,7 @@ def test_run_async():
         assert non_zero_count > 0
 
 
-## https://github.com/NVIDIA/cuda-quantum/issues/3095
+## Ref: https://github.com/NVIDIA/cuda-quantum/issues/3193
```
bettinaheim: I generally would prefer to mark a test as skipped rather than comment it out. I am rather averse to commented out code in the main branch. :) That would likely also have made the diff a lot more readable in this case.

amccaskey:
I like the consistency part here, but I had one thought though - since in simulation these kernels have to be run every single shot, should we instead opt for something lower like 100, in order to optimize for user-expectations on runtime?

python/tests/kernel/test_run_async_kernel.py
khalatepradnya: Fixed in fd5974fbedd7ab6ccfd33443363958d324a3258f

khalatepradnya:
> I like the consistency part here, but I had one thought though - since in simulation these kernels have to be run every single shot, should we instead opt for something lower like 100, in order to optimize for user-expectations on runtime?

Sure. Fixed in f015668ebcd3a8b0013836d5a998979bcc599fef

PR #3192, opened 2025-07-10
Part 2: merge more device_call code to main.

schweitzpgi:
#3180 must be merged first!


include/cudaq/Support/TargetConfig.h
```diff
@@ -116,7 +116,8 @@ struct BackendFeatureMap {
 };
```
amccaskey, line 116: currently, this header is not being installed. it will be needed as we progress. can we go ahead and add the cmake to get it installed now?

test/Quake/device_call.qke
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                 //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -canonicalize -distributed-device-call -cse %s | FileCheck %s
+
+func.func private @incrementInt(!cc.ptr<i32>) attributes {"cudaq-devicecall"}
```
amccaskey, line 11: is this incrementInt function left around from other tests? I don't see it used anywhere

include/cudaq/Support/TargetConfig.h
schweitzpgi: Is it part of the runtime then? We'll need to move it, if so.

test/Quake/device_call.qke
schweitzpgi: This test is a fraction of what it was.

PR #3204, opened 2025-07-14
Compatibility fixes for cupy 13.5+

1tnguyen:
### Symptom

Cupy integration tests (e.g., test_cupy_integration.py, dynamics tests/examples) failed with exception

```
TypeError: Accessing a CuPy ndarry on CPU is not allowed except when using system memory (on HMM or ATS enabled systems, need to set CUPY_ENABLE_UMP=1) or managed memory
```

### Root cause

This exception is triggered by the `PyObject_GetBuffer` call when we call `.request()` on a `py::buffer`.

For cupy, it will throw: 

https://github.com/cupy/cupy/blob/3ce2a11076db8ec26bd2a9ec617d0a1806b69951/cupy/_core/core.pyx#L408-L415

These `py::buffer` overloads are **not** intended for CuPy arrays. We have a separate `py::object` overload for CuPy arrays.

With CuPy v13.5+, Pybind11 will pick this `py::buffer` overload for CuPy arrays.

### Fixes

Add a helper to handle CuPy arrays in these `py::buffer` overloads

We still keep the generic `py::object` overload to support older CuPy versions.

### Tested by:

Running test_cupy_integration.py and dynamics tests.

Related to #3203 (where we constrained cupy versions to mitigate this)

### DIFF
--- a/python/runtime/cudaq/algorithms/py_state.cpp
+++ b/python/runtime/cudaq/algorithms/py_state.cpp
@@ -182,6 +182,58 @@ state pyGetStateLibraryMode(py::object kernel, py::args args) {
   });
 }
 
+static py::buffer_info getCupyBufferInfo(py::buffer cupy_buffer) {
+  // Note: cupy 13.5+ arrays will bind (overload resolution) to a py::buffer
+  // type. However, we cannot access the underlying buffer info via a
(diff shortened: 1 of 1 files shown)


mitchdz:
Pulling this in to https://github.com/NVIDIA/cuda-quantum/pull/3544

Issue #3160, opened 2025-07-07
Bug: Mid-Circuit Measurement Fails When qvector Is Passed Between Kernels


hrushikesh890:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When performing mid-circuit measurements in CUDA-Q, passing a qvector as a parameter into another kernel results in incorrect behavior. Specifically, measurement outcomes are not properly registered or propagated, which breaks control-flow logic such as conditionals based on measurement results.

### Steps to reproduce the bug

```
import cudaq

@cudaq.kernel
def kernel(register: cudaq.qview):
    for i in range(4):
        if i % 2 == 0:
            x(register[I])

        m = mz(register[I])
        reset(register[I])

        if m:
            x(register[I])
        else:
            h(register[I])

@cudaq.kernel
def kernel_top():
    qr = cudaq.qvector(4)
    kernel(qr)

@cudaq.kernel
def kernel_no_meas():
    register = cudaq.qvector(4)
    for i in range(4):
        if i % 2 == 0:
            x(register[I])

        m = mz(register[I])
        reset(register[I])

        if m:
            x(register[I])
        else:
            h(register[I])

print(cudaq.sample(kernel_no_meas))
print(cudaq.sample(kernel_top))
```

### Expected behavior

Both kernel_no_meas and kernel_top implement the same logic and should return the same output distribution. The mid-circuit measurement and conditional branching based on the result should behave identically.

✅ Monolithic Kernel (kernel_no_meas) **(Expected Correct output)**

```
{
  "__global__": { "1010": 262, "1011": 261, "1110": 238, "1111": 239 },
  "m": { "1010": 1000 }
}
```

❌ With qvector as Parameter (kernel_top) (Incorrect)

```
{
  "__global__": { "1111": 1000 },
  "m": { "0": 1000 }
}
```

Notice how the measured mid circuit result output is always 0

[...]

PR #3217, opened 2025-07-17
Propagate some conditional feedback metadata from callees to callers

annagrin:
### Description
Propagate some conditional feedback metadata from callees to callers.

In python, we add metadata to the quake IR after it is created, but run an inliner later, and the metadata is not present on the caller if only the callee had it. The current fix propagates metadata from callees to all the callers up the callgraph.

- Add a pass to propagate metadata to callers
- Add tests

Closes: https://github.com/NVIDIA/cuda-quantum/issues/3160


schweitzpgi:
Should this one be closed in favor of #3219 ?

PR #3219, opened 2025-07-17
Propagate conditional feedback metadata from callees to callers

annagrin:
### Description
Propagate conditional feedback metadata from callees to callers.

In python, we add metadata to the quake IR after it is created, but run an inliner later, and the metadata is not present on the caller if only the callee had it. The current fix propagates metadata from callees to all the callers up the callgraph.

- Add a pass to propagate metadata to callers
- Add tests

Closes: https://github.com/NVIDIA/cuda-quantum/issues/3160


lib/Optimizer/Transforms/QuakePropagateMetadata.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "cudaq/Optimizer/CallGraphFix.h"
+#include "cudaq/Optimizer/Dialect/CC/CCOps.h"
+#include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
+#include "cudaq/Optimizer/Transforms/Passes.h"
+#include "cudaq/Todo.h"
+#include "llvm/ADT/DepthFirstIterator.h"
+#include "llvm/ADT/PostOrderIterator.h"
+#include "llvm/Support/Debug.h"
+#include "mlir/Analysis/CallGraph.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
```
schweitzpgi: Do we really need all of these headers for this pass? It doesn't even need the control flow dialect, for instance. Or LLVM dialect or types. Maybe spend a couple minutes removing as many of these as you can...

annagrin: done, thanks!

schweitzpgi (approved):
This is a great addition. Thanks, Anna.

PR #3221, opened 2025-07-18
Move to miniforge

mitchdz:
Move from miniconda to minforge.


mitchdz:
> /ok to test f1ded4f

PR #3223, opened 2025-07-18
Add env variable for updated QIR

annagrin:
### Description
Add env variable for updated QIR

Some backend are still expecting our current QIR format, but https://github.com/NVIDIA/cuda-quantum/pull/3169 changes the output to the current QIR spec under development. 

- Output current format by default
- Add environment variable `CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT` to enable QIR format that is under development.


include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -87,7 +87,12 @@ def QIRToQIRProfilePrep : Pass<"qir-profile-prep", "mlir::ModuleOp"> {
     conflicts when rewriting matching DAGs independently.
   }];
 
-  let constructor = "cudaq::opt::createQIRProfilePreparationPass()";
+let options = [
+
```
schweitzpgi:
```suggestion
```

schweitzpgi:
Instead of passing a `bool` flag everywhere, we should refactor this code to use the already existing string argument and add a new version field to it. So instead of using an ambiguous `"qir-adaptive"`, it should pass something like `"qir:adaptive:0.2"` to give the spec, profile, and version. We will want to extend this for adaptive profile to include a features list at some point as well: `"qir:adaptive[integer]:0.2"` for example might enable integer LLVM operations.

Also, it's not clear what happens if the environment variable and the command-line option are (maliciously) made to disagree. Instead of an error-prone environment variable, it is probably better to place the QIR spec version directly in the IR as a ModuleOp attribute. This attribute should then be used at JIT time, etc. The Python bridge also would need this information as a property to place the proper attribute on the Quake IR.

python/runtime/cudaq/algorithms/py_translate.cpp
```diff
@@ -27,6 +28,9 @@ std::string getASM(const std::string &name, MlirModule module,
 std::string pyTranslate(py::object &kernel, py::args args,
                         const std::string &format) {
 
+  bool qirVersionUnderDevelopment =
+      getEnvBool("CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT", false);
```
bmhowe23, line 32: We're going to merge this for now under the premise that #3312 will include follow-up work related to these environment variables.

bmhowe23: Marking as resolved for now, but will make a top-level comment so it is easier to see this comment in the future.

Issue #2937, opened 2025-05-21
Wrong simulation result with the tensornet/tensornet-mps target


nadbp:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

For my specific circuit shown in the code below, there should be 16 items in the count result, but there are only 8 items with the tensornet or tensornet-mps target. With the nvidia target, I can get the correct simulation result. I know the tensornet target outputs a sample result, but it seems that the result is far from the correct one. Please let me know if this issuce could be fixed by some simple configuration.

### Steps to reproduce the bug

The Python code to reproduce the bug:

import cudaq
import numpy as np

qubit_count = 20

cudaq.set_target("tensornet")

kernel = cudaq.make_kernel()

qubits = kernel.qalloc(qubit_count)

depth = qubit_count // 5
for i in range(0, depth):
    kernel.h(qubits[i*5])
    for j in range(4):
        kernel.cx(qubits[i*5+j], qubits[i*5+j+1])

kernel.mz(qubits)

result = cudaq.sample(kernel, shots_count=10240)
result.dump()

The wrong simulation result is:
{ 00000000000000000000:1233 00000000001111111111:1311 00000111110000000000:1281 00000111111111111111:1283 11111000000000000000:1309 11111000001111111111:1241 11111111110000000000:1244 11111111111111111111:1338 }

### Expected behavior

{ 00000000000000000000:637 00000000000000011111:661 00000000001111100000:597 00000000001111111111:650 00000111110000000000:640 00000111110000011111:638 00000111111111100000:613 00000111111111111111:661 11111000000000000000:634 11111000000000011111:659 11111000001111100000:634 11111000001111111111:633 11111111110000000000:653 11111111110000011111:640 11111111111111100000:642 11111111111111111111:648 }

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**:  0.10.0 and 0.11.0
[...]

PR #3225, opened 2025-07-21
Fixes for cuTensorNet v2.8.0

1tnguyen:
### Description

[cuTensorNet v2.8.0](https://docs.nvidia.com/cuda/cuquantum/latest/cutensornet/release-notes.html#cutensornet-v2-8-0) introduces a change to the way to use the sampler:

>  "Fixed a bug in the sampling functionalities cutensornetSamplerSample() where setting CUTENSORNET_SAMPLER_CONFIG_DETERMINISTIC can lead to incorrect results for large tensor network state."

This requires us to update the sampling code in the CUDA Quantum to always configure the seed before calling `cutensornetSamplerSample`.

Also, add a test case for #2937, which is fixed in cuTensorNet v2.8.0.

### DIFF
--- a/runtime/nvqir/cutensornet/tensornet_state.inc
+++ b/runtime/nvqir/cutensornet/tensornet_state.inc
@@ -262,16 +262,6 @@ TensorNetState<ScalarType>::prepareSample(
     HANDLE_CUTN_ERROR(cutensornetSamplerConfigure(
         m_cutnHandle, sampler, CUTENSORNET_SAMPLER_CONFIG_NUM_HYPER_SAMPLES,
         &numHyperSamples, sizeof(numHyperSamples)));
-
-    // Generate a random seed from the backend simulator's random engine.
-    // Note: Even after a random seed setting at the user's level,
-    // consecutive `cudaq::sample` calls will still return different results
-    // (yet deterministic), i.e., the seed that we send to cutensornet should
-    // not be the user's seed.
-    const int32_t rndSeed = m_randomEngine();
-    HANDLE_CUTN_ERROR(cutensornetSamplerConfigure(
-        m_cutnHandle, sampler, CUTENSORNET_SAMPLER_CONFIG_DETERMINISTIC,
-        &rndSeed, sizeof(rndSeed)));
   }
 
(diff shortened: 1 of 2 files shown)


1tnguyen:
> /ok to test 4189dcf386973667a3efe5d82c853e642ddd61da

runtime/nvqir/cutensornet/tensornet_state.inc
```diff
@@ -345,11 +335,23 @@ TensorNetState<ScalarType>::executeSample(
...
+      // Generate a random seed from the backend simulator's random engine.
+      // Note: Even after a random seed setting at the user's level,
+      // consecutive `cudaq::sample` calls will still return different results
+      // (yet deterministic), i.e., the seed that we send to cutensornet should
+      // not be the user's seed.
+      const int32_t rndSeed = m_randomEngine();
+      HANDLE_CUTN_ERROR(cutensornetSamplerConfigure(
+          m_cutnHandle, sampler, CUTENSORNET_SAMPLER_CONFIG_DETERMINISTIC,
+          &rndSeed, sizeof(rndSeed)));
+    }
```
sacpis, line 353: Seems a bit different as now we are configuring for each shot. I think experts should weight in here.

khalatepradnya (approved):
LGTM.

runtime/nvqir/cutensornet/tensornet_state.inc
1tnguyen: We only do shot-by-shot when there are noise channels. Otherwise, it will sample all the shots at once.
There is a change in the way deterministic config is reset when `cutensornetSamplerSample` is executed in the new release, thus requiring this change.

1tnguyen:
> /ok to test b94f8a95bb5cfcb3d6c1c11cca755efaa6ea5621

1tnguyen:
> /ok to test f3f1dc502762c98e28e664a7f613ed70c9bf1e9c

bmhowe23:
I will directly merge this because nothing significant has changed in the code since the last successful CI.

PR #3237, opened 2025-07-22
Nexus job creation

khalatepradnya:


runtime/common/Executor.cpp
```diff
@@ -20,7 +20,9 @@ details::future Executor::execute(std::vector<KernelExecution> &codesToExecute,
 
   // Create the Job Payload, composed of job post path, headers,
   // and the job json messages themselves
-  auto [jobPostPath, headers, jobs] = serverHelper->createJob(codesToExecute);
+  auto payload = serverHelper->createJob(codesToExecute);
+  auto fullPayload = cudaq::toServerJobPayload(payload);
+  auto [jobPostPath, headers, jobs, cookies] = fullPayload;
```
1tnguyen: We added `getCookies` method to `ServerHelper` (returns empty by default). Perhaps we can keep the `createJob` API as-is and use `serverHelper->getCookies()` here to get the cookies?

khalatepradnya: Thanks! Fixed in d807906718974e800b95018ed80d5f3446e86d06.

PR #3243, opened 2025-07-23
Adding an error for invalid target attribute

### LINKED ISSUE #3238: cudaq.set_target silently fails when user provides an invalid arg
Just noticed this: when I was using python API `cudaq.set_target`, cudaq wouldn't complain if I had a typo in the option I provided. 

Example:
Correct argument should be `cudaq.set_target('nvidia', option='mqpu')`,
but I had this `cudaq.set_target('nvidia', options='mqpu')` (**options** instead of **option**), and it seems like `cudaq.set_target` simply ignored the `option` setting without throwing an error.

It would be helpful if cudaq can thrown an "unknown option given" type of error.

sacpis:
Adding an error for invalid target attribute options.

Closes #3238.

### DIFF
--- a/python/runtime/cudaq/target/py_runtime_target.cpp
+++ b/python/runtime/cudaq/target/py_runtime_target.cpp
@@ -15,6 +15,32 @@
 
 namespace cudaq {
 
+std::map<std::string, std::string>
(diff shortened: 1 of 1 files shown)


sacpis:
Closing this PR. Messed up signing the commits.

PR #3247, opened 2025-07-23
Error on non-callable types in a kernel

### LINKED ISSUE #3181: Silent failure when applying pauli_word argument as a gate in a kernel
### Required prerequisites

### Describe the bug

When a `cudaq.pauli_word` is passed as an argument to a CUDA-Q kernel and then applied as a gate to a qubit, the operation is silently ignored. It does not raise an error, but it also does not apply the specified Pauli gate, 

Steps to Reproduce:

1. Define a kernel that accepts a `cudaq.pauli_word` as an argument.
2. Inside the kernel, apply this pauli_word to a qubit as if it were a standard gate.
3. Execute the kernel using `cudaq.sample` or visualize it with `cudaq.draw`.

```
import cudaq

@cudaq.kernel()
def kernel(op: cudaq.pauli_word):
    q = cudaq.qvector(2)
    x(q[1])
    op(q[0])


print(cudaq.draw(kernel, cudaq.pauli_word("X")))
result = cudaq.sample(kernel, cudaq.pauli_word("X"))
print(result)
```

### Expected behavior

A valid alternative, if this functionality is not intended to be supported, would be for the program to fail with a clear error message instead of failing silently.

Not a regression

### Environment

- **CUDA-Q version**: CUDA-Q Version cu11-latest (https://github.com/NVIDIA/cuda-quantum 615d91910310605c83ea59f6afe6e7ae6dfecd28)

### Suggestions

_No response_

sacpis:
Throwing error on non-callable types in a kernel both in the kernel and the builder mode.

Closes #3181 

# Output

```
cudaq.kernel.ast_bridge.CompilerError: test.py:7: error: `op` is not a callable type, found symbol of type !cc.charspan
         (offending source -> op(q[0]))
```

### DIFF
--- a/.github/workflows/publishing.yml
+++ b/.github/workflows/publishing.yml
@@ -959,20 +959,24 @@ jobs:
           dpkg -i cuda-keyring_1.1-1_all.deb
           cuda_version_suffix="$(echo ${{ matrix.cuda_version }} | tr . -)"
           apt-get update
-          # Install cublas first so it can be found. For some reason, cublas is
-          # not found if we install it after the other CUDA libraries.
-          apt-get install -y --no-install-recommends \
-            libcublas-$cuda_version_suffix
-          if [ $(echo ${{ matrix.cuda_version }} | cut -d . -f1) -gt 11 ]; then 
+          if [ $(echo ${{ matrix.cuda_version }} | cut -d . -f1) -gt 11 ]; then
             apt-get install -y --no-install-recommends \
-            libnvjitlink-$cuda_version_suffix
-          fi
-          apt-get install -y --no-install-recommends \
-            libcurand-$cuda_version_suffix \
             cuda-cudart-$cuda_version_suffix \
-            libcusolver-$cuda_version_suffix \
+            cuda-nvrtc-$cuda_version_suffix \
+            libnvjitlink-$cuda_version_suffix \
+            libcurand-$cuda_version_suffix \
+            libcublas-$cuda_version_suffix \
             libcusparse-$cuda_version_suffix \
-            cuda-nvrtc-$cuda_version_suffix
+            libcusolver-$cuda_version_suffix
+          else
+            apt-get install -y --no-install-recommends \
+            cuda-cudart-$cuda_version_suffix \
+            cuda-nvrtc-$cuda_version_suffix \
+            libcurand-$cuda_version_suffix \
+            libcublas-$cuda_version_suffix \
+            libcusparse-$cuda_version_suffix \
+            libcusolver-$cuda_version_suffix
+          fi
 
       - name: Runtime dependencies (dnf)
         if: startsWith(matrix.os_image, 'redhat')
--- a/runtime/common/RuntimeMLIRCommonImpl.h
+++ b/runtime/common/RuntimeMLIRCommonImpl.h
@@ -309,6 +309,7 @@ mlir::LogicalResult verifyLLVMInstructions(llvm::Module *llvmModule,
                                            bool integerComputations,
                                            bool floatComputations) {
   bool isAdaptiveProfile = !isBaseProfile;
+  bool allowAllInstructions = getEnvBool("QIR_ALLOW_ALL_INSTRUCTIONS", false);
   for (llvm::Function &func : *llvmModule)
     for (llvm::BasicBlock &block : func)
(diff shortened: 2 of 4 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2189,6 +2189,10 @@ def bodyBuilder(iterVal):
                     callable = cc.CallableFuncOp(callableTy, val).result
                     func.CallIndirectOp([], callable, values)
                     return
+                else:
+                    self.emitFatalError(
+                        f"`{node.func.id}` is not a callable type, found symbol of type {val.type}",
```
khalatepradnya:
```suggestion
                        f"`{node.func.id}` object is not callable, found symbol of type {val.type}",
```

python/tests/kernel/test_kernel_features.py
```diff
@@ -2401,6 +2401,19 @@ def caller():
     assert "1010" in mCounts
 
 
+def test_error_on_non_callable_type():
+
+    @cudaq.kernel(verbose=True)
+    def kernel(op: cudaq.pauli_word):
+        q = cudaq.qvector(2)
+        x(q[1])
+        op(q[0])
```
khalatepradnya, line 2410: Q: Do we have this issue in builder mode?

python/tests/kernel/test_kernel_features.py
```diff
@@ -2401,6 +2401,19 @@ def caller():
     assert "1010" in mCounts
 
 
+def test_error_on_non_callable_type():
+
+    @cudaq.kernel(verbose=True)
```
khalatepradnya: Do we need the `verbose=True` annotation for the test? May fill up the console unnecessarily...

sacpis: Thanks for catching. Leftover from debugging. Removing it.

python/tests/kernel/test_kernel_features.py
poojarao8: I believe something similar, when I tried to do this

```
import cudaq
kernel, op = cudaq.make_kernel(cudaq.pauli_word)
q = kernel.qalloc(2)
kernel.op(q[0])

result = cudaq.sample(kernel, cudaq.pauli_word("X"))
```

I get the error: `RecursionError: maximum recursion depth exceeded while calling a Python object`

sacpis: Seems like we have not validated as what all arg types are allowed in the make_kernel function. This should not accept cudaq.pauli_word for now.

python/cudaq/kernel/ast_bridge.py
sacpis: Fixed in a561825.

python/tests/kernel/test_kernel_features.py
sacpis: Fixed in a561825.

python/tests/kernel/test_kernel_features.py
sacpis: Fixed in a561825.

sacpis: Working on fixing the callable type in builder mode.

sacpis: Calling op on kernel like `kernel.op` is fixed in [fe9d80f](https://github.com/NVIDIA/cuda-quantum/pull/3247/commits/fe9d80fd1102d888e5436b11349cff4859e434f7). Meaning we are throwing an error now for such calls.

marwafar: @sacpis would it be possible to check if this PR will be applied for this [krylov example](https://nvidia.github.io/cuda-quantum/latest/applications/python/krylov.html) too? It would make the code cleaner if we can simply apply the spin hamiltonian inside a quantum kernel.

sacpis: > @sacpis would it be possible to check if this PR will be applied for this [krylov example](https://nvidia.github.io/cuda-quantum/latest/applications/python/krylov.html) too? It would make the code cleaner if we can simply apply the spin hamiltonian inside a quantum kernel.

@marwafar Let me try to get an error thrown at such calls. Then I will work on making it work. I will keep in the loop on updates.

PR #3248, opened 2025-07-23
Try installing cublas first

mitchdz:
We have seen strange behavior in CI that is not reproducible locally.

The error is:
```
./a.out: error while loading shared libraries: libcublas.so.12: cannot open shared object file: No such file or directory
```
e.g. https://github.com/NVIDIA/cuda-quantum/actions/runs/16464800187/job/46543007819

There is a theory that installing cublas first will fix this.


mitchdz:
> /ok to test a0d04a0

PR #3257, opened 2025-07-24
Separating installation libs for cuda version 11 and version 12

sacpis:
Separating installation steps for cuda version 11 and version 12, as having a condition for `libnvjitlink` is causing apt-get install to install libs in diff order. 

Apt determines the dependencies, build a plan to install packages and dependencies and installs them in an order.


1tnguyen (approved):
👍

PR #3260, opened 2025-07-25
Populate sequential data for all registers

1tnguyen:
### Description

- Populate the sequential data for all sub-registers. Use the sequential `counts` data.

- Add a result parsing test using data collected offline.

- Add corresponding logic to the mock server.


unittests/backends/quantinuum/QuantinuumUtilTester.cpp
```diff
@@ -0,0 +1,45 @@
...
+    if (reg == "r0000") {
+      EXPECT_EQ(result.count("1", reg), 1000);
+      for (const auto &shot : result.sequential_data(reg)) {
+        EXPECT_EQ(shot, "1");
+      }
+    } else {
+      EXPECT_EQ(result.count("0", reg), 1000);
+      for (const auto &shot : result.sequential_data(reg)) {
+        EXPECT_EQ(shot, "0");
+      }
```
khalatepradnya, line 42: Nice test! 👍🏽

runtime/cudaq/platform/default/rest/helpers/quantinuum/QuantinuumHelper.h
```diff
@@ -0,0 +1,102 @@
...
+  cudaq::CountsDictionary globalCounts;
+  std::vector<std::string> globalSequentialData;
+  globalSequentialData.reserve(numShots);
+  for (const auto &outcomeArray : shotsOutcomeArray) {
+    // Convert the outcome array to a bit string
+    const auto bitString = outComeArrayToBitString(outcomeArray, numBits);
+    // Populate the register results
+    for (std::size_t i = 0; i < registerNames.size(); ++i) {
+      const auto bit = bitString[i];
+      registerResults[i][std::string{bit}]++;
```
khalatepradnya, line 80: Q: Do we need to check `assert(bitString.length() == outputNames.size());` ? (I had disabled it in the previous implementation)

1tnguyen: Good point! Yea, we should add it back (https://github.com/NVIDIA/cuda-quantum/pull/3260/commits/dcd7033edcf6e8f1c4ac5ba8913686ed36899e06)

PR #3263, opened 2025-07-25
[DNM] Adding a check to fetch the available architecture for IQM

sacpis:
Adding a check to fetch the available architecture for IQM and only submit the nightly integration test job if we support that architecture.


sacpis:
Please do not merge this PR yet.

Issue #3264, opened 2025-07-25
cudaq::to_integer call on a copy of a vector does not convert to QIR successfully


annagrin:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Using `cudaq::to_integer` crashes the compiler

### Steps to reproduce the bug

```
#include <cudaq.h>
#include <iostream>
#include <string>
#include <vector>

// Run with: CUDAQ_ENABLE_QUANTUM_DEVICE_RUN=1 nvq++ -fenable-cudaq-run --target quantinuum --emulate targettests/execution/tmp.cpp -o t.x && CUDAQ_DUMP_JIT_IR=1 CUDAQ_ENABLE_QUANTUM_DEVICE_RUN=1 ./t.x

__qpu__ uint64_t foo() {
  cudaq::qvector v(2);
  x(v);
  auto results = mz(v);

  std::vector<cudaq::measure_result> copy(2);
  int i = 0;
  for (auto s: results) {
    copy[i] = s;
  }
  return cudaq::to_integer(copy);
}

int main() {
  auto results = cudaq::run(1, foo);
  std::cout << results[0] << std::endl;
  return 0;
}
```

Results in:
```
; ModuleID = 'LLVMDialectModule'
source_filename = "LLVMDialectModule"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

%Qubit = type opaque
%Result = type opaque

@cstr.69363400 = private constant [4 x i8] c"i64\00"

define i64 @__nvqpp__mlirgen__function_foo._Z3foov() local_unnamed_addr #0 {
"0":
  tail call void @__quantum__qis__x__body(%Qubit* null)
  tail call void @__quantum__qis__x__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
[...]

PR #3266, opened 2025-07-25
Make to_integer intrinsic LLVM opt friendly

annagrin:
### Description
Make to_integer intrinsic LLVM opt friendly

Currently the casting from `i1*` to `i8*` of the vector data prevents some LLVM opts from optimizing the code (see example and the investigation notes in the issue below).

Update the intrinsic code to cast the elements of the vector instead of pointers to data.

Closes: https://github.com/NVIDIA/cuda-quantum/issues/3264


targettests/execution/to_integer_emulation.cpp
```diff
@@ -0,0 +1,89 @@
...
+    auto r = cudaq::run(1, copy_to_integer, test);
+    if (r[0] != 16777215) {
+      printf("1111 1111 1111 1111 1111 1111 1111 1111 1111 has to map to "
+             "16777215, but got %lu.\n",
+             r[0]);
+      return 1;
+    }
+  }
+
+  // FIXME: Fails on `--target quantinuum --emulate`:
```
bmhowe23, line 72: Does this still fail after applying the code changes in the PR, or do the code changes in this PR fix this commented out test?

annagrin: It is a new test that still fails after the PR (llvm opts are bailing out)

schweitzpgi (changes requested):
This is removing the code that takes advantage of knowledge about the layout of a `vector<bool>` on the device and introducing more casts. Since the LLVM optimizer doesn't seem to deal with `i1` values in memory well (or at all), it's hard to see how this isn't a regression. I'll try some experiments...

More specifically: the data in memory _is_ `i8` (bytes). So the description above has it backwards. It isn't casting to `i8`, the problems are casting pointers to bytes (LLVM understands those) to pointers to bits (LLVM doesn't understand those well).

schweitzpgi:
I think these issues have been resolved now in other PRs. Closing this one.

Issue #3270, opened 2025-07-26
Quantinuum nightly integration tests are failing


sacpis:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Following Quantinuum nightly integration tests are failing

1. qspan_slices.cpp: https://github.com/NVIDIA/cuda-quantum/actions/runs/16513124788/job/46698768334#step:8:654

   Error: 
   terminate called after throwing an instance of 'std::runtime_error'
     what():  Job failed to execute msg = [1000: Compile error: -1 Duplicate result tag: result]

2. sudoku_2x2-reg_name.cpp: https://github.com/NVIDIA/cuda-quantum/actions/runs/16513124788/job/46698768334#step:8:860

   Error:
   terminate called after throwing an instance of 'std::runtime_error'
     what():  Job failed to execute msg = [1000: Compile error: -1 Duplicate result tag: groverQubits]

This is because of the QIR missing the expansion of measurements pass. Seems like this pass was removed from quantinuum.yml with [PR 3130](https://github.com/NVIDIA/cuda-quantum/pull/3130/files#diff-e412df4af1913a62f6b4f62b9d8de065e5aba5592e8f64dbadacef992c023cc2).

Here is the QIR for the qspan_slices.cpp

```
source_filename = "LLVMDialectModule"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

%Qubit = type opaque
%Result = type opaque

@cstr.726573756C7400 = private constant [7 x i8] c"result\00"

define void @__nvqpp__mlirgen__function_foo._Z3foov() local_unnamed_addr #0 {
"0":
  tail call void @__quantum__qis__x__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*))
[...]

PR #3276, opened 2025-07-28
[bug-fix] Restore the pass to expand measurements for `quantinuum` target

khalatepradnya:
### Description
Addresses issue #3270 

Follow-up to PR #3130 

The order of passes should be - expand measurements, loop unrolling and then update register names (the last two are part of classical optimization pipeline).


sacpis:
@khalatepradnya Seems like you would have to add this pass to the config file as well
https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/TargetConfig/RegressionValidation/quantinuum.config#L21

PR #3281, opened 2025-07-29
Fix IQM arch name

sacpis:
Fix IQM arch name as per the new IQM chip convention.

Adonis -> Crystal 5
Apollo -> Crystal 20
Aphrodite -> Crystal 54

### DIFF
--- a/unittests/backends/iqm/IQMTester.cpp
+++ b/unittests/backends/iqm/IQMTester.cpp
@@ -113,7 +113,7 @@ CUDAQ_TEST(IQMTester, architectureMismatched) {
         platform.setTargetBackend(backendString);
       },
       testing::ThrowsMessage<std::runtime_error>(testing::StrEq(
-          "IQM QPU architecture mismatch: Expecting Crystal_5 architecture.")));
+          "IQM QPU architecture mismatch: Crystal_5 != Crystal_20")));
 }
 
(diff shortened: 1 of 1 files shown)


targettests/braket/cudaq_observe_qvector.cpp
bettinaheim, line 1: I am a bit confused about why this file is checked in under braket targettests - it seems that everything it is testing is unrelated to braket. Are these new tests, or was this file accidentally checked in?

targettests/execution/IQMTester.cpp
bettinaheim, line 1: It looks like this file is a copy of the unitttests/backends/iqm/IQMTester.cpp file. What was the reason to duplicate here? We don't have any similar tests for other backends under targettests/execution.

targettests/braket/cudaq_observe_qvector.cpp
sacpis: Seems like these files got in automatically. Let me remove those.

targettests/execution/IQMTester.cpp
sacpis: Seems like these files got in automatically. Let me remove those.

targettests/braket/cudaq_observe_qvector.cpp
sacpis: There should not be addition of any new file in this PR. Fixing it.

1tnguyen (approved):

PR #3291, opened 2025-07-31
Fix a potential race condition in `quantum_platform` thread Id to QPU Id map

1tnguyen:
### Description

The `threadToQpuId` map may be subjected to multi-threaded access:

- `runObservationAsync` will post functors to multiple threads that execute `runObservation`. 

- `runObservation` will then call `platform.set_current_qpu(qpu_id)`, which modified the `threadToQpuId` map => race condition (e.g., segfault if the map needs to be rehashed) 

Hence, adding a mutex to prevent race condition: locks are added to all places in the `quantum_plaform.cpp`. I think the other APIs (e.g., `launchVQE`, etc.) may not be subjected to this race condition but it's worth guarding them with the `shared_lock`.

Note: This `threadToQpuId` map is also accessed during the platform initialization (`setTargetBackend`). It's safe to assume no multi-threading during that stage.


1tnguyen:
> /ok to test 747768dcbe87e5d027b9aa5e8b7b23c4ec010f6a

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -93,11 +93,14 @@ void quantum_platform::set_current_qpu(const std::size_t device_id) {
   }
   platformCurrentQPU = device_id;
```
sacpis: Should we lock this as well?

```
platformCurrentQPU = device_id;
```

1tnguyen: Yea, it's worth guarding that as well. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3291/commits/5d64b47c14ea76aea893d141775abb3ed6e1b1c5

1tnguyen:
> /ok to test 5d64b47c14ea76aea893d141775abb3ed6e1b1c5

1tnguyen:
> /ok to test e0f6883fc829c012d46e9505e8d60b6a023d55a9

PR #3292, opened 2025-07-31
Ability to inject extra payload to REST requests

nvidia-dobri:
- Introduce `ExtraPayloadProvider` class to provide a mechanism for amending the REST JSON payload.

- The target can specify a provider that should be used during job request construction.

- Move the `RuntimeTarget` class from Python to `cudaq/common` to be used as the class that encapsulates target information.

- Add a dummy test (mock server) to verify the job JSON construction.

### DIFF
--- a/runtime/common/ExtraPayloadProvider.cpp
+++ b/runtime/common/ExtraPayloadProvider.cpp
@@ -8,6 +8,7 @@
 
 #include "ExtraPayloadProvider.h"
 #include <algorithm>
+#include <string>
 
 namespace cudaq {
 


python/utils/CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 SET(CMAKE_INSTALL_RPATH "$ORIGIN:$ORIGIN/..")
 
 add_library(cudaq-py-utils SHARED LinkedLibraryHolder.cpp)
-target_link_libraries(cudaq-py-utils PRIVATE nvqir cudaq fmt::fmt-header-only CUDAQTargetConfigUtil)
-target_include_directories(cudaq-py-utils PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/python/utils>)
+target_link_libraries(cudaq-py-utils PRIVATE nvqir cudaq fmt::fmt-header-only CUDAQTargetConfigUtil cudaq-common)
```
schweitzpgi: nit: is this longer than 80 columns?

runtime/common/ExtraPayloadProvider.cpp
```diff
@@ -0,0 +1,28 @@
...
+  extraPayloadProviders.push_back(std::move(provider));
+}
+
+const std::vector<std::unique_ptr<ExtraPayloadProvider>> &
+getExtraPayloadProviders() {
+  return extraPayloadProviders;
+}
+
+} // namespace cudaq
+
```
schweitzpgi: nit: github whining about file not ending with a CR.

runtime/common/RuntimeTarget.cpp
```diff
@@ -0,0 +1,50 @@
+/****************************************************************-*- C++ -*-****
```
schweitzpgi:
```suggestion
/*******************************************************************************
```

python/utils/CMakeLists.txt
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3292/commits/926a15b80a0150cc45270c4fa7f5e75d0ab9d821

runtime/common/ExtraPayloadProvider.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3292/commits/926a15b80a0150cc45270c4fa7f5e75d0ab9d821

1tnguyen:
> /ok to test 7becf1ff3f806da984ead5a51146f9886923acb8

unittests/backends/extra_payload_provider/horizon.yml
```diff
@@ -0,0 +1,29 @@
...
+  # Add the rest-qpu library to the link list
+  link-libs: ["-lcudaq-rest-qpu"]
+  # Define the lowering pipeline
+  platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,func.func(canonicalize),apply-op-specialization{constant-prop=1},aggressive-early-inlining,expand-measurements,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(canonicalize,multicontrol-decomposition)"
+  # Tell the rest-qpu that we are generating QIR.
+  codegen-emission: qir-base
+  # Library mode is only for simulators, physical backends must turn this off
+  library-mode: false
+  # Specify the extra payload provider to use
+  extra-payload-provider: sunrise
```
amccaskey: I was really hoping we would not need to manually modify the target definition to use extra payload providers. This could really lead to an explosion of unnecessary target definitions (they may be more than one valid extra payload provider per backend). If we want to keep this level of specificity, it may be better to think about the target-options, there at least a downstream library could add the required flags to nvq++ for the payload's unique feature.

1tnguyen: Good point. It makes a lot of sense to use target-options for this and I have updated the implementation.
Please let me know what you think.

1tnguyen:
> /ok to test 234b64c869769e20ce79ae96bbe707f5e47972fd

runtime/common/ExtraPayloadProvider.h
```diff
@@ -0,0 +1,41 @@
...
+  ExtraPayloadProvider() = default;
+  virtual ~ExtraPayloadProvider() = default;
+
+  /// @brief Get the name of the extra payload provider.
+  virtual const std::string name() const = 0;
+
+  /// @brief Inject extra payload into the server message.
+  /// @param target The RuntimeTarget to generate the payload for.
+  /// @param msg The server message to inject extra payload into.
+  virtual void injectExtraPayload(const RuntimeTarget &target,
```
bettinaheim: As discussed, it would be good to change this to return the extra payload instead of modifying the server message such that all logic that is specific to a particular service can be encapsulated into the server helper.

1tnguyen:
> /ok to test 7c0ecdff43932af8da004f89d86d857cafa0bc26

1tnguyen:
> /ok to test ca9edb0906b735c3ed6fe58284a9fc0c14094d96

1tnguyen:
> /ok to test 2f4b7ec4e1d3f4af4522a3c2fb4167941a751c19

1tnguyen:
> /ok to test 3da02964d966b187acd961823404da7e7da1bf0a

1tnguyen:
> /ok to test 2d8bbf2c1f531a0014d971441c2d78738234ad46

1tnguyen:
> /ok to test e6ca52e1774307eb05d5b98e9bcac8257d01bbf9

PR #3293, opened 2025-08-01
Simplify dependency structure of `RuntimeTarget`

1tnguyen:
### Description

- Separate YAML support from `TargetConfig` data classes.

- Move those platform's queries (based on the current platform at runtime) into `platform.h` rather than bundling them in `RuntimeTarget` class.

### DIFF
--- a/runtime/common/RuntimeTarget.cpp
+++ b/runtime/common/RuntimeTarget.cpp
@@ -26,4 +26,3 @@ std::string RuntimeTarget::get_target_args_help_string() const {
 }
 
 } // namespace cudaq
-                     


amccaskey:
Splitting out the YAML is a nice change. While we're here it would be helpful for libraries if we also installed `cudaq/Support/TargetConfig.h`. Can we add that here? (TBH it would be great if we installed all the compiler / dialect headers, but that's a bigger thing than this PR)

1tnguyen:
> Splitting out the YAML is a nice change. While we're here it would be helpful for libraries if we also installed `cudaq/Support/TargetConfig.h`. Can we add that here? (TBH it would be great if we installed all the compiler / dialect headers, but that's a bigger thing than this PR)

Definitely: install `cudaq/Support/TargetConfig.h` in https://github.com/NVIDIA/cuda-quantum/pull/3293/commits/4bbad75ebd4e171967edaff504958ec7e7637bba

PR #3296, opened 2025-08-02
Adding a decomposition pattern for controlled-Y gate (CY)

khalatepradnya:
* This pattern is needed by `quantinuum-gate-set-mapping`
* Python test to check decomposition


anthony-santana (approved):
Looks good to me 😄

PR #2838, opened 2025-04-24
Support for Quantum Machines backends

HanaGiatQM:

PR #3301, opened 2025-08-04
[docs] Adding Quantum Machines doucmentation

khalatepradnya:
Follow-up to PR https://github.com/NVIDIA/cuda-quantum/pull/2838


docs/sphinx/using/backends/hardware/aggregator.rst
```diff
@@ -0,0 +1,74 @@
...
+                        api_key="your_api_key",
+                        executor="mock")
+
+    Parameters:
+
+    - ``url``: The URL of the Quantum Machines server
+    - ``executor``: The name of the executor/backend to use (defaults to "mock")
+    - ``api_key``: Your API key (optional if set via environment variable)
+
+    To see a complete example for using Quantum Machines backends, take a look at our :doc:`Python examples <../../examples/examples>`.
```
efratshabtai: Is this part of this PR?

docs/sphinx/using/backends/hardware/aggregator.rst
```diff
@@ -0,0 +1,74 @@
+Aggregators
```
HanaGiatQM: Why is the term "Aggregator"? I think that "Quantum control system" is a more accurate title. WDYT?

as for the description below, how about:

Quantum Machines provides a unified quantum computing control system that enables
users to execute quantum programs on various QPU types (hardware backends) through their
Quantum Orchestration Platform (QOP) - a platform that controls different quantum hardware implementations through a unified software API

khalatepradnya: Fixed in [dad2f71](https://github.com/NVIDIA/cuda-quantum/pull/3301/commits/dad2f71bcde3431802f6e437ae49400caf3c060d).

The term `aggregator` is used in docs for backends that allow access to multiple QPUs.

docs/sphinx/using/backends/hardware/qcontrol.rst
```diff
@@ -0,0 +1,74 @@
...
+                        api_key="your_api_key",
+                        executor="mock")
+
+    Parameters:
+
+    - ``url``: The URL of the Quantum Machines server
+    - ``executor``: The name of the executor/backend to use (defaults to "mock")
+    - ``api_key``: Your API key (optional if set via environment variable)
+
+    To see a complete example for using Quantum Machines backends, take a look at our :doc:`Python examples <../../examples/examples>`.
```
efratshabtai: Is this supposed to point to a specific example using QM backend?
It currently points to general python example index. In this case maybe we should say something like "for complete CUDA-Q examples, take a look at..." or "look at ... for CUDA-Q examples you can run with the QM backend"

efratshabtai (approved):

docs/sphinx/using/backends/hardware/aggregator.rst
khalatepradnya: Yes, those are updates to `docs/sphinx/using/examples/hardware_providers.rst`

docs/sphinx/using/backends/hardware/qcontrol.rst
khalatepradnya: This takes to the generic examples page (same as other backends). I will update links so each can point to specific target.

khalatepradnya: Updated in 010a29b. 
Note that I will do a follow up PR to match this for other backends' docs because it needs additional movement of files that are unrelated to QM.

bmhowe23 (approved):
I only looked at the items in the .github folder. Hopefully others can review the other items?

bmhowe23:
> I only looked at the items in the .github folder. Hopefully others can review the other items?

Oh, I see that Efrat already has. Great!

docs/sphinx/using/backends/hardware/qcontrol.rst
khalatepradnya: See https://github.com/NVIDIA/cuda-quantum/pull/3339

PR #3304, opened 2025-08-05
NVQC CUDA-Q docker image with tpls source code

sacpis:
NVQC CUDA-Q docker image with tpls source code

Created both the original NVQC CUDA-Q docker image and NVQC CUDA-Q + tpls docker image.
https://registry.ngc.nvidia.com/orgs/pnyjrcojiblh/teams/cuda-quantum/containers/cuda-quantum/tags

### DIFF
--- a/docs/sphinx/using/extending/backend.rst
+++ b/docs/sphinx/using/extending/backend.rst
@@ -225,7 +225,7 @@ Create a ``YAML`` configuration file for your target:
       preprocessor-defines: ["-D CUDAQ_QUANTUM_DEVICE"]
       # Define the lowering pipeline
       # This will cover applying hardware-specific constraints since each provider may have different native gate sets, requiring custom mappings and decompositions. You may need assistance from the CUDA-Q team to set this up correctly.
-      platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(state-prep),unitary-synthesis,canonicalize,apply-op-specialization,aggressive-early-inlining,classical-optimization-pipeline,lower-to-cfg,func.func(canonicalize,multicontrol-decomposition),decomposition{enable-patterns=U3ToRotations},symbol-dce,<provider_name>-gate-set-mapping"
+      platform-lowering-config: "classical-optimization-pipeline,globalize-array-values,func.func(state-prep),unitary-synthesis,canonicalize,apply-op-specialization,aggressive-inlining,classical-optimization-pipeline,lower-to-cfg,func.func(canonicalize,multicontrol-decomposition),decomposition{enable-patterns=U3ToRotations},symbol-dce,<provider_name>-gate-set-mapping"
       # Tell the rest-qpu that we are generating QIR base profile.
       # As of the time of this writing, qasm2, qir-base and qir-adaptive are supported.
       codegen-emission: qir-base
--- a/include/cudaq/Optimizer/InitAllPasses.h
+++ b/include/cudaq/Optimizer/InitAllPasses.h
@@ -20,7 +20,7 @@ inline void registerCudaqPassesAndPipelines() {
   opt::registerOptTransformsPasses();
 
   // CUDA-Q pipelines
-  opt::registerAggressiveEarlyInliningPipeline();
+  opt::registerAggressiveInliningPipeline();
   opt::registerUnrollingPipeline();
   opt::registerClassicalOptimizationPipeline();
   opt::registerToExecutionManagerCCPipeline();
--- a/include/cudaq/Optimizer/Transforms/Passes.h
+++ b/include/cudaq/Optimizer/Transforms/Passes.h
@@ -21,9 +21,8 @@ namespace cudaq::opt {
 /// Add a pass pipeline to transform call between kernels to direct calls that
 /// do not go through the runtime layers, inline all calls, and detect if calls
 /// to kernels remain in the fully inlined into entry point kernel.
-void addAggressiveEarlyInlining(mlir::OpPassManager &pm,
-                                bool fatalCheck = false);
-void registerAggressiveEarlyInliningPipeline();
+void addAggressiveInlining(mlir::OpPassManager &pm, bool fatalCheck = false);
+void registerAggressiveInliningPipeline();
 
 void registerUnrollingPipeline();
 void registerClassicalOptimizationPipeline();
--- a/lib/Optimizer/CodeGen/Pipelines.cpp
+++ b/lib/Optimizer/CodeGen/Pipelines.cpp
@@ -17,7 +17,7 @@ void cudaq::opt::commonPipelineConvertToQIR(PassManager &pm,
   auto passConfigFields = passConfigAs.split(':');
 
   pm.addNestedPass<func::FuncOp>(createApplyControlNegations());
-  addAggressiveEarlyInlining(pm);
+  addAggressiveInlining(pm);
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(createUnwindLowering());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
@@ -40,7 +40,7 @@ void cudaq::opt::commonPipelineConvertToQIR(PassManager &pm,
   pm.addPass(createApplySpecialization());
   // If there was any specialization, we want another round in inlining to
   // inline the apply calls properly.
-  addAggressiveEarlyInlining(pm);
+  addAggressiveInlining(pm);
   addLowerToCFG(pm);
   pm.addNestedPass<func::FuncOp>(createCombineQuantumAllocations());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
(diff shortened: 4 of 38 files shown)


scripts/install_prerequisites.sh
```diff
@@ -35,16 +35,26 @@
 toolchain=''
 exclude_prereq=''
 install_all=true
+keep_sources=false
+tpls_dir="/opt/cudaq/tpls"
```
bettinaheim: rather than hardcode this, it would probably be good to look if CUDAQ_INSTALL_PREFIX is defined.

scripts/install_prerequisites.sh
```diff
@@ -340,13 +385,67 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
...
   else
     echo "AWS SDK already installed in $AWS_INSTALL_PREFIX."
   fi
 fi
 
+lookup_tpls_sha() {
+  local path="$1"
+
+  # Using lock file
+  if [[ -f /tmp/tpls_commits.lock ]]; then
```
bettinaheim: This seems rather specific to the nvqc dockerfile. What was the motivation for not just directly using git here (instead of in the dockerfile) to get the shas?

docker/release/cudaq.nvqc.Dockerfile
```diff
@@ -10,12 +10,45 @@
...
 
+# Flag to control whether to install/clone prerequisites
+ARG WITH_TPLS=false
+
+# COPY install_prerequisites into the image
+RUN sudo mkdir -p /tmp
+COPY --chmod=0755 scripts/install_prerequisites.sh /tmp/install_prerequisites.sh
+COPY .gitmodules /tmp/.gitmodules
+
+# Create a tpls_commits.lock file and move it to /tmp
```
bettinaheim: Why is this commented out?

scripts/install_prerequisites.sh
```diff
@@ -340,13 +385,67 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
...
+    echo "Processing submodule $lib at path $path ..."
+    repo="$(git config --file=.gitmodules submodule.$path.url)"
+    echo "Repository URL: $repo"
+
+    echo "Adding $dest as a safe.directory..."
+    sudo git config --global --add safe.directory "$dest"
+
+    commit="$(lookup_tpls_sha "$path")" || {
+      echo "ERROR: could not resolve pinned commit for $path. Aborting $lib." >&2
+      continue
```
bettinaheim: maybe fail the script instead?

scripts/install_prerequisites.sh
sacpis: Fixed in [7f7cdd6](https://github.com/NVIDIA/cuda-quantum/pull/3304/commits/7f7cdd6483c023aa5698c19083f8a1fbe0ba307a).

scripts/install_prerequisites.sh
sacpis: Fixed in [7f7cdd6](https://github.com/NVIDIA/cuda-quantum/pull/3304/commits/7f7cdd6483c023aa5698c19083f8a1fbe0ba307a).

docker/release/cudaq.nvqc.Dockerfile
sacpis: Cleaned up in [7f7cdd6](https://github.com/NVIDIA/cuda-quantum/pull/3304/commits/7f7cdd6483c023aa5698c19083f8a1fbe0ba307a).

scripts/install_prerequisites.sh
sacpis: Fixed in [7f7cdd6](https://github.com/NVIDIA/cuda-quantum/pull/3304/commits/7f7cdd6483c023aa5698c19083f8a1fbe0ba307a).

sacpis:
Seems like we will need to pass the commit sha from the Dockerfile to install_prereq script as the pinned commit cannot be retrieved inside the docker layer.

sacpis:
The git commands used in install_prereq are accessing url, paths from .gitmodules.

scripts/install_prerequisites.sh
```diff
@@ -35,16 +35,34 @@
 toolchain=''
 exclude_prereq=''
 install_all=true
+tpls_root="${CUDAQ_INSTALL_PREFIX:-/opt/cuda}"
+tpls_dir="$tpls_root/tpls"
+sudo mkdir -p "$tpls_dir"
```
bettinaheim: This script should not directly use sudo. Using sudo within a script is not a good practice - generally, if a script needs elevated permissions, then the script itself should be run as such. Please remove all sudo here (you can e.g. run this in an elevated shell in the nvqc dockerfile instead - `sudo bash <script>`)

scripts/install_prerequisites.sh
```diff
@@ -55,6 +73,63 @@ while getopts ":e:t:m" opt; do
 done
 OPTIND=$__optind__
 
+# Check if the lock file is present and then copy it to /tmp
```
bettinaheim: comment is no longer accurate

scripts/install_prerequisites.sh
```diff
@@ -55,6 +73,63 @@ while getopts ":e:t:m" opt; do
...
+    sub=${sub%.url}               # -> "tpls/foo"
+    path=$(git config --file .gitmodules --get "submodule.$sub.path")
+    lib=$(basename "$path")       # -> "foo"
+    dest="$tpls_dir/$lib"
+
+    echo "Processing submodule $lib at path $path ..."
+    repo="$(git config --file=.gitmodules submodule.$path.url)"
+    echo "Repository URL: $repo"
+
+    echo "Adding $dest as a safe.directory..."
```
bettinaheim: Why do you need to add it as safe directory? I find it rather invasive for a script to silently do that.

scripts/install_prerequisites.sh
```diff
@@ -55,6 +73,63 @@ while getopts ":e:t:m" opt; do
 done
 OPTIND=$__optind__
 
+# Check if the lock file is present and then copy it to /tmp
+if [ -n "$lock_file" ]; then
+  if [ ! -f "$lock_file" ]; then
```
bettinaheim: This block can be moved inside the `if [ -n "$lock_file" ]; then` block below.

scripts/install_prerequisites.sh
```diff
@@ -35,16 +35,34 @@
 toolchain=''
 exclude_prereq=''
 install_all=true
+tpls_root="${CUDAQ_INSTALL_PREFIX:-/opt/cuda}"
```
bettinaheim: Please move these lines inside the `if [ -n "$lock_file" ]; then` block below such that directories are created and variables are defined only if this option is actually requested. By "these lines" I mean all changes up here expect for the actual command line argument and the lock_file variable.

docker/release/cudaq.nvqc.Dockerfile
```diff
@@ -10,12 +10,44 @@
...
+# Use the following command to generate the tpls_commits.lock file which
+# contains the pinned commits for the submodules in /tpls directory. This is
+# needed to generate NVQC image with tpls source code.
+# git config --file .gitmodules --get-regexp '^submodule\..*\.path$' \
+#         | awk '{print $2}' \
+#         | while read p; do printf "%s %s\n" "$(git rev-parse HEAD:$p)" "$p"; done \
+#         > tpls_commits.lock
+
+# Uncomment the following line to copy the tpls_commits.lock file into the image
+# COPY tpls_commits.lock /tmp/tpls_commits.lock
```
bettinaheim: In case we want this in a pipeline in the future, you can avoid having to do any manual edits to the file and instead have the copy only depending on an argument by doing something like this:

```
ARG WITH_TPLS=false

FROM nvcf_image AS with_tpls_false
RUN echo "Default build without prerequisites."
FROM nvcf_image AS with_tpls_true
COPY tpls_commits.lock /tmp/tpls_commits.lock
RUN echo "adding install_prerequisites.sh"; \
         # ...etc.

FROM with_tpls_${WITH_TPLS}
```

scripts/install_prerequisites.sh
sacpis: Fixed in 526ac79 and 68eaecd.

scripts/install_prerequisites.sh
sacpis: Fixed in 526ac79 and 68eaecd.

scripts/install_prerequisites.sh
sacpis: Fixed in 526ac79 and 68eaecd.

scripts/install_prerequisites.sh
sacpis: Fixed in 526ac79 and 68eaecd.

docker/release/cudaq.nvqc.Dockerfile
sacpis: Fixed in 526ac79 and 68eaecd.

scripts/install_prerequisites.sh
sacpis: Fixed in 526ac79 and 68eaecd.

PR #3306, opened 2025-08-05
Phase Folding Optimization

atgeller:
Implements the phase polynomial based rotation merging optimization from https://arxiv.org/abs/1710.07345.

Multiple `quake.rz` ops can be merged into a single gate if they occur at the same phase. In a circuit with only Swap, Not, CNot, and Rz gates, a phase represents the state of a qubit based on the Swap, Not, and CNot operations applied to it.

First, a preprocessing step identifies each subcircuit consisting only of Swap, Not, CNot, and Rz gates. This preprocessing step starts by inspecting each CNot gate and working outwards, establishing a termination boundary determining the extent of the subcircuit based on unsupported operations.

At the beginning of the subcircuit, each qubit is assigned a unique phase variable. A phase is an exclusive sum of the phase variables determining the current state of a qubit, calculated as follows. While walking the subcircuit:
* A Swap gate swaps the phases for two qubits.
* A Not gate inverts the phase for a qubit by (un)setting a flag.
* A CNot gate sets the phase of the target to the exclusive sum of the phase variables in the phases for the control and target qubits.
* If multiple `rz` ops are found at the same phase, they will be combined by replacing the rotation angle of the latter occurring op to the sum of the rotation angles of both ops, and removing the prior op.

Phase folding is currently supported for the `remote-mqpu` target using the `CUDAQ_PHASE_FOLDING` environment variable.

Some examples can be found in `test/Quake/phase_folding.qke`.

This PR includes a random test generator python script in `unittests/phase_folding/random_gen.py`, which is used to generate random test cases based on templates.

# Benchmarks

Scripts are available at https://gist.github.com/atgeller/b56c0cd3d6db4665ac69e40aa1c54fe0. Copy them into your repo root and follow the instructions below.

## Benchmark results
Preliminary benchmark results can be found here: [Phase Folding Optimization Results.xlsx](https://github.com/user-attachments/files/22052216/Phase.Folding.Optimization.Results.xlsx)

## Randomly Generating Circuits

The script `random_gen.py` is a tool for randomly generating circuits based
on a template, which parameters for the block length, rz weight (percentage of
rotations compared to other operations in subcircuits), and number of qubits.

A template should look mostly like a standard cudaq program, except that in lieu
of quantum operations, the following directives can be used to randomly generate
a circuit:

* `GEN-QALLOC: nqubits=<n>` will generate a `qvector` allocation with size `n`.
In addition, each of the qubits in the `qvector` will automatically be added to
a pool of qubits to be randomly selected from when generating operations.
* `GEN-BLOCK` will generate a sequence of random operations with the length
determined by the `--block-length` CLI option.
* `GEN-SUBCIRCUIT` will generate a sequence of random valid subcircuit operations,
(i.e., NOT, CNOT, Swap, and rz) with the length determined by the `--block-length`
CLI option.
* `GEN-MEASURES` will insert measurements for all qubits in the active pool.
* `GEN:<nqubits>` will be replaced with the number of quibts in the active pool.
* `GEN:<qubit>` will be replaced with a random qubit from the active pool.
* `GEN:<qvector>` will be replaced with a random `qvector` from the active pool.

The active pool of `qvector`s/qubits is tracked via curly braces, so it is important
to always use curly braces for blocks, even when they just have a single instruction.

## Benchmarking With Randomly Generating Subcircuits

The script `run_phase_folding_benchmarking.py` will randomly generate and benchmark
a suite of kernels based on `simple.template` and sets of configuration options
for each of the parameters for randomly generating a script: block length, rz
weight, and number of qubits. This script will construct multiple random
benchmarks with different random seeds for each configuration, and run each such
benchmark several times, including a warmup run to mitigate noise from caching.

For example, the following will randomly generate templates with block lengths of
100 and 200 (`--block-lengths 100 200`), rotation operation weights of 0.5 and 0.6
(`--rz-weights 0.5 0.6`), and 5 and 10 qubits each (`--n-qubits 5 10`), and 2 seeds
per configuration (`--n-seeds=2`) for a total of 16 different benchmarks. Each
benchmark will be run once (`--iterations=1`). Results per run will be dumped to
`raw.csv` (`--raw-data-file=raw.csv`, by default results per run won't be kept).
Summary results per configuration are dumped to `results.csv` by default, but this
can be overridden by providing a results file name.

```
python3 run_opt_benchmarks.py --block-lengths 100 200 --rz-weights 0.5 0.6 \
    --n-qubits 5 10 --seed 489 --n-seeds=2 --iterations=1 --raw-data-file=raw.csv
```

In addition, the `--targets TARGET [TARGETS]` option can be used to specify
a particular target or targets to run the benchmarks on. The default target
is currently `remote-mqpu`.

## Running Benchmarks in a Docker Contaienr
Here's an example of running the above scripts in the cuda-quantum development container and extracting the results. It must be run from the repo root.
```
docker build -t cudaq-dev -f docker/build/cudaq.dev.Dockerfile . --build-arg install="CMAKE_BUILD_TYPE=Release CUDA_QUANTUM_VERSION=latest"

nvidia-docker run -dit --name runner --rm cudaq-dev 

docker exec runner python3 -u run_phase_folding_benchmarks.py --block-lengths 5 10 25 50 100 250 500 1000 --rz-weights 0.15 0.3 0.5 0.7 0.85 --n-qubits 3 5 10 15 20 --seed 489 --n-seeds=5 --iterations=3 --raw-data-file=raw.csv

docker cp runner:/workspaces/cuda-quantum/results.csv .
docker cp runner:/workspaces/cuda-quantum/raw.csv .

docker stop runner
```


lib/Optimizer/Transforms/PhaseFolding.h
```diff
@@ -0,0 +1,510 @@
+#pragma once
```
schweitzpgi: nit: missing banner

schweitzpgi: nit: why is some of this implemented in a header file? That doesn't follow the coding convention.

runtime/common/BaseRestRemoteClient.h
```diff
@@ -212,11 +212,16 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
     std::string errMsg;
     llvm::raw_string_ostream os(errMsg);
-    const std::string pipeline =
+
+    std::string pipeline =
         std::accumulate(clientPasses.begin(), clientPasses.end(), std::string(),
                         [](const auto &ss, const auto &s) {
                           return ss.empty() ? s : ss + "," + s;
                         });
+    if (getEnvBool("CUDAQ_PHASE_FOLDING", false))
```
schweitzpgi: nit: This is clever, but we should probably avoid more bonus envvars. heh

test/Transforms/PhaseFolding/qpe2qubits.qke
```diff
@@ -0,0 +1,70 @@
+// ========================================================================== //
+// Copyright (c) 2025 NVIDIA Corporation & Affiliates.                        //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// ========================================================================== //
```
schweitzpgi: nit: double banners in this one. heh

test/Transforms/PhaseFolding/simple.qke
```diff
@@ -0,0 +1,136 @@
+// ========================================================================== //
```
schweitzpgi, line 1: nit: prefer to not overdo the subdirectories under test. Can we just name these `test/Transforms/phase_folding_simple.qke` etc.?

runtime/common/BaseRestRemoteClient.h
atgeller: Is there an opt-level flag that can be set when using `nvq++`? And if so can it be accessed during JIT compilation?

lib/Optimizer/Transforms/PhaseFolding.h
atgeller: That is an artifact of having this implemented as two separate passes in separate files. I'll move the definitions into the cpp file.

test/Transforms/PhaseFolding/qpe2qubits.qke
atgeller: Whoops, fixed :)

test/Transforms/PhaseFolding/simple.qke
atgeller: Done

runtime/common/BaseRestRemoteClient.h
atgeller: Ah, actually, part of the reason we wanted the envvar is because it can be set dynamically, so we can run the circuit with and without the optimization, for testing and benchmarking. I'm not aware of another way to provide a dynamic configuration option to the JIT.

atgeller:
> /ok to test 90c244e

atgeller:
> /ok to test 14144c5

atgeller:
> /ok to test 1d18c09

runtime/common/BaseRestRemoteClient.h
bettinaheim: I don't think we currently have a way - would be good to add one. For now the environment variable sounds reasonable, but we should come back to it as part of introducing runtime configurable options more broadly.

atgeller:
> /ok to test e755d9e

atgeller:
> /ok to test de202a2

runtime/common/BaseRestRemoteClient.h
atgeller: I added a TODO

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+
+  // If any values are unsupported, the operation is also unsupported
+  for (auto operand : quake::getQuantumOperands(op))
+    if (!isSupportedValue(operand))
+      return true;
+
+  return false;
+}
+
+inline bool isTwoQubitOp(Operation *op) {
```
1tnguyen, line 89: nit: since we have `inline` here, perhaps we can add `inline` to some of the above one-liner functions as well, e.g. `processed`, `markProcessed`, etc., for consistency.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+  return quake::getQuantumOperands(op).size() == 2;
+}
+
+class Netlist {
+  SmallVector<SmallVector<Operation *>> netlists;
```
1tnguyen, line 100: Some code comments here would be nice.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+  public:
+    NetlistWrapper(Subcircuit *subcircuit, SmallVector<Operation *> *nl,
+                   Operation *anchor_point, Value def) {
+      this->nl = nl;
+      this->subcircuit = subcircuit;
+      this->def = def;
+      processFrom(getIndexOf(anchor_point));
+    }
```
1tnguyen, line 259:
```suggestion
    NetlistWrapper(Subcircuit *subcircuit, SmallVector<Operation *> *nl,
                   Operation *anchor_point, Value def): nl(nl), subcircuit(subcircuit), def(def) {
      processFrom(getIndexOf(anchor_point));
    }
```

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+
+public:
+  /// @brief Constructs a subcircuit containing only RZ, NOT, CNOT, and Swap
+  /// gates, using the Netlist representation `netlist`.
+  /// @details A subcircuit is an connected portion of the netlist containing
+  /// only RZ, NOT, CNOT, and Swap gates.
+  ///
+  /// First, we construct an initial subcircuit:
+  /// We start by walking forward and backward along the netlist from the
+  /// initial anchor point, which is the control of `cnot`, and add the
```
1tnguyen: Just checking if the comment here is correct?

```suggestion
  /// initial anchor point, which is a `cnot`, and add the
```

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+  void doPhaseFolding(Subcircuit *subcircuit) {
+    SmallVector<PhaseVariable *> phase_vars;
+    SmallVector<Phase> current_phases;
+    PhaseStorage store;
+    size_t i = 0;
+    // Initialize the phases and phase variables for each qubit in the circuit,
+    // the initial phase contains only the phase variable for that qubit
+    for (auto ref : subcircuit->getRefs()) {
+      auto phase_idx = i++;
+      auto new_phase_var = new PhaseVariable(phase_idx, ref);
```
1tnguyen, line 558: Q: Where do we clean up this `PhaseVariable`?

I saw that we pushed it to `phase_vars`, perhaps for tracking and cleaning purposes?

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+        new_phase.vars.insert(var);
+    new_phase.isInverted = (p1.isInverted != p2.isInverted);
+    return new_phase;
+  }
+
+  /// @brief Returns a new phase equal to `p1` with the opposite inversion flag
+  static Phase invert(Phase &p1) {
+    auto new_phase = Phase();
+    for (auto var : p1.vars)
+      new_phase.vars.insert(var);
```
1tnguyen: Q: Could we write this as:

```suggestion
    new_phase.vars.insert(p1.vars.begin(), p1.vars.end());
```
?

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,656 @@
...
+  /// operation, which will later be walked. If a disallowed gate is
+  /// encountered, we stop walking and add a termination point.
+  ///
+  /// Then, we prune the subcircuit, starting at the earlist ending termination
+  /// point (i.e., earliest termination point encountered while walking forward)
+  /// along each qubit, and walk forward, adjusting the termination boundary for
+  /// any connected qubits, and removing operations after the termination
+  /// boundary from the subcircuit.
+  Subcircuit(Operation *cnot, Netlist *netlist) {
+    start = cnot;
```
1tnguyen: I'm wondering whether we should add `assert(isCNOT(op))` here.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: An anchor point is a particular point on the netlist, along some qubit at an operation. I think it is more correct to say "at `cnot`", but also to clarify that it is for a particular qubit. I'll work on clarifying the explanation here.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Hmm, it doesn't look like it is cleaned up currently. I added a loop to `delete` all the `PhaseVariable`s in `phase_vars` at the end.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Make sense, done.

atgeller:
> /ok to test 2e915e1

atgeller:
> /ok to test 7c98691

atgeller:
> /ok to test f4f009e

atgeller:
> /ok to test 6892df4

atgeller:
> /ok to test 786d39e

atgeller:
> /ok to test b9c96dd

runtime/common/BaseRestRemoteClient.h
```diff
@@ -213,20 +213,35 @@ class BaseRemoteRestRuntimeClient : public RemoteRuntimeClient {
...
+
+    std::string pipeline =
         std::accumulate(clientPasses.begin(), clientPasses.end(), std::string(),
                         [](const auto &ss, const auto &s) {
                           return ss.empty() ? s : ss + "," + s;
                         });
+    // TODO: replace environment variable with runtime configuration
+    if (getEnvBool("CUDAQ_PHASE_FOLDING", true))
+      pipeline = pipeline + "func.func(factor-quantum-alloc,canonicalize,cse,"
+                            "phase-folding,canonicalize)";
```
schweitzpgi: What's the reason for this being controlled by an environment variable?

targettests/lit.cfg.py
```diff
@@ -20,7 +20,7 @@
 # `testFormat`: The test format to use to interpret tests.
 config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
 
-config.suffixes = ['.cpp', '.config']
+config.suffixes = ['.cpp', '.config', '.template']
```
schweitzpgi: Are .template files what we really run? From the name, I would assume these are like `foo.cpp.in` files that get preprocessed into A set of `foo-n.cpp` files and then lit would run the `foo-N.cpp` files.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+inline bool processed(Operation *op) { return op->hasAttr("processed"); }
+
+inline void markProcessed(Operation *op) {
+  op->setAttr("processed", OpBuilder(op).getUnitAttr());
+}
+
+// AXIS-SPECIFIC: could allow controlled y and z here
+inline bool isCNOT(Operation *op) {
+  return isa<quake::XOp>(op) && op->getNumOperands() == 2;
```
schweitzpgi: You could check that the XOp has one control operand.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+}
+
+/// Currently, only `!quake.ref`s generated directly from
+/// `quake.alloca`s are supported. This is with the assumption that
+/// the `factor-quantum-alloc` pass was run before, so any veqs, etc...
```
schweitzpgi, line 42: This assumption can be "forced" (or better controlled anyway) by creating a named pipeline that always executes the required passes together.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+namespace {
+#define RAW(X) quake::X
+// AXIS-SPECIFIC: Defines which operations break a circuit into subcircuits
+#define CIRCUIT_BREAKERS(MACRO)                                                \
+  MACRO(YOp), MACRO(ZOp), MACRO(HOp), MACRO(R1Op), MACRO(RxOp),                \
+      MACRO(PhasedRxOp), MACRO(RyOp), MACRO(U2Op), MACRO(U3Op)
+#define RAW_CIRCUIT_BREAKERS CIRCUIT_BREAKERS(RAW)
+
+inline bool processed(Operation *op) { return op->hasAttr("processed"); }
```
schweitzpgi: Adding processed attributes seems like it will be expensive. It is more typical to keep a worklist and deplete it or keep a hashing set of processed ops.

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+  // If any values are unsupported, the operation is also unsupported
+  for (auto operand : quake::getQuantumOperands(op))
+    if (!isSupportedValue(operand))
+      return true;
+
+  return false;
+}
+
+inline bool isTwoQubitOp(Operation *op) {
```
schweitzpgi, line 89: nit: Style-wise these little functions should be `static` and not in an anonymous namespace. (LLVM style.)

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+  }
+
+  bool isTerminationPoint(Operation *op) {
+    // Currently, each operation can only be part of one subcircuit (hence the
+    // check for the processed flag)
+    return (op->getBlock() != start->getBlock()) || processed(op) ||
+           isCircuitBreaker(op);
+  }
+
+  class NetlistWrapper {
```
schweitzpgi, line 169: Does this need to be a nested class?

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+    }
+
+    bool hasOps() { return end_point > start_point; }
+
+    void prune() { pruneFrom(end_point); }
+
+    Value getDef() { return def; }
+  };
+
+  Netlist *container;
```
schweitzpgi: nit: default initializer?

lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -0,0 +1,670 @@
...
+  SmallVector<NetlistWrapper *> qubits;
+  SetVector<Operation *> ops;
+  SmallVector<Operation *> ordered_ops;
+  Operation *start;
```
schweitzpgi: nit: default init?

atgeller:
> /ok to test 19e5a1d

atgeller:
> /ok to test b60f84e

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: It's quite intertwined with `Subcircuit`, so it would be a fair bit of work to break them apart into two, and then they would still probably need to be friends. The `NetlistWrapper` really shouldn't be used outside `Subcircuit` anyway.

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: Done

lib/Optimizer/Transforms/PhaseFolding.cpp
atgeller: I opted for a hashing set.

runtime/common/BaseRestRemoteClient.h
atgeller: It's necessary for the tests and benchmarking to be able to run with and without phase folding.

targettests/lit.cfg.py
atgeller: I added a `CMakeFile` to generate the tests from the templates, so we don't run `.template` files anymore.

atgeller:
> /ok to test 958f186

runtime/common/BaseRestRemoteClient.h
atgeller: I've add #3405 to track this so they can be removed once a runtime configuration is set up.

atgeller:
> /ok to test 958f186

atgeller:
> /ok to test cb16062

targettests/Remote-Sim/phase-folding/aliasing.cpp
```diff
@@ -0,0 +1,155 @@
...
+  auto counts2 = cudaq::estimate_resources(kernel);
+  assert(counts2.count("rz") == 2);
+
+  assert(state1.get_num_qubits() == state2.get_num_qubits());
+  auto result = state1.overlap(state2);
+  ASSERT_NEAR(result.real(), 1, 0.000001);
+  ASSERT_NEAR(result.imag(), 0, 0.000001);
+}
+
+void check4() {
```
bettinaheim: This test compiles with the fix in https://github.com/NVIDIA/cuda-quantum/pull/3443. However, it currently fails and needs further investigation.

bettinaheim: Oh, the issue is just that the random seed is not set.

bettinaheim:
> /ok to test b1e6d63

bettinaheim:
> /ok to test 0773727

PR #3315, opened 2025-08-07
Bump to Fedora:41 as Fedora:38/39 are EOL

mitchdz:
We are seeing network errors with Fedora:39 https://github.com/NVIDIA/cuda-quantum/actions/runs/16786910932/job/47541785386?pr=3309#step:6:67

Fedora:39 has been EOL for almost a year. Update our runners to use a supported version.

Also, remove suggestion for supported on Fedora:38/39 and update to Fedora:41

Upstream fedora bug - https://discussion.fedoraproject.org/t/error-504-404-when-updating-fedora-39/161798/2

PR #3316, opened 2025-08-07
ghcr_config: cache fedora:41

mitchdz:
Testing for https://github.com/NVIDIA/cuda-quantum/pull/3315


bettinaheim (approved):
I assume you would want to also update the validation confit in the same folder to use the newer fedora for the Python 3.13 (and other) validation(s)?

mitchdz:
Yes, I just want to update the cache so my per-PR CI can use the cached image.

mitchdz:
Closing in favor of https://github.com/NVIDIA/cuda-quantum/pull/3319

PR #3318, opened 2025-08-07
Throw a warning for use of unsupported `draw` API on hardware targets.

khalatepradnya:
This PR improves the usability and robustness of the `draw` function in the CUDA-Q, specifically handling cases where the function is invoked on hardware backends instead of simulators. The changes ensure that users receive a warning and an empty result rather than an exception. Also update the docs to clarify that this is simulator-only feature.


khalatepradnya:
Converting back to draft to finish some internal discussion.

khalatepradnya:
@bettinaheim - Please confirm if this is ready to go.

PR #3320, opened 2025-08-08
Cache Fedora 42 && redhat10

mitchdz:
Fedora 42 is out as the latest version, and so is redhat10.

Fedora always has a 13 month cadence to EOL versions. Fedora 41 will EOL this year, 2025 on November 19th[0].

[0] - https://fedorapeople.org/groups/schedule/f-41/f-41-key-tasks.html


mitchdz:
> /ok to test 7958829

PR #3332, opened 2025-08-12
Support for QIR output format on Quantinuum

1tnguyen:
### Description

This PR enables the receiving and processing of the QIR output log from Quantinuum's QSYS result. This allows `cudaq::run` on Quantinuum devices that return QSYS results. 

(1) Updates Quantinuum server helper to distinguish result type (PYTKET (legacy) vs. QSYS). 

- Use the appropriate endpoint to retrieve QIR output log if the device returns QSYS results.

- For `cudaq::sample`, update the QIR log parser to handle `RESULT` data type, which is logged by `__quantum__rt__result_record_output`. The parser converts this to bit strings for the count map and sequential data.
Note: we don't populate the per-register count map here. In most cases, they are auto-generated classical register names, hence wouldn't be needed by users. 

- For `cudaq::run`: adding the ability for the server helper to retrieve the QIR output log from the job response if available. This will be propagated via the execution context to the runtime so that we can process and parse it into `vector<T>` for return.

(2) Mock server update to produce QSYS result

Update the mock testing infrastructure to support `run`. Use the simulator-generated QIR output to construct QSYS job result.

(3) Add simple integration tests for nightly validation.


1tnguyen:
> /ok to test 3065ec13196823252fdd2cc8746a53d219dacfee

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -939,7 +939,8 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
       // Allow developer to disable remote sending (useful for debugging IR)
       if (getEnvBool("DISABLE_REMOTE_SEND", false))
         return;
-      future = executor->execute(codes, isObserve);
+      future = executor->execute(codes, isObserve, isRun,
```
schweitzpgi: Below there is an `enum` defined. Why thread around a bag of `bool` arguments?

1tnguyen:
> /ok to test e73718937ce75439b0243e226e63153c9f4e6fc5

runtime/common/BaseRemoteRESTQPU.h
1tnguyen: Good point! Fixed in e73718937ce75439b0243e226e63153c9f4e6fc5

1tnguyen:
> /ok to test 9b6c952c876aef9f105a33f44b63dfcf32319852

1tnguyen:
> /ok to test eea7bb8cd52d90991a2627b602e5a5a26d0f3827

1tnguyen:
> /ok to test 7078dcac6ad24b5757efdb4f10b72c1f9d26a052

1tnguyen:
> /ok to test f7424a1603763fa15bf788855f06c45daab36861

schweitzpgi:
Regarding the TODO, please see #3365 as I think it addresses the issue already.

1tnguyen:
> Regarding the TODO, please see #3365 as I think it addresses the issue already.

Indeed, it addresses most of the issues already, i.e., the ability to propagate the QIR code gen configuration throughout the compiler.

I think we'd still need a bit of glue code in the YAML file/target configuration to support customizing this for different 'device architectures' under the same target. For example, for the `quantinuum` target, existing devices (machines) would need QIR 1.0 + integer computation, and next-gen devices would support under-development QIR + integer and float computations.

.github/workflows/integration_tests.yml
```diff
@@ -445,6 +445,61 @@ jobs:
...
+          test_err_sum=0
+          for filename in ${{ steps.gettests.outputs.testlist }}; do
+            # TODO: merge the sample and run tests as we unified the compilation pipeline
+            case $filename in
+              targettests/quantinuum_ng/sample*)
+                [ -e "$filename" ] || echo "::error::Couldn't find files ($filename)"
+                nvq++ -v $filename --target quantinuum --quantinuum-machine Helios-1E --quantinuum-project ${{ secrets.QUANTINUUM_NEXUS_PROJECT_ID }} --quantinuum-max-cost 10 --quantinuum-noisy-simulation false
+                test_status=$?
+                if [ $test_status -eq 0 ]; then
+                  CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT=1 ./a.out
```
khalatepradnya: Does this environment variable go away with PR https://github.com/NVIDIA/cuda-quantum/pull/3365

utils/mock_qpu/quantinuum/__init__.py
```diff
@@ -160,16 +172,20 @@ async def create_job(job: dict):
     m = llvm.module.parse_bitcode(decoded)
     mstr = str(m)
     assert ('entry_point' in mstr)
+    print("Code")
+    print(mstr)
```
khalatepradnya: Are these print statements required?

runtime/common/RecordLogParser.cpp
```diff
@@ -83,8 +88,45 @@ void cudaq::RecordLogParser::handleOutput(
...
+
+    currentOutput = OutputType::RESULT;
+    const bool isUninitializedContainer =
+        (containerMeta.m_type == ContainerType::NONE) ||
+        (containerMeta.m_type == ContainerType::ARRAY &&
+         containerMeta.elementCount == 0);
+    if (isUninitializedContainer) {
+      // Currently, our QIR for sampled kernel only has a sequence of RESULT
+      // records, not wrapped in an ARRAY. Hence, we treat it as an array of
+      // results.
```
khalatepradnya, line 108: Q: Is this generally true for `RESULT` type or may be applicable to specific version that can be deduced from metadata?

runtime/common/RecordLogParser.cpp
```diff
@@ -183,6 +227,10 @@ void cudaq::RecordLogParser::preallocateTuple() {
 void cudaq::RecordLogParser::processSingleRecord(const std::string &recValue,
                                                  const std::string &recLabel) {
   auto label = recLabel;
+  // For result type, we don't use the record label (register name) as the type
+  // annotation.
```
khalatepradnya, line 231: Same Q here - is this from spec or applicable for this QIR version?

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
```diff
@@ -20,7 +20,7 @@ config:
   # Define the lowering pipeline
   platform-lowering-config: "func.func(erase-noise),classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,func.func(canonicalize),apply-op-specialization{constant-prop=1},aggressive-early-inlining,expand-measurements,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(canonicalize,multicontrol-decomposition),quantinuum-gate-set-mapping,distributed-device-call"
   # Tell the rest-qpu that we are generating Adaptive QIR.
-  codegen-emission: qir-adaptive[int_computations]
+  codegen-emission: qir-adaptive[int_computations,float_computations]
```
khalatepradnya: Q: Is this device-specific? Different code for different machines?

.github/workflows/integration_tests.yml
1tnguyen: Yes, it will go away, and we need to find a different solution. After 3365, we need to either create a separate `.yml` file for Helios so that we can change the qir version or extend the target configuration to handle different configs based on device architectures for the same target/provider.

runtime/common/Future.cpp
```diff
@@ -40,9 +40,28 @@ sample_result future::get() {
...
 
-    if (isObserve) {
+    if (resultType == ExecutionContextType::run) {
+      QirServerHelper *qirServerHelper =
+          dynamic_cast<QirServerHelper *>(serverHelper.get());
+      if (!qirServerHelper)
+        throw std::runtime_error(
+            "cudaq::details::future::getRawOutput() requires QIR Server Helper "
+            "but QPU named " +
+            qpuName + " is not implementing it.");
```
khalatepradnya: Can we make this message clearer and actionable, something like - 
```suggestion
        throw std::runtime_error("To support `run` API, " + qpuName + "must inherit `QirServerHelper` class");
```

runtime/common/RecordLogParser.cpp
1tnguyen: I think this is specific to the way CUDA-Q generates QIR for `sample`: we convert `mz` to a sequence of `__quantum__qis__mz__body` -> `__quantum__rt__result_record_output`. Hence, they are logged as a sequence of `RESULT`.

targettests/quantinuum_ng/run_rwpe.cpp
```diff
@@ -0,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// clang-format off
+// RUN: nvq++ %cpp_std --target quantinuum               --emulate %s -o %t && %t | FileCheck %s
```
khalatepradnya: nit
```suggestion
// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t | FileCheck %s
```

targettests/quantinuum_ng/sample_grover.cpp
```diff
@@ -0,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// clang-format off
+// RUN: nvq++ %cpp_std --target quantinuum               --emulate %s -o %t && %t | FileCheck %s
```
khalatepradnya: nit
```suggestion
// RUN: nvq++ %cpp_std --target quantinuum --emulate %s -o %t && %t | FileCheck %s
```

unittests/backends/quantinuum/QuantinuumNGTester.cpp
```diff
@@ -0,0 +1,237 @@
...
+  auto result = future.get();
+  result.dump();
+
+  printf("ENERGY: %lf\n", result.expectation());
+  EXPECT_TRUE(isValidExpVal(result.expectation()));
+}
+
+CUDAQ_TEST(QuantinuumNGTester, checkControlledRotations) {
+  // Checks for more advanced controlled rotations that only
+  // work in emulation.
```
khalatepradnya: Q: Does this comment apply to the next gen device as well?

unittests/backends/quantinuum/QuantinuumNGTester.cpp
```diff
@@ -0,0 +1,237 @@
...
+    kernel.rx<cudaq::ctrl>(M_PI, controls1, controls2, control3, target);
+    kernel.mz(controls1);
+    kernel.mz(controls2);
+    kernel.mz(control3);
+    kernel.mz(target);
+    auto counts = cudaq::sample(kernel);
+    counts.dump();
+    EXPECT_EQ(counts.count("11111111"), 1000);
+  }
+}
```
khalatepradnya, line 214: Q: Whay doesn't this need the Google test init ?

unittests/output_record/RecordParserTester.cpp
```diff
@@ -463,3 +463,108 @@ CUDAQ_TEST(ParserTester, checkFailureCases) {
     EXPECT_ANY_THROW(parser.parse(missingIndex));
   }
 }
+
+CUDAQ_TEST(ParserTester, checkResultType) {
+  const std::string log =
+      "HEADER\tschema_id\tlabeled\nHEADER\tschema_version\t1."
+      "0\nSTART\nMETADATA\tentry_point\nMETADATA\tqir_profiles\tadaptive_"
+      "profile\nMETADATA\trequired_num_qubits\t10\nMETADATA\trequired_num_"
+      "results\t10\nOUTPUT\tRESULT\t0\tr00000\nOUTPUT\tRESULT\t0\tr00001\nOUTPU"
```
khalatepradnya, line 472: So there's no delimiter `START` / `END` in between metadata and the first result?

khalatepradnya (commented):
Looking good! Thanks, @1tnguyen .
Some Qs and comments.

runtime/common/RecordLogParser.cpp
1tnguyen: In all the current spec versions, it should be safe to associate `RESULT` type with a bool (`i1` type).
I think the sub-type annotation (e.g., different integer types) in the label is somewhat implementation-specific, not specified in the spec.

.github/workflows/integration_tests.yml
khalatepradnya: Something like the `configuration-matrix` in `nvidia.yml`?

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
khalatepradnya: This was covered in internal discussion.

1tnguyen: As mentioned in the TODO, this is a temporary extension to the validation criteria applied to both legacy and NG devices. After #3365, we would need a different solution to adapt both the QIR code version (`CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT`) and any QIR extension capabilities based on `machine`.

unittests/backends/quantinuum/QuantinuumNGTester.cpp
1tnguyen: Ah, we combine this and the existing tester file as a single test app, hence we only need a single `main`.

unittests/output_record/RecordParserTester.cpp
1tnguyen: No, I think `START` / `END` is only used to wrap around a single shot result, and metadata goes in the first shot.

1tnguyen:
> /ok to test 8787cc75879f624e6d9b4afdd1f533feb5da383c

unittests/backends/quantinuum/QuantinuumNGTester.cpp
1tnguyen: Good point. No, it's not. I've removed it in 8787cc75879f624e6d9b4afdd1f533feb5da383c

utils/mock_qpu/quantinuum/__init__.py
1tnguyen: In 8787cc75879f624e6d9b4afdd1f533feb5da383c, I've wrapped all these print statements in a `verbose` flag (disabled by default).
I reckon it might be helpful for debugging purposes as the code gen pipeline is rapidly evolving :)

unittests/output_record/RecordParserTester.cpp
khalatepradnya: Thanks for the clarification. It seems odd, IMO.

1tnguyen:
> /ok to test 082b2ddc8b580ec434ef0f542b251fcad91ae98d

1tnguyen:
> /ok to test eb1343aa87a4ef7ebe0f9ac0351fcd3ac251fead

khalatepradnya (approved):

1tnguyen:
> /ok to test 05b371b74757046051aaebbd2d54560f00c840b5

1tnguyen:
> /ok to test 99fa2486e11db9af0a873a376ec1ae9378cb5453

1tnguyen:
> /ok to test 5815ed5dd0ce55bf51232921592c56f8a7d647a6

1tnguyen:
> /ok to test 1bcb7b59b8072ffd4611454f252222cf9cb43471

bmhowe23 (approved):
I reviewed the .github folder.

1tnguyen:
> /ok to test 75fbb30ea84b87fe4e0ea0429a5a6e78105c1e3c

PR #3335, opened 2025-08-12
[py][nfc] Move a couple of analysis to the proper place

boschmitt:
Moves the implementations of `ValidateArgumentAnnotations` and `ValidateReturnStatements` to the `analysis` module.


python/cudaq/kernel/ast_bridge.py
```diff
@@ -4605,8 +4539,8 @@ def compile_to_mlir(astModule, capturedDataStorage: CapturedDataStorage,
                          locationOffset=lineNumberOffset,
                          capturedVariables=parentVariables)
 
-    # First validate the arguments, make sure they are annotated
-    bridge.validateArgumentAnnotations(astModule)
+    ValidateArgumentAnnotations(bridge).visit(astModule)
```
sacpis, line 4542: nit: Should it be `bridge.ctx`?

boschmitt: Thanks for the comment. Why would it be `bridge.ctx`? These classes take the bridge itself as input in their initializers, even when they were inside the bridge they were created by passing `self`. Also `bridge.ctx` is the MLIR context object, these analyses work on the Python AST.

PR #3337, opened 2025-08-12
[py] Fix loop creation in PyASTBridge

boschmitt:
According to our definitions: A "invariant loop" is a loop that is _guaranteed_ to execute the body of the loop `totalIterations` times. _Early exits are not allowed_.

Loops created through `createInvariantForLoop` are not guaranteed to be invariant. This change renames `createInvariantForLoop` to `createForLoop` and add a flag parameter that signals whether the loop is indeed invariant.

### DIFF
--- a/python/tests/mlir/qreg_iterable.py
+++ b/python/tests/mlir/qreg_iterable.py
@@ -43,6 +43,6 @@ def test_qreg_iter():
 # CHECK:           ^bb0(%[[VAL_12:.*]]: i64):
 # CHECK:             %[[VAL_13:.*]] = arith.addi %[[VAL_12]], %[[VAL_1]] : i64
 # CHECK:             cc.continue %[[VAL_13]] : i64
-# CHECK:           } {invariant}
+# CHECK:           }
 # CHECK:           return
 # CHECK:         }
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -627,7 +627,7 @@ class PyASTBridge(ast.NodeVisitor):
                                             [iterVar], rawIndex).result
             cc.StoreOp(castedEle, targetEleAddr)
 
-        self.createInvariantForLoop(sourceSize, bodyBuilder)
+        self.createForLoop(sourceSize, bodyBuilder, invariant=True)
         return cc.StdvecInitOp(targetVecTy, targetPtr, length=sourceSize).result
 
     def __insertDbgStmt(self, value, dbgStmt):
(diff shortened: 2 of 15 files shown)


python/tests/mlir/qreg_iterable.py
```diff
@@ -43,6 +43,6 @@ def foo(N: int):
 # CHECK:           ^bb0(%[[VAL_12:.*]]: i64):
 # CHECK:             %[[VAL_13:.*]] = arith.addi %[[VAL_12]], %[[VAL_1]] : i64
 # CHECK:             cc.continue %[[VAL_13]] : i64
-# CHECK:           } {invariant}
+# CHECK:           }
```
schweitzpgi, line 46: This loop is invariant as it will iterate once on each qubit in the qvector. Did you mean to omit the attribute?

On the other hand, loop analysis will detect that it is an invariant loop, so there's no correctness issue.

schweitzpgi:
The comments in loop analysis seem to be slightly behind the implementation. I should fix that.

We don't allow early exits for counted loops. But invariant and monotonic loops can either allow or disallow early exits. There is an argument to their analysis to select the exact species precisely.

Generally speaking, one shouldn't be adding the loop analysis attributes to cc.loop Ops a priori. It's better to let loop analysis do so. In some rare cases, the front end may add an attribute to a trivial construct because it understands that something like `mz(qs);` must iterate over all the qubits in the qvector.

python/tests/mlir/qreg_iterable.py
boschmitt: It's just that the AST bridge don't analyze the loop to know whether is indeed invariant or not. So it's kind of a best-effort sort of tagging.

boschmitt:
> Generally speaking, one shouldn't be adding the loop analysis attributes to cc.loop Ops a priori. It's better to let loop analysis do so. In some rare cases, the front end may add an attribute to a trivial construct because it understands that something like `mz(qs);` must iterate over all the qubits in the qvector.

Perhaps it is better to remove adding this attribute from the bridge. I wasn't sure whether that would be a good idea. I suppose this loop creation helper is a Python implementation of:

https://github.com/NVIDIA/cuda-quantum/blob/9653cc9473b4a08cfcc52593e693d6260a4fbbf4/lib/Optimizer/Builder/Factory.cpp#L200-L234

schweitzpgi (approved):
LGTM

schweitzpgi:
@boschmitt Do you want to take another crack at this?

boschmitt:
> @boschmitt Do you want to take another crack at this?

I will close this since the bridge has change quite a bit since then. I would need to take a look to see if this still an issue.

PR #3340, opened 2025-08-12
[core] Reimplement cudaq::run for C++ and Python.

schweitzpgi:
For certain transport layers, there is a hard restriction that a kernel cannot return a value. In CUDA-Q however, kernels have always been allowed to return values. Moreover, kernels used with cudaq::run() are *required* to return values.

Prior to these changes, the hard restrictions of the transport layer were not being met. With these changes an entirely new kernel launch stack, specifically for use with cudaq::run, will be autogenerated. This allows for the kernel to exist as a CUDA-Q kernel and co-exist in a functionally complete and semantically correct way with the host language, C++. At the same time, the new second launch stack can exist and break the C++ semantics so that new launch stack will unceremoniously drop return values on the floor. This achieves the restriction imposed by the transport layer at the expense of (mostly) duplicated code.

Additionally, instead of injecting output logging code into all entry point kernels, the output logging code will be injected into the cudaq::run kernels.

Add new op for output logging.


schweitzpgi:
This requires more work for Python, which has a different implementation of `cudaq.run`.

runtime/cudaq/cudaq.cpp
```diff
@@ -246,12 +246,33 @@ static std::vector<std::string> kernelRegistry;
...
+  auto iter = runnableKernelRegistry.find(kernelName);
+  return (iter != runnableKernelRegistry.end()) ? iter->second : nullptr;
+}
+
+void *
+cudaq::registry::__cudaq_getRunnableKernel(const std::string &kernelName) {
+  void *result = getRunnableKernelOrNull(kernelName);
+  if (!result)
+    throw std::runtime_error("runnable kernel is not present: kernel cannot be "
+                             "called by cudaq::run");
```
bmhowe23, line 272: This new error is getting tripped in some of my local CUDA-QX tests if I remove `-fenable-cudaq-run` from the compilation of the top-level file.

bmhowe23: Marking as resolved because I'll just leave the `-fenable-cudaq-run` settings alone.

schweitzpgi: It might be worth filing an issue to keep track of this rather weird set of flaming hoop leap throughs needed to make the code run with run. But I left all that alone for now.

test/AST-Quake/cudaq_run.cpp
```diff
@@ -20,309 +21,184 @@ struct K9 {
...
-// CHECK:           %[[VAL_25:.*]] = cc.cast %[[VAL_24]] : (!cc.ptr<!cc.array<i8 x 4>>) -> !cc.ptr<i8>
-// CHECK:           call @__quantum__rt__bool_record_output(%[[VAL_23]], %[[VAL_25]]) : (i1, !cc.ptr<i8>) -> ()
-// CHECK:           return {cc.cudaq.run} %[[VAL_14]] : !cc.stdvec<i1>
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__K9.run()
+// CHECK:           %[[VAL_0:.*]] = call @__nvqpp__mlirgen__K9() : () -> !cc.stdvec<i1>
+// CHECK:           return
 // CHECK:         }
-// clang-format on
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__K9.run.entry(
```
khalatepradnya, line 72: Q: Why doesn't this have the attribute `cudaq-entrypoint`?

khalatepradnya: (I feel like that's the point of this PR! 😅)

schweitzpgi: It's not a kernel on the device side. We do not tag host side code with `cudaq-entrypoint`.

bmhowe23 (commented):
This LGTM. Hopefully @khalatepradnya and/or @annagrin can give it the official review and approval?

runtime/cudaq/algorithms/sample.h
```diff
@@ -16,7 +16,7 @@
 
 namespace cudaq {
 bool kernelHasConditionalFeedback(const std::string &);
-namespace __internal__ {
+namespace detail {
```
sacpis, line 19: nit: should the namespace be `details`?

schweitzpgi: No. We're changing all these `details` and `__internal__` to simply `detail`.

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -87,33 +105,32 @@ RunResultSpan pyRunTheKernel(const std::string &name, MlirModule module,
...
+      platform, name, origName, shots_count, qpu_id);
 
   std::free(rawArgs);
   return results;
 }
 
-std::vector<py::object> pyReadResults(RunResultSpan results, MlirModule module,
-                                      func::FuncOp funcOp,
-                                      std::size_t shots_count) {
+static std::vector<py::object> pyReadResults(details::RunResultSpan results,
```
sacpis, line 114: nit: As we changing `details` and `__internal__`, should we make it `detail` here as well?

runtime/cudaq/algorithms/sample.h
sacpis: Thanks @schweitzpgi. I see `details` word at other places. Should we change that to `detail` as well?

python/runtime/cudaq/algorithms/py_run.cpp
schweitzpgi: No. This PR isn't about making those naming changes everywhere in the code. That should (and was) done in another PR that got derailed.

sacpis: Got it. Thank you.

test/AST-Quake/cudaq_run.cpp
```diff
@@ -7,11 +7,12 @@
  ******************************************************************************/
 
 // clang-format off
-// RUN: cudaq-quake --cudaq-run=1 %cpp_std %s | cudaq-opt --add-dealloc --expand-measurements --factor-quantum-alloc --expand-control-veqs --cc-loop-unroll --canonicalize --multicontrol-decomposition --lower-to-cfg --cse --decomposition=enable-patterns="CCXToCCZ,CCZToCX" --combine-quantum-alloc --canonicalize --convert-to-qir-api --return-to-output-log --symbol-dce --canonicalize | FileCheck %s
+// RUN: cudaq-quake --cudaq-run=1 %cpp_std %s | cudaq-opt --kernel-execution=generate-run-stack=1 --add-dealloc --expand-measurements --factor-quantum-alloc --expand-control-veqs --cc-loop-unroll --canonicalize --multicontrol-decomposition --lower-to-cfg --cse --decomposition=enable-patterns="CCXToCCZ,CCZToCX" --combine-quantum-alloc --canonicalize --convert-to-qir-api --return-to-output-log --symbol-dce --canonicalize | FileCheck %s
 // clang-format on
 
 #include <cudaq.h>
 
+// FIXME: Results that are vectors are dropped on the floor!
```
khalatepradnya, line 15: Is there a task to capture this `FIXME`?

python/tests/kernel/test_run_kernel.py
```diff
@@ -888,12 +888,12 @@ class MyClass:
         y: bool
 
     @cudaq.kernel
-    def simple_struct(t: MyClass) -> MyClass:
+    def simple_strucA(t: MyClass) -> MyClass:
```
khalatepradnya, line 891: About need to update names in the tests, is this issue same as https://github.com/NVIDIA/cuda-quantum/issues/3347?

khalatepradnya (approved):
LGTM. 👍🏽Thanks, Eric! 
Please capture follow-ups / TODOs for tracking.

python/tests/kernel/test_run_kernel.py
schweitzpgi: Yes. 👍

test/AST-Quake/cudaq_run.cpp
schweitzpgi: This is a carry-over from the existing implementation. I am not sure if we ever made an issue for it. I assumed that there was one.

schweitzpgi: #3013 looks like the grab bag of issues.

mitchdz:
> /ok to test 627bcfe

PR #3353, opened 2025-08-16
Allow .so files with `__qpu__` kernels be well isolated from CUDA-Q simulator and platform libs

bmhowe23:
This change makes it so that .so files that contain `__qpu__` kernels no longer have to bring in so many unnecessary additional dependencies into applications if the execution of those kernels isn't actually required in that application.

One might argue that `__qpu__` kernels are themselves unnecessary backage if they aren't really needed in an application (and therefore *they* should be the ones removed), and while this is - to some extent - true, the `__qpu__` kernels themselves are very lightweight baggage whereas the full CUDA-Q runtime is _much_ larger. It is more prudent to break the connection here rather than to require that every .so file that contains `__qpu__` kernels to carry so many dependencies with it.

Most of the changes in this PR are moving code from `cudaq.cpp` (`libcudaq.so`) to `runtime/common/DeviceCodeRegistry.cpp` (`libcudaq-common.so`). The latter is a lightweight library that doesn't require NVQIR+simulators while the former is not.


lib/Optimizer/CodeGen/ConvertCCToLLVM.cpp
```diff
@@ -133,6 +133,18 @@ struct CCToLLVM : public cudaq::opt::impl::CCToLLVMBase<CCToLLVM> {
...
     }
+
+    // Set the __ctl functions to extern_weak linkage. These need special
+    // linkage because they are passed to
+    // generalizedInvokeWithRotationsControlsTargets as a function parameter.
+    // Without this change, one needs to bring the entire NVQIR world (with
+    // simulators and platforms) in order to dlopen a .so file that contains
+    // __qpu__ kernels.
+    auto mod = getOperation();
+    mod.walk([&](LLVM::LLVMFuncOp funcOp) {
```
schweitzpgi: There is no need to walk the entire depth and breadth of the IR to find functions. They appear at the top-level under a Module. So a simple iteration over the module is better.
```c++
  for (auto f : mod)
    if (isa<LLVM::LLVMFuncOp>(f))
```

bmhowe23: Updated in 97f568e46e

lib/Optimizer/CodeGen/ConvertCCToLLVM.cpp
```diff
@@ -133,6 +133,17 @@ struct CCToLLVM : public cudaq::opt::impl::CCToLLVMBase<CCToLLVM> {
       LLVM_DEBUG(getOperation().dump());
       signalPassFailure();
     }
+
+    // Set the __ctl functions to extern_weak linkage. These need special
```
schweitzpgi, line 137: This is a hack that papers over the actual problem. Please provide a way to reproduce the dlopen issue itself.

bmhowe23: This reproducer is copied from our Slack thread from late last week:
```
################### Put this into repro.cpp:
#include <cudaq.h>

__qpu__ void kernel(cudaq::qview<> q) {
  cudaq::x(q[0]);
  x<cudaq::ctrl>(q[0], q[1]);
}

extern "C" int my_other_func(int x) { return 2 * x; }

################### Put this into repro_main.cpp
#include <stdio.h>
#include <dlfcn.h>

int main() {
  // Dynamically load the .so file.
  printf("Loading librepro.so\n");
  void *handle = dlopen("librepro.so", RTLD_LAZY);
  if (!handle) {
    printf("Failed to load librepro.so: %s\n", dlerror());
    return 1;
  }
  printf("Loaded librepro.so\n");

  // Get the address of the function my_other_func.
  auto my_other_func =
      reinterpret_cast<int (*)(int)>(dlsym(handle, "my_other_func"));
  if (!my_other_func) {
    printf("Failed to get address of my_other_func: %s\n", dlerror());
    return 1;
  }

  printf("my_other_func(1): %d\n", my_other_func(1));
}

############ Build commands for the .so file
nvq++ -v --save-temps -c -fPIC --enable-mlir repro.cpp -o repro.o
/usr/local/llvm/bootstrap/cxx -fPIC -O2 -Wl,--no-as-needed -lcudaq-common -lcudaq-operator -shared -o librepro.so repro.o
############ Build commands for the application
g++ repro_main.cpp -ldl

############ Run the application
LD_LIBRARY_PATH=/usr/local/cudaq/lib:. ./a.out
```

bmhowe23:
Closing in lieu of a yet-to-be-determined solution.

PR #3358, opened 2025-08-18
Use legacy cgi

mitchdz:
cgi has been deprecated from python3.13 https://peps.python.org/pep-0594/. Fall back to legacy-cgi for python3.13


mitchdz:
> /ok to test 8132283

PR #3360, opened 2025-08-19
Support resource estimation on remote simulators

atgeller:
### Description

Expands https://github.com/NVIDIA/cuda-quantum/pull/3051 to handle remote simulators as well.


atgeller:
> /ok to test 062ed21

runtime/common/RemoteKernelExecutor.h
```diff
@@ -85,6 +86,13 @@ class RemoteRuntimeClient
   // Return the remote capabilities of the server.
   virtual RemoteCapabilities getRemoteCapabilities() const = 0;
 
+  virtual mlir::ModuleOp lowerKernel(mlir::MLIRContext &mlirContext,
```
1tnguyen, line 93: nit: could we add a comment here about its functionality?

runtime/common/BaseRemoteSimulatorQPU.h
```diff
@@ -11,15 +11,26 @@
...
 #include "cudaq/algorithms/optimizer.h"
 #include "cudaq/platform/qpu.h"
 #include "cudaq/platform/quantum_platform.h"
+#include "mlir/ExecutionEngine/ExecutionEngine.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
 #include <fstream>
 
+namespace nvqir {
+cudaq::Resources *getResourceCounts();
```
amccaskey: we really can't be coupling all these libraries to the NVQIR library

runtime/common/RemoteKernelExecutor.h
atgeller: Done

runtime/common/BaseRemoteSimulatorQPU.h
atgeller: I can work around this here in the short term. For the longer term it may be nicer to attach the resource counter and/or relevant metadata to either the execution manager or platform to avoid needing the NVQIR library in the QPUs to nicely handle resource counting. Thoughts @1tnguyen?

atgeller: I've removed the dependency

atgeller:
> /ok to test 1425533

atgeller:
> /ok to test 6259187

runtime/common/BaseRestRemoteClient.h
```diff
@@ -118,14 +118,12 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
     return cudaq::RestRequest::REST_PAYLOAD_VERSION;
   }
 
-  std::string constructKernelPayload(
-      mlir::MLIRContext &mlirContext, const std::string &name,
-      void (*kernelFunc)(void *), const void *args, std::uint64_t voidStarSize,
```
schweitzpgi, line 123: Why was the `kernelFunc` parameter removed here?

schweitzpgi: Kernel launch is intended to always supply sufficient arguments for various forms of call. Here, we probably aren't going to call the function in the current process, but it's not great to have its own special signature either.

PR #3361, opened 2025-08-19
Remove poor practice of using namespaces in OpaqueArguments.h

### LINKED ISSUE #3336: header file OpaqueArguments.h uses poor practice
The header file `python/utils/OpaqueArguments.h` uses poor practice and opens two namespaces with `using namespace X;` commands.

Both of these lines must be deleted and any erroneous code in other files that happens to compile by these namespaces being left open must be repaired as well by explicitly adding the namespace tag(s).

hrushikesh890:
### Description

This issue resolves #3336 .
Removed `using namespace std::chrono_literals` and `using namespace mlir` from OpaqueArguments.h
Added scope resolution operators in affected functions and files for mlir. No changes needed for chrono_literals.


hrushikesh890:
Updated: `py_run.cpp` now compatible with recent `main`-branch changes
Updated: Fixed clang format issues

hrushikesh890:
Fixed clang formatting issues again. Quick question how can check the clang format issues before pushing my commits? I tried using the run_clang_format.sh in scripts but that did not show me any errors.

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -60,16 +60,16 @@ getKernelLaunchParameters(py::object &kernel, py::args args) {
 
   // Lookup the runnable kernel.
   auto mod = unwrap(kernelMod);
-  auto runKern = mod.lookupSymbol<func::FuncOp>(
+  auto runKern = mod.lookupSymbol<mlir::func::FuncOp>(
```
schweitzpgi, line 63: For clarity: With an implementation file (`.cpp`), it is acceptable to have a single `using namespace` _foo_`;` at the top of the module after the `#include` directives.

schweitzpgi: In this case, there aren't that many `mlir::` namespaces being added, so it doesn't make much difference.

schweitzpgi:
> /ok to test 1f0e7e4

schweitzpgi:
> /ok to test [3661cf9](https://github.com/NVIDIA/cuda-quantum/pull/3361/commits/3661cf9cde18ff7431f7c5f9793e7b796293b24c)

PR #3363, opened 2025-08-19
[DNM] use --find-links and drop pypiserver

mitchdz:
Testing using pip's native `--find-links` instead of pypiserver.


mitchdz:
this will be included in https://github.com/NVIDIA/cuda-quantum/pull/3421, dropping.

PR #3308, opened 2025-08-06
Generate md files for CUDA-Q

sacpis:
Generate markdown files corresponding to .html files in CUDA-Q. This is needed to improve the search for CUDA-Q keywords when searching through LLMs.

LLMs use llms.txt file to search for the knowledge base related to the search terms.

PR #3364, opened 2025-08-19
Adding llms with paths

sacpis:
Adding llms with paths to resources.

Following PR should go in before this one
https://github.com/NVIDIA/cuda-quantum/pull/3308


latest/llms.txt
```diff
@@ -0,0 +1,29 @@
+# NVIDIA CUDA-Q
+
+> CUDA-Q streamlines hybrid application development and promotes productivity and scalability in quantum computing. It offers a unified programming model designed for a hybrid setting—that is, CPUs, GPUs, and QPUs working together. CUDA-Q contains support for programming in Python and in C++.
```
schweitzpgi: What do you think about using the word "heterogeneous" instead of "hybrid" here?

sacpis: As per the product requirement, it should directly come from [CUDA-Q doc](https://developer.nvidia.com/cuda-q#section-benefits).

@efratshabtai for viz.

latest/llms.txt
```diff
@@ -0,0 +1,29 @@
...
+
+- [Quick Start](https://nvidia.github.io/cuda-quantum/latest/using/quick_start.md)
+
+## API
+
+- [API Reference](https://nvidia.github.io/cuda-quantum/latest/api/api.md)
+
+## Examples
+
+- [Basics](https://nvidia.github.io/cuda-quantum/latest/using/basics/basics.md)
```
bettinaheim: Instead of a full path here, I think we can and should simply use the relative path `latest/using/basics/basics.md`

sacpis: Fixed in [608a7a0](https://github.com/NVIDIA/cuda-quantum/pull/3364/commits/608a7a08ea5143006915a1c29127335649afed02).

latest/llms.txt
sacpis: @schweitzpgi Fixed in [529315c](https://github.com/NVIDIA/cuda-quantum/pull/3364/commits/529315c360dd0c317eccc4819c4b267eab990f8a).

Issue #3312, opened 2025-08-07
Follow up work to clean up the initial implementations for run and device calls


bettinaheim:
Creating an issue to track some items that are left as todos in PRs:

- [ ] In runtime/common/BaseRemoteRestQPU.h,: TODO - clean this up to not have to do this. Considering the module's map, or cudaq::details::getKernelName(). But make sure it works for
- [ ] In runtime/common/RuntimeMLIR.cpp: FIXME: don't use a recursive `walk` to DFS for FuncOps, which appear as children, in a Module.
- [ ] Add proper validation to not permit arbitrary/non-profile compliant calls in lib/Optimizer/CodeGen/VerifyQIRProfile.cpp
- [ ] reexamine filterSpecificCodePatterns in runtime/common/RuntimeMLIRCommonImpl.h 
- [ ] get rid of --enable-run
- [ ] eliminate the use of a range of different `CUDAQ_` environment variables to manually configure the compiler

PR #3365, opened 2025-08-19
Follow up work from #3223.

schweitzpgi:
See also #3312.


lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -578,23 +578,27 @@ struct DiscriminateOpRewrite
...
 
   LogicalResult
   matchAndRewrite(quake::DiscriminateOp disc, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     if constexpr (M::discriminateToClassical) {
-      rewriter.replaceOpWithNewOp<func::CallOp>(
-          disc, rewriter.getI1Type(),
-          cudaq::opt::getQIRReadResultBody(qirVersionUnderDevelopment),
-          adaptor.getOperands());
+      if constexpr (M::qirVersion == QirVersion::version_0_2) {
```
annagrin, line 600: Should we have a helper that returns the correct `ReadResult` depending on the version, so we don't have to test here and other places?

utils/mock_qpu/anyon/__init__.py
```diff
@@ -39,16 +38,16 @@ class Job(BaseModel):
...
 
-qirVersionUnderDevelopment = os.environ.get(
-    "CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT", False)
-requiredQubits = "required_num_qubits" if qirVersionUnderDevelopment else "requiredQubits"
-
 
 def getNumRequiredQubits(function):
     for a in function.attributes:
-        if requiredQubits in str(a):
+        if "required_num_qubits" in str(a):
```
annagrin, line 44: Looks like the tests will pass for either value, not depending on the current version, so they won't catch incorrect fields in CI... Do we want them to be stricter and only pass for the current version of QIR?

include/cudaq/Optimizer/CodeGen/Passes.h
```diff
@@ -72,8 +67,7 @@ mlir::LLVM::LLVMStructType lambdaAsPairOfPointers(mlir::MLIRContext *context);
 /// before conversion to the LLVM-IR dialect.
 void registerToQIRAPIPipeline();
 void addConvertToQIRAPIPipeline(mlir::OpPassManager &pm, mlir::StringRef api,
-                                bool opaquePtr,
-                                bool qirVersionUnderDevelopment);
+                                bool opaquePtr = false);
```
khalatepradnya, line 73: Can you add comment on why this argument is now `false` by default?

include/cudaq/Optimizer/CodeGen/Peephole.h
```diff
@@ -56,14 +56,13 @@ inline mlir::Value createMeasureCall(mlir::PatternRewriter &builder,
 }
 
 inline mlir::Value createReadResultCall(mlir::PatternRewriter &builder,
-                                        mlir::Location loc, mlir::Value result,
-                                        bool qirVersionUnderDevelopment) {
-
+                                        mlir::Location loc,
+                                        mlir::Value result) {
+  // NB: This code is only used from a deprecated pass.
```
khalatepradnya, line 61: Q: Is it time to delete / clean up deprecated pass? Can be follow-up, might be helpful to call out which pass that is.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -608,8 +612,7 @@ struct DiscriminateOpToCallRewrite
     return success();
   }
 
-private:
-  bool qirVersionUnderDevelopment;
+  const std::string version;
```
khalatepradnya, line 622: nit: Variable name `qirVersion`?

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -12,8 +12,10 @@ using namespace mlir;
 
 void cudaq::opt::commonPipelineConvertToQIR(PassManager &pm,
                                             StringRef codeGenFor,
-                                            StringRef passConfigAs,
-                                            bool qirVersionUnderDevelopment) {
+                                            StringRef passConfigAs) {
+  auto codeGenFields = codeGenFor.split(':');
+  auto passConfigFields = passConfigAs.split(':');
```
annagrin, line 17: Should we create a helper to get out a version and format, so the code is easier to understand?

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
```diff
@@ -550,8 +543,7 @@ struct WireSetToProfileQIRPrepPass
     addBodyDecl("mz", measTy);
     auto readResTy = FunctionType::get(ctx, TypeRange{resTy},
                                        TypeRange{builder.getI1Type()});
-    createNewDecl(cudaq::opt::getQIRReadResultBody(qirVersionUnderDevelopment),
-                  readResTy);
+    createNewDecl("__quantum__qis__read_result__body", readResTy);
```
khalatepradnya: Q: Why is this hard-coded?

khalatepradnya (commented):
First pass

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
```diff
@@ -372,8 +370,7 @@ struct DiscriminateRewrite : OpConversionPattern<quake::DiscriminateOp> {
         loc, std::nullopt, cudaq::opt::QIRRecordOutput,
         ValueRange{adaptor.getMeasurement(), nameValCStr});
     if (isAdaptiveProfile) {
-      std::string funcName =
-          cudaq::opt::getQIRReadResultBody(qirVersionUnderDevelopment);
+      std::string funcName = toQisBodyName(std::string("read_result"));
```
annagrin, line 375: It looks like the changes in this file only can create QIR version 0.1, what do we do for cases where version 0.2 is needed?

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
annagrin: Should this be a constant instead?

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -314,11 +314,12 @@ class BaseRemoteRESTQPU : public cudaq::QPU {
...
                     config.BackendConfig->CodegenEmission);
-        auto [codeGenName, codeGenOptions] = parseCodeGenTranslationString(
-            config.BackendConfig->CodegenEmission);
-        codegenTranslation = codeGenName;
-        if (codegenTranslation == "qir-adaptive") {
-          for (const auto &option : codeGenOptions) {
+        auto [codeGenName, codeGenVersion, codeGenOptions] =
+            parseCodeGenTranslationString(
+                config.BackendConfig->CodegenEmission);
+        codegenTranslation = config.BackendConfig->CodegenEmission;
```
annagrin, line 317: Nit: can this be moved above the previous statement and use `codegenTranslation` as a parameter in `parseCodeGenTranslationString`?

runtime/common/BaseRestRemoteClient.h
```diff
@@ -234,14 +234,11 @@ class BaseRemoteRestRuntimeClient : public cudaq::RemoteRuntimeClient {
...
                                      const std::vector<void *> *rawArgs) {
-    bool qirVersionUnderDevelopment =
-        getEnvBool("CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT", false);
-
     auto moduleOp = lowerKernel(mlirContext, name, args, voidStarSize,
                                 startingArgIdx, rawArgs);
 
     mlir::PassManager pm(&mlirContext);
-    opt::addPipelineConvertToQIR(pm, qirVersionUnderDevelopment);
+    opt::addPipelineConvertToQIR(pm, "qir:0.1");
```
annagrin, line 243: I think we need this to be configurable, probably by an option to the compiler - is there a follow up to support it?

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -394,19 +324,15 @@ std::size_t getArgAsInteger(llvm::Value *arg) {
...
   for (llvm::Function &func : *llvmModule) {
-    if (func.hasFnAttribute(cudaq::opt::QIREntryPointAttrName)) {
-      requiredNumQubits = func.getFnAttributeAsParsedInteger(
-          cudaq::opt::getQIRRequiredQubitsAttrName(qirVersionUnderDevelopment),
-          requiredNumQubits);
-      requiredNumResults = func.getFnAttributeAsParsedInteger(
-          cudaq::opt::getQIRRequiredResultsAttrName(qirVersionUnderDevelopment),
-          requiredNumResults);
+    if (func.hasFnAttribute("entry_point")) {
+      required_num_qubits = func.getFnAttributeAsParsedInteger(
```
annagrin, line 334: Should the function names depend on the QIR version?

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -621,6 +543,26 @@ qirProfileTranslationFunction(const char *qirProfile, mlir::Operation *op,
                             "dynamic_qubit_management", falseValue);
   llvmModule->addModuleFlag(llvm::Module::ModFlagBehavior::Error,
                             "dynamic_result_management", falseValue);
+  if (isAdaptiveProfile) {
```
annagrin, line 562: Should we support version 0.2 as well here?

include/cudaq/Optimizer/CodeGen/Peephole.h
schweitzpgi: If it were only so easy. :) We have a number of codegen passes sitting around in this directory and most of them are deprecated. But it may be that some of them are still used in one-offs. Yes, we need to clean that up. No, I'm not going to do it as part of this PR.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
schweitzpgi: I don't see that as really needed since everything in the compilation unit is QIR.

lib/Optimizer/CodeGen/Pipelines.cpp
schweitzpgi: Yes. But I want to do that as part of the next refactoring in this ongoing PR...

This is trivial code that belongs in some "base" functionality.

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
schweitzpgi: This is an old codegen pass. If someone is using it to generate "new QIR" then they're mistaken.

runtime/common/BaseRestRemoteClient.h
schweitzpgi: I agree. This needs to come from the .yml file settings. But it should not be reconfigured by, potentially accidental, settings like environment variables.

runtime/common/RuntimeMLIRCommonImpl.h
schweitzpgi: This reminds me ... I didn't find any tests. Do we have a target that wants QIR 0.2?

utils/mock_qpu/anyon/__init__.py
schweitzpgi: This is the server, not a test. I think we need a specific test and not hijack functions in the server, eh?

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
schweitzpgi: Thanks! Good catch.

runtime/common/RuntimeMLIRCommonImpl.h
schweitzpgi: Good catch. Thanks.

include/cudaq/Optimizer/CodeGen/Passes.h
schweitzpgi: a281d090

lib/Optimizer/CodeGen/WireSetsToProfileQIR.cpp
schweitzpgi: a281d090

schweitzpgi: a281d090

runtime/common/BaseRemoteRESTQPU.h
schweitzpgi: a281d090

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
schweitzpgi: This is a compile-time `if constexpr`, so the test isn't a runtime issue. As the mixin is part of the structure of this class and there aren't a ton of places that we need to distinguish the version, leaving it here seems fine.

If and when this gets more complicated, we can write helper templates, but those would be more lines of code for this.

lib/Optimizer/CodeGen/Pipelines.cpp
schweitzpgi: f260c47e

runtime/common/BaseRestRemoteClient.h
schweitzpgi: Discussed with Thien. The server apparently expects full QIR and this wasn't designed to handle different versions. Will open an issue.

khalatepradnya (commented):
LGTM. All my comments have been addressed.
Perhaps @annagrin give final approval?

schweitzpgi:
I think this one has marinated in CI long enough. I'll add follow-up issues and plan to add some directed tests in another PR. Stay tuned.

bmhowe23 (approved):

PR #3371, opened 2025-08-20
[core] Fix a type error in SROA pass.

### LINKED ISSUE #3370: Runtime error when modifying tuples inside kernels
### Required prerequisites

### Describe the bug

Copying and modifying tuples inside kernels fails after SROA optimization, but only after some other kernel runs.

```c++
// clang-format off
// RUN: nvq++ -fenable-cudaq-run --target quantinuum --emulate test.cpp -o t.x && CUDAQ_ENABLE_QUANTUM_DEVICE_RUN=1 ./t.x
// clang-format on

#include <cudaq.h>

struct MyTuple {
  bool boolVal;
  std::int64_t i64Val;
  double f64Val;
};

auto struct_test = []() __qpu__ {
  MyTuple t = {true, 654, 9.123};
  t.boolVal = false;
  t.i64Val = 13;
  t.f64Val = 3.14;
  return t;
};

auto struct_test_param = [](MyTuple t1) __qpu__ {
  MyTuple t = t1;
  t.boolVal = false;
  t.i64Val = 13;
  t.f64Val = 3.14;
  return t;
};

int main() {
  int c = 0;
  {
    MyTuple t = {true, 654, 9.123};
    const auto results1 = cudaq::run(3, struct_test); // Commenting this line out makes the test work
    const auto results = cudaq::run(3, struct_test_param, t);
    if (results.size() != 3) {
      printf("FAILED! Expected 3 shots. Got %lu\n", results.size());
    } else {
      c = 0;
      for (auto i : results)
        printf("%d: {%s, %ld, %f}\n", c++, i.boolVal ? "true" : "false",
               i.i64Val, i.f64Val);
      printf("success!\n");
    }
  }

  return 0;
}
```

Output:
```
loc("-":9:10): error: 'cc.compute_ptr' op result type does not match input
  what():  Remote rest platform Quake lowering failed.
Aborted (core dumped)
```

### Expected behavior

Test should run and output:

```
success!
```

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

schweitzpgi:
Add CHECK lines to the commits.

Fixes #3370.

### DIFF
--- a/lib/Optimizer/Transforms/SROA.cpp
+++ b/lib/Optimizer/Transforms/SROA.cpp
@@ -168,7 +168,6 @@ public:
     // It's a match. We have a chain of insert_values back to a undef. Let's
     // replace this store.
     Value dest = storeOp.getPtrvalue();
-    Type baseTy = dest.getType();
     for (auto insVal : stack) {
       // Each insert_value is converted to a compute_ptr and store.
       auto v = insVal.getValue();
--- a/lib/Optimizer/Transforms/SROA.cpp
+++ b/lib/Optimizer/Transforms/SROA.cpp
@@ -176,8 +176,9 @@ public:
       for (std::int32_t off : insVal.getPosition())
         args.push_back(off);
       auto loc = insVal.getLoc();
+      auto vTy = cudaq::cc::PointerType::get(v.getType());
       auto toAddr =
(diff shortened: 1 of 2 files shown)


annagrin (approved):
LGTM, Thanks Eric!

PR #3389, opened 2025-08-25
[CI] [validation script] Update regex to match single quoted strings as well

khalatepradnya:
* Allow use of single-quotes as well since both `cudaq.set_target("foo")` and `cudaq.set_target('foo')` are valid and
the script needs to skip both.

### DIFF
--- a/scripts/validate_container.sh
+++ b/scripts/validate_container.sh
@@ -313,7 +313,7 @@ do
     let "samples+=1"
 
     skip_example=false
-    explicit_targets=`cat $ex | grep -Po '^\s*cudaq.set_target\("\K.*(?=")'`
+    explicit_targets=`cat $ex | grep -Po '^\s*cudaq.set_target\(['"]\K[^'"]+(?=['"])'`
     for t in $explicit_targets; do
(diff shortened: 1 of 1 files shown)


scripts/validate_container.sh
```diff
@@ -313,7 +313,7 @@ do
     let "samples+=1"
 
     skip_example=false
-    explicit_targets=`cat $ex | grep -Po '^\s*cudaq.set_target\("\K.*(?=")'`
+    explicit_targets=`cat $ex | grep -Po '^\s*cudaq.set_target\(['"]\K[^'"]+(?=['"])'`
```
khalatepradnya: oops.. has a typo.
https://github.com/NVIDIA/cuda-quantum/actions/runs/17220850468/job/48857274674?pr=3389

sacpis: Seems like these examples do not have cudaq.set_target. Are these files being skipped from running through CI?

Will paste the corrected regex in few min.

sacpis: Please try this regex 

```
"^\s*cudaq\.set_target\((['\"])\K.*?(?=\1)"
```

mitchdz (approved):
LGTM.

PR #3390, opened 2025-08-26
Support architecture-specific customization in target config

1tnguyen:
### Description

Due to the evolution of the QIR spec and hardware backend providers, we need to provide customization points within the same target configuration.

Note: this would provide a unified user experience, e.g., specifying different machines/devices within a target, target name is usually associated with a provider. In other words, we don't need to introduce new targets.

Details:

- Provide a machine config schema type, which allows us to annotate an argument as the one associated with device names. This field then allows matching based on an explicit list or a regex pattern. 

- The fields that we allow overriding are extendable. Currently, I only add code gen emission.

- Add a helper to determine the value based on nvq++ CLI/Python `set_target` arguments.

- Extend quantinuum target config to support new QIR spec (0.2) for next-gen Helios devices. https://github.com/NVIDIA/cuda-quantum/pull/3332 depends on this as the `CUDAQ_QIR_VERSION_UNDER_DEVELOPMENT` environment variable has been deprecated

- Add tests.


include/cudaq/Support/TargetConfig.h
```diff
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <map>
```
schweitzpgi, line 11: nit: this isn't used in these header changes AFAICT.

include/cudaq/Support/TargetConfig.h
```diff
@@ -133,6 +162,11 @@ class TargetConfig {
   std::optional<BackendEndConfigEntry> BackendConfig;
   /// Additional configuration mapping (if this is a multi-configuration target)
   std::vector<BackendFeatureMap> ConfigMap;
+
+public:
```
schweitzpgi:
```suggestion
```
This entire class is already public.

lib/Support/Config/TargetConfig.cpp
```diff
@@ -342,5 +353,97 @@ void MappingTraits<cudaq::config::TargetConfig>::mapping(
   io.mapOptional("configuration-matrix", info.ConfigMap);
 }
 
+std::string MappingTraits<cudaq::config::TargetConfig>::validate(
+    IO &io, cudaq::config::TargetConfig &info) {
+  // There should only ever be 1 machine-configuration entry in the target
+  // arguments.
+  int count = 0;
```
schweitzpgi:
```suggestion
  unsigned count = 0;
```
Just to get in the habit. Much of LLVM/MLIR uses unsigned values for lengths and counts...

schweitzpgi (approved):
LGTM. Thanks for helping tidy up this feature, Thien.

python/tests/mlir/qir_profile.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # This device requires qir:0.2
+    cudaq.set_target('quantinuum', emulate=True, machine='Helios-1E')
+    result = cudaq.sample(my_kernel)
+
+
+# CHECK: requiredQubits
+# CHECK: requiredResults
+# CHECK-NG: required_num_qubits
+# CHECK-NG: required_num_results
```
khalatepradnya, line 44: Nice test!

include/cudaq/Support/TargetConfig.h
1tnguyen: We use `std::map` as the argument type for `getCodeGenSpec` :)

PR #3392, opened 2025-08-27
[core] Add a pass to insert measurements on allocated qubits

khalatepradnya:
This PR introduces a new compiler pass that automatically inserts measurement operations into quantum kernels that lack explicit measurements. 
This behavior would ensure compliance with the sampling API specification that all allocated qubits are measured at the end of the kernel if no measurements are present.


lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+namespace {
+
+/// Analysis class that examines a function to determine whether it contains
+/// measurement operations and collects all qubit allocations.
+struct Analysis {
+  Analysis() = default;
+  Analysis(const Analysis &) = delete;
+  Analysis(Analysis &&) = delete;
+  Analysis &operator=(const Analysis &) = delete;
```
schweitzpgi: Since this class is only defined and used in this module, you don't really need these. It's clear that the Analysis object isn't being assigned, copied, or moved.

lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,149 @@
...
+  bool hasMeasurement = false;
+  SmallVector<quake::AllocaOp> allocations;
+
+  bool hasQubitAlloca() const { return !allocations.empty(); }
+
+  LogicalResult analyze(func::FuncOp func) {
+    func.walk([&](Operation *op) {
+      if (op->hasTrait<cudaq::QuantumMeasure>()) {
+        hasMeasurement = true;
+        return WalkResult::interrupt();
```
schweitzpgi, line 37: Don't you want to collect all the `quake::AllocaOp`s? This will terminate the walk possibly before all the AllocaOp have been visited.

schweitzpgi: Consider moving this code to the constructor on the Analysis.

```c++
explicit Analysis(func::FuncOp func) {
  ...
}
```

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -11,6 +11,22 @@
...
+    measurements.
+
+    // TODO: Determine the appropriate workflow to invoke this pass. It should 
+    be run in the sampling context, so that we comply with the spec - If no 
+    measurements appear in the kernel, then the `__global__` register is formed
+    with implicit measurements being added for all the qubits defined in the 
+    kernel, and the measurements all occur at the end of the kernel.
+
+  }];
+  let dependentDialects = ["cudaq::cc::CCDialect", "quake::QuakeDialect"];
```
schweitzpgi: We add ops from the `cf` dialect. Might want to add that.

lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  void runOnOperation() override {
+    func::FuncOp func = getOperation();
+    if (!func || func.empty())
+      return;
+
+    if (!func->hasAttr(cudaq::entryPointAttrName))
+      return;
+
+    /// NOTE: Having a conditional on a measurement indicates that a measurement
```
schweitzpgi, line 116: Do you want these `NOTE` comments to be part of the doxygen documentation?

lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,149 @@
...
+        return;
+    }
+
+    // Check if the function has any measurement operations, if yes, we don't do
+    // anything. If not, then check if the function has any qubit allocations,
+    // if yes, then we want to add measurements to it.
+    /// NOTE: Having an explicit measurement does not guarantee that all the
+    /// allocated qubits are measured.
+    Analysis analysis;
+    if (failed(analysis.analyze(func))) {
```
schweitzpgi: nit: It looks like this will never fail.

include/cudaq/Optimizer/Transforms/Passes.td
schweitzpgi: I didn't see any ops from `cudaq::cc`. So that could be removed? Arguably, `quake` must already be present since we're looking for `quake::AllocaOp` and Quake measurements.

test/Quake/add_measurements-1.qke
```diff
@@ -0,0 +1,102 @@
...
+
+// RUN: cudaq-opt --add-measurements %s | FileCheck %s
+
+module attributes {quake.mangled_name_map = {__nvqpp__mlirgen__caller = "__nvqpp__mlirgen__caller_PyKernelEntryPointRewrite"}} {
+  func.func @__nvqpp__mlirgen__subkernel() attributes {"cudaq-entrypoint", "cudaq-kernel"} {
+    %0 = quake.alloca !quake.ref
+    quake.x %0 : (!quake.ref) -> ()
+    %measOut = quake.mz %0 : (!quake.ref) -> !quake.measure
+    return
+  }
```
schweitzpgi, line 17: Probably should have a test with multiple `return` Ops. `return %5 : T` would be a good idea as well.

schweitzpgi (commented):
NIcely done. Thanks, Pradnya.

include/cudaq/Optimizer/Transforms/Passes.td
khalatepradnya: Fixed in 20d7785bd622902e4cdca3287a9043b6b3ab26ba
I was getting `LLVM ERROR: can't create type 'cudaq::cc::StdvecType' because storage uniquer isn't initialized: the dialect was likely not loaded, or the type wasn't added with addTypes<...>() in the Dialect::initialize() method.` error before adding `cudaq::cc::CCDialect`. I removed it and no error anymore.

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: Fixed in 20d7785bd622902e4cdca3287a9043b6b3ab26ba

test/Quake/add_measurements-1.qke
khalatepradnya: Added in 20d7785bd622902e4cdca3287a9043b6b3ab26ba

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: Fixed in 20d7785bd622902e4cdca3287a9043b6b3ab26ba.
I am skipping any processing if a measurement is found, so no, not collecting allocas in that case.

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: Removed in 20d7785bd622902e4cdca3287a9043b6b3ab26ba

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: Not sure. I will revisit after internal discussions .

lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!func->hasAttr(cudaq::entryPointAttrName))
+      return;
+
+    /// NOTE: Having a conditional on a measurement indicates that a measurement
+    /// is present, however, it does not guarantee that all the allocated qubits
+    /// are measured.
+    if (auto boolAttr = func->getAttr("qubitMeasurementFeedback")
+                            .dyn_cast_or_null<mlir::BoolAttr>()) {
+      if (boolAttr.getValue())
```
schweitzpgi, line 121: Does this need to be a Boolean attribute? It seems like if the attribute is there, that might be enough of an indication.

lib/Optimizer/Transforms/AddMeasurements.cpp
```diff
@@ -0,0 +1,139 @@
...
+  ArrayRef<Type> returnTypes = funcOp.getFunctionType().getResults();
+  if (!returnTypes.empty()) {
+    SmallVector<Location> argLocs(returnTypes.size(), loc);
+    newBlock->addArguments(returnTypes, argLocs);
+  }
+
+  // Replace every func.return in the function with a branch to the new block.
+  SmallVector<func::ReturnOp> returnsToReplace;
+  funcOp.walk(
+      [&](func::ReturnOp returnOp) { returnsToReplace.push_back(returnOp); });
```
schweitzpgi: You could collect the ReturnOp ops in the analysis as well as the AllocaOp. Save having to rescan the entire function.

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: I want to account for presence of `qubitMeasurementFeedback = false`.

lib/Optimizer/Transforms/AddMeasurements.cpp
khalatepradnya: Fixed in [c38f308](https://github.com/NVIDIA/cuda-quantum/pull/3392/commits/c38f308f24c280b784a5df9bdf61e122fb924706)

schweitzpgi (approved):
Thanks for working on this new pass. It's going to be great.

PR #3393, opened 2025-08-27
update default values for rest endpoint

qci-opensource:
@khalatepradnya please review this change to the default values on the rest endpoint


khalatepradnya:
> /ok to test d43cc45

PR #3396, opened 2025-08-27
[NFC] cleanup some idioms when using isa<T>

schweitzpgi:


atgeller:
LGTM, thanks @schweitzpgi! I think similar changes may be needed for #3306 before it gets merged...

atgeller (approved):

PR #3397, opened 2025-08-27
[CI] Experimenting with test invocation

khalatepradnya:


khalatepradnya:
The `targettests` time out on x86, gcc11/12.

PR #3398, opened 2025-08-28
Phase Folding Optimization for Value Semantics

atgeller:
### Description

A version of #3306 that works on the value semantics instead of the reference semantics.

Some todos:
- Definitions are split across files, it would be better to combine them
- We ultimately decided against outlining subcircuits into functions, so that will need to be disabled and tests updated.


atgeller:
Subsumed by #5021.

PR #3401, opened 2025-08-28
[testing] [Python] Update test fixture

khalatepradnya:
This PR improves the reliability and setup of the `test_mqpu.py` test suite by enhancing the test fixture for proper MQPU environment preparation and cleanup.

Tested locally, with this fix, the `test_mqpu.py` succeeded when run as part of whole test suite 3 times in a row.

### DIFF
--- a/python/tests/parallel/test_mqpu.py
+++ b/python/tests/parallel/test_mqpu.py
@@ -6,9 +6,15 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-import cudaq, os, pytest, random, timeit
-from cudaq import spin
+import os
+import random
+import time
+import timeit
+
+import cudaq
 import numpy as np
+import pytest
+from cudaq import spin
 
 skipIfNoMQPU = pytest.mark.skipif(
     not (cudaq.num_available_gpus() > 0 and cudaq.has_target('nvidia-mqpu')),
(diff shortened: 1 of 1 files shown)


sacpis:
@khalatepradnya I think this is a fix for https://nvbugspro.nvidia.com/bug/5415866 as well, right?

khalatepradnya:
> @khalatepradnya I think this is a fix for https://nvbugspro.nvidia.com/bug/5415866 as well, right?

Yes, mostly. I ran the test suite a 4th time and this time `test_mqpu.py` as skipped, but it didn't cause unexpected failure.

sacpis:
Hi @khalatepradnya, I don't think that we should call reset_target at the start of do_something. Seems like we need to make sure that any test which ran before this one called reset_target in its do_something method.

khalatepradnya:
> Hi @khalatepradnya, I don't think that we should call reset_target at the start of do_something. Seems like we need to make sure that any test which ran before this one called reset_target in its do_something method.

I don't know how we can guarantee that, unless all tests are updated - which will be a good idea. My attempt here is that this test (a) shouldn't fail when we run the whole suite, instead be skipped if the set up has failed (b) give the set up more time since this is run with debug build and known to be sluggish.

mitchdz:
Honestly I'm fine with this implementation. I like to write all my tests with the Arrange-Act-Assert methodology, and this implementation forces the `Arrange` portion on all the tests.

sacpis:
> > Hi @khalatepradnya, I don't think that we should call reset_target at the start of do_something. Seems like we need to make sure that any test which ran before this one called reset_target in its do_something method.
> 
> I don't know how we can guarantee that, unless all tests are updated - which will be a good idea. My attempt here is that this test (a) shouldn't fail when we run the whole suite, instead be skipped if the set up has failed (b) give the set up more time since this is run with debug build and known to be sluggish.

Can we make sure the reset_target is called in the previous test and does its job of resetting the target? We can just check if any target is present in do_something() (on the first run).

khalatepradnya:
> Let me know if I can try to play with this one.

Go ahead!

PR #3410, opened 2025-09-02
fixes issue 3409

mmvandieren:

### DIFF
--- a/docs/sphinx/examples/python/executing_kernels.py
+++ b/docs/sphinx/examples/python/executing_kernels.py
@@ -199,7 +199,11 @@ Angle 1.4: 5/10 ones measured
  [End `RunAsyncOutput`] '''
 
 # [Begin Observe]
+import cudaq
+
 from cudaq import spin
+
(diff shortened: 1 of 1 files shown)


sacpis:
Thanks @mmvandieren. `qubit_count` is defined on line 2 of the file. If we were to make every block executable on its own, we will have to make sure that every block has cudaq import and necessary variables initialized.

mmvandieren:
OK.  wasn't sure what the standard was.

sacpis:
Seems like @mawolf2023 would know.

sacpis:
Seems like @mawolf2023 is OOO. In this case, I think @bettinaheim could make a call.

schweitzpgi:
@sacpis what's the status of this one?

sacpis:
> @sacpis what's the status of this one?

@schweitzpgi Seems like we will need to check with the product as whether to include multiple imports in a single file for every cell. Will wait for @efratshabtai.

PR #3381, opened 2025-08-21
Issue 3041: s/aggressive-early-inlining/aggressive-inlining/

vdFLunN5FWuz:
### Description

Remove the adjective "early" from inlining pass/pipeline #3041

schweitzpgi - 2025-06-11 - The use of "early" is a bit of historical legacy and it no longer really makes sense and could actually create confusion. It is recommended that the name be shortened to "aggressive-inlining".

Rename methods, types, build scripts, config, and files containing types. Update *.rst documentation.

ctest before
99% tests passed, 2 tests failed out of 1294
The following tests did not run:
        1222 - iqm-tests (Skipped)
The following tests FAILED:
          5 - ctest-nvqpp (Failed)
          6 - ctest-targettests (Failed)

ctest after
99% tests passed, 2 tests failed out of 1294
The following tests did not run:
        1222 - iqm-tests (Skipped)
The following tests FAILED:
          5 - ctest-nvqpp (Failed)
          6 - ctest-targettests (Failed)

PR #3411, opened 2025-09-02
[fix] Update example - reset qubit after measurement

khalatepradnya:
Follow-up to similar change in PR https://github.com/NVIDIA/cuda-quantum/pull/3332
The change improves the correctness of the quantum walk by resetting the `aux` qubit after each iteration.


1tnguyen (approved):
👍

schweitzpgi:
Let's see if we can get #3381 to merge ...

PR #3413, opened 2025-09-03
[NFC] Change the target tests to make the CI faster

schweitzpgi:
Get rid of "target" tests run without a target and make file structure coherent.

Convert all tests that are run dozens of times from symbolic links to use include and remove the llvm-lit RUN lines. Now the tests will RUN once, but still be available for the other parts of CI that will cherry-pick the tests by name to run in special ways.


targettests/anyon/bug_qubit.cpp
bmhowe23, line 1: The root cause of the problem is that the `local_excludes` variable in `targettests/lit.cfg.py` was not kept up to date as new targets were added. The following trivial change (just a few lines for the whole PR) would remove all the duplicate tests AFAICT. Of course, if `#include`ing .cpp files is still preferred, that's functional, too.

https://gist.github.com/bmhowe23/74db56461a12d79eb426b11ba8705cbd

schweitzpgi: Yes, but as you say the lit config wasn't being maintained properly and neither were those directories themselves. ☹️ 

Making it inescapably obvious by using #include here, hopefully, makes it clearer to people what to do ... assuming it works across CI. 🤞

bmhowe23: In order to reduce future confusion, we should remove the target-specific folders from `local_excludes` then if we don't expect anybody to maintain that list (doesn't have to be for this PR).

bmhowe23 (commented):
Just curious - what was the failure? It looked correct to me when I ran `llvm-lit -v --show-tests --param nvqpp_site_config=build/targettests/lit.site.cfg.py build/targettests`. Does `--show-tests` not work correctly?

schweitzpgi:
> > Commit title: Include Ben's patch, but that approach has proven to be a failure
> 
> Just curious - what was the failure? It looked correct to me when I ran `llvm-lit -v --show-tests --param nvqpp_site_config=build/targettests/lit.site.cfg.py build/targettests`. Does `--show-tests` not work correctly?

I mean process failure. No one maintains the config file. No one understands the special nature of the directory structure (to wit, one of the directories included in your patch was not symbolic links to anything). These changes are about _process_. It codifies that files are being used in different ways that casual symbolic links do not convey. We are looking for a new approach entirely here.

PR #3416, opened 2025-09-03
update to use CPU runners with more memory

mitchdz:
I believe the CI tests might be running out of memory on the smaller nodes. This is checking the test speed with more memory.


sacpis:
Just curious, what is 8 vs 8m? how much memory each has?

mitchdz:
@sacpis https://github.com/nv-gha-runners/enterprise-runner-configuration/blob/main/docs/runner-groups.md
It is upgrading from 32GB to 64GB memory

sacpis:
> @sacpis https://github.com/nv-gha-runners/enterprise-runner-configuration/blob/main/docs/runner-groups.md It is upgrading from 32GB to 64GB memory

Nice. Thanks @mitchdz.

mitchdz:
we do not currently have permission to use these runners.

PR #3417, opened 2025-09-03
[NFC] Tidy of local_excludes since they are not being used

bmhowe23:


targettests/lit.cfg.py
```diff
@@ -24,10 +24,7 @@
 
 # Exclude a list of directories from the test suite:
 #   - 'Inputs' contain auxiliary inputs for various tests.
-local_excludes = [
-    'anyon', 'braket', 'ionq', 'iqm', 'oqc', 'qci', 'quantinuum',
-    'quantum_machines', 'fermioniq', 'infleqtion', 'Inputs'
-]
+local_excludes = ['Inputs']
```
schweitzpgi, line 27: There is some potential savings to leaving these as the directories won't be run at all (not even the `echo` commands).

On the other hand, some people are sticking entirely new tests in these directories, apparently because it's not clear they are intended only for the CI scripts (hence one failure when I consolidated the tests).

bmhowe23: > There is some potential savings to leaving these as the directories won't be run at all (not even the `echo` commands).
> 
> On the other hand, some people are sticking entirely new tests in these directories, apparently because it's not clear they are intended only for the CI scripts (hence one failure when I consolidated the tests).

In my opinion, it just causes maintenance confusion to leave them here, leaving them in a half right, half wrong state. Either the process is broken and we should eliminate it, or our code reviews should do a better job enforcing the process. Based on the last PR, I gathered it was best to just eliminate it.

PR #3418, opened 2025-09-04
Update BraketServerHelper to use ccnot instead of ccx

jacofeld:
Checklist:
- [ ] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.

### Description
I added a mapping from `ccx`->`ccnot` since that's the name of the gate on Braket in the exact same manner `cx` is already mapped to `cnot`.

This problem was mentioned on this previous issue:
https://github.com/NVIDIA/cuda-quantum/issues/2325#issuecomment-2518222251

### DIFF
--- a/runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
+++ b/runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
@@ -14,6 +14,8 @@ std::string prepareOpenQasm(std::string source) {
   source = std::regex_replace(source, includeRE, "");
   const std::regex cxToCnot{"\\scx\\s"};
   source = std::regex_replace(source, cxToCnot, " cnot ");
+  const std::regex ccxToCcnot{"\\sccx\\s"};
+  source = std::regex_replace(source, ccxToCcnot, " ccnot ");
   return source;
 }
 
--- a/python/tests/backends/test_braket.py
+++ b/python/tests/backends/test_braket.py
@@ -498,6 +498,7 @@ def test_exp_pauli():
 
 
 def test_toffoli():
+
     @cudaq.kernel
     def kernel():
         q = cudaq.qvector(3)
--- a/python/tests/backends/test_braket.py
+++ b/python/tests/backends/test_braket.py
@@ -497,6 +497,20 @@ def test_exp_pauli():
     assert not '10' in counts
 
 
+def test_toffoli():
+    @cudaq.kernel
+    def kernel():
+        q = cudaq.qvector(3)
+        x(q)
+        x.ctrl([q[0], q[1]], q[2])
+        mz(q)
+
+    counts = cudaq.sample(kernel)
+    counts.dump()
+    assert '110' in counts
+    assert len(counts) == 1
+
+
 # leave for gdb debugging
 if __name__ == "__main__":
     loc = os.path.abspath(__file__)


runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
```diff
@@ -14,6 +14,8 @@ std::string prepareOpenQasm(std::string source) {
   source = std::regex_replace(source, includeRE, "");
   const std::regex cxToCnot{"\\scx\\s"};
   source = std::regex_replace(source, cxToCnot, " cnot ");
+  const std::regex ccxToCcnot{"\\sccx\\s"};
+  source = std::regex_replace(source, ccxToCcnot, " ccnot ");
```
khalatepradnya, line 18: This change looks good. Thanks for the update.

How about we add a simple test like -
```
diff --git a/python/tests/backends/test_braket.py b/python/tests/backends/test_braket.py
index 6b69adc029..892d7183ff 100644
--- a/python/tests/backends/test_braket.py
+++ b/python/tests/backends/test_braket.py
@@ -497,6 +497,21 @@ def test_exp_pauli():
     assert not '10' in counts


+def test_toffoli():
+
+    @cudaq.kernel
+    def kernel():
+        q = cudaq.qvector(3)
+        x(q)
+        x.ctrl([q[0], q[1]], q[2])
+        mz(q)
+
+    counts.dump()
+    assert '110' in counts
+    assert len(counts) == 1
+
+
 # leave for gdb debugging
 if __name__ == "__main__":
     loc = os.path.abspath(__file__)
```

I tested `test_braket.py` file locally, `30 passed`.

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
```diff
@@ -14,6 +14,8 @@ std::string prepareOpenQasm(std::string source) {
...
   return source;
```
khalatepradnya, line 19: Note that I tried the `sudoku` example as well, it compiles and runs but gives incorrect result.

Before the changes in this PR we see the following failure:
```
# bin/nvq++ --target braket ../targettests/execution/sudoku_2x2.cpp && ./a.out
  what():  [line 13] uses a gate: ccx which is not supported by the device or defined via a defcal
Aborted (core dumped)
```
With the changes in this PR, we get - 
```
# bin/nvq++ --target braket ../targettests/execution/sudoku_2x2.cpp && ./a.out
1101
0101
a.out: sudoku_2x2.cpp:70: int main(): Assertion `most_probable.count("1001") == 1' failed.
Aborted (core dumped)
```
Since the simple Toffoli gate test passed, I think this problem is orthogonal, but it still needs fixing (can be a follow-up).

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
jacofeld: Thanks for the suggestion, I copied this test in!

khalatepradnya:
Missing DCO on last commit. Here's how to fix: https://github.com/NVIDIA/cuda-quantum/pull/3418/checks?check_run_id=49638355822

runtime/cudaq/platform/default/rest/helpers/braket/BraketServerHelper.cpp
khalatepradnya: Tracking this in issue [#2325](https://github.com/NVIDIA/cuda-quantum/issues/2325#issuecomment-3255878245)

khalatepradnya:
> /ok to test 03c4434

PR #3419, opened 2025-09-04
[Python] Add utility to register callbacks for Python target changes 

1tnguyen:
### Description

Add functions to register/unregister callbacks to be executed when the runtime target changes.

Upon registration, we also invoke the callback with the current target.

### DIFF
--- a/python/runtime/cudaq/target/py_runtime_target.cpp
+++ b/python/runtime/cudaq/target/py_runtime_target.cpp
@@ -11,8 +11,36 @@
 #include "common/FmtCore.h"
 #include "common/Logger.h"
 #include "cudaq/platform.h"
+#include <functional>
+#include <pybind11/functional.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
+
+namespace {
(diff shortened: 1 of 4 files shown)


python/runtime/cudaq/target/py_runtime_target.cpp
```diff
@@ -11,9 +11,37 @@
...
+#include <pybind11/functional.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+namespace {
+using SetTargetCallbackFnTy = std::function<void(const cudaq::RuntimeTarget &)>;
+// List of `set_target` callbacks to be executed during target changes.
+// We use a string identifier for so that the subscriber can identify itself,
+// e.g., to unsubscribe/replace the callback.
+static std::unordered_map<std::string, SetTargetCallbackFnTy> g_callbacks;
```
sacpis, line 25: Do we need to lock this map?

1tnguyen: Good point. Fixed in f020bb457f81eb147ce7d0900d744a05f0719556

PR #3420, opened 2025-09-04
[DRAFT] Update dev images to Ubuntu 24.04

bmhowe23:
Update the dev image to use Ubuntu 24.04. By implication, this also advances CUDA and Python in those images to 12.6 and 3.12, respectively.

Correspondingly, it drops Python 3.10 from the wheel builds.


sacpis:
Is it like we will be supporting min CUDA version as 12.6? Just FYI, Jetson has CUDA 12.4 on it.

bmhowe23:
> Is it like we will be supporting min CUDA version as 12.6? Just FYI, Jetson has CUDA 12.4 on it.

This is really just for our Docker images, which have to use one particular version of every software package that we support. Our Python wheels and installer will support running alongside many other versions (same as before).

This is just a draft for now. We may decide to go in a different direction eventually.

mitchdz:
Closing since implemented and included in my PR.

PR #3422, opened 2025-09-05
[sample] Automatic measurement on remote targets

khalatepradnya:
This PR updates measurement handling in sample execution contexts for remote backend targets, improving compliance with the `sample` API specification and backend consistency.

* Modified `BaseRemoteRESTQPU.h` to invoke the `add-measurements` pass only for sample execution contexts.
* Added a test in `test_braket.py` to demonstrate invoking `sample` and `observe` on same kernel.
```
 // Before changes in this PR
  $ python3 test_with_braket.py 
   -1.8324588900000012
   RuntimeError: At least one of the used qubits in the program need to be measured
 // After changes in this PR
  $ python3 test_with_braket.py 
  -1.7036210699999992
  { 01:83 10:917 }
```
* Removed explicit calls to `mz` (measurement) in the sample kernels for the `quantinuum`, `braket`, and `infleqtion` targets, allowing implicit measurement insertion by the framework. Verified manually against integration servers.


1tnguyen (approved):
👍

PR #3423, opened 2025-09-08
Part 1: Restructuring of compiler pipelines.

schweitzpgi:
There are a variety of compiler pass pipelines that exist in the project. It has long been acknowledged that they don't meet our needs very well.

For the C++ AOT compiler, there are two sets of pipelines primarily. There is the pipeline that is built in the compiler driver and then a pipeline that is used to generate code for default simulation.

For the C++ JIT compiler, every target can (and did) specify its own special pipeline, leading to inconsistencies and maintenance problems. Furthermore, this pipeline may have been incorrect and patched on-the-fly by the JIT compilation process. Finally, it was augmented by running the same pipeline as the AOT compiler for code generation with various minor changes. This situation was both far too flexible (anyone could run any pass in any order in the JIT pipeline) and far too inflexible (high-level and low-level passes were thrown together in ad hoc ways, preventing downstream passes from running correctly).

For Python, the pipelines are done in a completely different way for the most part, but some of the pipelines are the same. For example, the code generation pipeline remains largely the same, but has its own minor tweaks. Python doesn't really distinguish between AOT and JIT compilation.

The changes:

  - Instead of a single "lowering pipeline" that was used by the JIT compiler, the JIT pipeline has been decomposed into target agnostic and target dependent passes. Furthermore, the JIT pipeline has been broken into 3 levels: high level, mid level, and low level. Each of these levels runs the target dependent passes (from the .yml target config) followed by the target independent passes for that level.

  - The code generation pipeline now has distinct AOT vs. JIT separation. This eliminates the running of passes at JIT time that are not required.


targettests/TargetConfig/RegressionValidation/anyon.config
```diff
@@ -17,10 +17,11 @@
 # Add the rest-qpu library to the link list
 # CHECK-DAG: LINKLIBS="${LINKLIBS} -lcudaq-rest-qpu"
 
-# Define the lowering pipeline. telegraph-8q has an 8-qubit ring topology, so mapping
+# Define the JIT lowering pipeline. telegraph-8q has an 8-qubit ring topology, so mapping
```
khalatepradnya: Orthogonal to this PR, but I wonder if we need to keep maintaining these `*.config` regression tests anymore. The original purpose, if I am not mistaken, was to ensure nothing broke when we transitioned to YAML files. May be there's more to it, @1tnguyen ?

runtime/cudaq/platform/default/rest/helpers/ionq/ionq.yml
```diff
@@ -17,8 +17,8 @@ config:
   preprocessor-defines: ["-D CUDAQ_QUANTUM_DEVICE"]
   # Add the rest-qpu library to the link list
   link-libs: ["-lcudaq-rest-qpu"]
-  # Define the lowering pipeline
-  platform-lowering-config: "func.func(erase-noise),classical-optimization-pipeline,globalize-array-values,func.func(canonicalize,state-prep),unitary-synthesis,func.func(canonicalize),apply-op-specialization{constant-prop=1},aggressive-inlining,classical-optimization-pipeline,decomposition{enable-patterns=U3ToRotations},func.func(canonicalize,multicontrol-decomposition),ionq-gate-set-mapping,distributed-device-call"
+  # Define the JIT lowering pipeline
+  jit-mid-level-pipeline: "ionq-gate-set-mapping"
   # Tell the rest-qpu that we are generating QIR.
```
khalatepradnya, line 24: Q: What 'level' is `post-codegen-passes: "remove-measurements"`?
(see line#26)

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -2040,42 +2041,47 @@ struct QuakeToQIRAPIPass
...
           processOperation<
               BaseProfileQIR</*opaquePtr=*/true, QirVersion::version_0_2>>(
               typeConverter);
-        processOperation<
-            BaseProfileQIR</*opaquePtr=*/false, QirVersion::version_0_2>>(
-            typeConverter);
+        else
+          processOperation<
+              BaseProfileQIR</*opaquePtr=*/false, QirVersion::version_0_2>>(
+              typeConverter);
```
1tnguyen, line 2055: I wonder whether we should shorten this `if`-`else` to:

 ```
 processOperation<
              BaseProfileQIR</*opaquePtr=*/opaquePtr, QirVersion::version_0_2>>(
              typeConverter);
 ```

targettests/TargetConfig/RegressionValidation/anyon.config
1tnguyen: That's a good point. The original intention of these tests is to catch any regression when transitioning from the `.config` to `.yml` file format.
It has been a couple of releases since then, so I think it's time for us to remove these regression tests.

schweitzpgi: I've been wondering the same thing actually. Is it worth opening an Issue?

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
schweitzpgi: I don't think you can pass the `bool` variable as a template argument like that.

runtime/cudaq/platform/default/rest/helpers/ionq/ionq.yml
schweitzpgi: Let me make this an issue. It would be good to do it before QIR codegen, I think. But there may be a reason not to.

targettests/TargetConfig/RegressionValidation/anyon.config
1tnguyen: Done. Posted https://github.com/NVIDIA/cuda-quantum/issues/3425

khalatepradnya: I will create a PR to remove these.

1tnguyen: @schweitzpgi has posted https://github.com/NVIDIA/cuda-quantum/issues/3426 :) 
I closed mine as duplicate!

PR #3429, opened 2025-09-10
[Testing][NFC] Using the conventional mock server port number series to improve stability

1tnguyen:
### Description

This payload provider mock server failed intermittently due to port availability.

Other mock servers often use `6244x` port numbers. Hence, changed the port number to that series (up to 62449 has been used, thus 62450)


schweitzpgi (approved):
Did you create an issue for using a delayed backoff and/or random port scheme?

khalatepradnya (approved):
Yes, please! 
I occasionally see the failure `[Errno 98] error while attempting to bind on address ('0.0.0.0', 56686): address already in use`  on `extra-payload-provider-tests` and I think this change might mitigate it.

PR #3431, opened 2025-09-10
use -j$(nproc) for ctest

mitchdz:
note that llvm-lit is not explicitly added as well since it by default uses all available cpus[0].

[0] - https://llvm.org/docs/CommandGuide/lit.html#cmdoption-lit-j

### DIFF
--- a/.github/workflows/test_in_devenv.yml
+++ b/.github/workflows/test_in_devenv.yml
@@ -98,7 +98,7 @@ jobs:
               echo "::error file=test_in_devenv.yml::Python interop tests failed with status $pytest_status."
               exit 1
             fi
-            ctest --output-on-failure --test-dir build -E "ctest-nvqpp|ctest-targettests"
+            ctest -j$(nproc) --output-on-failure --test-dir build -E "ctest-nvqpp|ctest-targettests"
             ctest_status=$?
             $LLVM_INSTALL_PREFIX/bin/llvm-lit -v --time-tests --param nvqpp_site_config=build/test/lit.site.cfg.py build/test
             lit_status=$?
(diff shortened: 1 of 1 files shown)


.github/workflows/test_in_devenv.yml
```diff
@@ -98,7 +98,7 @@ jobs:
               echo "::error file=test_in_devenv.yml::Python interop tests failed with status $pytest_status."
               exit 1
             fi
-            ctest --output-on-failure --test-dir build -E "ctest-nvqpp|ctest-targettests"
+            ctest -j$(nproc) --output-on-failure --test-dir build -E "ctest-nvqpp|ctest-targettests"
```
schweitzpgi, line 101: Amazing idea! 😄

mitchdz: I had some inspiration 🙂

mitchdz:
> Hope this speeds things up without causing thrashing or OOM errors.

That's my concern. I wonder if this was never done for reasons of tests not setting down properly, but we should see soon.

bmhowe23:
> > Hope this speeds things up without causing thrashing or OOM errors.
> 
> That's my concern. I wonder if this was never done for reasons of tests not setting down properly, but we should see soon.

I know that it causes issues on my machine (despite having 128GB memory), but that will be machine specific. It is also not very reliable to do this on a machine with a GPU because at least some of our tests expect to have access to the GPU's full memory, so the random run-to-run variation of timing will cause problems in that situation. (Still worth testing in the CI, though.)

mitchdz:
These tests are CPU only tests, so we should be good with GPU interference. Memory is a valid concern though, especially on shared runners.

mitchdz:
Looks like it doesn't really improve the time that much 🙁

mitchdz:
Closing since no improvements seen.

PR #3432, opened 2025-09-10
[Python] [tests] Clear kernel registries in test set up

khalatepradnya:
WIP: Issue#1954


python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -6,18 +6,20 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
+import os
+from typing import List
+
```
schweitzpgi, line 11: Q: Is there a reason for blank lines between import statements in python?

(We do that in C++ often enough, which is actually a problem, since we don't want to hide bugs in the header files by trying to force particular orderings and faking out clang-format.)

khalatepradnya: It is PEP8 guideline - https://peps.python.org/pep-0008/#imports
And tools such as `isort` , `autopep8` can be used to automatically apply.

python/tests/backends/test_Infleqtion.py
```diff
@@ -17,10 +20,10 @@
 
 
 @pytest.fixture(scope="session", autouse=True)
-def do_something():
+def setup_test_environment():
```
sacpis, line 23: Thanks @khalatepradnya for renaming this one :)

python/tests/backends/test_Infleqtion.py
```diff
@@ -17,10 +20,10 @@
...
+    cudaq.__clearKernelRegistries()
     cudaq.set_target("infleqtion")
     yield "Running the tests."
-    cudaq.__clearKernelRegistries()
```
sacpis, line 23: Don't we need to call `__clearKernelRegistries` after the last test?

python/tests/domains/test_qnn.py
```diff
@@ -7,16 +7,21 @@
...
 
 skipIfNoMQPU = pytest.mark.skipif(
     not (cudaq.num_available_gpus() > 0 and cudaq.has_target('nvidia-mqpu')),
     reason="nvidia-mqpu backend not available")
 
 
+@pytest.fixture(autouse=True)
+def setup_test_environment():
+    cudaq.__clearKernelRegistries()
+    yield
```
sacpis, line 22: Same here after the yield, we will need to clean up the kernel registries, right?

python/tests/domains/test_uccsd.py
```diff
@@ -0,0 +1,161 @@
...
+                               molecule,
+                               optimizer,
+                               parameter_count=num_parameters)
+    print(energy, params)
+    assert np.isclose(-1.137, energy, rtol=1e-3)
+
+
+def test_uccsd_kernel():
+
+    from cudaq.kernels import uccsd
```
sacpis, line 104: We can make this common import as it is being used in other tests as well?

python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
schweitzpgi: My question doesn't seem to be answered at that link?

I'm asking why
```python
import foo

import bar
```
instead of
```python
import foo
import bar
```

Maybe the answer is python programmers just like double tapping the enter key?

sacpis: Yes, I don't think we need to have a blank lines between imports. I think it is individual programmers style.
One reason I can think of having a blank line is to group the imports (Standard, third-party, local).

khalatepradnya: It says, 
```
Imports should be grouped in the following order:

- Standard library imports.
- Related third party imports.
- Local application/library specific imports.

You should put a blank line between each group of imports.
```

python/tests/backends/test_Infleqtion.py
khalatepradnya: We can, but, I don't think we need to since we clear it before invoking individual tests.

python/tests/domains/test_qnn.py
khalatepradnya: No, unless I'm missing why?

python/tests/domains/test_uccsd.py
khalatepradnya: I left it here because I want the set up to override which `uccsd` module to import. If running stand-alone then yes, it makes sense, but if running as a suite, we may need to make a copy of the source file and import than and use it. Hence keeping this here. Would you rather have it moved to the set up?

python/tests/backends/test_Infleqtion.py
sacpis: I mean the last test in the suite.

python/tests/domains/test_uccsd.py
sacpis: Most prob to the global import as in setup we would have to import on every test call, right? Which seems unnecessary, unless I am missing anything here.

khalatepradnya: yes, but there are different modules being loaded depending on whether registry was cleared. Hence, this is single common place to load (but, per test).

khalatepradnya:
Closing this as it is out of sync. Consider follow-up with smaller parts.

PR #3436, opened 2025-09-15
Update to cuquantum 25.09 

1tnguyen:
### Description

- Update cuquantum dependency to 25.09.

- Fix deprecation warnings in tensornet backend (compiler errors, as we have warnings as errors as the default).

- Adjust build configuration as `libcustatevec.so` now has linkage to `libnvidia-ml.so.1` (provided by NVIDIA GPU drivers). In a build environment without a GPU, this file is not available. This causes issues for (1) building the test executables and (2) gtest discovery (`gtest_discover_tests`), which invokes the executable to list all the test cases. Thus, use CMake to inject the file from the stub library during build and remove it afterward.

### Pipeline

Deployment: https://github.com/NVIDIA/cuda-quantum/actions/runs/18182323457


mitchdz:
Thanks for this Thien. It's unfortunate with the `libnvidia-ml.so.1` linkage which makes things a little gross, hopefully we can remove that with a future custatevec update.

mitchdz (approved):

bmhowe23:
I'll run this branch through the CUDA-QX CI to see what breaks.

https://github.com/NVIDIA/cudaqx/pull/316

https://github.com/NVIDIA/cudaqx/actions/runs/18316833547?pr=316

bmhowe23:
> I'll run this branch through the CUDA-QX CI to see what breaks.
> 
> [NVIDIA/cudaqx#316](https://github.com/NVIDIA/cudaqx/pull/316)
> 
> https://github.com/NVIDIA/cudaqx/actions/runs/18316833547?pr=316

Surprisingly, it looks like nothing broke, despite parts of CUDA-QX calling out 25.06. In any case, looks like there is no need to wait for CUDA-QX on anything before merging this.

PR #3440, opened 2025-09-15
devdeps.ext.Dockerfile: pin 25.06

mitchdz:
now that cuquantum 25.09 is out, deployments is failing, since that is being pulled automatically.


1tnguyen (approved):
Thanks @mitchdz.

PR #3443, opened 2025-09-16
Bug fix to allow select on qubits

bettinaheim:
... and other random fixes.

### DIFF
--- a/lib/Support/Config/TargetConfig.cpp
+++ b/lib/Support/Config/TargetConfig.cpp
@@ -385,7 +385,7 @@ void MappingTraits<cudaq::config::TargetArchitectureSettings>::mapping(
 
 void MappingTraits<cudaq::config::MachineArchitectureConfig>::mapping(
     IO &io, cudaq::config::MachineArchitectureConfig &info) {
-  io.mapRequired("arch-name", info.Name);
+  io.mapOptional("arch-name", info.Name);
   io.mapOptional("machine-names", info.MachineNames);
   io.mapOptional("pattern", info.MachinePattern);
   io.mapRequired("config", info.Configuration);
--- a/runtime/cudaq/platform/default/rest/helpers/qci/QCIServerHelper.cpp
+++ b/runtime/cudaq/platform/default/rest/helpers/qci/QCIServerHelper.cpp
@@ -44,7 +44,11 @@ private:
       "https://aqumen-service-dev.quantumcircuitsinc.net/";
 
   /// @brief Default machine, the simulator.
+  /// TODO: Update this to `AquSim`
   const std::string DEFAULT_MACHINE = "simulator";
+
+  /// @brief Default action to perform
+  const std::string DEFAULT_METHOD = "simulate";
 
   /// @brief Polling interval for job status via QCI's CUDA-Q endpoint in
   /// microseconds.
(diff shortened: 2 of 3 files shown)


bettinaheim:
> /ok to test b81e92b

1tnguyen (approved):
LGTM 👍

bettinaheim:
> /ok to test da14089

lib/Optimizer/Transforms/CombineQuantumAlloc.cpp
```diff
@@ -62,11 +62,12 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
           Value hi = rewriter.create<arith::ConstantIntOp>(
               alloc.getLoc(), os.first + os.second - 1, rewriter.getI64Type());
+          // trying to print alloc after the replace gives a segfault
+          LLVM_DEBUG(llvm::dbgs() << "replace " << alloc);
           [[maybe_unused]] Value subveq =
               rewriter.replaceOpWithNewOp<quake::SubVeqOp>(
                   alloc, alloc.getType(), analysis.newAlloc, lo, hi);
-          LLVM_DEBUG(llvm::dbgs()
-                     << "replace " << alloc << " with " << subveq << '\n');
+          LLVM_DEBUG(llvm::dbgs() << " with " << subveq << '\n');
```
schweitzpgi, line 70: I don't think we need this change to the debug output. Since this can be multithreaded, splitting the output is more confusing potentially.

test/NVQPP/qir_gen.cpp
```diff
@@ -0,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2025 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// RUN: nvq++ %cpp_std --emit-qir %s && cat qir_gen.qir.ll | \
```
schweitzpgi, line 9: The structure of this test is very broad and less focused on the exact issue. That is, something in a completely unrelated area of the compiler, or even runtime headers, might change this such that the focus of the test vanishes.

Can we instead capture the Quake code that has the arith.select operation and run `cudaq-translate --convert-to=qir %s -o %t` on that? That will guarantee that other changes don't introduce false positive results.

test/NVQPP/struct_arg.cpp
```diff
@@ -7,7 +7,7 @@
  ******************************************************************************/
 
 // RUN: nvq++ %cpp_std --emit-qir %s && cat struct_arg.qir.ll | \
-// RUN: FileCheck %s ; rm struct_arg.qir.ll
+// RUN: FileCheck %s && rm struct_arg.qir.ll
```
schweitzpgi, line 10: Don't we always want to erase the intermediate file? Although we shouldn't return the return code of `rm`. heh

lib/Optimizer/Transforms/CombineQuantumAlloc.cpp
bettinaheim: See the comment I added. Do you have other suggestions for how to avoid the segfault?

test/NVQPP/qir_gen.cpp
bettinaheim: I can update the test to just look for the relevant select lines.

test/NVQPP/struct_arg.cpp
bettinaheim: I thought if the test failed I might as well have the .ll file handy. But either way is fine. But yes, the return code of rm is what made the test always pass...

bettinaheim:
> /ok to test e6d1584

bettinaheim:
> /ok to test 5f19834

test/NVQPP/qir_gen.cpp
schweitzpgi: Yeah, that's good. I was asking about making the test take a .qke file as input as well. If you want, you can open this comment as a new issue for later. It's a good first issue. :)

lib/Optimizer/Transforms/CombineQuantumAlloc.cpp
schweitzpgi: Ah. No, I missed the comment, but it makes sense. heh

test/NVQPP/qir_gen.cpp
bettinaheim: Good point, I added a quake test. I kept the one here - not a bad thing to check if that kind of structure simplifies properly.

bettinaheim:
> /ok to test 80eb5a2

bettinaheim:
> /ok to test a2aa075

PR #3445, opened 2025-09-16
[qci] Make the code gen flexible per machine type

khalatepradnya:
This PR updates the target configuration for `qci` backends. This enables granular control over code generation settings per machine and adds support for a new `method` argument.

* Updated the `machine` argument type in `qci.yml` to `machine-config`, allowing for machine-specific configuration blocks (e.g., for `AquSim` and `Seeker`), and added a new optional `method` argument to specify circuit execution type (`simulate` or `execute`).
* Updated the mapping logic in `TargetConfig.cpp` to make the `"arch-name"` field optional instead of required, allowing for more flexible architecture configuration.
* Updated job creation in the `QCIServerHelper` to include the selected `method` in the job payload, ensuring the backend receives the correct execution type for each job. Also override the machine to `simulator` if `AquSim` is specified (as a temporary measure till service is updated).

### DIFF
--- a/runtime/cudaq/platform/default/rest/helpers/qci/QCIServerHelper.cpp
+++ b/runtime/cudaq/platform/default/rest/helpers/qci/QCIServerHelper.cpp
@@ -47,11 +47,8 @@ private:
   /// TODO: Update this to `AquSim`
   const std::string DEFAULT_MACHINE = "simulator";
 
-  /// @brief Allowed methods
-  const std::vector<std::string> ALLOWED_METHODS = {"simulate", "execute"};
-
   /// @brief Default action to perform
-  const std::string DEFAULT_METHOD = ALLOWED_METHODS[0];
+  const std::string DEFAULT_METHOD = "simulate";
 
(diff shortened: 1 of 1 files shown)


lib/Support/Config/TargetConfig.cpp
```diff
@@ -375,7 +375,7 @@ void MappingTraits<cudaq::config::TargetArchitectureSettings>::mapping(
 
 void MappingTraits<cudaq::config::MachineArchitectureConfig>::mapping(
     IO &io, cudaq::config::MachineArchitectureConfig &info) {
-  io.mapRequired("arch-name", info.Name);
+  io.mapOptional("arch-name", info.Name);
```
schweitzpgi, line 388: What happens if no `arch-name` is specified?

khalatepradnya: This change makes it an optional field, so the changes in `qci.yml` file, with no `arch-name` work.

1tnguyen (approved):
LGTM 👍

PR #3446, opened 2025-09-17
Follow up to #3423

schweitzpgi:
Found the mystery crash was caused by ListOption pass options not being correctly initialized by default.

Changes remove the deprecated platform-lowering-pipeline (yay!) and add a helper to avoid and document the issue.

### DIFF
--- a/lib/Optimizer/Transforms/Pipelines.cpp
+++ b/lib/Optimizer/Transforms/Pipelines.cpp
@@ -103,6 +103,8 @@ static void registerEmulationTargetPrepPipeline() {
 void cudaq::opt::addDecompositionPass(OpPassManager &pm,
                                       ArrayRef<std::string> enabledPats,
                                       ArrayRef<std::string> disabledPats) {
+  // NB: Both of these ListOption *must* be set here or they may contain garbage
+  // and the compiler may crash.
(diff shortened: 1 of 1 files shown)


schweitzpgi:
#3423

lib/Optimizer/Transforms/Pipelines.cpp
```diff
@@ -46,7 +51,7 @@ static void createTargetPrepPipeline(OpPassManager &pm,
   // pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddDeallocs());
   // pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddMetadata());
   // pm.addNestedPass<func::FuncOp>(cudaq::opt::createUnwindLowering());
-  pm.addPass(cudaq::opt::createQuakePropagateMetadata());
+  // pm.addPass(cudaq::opt::createQuakePropagateMetadata());
```
khalatepradnya, line 54: Can we remove these?

schweitzpgi: I'm going to add them back in short order ... assuming I can ... or, yeah, they're goners.

PR #3448, opened 2025-09-19
Support for QIR output on QCI

khalatepradnya:
* Parse QIR output

### DIFF
--- a/runtime/common/Future.cpp
+++ b/runtime/common/Future.cpp
@@ -46,7 +46,7 @@ sample_result future::get() {
           dynamic_cast<QirServerHelper *>(serverHelper.get());
       if (!qirServerHelper)
         throw std::runtime_error("To support `run` API, " + qpuName +
-                                 "must inherit `QirServerHelper` class");
+                                 " must inherit `QirServerHelper` class");
       if (!inFutureRawOutput)
         throw std::runtime_error(
             "cudaq::details::future::get() for 'run' requires a raw output "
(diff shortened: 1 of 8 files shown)


utils/mock_qpu/qci/__init__.py
```diff
@@ -64,9 +65,25 @@ def getKernelFunction(module):
...
-        if "requiredQubits" in str(a):
+        if "required_num_qubits" in str(a):
             return int(
-                str(a).split("requiredQubits\"=")[-1].split(" ")[0].replace(
+                str(a).split(f'required_num_qubits\"=')[-1].split(" ")
+                [0].replace("\"", "").replace("'", ""))
+        elif "requiredQubits" in str(a):
+            return int(
+                str(a).split(f'requiredQubits\"=')[-1].split(" ")[0].replace(
+                    "\"", "").replace("'", ""))
```
sacpis: Wondering if we can have a simple function here to return the number of qubits as the logic in all functions seems similar. We can defined this common function something like this

```
def get_qubits(txt, var):
    return str(txt).split(var)[-1].split(" ")
                [0].replace("\"", "").replace("'", ""))
```

and then we would just need to call this function.

khalatepradnya: This is temporary code to support both QIR v0.1 and v0.2, will be removed to support only QIR v1.0.

sacpis: Sorry, did not mean to comment on this one. Was just referring.

khalatepradnya: No worries!

khalatepradnya: See 57c9e35

khalatepradnya:
Temporarily hold off merging till server-side changes are enabled.

PR #3449, opened 2025-09-19
MergeModules utility function

sacpis:
This fixes bug in `checkRegisteredCppDeviceKernel` which dropped everything on the floor which is not `func::FuncOp`.

* Adding mergeModules utility function
* Avoid using walk as it is wrong. Filtering artifacts by their IR node type is incorrect.
* Adding unit tests

co-authored by: Eric Schweitz <eschweitz@nvidia.com>

### DIFF
--- a/include/cudaq/Optimizer/Builder/Factory.h
+++ b/include/cudaq/Optimizer/Builder/Factory.h
@@ -294,19 +294,7 @@ std::pair<mlir::func::FuncOp, /*alreadyDefined=*/bool>
 getOrAddFunc(mlir::Location loc, mlir::StringRef funcName,
              mlir::FunctionType funcTy, mlir::ModuleOp module);
 
-inline void mergeModules(mlir::ModuleOp into, mlir::ModuleOp from) {
-  for (mlir::Operation &op : *from.getBody()) {
-    auto sym = llvm::dyn_cast<mlir::SymbolOpInterface>(&op);
-    if (!sym)
-      continue; // Only merge named symbols, avoids duplicating anonymous ops.
-
-    // If `into` already has a symbol with this name, skip it.
-    if (mlir::SymbolTable::lookupSymbolIn(into, sym.getName()))
-      continue;
-
-    into.push_back(op.clone());
-  }
-}
+void mergeModules(mlir::ModuleOp into, mlir::ModuleOp from);
 } // namespace factory
 
 std::size_t getDataSize(llvm::DataLayout &dataLayout, mlir::Type ty);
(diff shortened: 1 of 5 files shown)


include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -294,6 +294,19 @@ std::pair<mlir::func::FuncOp, /*alreadyDefined=*/bool>
 getOrAddFunc(mlir::Location loc, mlir::StringRef funcName,
              mlir::FunctionType funcTy, mlir::ModuleOp module);
 
+inline void mergeModules(mlir::ModuleOp into, mlir::ModuleOp from) {
```
schweitzpgi: I don't think this is performance critical. We can move the definition to Factory.cpp.

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -294,6 +294,19 @@ std::pair<mlir::func::FuncOp, /*alreadyDefined=*/bool>
...
+  for (mlir::Operation &op : *from.getBody()) {
```
schweitzpgi:
```suggestion
  for (Operation *op : from.getBody()) {
```
Moving the code to the Factory.cpp file has the bonus that you don't have to add the namespace everywhere.

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -294,6 +294,19 @@ std::pair<mlir::func::FuncOp, /*alreadyDefined=*/bool>
...
+    auto sym = llvm::dyn_cast<mlir::SymbolOpInterface>(&op);
```
schweitzpgi:
```suggestion
    auto sym = dyn_cast<mlir::SymbolOpInterface>(op);
```

include/cudaq/Optimizer/Builder/Factory.h
sacpis: Fixed in 408aacb.

include/cudaq/Optimizer/Builder/Factory.h
sacpis: Fixed in 408aacb.

include/cudaq/Optimizer/Builder/Factory.h
sacpis: If we go with the suggested change, then we get this build error

```
/workspaces/cuda-quantum/lib/Optimizer/Builder/Factory.cpp: In function ‘void cudaq::opt::factory::mergeModules(mlir::ModuleOp, mlir::ModuleOp)’:
/workspaces/cuda-quantum/lib/Optimizer/Builder/Factory.cpp:778:37: error: ‘begin’ was not declared in this scope
  778 |   for (Operation *op : from.getBody()) {
```

PR #3450, opened 2025-09-19
[WIP

1tnguyen:

### DIFF
--- a/include/cudaq/Optimizer/Transforms/Passes.h
+++ b/include/cudaq/Optimizer/Transforms/Passes.h
@@ -56,6 +56,8 @@ std::unique_ptr<mlir::Pass>
 createArgumentSynthesisPass(mlir::ArrayRef<mlir::StringRef> funcNames,
                             mlir::ArrayRef<mlir::StringRef> substitutions);
 
+std::unique_ptr<mlir::Pass> createQubitResetBeforeReuse();
+
 // declarative passes
 #define GEN_PASS_DECL
 #define GEN_PASS_REGISTRATION
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -1360,4 +1360,16 @@ def WriteAfterWriteElimination : Pass<"write-after-write-elimination"> {
   }];
 }
 
+def QubitResetBeforeReuse : Pass<"qubit-reset-before-reuse", "mlir::func::FuncOp"> {
+  let summary = "Add qubit reset and conditional initialization after measurement if qubit is to be reused.";
+  let description = [{
+   This pass adds qubit reset and conditionally applies an X gate if the measurement result is 1 
+   to initialize qubit into the correct state after measurement. This is only activated when the measured qubit
+   is to be reused.
+   Note: if the measurement is already accompanied by a reset, we won't add any extra reset.
+  }];
+
+  let constructor = "cudaq::opt::createQubitResetBeforeReuse()";
+}
+
 #endif // CUDAQ_OPT_OPTIMIZER_TRANSFORMS_PASSES
--- a/lib/Optimizer/Transforms/CMakeLists.txt
+++ b/lib/Optimizer/Transforms/CMakeLists.txt
@@ -70,6 +70,7 @@ add_cudaq_library(OptTransforms
   VariableCoalesce.cpp
   WiresToWiresets.cpp
   WriteAfterWriteElimination.cpp
+  ResetBeforeReuse.cpp
 
   DEPENDS
     OptTransformsPassIncGen
--- a/lib/Optimizer/Transforms/ResetBeforeReuse.cpp
+++ b/lib/Optimizer/Transforms/ResetBeforeReuse.cpp
@@ -83,7 +83,6 @@ public:
                        << "\n";
 
           const auto orderedUsers = sortUsers(reg.getUsers(), dom);
-
           // Find the next op on the register
           assert(orderedUsers.size() > 0);
           bool foundThisExtract = false;
@@ -118,8 +117,6 @@ public:
           assert(foundThisExtract);
         }
       }
-      llvm::outs() << "Hey: " << qubit.getType() << "\n";
-
       return nullptr;
     };
     OpBuilder builder(funcOp);
--- a/lib/Optimizer/Transforms/ResetBeforeReuse.cpp
+++ b/lib/Optimizer/Transforms/ResetBeforeReuse.cpp
@@ -25,6 +25,27 @@ using namespace mlir;
 
 namespace {
 
+static SmallVector<Operation *, 8> sortUsers(const Value::user_range &users,
+                                             const DominanceInfo &dom) {
+  SmallVector<Operation *, 8> orderedUsers;
+
+  for (auto *user : users) {
+    if ([&]() {
+          for (auto iter = orderedUsers.begin(), iterEnd = orderedUsers.end();
+               iter != iterEnd; ++iter) {
+            assert(*iter);
+            if (dom.dominates(user, *iter)) {
+              orderedUsers.insert(iter, user);
+              return false;
+            }
+          }
+          return true;
+        }())
+      orderedUsers.push_back(user);
+  }
+  return orderedUsers;
+}
+
 class QubitResetBeforeReusePass
     : public cudaq::opt::QubitResetBeforeReuseBase<QubitResetBeforeReusePass> {
 public:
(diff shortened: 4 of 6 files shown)


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -1360,4 +1360,16 @@ def WriteAfterWriteElimination : Pass<"write-after-write-elimination"> {
...
+def QubitResetBeforeReuse : Pass<"qubit-reset-before-reuse", "mlir::func::FuncOp"> {
+  let summary = "Add qubit reset and conditional initialization after measurement if qubit is to be reused.";
+  let description = [{
+   This pass adds qubit reset and conditionally applies an X gate if the measurement result is 1 
+   to initialize qubit into the correct state after measurement. This is only activated when the measured qubit
+   is to be reused.
+   Note: if the measurement is already accompanied by a reset, we won't add any extra reset.
+  }];
+
+  let constructor = "cudaq::opt::createQubitResetBeforeReuse()";
```
schweitzpgi:
```suggestion
```
Let's use the new automatic code generation facilities here, since this is a new pass.  See StackFramePrealloc.cpp for an example of what to do.

include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -85,6 +85,8 @@ std::unique_ptr<mlir::Pass>
 createArgumentSynthesisPass(mlir::ArrayRef<mlir::StringRef> funcNames,
                             mlir::ArrayRef<mlir::StringRef> substitutions);
 
+std::unique_ptr<mlir::Pass> createQubitResetBeforeReuse();
```
schweitzpgi:
```suggestion
```

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+class QubitResetBeforeReusePass
+    : public cudaq::opt::QubitResetBeforeReuseBase<QubitResetBeforeReusePass> {
+public:
+  QubitResetBeforeReusePass() = default;
+
+  void runOnOperation() override {
+    func::FuncOp funcOp = getOperation();
+    if (!funcOp || funcOp.empty())
```
schweitzpgi:
```suggestion
    if (funcOp.empty())
```
The operation has to exist to get here, so we don't need to re-check whether it exists.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,174 @@
...
+              foundThisExtract = true;
+            }
+          }
+          assert(foundThisExtract);
+        }
+      }
+      return nullptr;
+    };
+    OpBuilder builder(funcOp);
+    funcOp->walk([&](quake::MzOp mz) {
```
schweitzpgi: Let's use a rewriter pattern.

1tnguyen: Thanks! Fixed as suggested.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,242 @@
...
+using namespace mlir;
+
+namespace {
+
+static SmallVector<Operation *, 8> sortUsers(const Value::user_range &users,
+                                             const DominanceInfo &dom) {
+  SmallVector<Operation *, 8> orderedUsers;
+
+  for (auto *user : users) {
+    if ([&]() {
```
sacpis, line 40: Wondering why do we need this `if` loop.

schweitzpgi (commented):
This pass should be written when the IR is in explicit linear use form, as it will be much simpler to understand, write, etc.

It actually becomes trivial. A measurement op must produce 1 or more wires. If those wires are used by anything that is not a sink operation, they are reused. In those case, if the reuse is not a reset operation, then insert a reset operation. Done.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,242 @@
...
+                  LLVM_DEBUG(llvm::dbgs()
+                             << "Next use: " << *extractedQubitOrderedUsers[0]
+                             << "\n");
+                  return extractedQubitOrderedUsers[0];
+                }
+              }
+            }
+            if (v.value() == extractOp) {
+              foundThisExtract = true;
+            }
```
sacpis: Instead of using foundThisExtract flag, we can start iterating with index 1 and just check if index 0's value is extractOp, right?

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,242 @@
...
+                      sortUsers(extractedQubit.getUsers(), dom);
+                  assert(!extractedQubitOrderedUsers.empty());
+                  LLVM_DEBUG(llvm::dbgs()
+                             << "Next use: " << *extractedQubitOrderedUsers[0]
+                             << "\n");
+                  return extractedQubitOrderedUsers[0];
+                }
+              }
+            }
+            if (v.value() == extractOp) {
```
schweitzpgi: nit: no braces

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,242 @@
...
+              foundThisExtract = true;
+            }
+          }
+          assert(foundThisExtract);
```
schweitzpgi: pass should not assert (unless this is for debugging, in which case it should be deleted before final commit).

A pass should always run. The correct failure mode is to not change the IR.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -0,0 +1,242 @@
...
+      for (auto v : llvm::enumerate(orderedUsers)) {
+        if (v.value() == op && v.index() < (orderedUsers.size() - 1) &&
+            dom.dominates(op, orderedUsers[v.index() + 1])) {
+          return orderedUsers[v.index() + 1];
+        }
+      }
+    }
+
+    // No next use is found, check if this is an extracted qubit.
+    if (isa<quake::RefType>(qubit.getType())) {
```
schweitzpgi, line 162: In reference semantics, you will have to account for every possible case of aliasing. References can alias.

So it's not as simple as looking at alloca and extract_ref. Correctness here means you have to consider calls, block arguments, various arithmetic operations, dynamic qubit refreences, and so on.

On the other hand, value semantics has already been built and makes all of this much, much easier.

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
```diff
@@ -20,6 +20,7 @@ config:
   # Define the JIT lowering pipeline
   jit-high-level-pipeline: "expand-measurements"
   jit-mid-level-pipeline: "quantinuum-gate-set-mapping"
+  jit-low-level-pipeline: "func.func(qubit-reset-before-reuse)"
```
schweitzpgi: Is this meant to be a Quantinuum only pass?

targettests/execution/estimate_resources_break.cpp
```diff
@@ -6,8 +6,9 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
```
schweitzpgi, line 6: nit: why a dash in the name of this test?

test/AST-Quake/reset_after_measure.cpp
```diff
@@ -0,0 +1,123 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// REQUIRES: c++20
+// RUN: cudaq-quake %s | cudaq-opt --qubit-reset-before-reuse | FileCheck %s
+
+#include <cudaq.h>
+// CHECK:   func.func @__nvqpp__mlirgen__function_no_reuse
```
schweitzpgi: nit: can we group the CHECK lines in blocks after the code instead of shuffled into the code. It makes fixing the tests after changes much, much easier.

test/AST-Quake/reset_after_measure_reg.cpp
```diff
@@ -0,0 +1,99 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// REQUIRES: c++20
```
schweitzpgi: Does this code not compile under c++17?

test/AST-Quake/reset_after_measure_reg.cpp
```diff
@@ -0,0 +1,99 @@
...
+// clang-format off
+// RUN: cudaq-quake %s | cudaq-opt --indirect-to-direct-calls --inline --expand-measurements --memtoreg=quantum=0 --canonicalize --cc-loop-normalize --cc-loop-unroll --qubit-reset-before-reuse | FileCheck %s
```
schweitzpgi: This is such a long list of passes, etc. But I don't think we're actually interested in testing them all? (I may be wrong there.) Anyway, instead of introducing all these passes as variables to the test itself, we can just run them once up front, save the .qke file and then use the .qke file as the basis for the test along with the pass we're testing. This will eliminate chances that changes to unrelated parts of the core will break your tests.

lib/Optimizer/Transforms/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ add_cudaq_library(OptTransforms
   VariableCoalesce.cpp
   WiresToWiresets.cpp
   WriteAfterWriteElimination.cpp
+  ResetBeforeReuse.cpp
```
khalatepradnya: nit: I think these are listed in alphabetical order.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
1tnguyen: It's just a shorthand to use the lambda return value as a flag to denote that we have found a place to insert the new element, rather than pushing it to the back of the list.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
1tnguyen: Not really, the `extractOp` of interest can be anywhere in the ordered list of users of the `qvec` register.
The list of users is the entire list from the allocation to the deallocation, as we are in the reference semantics.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
1tnguyen: Good point. I've removed the debug asserts in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/f5d4ef8ba5dd46c8937484f1d3eb6415961508f6

runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
1tnguyen: Yea, currently, it's the one that has this requirement.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
1tnguyen: Removed some unnecessary braces in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/f5d4ef8ba5dd46c8937484f1d3eb6415961508f6

targettests/execution/estimate_resources_break.cpp
1tnguyen: Good point. I've renamed them in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/21f07dec7c5b2e859008035db90b427f252eab5d

lib/Optimizer/Transforms/CMakeLists.txt
1tnguyen: Good catch. I didn't notice that :) 

Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/25ebb59f4d9972d2411e471d3e092899de5dc581

test/AST-Quake/reset_after_measure.cpp
1tnguyen: Good point. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/a4811a5f5d0552d2c0e39666bfc43a1a4c3357bd

test/AST-Quake/reset_after_measure_reg.cpp
1tnguyen: It should. Removed it in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/a4811a5f5d0552d2c0e39666bfc43a1a4c3357bd

test/AST-Quake/reset_after_measure_reg.cpp
1tnguyen: Thanks for the tip. Refactored it into a `.qke` test in https://github.com/NVIDIA/cuda-quantum/pull/3450/commits/a4811a5f5d0552d2c0e39666bfc43a1a4c3357bd

test/AST-Quake/reset_after_measure.cpp
```diff
@@ -0,0 +1,126 @@
...
+// RUN: cudaq-quake %s | cudaq-opt --qubit-reset-before-reuse | FileCheck %s
+
+#include <cudaq.h>
+void no_reuse() __qpu__ {
+  cudaq::qubit q;
+  h(q);
+  mz(q);
+}
+
+// CHECK:   func.func @__nvqpp__mlirgen__function_no_reuse
```
schweitzpgi: Let's use the generate-test-checks.py script (tpls/llvm/mlir/utils) to generate more comprehensive CHECK lines.

1tnguyen: Definitely, fixed in 21211db9ab76682f28a752ded16648d0dcdc9440

1tnguyen:
> This pass should be written when the IR is in explicit linear use form (aka, value semantics), as it will be much simpler to understand, write, etc.
> 
> It actually becomes trivial. A measurement op must produce 1 or more wires. If those wires are used by anything that is not a sink operation, they are reused. In those case, if the reuse is not a reset operation, then insert a reset operation. Done.

I posted https://github.com/NVIDIA/cuda-quantum/issues/3565 to implement this pass in the linear SSA form.
In the linear form, we would also need to find a good solution for potential qubit wire aliasing, e.g., from qubit vector slicing. 

Hence, it might be better to do it as a follow-on task.

schweitzpgi (approved):
Giving temporary approval. This ought to be re-done using the wire types.

PR #3451, opened 2025-09-19
Ability to send on the `QuantinuumConfig.simulator` field

1tnguyen:
### Description

This `simulator` field is an allowed field in the `QuantinuumConfig` object [schema](https://nexus.quantinuum.com/api-docs).
Hence, this PR enables users to set that in the payload.


runtime/cudaq/platform/default/rest/helpers/quantinuum/quantinuum.yml
```diff
@@ -82,3 +82,8 @@ target-arguments:
     type: string
     platform-arg: extra_payload_provider
     help-string: "Specify the extra payload provider if any."
+  - key: simulator
+    required: false
+    type: string
+    platform-arg: simulator
+    help-string: "If machine is a simulator, the type of simulator to use."
```
khalatepradnya, line 89: Q: Should this spell out what the allowed values are?
(Okay if clarified in docs elsewhere)

khalatepradnya: Never mind, found it in the Nexus API docs.

PR #3452, opened 2025-09-19
[Python] Make sample API spec compliant

khalatepradnya:
[CUDA-Q spec](https://nvidia.github.io/cuda-quantum/latest/specification/cudaq/algorithmic_primitives.html#cudaq-sample) for `sample` says -

>  kernels passed to this function must be entry-point kernels and return void. 

This pull request introduces stricter validation to the Python `cudaq.sample` and `cudaq.sample_async` APIs, ensuring they are only used with kernels that have a void return type. 

Note: marked this as a breaking change because prior versions incorrectly / unintentionally supported sampling on non-void kernels.


python/cudaq/runtime/sample.py
```diff
@@ -71,6 +71,11 @@ def sample(kernel,
     has_conditionals_on_measure_result = False
 
     if isinstance(kernel, PyKernelDecorator):
+        if kernel.returnType is not None:
+            raise RuntimeError(
+                f"The `sample` API only supports kernels with void return type. "
```
1tnguyen: I wonder whether we should replace the word 'void' with something more Pythonic.

1tnguyen (approved):
LGTM 👍

python/cudaq/runtime/sample.py
khalatepradnya: How about returns `None`?

khalatepradnya: See f624bc0

python/tests/kernel/test_kernel_call_return.py
```diff
@@ -41,8 +42,10 @@ def caller() -> bool:
     result = caller()
     assert result == True or result == False
 
-    counts = cudaq.sample(caller)
-    assert '1' in counts and len(counts) == 1
+    with pytest.raises(RuntimeError) as error:
+        cudaq.sample(caller)
+    assert "The `sample` API only supports kernels that return None (void)" in repr(
+        error)
```
schweitzpgi:
```suggestion
    assert ("The `sample` API only supports kernels that return None (void)" in repr(
        error))
```

python/tests/kernel/test_run_kernel.py
```diff
@@ -992,12 +992,14 @@ def bell_pair() -> int:
...
-    sample_results = cudaq.sample(bell_pair, shots_count=10)
-    assert len(sample_results) == 2
-
     direct_call_result = bell_pair()
     assert direct_call_result is not None
 
+    with pytest.raises(RuntimeError) as error:
+        cudaq.sample_async(bell_pair, shots_count=10)
+    assert "The `sample_async` API only supports kernels that return None (void)" in repr(
+        error)
```
schweitzpgi:
```suggestion
    assert ("The `sample_async` API only supports kernels that return None (void)" in repr(
        error))
```

PR #3454, opened 2025-09-22
install compiler headers for external plugin development

amccaskey:
Add Quake and CC headers (including tablegen headers) to the install target. This is required to enable downstream development of compiler passes.


schweitzpgi:
Looks like the commit needs to be signed.

PR #3462, opened 2025-09-24
Fixes publishing Python metapackage validation

sacpis:
Fixes publishing Python metapackage validation

https://github.com/NVIDIA/cuda-quantum/actions/runs/17964651031/job/51098367212#step:7:8884

Co-authored by: @1tnguyen.


mitchdz:
We should probably look for a missing RPATH somewhere, and if we decide to go down this route, it should be annotated in the pypi docs so users can follow it without much frustration.

PR #3463, opened 2025-09-24
[parser] Ignore unsuccessful shot in output log

khalatepradnya:
This PR updates the handling of shot records in the `RecordLogParser` to ensure that only successful shots are processed and ignores any failed shots (instead of throwing error). 
Also removes unnecessary methods, consolidates shot processing logic, and adds unit tests to verify correct behavior for both successful and failed shots.


runtime/common/RecordLogParser.cpp
```diff
@@ -16,29 +16,47 @@ void cudaq::RecordLogParser::parse(const std::string &outputLog) {
...
+    else if (recordType == "METADATA")
+      handleMetadata(entries);
+    else if (recordType == "START") {
+      processingShot = true;
+      shotData.clear();
+    } else if (recordType == "OUTPUT") {
+      if (processingShot)
+        shotData.push_back(entries);
+      else
+        handleOutput(entries);
```
1tnguyen, line 43: Q: Do we expect a situation where `processingShot` is `false` here? 
One case I can imagine is that there is no `START`. Perhaps we should throw an error instead?

1tnguyen (approved):
👍

runtime/common/RecordLogParser.cpp
khalatepradnya: Simulator output does not have `START`.

runtime/common/RecordLogParser.cpp
```diff
@@ -16,29 +16,47 @@ void cudaq::RecordLogParser::parse(const std::string &outputLog) {
...
+            handleOutput(shotLine);
+        }
+      } else {
+        CUDAQ_INFO("Discarding shot data due to non-zero END status.");
+      }
+      processingShot = false;
+      shotData.clear();
+      containerMeta.reset();
+    } else
+      throw std::runtime_error("Invalid record type: " + recordType);
```
schweitzpgi:
```suggestion
    } else {
      throw std::runtime_error("Invalid record type: " + recordType);
    }
```

runtime/common/RecordLogParser.cpp
```diff
@@ -16,29 +16,47 @@ void cudaq::RecordLogParser::parse(const std::string &outputLog) {
...
+    } else if (recordType == "OUTPUT") {
+      if (processingShot)
+        shotData.push_back(entries);
+      else
+        handleOutput(entries);
+    } else if (recordType == "END") {
+      if (entries.size() < 2)
+        throw std::runtime_error("Missing shot status");
+      if (entries[1] == "0") {
+        if (processingShot) {
```
schweitzpgi, line 48: Rather than build this state machine flag into the loop, can't we pull these loops apart? Instead of looping and copying everything to another set of set of strings, we could (1) loop over the initial output buffer to see if the END marker indicates this is a good shot, then (2) go back and rescan the OUTPUT markers and process them or (3) if not good go to the next shot (if any). I believe this is already a vector of strings, so saving the index into the vector from where you start the lookahead should be easy enough.

khalatepradnya: Right, we have all the data in the `lines`. I've updated the logic in c033c70 to capture just the start index. 
To separate out the loops isn't straightforward IMO because of the edge cases - like we might just get shots w/o `START` and `END`; or the first record has `METADATA` included; and the `RESULT` type is any array w/o the container metadata.

PR #3465, opened 2025-09-25
Fix bug in unitary mixture calculation when the matrix values are small due to low error rate

1tnguyen:
### Description

There is a bug in `isScaledUnitary` implementation whereby we escape early based on the element value compared to the tolerance. 

The tolerance value shouldn't be used there, as this will skip the case where the error rate is small.

### DIFF
--- a/include/cudaq/Support/TargetConfig.h
+++ b/include/cudaq/Support/TargetConfig.h
@@ -97,7 +97,6 @@ struct BackendEndConfigEntry {
   /// Enable/disable the library mode if provide.
   std::optional<bool> LibraryMode;
   /// IR lowering configuration (hardware REST QPU)
-  std::string PlatformLoweringConfig; // deprecated
   std::string JITHighLevelPipeline;
   std::string JITMidLevelPipeline;
   std::string JITLowLevelPipeline;


1tnguyen:
> /ok to test a06447794fcc9c90d28ae0039ac9c3cfc6867a45

unittests/common/NoiseModelTester.cpp
```diff
@@ -74,3 +74,23 @@ CUDAQ_TEST(NoiseModelTester, checkConstruction) {
...
+
+#if defined(CUDAQ_SIMULATION_SCALAR_FP64)
+CUDAQ_TEST(NoiseModelTester, checkUnitaryDetection) {
+  EXPECT_TRUE(amplitude_damping(0.5).unitary_ops.empty());
+  EXPECT_FALSE(depolarization1(0.1).unitary_ops.empty());
+  // Small probability depolarization
+  EXPECT_FALSE(depolarization1(1e-6).unitary_ops.empty());
+  EXPECT_EQ(depolarization1(1e-6).unitary_ops.size(), 4);
+  EXPECT_EQ(depolarization1(1e-6).probabilities.size(), 4);
+  EXPECT_EQ(depolarization1(1e-6).probabilities.size(), 4);
```
sacpis: Are these same?

1tnguyen: Good catch! Removed the duplicated check line.

1tnguyen:
> /ok to test 2ec57eb4e4691337277f9bab09aad4ac4889060b

PR #3466, opened 2025-09-26
[NFC] Formatting of doc strings and comments.

schweitzpgi:
Apparently yapf doesn't do a very good job with formatting these to the 80 column coding standard.


sacpis:
Thanks @schweitzpgi for catching this. Which tool have you used to format doc strings?

sacpis:
Seems like few docs strings still need to be made within the column length of 80 (like the docstring in `class PyASTBridge` etc). `docformatter` will help with that.

schweitzpgi:
> Seems like few docs strings still need to be made within the column length of 80 (like the docstring in `class PyASTBridge` etc). `docformatter` will help with that.

I'm just pulling out an assortment of these from my other changes for the moment. Might as well get them in separately rather than clutter up code changes.

PR #3467, opened 2025-09-26
[refactoring] Common code for QIR processing

khalatepradnya:
This PR refactors how sample results are created from QIR output logs in the `QuantinuumServerHelper`, moving logic from the implementation file into a reusable method in the base server helper class.

Attempt to break up large PR #3448 into smaller chunks.

### DIFF
--- a/runtime/common/ServerHelper.h
+++ b/runtime/common/ServerHelper.h
@@ -8,13 +8,13 @@
 
 #pragma once
 
-#include "nlohmann/json.hpp"
-
 #include "ExecutionContext.h"
 #include "Future.h"
 #include "Registry.h"
 #include "RuntimeTarget.h"
 #include "SampleResult.h"
+#include "common/RecordLogParser.h"
+#include "nlohmann/json.hpp"
 #include <filesystem>
 
 namespace cudaq {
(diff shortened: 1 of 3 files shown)


runtime/common/ServerHelper.h
```diff
@@ -168,5 +168,42 @@ class QirServerHelper {
...
+    char *buffer = static_cast<char *>(malloc(bufferSize));
+    std::memcpy(buffer, origBuffer, bufferSize);
+
+    std::vector<std::vector<bool>> results = {
+        reinterpret_cast<std::vector<bool> *>(buffer),
+        reinterpret_cast<std::vector<bool> *>(buffer + bufferSize)};
+    const auto numShots = results.size();
+    // Create the counts dictionary
+    cudaq::CountsDictionary globalCounts;
+    std::vector<std::string> globalSequentialData;
```
sacpis, line 191: Sorry I might have have missed something but wondering why do we need globalSequentialData here? Just checked the code in main and found out that it is assigned to `result.sequentialData`. But here as we don't have that, do we need to store globalSequentialData?

khalatepradnya: Good catch! That's used in a different type of output (non-QIR) processing. (And here it _may_ not be required.) 
I simply moved code, so before I proceed to delete, @1tnguyen can you comment in case I'm missing something?

1tnguyen: Good catch @sacpis 🥇 

This is an oversight. We should have assigned it here as well.

@khalatepradnya Could you please fix that in this PR as well? Thanks!

1tnguyen (approved):
Thanks @khalatepradnya!

runtime/common/ServerHelper.h
khalatepradnya: See 5b3e467

PR #3475, opened 2025-09-29
Expose runtime target information of quantum platform

1tnguyen:
### Description

This information was already populated for internal usage, e.g., to determine codegen spec, etc. 
Now, expose it for downstream code to query it if needed (e.g., for validation purposes).

### DIFF
--- a/docs/sphinx/api/languages/cpp_api.rst
+++ b/docs/sphinx/api/languages/cpp_api.rst
@@ -356,6 +356,8 @@ Platform
 
 .. doxygenstruct:: cudaq::CodeGenConfig
 
+.. doxygenstruct:: cudaq::RuntimeTarget
+
 Utilities
 =========
 


runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -280,6 +280,10 @@ cudaq::CodeGenConfig quantum_platform::get_codegen_config() {
   return config;
 }
 
+const RuntimeTarget *quantum_platform::get_runtime_target() const {
+  return runtimeTarget.get();
```
sacpis, line 284: Wondering what needs to happen if we call `get_runtime_target` before/without `set_target`?

1tnguyen: We always `set_target` in Python as part of the static initialization of the module during `LinkedLibraryHolder` constructor: https://github.com/NVIDIA/cuda-quantum/blob/25c051977eb1b3046527f1b325281b7d21ff7034/python/utils/LinkedLibraryHolder.cpp#L335

sacpis:
Should we add pybinding for `get_target` in `cudaq/python/runtime/cudaq/target/py_runtime_target.cpp`? So that it can be used in Python as well.

runtime/cudaq/platform/quantum_platform.cpp
schweitzpgi: Might want to add an `assert` in case this accessor is somehow reached in a context other than Python.

1tnguyen:
> Should we add pybinding for `get_target` in `cudaq/python/runtime/cudaq/target/py_runtime_target.cpp`? So that it can be used in Python as well.

In Python, the current `cudaq.get_target()` should provide target information (Python wrapper of the `RuntimeTarget` info) as well

PR #3478, opened 2025-09-30
Fix memory leak in `Array` allocation

1tnguyen:
### Description

In our generated QIR, we allocate `Array*`, e.g., with `__quantum__rt__array_create_1d` to pass a list of qubit references to a function call.
However, we don't have a mechanism to delete them.

This leads to memory leak. 

Some scenarios that it becomes problematic/noticeable (reported by @cuMikeBrown):

(1) Kernel that call another one a large number (could be exponential) of times. For example, implementing `U^(2^k)` via repeating application of `U` kernel.

(2) Many calls to kernel launch, e.g., when doing benchmarking (many repeats to compute the average).

The fix is to track and clean these `Array` when we finish the kernel launch.

**Note**: a more efficient clean-up strategy, e.g., tracking the lifetime of `Array` instances _within_ the kernel, would be deferred due to its complexity.


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -630,6 +630,11 @@ class GenerateKernelExecution
       return;
     }
 
+    // After kernel launch, we clean up any dangling arrays that are allocated
+    // during kernel execution.
+    builder.create<func::CallOp>(loc, std::nullopt,
+                                 cudaq::runtime::cleanupArrays, ValueRange{});
```
schweitzpgi: We'll need this in Python as well, since the leaking allocation is at the C layer underneath the interpreter.

runtime/nvqir/QIRTypes.cpp
```diff
@@ -19,6 +19,42 @@ extern bool initialized;
...
+    return instance;
+  }
+
+  // Track the given array
+  void track(Array *arr) { allocated_arrays.push_back(arr); }
+  // Untrack the given array (manually delete outside)
+  void untrack(Array *arr) {
+    auto it = std::find(allocated_arrays.begin(), allocated_arrays.end(), arr);
+    if (it != allocated_arrays.end())
+      allocated_arrays.erase(it);
```
schweitzpgi: Should the `else` case here raise an error? It's arguably a very broken program to deallocate arrays that aren't arrays or already dead. Also, it would be cheaper in terms of performance to replace the pointer with a `nullptr` than the resize the vector.

runtime/nvqir/QIRTypes.cpp
```diff
@@ -19,6 +19,42 @@ extern bool initialized;
 extern bool verbose;
 }
 
+namespace {
+// Singleton struct to track allocated arrays
+struct ArrayTracker {
+  // Get the singleton instance of ArrayTracker
+  static ArrayTracker &getInstance() {
+    static ArrayTracker instance;
```
schweitzpgi: I suppose we could use the C++ run once stuff to avoid the static initializer here. Kind of nitpicky though.

schweitzpgi (commented):
Thanks for working on this, Thien. Looking good.

lib/Optimizer/Transforms/GenKernelExecution.cpp
1tnguyen: Good point. I've thought that this will cover Python kernels, but apparently not :) 
I added this to `thunk` function, which is used by Python (see https://github.com/NVIDIA/cuda-quantum/pull/3478/commits/ed67a7a475b507c4c1ef667e0bd31e692a424936)

runtime/nvqir/QIRTypes.cpp
1tnguyen: I've updated the code to use `nullptr` instead of resizing. I updated the code to emit a warning if an unexpected pointer is provided here. As a clean-up utility, I'm a bit hesitant to make it a hard error.

runtime/nvqir/QIRTypes.cpp
1tnguyen: Yeah, I think I'll stick with Meyer’s singleton pattern here for its simplicity :)

lib/Optimizer/Transforms/GenKernelExecution.cpp
schweitzpgi: I don't see why it needs to be in both functions. Arguably putting it in the host side code (which is being done here) isn't really correct, since the device isn't the host.

runtime/nvqir/NVQIR.cpp
```diff
@@ -130,6 +130,14 @@ thread_local static std::vector<std::unique_ptr<Array>> allocatedArrays;
...
 
 /// @brief Utility function mapping qubit ids to a QIR Array pointer
+// Qubits are newly constructed in the returned array. Hence, the caller must
+// manage their lifetime.
+// For example, this can be used for allocating qubit arrays whereby the
+// corresponding deallocation (`__quantum__rt__qubit_release_array`) would clean
+// up the qubits.
+//
+// IMPORTANT: don't use this for borrowed qubit indices as that
+// will result in leak.
```
schweitzpgi, line 138: OK, this is interesting. The runtime is already caching the allocations in `allocatedArrays`. So we shouldn't need to do it a second time. In fact, that's a bad idea for maintenance reasons.
It looks like the cleanup routine should just clear out the `allocatedArrays` `vector`.

test/Translate/return_values.qke
```diff
@@ -475,7 +476,8 @@ func.func @test_5(%sret: !cc.ptr<!cc.struct<{i64, f64}>> {llvm.sret = !cc.struct
 
 // CHECK-LABEL: define { i8*, i64 } @test_2.thunk(i8* nocapture writeonly 
 // CHECK-SAME:                                                            %[[VAL_0:.*]], i1
-// CHECK-SAME:                                                            %[[VAL_1:.*]]) {{.*}} {
+// CHECK-SAME:                                                            %[[VAL_1:.*]]) {
+// CHECK:         tail call void @__nvqpp_cleanup_arrays()
```
schweitzpgi, line 480: Why is this done *before* the kernel runs? Weird.

schweitzpgi: Also why is the output different? This removes a regex that was apparently matching something previously. 😕

1tnguyen: It was having a `#k` (e.g., `#1`, `#2`) before. I'm not sure why it changed as we added a call.
Also, I noticed that we explicitly want a no-inline for the kernel call, but I don't see it either.

lib/Optimizer/Transforms/GenKernelExecution.cpp
1tnguyen: I see. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3478/commits/75e60570f1854a9483a2d2d87f1956d75071e243 (removed this function, just kept the one in `thunk`)

runtime/nvqir/NVQIR.cpp
1tnguyen: I merged this into the tracker (see https://github.com/NVIDIA/cuda-quantum/pull/3478/commits/75e60570f1854a9483a2d2d87f1956d75071e243) and removed a couple of unused functions.

schweitzpgi: Does the tracker need to be thread safe since this original cache of Arrays is?

runtime/nvqir/QIRTypes.cpp
```diff
@@ -19,6 +20,37 @@ extern bool initialized;
 extern bool verbose;
 }
 
+nvqir::ArrayTracker &nvqir::ArrayTracker::getInstance() {
+  static thread_local nvqir::ArrayTracker instance;
```
schweitzpgi, line 24: OK. You added it already. :)

runtime/nvqir/QIRTypes.cpp
```diff
@@ -73,30 +107,34 @@ std::vector<int64_t> getRangeValues(Array *in_array, const Range &in_range) {
   // Convert to absolute index.
   const auto start_idx = convertIndex(in_range.start);
   const auto end_idx = convertIndex(in_range.end);
-
+  Array *result = new Array(0, sizeof(Qubit *));
```
schweitzpgi, line 110: Do we really always want to allocate memory? It looks like the old algorithm returned a vector of null pointers originally.

runtime/nvqir/QIRTypes.cpp
```diff
@@ -19,6 +20,37 @@ extern bool initialized;
...
+// Untrack the given array (manually delete outside)
+void nvqir::ArrayTracker::untrack(Array *arr) {
+  auto it = std::find(allocated_arrays.begin(), allocated_arrays.end(), arr);
+  // Use nullptr to indicate untracked arrays (manually deleted outside) to
+  // prevent vector shrink.
+  if (it != allocated_arrays.end())
+    if (*it != nullptr)
+      *it = nullptr;
+    else
+      CUDAQ_WARN("Attempting to untrack an Array that is already untracked.");
```
schweitzpgi: I think you could just test that `arr` was `nullptr` when passed in to get this warning condition.

runtime/nvqir/QIRTypes.cpp
1tnguyen: Yea :)

runtime/nvqir/NVQIR.cpp
1tnguyen: Indeed. I made the array tracker `thread_local`.

runtime/nvqir/QIRTypes.cpp
1tnguyen: Actually, the original algorithm returned an empty vector of indices, which will turn into an `Array*` of zero size, like what we have here.

schweitzpgi: Yes. An "empty vector" is 3 pointers initialized to `nullptr`. :)

runtime/nvqir/QIRTypes.cpp
1tnguyen: Good point. I think I was confused when writing this code. Please take a look at the revision https://github.com/NVIDIA/cuda-quantum/pull/3478/commits/7d41dd809c76f716f17397aa26322045c11bfc74

PR #3488, opened 2025-10-02
Fix docs

mitchdz:
This was missed when we deprecated cuda 11. Build docs only for first index of cuda 12.6.


bmhowe23 (approved):

bmhowe23:
I'm just pushing this in since the CI already ran successfully here: https://github.com/NVIDIA/cuda-quantum/pull/3482

PR #3489, opened 2025-10-02
Improved list comprehension and other python fixes

bettinaheim:
More details will follow...

### DIFF
--- a/lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
+++ b/lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
@@ -1516,6 +1516,18 @@ struct AllocaOpPattern : public OpConversionPattern<cudaq::cc::AllocaOp> {
   }
 };
 
+struct ReturnOpPattern : public OpConversionPattern<func::ReturnOp> {
+  using Base = OpConversionPattern<func::ReturnOp>;
+  using Base::Base;
+
+  LogicalResult
+  matchAndRewrite(func::ReturnOp op, typename Base::OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<func::ReturnOp>(op, adaptor.getOperands());
+    return success();
+  }
+};
+
 /// Convert the quake types in `func::FuncOp` signatures.
 struct FuncSignaturePattern : public OpConversionPattern<func::FuncOp> {
   using OpConversionPattern::OpConversionPattern;
@@ -1738,7 +1750,7 @@ static void commonQuakeHandlingPatterns(RewritePatternSet &patterns,
                                         TypeConverter &typeConverter,
                                         MLIRContext *ctx) {
   patterns.insert<ApplyOpTrap, GetMemberOpRewrite, MakeStruqOpRewrite,
-                  RelaxSizeOpErase, VeqSizeOpRewrite>(typeConverter, ctx);
+                  ReturnOpPattern, RelaxSizeOpErase, VeqSizeOpRewrite>(typeConverter, ctx);
 }
 
 //===----------------------------------------------------------------------===//
@@ -1995,7 +2007,7 @@ struct QuakeToQIRAPIPass
           return true;
         });
     target.addDynamicallyLegalOp<
-        func::CallOp, func::CallIndirectOp, cudaq::cc::NoInlineCallOp,
+        func::CallOp, func::CallIndirectOp, func::ReturnOp, cudaq::cc::NoInlineCallOp,
         cudaq::cc::VarargCallOp, cudaq::cc::CallCallableOp,
         cudaq::cc::CallIndirectCallableOp, cudaq::cc::CastOp,
         cudaq::cc::FuncToPtrOp, cudaq::cc::StoreOp, cudaq::cc::LoadOp>(
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -3056,120 +3056,7 @@ class PyASTBridge(ast.NodeVisitor):
 
         if not isinstance(node.generators[0].target, ast.Name):
             self.emitFatalError(
-                "only support named targets in list comprehension", node)
-
-        def process_void_list():
-            # NOTE: This does not actually a create a valid value, and will fail is something
-            # tries to use the value that this was supposed to create later on. 
-            # Keeping this to keep existing functionality, but this is questionable imo.
-            # Aside from no list being produced, this should work regardless of what we
-            # iterate over or what the expression is we evaluate.
-            self.emitWarning("produced elements in list comprehension contain None - expression will be evaluated but no list is generated", node)
-            forNode = ast.For()
-            forNode.iter = node.generators[0].iter
-            forNode.target = node.generators[0].target
-            forNode.body = [node.elt]
-            forNode.orelse = []
-            self.visit_For(forNode)
-
-        # We need to know the element type of the list we are creating.
-        # Unfortunately, dynamic typing makes this a bit painful.
-        # We hence limit what kinds of expressions may be used.
-        def get_item_type(pyval):
-            if isinstance(pyval, ast.Name):
-                self.visit(pyval)
-                item = self.popValue()
-                return item.type
-            elif isinstance(pyval, ast.Constant):
-                return mlirTypeFromPyType(type(pyval.value), self.ctx)
-            elif isinstance(pyval, ast.List):
-                # FIXME: CHECK FOR EMPTY LIST (same in visit_List)
-                elts = [get_item_type(v) for v in pyval.elts]
-                if None in elts: return None
-                base_elTy = elts[0]
-                isHomogeneous = False not in [base_elTy == t for t in elts]
-                if not isHomogeneous:
-                    isArithmetic = False not in [self.isArithmeticType(t) for t in elts]
-                    if not isArithmetic:
-                        self.emitFatalError(
-                            "non-homogenous list not allowed - must all be same type: {}".
-                            format(elts), node)
-                    for t in elts[1:]:
-                        base_elTy = self.__get_superior_type(base_elTy, t)
-                return cc.StdvecType.get(base_elTy)
-            elif isinstance(pyval, ast.Call) and isinstance(pyval.func, ast.Name):
-                # supported for calls but not here:
-                # 'range', 'enumerate', 'list'
-                if pyval.func.id == 'len' or pyval.func.id == 'int': 
-                    return IntegerType.get_signless(64)
-                elif pyval.func.id == 'complex': 
-                    return self.getComplexType()
-                elif self.__isUnitaryGate(pyval.func.id) or pyval.func.id == 'reset': 
-                    process_void_list(pyval)
-                    return None
-                elif self.__isMeasurementGate(pyval.func.id):
-                    # It's tricky to know if we are calling a measurement on a single qubit, 
-                    # or on a vector of qubits, e.g. consider the case [mz(qs[i:]) for i in range(n)], 
-                    # or [mz(qs) for _ in range(n)], or [mz(qs) for _ in qs].
-                    # We hence limit support to iterating over a vector of qubits, and check that the
-                    # iteration variable is passed directly to the measurement.
-                    isIterOverVeq = isinstance(node.generators[0].iter, ast.Name) and \
-                                    node.generators[0].iter.id in self.symbolTable and \
-                                    quake.VeqType.isinstance(
-                                    self.symbolTable[node.generators[0].iter.id].type)
-                    if not isIterOverVeq:
-                        self.emitFatalError(
-                            "performing measurements in list comprehension expressions is only supported when iterating over a vector of qubits", node)
-                    iterVarPassedAsArg = len(pyval.args) == 1 and \
-                            isinstance(pyval.args[0], ast.Name) and \
-                            isinstance(node.generators[0].target, ast.Name) and \
-                            pyval.args[0].id == node.generators[0].target.id
-                    if not iterVarPassedAsArg:
-                        self.emitFatalError(
-                            "unsupported argument to measurement in list comprehension", node)
-                    return IntegerType.get_signless(1)
-                elif pyval.func.id in globalKernelRegistry:
-                    # Not necessarily unitary
-                    resTypes = globalKernelRegistry[pyval.func.id].type.results
-                    if len(resTypes) == 0:
-                        process_void_list(pyval)
-                        return None
-                    if len(resTypes) != 1 or cc.StructType.isinstance(resTypes[0]):
-                        # TODO: support...
-                        self.emitFatalError(
-                            "Only variables, constants, and calls to built-ins can be used to populate values in list comprehension expressions",
-                            node)
-                    return resTypes[0]
-                    
-                elif pyval.func.id in globalRegisteredTypes.classes:
-                    _, annotations = globalRegisteredTypes.getClassAttributes(
-                        pyval.func.id)
-                    structTys = [
-                        mlirTypeFromPyType(v, self.ctx)
-                        for _, v in annotations.items()
-                    ]
-                    # no need to do much verification on the validity of the type here -
-                    # this will be handled when we build the body
-                    isStruq = any((self.isQuantumType(t) for t in structTys))
-                    if isStruq:
-                        return quake.StruqType.getNamed(pyval.func.id, structTys)
-                    else:
-                        return cc.PointerType.get(cc.StructType.getNamed(pyval.func.id, structTys))
-            else:
-                self.emitFatalError(
-                    "Only variables, constants, and calls to built-ins can be used to populate values in list comprehension expressions",
-                    node)
-
-        listElemTy = get_item_type(node.elt)
-        if listElemTy is None: return
-        listTy = cc.ArrayType.get(listElemTy)
-
-        # General case of
-        # `listVar = [expr(i) for i in iterable]`
-        # Need to think of this as
-        # `listVar = stdvec(iterable.size)`
-        # `for i, r in enumerate(listVar):`
-        # `   listVar[i] = expr(r)`
+                "CUDA-Q only support a single named targets in list comprehension", node)
 
         # Let's handle the following `listVar` types
         # `   %9 = cc.alloca !cc.array<!cc.stdvec<T> x 2> -> ptr<array<stdvec<T> x N>`
@@ -3178,12 +3065,12 @@ class PyASTBridge(ast.NodeVisitor):
         self.visit(node.generators[0].iter)
 
         if len(self.valueStack) == 1:
-            iterable = self.popValue()
+            iterable = self.ifPointerThenLoad(self.popValue())
             iterableSize = None
             if cc.StdvecType.isinstance(iterable.type):
                 iterableSize = cc.StdvecSizeOp(self.getIntegerType(), iterable).result
                 iterTy = cc.StdvecType.getElementType(iterable.type)
-                iterArrPtrTy = cc.PointerType.get(cc.ArrayType.get(cc.PointerType.get(iterTy)))
+                iterArrPtrTy = cc.PointerType.get(cc.ArrayType.get(iterTy))
                 iterable = cc.StdvecDataOp(iterArrPtrTy, iterable).result
             elif quake.VeqType.isinstance(iterable.type):
                 iterableSize = quake.VeqSizeOp(self.getIntegerType(), iterable).result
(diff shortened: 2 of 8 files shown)


python/cudaq/kernel/utils.py
```diff
@@ -185,12 +205,7 @@ def emitFatalErrorOverride(msg):
     if isinstance(annotation,
                   ast.Subscript) and (annotation.value.id == 'tuple' or
                                       annotation.value.id == 'Tuple'):
-        localEmitFatalError(
-            f"Use of tuples is not supported in kernels ({ast.unparse(annotation) if hasattr(ast, 'unparse') else annotation})."
-        )
-
-        #FIXME: re-enable tuple support after we have the spec.
```
bettinaheim, line 192: We are already using tuples liberally within quantum kernels, and we also are using the same representation for dataclasses, which we allow in signatures. It hence seems to me that enabling them doesn't introduce problems that we don't already have with the support as it is. But if someone can think of what needs to be fixed first, let me know.

python/tests/mlir/ast_elif.py
```diff
@@ -19,14 +19,14 @@ def test_elif():
     def cost(thetas: np.ndarray):  # can pass 1D ndarray or list
         q = cudaq.qvector(4)
         for i, theta in enumerate(thetas):
-            if i % 2.0:  # asserting we convert 2.0 to 2
+            if i % 2.0:  # asserting we convert i to float
```
bettinaheim: Converting float to int upon modulo is not matching the behavior that python has. Not sure why we did that.

python/tests/mlir/ast_list_int.py
```diff
@@ -17,7 +17,7 @@ def test_list_int():
     def oracle(register: cudaq.qview, auxillary_qubit: cudaq.qubit,
                hidden_bitstring: list[int]):
         # Also test out len() here, should convert to stdvec_size
-        x = len(hidden_bitstring)
+        n = len(hidden_bitstring)
```
bettinaheim, line 20: It seems that before, x.ctrol was assuming x to be the global x gate despite a variable shadowing it. This does not match python behavior, and we either need to forbit the creation of any identifiers that would shadow built in gates, or we respect python behavior (at least more and more as we improve the bridge)

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3020,97 +3060,233 @@ def visit_ListComp(self, node):
...
+                "CUDA-Q only supports list comprehension on ranges and arrays",
                 node)
 
-        arrayTy = cc.PointerType.getElementType(iterable.type)
-        if not cc.ArrayType.isinstance(arrayTy):
-            self.emitFatalError(
-                "CUDA-Q only considers general list comprehension on iterables from range(...)",
-                node)
+        def process_void_list():
+            # NOTE: This does not actually a create a valid value, and will fail is something
```
sacpis: nit: This does not actually create a valid value, and will fail if something

python/cudaq/kernel/ast_bridge.py
```diff
@@ -787,28 +810,6 @@ def mlirTypeFromAnnotation(self, annotation):
         if msg is not None:
             self.emitFatalError(msg, annotation)
 
-    def argumentsValidForFunction(self, values, functionTy):
```
bettinaheim, line 793: argumentsValueForFunction and checkControlAndTargetTypes are replaced by local functions in visit_Call

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1010,6 +1055,12 @@ def visit(self, node):
         super().visit(node)
         self.indent_level -= 1
 
+    # FIXME: using generic_visit the way we do seems incredibly dangerous;
```
bettinaheim, line 1119: Better to tackle that in a separate PR - this touches everything all over the place.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -508,6 +509,41 @@ def isArithmeticType(self, type):
         return IntegerType.isinstance(type) or F64Type.isinstance(
             type) or F32Type.isinstance(type) or ComplexType.isinstance(type)
 
+    def __isSupportedNumpyFunction(self, id):
+        return id in ['sin', 'cos', 'sqrt', 'ceil', 'exp']
+    
+    def __isSupportedVectorFunction(self, id):
+        return id in ['front', 'back', 'append']
```
bettinaheim: This merely indicates that visit_Call will handle these appropriately. In the case of append, we give an error that it is not supported.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3882,12 +4044,20 @@ def visit_Compare(self, node):
...
+                superior_type = self.__get_superior_type(left_val.type, element.type)
+                promoted_left = self.tryChangeOperandToType(superior_type, left_val)
+                promoted_element = self.tryChangeOperandToType(superior_type, element)
+                if promoted_left.type != superior_type or element.type != superior_type:
+                    self.emitFatalError("failed to convert type for comparison", node)
 
                 iCondPred = IntegerAttr.get(self.getIntegerType(), 0)
                 fCondPred = IntegerAttr.get(self.getIntegerType(), 0)
+
+                # TODO: support... (i.e. adjust cmp_result below)
```
bettinaheim, line 4056: Preexisting todo - I merely moved code around and haven't looked at this.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3907,67 +4077,6 @@ def check_element(idx):
 
             return
 
-    def visit_AugAssign(self, node):
```
bettinaheim, line 3928: I split out and updated what was in visit_BinOp before into a separate function that both visit_BinOp and visit_AugAssign call into. It makes no sense to have augassign duplicate the logic, and it was correspondingly missing plenty.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -1516,6 +1516,18 @@ struct AllocaOpPattern : public OpConversionPattern<cudaq::cc::AllocaOp> {
   }
 };
 
+struct ReturnOpPattern : public OpConversionPattern<func::ReturnOp> {
```
schweitzpgi, line 1519: This appears to be a nop (replacing an operation with a clone of itself) and is not necessary or desired. All `func` dialect operations will be handled at the appropriate time.

python/tests/mlir/ast_elif.py
schweitzpgi: Can we break this bug fix out into a separate PR please?

python/tests/mlir/quantum_struct.py
bettinaheim, line 1: The tests here has been moved rather than deleted.

python/tests/mlir/tuple_assign.py
bettinaheim, line 1: The tests here has been moved rather than deleted.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
bettinaheim: It is not a no-op. This change was necessary to convert any quake types in the return value. If you remove it, you can see a failing test in this PR. Without this change, it is not possible to return a struq value from a quantum kernel.

bettinaheim:
> /ok to test 8f5e619

bettinaheim:
> /ok to test 383584b

python/cudaq/kernel/utils.py
bettinaheim, line 1: The changes here look like more than they are. The main changes are:
- make sure that context is passed everywhere in mlirTypeFromAnnotation (achieved with `with ctx:...`)
- remove the errors to prevent accepting tuples as valid kernel data types
- split out a new mlirTryCreateStructType helper function
- add quake.StruqType to mlirTypeToPyType

python/tests/mlir/ast_elif.py
bettinaheim: https://github.com/NVIDIA/cuda-quantum/pull/3513

bettinaheim:
> /ok to test d4f34a5

schweitzpgi:
We discussed offline, but it is worth mentioning here perhaps as well.

Python tuples are _not_ C++ tuples. Python tuples are quasi-values[*]. C++ tuples are variables. (There are other differences as well.)

This means that semantically, the Python bridge _must_ lower a Python tuple to an SSA value. So input such as
```python
(1, 2, 3.0)
```
will necessarily be required to be translated by the bridge to
```mlir
  %65 = cc.undef !cc.struct<{i64, i64, f64}>
  %66 = cc.insert_value %65[0], %c1 : (!cc.struct<{i64, i64, f64}>, i64) -> !cc.struct<{i64, i64, f64}>
  %67 = cc.insert_value %66[0], %c2 : (!cc.struct<{i64, i64, f64}>, i64) -> !cc.struct<{i64, i64, f64}>
  %68 = cc.insert_value %67[0], %c3_f : (!cc.struct<{i64, i64, f64}>, f64) -> !cc.struct<{i64, i64, f64}>
```

It is incorrect to lower this to a `cc.alloca` and use `cc.store` to initialize that variable.

[*]: Python tuples are always values, but values may have "reference types", hence the "quasi-value" term. A value of reference type (such as a list or object of a user-defined class) requires more consideration. Values of reference type must be lowered as reference values, which are defined to be `!cc.ptr<T>` type in the quake IR. Effectively, the tuple itself must be constructed as explained above—in the value domain—but the values that compose the product type may themselves by references to _variables_.

bettinaheim:
> /ok to test 277c7f7

bettinaheim:
> /ok to test 91ee130

bettinaheim:
> /ok to test 81e48f6

bettinaheim:
> /ok to test 2497c9f

schweitzpgi:
Q: why can't #3513 be merged first?

bettinaheim:
> Q: why can't #3513 be merged first?

There may be some tweaks that this PR adds to e.g. handle the result of modulo in if-conditions. If we really really want to, i can further disentangle, but I don't quite see why we would want to go through the trouble.

bettinaheim:
> /ok to test 0cdf008

python/runtime/cudaq/algorithms/py_draw.cpp
```diff
@@ -20,6 +20,9 @@ namespace cudaq {
 namespace details {
 std::tuple<std::string, MlirModule, OpaqueArguments *>
 getKernelLaunchParameters(py::object &kernel, py::args args) {
+  if (!py::hasattr(kernel, "arguments"))
+    throw std::runtime_error(
+        "unrecognized kernel - did you forget the @kernel attribute?");
```
schweitzpgi, line 25: Might be better if it mentioned `arguments` as the specific attribute that is missing.

python/tests/kernel/test_direct_call_return_kernel.py
```diff
@@ -384,7 +382,7 @@ def simple_tuple_int_float_assign(
 
     with pytest.raises(RuntimeError) as e:
         simple_tuple_int_float_assign(2, (-13, 42.3))
-    assert 'Use of tuples is not supported in kernels' in str(e.value)
+    assert 'indexing into struct elements must not modify value' in str(e.value)
```
schweitzpgi: This code is actually passing a Python tuple. Might be better to use `tuple` in the error message. (In python `struct` refers to some package to interface between C and Python product types.)

python/tests/kernel/test_direct_call_return_kernel.py
```diff
@@ -456,25 +448,34 @@ def simple_tuple_int32_bool_no_args() -> tuple[np.int32, bool]:
...
         return (np.int32(-13), True)
 
-    with pytest.raises(RuntimeError) as e:
-        simple_tuple_int32_bool_no_args1()
-    assert 'Use of tuples is not supported in kernels' in str(e.value)
+    result = simple_tuple_int32_bool_no_args1()
+    # Note: printing the kernel correctly shows the MLIR
+    # values return type as "tuple" {i32, i1}, but we don't
+    # actually create numpy values even when these are requested
+    # in the signature.
```
schweitzpgi, line 465: Can we make this an issue if it isn't already?

runtime/common/ArgumentConversion.cpp
```diff
@@ -559,7 +560,8 @@ Value dispatchSubtype(OpBuilder &builder, Type ty, void *p, ModuleOp substMod,
         case 64:
           return genConstant(builder, *static_cast<std::int64_t *>(p));
         default:
-          return {};
+          throw std::runtime_error("unsupported integer width " +
```
schweitzpgi: This is an internal compiler error and common code. The exception should be caught by the python bindings, but for C++ an exception isn't going to be handled. (In the compiler, we don't use exceptions.)

Maybe this should just call `mlir::emitError` and abort.

runtime/common/ArgumentConversion.cpp
```diff
@@ -614,12 +616,26 @@ static std::size_t getHostSideElementSize(Type eleTy,
 ArrayAttr genRecursiveConstantArray(OpBuilder &builder,
                                     cudaq::cc::StdvecType vecTy, void *p,
                                     llvm::DataLayout &layout) {
+
+  auto eleTy = vecTy.getElementType();
+  // Need special handling for boolean arrays
+  if (auto intTy = dyn_cast<IntegerType>(eleTy)) {
+    if (intTy.getWidth() == 1) {
+      SmallVector<Attribute> members;
+      auto *vecPtr = static_cast<std::vector<bool> *>(p);
```
schweitzpgi, line 627: It appears that this code assumes the `std::vector<bool>` was already converted to something more practical, `std::span<char>`. But looking at `runtime/test/test_argument_conversion.cpp` and it doesn't test `std::vector<bool>`.

So, a `std::vector<bool>` test must be added in the C++ test (see above). Hopefully, that will work.

If however, the original code is correct and there really isn't a `std::vector<bool>` in use here, then we have a problem in Python will converting lists of bool into the wrong type. Generally, using `std::vector<bool>` is a bad idea as it means we have all these special case handling cases littering our code. Someday it would be nice for CUDA-Q to stop using `std::vector<bool>` entirely.

python/cudaq/kernel/utils.py
```diff
@@ -101,6 +101,27 @@ def emitWarning(msg):
...
+    tuples and `dataclass` structs of quantum and classical types. Returns
+    None if the given element types don't satisfy the restrictions imposed
+    on these types.
+    """
+
+    def isQuantumType(ty):
+        return quake.RefType.isinstance(ty) or quake.VeqType.isinstance(
+            ty) or quake.StruqType.isinstance(ty)
+
+    numQuantumMembers = sum((isQuantumType(t) for t in mlirEleTypes))
```
schweitzpgi, line 116: Q: is there a short-circuiting way of doing this? We don't really care how many, just is 1 is present.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1636,29 +1853,20 @@ def get_full_module_path(partial_path):
             # If we did have module names, then this is what we are looking for
             if len(moduleNames):
                 name = node.func.attr
-                if not name in globalKernelRegistry:
```
schweitzpgi, line 1639: There's going to be some conflicts here with my reworking the call stacks, etc. in Python, but these changes are an improvement nonetheless.

schweitzpgi (commented):
This is going to be great. Glad to see some of these bugs getting squished. Thanks, Bettina.

python/runtime/cudaq/algorithms/py_draw.cpp
bettinaheim: It is likely more comprehensive that way - you get this error if you are missing the @kernel attribute on a function and try to run/sample it. I believe arguments is something that our tools (this attribute) injects, and not something a user would know exists. The error message tells them how to fix the issue they are seeing, and stating arguments is missing does give little (no) information about what steps a user should take.

runtime/common/ArgumentConversion.cpp
schweitzpgi: After thinking more about this, I think throwing the exception is the better choice here given the contexts of how this code is used in both Python and C++.

docs/sphinx/examples/python/executing_kernels.ipynb
```diff
@@ -147,12 +147,12 @@
...
     "    # Measure and return total number of qubits in state |1⟩\n",
-    "    result = 0\n",
+    "    res = 0\n",
     "    for i in range(num_qubits):\n",
     "        if mz(qubits[i]):\n",
-    "            result += 1\n",
+    "            res += 1\n",
     "\n",
-    "    return result\n",
+    "    return res\n",
```
sacpis, line 155: Wondering if this change is needed?

docs/sphinx/examples/python/executing_kernels.py
```diff
@@ -62,12 +62,12 @@ def simple_ghz(num_qubits: int) -> int:
...
     # Measure and return total number of qubits in state |1⟩
-    result = 0
+    res = 0
     for i in range(num_qubits):
         if mz(qubits[i]):
-            result += 1
+            res += 1
 
-    return result
+    return res
```
sacpis, line 70: Same here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -359,12 +368,12 @@ def getConstantInt(self, value, width=64):
...
+    def changeOperandToType(self, ty, operand, allowDemotion=False):
         """
-        Change the type of an operand to a specified type. This function
-        primarily handles type conversions and promotions to higher types
-        (complex > float > int). Demotion of floating type to integer is allowed
-        by default.
+        Change the type of an operand to a specified type. This function primarily 
+        handles type conversions and promotions to higher types (complex > float > int). 
+        Demotion of floating type to integer is not allowed by default.
+        Regardless of whether demotion is allowed, types will be cast to smaller widths.
```
sacpis, line 376: nit: casted

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2283,746 +2453,741 @@ def bodyBuilder(iterVal):
...
+                                'controlled operation requested without any control argument(s).',
+                                node)
+                        negatedControlQubits = None
+                        if len(self.controlNegations):
+                            negCtrlBools = [None] * len(controls)
+                            for i, c in enumerate(controls):
+                                negCtrlBools[i] = c in self.controlNegations
+                            negatedControlQubits = DenseBoolArrayAttr.get(
+                                negCtrlBools)
+                            self.controlNegations.clear()
```
sacpis: Seems like this has been repeated several times. Should we make a function of it and just call it here?

sacpis (approved):
LGTM. Thanks @bettinaheim. Left a few comments.

python/tests/kernel/test_direct_call_return_kernel.py
bettinaheim: https://github.com/NVIDIA/cuda-quantum/issues/3524

python/cudaq/kernel/utils.py
bettinaheim: We do care; two lines below is the check `numQuantumMembers != len(mlirEleTypes)` to give an error if you have a mix of quantum and non-quantum members.

python/cudaq/kernel/ast_bridge.py
bettinaheim: Ok, happy to help revise to incorporate your changes.

docs/sphinx/examples/python/executing_kernels.ipynb
bettinaheim: Yes, it is. The reason is that the file in its entirety contains a global variable outside the kernel that is called result. As it is implemented right now, the builder does not create a new variable when you define a local variable with the same name inside a kernel. This is another bug in the bridge that does not match python behavior, but I haven't gotten to revising it.

python/cudaq/kernel/ast_bridge.py
bettinaheim: Not a native English speaker, but the dictionary says cast is correct?

python/cudaq/kernel/ast_bridge.py
bettinaheim: Oh wow, you found another one... Looks like we are not consistently handling negated control qubits. Yeah, splitting out a function and making a pass over visit_Call to actually respect negated control qubits across the board...
This is why this PR keeps getting bigger...

python/cudaq/kernel/ast_bridge.py
sacpis: My bad.

runtime/common/ArgumentConversion.cpp
bettinaheim: @schweitzpgi I added a tests to the argument conversion with a vector bool. Looks like it should be working fine to me. Can you please double check that this all looks good and we have the test coverage we need?
Edit: It looks like it is unhappy with a vector of vector of bool. I'll look closer a bit later.

bettinaheim: So, I did confirm a couple of things:
1) the implementation prior to my change does not work for C++ either.
2) the vect<vect<bool>> issue is at the heart the same as the vect<bool> issue; under the hood, the outer vector is not an array with 3 char*.
3) vect<vect<bool>> works in C++ if I follow the same approach as for the inner one and stupidly cast the *p to static_cast<std::vector<std::vector<bool>> *>. 
4) haven't checked vect vect bool in python yet
5) I am not sure what the best solution here is. maybe we should convert vect<bool> into something more standard before it gets to arg synthesis. However, that requires updating a bunch of other things, including adjusting the logic of how __nvqpp_initializer_list_to_vector_bool is used when targeting hw backends.

bettinaheim:
> /ok to test b89d8f3

runtime/common/ArgumentConversion.cpp
bettinaheim: deferred for a separate PR per offline discussions. For now, the argument conversion will emit an error when we have a vector<vector<bool>>.

PR #3490, opened 2025-10-02
Update mgpu SHA to pick up a downstream bug fix

1tnguyen:
### Description

The input stage gate fusion (https://github.com/NVIDIA/cuda-quantum/pull/3479) has a bug: incorrect fusion in some scenarios.

This has been rectified in the downstream library. Hence, updating the SHA here to pick up the fix.


1tnguyen:
Closing as there should be another sha update fixing the downstream build script.

PR #3496, opened 2025-10-06
[runtime] Fix interface bugs.

schweitzpgi:
Fixes kernel_builder to use the correct declaration of altLaunchKernel and pass it the correct arguments.


runtime/cudaq/platform/quantum_platform.h
```diff
@@ -15,14 +15,14 @@
 #include "common/ThunkInterface.h"
 #include "cudaq/remote_capabilities.h"
 #include "cudaq/utils/cudaq_utils.h"
+#include "nvqpp_interface.h"
```
khalatepradnya, line 18: With the addition of this new header file, those declarations in this file can be removed? (lines 239-258)

PR #3498, opened 2025-10-07
[NFC] Add timing tag for tensornet API calls

1tnguyen:
### Description

Add a specific timing tag for certain TensorNet API calls of interest, such as `prepare` (contraction path optimization) and `compute` (contraction).


mitchdz (approved):
Thanks for this @1tnguyen, this will be very useful for tensornet debugging.

PR #3504, opened 2025-10-08
Handle `Helios-1SC` empty results

1tnguyen:
### Description

- Syntax checker for Helios wouldn't return any results, hence need to handle it differently.

- Add a note to our docs to describe the expected result when using the Helios syntax checker.

### DIFF
--- a/runtime/cudaq/platform/default/rest/helpers/quantinuum/QuantinuumServerHelper.cpp
+++ b/runtime/cudaq/platform/default/rest/helpers/quantinuum/QuantinuumServerHelper.cpp
@@ -95,6 +95,11 @@ protected:
   std::string extractOutputLog(ServerMessage &postJobResponse,
                                std::string &jobId) override;
 
+  /// @brief Helper to determine if a completed job returns a result.
+  // Some jobs, such as syntax checker jobs, may complete without returning a
+  // result.
+  bool jobReturnsResult(ServerMessage &postJobResponse) const;
+
 public:
   /// @brief Return the name of this server helper, must be the
   /// same as the qpu config file.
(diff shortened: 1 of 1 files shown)


runtime/cudaq/platform/default/rest/helpers/quantinuum/QuantinuumServerHelper.cpp
```diff
@@ -559,7 +575,17 @@ std::string QuantinuumServerHelper::extractOutputLog(ServerMessage &jobResponse,
                                                      std::string &jobId) {
   const auto [resultType, resultId] = getResultId(jobResponse);
   if (resultId.empty()) {
-    throw std::runtime_error("Job completed but no result ID found.");
+    const std::string deviceName =
+        jobResponse["data"]["attributes"]["definition"]["backend_config"]
+                   ["device_name"]
+                       .get<std::string>();
+    // NG-device (Helios) syntax checker jobs will not return results.
+    if (deviceName.starts_with("Helios") && deviceName.ends_with("SC")) {
```
khalatepradnya: Since this check occurs multiple times and contains hard-coded strings, I wonder if we can have a common function like `willReturnResults` or `waitForResults` (or `wontReturnResults`)?

1tnguyen: Good point! Refactored into a helper as suggested in https://github.com/NVIDIA/cuda-quantum/pull/3504/commits/aa9fbb03c238229639576276c054bca5193de8c7

runtime/cudaq/platform/default/rest/helpers/quantinuum/QuantinuumServerHelper.cpp
```diff
@@ -684,6 +699,20 @@ cudaq::ExtraPayloadProvider *QuantinuumServerHelper::getExtraPayloadProvider() {
   return it->get();
 }
 
+bool QuantinuumServerHelper::jobReturnsResult(
+    ServerMessage &jobResponse) const {
+  // Retrieve the device name if available.
+  auto deviceNamePath =
+      "/data/attributes/definition/backend_config/device_name"_json_pointer;
```
khalatepradnya, line 706: nit: May be move this endpoint to the constants at top of this file.

1tnguyen: After looking at the code, I think we better keep it here as those constants are API endpoints rather than Json field paths.

PR #3505, opened 2025-10-08
[CI] [nightly] Run the hardware integration tests parallelly for each provider

khalatepradnya:
* Order alphabetically by provider name
* Enable all tests to check the status


.github/workflows/integration_tests.yml
```diff
@@ -821,35 +725,78 @@ jobs:
...
+              ;;
+          esac
+          
           set -e # Re-enable exit code error checking
           if [ ! $test_err_sum -eq 0 ]; then
             echo "::error::${test_err_sum} tests failed. See step summary for a list of failures"
             exit 1
           fi
-
-      - name: Submit to ${{ inputs.target }}
```
khalatepradnya, line 833: Removed this since I can't find any use of it. @mitchdz - let me know if this needs to be retained.

.github/workflows/integration_tests.yml
```diff
@@ -334,35 +356,127 @@ jobs:
...
+                filelist="docs/sphinx/targets/cpp/pasqal.cpp docs/sphinx/targets/python/pasqal.py"
+                ;;
+              qci)
+                filelist="targettests/qci/*.cpp"
+                ;;
+              quantinuum)
+                # Include both legacy Quantinuum and Quantinuum-NG tests
+                filelist="targettests/quantinuum/*.cpp targettests/quantinuum_ng/*.cpp docs/sphinx/targets/python/quantinuum.py"
+                ;;
+            esac
```
sacpis, line 396: I think this can be made to fetch all files in a folder (as per the provider) as it will be a lengthly lists of tests if the number of tests increases.

sacpis: That way we can avoid hard coding the filenames as well.

.github/workflows/integration_tests.yml
```diff
@@ -334,35 +356,127 @@ jobs:
...
+          case "${{ matrix.provider }}" in
+            anyon)
+              echo "### Setting up Anyon account" >> $GITHUB_STEP_SUMMARY
+              curl -X POST --user "${{ secrets.ANYON_USERNAME }}:${{ secrets.ANYON_PASSWORD }}" \
+                -H "Content-Type: application/json" https://api.anyon.cloud:5000/login > credentials.json
+              id_token=`cat credentials.json | jq -r '."id_token"'`
+              refresh_token=`cat credentials.json | jq -r '."refresh_token"'`
+              echo "key: $id_token" > ~/.anyon_config
+              echo "refresh: $refresh_token" >> ~/.anyon_config
+              ;;
```
sacpis, line 412: For each of these provider, the env commands can go into separate env files, which we can call here and execute those.

sacpis: That way we just need to have one case or rather a call like `execute_env(${{ matrix.provider }})`.

.github/workflows/integration_tests.yml
```diff
@@ -593,219 +630,86 @@ jobs:
...
+                if [ $test_status -eq 0 ]; then
+                  echo ":white_check_mark: Successfully ran test: $filename" >> $GITHUB_STEP_SUMMARY
+                else
+                  echo ":x: Test failed (failed to execute): $filename" >> $GITHUB_STEP_SUMMARY
+                  test_err_sum=$((test_err_sum+1))
+                fi
+              done
+              ;;
+              
+            qci)
```
sacpis, line 710: We can do similar thing with test execution where the commands would go in separate files and we call those files here to execute those commands.

sacpis:
Just to make it dynamic, where we can plug and pull any tests from the integration suite, those tests names can be fetched from the file names (with provider names).

.github/workflows/integration_tests.yml
mitchdz: If you see no good use for it, then it could be fine to remove. You will see it doesn't get ran on the automated test because of the `github.event_name == 'workflow_dispatch'` check. I'm not fully familiar with Quantinuum, so maybe it is useful to keep this in for manual testing from time to time?

mitchdz (approved):
LGTM. There are some refactor comments that we can circle back to later.

.github/workflows/integration_tests.yml
khalatepradnya: Opened new issue https://github.com/NVIDIA/cuda-quantum/issues/3554

.github/workflows/integration_tests.yml
khalatepradnya: Removing since it is not being used.

1tnguyen (approved):
👍

PR #3506, opened 2025-10-09
[testing] [Python] Make the imports independent and non-fatal

khalatepradnya:
This PR improves the robustness of the `utils/mock_qpu` module's initialization by ensuring that import errors for individual mock QPU backends do not prevent the rest of the module from loading. Instead, warnings are issued for any failed imports.


khalatepradnya:
I got the `python/tests/backends/*.py` tests to run locally, but they are still not running in the CI as expected.
I will add a follow-up PR to address this.

PR #3507, opened 2025-10-10
Various fixes for dynamic kernel loading

1tnguyen:
### Description

(1) Allow override in `registerDeviceKernel`

(2) Clear kernel decorator's `module` (parsed from walking the Python AST) as we switch target. 
The lack of this `module` attribute tells the decorator to perform a proper `compile()`.

Note: Other global maps might be populated during kernel constructor (i.e., as Python sees kernels); hence might not be safe to clear.

(3) Python to honor the default simulator environment variable for hardware target emulation. It was hardcoded to `qpp` for those targets that don't define a simulator.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -434,6 +434,7 @@ class PyASTBridge(ast.NodeVisitor):
                                  operand,
                                  sint=operand_width != 1,
                                  zint=operand_width == 1).result
+
         self.emitFatalError(
             f'cannot convert value of type {operand.type} to the requested type {ty}',
             self.currentNode)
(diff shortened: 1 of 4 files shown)


python/utils/LinkedLibraryHolder.cpp
```diff
@@ -115,7 +115,7 @@ void parseRuntimeTarget(const std::filesystem::path &cudaqLibPath,
     }
   }
   target.platformName = foundPlatformName.value_or("default");
-  target.simulatorName = foundSimulatorName.value_or("qpp");
+  target.simulatorName = foundSimulatorName.value_or("");
```
sacpis, line 118: Does this mean the default won't be `qpp` simulator anymore? Just curious to know what would the default option be?

python/utils/LinkedLibraryHolder.cpp
```diff
@@ -422,8 +422,27 @@ void LinkedLibraryHolder::setTarget(
 
   CUDAQ_INFO("Setting target={} (sim={}, platform={})", targetName,
              target.simulatorName, target.platformName);
+  std::string simName = target.simulatorName;
+  if (simName.empty()) {
+    // This target doesn't a simulator defined, e.g., hardware targets.
```
sacpis: nit: doesn't have a

python/utils/LinkedLibraryHolder.cpp
```diff
@@ -422,8 +422,27 @@ void LinkedLibraryHolder::setTarget(
...
+    // We still need a simulator in case of local emulation.
+    auto &defaultTargetInfo = targets[defaultTarget];
```
sacpis, line 429: Got it. Here is the default simulator.

1tnguyen: Yeah, without this, we're stuck with `qpp` for all hardware emulation when running on Python, e.g., cannot use `stim` for emulation of Clifford circuits.

python/utils/LinkedLibraryHolder.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3507/commits/9d9228098ee2485e6ca0eb81a14caebf4b297bdf

python/cudaq/__init__.py
```diff
@@ -213,6 +214,11 @@ def __clearKernelRegistries():
     globalRegisteredOperations.clear()
 
 
+cudaq_runtime.register_set_target_callback(
+    lambda _:
+    [setattr(kernel, "module", None) for kernel in globalKernelDecorators],
+    "clearKernelDecoratorModules")
```
sacpis, line 220: Can we get rid of hardcoded values here?

1tnguyen: This string is just used for `info` logging of the callback. As this is only used once here, it might make sense to keep it as a string literal, e.g., vs. declaring an additional variable for it.

schweitzpgi:
I'm going to very likely revert this with my changes.

Yes. I'll revert these changes in my total rewrite of the python handling of builders, decorators, compilation, and launching.

PR #3508, opened 2025-10-10
[testing] [Python] Make the test set up work with the latest version of `llvmlite`

khalatepradnya:
Fix for following error that prevents from running the `backend tests in CI - 
```
llvmlite.binding.initialize() is deprecated and will be removed. LLVM initialization is now handled automatically. Please remove calls to this function from your code and check for other behavioral changes that may have occurred due to LLVM updates.
```

We should refactor this test set up to reduce code duplication, but that is out-of-scope for this PR.


khalatepradnya:
See: https://github.com/NVIDIA/cuda-quantum/pull/3541

khalatepradnya:
This PR is not required if we pin the `llvmlite` version to `0.42.0` (See devdeps images)

PR #3511, opened 2025-10-14
Handle additional target arguments for QCI

khalatepradnya:
This PR enhances the `qci` backend by introducing support for new target arguments and updating how job options are structured. 
* Added the `noisy` and `repeat_until_shots_requested` parameters


docs/sphinx/using/backends/hardware/superconducting.rst
```diff
@@ -333,11 +333,43 @@ named :code:`QCI_AUTH_TOKEN` before running your CUDA-Q program.
...
+        You can control the execution method using the :code:`method` parameter:
+
+        .. code:: python
+
+            # For simulation (default)
+            cudaq.set_target('Seeker', method='simulate')
+            # For hardware execution
+            cudaq.set_target('Seeker', method='execute')
+
+        For noisy simulation, you can enable the :code:`noisy` parameter:
```
1tnguyen, line 354: There is a sentence above (in the introduction) that says noisy simulation is not supported. We may need to update that as well.

docs/sphinx/using/backends/hardware/superconducting.rst
```diff
@@ -333,11 +333,43 @@ named :code:`QCI_AUTH_TOKEN` before running your CUDA-Q program.
...
+
+        .. code:: python
+
+            cudaq.set_target('qci', noisy=True)
+
+        When collecting shots, you can ensure the requested number of shots are obtained
+        by enabling the :code:`repeat_until_shots_requested` parameter:
```
1tnguyen, line 361: It might be worth adding a note that the default is only returning successful shots, which may be fewer than the requested shots.

docs/sphinx/using/backends/hardware/superconducting.rst
khalatepradnya: Good catch! Updated in 2252e9c to also include QPU description.

docs/sphinx/using/backends/hardware/superconducting.rst
khalatepradnya: Good point! Added in https://github.com/NVIDIA/cuda-quantum/commit/2252e9c6c0ce4465a21a12a3ba4cfbdfaf5c3fa0.

PR #3515, opened 2025-10-14
Fix the publishing pipeline

mitchdz:
The publishing pipeline has 2 errors:

1) various python notebooks failed to find a cuda header. This is remedied by installing the right cuda runtime headers from pypi.

2) Test installation error was failing for python3 .12 wheels
 - There is no `0.8.0` wheels for `cuda-quantum` package anymore, so this test is invalid. This PR just skips that test now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
   'nvidia-cublas-cu12 ~= 12.0',
   'nvidia-curand-cu12 ~= 10.3',
   'nvidia-cusparse-cu12 ~= 12.5',
-  'nvidia-cuda-runtime-cu12 ~= 12.0',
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
   'nvidia-cublas-cu12 ~= 12.0',
   'nvidia-curand-cu12 ~= 10.3',
   'nvidia-cusparse-cu12 ~= 12.5',
-  'nvidia-cuda-runtime-cu12 ~= 12.0',
+  'nvidia-cuda-runtime-cu12 ~= 12.6',
```
bmhowe23, line 28: Per internal discussion, we may need to rethink this change.

mitchdz:
This particular PR is no longer needed. Closing.

PR #3517, opened 2025-10-15
[sample] When targeting hardware backends, launch kernel only once.

khalatepradnya:
This PR updates the sampling logic for hardware and emulated QPU platforms by ensuring that sampling is executed only once in those cases, instead of in a loop. This avoids automatic multiple executions on hardware.

- [x] Verify the expected behavior in emulation mode: All tests pass

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -34,7 +34,7 @@ on:
       target_machine:
         type: string
         required: false
-        description: 'Target machine (e.g., H1-1E).'
+        description: 'Target machine (e.g., H2-1E).'
       cudaq_test_image:
         type: string
         required: false
@@ -407,7 +407,7 @@ jobs:
             case $filename in
               targettests/quantinuum/*)
                 [ -e "$filename" ] || echo "::error::Couldn't find files ($filename)"
-                nvq++ -v $filename -DSYNTAX_CHECK --target quantinuum --quantinuum-machine H1-1SC --quantinuum-project ${{ secrets.QUANTINUUM_NEXUS_PROJECT_ID }}
+                nvq++ -v $filename -DSYNTAX_CHECK --target quantinuum --quantinuum-machine H2-1SC --quantinuum-project ${{ secrets.QUANTINUUM_NEXUS_PROJECT_ID }}
                 test_status=$?
                 if [ $test_status -eq 0 ]; then
                   ./a.out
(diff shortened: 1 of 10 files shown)


runtime/cudaq/algorithms/sample.h
```diff
@@ -95,6 +95,17 @@ runSampling(KernelFunctor &&wrappedKernel, quantum_platform &platform,
   platform.set_exec_ctx(ctx.get(), qpu_id);
   platform.set_current_qpu(qpu_id);
 
+  // If target is hardware backend, launch once on the QPU
+  if (platform.is_remote() || platform.is_emulated()) {
```
khalatepradnya: Need to exclude remote simulators...

python/cudaq/runtime/sample.py
```diff
@@ -116,6 +116,13 @@ def sample(kernel,
     ctx.explicitMeasurements = explicit_measurements
     cudaq_runtime.setExecutionContext(ctx)
 
+    # If the platform is a hardware QPU, launch once
+    if cudaq_runtime.isQuantumDevice():
+        kernel(*args)
+        cudaq_runtime.resetExecutionContext()
+        cudaq_runtime.unset_noise()
+        return ctx.result
+
```
schweitzpgi: The code below on lines 123-128 is intended to do this. Is the code below incorrect, not working for some reason, or ?

khalatepradnya: That code runs in a while loop which keeps invoking the kernel until the shots count is reached - this is frequently used in simulation of kernels with mid-circuit measurement which are run shot-by-shot. But, when sending the kernel to remote hardware backends, automatically re-running the kernel till shot count is reached (because real hardware may have <100% successful shots) may cause unexpected cost to the user.

schweitzpgi: It doesn't matter that it is the while loop, since the point of it is to exit the loop after 1 iteration.

It still feels redundant somehow. You add a test to avoid a loop completely and yet the loop has a test in it to -- avoid iterating the loop completely. Can't these things be combined with a better conditional test?

khalatepradnya: The code can surely be refactored to early exit in case of quantum devices - even if all requested shots haven't been received. 
The purpose of this PR is to facilitate internal testing (still ongoing). I will address these concerns and mark ready.

bettinaheim (approved):

PR #3502, opened 2025-10-08
Support CUDA 13

mitchdz:
## CUDA 13 Support

This update adds **CUDA 13.0** build support.

Working deployments test - https://github.com/NVIDIA/cuda-quantum/actions/runs/18562671783

### Changes
- Add support for **CUDA 13**, including **CC 100** (Blackwell) only when CUDA ≥ 13.0  
  - CUDA 12.6 does not recognize CC 100; future updates will move to CUDA 12.9  
- Remove **Volta (CC 70)** from CMake targets  
- Add **Turing (CC75)** to CMake targets
- Update docs to remove **CUDA 11.x** references  
- **UCX:** `v1.16.0` → `v1.19.0`  
- **CuPy:** `13.4.1` → `13.6.0` for both `cupy-cuda12x` and `cupy-cuda13x`  
- Default build targets **CUDA 13** in `pyproject.toml`  
- Update wheel Dockerfile to handle CUDA 13 ↔ CUDA 12 deps  
- (Thanks @1tnguyen) Removed deprecated Thrust calls per [CCCL v3.0.0](https://github.com/NVIDIA/cccl/releases/tag/v3.0.0)
- prevent libc++ from redefining conflicting headers from CCCL https://github.com/NVIDIA/cuda-quantum/pull/3502/commits/1934d3af7e9b2fb2bffa228d156f4b9b9b4a61df
  - functions are redefined in the new CCL  in https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/std/__new/allocate.h which conflict with the definitions from the thrust library.
- fix publishing pipeline
  - for python 3.12 skip the test installation error, as the old cuda-quantum pypi packages for `0.8.0` does not exist for those wheels.
- update mgpu sha for cusvim cuda13 support

Open issues (to be address in future PRs)
- publishing pipeline does not work with notebooks
- install nightly torch for cuda13 support
- better handle CCCL update https://github.com/NVIDIA/cuda-quantum/issues/3519
- Create 2 pyproject.toml files, `pyproject.toml.cu{12,13}` https://github.com/NVIDIA/cuda-quantum/issues/3520

PR #3518, opened 2025-10-15
update mgpu commit sha

mitchdz:
this mgpu update adds CUDA 13 support in cusvim.


mitchdz:
combined in https://github.com/NVIDIA/cuda-quantum/pull/3502

PR #3530, opened 2025-10-20
[fix] [sample] Correctly parse QIR log for results that are out of order

khalatepradnya:
This PR improves the parsing and handling of output labels in the `RecordLogParser` and adds a new unit test to verify correct ordering of parsed results.

### DIFF
--- a/runtime/common/RecordLogParser.cpp
+++ b/runtime/common/RecordLogParser.cpp
@@ -103,6 +103,7 @@ void cudaq::RecordLogParser::handleOutput(
   const std::string &recType = entries[1];
   const std::string &recValue = entries[2];
   std::string recLabel = (entries.size() == 4) ? entries[3] : "";
+  cudaq::trim(recLabel);
   if (recType == "RESULT") {
     // Sample-type QIR output, where we have an array of `RESULT` per shot. For
     // example,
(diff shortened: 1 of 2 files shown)


runtime/common/RecordLogParser.cpp
```diff
@@ -128,11 +129,18 @@ void cudaq::RecordLogParser::handleOutput(
...
-    processArrayEntry(recValue,
-                      fmt::format("[{}]", containerMeta.processedElements));
+    // Get the index from the label, if available.
+    if (!recLabel.empty() && recLabel[0] == 'r' && recLabel.back() >= '0' &&
+        recLabel.back() <= '9') {
+      recLabel = recLabel.substr(1);
+    } else {
+      // Note: we expect the results are sequential in the same order that mz
+      // operations are called. This may include results in named registers
+      // (specified in kernel code) and other auto-generated register names.
```
qci-amos, line 139: What are the circumstances where this might happen?

khalatepradnya: Ordered schema.

qci-amos: Can you give me an idea how this would manifest in a bc on my end if this is something I need to be aware of?

Currently the only thing I really guarantee is that the label is preserved.

khalatepradnya: I don't think any change on your end is required.

PR #3533, opened 2025-10-21
Various fixes for vector support in Python

1tnguyen:
### Description

(1) Handle i1 type, especially `std_vec<i1>`, similar to C++ to make sure interoperability. The underlying pointer data access should be cast to `i8*`.

(2) Expose `to_integer` intrinsic to Python. It now works thanks to (1). Add a Python test using the reference C++ test case. 

(3) Add the missing `__nvqpp_vectorCopyToStack` call in Python-generated IR. This makes the Python-generated IR equivalent to the one generated from C++, hence allowing for more optimal code generation.


python/cudaq/kernel/ast_bridge.py
```diff
@@ -578,27 +579,39 @@ def ifNotPointerThenStore(self, value):
         if not cc.PointerType.isinstance(value.type):
             slot = cc.AllocaOp(cc.PointerType.get(value.type),
                                TypeAttr.get(value.type)).result
+            assert cc.PointerType.get(value.type) == slot.type
```
khalatepradnya: Can we replace these `assert` statements with explicit exception handling - like raise `ValueError` ?

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -371,6 +371,136 @@ static constexpr IntrinsicCode intrinsicTable[] = {
     {cudaq::stdvecBoolCtorFromInitList, {}, R"#(
   func.func private @__nvqpp_initializer_list_to_vector_bool(!cc.ptr<none>, !cc.ptr<none>, i64) -> ())#"},
 
+    {"__nvqpp_internal_number_of_digits", {}, R"#(
```
khalatepradnya: Can we add quake test(s) to showcase how the new intrinsics are exercised?

python/tests/kernel/test_run_kernel.py
```diff
@@ -511,18 +534,100 @@ def test_return_list_float64():
...
+        shots = 2
+        results = cudaq.run(measure_all_qubits, numQubits, shots_count=shots)
+        assert len(results) == shots
+        for res in results:
+            assert len(res) == numQubits
+            for i in range(numQubits):
+                if i % 2 == 0:
+                    assert res[i] == True
+                else:
+                    assert res[i] == False
```
khalatepradnya: Nice set of tests! 👍

python/cudaq/kernel/ast_bridge.py
1tnguyen: After reviewing these assertions, I realized that they are redundant and wouldn't catch anything. Hence, I removed them accordingly (https://github.com/NVIDIA/cuda-quantum/pull/3533/commits/a5bb0c273ae94d5992096106fe1d46c3e6a0a134). 

These are probably left over unintentionally from my debug code.

lib/Optimizer/Builder/Intrinsics.cpp
1tnguyen: Good point! Added in https://github.com/NVIDIA/cuda-quantum/pull/3533/commits/43c1d9a54886875689355e95b64bd5332e878fc7

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -158,7 +162,55 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
                 Value w = rewriter.create<cudaq::cc::LoadOp>(loc, v);
                 genOutputLog(loc, rewriter, w, offset);
               }
+              return;
             }
+
+          // If we reach here and we cannot determine the constant size of the
+          // buffer, then we will not generate dynamic output logging with a for
```
schweitzpgi:
```suggestion
          // buffer, then we will generate dynamic output logging with a for
```
?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -585,20 +586,31 @@ def ifNotPointerThenStore(self, value):
     def __createStdvecWithKnownValues(self, size, listElementValues):
         # Turn this List into a StdVec<T>
         arrSize = self.getConstantInt(size)
-        arrTy = cc.ArrayType.get(listElementValues[0].type)
+        elemTy = listElementValues[0].type
```
schweitzpgi, line 589: This probably ought to be 2 PRs. 1 for the bool fixes and 1 for the logging of vector changes.

schweitzpgi: I think that means breaking out (3) as a separate PR.

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -371,6 +371,136 @@ static constexpr IntrinsicCode intrinsicTable[] = {
...
+  func.func private @__nvqpp_internal_number_of_digits(%arg0: i64) -> i64 {
```
schweitzpgi: Is this function only used for the integer->string conversion?

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -371,6 +371,136 @@ static constexpr IntrinsicCode intrinsicTable[] = {
...
+      cc.continue 
+    } 
+    %4 = cc.load %2 : !cc.ptr<i64> 
+    return %4 : i64 
+  } 
+  )#"},
+
+    // __nvqpp_internal_tostring
+    {"__nvqpp_internal_tostring", {}, R"#(
+  func.func private @__nvqpp_internal_tostring(%buf: !cc.stdvec<i8>, %val: i64) {
```
schweitzpgi: Here is a tighter function that doesn't need to compute the length and is already lowered to CFG, etc.

```mlir
  // Convert a non-negative i64 value to a string base 10 for printing.
  // argument 0: buffer from cc.alloca !cc.array<i8 x 32>
  // argument 1: the unsigned value
  func.func @__nvqpp__to_string(%arg0: !cc.ptr<!cc.array<i8 x ?>>, %arg1: i64) {
    %c48_i64 = arith.constant 48 : i64
    %c0_i8 = arith.constant 0 : i8
    %c10_i64 = arith.constant 10 : i64
    %c1_i32 = arith.constant 1 : i32
    %c31_i32 = arith.constant 31 : i32
    %false = arith.constant false
    %c48_i8 = arith.constant 48 : i8
    %c0_i32 = arith.constant 0 : i32
    %0 = cc.cast %arg0 : (!cc.ptr<!cc.array<i8 x ?>>) -> !cc.ptr<i8>
    cc.store %c48_i8, %0 : !cc.ptr<i8>
    cf.br ^bb1(%c0_i32, %arg1 : i32, i64)
  ^bb1(%1: i32, %2: i64):  // 2 preds: ^bb0, ^bb4
    %3 = arith.cmpi ne, %2, %c0_i64 : i64
    %4 = arith.cmpi eq, %3, %false : i1
    cf.cond_br %4, ^bb3(%false : i1), ^bb2
  ^bb2:  // pred: ^bb1
    %5 = arith.cmpi slt, %1, %c31_i32 : i32
    cf.br ^bb3(%5 : i1)
  ^bb3(%6: i1):  // 2 preds: ^bb1, ^bb2
    cf.cond_br %6, ^bb4(%1, %2 : i32, i64), ^bb5(%1 : i32)
  ^bb4(%7: i32, %8: i64):  // pred: ^bb3
    %9 = arith.addi %7, %c1_i32 : i32
    %10 = cc.compute_ptr %arg0[%7] : (!cc.ptr<!cc.array<i8 x ?>>, i32) -> !cc.ptr<i8>
    %11 = arith.remui %8, %c10_i64 : i64
    %12 = arith.addi %11, %c48_i64 : i64
    %13 = cc.cast %12 : (i64) -> i8
    cc.store %13, %10 : !cc.ptr<i8>
    %14 = arith.divui %8, %c10_i64 : i64
    cf.br ^bb1(%9, %14 : i32, i64)
  ^bb5(%15: i32):  // pred: ^bb3
    %16 = arith.cmpi eq, %15, %c0_i32 : i32
    cf.cond_br %16, ^bb6, ^bb7(%15 : i32)
  ^bb6:  // pred: ^bb5
    %17 = cc.compute_ptr %arg0[1] : (!cc.ptr<!cc.array<i8 x ?>>) -> !cc.ptr<i8>
    cc.store %c0_i8, %17 : !cc.ptr<i8>
    return
  ^bb7(%18: i32):  // pred: ^bb5
    %19 = cc.compute_ptr %arg0[%18] : (!cc.ptr<!cc.array<i8 x ?>>, i32) -> !cc.ptr<i8>
    cc.store %c0_i8, %19 : !cc.ptr<i8>
    %20 = arith.subi %18, %c1_i32 : i32
    cf.br ^bb8(%20, %c0_i32 : i32, i32)
  ^bb8(%21: i32, %22: i32):  // 2 preds: ^bb7, ^bb9
    %23 = arith.cmpi slt, %22, %21 : i32
    cf.cond_br %23, ^bb9(%21, %22 : i32, i32), ^bb10
  ^bb9(%24: i32, %25: i32):  // pred: ^bb8
    %26 = cc.compute_ptr %arg0[%25] : (!cc.ptr<!cc.array<i8 x ?>>, i32) -> !cc.ptr<i8>
    %27 = cc.load %26 : !cc.ptr<i8>
    %28 = cc.compute_ptr %arg0[%24] : (!cc.ptr<!cc.array<i8 x ?>>, i32) -> !cc.ptr<i8>
    %29 = cc.load %28 : !cc.ptr<i8>
    cc.store %29, %26 : !cc.ptr<i8>
    %30 = arith.subi %24, %c1_i32 : i32
    cc.store %27, %28 : !cc.ptr<i8>
    %31 = arith.addi %25, %c1_i32 : i32
    cf.br ^bb8(%30, %31 : i32, i32)
  ^bb10:  // pred: ^bb8
    return
  }
```

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -98,6 +98,7 @@ void createTargetCodegenPipeline(PassManager &pm,
   pm.addNestedPass<func::FuncOp>(createCSEPass());
   ::addQIRConversionPipeline(pm, options.target);
   pm.addPass(cudaq::opt::createReturnToOutputLog());
+  cudaq::opt::addLowerToCFG(pm);
```
schweitzpgi: not needed with the above function

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
1tnguyen: Good catch! My error when editing the original error message.

lib/Optimizer/Builder/Intrinsics.cpp
1tnguyen: Yeah, it's only used for the integer->string conversion.
Originally, I wanted to make the `to_string` produce a compact string. This string then needs to be assembled with prefix and postfix, thus having this as a separate function to help with the allocation of the integer string and the final string buffers.

python/cudaq/kernel/ast_bridge.py
1tnguyen: That makes sense. Let me do that.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3679,6 +3771,9 @@ def fix_negative_idx(idx, get_size):
                                    upper=upperVal).result)
             elif cc.StdvecType.isinstance(var.type):
                 eleTy = cc.StdvecType.getElementType(var.type)
+                isBool = eleTy == self.getIntegerType(1)
+                if isBool:
+                    eleTy = self.getIntegerType(8)
```
khalatepradnya: nit / style: 
```
eleTy = self.getIntegerType(8) if eleTy == self.getIntegerType(1) else cc.StdvecType.getElementType(var.type)
```

lib/Optimizer/Builder/Intrinsics.cpp
1tnguyen: We've moved this into an (upcoming) separate PR.

lib/Optimizer/Builder/Intrinsics.cpp
1tnguyen: This will be incorporated in the upcoming separate PR dealing with returning vectors from Python `cudaq.run`.

python/cudaq/kernel/ast_bridge.py
1tnguyen: Thanks! I've edited this code in https://github.com/NVIDIA/cuda-quantum/pull/3533/commits/1c8cddced2b54870540a9a9605a8fb23cf13d6a0. 

Indeed, this `isBool` variable is not needed, so it has been removed. I still keep the `if` then reassign pattern as I think it's a bit easier to read (shorter). The true one-liner solution is to combine line 3773 as well, but it'd become a bit too crowded :)

PR #3536, opened 2025-10-22
Fixes bugs in Python argument conversion

schweitzpgi:
Argument conversion is done on code that is to run on the device. As such there shall never be any `std::vector<bool>` template specialization data values on the device. These changes make argument conversion portable across any `std::vector<T>`, enable the `std::vector<std::vector<bool>>` test, and otherwise fix the tests and conversion.

It is the expectation of the implementation that all arguments shall be marshaled correctly for operation on the device. This includes conversions prior to argument synthesis and kernel specialization.

Also changes the calls to `mlir::emitError()` to `throw std::runtiem_error()` which was already being used in some cases here for consistency.

### DIFF
--- a/python/cudaq/runtime/sample.py
+++ b/python/cudaq/runtime/sample.py
@@ -120,8 +120,9 @@ Returns:
     while counts.get_total_shots() < shots_count:
         kernel(*args)
         cudaq_runtime.resetExecutionContext()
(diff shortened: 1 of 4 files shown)


schweitzpgi:
Making this a draft until all the various places the python implementation needs to be updated can be tracked down.

PR #3537, opened 2025-10-22
Python language translation revision

bettinaheim:
## Summary of changes:

- Added additional validation and type conversions for arguments and return values when necessary
- Added proper error messages to indicate when a copy of a reference type is required
- Added proper error messages for assignments that are not supported
- Added a `copy` helper function for lists, np.arrays, tuples, and dataclasses to create deep and shallow copies

### Bug Fixes
- Fixes an issue where kernels would not be found in certain cases (due to analysis not adding them or dependencies not being correctly accumulated)
- Fixes incorrect errors when building certain nested expressions (e.g. constructing a list within a call)
- Fixed various issues with passing nested containers across device kernels
- Added some missing overloads for some gate invocations and kernel calls
- Fixed some inconsistencies for variable assignments
- Added support for chaining item and attribute access and fixed related issues with item assignments
- Fixed an issue where lists and arrays were not properly copied when a copy constructor was invoked
- Added a comprehensive error for some cases that are not yet fully supported and resulted in a segfault
- Fixed a bug that would cause Boolean operations with more than two operands to not be correctly evaluated
- Fixed an issue that would cause a crash when building a list from a range in some cases

The rest of this description contains a breakdown of the PR content with explanations for developers. 
The PR focuses on the Python bridge and its use of data types. It intentionally does not modify anything related to host-device data transfer, the symbol table, or the representation of callables and states, since the revisions by @schweitzpgi will majorly change that. 

## Changes to ast_bridge.py:

There are four main things revised
1. The value stack during IR construction
2. Loops
3. Calls
4. Assignments
 
### Value stack

The previous implementation was using a single deque to propagate values across node visits. This is problematic, since it looses the association of values with specific node. Correspondingly, the previous implementation would fail for (some) nested expressions (e.g. a list built inside a call expression). It was also highly problematic since a failure to produce a value would go undetected and a previously pushed value may be used instead, leading to incorrect code (either failing to compile, or worse, compiling but not matching the application code). 

This PR hence introduces a proper PyStack to propagate MLIR values across visits. For each node visit, a frame is pushed to that stack, and the new implementation validates that the correct number of values is produced by each node. To further make the detection of incorrect or incorrectly processed code more robust, each node now pushes a single value. 
The relevant pieces of code to look at are the PyStack class and the definition of visit. The override for generic_visit is removed and no longer used. Related changes:
- Calls to 'range' and 'enumerate' in visit_Call push a vector. 
- The processing of KrausChannels is removed from visit_Name and visit_Attribute and instead implemented directly inside `apply_noise` in visit_Call; the use of KrausChannels is only ever valid as part of a call (for both the previous and new implementation)

### Loops

In the previous implementation, all loops were indiscriminately marked as invariant. This was incorrect for the case of e.g. visit_For, where the for-loop may contain break statements. The new implementation defines a general helper for loop-creation (createForLoop), as well as createMonotonicForLoop and createInvariantForLoop calling into that. These are used in the implementation of visit_For, visit_While, visit_Compare (for 'In/NotIn' comparison), visit_List_Comp, in various Call expressions, as well as in migrateLists (explained further below). 

### Calls

The previous implementation did very little checks around the correctness of arguments to calls. This PR largely refactors the entire implementation of visit_Call to have most of the cases call into shared helper functions defined at the beginning of visit_Call. These helpers ensure that (with few exceptions that I haven't refactored) all arguments to calls are checked for correctness and type conversions are applied when appropriate. There are two general helpers that are also used in other parts of the code base; `changeOperandToType` and `__groupValues`. The former is responsible for all type conversions throughout the bridge. The latter is used to process a list of AST nodes and validate that we have an appropriate number of them. In addition to the nodes to process, it takes a `list[int | tuple[int, int]]` argument indicating how many values to expect and how to group them. See the doc comment for more details.

### Assignments

The changes to assignment make up the largest part of the changes (and tests). For one, the new implementation treats all of the following as assignments: assignments as part of visit_Assign and visit_AugAssign, the definition of function arguments, and the definition of loop iteration variables. This change also introduces a clear set of rules for the use of pointers. The outline given here only applies to the values produced to represent objects in the Python source code, not internal representations for data types (e.g. pointers to data arrays contained in a vector). The distinction is important since internal representation for certain data types should be opaque to the python bridge for the sake of encapsulation. 
In the new implementation, the only data type that is always passed by pointer is a State. I won't elaborate on states any further; the support for them is currently in a somewhat inconsistent state and needs to be reexamined after the merge of the Python compiler changes. I hence left it as is in this PR. All other data types are passed by value (both in the previous and in the new implementation); the paragraphs further below elaborate how the new implementation deals with reference types in Python. Aside from State objects (and internal representations), pointers are exclusively used to represent variables. When variables are created, a pointer to a stack slot is created and push to the symbol table. When variables are used the current value of the variable is loaded and pushed to the value stack. This leads to a clean and consistent handling of all value types in Python. They behave as they should and are not subject to any restrictions in their use. The subsequent paragraphs discuss python reference types and quantum types, for which we have to impose certain restrictions to ensure that any code that successfully compiles matches the expected Python behavior.

#### Quantum Types and Measurements

Quantum types (qubits, qvectors, quantum structs) and measurement results are stored as values - not pointers! - in the symbol table (both in the previous and in the new implementation). This necessarily requires that we impose restrictions regarding assignments to variables of these types. While these restrictions are likely desirable for quantum types, we could reexamine these restrictions once we introduce a proper type distinction between boolean values and measurements throughout the stack. As it is currently, storing them as values is needed to enable sampling with explicit measurements (to an extend - the current support is incomplete and full support requires the type distinction). 

The restriction for assignments to variables of these types specifically is that values of these types cannot be assigned to variables in a parent scope, whether directly or indirectly (meaning item assignment). A direct assignment is impossible to support with the current representation in the symbol table since there is no way to conditionally update the value in the parent scope depending on whether the child scope was executed. We could lift this restriction in a future version e.g. by leveraging phi nodes. An indirect assignment (i.e. assigning to an item of a vector or struct in the parent scope) would technically be possible for measurements but would lose the information of where the value came from - I left it as is and don't support that; to be reexamined in a future version. For quantum values, any item assignments are intentionally forbidden by design of the CUDA-Q language. Other than that, no restriction exists for assignments to variables in the same scope. 

#### Python Reference Types

Currently supported types within kernels that are reference types in Python are lists, numpy arrays, and dataclasses. 

NOTE: Much like States, Callables were left as is in this PR, with the expectation to reexamine them after the Python compiler changes were merged. Left as is specifically means they are currently stored as values (not pointers) in the symbol table and subject to the same restrictions for assignments that were discussed for vectors. 

*Lists and numpy arrays:*

Lists and numpy arrays share the same representation in the IR. Everything stated for lists should be taken to apply also for numpy arrays.
Lists are represented as `stdvec` objects in the IR. Their internal representation contains a pointer to a data array and an integer indicating their size. This ensures that even though we pass `stdvec` objects by value across function boundaries, they indeed follow reference behavior. There are two issues with the representation as is (not modified in this PR) that are left for consideration to revise in a future version: 
1. `stdvec` objects are also stored as values (not pointers) in the symbol table, and both their data pointer and their size is immutable as far as I saw (no IR expression to update the size of a `stdvec` without constructing a new one). As such, we have the similar issues as we have for quantum types when it comes to direct assignments to variables in the parent scope. Indirect assignments to items of variables in the parent scope do not suffer from this and work as expected. 
2. [only if the kernel returns lists]: The memory to store the data of an `stdvec` is allocated on the stack. When we return an `stdvec` we hence need to make sure to copy that memory to the heap. This copy is inconsistent with the behavior one would expect from python in the case when the returned list was in fact passed as argument (aside from the data in this case in fact being caller allocated). After careful consideration of all options, the best path seems to be to make sure that we keep track of vectors that come from arguments (see container item restrictions below) and give an error when we return one that was passed as argument (more details are in comments in code). In practice, I believe this is not encountered all that often and in most cases, the performance after optimization should be the same.

*Classical Dataclasses:*

In a future version, it may be nice to have a similar representation for dataclasses as we have for vectors (a value encapsulating a pointer). Keeping the representation as it was, and passing the classical `structs` as values across function boundaries, we have two main kinds/sources of restrictions we need to impose. These restrictions ensure that any valid code behaves as one would expect Python code to behave. 
1. We cannot assign an `lvalue` to a variable in the parent scope (directly or indirectly). This is because the current representation makes it impossible to ensure that the two handles (lvalue that is assigned and lvalue we are assigning to) indeed access the same data. For the same reason, we impose that a dataclass `lvalue` cannot be used as a container item (see container item restrictions below).
2. The new implementation imposes additional checks for dataclasses passed as function arguments to force that the copy that we effectively do as part of assigning it to an argument is explicit in the Python code. Specifically, we require an explicit `.copy` in the python code when directly or indirectly assigning a dataclass function argument to a local variable. 

NOTE: Tuples are value types in Python. While they stare the same representation with dataclasses as `structs` in the IR, we forbid assignments to tuple items (matching python behavior), but/therefore do not otherwise need to impose any restrictions on them.

*Restrictions on Container Items:*

For the reasons outlined above, and in more detail in code comments and tests, we impose the following restrictions for any container items:
- Dataclass `lvalues` cannot be used as container items - an explicit `.copy` must be made to store them in the container.
- [only if the kernel returns lists]: Lists passed as function arguments (or as an function argument item) cannot be used as container items - an explicit `.copy` must be made to store them in the container.
- Container items cannot be pointers - i.e. containers cannot contain States (since this is the only type where we represent a Python value as a pointer). This restriction was already made in the previous implementation and I kept any behavior related to State and Callables unmodified.

Currently supported containers in the Python bridge are tuples, dataclasses and lists/numpy.arrays. The restrictions above are enforced by a common helper function (`__validate_container_entry`) that is called whenever a container is created. I.e. it is called during item assignment, list comprehension, and constructor calls, copy constructors, or literals of containers.

The most relevant pieces of code to look at to understand the changes with regards to assignments are visit_Assign, visit_Name, and visit_Return, as well as the tests in test_assignments.py.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -454,12 +454,10 @@ jobs:
               echo "### Setting up OQC account" >> $GITHUB_STEP_SUMMARY
               echo "::add-mask::${{ secrets.OQC_URL }}"
               echo "::add-mask::${{ secrets.BACKEND_LOGIN_EMAIL }}"
-              echo "::add-mask::${{ secrets.OQC_PASSWORD }}"
               echo "::add-mask::${{ secrets.OQC_DEVICE }}"
               echo "::add-mask::${{ secrets.OQC_AUTH_TOKEN }}"
               echo "OQC_URL=${{ secrets.OQC_URL }}" >> $GITHUB_ENV
               echo "OQC_EMAIL=${{ secrets.BACKEND_LOGIN_EMAIL }}" >> $GITHUB_ENV
-              echo "OQC_PASSWORD=${{ secrets.OQC_PASSWORD }}" >> $GITHUB_ENV
               echo "OQC_DEVICE=${{ secrets.OQC_DEVICE }}" >> $GITHUB_ENV
               echo "OQC_AUTH_TOKEN=${{ secrets.OQC_AUTH_TOKEN }}" >> $GITHUB_ENV
               ;;
--- a/lib/Optimizer/Transforms/DecompositionPatterns.cpp
+++ b/lib/Optimizer/Transforms/DecompositionPatterns.cpp
@@ -6,6 +6,22 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
+/**
+ * This file contains the decomposition patterns that match single gates and
+ * decompose them into a sequence of other gates.
+ *
+ * Each pattern definition contains 3 elements:
+ * 1. The pattern itself, which defines what ops to match and how to replace
+ * them. It must inherit from DecompositionPattern<PatternType, Op>.
+ * 2. The pattern type, which contains the pattern metadata. It must inherit
+ * from DecompositionPatternType.
+ * 3. A call to the CUDAQ_REGISTER_TYPE macro to register the pattern in the
+ * registry.
+ *
+ * Writing 2 and 3 manually is a bit verbose. The REGISTER_DECOMPOSITION_PATTERN
+ * macro can be used for this purpose instead.
+ */
+
 #include "DecompositionPatterns.h"
 #include "cudaq/Optimizer/Builder/Factory.h"
 #include "cudaq/Optimizer/Dialect/CC/CCOps.h"
@@ -13,8 +29,17 @@
 #include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Rewrite/FrozenRewritePatternSet.h"
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/ADT/StringMap.h>
+#include <llvm/ADT/StringRef.h>
+#include <llvm/Support/Casting.h>
+#include <llvm/Support/Error.h>
+#include <llvm/Support/TypeName.h>
+#include <memory>
 
 using namespace mlir;
+
+LLVM_INSTANTIATE_REGISTRY(cudaq::DecompositionPatternType::RegistryType)
 
 namespace {
 
@@ -283,6 +308,36 @@ LogicalResult checkAndExtractControls(quake::OperatorInterface op,
   return success();
 }
 
+// From here on, we define the decomposition patterns ==========================
+
+/// Macro to register a decomposition pattern with its metadata
+/// Usage: REGISTER_DECOMPOSITION_PATTERN(PatternName, "source_op", "target1",
+/// "target2", ...)
+/// where "source_op" is the operation that the pattern matches and
+/// {"target1", "target2", ...} are the operations that the pattern may produce.
+#define REGISTER_DECOMPOSITION_PATTERN(PATTERN, SOURCE_OP, ...)                \
+  struct PATTERN##Type : public cudaq::DecompositionPatternType {              \
+    using cudaq::DecompositionPatternType::DecompositionPatternType;           \
+    llvm::StringRef getSourceOp() const override { return SOURCE_OP; }         \
+    llvm::ArrayRef<llvm::StringRef> getTargetOps() const override {            \
+      static constexpr llvm::StringRef ops[] = {__VA_ARGS__};                  \
+      return ops;                                                              \
+    }                                                                          \
+    llvm::StringRef getPatternName() const override { return #PATTERN; }       \
+    std::unique_ptr<mlir::RewritePattern>                                      \
+    create(mlir::MLIRContext *context,                                         \
+           mlir::PatternBenefit benefit = 1) const override {                  \
+      std::unique_ptr<mlir::RewritePattern> pattern =                          \
+          RewritePattern::create<PATTERN>(context, benefit);                   \
+      return pattern;                                                          \
+    }                                                                          \
+  };                                                                           \
+  CUDAQ_REGISTER_TYPE(cudaq::DecompositionPatternType, PATTERN##Type, PATTERN)
+
+// TODO: The decomposition patterns "SToR1", "TToR1", "R1ToU3", "U3ToRotations"
+// can handle arbitrary number of controls, but currently metadata cannot
+// capture this. The pattern types therefore only advertise them for 0 controls.
+
 //===----------------------------------------------------------------------===//
 // HOp decompositions
 //===----------------------------------------------------------------------===//
@@ -291,10 +346,14 @@ LogicalResult checkAndExtractControls(quake::OperatorInterface op,
 // ───────────────────────────────────
 // quake.phased_rx(π/2, π/2) target
 // quake.phased_rx(π, 0) target
-struct HToPhasedRx : public OpRewritePattern<quake::HOp> {
-  using OpRewritePattern<quake::HOp>::OpRewritePattern;
 
-  void initialize() { setDebugName("HToPhasedRx"); }
+struct HToPhasedRxType; // forward declare the pattern type, defined in the
+                        // macro below
+struct HToPhasedRx
+    : public cudaq::DecompositionPattern<HToPhasedRxType, quake::HOp> {
+
+  using cudaq::DecompositionPattern<HToPhasedRxType,
+                                    quake::HOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::HOp op,
                                 PatternRewriter &rewriter) const override {
@@ -323,14 +382,18 @@ struct HToPhasedRx : public OpRewritePattern<quake::HOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(HToPhasedRx, "h", "phased_rx");
 
 // quake.exp_pauli(theta) target pauliWord
 // ───────────────────────────────────
 // Basis change operations, cnots, rz(theta), adjoint basis change
-struct ExpPauliDecomposition : public OpRewritePattern<quake::ExpPauliOp> {
-  using OpRewritePattern::OpRewritePattern;
-
-  void initialize() { setDebugName("ExpPauliDecomposition"); }
+struct ExpPauliDecompositionType; // forward declare the pattern type, defined
+                                  // in the macro below
+struct ExpPauliDecomposition
+    : public cudaq::DecompositionPattern<ExpPauliDecompositionType,
+                                         quake::ExpPauliOp> {
+  using cudaq::DecompositionPattern<ExpPauliDecompositionType,
+                                    quake::ExpPauliOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::ExpPauliOp expPauliOp,
                                 PatternRewriter &rewriter) const override {
@@ -496,12 +559,18 @@ struct ExpPauliDecomposition : public OpRewritePattern<quake::ExpPauliOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(ExpPauliDecomposition, "exp_pauli", "rx", "h",
+                               "x(1)", "rz");
 
 // Naive mapping of R1 to Rz, ignoring the global phase.
 // This is only expected to work with full inlining and
 // quake apply specialization.
-struct R1ToRz : public OpRewritePattern<quake::R1Op> {
-  using OpRewritePattern::OpRewritePattern;
+struct R1ToRzType; // forward declare the pattern type, defined in the macro
+                   // below
+struct R1ToRz : public cudaq::DecompositionPattern<R1ToRzType, quake::R1Op> {
+  using cudaq::DecompositionPattern<R1ToRzType,
+                                    quake::R1Op>::DecompositionPattern;
+
   LogicalResult matchAndRewrite(quake::R1Op r1Op,
                                 PatternRewriter &rewriter) const override {
     if (!r1Op.getControls().empty())
@@ -513,15 +582,17 @@ struct R1ToRz : public OpRewritePattern<quake::R1Op> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(R1ToRz, "r1", "rz");
 
 // Naive mapping of R1 to U3
 // quake.r1(λ) [control] target
 // ───────────────────────────────────
 // quake.u3(0, 0, λ) [control] target
-struct R1ToU3 : public OpRewritePattern<quake::R1Op> {
-  using OpRewritePattern<quake::R1Op>::OpRewritePattern;
-
-  void initialize() { setDebugName("R1ToU3"); }
+struct R1ToU3Type; // forward declare the pattern type, defined in the macro
+                   // below
+struct R1ToU3 : public cudaq::DecompositionPattern<R1ToU3Type, quake::R1Op> {
+  using cudaq::DecompositionPattern<R1ToU3Type,
+                                    quake::R1Op>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::R1Op r1Op,
                                 PatternRewriter &rewriter) const override {
@@ -533,14 +604,17 @@ struct R1ToU3 : public OpRewritePattern<quake::R1Op> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(R1ToU3, "r1", "u3");
 
 // quake.r1<adj> (θ) target
 // ─────────────────────────────────
 // quake.r1(-θ) target
-struct R1AdjToR1 : public OpRewritePattern<quake::R1Op> {
-  using OpRewritePattern<quake::R1Op>::OpRewritePattern;
-
-  void initialize() { setDebugName("R1AdjToR1"); }
+struct R1AdjToR1Type; // forward declare the pattern type, defined in the macro
+                      // below
+struct R1AdjToR1
+    : public cudaq::DecompositionPattern<R1AdjToR1Type, quake::R1Op> {
+  using cudaq::DecompositionPattern<R1AdjToR1Type,
+                                    quake::R1Op>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::R1Op op,
                                 PatternRewriter &rewriter) const override {
@@ -567,16 +641,19 @@ struct R1AdjToR1 : public OpRewritePattern<quake::R1Op> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(R1AdjToR1, "r1", "r1");
 
 // quake.swap a, b
 // ───────────────────────────────────
 // quake.cnot b, a;
 // quake.cnot a, b;
 // quake.cnot b, a;
-struct SwapToCX : public OpRewritePattern<quake::SwapOp> {
-  using OpRewritePattern<quake::SwapOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("SwapToCX"); }
+struct SwapToCXType; // forward declare the pattern type, defined in the macro
+                     // below
+struct SwapToCX
+    : public cudaq::DecompositionPattern<SwapToCXType, quake::SwapOp> {
+  using cudaq::DecompositionPattern<SwapToCXType,
+                                    quake::SwapOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::SwapOp op,
                                 PatternRewriter &rewriter) const override {
@@ -595,6 +672,7 @@ struct SwapToCX : public OpRewritePattern<quake::SwapOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(SwapToCX, "swap", "x(1)");
 
 // quake.h control, target
 // ───────────────────────────────────
@@ -605,10 +683,11 @@ struct SwapToCX : public OpRewritePattern<quake::SwapOp> {
 // quake.t<adj> target;
 // quake.h target;
 // quake.s<adj> target;
-struct CHToCX : public OpRewritePattern<quake::HOp> {
-  using OpRewritePattern<quake::HOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("CHToCX"); }
+struct CHToCXType; // forward declare the pattern type, defined in the macro
+                   // below
+struct CHToCX : public cudaq::DecompositionPattern<CHToCXType, quake::HOp> {
+  using cudaq::DecompositionPattern<CHToCXType,
+                                    quake::HOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::HOp op,
                                 PatternRewriter &rewriter) const override {
@@ -634,6 +713,7 @@ struct CHToCX : public OpRewritePattern<quake::HOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(CHToCX, "h(1)", "s", "h", "t", "x(1)");
 
 //===----------------------------------------------------------------------===//
 // SOp decompositions
@@ -644,10 +724,12 @@ struct CHToCX : public OpRewritePattern<quake::HOp> {
 // phased_rx(π/2, 0) target
 // phased_rx(-π/2, π/2) target
 // phased_rx(-π/2, 0) target
-struct SToPhasedRx : public OpRewritePattern<quake::SOp> {
-  using OpRewritePattern<quake::SOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("SToPhasedRx"); }
+struct SToPhasedRxType; // forward declare the pattern type, defined in the
+                        // macro below
+struct SToPhasedRx
+    : public cudaq::DecompositionPattern<SToPhasedRxType, quake::SOp> {
+  using cudaq::DecompositionPattern<SToPhasedRxType,
+                                    quake::SOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::SOp op,
                                 PatternRewriter &rewriter) const override {
@@ -681,6 +763,7 @@ struct SToPhasedRx : public OpRewritePattern<quake::SOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(SToPhasedRx, "s", "phased_rx");
 
 // quake.s [control] target
 // ────────────────────────────────────
@@ -688,10 +771,11 @@ struct SToPhasedRx : public OpRewritePattern<quake::SOp> {
 //
 // Adding this gate equivalence will enable further decomposition via other
 // patterns such as controlled-r1 to cnot.
-struct SToR1 : public OpRewritePattern<quake::SOp> {
-  using OpRewritePattern<quake::SOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("SToR1"); }
+struct SToR1Type; // forward declare the pattern type, defined in the macro
+                  // below
+struct SToR1 : public cudaq::DecompositionPattern<SToR1Type, quake::SOp> {
+  using cudaq::DecompositionPattern<SToR1Type,
+                                    quake::SOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::SOp op,
                                 PatternRewriter &rewriter) const override {
@@ -710,6 +794,7 @@ struct SToR1 : public OpRewritePattern<quake::SOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(SToR1, "s", "r1");
 
 //===----------------------------------------------------------------------===//
 // TOp decompositions
@@ -720,10 +805,12 @@ struct SToR1 : public OpRewritePattern<quake::SOp> {
 // quake.phased_rx(π/2, 0) target
 // quake.phased_rx(-π/4, π/2) target
 // quake.phased_rx(-π/2, 0) target
-struct TToPhasedRx : public OpRewritePattern<quake::TOp> {
-  using OpRewritePattern<quake::TOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("TToPhasedRx"); }
+struct TToPhasedRxType; // forward declare the pattern type, defined in the
+                        // macro below
+struct TToPhasedRx
+    : public cudaq::DecompositionPattern<TToPhasedRxType, quake::TOp> {
+  using cudaq::DecompositionPattern<TToPhasedRxType,
+                                    quake::TOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::TOp op,
                                 PatternRewriter &rewriter) const override {
@@ -758,6 +845,7 @@ struct TToPhasedRx : public OpRewritePattern<quake::TOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(TToPhasedRx, "t", "phased_rx");
 
 // quake.t [control] target
 // ────────────────────────────────────
@@ -765,10 +853,11 @@ struct TToPhasedRx : public OpRewritePattern<quake::TOp> {
 //
 // Adding this gate equivalence will enable further decomposition via other
 // patterns such as controlled-r1 to cnot.
-struct TToR1 : public OpRewritePattern<quake::TOp> {
-  using OpRewritePattern<quake::TOp>::OpRewritePattern;
-
-  void initialize() { setDebugName("TToR1"); }
+struct TToR1Type; // forward declare the pattern type, defined in the macro
+                  // below
+struct TToR1 : public cudaq::DecompositionPattern<TToR1Type, quake::TOp> {
+  using cudaq::DecompositionPattern<TToR1Type,
+                                    quake::TOp>::DecompositionPattern;
 
   LogicalResult matchAndRewrite(quake::TOp op,
                                 PatternRewriter &rewriter) const override {
@@ -786,6 +875,7 @@ struct TToR1 : public OpRewritePattern<quake::TOp> {
     return success();
   }
 };
+REGISTER_DECOMPOSITION_PATTERN(TToR1, "t", "r1");
 
 //===----------------------------------------------------------------------===//
 // XOp decompositions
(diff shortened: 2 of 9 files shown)


python/tests/mlir/test_output_qir.py
bettinaheim, line 1: This file/test was entirely redundant; it had the exact same content as the `test_output_translation_qir` (possibly a merge gone wrong and duplicating it at some point?)

python/tests/mlir/test_output_translate_openqasm.py
bettinaheim, line 1: I validated that the code both before an after my changes in this test was and is correct; the only difference arises due to  a change of order in the control qubits. Specifically (only) the output for `z.ctrl(ctrl1, ctrl2, target)` changed. The order of control qubits is entirely irrelevant for correctness. After my changes, the order of control qubits matches the source code, previously the order was reverted.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1422,18 +1524,11 @@ def process_assignment(target, value):
...
+
                 if not cc.PointerType.isinstance(ptrVal.type):
                     self.emitFatalError(
                         "Invalid CUDA-Q subscript assignment, variable must be a pointer.",
                         node)
-                # See if this is a pointer to an array, if so cast it
-                # to a pointer on the array type
-                ptrEleType = cc.PointerType.getElementType(ptrVal.type)
-                if cc.ArrayType.isinstance(ptrEleType):
-                    ptrVal = cc.CastOp(
```
bettinaheim, line 1461: This seems very odd to me - I need to check/test this further. The visit of the target in this case should yield a pointer to the right array element. That pointer should be of the correct type (a ptr of array element type). If it is not, then something seems wrong with our code.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1605,39 +1691,13 @@ def process_potential_ptr_types(value):
...
-        General calls to previously seen CUDA-Q kernels are supported. By this
-        we mean that an kernel can not be invoked from a kernel unless it was
-        defined before the current kernel.  Kernels can also be reversed or
-        controlled with `cudaq.adjoint(kernel, ...)` and `cudaq.control(kernel,
-        ...)`.
-
-        Finally, general operation modifiers are supported, specifically
-        `OPERATION.adj` and `OPERATION.ctrl` for adjoint and control synthesis
-        of the operation.
+        Map a Python Call operation to equivalent MLIR. This method handles
```
bettinaheim, line 2163: No point in having the doc comment be too specific about the order of implementation; this comment is outdated and would otherwise be outdated again very quickly.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4062,47 +3945,53 @@ def visit_BoolOp(self, node):
...
             # Note we want any `mz(q)` calls to push their
             # result value to the stack, so we set a non-None
             # variable name here.
             self.currentAssignVariableName = ''
             self.visit(node.values[0])
-            lhs = self.popValue()
-            zero = self.getConstantInt(0, IntegerType(lhs.type).width)
+            cond = self.__arithmetic_to_bool(self.popValue())
+
+            def process_boolean_op(prior, values):
```
bettinaheim, line 4335: The previous logic was plain wrong; it would simply ignore any values past the first two. To deal with all operands properly, we need this to be recursive.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4735,11 +4614,11 @@ def visit_Name(self, node):
                     return
                 loaded = cc.LoadOp(value).result
                 self.pushValue(loaded)
-            elif cc.CallableType.isinstance(
```
bettinaheim, line 4841: Removed the None return here for `cc.CallableType` that is not a `BlockArgument`. This case is no different than when we have a lambda expression in the code where we also push a value to the stack.

schweitzpgi:
> cc.Arrays are now exclusively used only as the underlying data of vectors and for strings; they are hence never valid data types for values passed between nodes

Not sure what this means. The `!cc.array` type is a perfectly valid type for a value. Values may have this type and storage locations may hold values of this type.

I also don't understand the prohibitions on productions. Some productions may not produce a value at all. Some productions may produce vectors of values. Are you saying every visit will produce 1 wrapped binding value? Even if there isn't one?

python/tests/kernel/test_run_kernel.py
```diff
@@ -311,7 +311,7 @@ def incrementer(i: int) -> int:
     def kernel_with_list_arg(arg: list[int]) -> list[int]:
         result = arg
         for i in result:
-            incrementer(i)
+            incrementer(i) # FIXME: THIS IS NOT HOW PYTHON SHOULD BEHAVE...!
```
khalatepradnya: The intent of this test was to have nested functions with list arguments being passed along. The function `incrementer` is called but it is not updating the values of the `result` (and is not expected to update the values). The final result is unchanged from original input arguments. 
The leftover wrong comment (while trying to come up with a test scenario) on line# 327,  `# 4+1 + 5+1 + 6+1 = 15` (it would have been `18`) should be deleted or updated to `# 4 + 5 + 6 = 15`. 
My bad. 😔

bettinaheim: Oh wow, I didn't do the math! 🤦 
Ok, reading the comments that had me very puzzed about how that would even happen that the result vector is updated. One less thing to worry about.

bettinaheim:
> > cc.Arrays are now exclusively used only as the underlying data of vectors and for strings; they are hence never valid data types for values passed between nodes
> 
> Not sure what this means. The `!cc.array` type is a perfectly valid type for a value. Values may have this type and storage locations may hold values of this type.
> 
> I also don't understand the prohibitions on productions. Some productions may not produce a value at all. Some productions may produce vectors of values. Are you saying every visit will produce 1 wrapped binding value? Even if there isn't one?

I meant they are not used (currently) to represent a value produced by any of the visits. If we make use of arrays a the representation for some node(s) in the AST, then a bunch of other nodes need to be updated to consistently deal with arrays. Right now, everything checks for stdvecttype and qvectypes. 
The only change that this PR does related to array usage is that it changes the visit for a call to 'range' and 'enumerate' to return an stdvect instead of an array and a size separately. The array that was returned wasn't statically sized in any case, and the additional vector init call seems like something that would be properly optimized by the compiler when appropriate.

Re produced values: Every visit produces either 0 or 1 value on the value stack. We have to allow for 0 because there are plenty of things that don't/can't produce a value (e.g. we don't have a representation for None, so a call to a void function will not produce a value). A produced value can be an array or a tuple, but there isn't a reason that there should be multiple separate values produced - if we do that, we are really abandoning any safety that we get from/need for the strong typing. I am not sure what you mean by "producing a binding". Each node/visit may produce a large amount of MLIR code. However, it has one final result value. That result can be of any type, but there aren't any expressions in the language that would result in having more than one "result" (the result of a call is it's return value, the result of a boolean operation is a boolean etc).

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4233,31 +4109,41 @@ def compare_equality(item1, item2):
...
-                    cc.ArrayType.isinstance(right.type)):
+            if not cc.StdvecType.isinstance(right.type):
                 self.emitFatalError(
                     "Right operand must be a list/vector for 'in' comparison")
+            vectSize = cc.StdvecSizeOp(self.getIntegerType(), right).result
 
             # Loop setup
             i1_type = self.getIntegerType(1)
+            trueVal = self.getConstantInt(1, 1)
             accumulator = cc.AllocaOp(cc.PointerType.get(i1_type),
```
bettinaheim, line 4514: @schweitzpgi Is this the proper representation that the accumulator is not passed as argument across the blocks? This is how it was before, but I am not entirely clear regarding whether everything used inside the loop should be passed as loop argument or whether the compiler will rewrite the loop as needed.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4784,43 +4456,33 @@ def visit_Name(self, node):
...
-                mlirValCreator = lambda: self.getConstantComplex(value,
-                                                                 width=32)
+                mlirVal = self.getConstantComplex(value, width=32)
             elif isinstance(value, complexType) or isinstance(
                     value, np.complex128):
-                mlirValCreator = lambda: self.getConstantComplex(value,
-                                                                 width=64)
-
-            if mlirValCreator != None:
-                with InsertionPoint.at_block_begin(self.entry):
```
bettinaheim, line 4917: Creating an alloc here to get a writable pointer is both unnecessary and plain wrong. Captured variables cannot be updated, so we don't need/want them pretending that they can. These are constant values within the kernel and should be treated as such.

python/tests/kernel/test_kernel_features.py
```diff
@@ -999,22 +1117,6 @@ def canCaptureList():
                       atol=1e-3)
 
 
-def test_capture_disallow_change_variable():
```
bettinaheim, line 1002: This test just move into test_assignments

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4784,68 +4592,47 @@ def visit_Name(self, node):
...
-                return
+            else:
 
-            errorType = type(value).__name__
-            if (isinstance(value, list)):
-                errorType = f"{errorType}[{type(value[0]).__name__}]"
-
-            try:
-                if issubclass(value, cudaq_runtime.KrausChannel):
-                    # Here we have a KrausChannel as part of the AST.  We want
```
bettinaheim, line 4935: Kraus channels are not generally valid variable types in CUDA-Q. They can only be used specifically in calls to apply noise. Hence, this logic migrated to exactly that point to make sure we get a proper error if you try to use them anywhere else.

schweitzpgi (commented):
This is just too big to review properly. I think it needs to be split up into multiple focused PRs.

lib/Optimizer/CodeGen/CCToLLVM.cpp
```diff
@@ -524,8 +524,7 @@ class SizeOfOpPattern : public ConvertOpToLLVMPattern<cudaq::cc::SizeOfOp> {
                   ConversionPatternRewriter &rewriter) const override {
     auto inputTy = sizeOfOp.getInputType();
     auto resultTy = sizeOfOp.getType();
-    if (quake::isQuakeType(inputTy) || cudaq::cc::isDynamicType(inputTy)) {
-      // Types that cannot be reified produce the poison op.
+    if (quake::isQuakeType(inputTy)) {
```
schweitzpgi: This change appears to be incorrect. Can you explain how these changes can statically determine the size of an object that's size is only known at runtime?

python/cudaq/kernel/ast_bridge.py
schweitzpgi: The bridge should create temporary variables (properly scoped) and use load and store operations to update the variable.

The compiler knows how to lift variables to SSA values. I would consider it a (lurking) bug to try and reimplement that functionality in the bridge.

lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1054,6 +1054,32 @@ class GenerateKernelExecution
           thunk = genThunkFunction(loc, builder, classNameStr, structTy, funcTy,
                                    funcOp);
 
+          // The `argsCreator` function doesn't properly deal with structs
+          // that contain dynamic types. Since we want to revise that in a 
+          // future version, I am opting to simply give a comprehensive enough
+          // error here rather than fixing that. If we don't fail here, the
+          // invocation of the `argsCreator` as it is will segfault.
```
schweitzpgi: This code doesn't belong here.

python/cudaq/kernel/analysis.py
```diff
@@ -110,9 +110,11 @@ def visit_Call(self, node):
                              'attr') and node.func.attr in globalAstRegistry:
                     self.depKernels[node.func.attr] = globalAstRegistry[
                         node.func.attr]
-                elif node.func.value.id == 'cudaq' and node.func.attr in [
-                        'control', 'adjoint'
-                ] and node.args[0].id in globalAstRegistry:
+                elif isinstance(node.func.value, ast.Name) and \
```
schweitzpgi: Do not use backslashes.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -104,6 +104,90 @@ def __init__(self, *args, **kwargs):
         RuntimeError.__init__(self, *args, **kwargs)
 
 
+class PyStack(object):
```
schweitzpgi, line 108: I expect given the amount of churn in this file that this will all need to be redone for the python rewrite.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -1,1140 +1,1164 @@
...
-              "cudaq::builder failed to get kernelReg function.");
-        }
-
-        return py::capsule(*funcPtr);
-      },
-      "JIT compile and return the C function pointer for the FuncOp of given "
-      "name.");
-}
-} // namespace cudaq
+/*******************************************************************************
```
schweitzpgi: Why is every single line of this file changed? How are we supposed to review this?

python/tests/mlir/bug_1777.py
```diff
@@ -1,80 +1,79 @@
...
-# CHECK:           } do {
-# CHECK:           ^bb0(%[[VAL_10:.*]]: i64):
-# CHECK:             %[[VAL_11:.*]] = quake.extract_ref %[[VAL_5]]{{\[}}%[[VAL_10]]] : (!quake.veq<2>, i64) -> !quake.ref
-# CHECK:             %[[VAL_12:.*]] = quake.mz %[[VAL_11]] name "res" : (!quake.ref) -> !quake.measure
-# CHECK:             %[[VAL_13:.*]] = quake.discriminate %[[VAL_12]] : (!quake.measure) -> i1
-# CHECK:             cc.store %[[VAL_13]], %[[VAL_6]] : !cc.ptr<i1>
-# CHECK:             %[[VAL_14:.*]] = arith.cmpi eq, %[[VAL_13]], %[[VAL_0]] : i1
-# CHECK:             cc.if(%[[VAL_14]]) {
-# CHECK:               %[[VAL_15:.*]] = quake.mz %[[VAL_5]] name "inner_mz" : (!quake.veq<2>) -> !cc.stdvec<!quake.measure>
-# CHECK:               %[[VAL_16:.*]] = quake.discriminate %[[VAL_15]] : (!cc.stdvec<!quake.measure>) -> !cc.stdvec<i1>
```
bettinaheim, line 55: Note that we still generate the discriminate in the AST bridge. However, the assignment previously did some odd things where it allocated a new stack slot, stored the value there, and then pushed the value (not the stack slot) to the symbol table regardless of whether the value existed in the parent scope. I have no idea why it created a stack slot that it then essentially ignored. The change to treat it appropriately during assignment causes the alloca and store to not be produced, and then the discriminate gets eliminated by the based pass manager passes done right after the bridge creates the module.

python/tests/mlir/bug_1775.py
```diff
@@ -1,65 +1,61 @@
...
-
-# CHECK-LABEL:   func.func @__nvqpp__mlirgen__test() attributes {"cudaq-entrypoint", "cudaq-kernel", qubitMeasurementFeedback = true} {
-# CHECK:           %[[VAL_0:.*]] = arith.constant true
-# CHECK:           %[[VAL_1:.*]] = quake.alloca !quake.ref
-# CHECK:           %[[VAL_2:.*]] = quake.mz %[[VAL_1]] name "res" : (!quake.ref) -> !quake.measure
-# CHECK:           %[[VAL_3:.*]] = quake.discriminate %[[VAL_2]] : (!quake.measure) -> i1
-# CHECK:           quake.h %[[VAL_1]] : (!quake.ref) -> ()
-# CHECK:           %[[VAL_4:.*]] = quake.mz %[[VAL_1]] name "res" : (!quake.ref) -> !quake.measure
-# CHECK:           %[[VAL_5:.*]] = quake.discriminate %[[VAL_4]] : (!quake.measure) -> i1
-# CHECK:           %[[VAL_6:.*]] = cc.alloca i1
```
bettinaheim, line 55: see comment on bug_1777

lib/Optimizer/Transforms/GenKernelExecution.cpp
schweitzpgi: I'm also not sure what this is doing really. We have long supported structs with vector elements as evidenced by the [tests as can be seen here.](https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/Kernel/signature-2.cpp#L102C1-L105C3)

python/cudaq/kernel/ast_bridge.py
```diff
@@ -461,15 +606,21 @@ def pushValue(self, value):
...
         """
-        val = self.valueStack.pop()
+        val = self.valueStack.popValue()
         self.debug_msg(lambda: f'pop {val}')
         return val
+    
+    def popAllValues(self, expectedNumVals):
+        values = [self.popValue() for _ in range(self.valueStack.currentNumValues)]
+        if len(values) != expectedNumVals:
+            self.emitFatalError("processing error - no valid value was created", self.currentNode)
```
1tnguyen: nit: the error message here is not very clear. I think it should mention the mismatch between the number of values on the stack vs. the expected.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -629,62 +777,65 @@ def getStructMemberIdx(self, memberName, structTy):
             )
         return structIdx, mlirTypeFromPyType(userType[memberName], self.ctx)
 
+    def __copyStructAndConvertElements(self, struct, expectedTy = None, conversion=None):
```
1tnguyen: nit: a code comment here, especially for the `conversion` callback argument, might be helpful.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1046,92 +1171,167 @@ def __processRangeLoopIterationBounds(self, argumentNodes):
...
+        Integer values have to be positive or -1, where -1
+        indicates that any number of values is acceptable.
+        Tuples of two integers (min, max) indicate that any number
+        of values in [min, max] is acceptable.
+        The list may only contain at most one negative integer or
+        tuple (enforced via assert only).
+
+        Emits a fatal error if any of the given `pyvals` did not
+        generate a value. Emits a fatal error if there are too
+        too many or too few values to satisfy the requested grouping.
```
1tnguyen: Typo: "too too many"

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1046,92 +1171,167 @@ def __processRangeLoopIterationBounds(self, argumentNodes):
...
+        '''
+
+        rootVal = self.__get_root_value(pyVal)
+        assert rootVal or not self.isFunctionArgument(mlirVal)
+
+        if cc.PointerType.isinstance(mlirVal.type):
+            # We do not allow to create container that contain pointers.
+            valTy = cc.PointerType.getElementType(mlirVal.type)
+            assert cc.StateType.isinstance(valTy)
+            if cc.StateType.isinstance(valTy):
```
1tnguyen, line 1426: I'm a bit confused about the intent of this `assert`.

python/tests/kernel/test_control_negations.py
```diff
@@ -1,281 +1,256 @@
...
-        cudaq.sample(control_simple_gate)
-    assert "unary operator ~ is only supported for values of type qubit" in str(
-        e.value)
-
-
-# leave for gdb debugging
-if __name__ == "__main__":
-    loc = os.path.abspath(__file__)
-    pytest.main([loc, "-rP"])
+# ============================================================================ #
```
1tnguyen: It seems like the GitHub diff UI is choking here as well.

lib/Optimizer/CodeGen/CCToLLVM.cpp
bettinaheim: Maybe we need a more fine-grained check here. Is dynamic fails for arrays of unknown size (I believe sizeof would need to fail in this case as well), span-like types, and any structs that contain these. 
Span-like types are vectors and char spans. For vectors, sizeof should not fail, since the size of the vector is perfectly known (three pointers) - not sure about char spans and how they are used. 
isDynamicType is also used in a bunch of contexts, however, for which vectors indeed need to be considered dynamic. I can create a new helper isDynamicallySizedType that indeed reflects when the size of the type itself is not known at compile time.

lib/Optimizer/Transforms/GenKernelExecution.cpp
bettinaheim: Interesting. The scenario segfaults for python - in that case, likely an issue with the python host-device data conversion. I'll remove it from here and figure out where a python specific location for this check might be.

python/cudaq/kernel/ast_bridge.py
bettinaheim: It's not too bad. Happy to port it.

python/cudaq/kernel/ast_bridge.py
bettinaheim: The intent is to reflect that the current implementation expects that the only type that is ever represented as a pointer is StateType. If we represent other values (not variables!) as pointers, this implementation (and the checks during assignment) probably need to be updated. As it is in the current implementation, we do not allow to create any containers that store pointers. This was already the case in the old implementation.

lib/Optimizer/Transforms/GenKernelExecution.cpp
bettinaheim: I looked more carefully. I think the argument conversion is just not correct for python structs. I made it fail for python only.

bettinaheim:
> /ok to test 639ebf2

bettinaheim:
> /ok to test eeea99a

bettinaheim:
> /ok to test 8de65f1

lib/Frontend/nvqpp/ConvertStmt.cpp
```diff
@@ -359,6 +359,9 @@ bool QuakeBridgeVisitor::VisitReturnStmt(clang::ReturnStmt *x) {
       if (!cudaq::cc::isDynamicType(eleTy))
         tySize = irb.getByteSizeOfType(loc, eleTy);
       if (!tySize) {
+        // TODO: we need to recursively create copies of all
+        // dynamic memory used within the type. See the
+        // implementation of `visit_Return` in the Python bride.
```
lmondada:
```suggestion
        // implementation of `visit_Return` in the Python bridge.
```

lmondada:
I agree that `.copy()` is a very useful shorthand. It also feels the "most natural" to me.

Perhaps unfortunately, the "most pythonic" would be to support the use of the `copy` and `deepcopy` free functions that can be imported from the `copy` module.

```python
import copy

v = ['a', 'b']
deep_v = ['a', ['a', 'b']]
w = copy.copy(v)
deep_w = copy.deepcopy(deep_v)
```

Given that we are parsing the AST, the fact that these functions are imported from the copy module makes this syntax an annoying complication, given the different ways these functions could be imported and used:
1.  `import copy` followed by `copy.copy` or `copy.deepcopy`
2. `from copy import copy, deepcopy`, followed by the unprefixed `copy(v)` or `deepcopy(v)`
3. `from copy import *`
4. worst of all, renaming: `from copy import copy as shallow_copy`

Supporting just 1. seems the simplest thing to do and is reasonable (even without checking for existence of the import). However, leaning into this Python convention without supporting at least 2. (which is how I'd use it) is likely to surprise users...

As a different point to consider, `numpy` supports both the [free function](https://numpy.org/doc/stable/reference/generated/numpy.copy.html) as well as providing a [`.copy()` method](https://numpy.org/doc/2.1/reference/generated/numpy.ndarray.copy.html#numpy.ndarray.copy):
- on the one hand, the numpy [docs page](https://numpy.org/doc/stable/reference/generated/numpy.copy.html) for the free function encourages the use of the `.copy()` method instead 
image

- on the other, there is no support for deep copying. [This page](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.copy.html#numpy.ndarray.copy) states that python's `copy.deepcopy` function should be used for that... 
image

python/tests/visualization/test_draw.py
```diff
@@ -138,11 +138,11 @@ def hw_kernel():
...
-     ╭───╮                                       ╭───╮╭─────╮╭───╮╭───╮
-q0 : ┤ h ├──●─────────────────────●──────────────┤ x ├┤ tdg ├┤ x ├┤ t ├
-     ╰───╯  │                     │              ╰─┬─╯╰─────╯╰─┬─╯├───┤
-q1 : ───────┼───────────●─────────┼───────────●────●───────────●──┤ t ├
-     ╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮ ╭───╮      ╰───╯
+     ╭───╮                                                        ╭───╮
+q0 : ┤ h ├──────────────●─────────────────────●────●───────────●──┤ t ├
+     ╰───╯              │                     │  ╭─┴─╮╭─────╮╭─┴─╮├───┤
+q1 : ───────●───────────┼─────────●───────────┼──┤ x ├┤ tdg ├┤ x ├┤ t ├
+     ╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮╭─┴─╮╭─────╮╭─┴─╮├───┤╰┬───┬╯╰───╯╰───╯
```
khalatepradnya, line 145: In this example, `q[0]` and `q[1]` are control qubits, so their order should not matter, but, it is not clear to me why the operations got swapped...
(Apologies if this was already discussed in the walk-through)

khalatepradnya: Saw this explanation - https://github.com/NVIDIA/cuda-quantum/pull/3537/files#r2452990751

python/tests/mlir/ast_continue.py
```diff
@@ -51,20 +54,40 @@ def kernel(x: float):
...
+            if idx < 3:
+                continue
+            ry(x, q)
+
+    res = cudaq.sample(kernel, numpy.pi)
+    assert len(res) == 1 and '000111' in res
+    print(f"output test1: {res}")
+
+
+# CHECK-LABEL:  output test1: { 000111:1000 }
```
khalatepradnya, line 93: nit: Do we need both `assert` on line# 89 and `CHECK` here?

python/tests/kernel/test_control_negations.py
```diff
@@ -207,6 +173,15 @@ def control_kernel():
...
+    def multi_control_kernel():
+        c, q = cudaq.qvector(4), cudaq.qubit()
+        x(c[0], c[3])
+        cudaq.control(custom_x, c[0], ~c[1], ~c[2], c[3], q)
+
+    counts = cudaq.sample(multi_control_kernel)
+    assert counts["10011"] == 1000
+
     # Note: calling cudaq.control on a registered operation
     # or on a built-in gate is not supported at the time of writing this
```
khalatepradnya, line 186: This comment is no longer valid, right? Because the test above does exactly that?

include/cudaq/Optimizer/Dialect/CC/CCTypes.h
```diff
@@ -38,10 +38,16 @@ inline bool SpanLikeType::classof(mlir::Type type) {
   return mlir::isa<StdvecType, CharspanType>(type);
 }
 
-/// Return true if and only if \p ty has dynamic extent. This is a recursive
+/// Returns true if and only if \p ty has dynamic extent. This is a recursive
 /// test on composable types.
 bool isDynamicType(mlir::Type ty);
 
+/// Returns true if and only if to memory needed to store a value of type
```
khalatepradnya: typo? `if to`
```suggestion
/// Returns true if and only if the memory needed to store a value of type
```

bettinaheim:
> I agree that `.copy()` is a very useful shorthand. It also feels the "most natural" to me.
> 
> Perhaps unfortunately, the "most pythonic" would be to support the use of the `copy` and `deepcopy` free functions that can be imported from the `copy` module.
> 
> ```python
> import copy
> 
> v = ['a', 'b']
> deep_v = ['a', ['a', 'b']]
> w = copy.copy(v)
> deep_w = copy.deepcopy(deep_v)
> ```
> 
> Given that we are parsing the AST, the fact that these functions are imported from the copy module makes this syntax an annoying complication, given the different ways these functions could be imported and used:
> 
> 1. `import copy` followed by `copy.copy` or `copy.deepcopy`
> 2. `from copy import copy, deepcopy`, followed by the unprefixed `copy(v)` or `deepcopy(v)`
> 3. `from copy import *`
> 4. worst of all, renaming: `from copy import copy as shallow_copy`
> 
> Supporting just 1. seems the simplest thing to do and is reasonable (even without checking for existence of the import). However, leaning into this Python convention without supporting at least 2. (which is how I'd use it) is likely to surprise users...
> 
> As a different point to consider, `numpy` supports both the [free function](https://numpy.org/doc/stable/reference/generated/numpy.copy.html) as well as providing a [`.copy()` method](https://numpy.org/doc/2.1/reference/generated/numpy.ndarray.copy.html#numpy.ndarray.copy):
> 
> * on the one hand, the numpy [docs page](https://numpy.org/doc/stable/reference/generated/numpy.copy.html) for the free function encourages the use of the `.copy()` method instead
> 

Summarizing a brief offline discussion: Matching the Python library functions for specific data types would require storing the Python data type along with the MLIR values during bridge construction. We definitively want to stick with the premise that "if it looks like Python code it should behave like Python code" for valid code that compiles. However, I think we can take a bit of liberty with additional functionality (i.e. adding member functions e.g. for lists that are not usually defined for Python lists). The rationale is that when we compile Python, we are mapping Python data types to other types that match Python behavior but may have a slightly different set of methods defined for convenience.

python/tests/mlir/ast_continue.py
bettinaheim: The test is run with llvm-lit. If this check is not here, then the test will silently fail since the output of the pytest process is piped into FileCheck. The assert is a bit redundant.

python/tests/kernel/test_control_negations.py
bettinaheim: The test above calls cudaq.control on kernel, not on a registered operation (custom unitary).

bettinaheim:
> /ok to test 1f1cfe8

bettinaheim:
> /ok to test 1f1cfe8

1tnguyen (approved):
LGTM  💯

PR #3538, opened 2025-10-22
Follow up to PR #3530

khalatepradnya:
This PR follows the work from PR #3530 by updating the logic for extracting result indices from output labels and adds a new unit test to verify correct handling of named results. The changes ensure that both auto-generated and user-defined result labels are interpreted correctly.

Such fixes will go away when we enable stricter check on sampled kernels targeting hardware backends to not have any explicit measurement operations present.

### DIFF
--- a/runtime/common/RecordLogParser.cpp
+++ b/runtime/common/RecordLogParser.cpp
@@ -151,9 +151,18 @@ void cudaq::RecordLogParser::handleOutput(
       // cause `Array index out of bounds` error. The proper fix is to disallow
       // explicit mz operations in sampled kernels. Also, `run` is appropriate
       // for getting sub-register results.
-      else if (recLabel.size() == 6 && recLabel[0] == 'r' &&
-               recLabel.back() >= '0' && recLabel.back() <= '9') {
-        idxLabel = recLabel.substr(1);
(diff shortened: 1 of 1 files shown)


runtime/common/RecordLogParser.cpp
```diff
@@ -129,18 +129,35 @@ void cudaq::RecordLogParser::handleOutput(
...
+        idxLabel = recLabel.substr(percentPos + 1);
+      }
+      // This logic is fragile; for example user may have only one mz assigned
+      // to variable like r00001 and it will be interpreted as index 1, and
+      // cause `Array index out of bounds` error. The proper fix is to disallow
+      // explicit mz operations in sampled kernels. Also, `run` is appropriate
+      // for getting sub-register results.
+      else if (recLabel.size() == 6 && recLabel[0] == 'r' &&
+               recLabel.back() >= '0' && recLabel.back() <= '9') {
+        idxLabel = recLabel.substr(1);
```
1tnguyen: Could we add a check that the substring is a number before using it? 
e.g., to further prevent cases like `rxxx_1`

khalatepradnya: Sure, see 5668807.

PR #3539, opened 2025-10-22
Set QIR v1.0 for base profile

khalatepradnya:
This PR updates to the `parseCodeGenTranslation` function to ensure that the QIR base profile is assigned a default version of `1.0`.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -34,7 +34,7 @@ on:
       target_machine:
         type: string
         required: false
-        description: 'Target machine (e.g., H1-1E).'
+        description: 'Target machine (e.g., H2-1E).'
       cudaq_test_image:
         type: string
         required: false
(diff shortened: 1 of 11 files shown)


runtime/common/CodeGenConfig.cpp
```diff
@@ -61,6 +61,13 @@ cudaq::parseCodeGenTranslation(const std::string &codegenTranslation) {
       .isBaseProfile = codeGenName == "qir-base",
   };
 
+  // Default version for base profile is 1.0
+  if (config.isBaseProfile) {
```
bettinaheim, line 65: The adaptive profile should also have a version number somewhere and it should also be 1.0 at this time. 
The legacy mode (using the old naming for metadata like num_required_qubits) is version 0.1

khalatepradnya: The adaptive profile version is handled in this function below at around lines 104-120.

PR #3540, opened 2025-10-22
Replace H1-* devices with H2-* devices

khalatepradnya:
This PR updates configuration, tests and documentation to use the newer Quantinuum H2-series machines instead of the older H1-series machines.


1tnguyen (approved):
Thanks @khalatepradnya 👍

PR #3545, opened 2025-10-24
Fix a bug in Python `run`, `run_async`, `sample_async`, and `observe_async`: callable `args` are dropped

1tnguyen:
### Description

Python's `run`, `run_async`, `sample_async`, and `observe_async` are handled entirely in PyBinding, which is different from `sample` and `observe` (Python implementation in `sample.py` and `observe.py`.

(Thanks @khalatepradnya, for pointing out all the places where we may have this oversight)

The main difference is that we're missing some logic in handling callable arguments, which is part of Python's `__call__` before invoking `pyAltLaunchKernel`. Hence, callable `args` are not packed. 

This PR fixes that and adds tests for this scenario.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -554,13 +554,12 @@ class PyASTBridge(ast.NodeVisitor):
         return id in ['mx', 'my', 'mz']
 
     def __isUnitaryGate(self, id):
-        return self.__isSimpleGate(id) or \
-            self.__isRotationGate(id) or \
-            self.__isAdjointSimpleGate(id) or \
-            self.__isControlledSimpleGate(id) or \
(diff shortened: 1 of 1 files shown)


python/tests/kernel/test_run_async_kernel.py
```diff
@@ -1015,6 +1016,34 @@ def kernel() -> bool:
...
+    @cudaq.kernel
+    def prep_1_state(qubits: cudaq.qvector):
+        x(qubits)
+
+    for num_qubits in [1, 2, 3, 4]:
+        results = cudaq.run_async(kernel,
+                                  prep_1_state,
+                                  num_qubits,
+                                  shots_count=10).get()
+        print("Results for prep_1_state with", num_qubits, "qubits:", results)
```
sacpis: Do we need this print statement?

PR #3549, opened 2025-10-26
Use `cudaq.complex()` in Python to math simulator precision

1tnguyen:
### Description

Use `cudaq.complex()`, which is equivalent to C++ `cudaq::complex` to take into account runtime target precision.

This is related to https://github.com/NVIDIA/cuda-quantum/pull/3507, where we reflect the default runtime simulator logic in the emulation workflow.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -17,16 +17,16 @@ on:
         options:
           - nightly
           - anyon
+          - fermioniq
           - infleqtion
           - ionq
           - iqm
           - oqc
+          - orca
           - pasqal
+          - qci
           - quantinuum
           - nvqc
-          - orca
-          - fermioniq
-          - qci
       single_test_name:
         type: string
         required: false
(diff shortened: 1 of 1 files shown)


python/tests/remote/test_remote_platform.py
```diff
@@ -456,7 +456,7 @@ def kernel():
 
 
 def test_capture_state():
-    s = cudaq.State.from_data(np.array([1., 0], dtype=np.complex128))
+    s = cudaq.State.from_data(np.array([1., 0], dtype=cudaq.complex()))
```
khalatepradnya, line 459: These changes are good to fix the broken tests. Can we add a few test cases where - `dtype` is specified as `np.complexxx` or `complex` and the default simulator is different (`qpp-cpu`, `nvidia`) to cover the expected behavior. Okay to do as a follow-up. Also, docs / examples should call this out - `use cudaq.complex()` , if not done already.

1tnguyen: Good point! I'll add them in this PR.

1tnguyen: Added in https://github.com/NVIDIA/cuda-quantum/pull/3549/commits/88211309db4f5dcfeb52d62712c16260c8caae8c

PR #3550, opened 2025-10-27
[NFC] Remove archaic style from python sources.

schweitzpgi:
"""
Line Continuation: Backslash at the end of a line allows you to break long lines of code. However, this style is generally discouraged unless necessary—using implicit continuation within parentheses, brackets, or braces is preferred for readability and maintenance. """

It also conforms to the existing style used in our source code.

### DIFF
--- a/docs/sphinx/applications/python/unitary_compilation_diffusion_models.ipynb
+++ b/docs/sphinx/applications/python/unitary_compilation_diffusion_models.ipynb
@@ -89,7 +89,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install genQC==0.2.3 torch --break-system-packages -q"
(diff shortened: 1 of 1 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -4818,6 +4820,7 @@ def visit_Name(self, node):
         table.
         """
 
+        from .kernel_decorator import isa_kernel_decorator
```
sacpis: Seems like you missed to check-in `kernel_decorator.py` with `isa_kernel_decorator` definition?

schweitzpgi: Didn;t mean to cherry pick that one.

PR #3552, opened 2025-10-28
Fix publishing failure

sacpis:
Fixes: https://github.com/NVIDIA/cuda-quantum/actions/runs/18803762479/job/53656710743#step:5:2574

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,12 +22,12 @@ dependencies = [
   'numpy >= 1.24',
   'scipy >= 1.10.1',
   'requests >= 2.31',
-  'nvidia-cublas-cu12 ~= 12.6',
+  'nvidia-cublas-cu12 >= 12.6',
   'nvidia-curand-cu12 ~= 10.3',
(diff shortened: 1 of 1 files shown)


sacpis:
> /ok to test da6a3e

sacpis:
> /ok to test dda6a3e

mitchdz:
If I'm not mistaken, this will still install `12.9` since it has the `~=` install parameter.

sacpis:
> /ok to test 0441896

mitchdz:
Do you have a publishing run that shows this being fixed?

In my PR I have updated the `pyproject.toml` to include 2 versions for each cuda version, `pyproject.toml.cu12` and `pyproject.toml.cu13`. For the cu13 build we should not be using the suffix at all as you can see at https://github.com/NVIDIA/cuda-quantum/blob/test-publishing/pyproject.toml.cu13#L28

Another concern I have with this approach, is does this happen to break compatibility when a user is doing driver forwarding with CUDA < 12.6? I don't particularly think it does, but I want to make sure that we don't break that use case.

PR #3553, opened 2025-10-28
[QIR] Insert array record call in sampling workflow

khalatepradnya:
This PR introduces a new module pass, `QirInsertArrayRecord`, that analyzes instruction patterns and inserts array recording instruction required for proper measurement result handling in QIR pipelines. The pass looks at existing result recording calls and ensures the QIR output declares the correct array size and type label for measurement results. 

This PR also enables the pass in compilation workflow.

### DIFF
--- a/utils/mock_qpu/quantinuum/__init__.py
+++ b/utils/mock_qpu/quantinuum/__init__.py
@@ -10,7 +10,7 @@ import cudaq
 from fastapi import FastAPI, HTTPException, Header, Query
 from fastapi.responses import JSONResponse
 from typing import Union
-import uvicorn, uuid, base64, ctypes
+import uuid, base64, ctypes
 from pydantic import BaseModel
 from llvmlite import binding as llvm
 
@@ -432,11 +432,3 @@ async def create_decoder_config(job: dict):
             }
         }
     }
-
-
-def startServer(port):
-    uvicorn.run(app, port=port, host='0.0.0.0', log_level="info")
-
-
-if __name__ == '__main__':
-    startServer(62440)
--- a/.github/workflows/config/gitlab_commits.txt
+++ b/.github/workflows/config/gitlab_commits.txt
@@ -1,2 +1,2 @@
 nvidia-mgpu-repo: cuda-quantum/cuquantum-mgpu.git
-nvidia-mgpu-commit: 7fcf926ca043440c72c5cf844c5b7030a0f69a3a
+nvidia-mgpu-commit: 7f4b197923b5b77d17edee56e1a5f1aadee7d837
--- a/.github/workflows/config/spellcheck_config.yml
+++ b/.github/workflows/config/spellcheck_config.yml
@@ -283,3 +283,6 @@ matrix:
       # Ignore word prefixes, e.g. "pre-" in pre-computed
       - open: '(^|\s)[A-Z]?[a-z]+-'
         close: '(?=.)'
+      # Ignore shebang (#!)
+      - open: '^!'
+        close: '$'
--- a/docs/sphinx/using/extending/backend.rst
+++ b/docs/sphinx/using/extending/backend.rst
@@ -307,7 +307,7 @@ Create unit tests for your server helper:
     #!/bin/bash
     
     # Start the mock server
-    python3 -m utils.mock_qpu.<provider_name> @PORT@ &
+    python3 utils/start_mock_qpu.py <provider_name> &
     SERVER_PID=$!
     
     # Wait for server to start
--- a/python/tests/dynamics/test_evolve_dynamics.py
+++ b/python/tests/dynamics/test_evolve_dynamics.py
@@ -154,6 +154,87 @@ def test_batching_bugs():
         assert len(evolution_result.intermediate_states()) == len(steps)
 
 
+def test_density_matrix_indexing():
+    """
+    Test that density matrix element access uses correct indexing.
+    
+    This is a regression test for a bug where the operator() function
+    used the total dimension (dim*dim) instead of single-side dimension (dim)
+    for bounds checking and linear index calculation.
+    
+    For a 2-qubit system (4x4 density matrix with 16 total elements):
+    - Valid indices should be 0, 1, 2, 3
+    - Accessing rho[i, j] should compute linear index as i * 4 + j
+    - The bug computed i * 16 + j, causing out-of-bounds access
+    """
+    from cudaq import spin
+    from cudaq.dynamics import Schedule
+
+    # 1-qubit system: 2x2 density matrix
+    psi0_1q = cudaq.State.from_data(cp.array([1.0, 0.0], dtype=cp.complex128))
+    hamiltonian_1q = 0.0 * spin.z(0)
+    steps = np.linspace(0.0, 0.1, 2)
+    schedule = Schedule(steps, ["t"])
+
+    result_1q = cudaq.evolve(
+        hamiltonian_1q,
+        {0: 2},
+        schedule,
+        psi0_1q,
+        collapse_operators=[spin.z(0)],
+        store_intermediate_results=cudaq.IntermediateResultSave.ALL,
+    )
+
+    rho_1q = result_1q.final_state()
+    # For |0> initial state, density matrix is |0><0|
+    # rho[0,0] = 1, rho[1,1] = 0
+    assert abs(rho_1q[0, 0] - 1.0) < 1e-10
+    assert abs(rho_1q[1, 1]) < 1e-10
+    assert abs(rho_1q[0, 1]) < 1e-10
+    assert abs(rho_1q[1, 0]) < 1e-10
+
+    # Test out-of-bounds access is rejected
+    with pytest.raises(RuntimeError, match="indices out of range"):
+        _ = rho_1q[2, 0]
+
+    # 2-qubit system: 4x4 density matrix
+    psi0_2q = cudaq.State.from_data(
+        cp.array([1.0, 0.0, 0.0, 0.0], dtype=cp.complex128))
+    hamiltonian_2q = 0.0 * spin.z(0) + 0.0 * spin.z(1)
+
+    result_2q = cudaq.evolve(
+        hamiltonian_2q,
+        {
+            0: 2,
+            1: 2
+        },
+        schedule,
+        psi0_2q,
+        collapse_operators=[spin.z(0)],
+        store_intermediate_results=cudaq.IntermediateResultSave.ALL,
+    )
+
+    rho_2q = result_2q.final_state()
+    # For |00> initial state, density matrix is |00><00|
+    # Only rho[0,0] = 1, all other elements = 0
+    assert abs(rho_2q[0, 0] - 1.0) < 1e-10
+    # These indices would cause out-of-bounds access with the old buggy code
+    assert abs(rho_2q[1, 1]) < 1e-10
+    assert abs(rho_2q[2, 2]) < 1e-10
+    assert abs(rho_2q[3, 3]) < 1e-10
+    # Off-diagonal elements
+    assert abs(rho_2q[0, 3]) < 1e-10
+    assert abs(rho_2q[3, 0]) < 1e-10
+
+    # Test out-of-bounds access is rejected
+    with pytest.raises(RuntimeError, match="indices out of range"):
+        _ = rho_2q[4, 0]
+    with pytest.raises(RuntimeError, match="indices out of range"):
+        _ = rho_2q[0, 4]
+    with pytest.raises(RuntimeError, match="indices out of range"):
+        _ = rho_2q[4, 4]
+
+
 # leave for gdb debugging
 if __name__ == "__main__":
     loc = os.path.abspath(__file__)
--- a/runtime/nvqir/cudensitymat/CuDensityMatState.cpp
+++ b/runtime/nvqir/cudensitymat/CuDensityMatState.cpp
@@ -160,9 +160,12 @@ CuDensityMatState::operator()(std::size_t tensorIdx,
     if (indices.size() != 2)
       throw std::runtime_error("CuDensityMatState holding a density matrix "
                                "supports only 2-dimensional indices");
-    if (indices[0] >= dimension || indices[1] >= dimension)
+    // For density matrix, dimension is the total size (dim*dim), so we need
+    // to compute the single-side dimension for bounds checking.
+    const std::size_t dim = static_cast<std::size_t>(std::sqrt(dimension));
+    if (indices[0] >= dim || indices[1] >= dim)
       throw std::runtime_error("CuDensityMatState indices out of range");
-    return extractValue(indices[0] * dimension + indices[1]);
+    return extractValue(indices[0] * dim + indices[1]);
   }
   if (indices.size() != 1)
     throw std::runtime_error(
--- a/unittests/backends/anyon/AnyonStartServerAndTest.sh.in
+++ b/unittests/backends/anyon/AnyonStartServerAndTest.sh.in
@@ -36,7 +36,7 @@ EOF
 
 # Uncomment this and connect to mock_qpu backend server within the container (mock_qpu/anyon) under main/utils
 # Launch the fake server
-PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/mock_qpu/anyon/__init__.py &
+PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/start_mock_qpu.py anyon &
 # we'll need the process id to kill it
 pid=$(echo "$!")
 n=0
--- a/unittests/backends/braket/BraketStartServerAndTest.sh.in
+++ b/unittests/backends/braket/BraketStartServerAndTest.sh.in
@@ -36,7 +36,7 @@ EOF
 
 # Uncomment this and connect to mock_qpu backend server within the container (mock_qpu/braket) under main/utils
 # Launch the fake server
-PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/mock_qpu/braket/__init__.py &
+PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/start_mock_qpu.py braket &
 # we'll need the process id to kill it
 pid=$(echo "$!")
 n=0
--- a/unittests/backends/infleqtion/InfleqtionStartServerAndTest.sh.in
+++ b/unittests/backends/infleqtion/InfleqtionStartServerAndTest.sh.in
@@ -21,7 +21,7 @@ EOF
 }
 
 # Launch the fake server
-PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/mock_qpu/infleqtion/__init__.py &
+PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/start_mock_qpu.py infleqtion &
 # we'll need the process id to kill it
 pid=$(echo "$!")
 n=0
--- a/unittests/backends/ionq/IonQStartServerAndTest.sh.in
+++ b/unittests/backends/ionq/IonQStartServerAndTest.sh.in
@@ -21,7 +21,7 @@ EOF
 }
 
 # Launch the fake server
-PYTHONPATH=@CMAKE_BINARY_DIR@/python @Python_EXECUTABLE@ @CMAKE_SOURCE_DIR@/utils/mock_qpu/ionq/__init__.py &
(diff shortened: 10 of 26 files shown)


include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -131,6 +131,17 @@ def QIRToQIRProfile : Pass<"convert-to-qir-profile"> {
...
 
+def QirInsertArrayRecord : Pass<"qir-insert-array-record", "mlir::ModuleOp"> {
+  let summary = "Analyze instruction patterns and insert new instructions for QIR";
+  let description = [{
+    This pass performs analysis on the instruction patterns between QIR prep and 
+    conversion passes, then inserts array recording instruction based on the analysis.
+  }];
+  let dependentDialects = ["quake::QuakeDialect", "cudaq::cc::CCDialect", 
+      "mlir::LLVM::LLVMDialect"];
+  let constructor = "cudaq::opt::createQirInsertArrayRecord()";
```
schweitzpgi: Do not use old tablegen pattern for new passes.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,219 @@
...
+  // We insert only one array record call
+  if (!allocaOps.empty())
+    builder.setInsertionPointAfter(allocaOps[0]);
+  else if (firstMeasureOp)
+    builder.setInsertionPoint(firstMeasureOp);
+  else
+    return failure();
+
+  // Create the label string: "array<i8 x N>"
+  std::string labelStr = "array<i8 x " + std::to_string(resultCount) + ">";
```
schweitzpgi: This isn't the documented label format, is it?

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+struct QirInsertArrayRecordPass
+    : public cudaq::opt::impl::QirInsertArrayRecordBase<
+          QirInsertArrayRecordPass> {
+
+  using QirInsertArrayRecordBase::QirInsertArrayRecordBase;
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    module.walk([&](func::FuncOp funcOp) {
```
schweitzpgi: You don't need to do a DFS over the entire IR to find top-level artifacts.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,219 @@
...
+          funcOp->hasAttr(cudaq::runtime::enableCudaqRun))
+        return WalkResult::advance();
+
+      AllocaMeasureStoreAnalysis analysis(funcOp);
+      if (analysis.arraySize == 0)
+        return WalkResult::advance();
+
+      LLVM_DEBUG(llvm::dbgs() << "Before adding array recording call:\n"
+                              << *funcOp);
+      if (failed(insertArrayRecordingCalls(funcOp, analysis.arraySize,
```
schweitzpgi: Should use the intrinsic loading instead of rolling your own. Aren't these already loaded by the QIR API Prep pass, though?

test/Translate/array_record_insert.qke
```diff
@@ -0,0 +1,248 @@
...
+// CHECK:           %[[VAL_2:.*]] = quake.extract_ref %[[VAL_0]][1] : (!quake.veq<4>) -> !quake.ref
+// CHECK:           quake.x {{\[}}%[[VAL_1]]] %[[VAL_2]] : (!quake.ref, !quake.ref) -> ()
+// CHECK:           %[[VAL_3:.*]] = quake.extract_ref %[[VAL_0]][3] : (!quake.veq<4>) -> !quake.ref
+// CHECK:           quake.h %[[VAL_3]] : (!quake.ref) -> ()
+// CHECK:           %[[VAL_4:.*]] = quake.extract_ref %[[VAL_0]][2] : (!quake.veq<4>) -> !quake.ref
+// CHECK:           quake.h %[[VAL_4]] : (!quake.ref) -> ()
+// CHECK:           quake.x {{\[}}%[[VAL_3]]] %[[VAL_4]] : (!quake.ref, !quake.ref) -> ()
+// CHECK:           quake.h %[[VAL_4]] : (!quake.ref) -> ()
+// CHECK:           %[[VAL_5:.*]] = cc.alloca !cc.array<i8 x 2>
+// CHECK:           %[[VAL_6:.*]] = cc.string_literal "array<i8 x 4>" : !cc.ptr<!cc.array<i8 x 14>>
```
schweitzpgi: I think the Run Spec used something like `"[i8 x 4]"`. We should stick with what was documented there.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,219 @@
...
+      LLVM_DEBUG(llvm::dbgs() << "After adding array recording call:\n"
+                              << *funcOp);
+
+      return WalkResult::advance();
+    });
+  }
+};
+} // namespace
+
+std::unique_ptr<Pass> cudaq::opt::createQirInsertArrayRecord() {
```
schweitzpgi: This is not needed. Tablegen will autogenerate the `create<Pass>` function for you.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
schweitzpgi: I confused two labels. My bad.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: I wanted to make this pass self-contained, since we cannot guarantee that the prep pass has run before this one. 
Should I move the `loadIntrinsic` call from line#176 here?

include/cudaq/Optimizer/CodeGen/Passes.td
khalatepradnya: Removed.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Replaced with for loop.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Removed.

include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -131,6 +131,16 @@ def QIRToQIRProfile : Pass<"convert-to-qir-profile"> {
   let constructor = "cudaq::opt::createQIRToQIRProfilePass(\"qir-base\")";
 }
 
+def QirInsertArrayRecord : Pass<"qir-insert-array-record", "mlir::ModuleOp"> {
```
bettinaheim, line 134: I assume this is a separate pass since it is used for sample and we don't have a return statement we replace there?

runtime/common/RecordLogParser.cpp
```diff
@@ -119,9 +119,14 @@ void cudaq::RecordLogParser::handleOutput(
...
-      // Currently, our QIR for sampled kernel only has a sequence of RESULT
-      // records, not wrapped in an ARRAY. Hence, we treat it as an array of
-      // results.
+      // NOTE: This is a temporary workaround until all backends consistently
+      // use the new transformation pass that wraps result records inside an
+      // array record output. For now, we permit "naked" RESULT records, i.e.,
+      // if the QIR produced by a sampled kernel emits a sequence of RESULT
+      // records without enclosing them in an ARRAY, we interpret them
+      // collectively as an array of results.
+      // NOTE: This assumption prevents us from correctly supporting `run` with
```
bettinaheim, line 129: What is needed to get rid of this workaround?

utils/mock_qpu/quantinuum/__init__.py
```diff
@@ -218,6 +218,9 @@ async def create_job(job: dict):
     funcPtr = engine.get_function_address(kernelFunctionName)
     kernel = ctypes.CFUNCTYPE(None)(funcPtr)
 
+    # Clear any leftover log from previous jobs
+    cudaq.testing.getAndClearOutputLog()
```
bettinaheim, line 222: Why do we need to explicitly clear the output log here? I.e. shouldn't that be done automatically as part of creating the correct cudaq data type to represent the result? Why do we explicitly clear it here but nowhere else?

include/cudaq/Optimizer/CodeGen/Passes.td
khalatepradnya: Yes, that is correct. We have logic in place to automatically insert measurements which in turn get converted to result recording call, but there is no "warpper" array recording call which is now being added by this pass.

runtime/common/RecordLogParser.cpp
khalatepradnya: I think the proper way / time would be when we disallow user-specified measurements in sampled kernels, then we can also drop the index from label logic on lines 144-172.

utils/mock_qpu/quantinuum/__init__.py
khalatepradnya: Great Q! It took me a while to debug this one. In general, the output log is cleared by the callers (`runTheKernel`). But, for testing with mock servers, we are using "low-level" APIs, from the `testing` submodule and this caller should clear the output. Why it is only needed here and now is because - For Quantinuum we support two types of outputs, legacy one and QIR log. This PR adds the array recording call, whose implementation in `NVQIR.cpp` adds it to the output log (Logic for result record is different). For the legacy tests, this log is not consumed and can spill into next test causing it to fail.  Hence, we clear the output here. I don't think this can happen in production since different APIs are used.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
schweitzpgi: Hmm.

Well, the QIR lowering prep(aration) pass should, by definition, always run prior to lowering things to QIR.  If we're generating QIR code/calls and don't know if the prep pass was run, we're doing something quite wrong.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,213 @@
...
+  return {};
+}
+
+// Walk a function to identify all the measure-discriminate-store patterns and
+// collect the associated `AllocaOp` when the measurement results are stored.
+// Collect only unique AllocaOps - since each may correspond to multiple
+// measurement operations. When there are no explicit stores, track the first
+// measurement operation and the get the total number of measurements.
+struct AllocaMeasureStoreAnalysis {
+  AllocaMeasureStoreAnalysis() = default;
```
schweitzpgi: Why is this default ctor needed? Shouldn't the analysis always be run?

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,213 @@
...
+    return {};
+  }
+  return {};
+}
+
+// Walk a function to identify all the measure-discriminate-store patterns and
+// collect the associated `AllocaOp` when the measurement results are stored.
+// Collect only unique AllocaOps - since each may correspond to multiple
+// measurement operations. When there are no explicit stores, track the first
+// measurement operation and the get the total number of measurements.
```
schweitzpgi: What does "unique AllocaOps" mean? What happens if the variable (the alloca slot) is used by other loads and stores, which may have nothing to do with any quantum operations? What happens if the location being stored to is not an AllocaOp? How do you count measurements? A measurement can measure multiple qubits.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,213 @@
...
+      }
+
+      // TODO: Check if more operations need to be added here.
+      if (!isa<quake::DiscriminateOp, cudaq::cc::CastOp>(op)) {
+        return WalkResult::advance();
+      }
+
+      // Find the operands derived from measurements
+      for (auto operand : op->getOperands()) {
+        if (valueToMeasurement.count(operand)) {
```
schweitzpgi: This doesn't look correct. It should be ok if the SSA value is lexically subsequent to the defining op, but that's not going to always be true in legal IR. Effectively, it does not consider the control flow issues.

I've said this in other PRs, this sort of roll-your-own pattern matching in each pass is going to become a maintenance obstacle as we go along. We really need to be using the factored use-def (value semantics) form of the IR to our advantage for these rewrites.

include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -131,6 +131,16 @@ def QIRToQIRProfile : Pass<"convert-to-qir-profile"> {
...
+  let summary = "Analyze instruction patterns and insert new instructions for QIR";
+  let description = [{
+    This pass performs analysis on the instruction patterns between QIR prep and 
+    conversion passes, then inserts array recording instruction based on the analysis.
+  }];
+  let dependentDialects = ["quake::QuakeDialect", "cudaq::cc::CCDialect", 
+      "mlir::LLVM::LLVMDialect"];
```
schweitzpgi: I don't see where the LLVMDialect is used. Did I miss it?

I think it would be ok to delete this dependentDialects since we know the code will have Quake and CC dialects in the input. (We don't need to check that those dialects were loaded.)

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: This analysis is no longer used and has been removed.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Removed this logic, using result recording calls directly.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Removed.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: I have restructured the flow, and the new pass runs after QIR Prep and API passes (before finalization). But the intrinsic for array record call isn't pre-loaded, hence I add it (if it is missing).

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+// Inserts a QIR array record output call to declare measurement result storage.
+// QIR requires `__quantum__rt__array_record_output()` be called before multiple
+// measurements to declare the output array size and type label. This is
+// required in `sample` API since it always returns a vector of measurement
+// results. The call is inserted before the first
+// `__quantum__rt__result_record_output` call. The label string is created as
+// "array<i1 x N>" where N is the total number of measurement results. The array
+// record output call is created as:
+// `__quantum__rt__array_record_output(N, label);`
+LogicalResult insertArrayRecordingCall(OpBuilder &builder, mlir::Location loc,
```
schweitzpgi:
```suggestion
static LogicalResult insertArrayRecordingCall(OpBuilder &builder, mlir::Location loc,
```
Prefer making functions static instead of anonymous scope. (It's a LLVM coding standard thing.)

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+namespace cudaq::opt {
+#define GEN_PASS_DEF_QIRINSERTARRAYRECORD
+#include "cudaq/Optimizer/CodeGen/Passes.h.inc"
+} // namespace cudaq::opt
+
+#define DEBUG_TYPE "qir-insert-array-record"
+
+using namespace mlir;
+
+namespace {
```
schweitzpgi: Can move this line down to the struct below.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+// Inserts a QIR array record output call to declare measurement result storage.
+// QIR requires `__quantum__rt__array_record_output()` be called before multiple
+// measurements to declare the output array size and type label. This is
+// required in `sample` API since it always returns a vector of measurement
+// results. The call is inserted before the first
+// `__quantum__rt__result_record_output` call. The label string is created as
+// "array<i1 x N>" where N is the total number of measurement results. The array
```
schweitzpgi: Does this line need backquotes rather than double quotes?

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+      builder.getContext(), builder.getI8Type(), labelStr.size() + 1));
+  Value lit = builder.create<cudaq::cc::CreateStringLiteralOp>(
+      loc, strLitTy, builder.getStringAttr(labelStr));
+  auto i8PtrTy = cudaq::cc::PointerType::get(builder.getI8Type());
+  Value label = builder.create<cudaq::cc::CastOp>(loc, i8PtrTy, lit);
+  Value size = builder.create<arith::ConstantIntOp>(loc, resultCount, 64);
+  builder.create<func::CallOp>(loc, TypeRange{},
+                               cudaq::opt::QIRArrayRecordOutput,
+                               ArrayRef<Value>{size, label});
+  return success();
```
schweitzpgi, line 54: It looks like this always return success ... so we don't need a return value here.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+          recordOutputCalls.push_back(callOp);
+      });
+
+      if (arrayRecordExists || recordOutputCalls.empty())
+        continue;
+
+      LLVM_DEBUG(llvm::dbgs() << "Before adding array recording call:\n"
+                              << *funcOp);
+      // Add the declaration of array recording call to the module
+      auto module = funcOp->getParentOfType<ModuleOp>();
```
schweitzpgi: This module is necessarily the same module as line 66, no?

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
```diff
@@ -0,0 +1,109 @@
...
+      if (!module.lookupSymbol(cudaq::opt::QIRArrayRecordOutput)) {
+        auto irBuilder = cudaq::IRBuilder::atBlockEnd(module.getBody());
+        if (failed(irBuilder.loadIntrinsic(module,
```
schweitzpgi: Just load this up front. It doesn't hurt anything to have the declaration sitting there in the module and its more efficient to just load it once than to check every time if it wasn't added yet.

runtime/common/RuntimeMLIRCommonImpl.h
```diff
@@ -181,7 +181,7 @@ static bool isValidOutputCallInstruction(llvm::Instruction &inst) {
 // block contains irreversible operations (measurements), and the blocks may not
 // overlap.
 // Reference:
-// https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md?plain=1#L237
+// https://github.com/qir-alliance/qir-spec/blob/684b17b9367ec793efdfbb7d804bb78bea1c51ed/specification/profiles/Base_Profile.md#L196
```
schweitzpgi: nit: consider using tinyurl here.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Fixed in 27e3547e37795126221c58fe49c3649b31e313f2

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Unless I misunderstood, simply dropping the return statement would give `control reaches end of non-void function` error.

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: yes! Fixed in 27e3547e37795126221c58fe49c3649b31e313f2

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Fixed in 27e3547e37795126221c58fe49c3649b31e313f2

lib/Optimizer/CodeGen/QirInsertArrayRecord.cpp
khalatepradnya: Fixed in 27e3547e37795126221c58fe49c3649b31e313f2

runtime/common/RuntimeMLIRCommonImpl.h
khalatepradnya: Shortened the commit ID.

PR #3555, opened 2025-10-29
Python support for returning lists of simple types from `cudaq.run`

1tnguyen:
### Description

- Lift the restriction in Python's `run` to allow for returning arrays.

- Add support for dynamic array logging, e.g., without arg synthesis.

- Enable all the Python tests for returning arrays from `cudaq.run`.

- Add some C++ tests for dynamic arrays.

### DIFF
--- a/docs/sphinx/applications/python/edge_detection.ipynb
+++ b/docs/sphinx/applications/python/edge_detection.ipynb
@@ -413,7 +413,7 @@
     "    x.ctrl(ancilla, qubits[0])\n",
     "    \n",
     "    for i in range(1, qubit_num):\n",
-    "        x.ctrl([ancilla,qubits[0:i]], qubits[i])\n",
+    "        x.ctrl([ancilla, *qubits[0:i]], qubits[i])\n",
     "            \n",
     "    h(ancilla)\n",
     "\n",
--- a/docs/sphinx/applications/python/quantum_volume.ipynb
+++ b/docs/sphinx/applications/python/quantum_volume.ipynb
@@ -68,7 +68,7 @@
     "    param_index = 0\n",
     "\n",
     "    for layer in range(n):\n",
-    "        for gate in range(n / 2):\n",
+    "        for gate in range(n // 2):\n",
     "            su4_gate(reg[permutations[layer * n + gate * 2]],\n",
     "                     reg[permutations[layer * n + gate * 2 + 1]],\n",
     "                     params[param_index:param_index + 21])\n",
--- a/docs/sphinx/using/backends/sims/tnsims.rst
+++ b/docs/sphinx/using/backends/sims/tnsims.rst
@@ -98,11 +98,16 @@ Use the following commands to enable distribution across multiple GPUs (adjust t
 Specific aspects of the simulation can be configured by setting the following of environment variables:
 
 * **`CUDA_VISIBLE_DEVICES=X`**: Makes the process only see GPU X on multi-GPU nodes. Each MPI process must only see its own dedicated GPU. For example, if you run 8 MPI processes on a DGX system with 8 GPUs, each MPI process should be assigned its own dedicated GPU via `CUDA_VISIBLE_DEVICES` when invoking `mpiexec` (or `mpirun`) commands. 
-* **`OMP_PLACES=cores`**: Set this environment variable to improve CPU parallelization.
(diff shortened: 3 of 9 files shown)


khalatepradnya:
nit: May be call them `list`s instead of arrays in context of Python.

targettests/execution/cudaq_run.cpp
```diff
@@ -42,6 +42,13 @@ __qpu__ std::vector<bool> vector_bool_test() {
   return vec;
 }
 
+__qpu__ std::vector<bool> vector_bool_dynamic_size_test(int size) {
+  std::vector<bool> vec(size);
+  for (int i = 0; i < size; i++)
+    vec[i] = (i % 2 == 0) ? true : false;
+  return vec;
+}
+
```
khalatepradnya, line 51: yay! 😊

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -100,10 +100,13 @@ pyRunTheKernel(const std::string &name, const std::string &origName,
...
-  if (returnTy.isa<cc::StdvecType>()) {
-    throw std::runtime_error("`cudaq.run` does not yet support returning "
-                             "`list` from entry-point kernels.");
+  // Disallow returning nested vectors from entry-point kernels.
+  if (auto vecTy = dyn_cast<cudaq::cc::StdvecType>(returnTy)) {
+    auto elemTy = vecTy.getElementType();
+    if (elemTy.isa<cudaq::cc::StdvecType>())
+      throw std::runtime_error(
+          "`cudaq.run` does not yet support returning nested `list` from "
+          "entry-point kernels.");
```
khalatepradnya, line 110: Q on scope: what about lists of structs (dataclass)?

1tnguyen: Yes, that's right. We should check for that as well.
Added in 2d54cdfb79550af15a1b40b2b5f3a56c4b6d0c97 along with some tests.

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -158,7 +162,55 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
...
+
+          // If we reach here and we cannot determine the constant size of the
+          // buffer, then we will generate dynamic output logging with a for
+          // loop.
+          Value vecSz = rewriter.template create<cudaq::cc::StdvecSizeOp>(
+              loc, rewriter.getI64Type(), val);
+          const std::string arrayLabelPrefix =
+              "array<" + translateType(vecTy.getElementType()) + " x ";
+          Value labelBuffer =
+              makeLabel(loc, rewriter, arrayLabelPrefix, vecSz, ">");
```
sacpis, line 176: I think we can put the logic of arrayLabelPrefix inside makeLabel. So that we don't have to hardcode values here.

1tnguyen: We have 2 use cases for making labels: the array label, e.g., `array<i8 x 10>`, and the array element index labels, e.g., `[7]`. Hence, the `makeLabel` function would need to handle generic prefix and postfix.

sacpis: Oh I meant this part of the code

```
const std::string arrayLabelPrefix =
              "array<" + translateType(vecTy.getElementType()) + " x ";
          Value labelBuffer =
              makeLabel(loc, rewriter, arrayLabelPrefix, vecSz, ">");
```

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -158,7 +162,55 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
...
+                        loc, i1PtrTy, eleAddr);
+                    return rewriter.create<cudaq::cc::LoadOp>(loc, i1Cast);
+                  }
+
+                  return rewriter.create<cudaq::cc::LoadOp>(loc, eleAddr);
+                }();
+                const std::string prefix = preStr + "[";
+                const std::string postfix = "]";
+                Value dynamicLabel =
+                    makeLabel(loc, rewriter, prefix, indexVar, postfix);
```
sacpis, line 211: Oh now I see it here.

sacpis (commented):
LGTM. Thanks @1tnguyen.

sacpis (approved):

PR #3563, opened 2025-10-31
Fix some notebooks to conform to stricter Python bridge

1tnguyen:
### Description

- Use integer div (`//`) (`range` requires integer value)

- Expand the qubit list with `*` to use with `control`: mix-and-match qubits and qubit arrays is not allowed.


docs/sphinx/applications/python/quantum_volume.ipynb
```diff
@@ -68,7 +68,7 @@
     "    param_index = 0\n",
     "\n",
     "    for layer in range(n):\n",
-    "        for gate in range(n / 2):\n",
+    "        for gate in range(n // 2):\n",
```
sacpis, line 71: It is good, but the datatype is already mentioned as int in the function definition. So wondering if it is needed?

sacpis (approved):
LGTM. Thanks @1tnguyen.

docs/sphinx/applications/python/quantum_volume.ipynb
1tnguyen: This is due to our recent update to the [bridge](https://github.com/NVIDIA/cuda-quantum/blob/b81fa8e3ff160805d0bd8b70644dfbfe5122239c/python/cudaq/kernel/ast_bridge.py#L3531-L3534
) to handle division operator in the same manner as [Python3 ](https://docs.python.org/3/whatsnew/2.2.html#pep-238-changing-the-division-operator), not as C (or C++). Hence, `n/2` would be a floating-point value.

PR #3570, opened 2025-11-04
[NFC] Formatting changes.

schweitzpgi:
These are some NFC changes from the rework of the python launchers, compilation, and bridge.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -4970,9 +4970,10 @@ class PyASTBridge(ast.NodeVisitor):
 
         # Throw an exception for the case that the name is not
         # in the symbol table
-        self.emitFatalError(f"Invalid variable name requested - '{node.id}' is "
-                            f"not defined within the quantum kernel it is used "
-                            f"in.", node)
+        self.emitFatalError(
+            f"Invalid variable name requested - '{node.id}' is "
+            f"not defined within the quantum kernel it is used "
+            f"in.", node)
 
 
 def compile_to_mlir(astModule, capturedDataStorage: CapturedDataStorage,


python/cudaq/kernel/ast_bridge.py
```diff
@@ -1221,9 +1236,10 @@ def visit_FunctionDef(self, node):
             self.name = node.name
             self.capturedDataStorage.name = self.name
 
-            # the full function name in MLIR is `__nvqpp__mlirgen__` + the function name
-            if not self.disableNvqppPrefix:
-                fullName = nvqppPrefix + node.name
+            # the full function name in MLIR is `__nvqpp__mlirgen__` + the
+            # function name
+            if self.disableNvqppPrefix:
+                fullName = node_name
```
1tnguyen: I don't see where this `node_name` variable is defined?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1639,6 +1655,29 @@ def process_potential_ptr_types(value):
             self.emitFatalError("unrecognized attribute {}".format(node.attr),
                                 node)
 
+    def find_unique_decorator_name(self, name):
+        mod = sys.modules[self.kernelModuleName]
```
1tnguyen: Is `self.kernelModuleName` a new property of `PyASTBridge`?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1639,6 +1655,29 @@ def process_potential_ptr_types(value):
...
+        return None
+
+    # TODO: move this out of the bridge and instead resolve these symbols at the
+    # callsites.
+    def get_kernel_from_context(self, node_func_id):
+        decorator = recover_kernel_decorator(node_func_id)
+        if decorator:
+            nodeFuncId = decorator.uniqName
+            if nodeFuncId in globalKernelUniqueModule:
+                registeredKernel = recover_func_op(
```
1tnguyen: It looks like `recover_func_op` is not part of this changeset.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4814,6 +4853,23 @@ def visit_AugAssign(self, node):
...
+        correctly implement Python's dynamic scope rules.
+
+        All device kernel references must be resolved at a kernel decorator call
+        site, and the resolution must include resolution of all lifted arguments
+        (recursively).
+        """
+        if not self.liftedArgs:
+            self.firstLiftedPos = len(self.entry.arguments)
+        if symbol not in self.liftedArgs:
+            self.liftedArgs.append(symbol)
```
1tnguyen: This is unused at the moment, isn't it?

python/cudaq/kernel/ast_bridge.py
schweitzpgi: This includes changes which will be used later.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: This includes changes that will be used later.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: This includes changes to be used later.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1774,7 +1772,7 @@ def processControlAndAdjoint(pyFuncVal, attrName):
                 if (not quake.RefType.isinstance(controls[0].type) and
                         not quake.VeqType.isinstance(controls[0].type)):
                     self.emitFatalError(
-                        f'invalid argument type for control operand', node)
+                        'invalid argument type for control operand', node)
```
sacpis, line 1775: Just curious, why have we removed `f` from here and kept it elsewhere?

sacpis: Got it.

sacpis (approved):
LGTM. Thanks @schweitzpgi.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: No variables were being formatted here, so it just looked like using a `printf` call when `puts` was simpler and better.

PR #3579, opened 2025-11-05
[CI] Publishing pipeline fixes continued

khalatepradnya:
This PR follows-up the work from https://github.com/NVIDIA/cuda-quantum/pull/3544

Successful pipeline - https://github.com/NVIDIA/cuda-quantum/actions/runs/19126706667

* Updated workflow and validation scripts to use full CUDA version strings (e.g., `12.4.0`, `13.0.2`) for conda channel selection for compatibility with available CUDA packages
* Improved detection logic for CUDA versions in `setup.py` by including CUDA 13 library names, ensuring correct identification of the environment. (`python/metapackages/setup.py`) 
* Refactored CUDA library path setup in `cudaq/__init__.py` to dynamically construct and set the `CUDAQ_DYNLIBS` environment variable based on the detected CUDA version, supporting both CUDA 12 and CUDA 13 and handling missing libraries gracefully. (`python/cudaq/__init__.py`)
* Updated Dockerfile installation logic to install the correct set of CUDA Python packages depending on the major CUDA version, improving container compatibility and reducing installation errors. (`docker/release/cudaq.ext.Dockerfile`)
* Enhanced notebook validation workflow to create a dedicated Jupyter kernel and virtual environment for container-based notebook execution, allowing notebooks to install their own dependencies and improving isolation and reproducibility. (`scripts/validate_container.sh`, `docs/notebook_validation.py`)
* Added missing Python packages (`python3-venv`, `python3-pip`) to the publishing workflow setup to ensure required tools are available for builds and validation. (`.github/workflows/publishing.yml`)
* Metapackage validation: (1) explicitly specify the conda channel, don't assume the `.0` patch version in the channel name (conda 13.0.0 channel is not a real CUDA 13 channel); (2) install torch (during validation) matching the version of cuda-q wheel dependencies (prevent conflicting dependencies).

### DIFF
--- a/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
+++ b/lib/Optimizer/CodeGen/QuakeToCodegen.cpp
@@ -97,9 +97,11 @@ public:
     auto statePtrTy = cudaq::cc::PointerType::get(stateTy);
     auto i8PtrTy = cudaq::cc::PointerType::get(rewriter.getI8Type());
     auto cast = rewriter.create<cudaq::cc::CastOp>(loc, i8PtrTy, buffer);
+    auto one = rewriter.create<arith::ConstantIntOp>(loc, 1, size.getType());
+    auto powsz = rewriter.create<arith::ShLIOp>(loc, size.getType(), one, size);
 
     rewriter.replaceOpWithNewOp<func::CallOp>(
-        createStateOp, statePtrTy, createStateFunc, ValueRange{cast, size});
+        createStateOp, statePtrTy, createStateFunc, ValueRange{cast, powsz});
     return success();
   }
 };


docs/sphinx/applications/python/quantum_pagerank.ipynb
```diff
@@ -13,16 +13,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip install torch==2.6.0 torch-geometric==2.6.1 -q"
```
khalatepradnya, line 16: Don't we need `torch-geometric`? @sacpis

docs/sphinx/applications/python/quantum_pagerank.ipynb
```diff
@@ -200,9 +190,18 @@
...
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 5,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
```
khalatepradnya: This may have been pushed in error? @sacpis

docs/sphinx/examples/python/dynamics/dynamics_intro_2.ipynb
```diff
@@ -552,7 +552,7 @@
     "for i in range(N):  \n",
     "    plt.plot(steps, exp_val[i],label=f'spin {i}')\n",
     "plt.legend(fontsize=12)\n",
-    "plt.ylabel(\"$\\sigma^z $\")\n",
+    "plt.ylabel(\"$\\\\sigma^z $\")\n",
```
khalatepradnya, line 555: If I understand correctly, this was the only change introduced, right? @sacpis

docs/sphinx/applications/python/quantum_pagerank.ipynb
sacpis: No.

docs/sphinx/examples/python/dynamics/dynamics_intro_2.ipynb
sacpis: yes

docs/sphinx/applications/python/quantum_pagerank.ipynb
sacpis: Removed it.

scripts/validate_container.sh
```diff
@@ -350,10 +350,50 @@ do
...
+    # Register the venv as a Jupyter kernel
+    # Notebooks will execute in this environment and can install their own packages
+    JUPYTER_KERNEL_NAME="cudaq_nb_validation_container"
+    python3 -m ipykernel install --user \
+        --name="$JUPYTER_KERNEL_NAME" \
+        --display-name="Python (CUDA-Q Container Notebook Validation)" \
+        2>/dev/null
+    
+    echo "Jupyter kernel '${JUPYTER_KERNEL_NAME}' registered."
+    echo "Notebooks will install their own dependencies during execution."
```
sacpis, line 379: I think this will be an overkill where these steps will be performed for each notebook for each run. Do we have any alternative here? Also, would you please remind me as why are we going with creating venv for each notebook?

sacpis (commented):
Overall this PR looks good. Just have one comment. Thanks @khalatepradnya.

scripts/validate_container.sh
khalatepradnya: Thank you for raising the concern. I introduced this change so as to run the notebooks in isolation and not pollute the environment, to eliminate that as the reason for original failures seen in the pipeline. I am okay reverting it if isn't beneficial and / or if you feel strongly against this approach.

sacpis: I don't see a strong reason to use venv. One more thing, we can do pip install just once for all notebooks. Lets move this pip install logic before the loop starts?

khalatepradnya: That install happens only once, before the `notebook_validation.py` is invoked.

sacpis: I think we are good to go then.

scripts/validate_pycudaq.sh
```diff
@@ -162,9 +168,21 @@ done
...
 # This is an optional integrator, which requires torch and torchdiffeq.
-# Install torch separately to match the cuda version.
+# Install torch separately to match the cuda version installed as CUDA-Q dependency.
 # Torch if installed as part of torchdiffeq's dependencies, may default to the latest cuda version. 
-python3 -m pip install torch --index-url https://download.pytorch.org/whl/cu$(echo $cuda_version | cut -d '.' -f-2 | tr -d .)
+ctk_version="$(python3 -m pip list | grep nvidia-cuda-runtime | tr -s " " | cut -d " " -f 2)" 
+torch_channel=""
+# For CUDA version 13, we need to use the nightly build.
+# This is required because the stable build (as of v2.9.0) depends on cuBlas 13.0,
+# while cuquantum-cu13 depends on cuBlas 13.1.
```
bmhowe23, line 177: Wow, I assume we need to make a similar change to CUDA-QX? Will this need to go int the docs?

docker/release/cudaq.wheel.Dockerfile
```diff
@@ -47,10 +47,12 @@ RUN cd /cuda-quantum && \
     esac
 
 # Create the README
+# Note: conda channel 13.0.0 does not contain CUDA 13 (still 12.x), so we map to 13.0.2
 RUN cd cuda-quantum && cat python/README.md.in > python/README.md && \
     package_name=cuda-quantum-cu$(echo ${CUDA_VERSION} | cut -d . -f1) && \
     cuda_version_requirement="\>= ${CUDA_VERSION}" && \
     cuda_version_conda=${CUDA_VERSION}.0 && \
+    cuda_version_conda=${cuda_version_conda/13.0.0/13.0.2} && \
```
bmhowe23, line 55: Wow this is getting complicated. Not for this PR, but I can't help but wonder if we should make a README.md.cu12 and README.md.cu13 file like was recently done for the pyproject.toml files.

scripts/validate_pycudaq.sh
1tnguyen: Definitely. I'll update the doc in a follow-up PR. Currently, it has a similar note about using the torch nightly channel for CUDA 12.8. I'll check whether it's still needed and add a note for CUDA 13.

docker/release/cudaq.wheel.Dockerfile
1tnguyen: Indeed, it might be a better long-term solution to have separate Python README files for each CUDA version.
We can experiment with the separate README files approach as a follow-up PR to see how it looks :)

khalatepradnya:
Merging to facilitate package creation...

scripts/validate_pycudaq.sh
bmhowe23: > while cuquantum-cu13 depends on cuBlas 13.1.

I do not see this dependency. When I install in a fresh environment, the pipdeptree that I see is this:
```
cuquantum-cu13==25.9.0
├── custatevec-cu13 [required: ==1.10.0, installed: 1.10.0]
├── cutensornet-cu13 [required: ==2.9.0, installed: 2.9.0]
│   └── cutensor-cu13 [required: >=2.3.1,<3, installed: 2.3.1]
└── cudensitymat-cu13 [required: ==0.3.0, installed: 0.3.0]
    ├── cutensor-cu13 [required: >=2.3.1,<3, installed: 2.3.1]
    └── cutensornet-cu13 [required: ~=2.9, installed: 2.9.0]
        └── cutensor-cu13 [required: >=2.3.1,<3, installed: 2.3.1]
```
Where do you see that cuquantum-cu13 requires cuBlas 13.1?

And the prerequisites that I see on https://docs.nvidia.com/cuda/cuquantum/latest/custatevec/index.html simply say 12.x and 13.x for CUDA 12 and 13, respectively.

bmhowe23: I do, on the other hand, see where *we*/CUDA-Q are declaring a dependency on 13.1, but perhaps we shouldn't be doing that?

https://github.com/NVIDIA/cuda-quantum/blob/9336ee256e4ab3e4f4f1f94b72b4f2ce0f5999cd/pyproject.toml.cu13#L25

bmhowe23: I think https://github.com/NVIDIA/cuda-quantum/pull/3589 needs to be incorporated to address the Torch issue. That allows us to use a released version of Torch rather than forcing our users to use a nightly version of Torch.

PR #3582, opened 2025-11-06
Fix bug related to incorrect vector size computation

schweitzpgi:
Codegen was computing the size of a veq incorrectly.


sacpis (approved):
LGTM. Thanks @schweitzpgi.

PR #3584, opened 2025-11-06
[core] Fix bug in sret type handling.

schweitzpgi:
Also add a `isa` type test for host side std::vector<bool> canonical type.


lib/Optimizer/Builder/Factory.cpp
```diff
@@ -357,6 +357,27 @@ static cc::StructType stlHostVectorType(Type eleTy) {
   return cc::StructType::get(ctx, ArrayRef<Type>{ptrTy, padout});
 }
 
+bool factory::isStlVectorBoolHostType(Type ty) {
```
sacpis, line 360: Just curious, where are we using it?

sacpis (approved):

lib/Optimizer/Builder/Factory.cpp
schweitzpgi: In code that hasn't been merged. :)

Issue #3581, opened 2025-11-06
Automatic selection of `RewritePattern`s in `decomposition` pass


lmondada:
### Required prerequisites

- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

Objective: specify declaratively the allowed quantum gate set in the hardware platforms' YAML config files, without having to list the actual rewrite patterns to be used.

## Context

The pass pipelines for individual hardware platforms is given in `.yml` files within the `runtime/cudaq/platform/default/rest/helpers` directory. One of the objectives of these pipelines is to decompose quantum gates into hardware-supported gates. The decomposition-specific passes are found within the `config.jit-mid-level-pipeline` property of the YAML, but other passes may be run at this step too (e.g. `lower-to-cfg`).

The gate decomposition is achieved by one of two passes:
1. DecompositionPass (aka `decomposition`). This pass takes a list of whitelisted and blacklisted patterns and applies all permissible RewritePatterns as many times as possible. This pass never fails and does not guarantee that all gates in the resulting program are contained within a target gate set. This pass is called directly from the YAML, for instance in [braket.yml](https://github.com/NVIDIA/cuda-quantum/blob/86e83e00b95b9ac458c64e74507ce8954eb3888e/runtime/cudaq/platform/default/rest/helpers/braket/braket.yml#L22).
2. BasisConversionPass (aka `basis-conversion`). On top of pattern whitelist and blacklist, this pass takes a `basis` argument that specifies the allowed set of quantum gates. The pass uses the permissible patterns to repeatedly rewrite ops, until all quantum ops are within the allowed gateset. If this cannot be achieved, the pass fails. This pass is not referenced directly from the YAML configs, but is used in the definitions of gate-set mapping pipelines such as `ionq-gate-set-mapping` used in [ionq.yml](https://github.com/NVIDIA/cuda-quantum/blob/86e83e00b95b9ac458c64e74507ce8954eb3888e/runtime/cudaq/platform/default/rest/helpers/ionq/ionq.yml#L21).

Currently, both passes require the user to specify the set of pattern rewrites to use. If patterns aren't explicitly whitelisted or blacklisted, there is a risk that cycles are formed in the rewrites, or that the passes' behavior changes as more patterns are added. 

## Proposal

[...]

PR #3587, opened 2025-11-07
[core] Add metadata map for decomposition patterns

lmondada:
### Description

This PR creates a system to register Decomposition patterns and associate metadata to them. This will be useful in an upcoming PR, where the metadata about what gate type is matched and what gate type is introduced by a rewrite will be used to select the subsets of pattern rewrites useful to target a specified target basis (see #3581).

#### With or without macro?

I tried to avoid using a macro here, but the same ~15 lines are repeated 31 times (once for each decomposition pattern). 
 A macro can remove this redundancy. What do you prefer?

To make reviewing easier, the first commit introduces all the changes without using a macro. The second commit then introduces the macro and removes redundancies. Getting rid of the macro is thus as simple as removing the second commit, if that's desired.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -454,12 +454,10 @@ jobs:
               echo "### Setting up OQC account" >> $GITHUB_STEP_SUMMARY
               echo "::add-mask::${{ secrets.OQC_URL }}"
               echo "::add-mask::${{ secrets.BACKEND_LOGIN_EMAIL }}"
-              echo "::add-mask::${{ secrets.OQC_PASSWORD }}"
               echo "::add-mask::${{ secrets.OQC_DEVICE }}"
               echo "::add-mask::${{ secrets.OQC_AUTH_TOKEN }}"
               echo "OQC_URL=${{ secrets.OQC_URL }}" >> $GITHUB_ENV
               echo "OQC_EMAIL=${{ secrets.BACKEND_LOGIN_EMAIL }}" >> $GITHUB_ENV
-              echo "OQC_PASSWORD=${{ secrets.OQC_PASSWORD }}" >> $GITHUB_ENV
               echo "OQC_DEVICE=${{ secrets.OQC_DEVICE }}" >> $GITHUB_ENV
               echo "OQC_AUTH_TOKEN=${{ secrets.OQC_AUTH_TOKEN }}" >> $GITHUB_ENV
               ;;
--- a/python/cudaq/dynamics/evolution.py
+++ b/python/cudaq/dynamics/evolution.py
@@ -544,7 +544,7 @@ def evolve_single_async(
         step_parameters, dt)
     if shots_count is None:
         shots_count = -1
-    if store_intermediate_results:
+    if store_intermediate_results != IntermediateResultSave.NONE:
         evolution = _evolution_kernel(
             num_qubits,
             compute_step_matrix,
@@ -579,12 +579,12 @@ def evolve_single_async(
             return cudaq_runtime.evolve_async(initial_state, kernel)
         # FIXME: permit to compute expectation values for operators defined as matrix
         if len(collapse_operators) > 0:
-            cudaq_runtime.evolve_async(initial_state,
-                                       kernel,
-                                       parameters[-1],
-                                       observable_spinops,
-                                       noise_model=noise,
-                                       shots_count=shots_count)
+            return cudaq_runtime.evolve_async(initial_state,
+                                              kernel,
+                                              parameters[-1],
+                                              observable_spinops,
+                                              noise_model=noise,
+                                              shots_count=shots_count)
         return cudaq_runtime.evolve_async(initial_state,
                                           kernel,
                                           parameters[-1],
--- a/python/tests/dynamics/test_evolve_simulators.py
+++ b/python/tests/dynamics/test_evolve_simulators.py
@@ -331,6 +331,134 @@ def test_evolve_async():
                                atol=0.1)
 
 
+def test_evolve_no_intermediate_results():
+    """Test evolve with store_intermediate_results=NONE 
+    to verify the else branch in evolve_single is working."""
+
+    # Qubit Hamiltonian
+    hamiltonian = 2 * np.pi * 0.1 * spin.x(0)
+
+    # Dimensions
+    dimensions = {0: 2}
+
+    # Initial state
+    rho0 = cudaq.State.from_data(
+        np.array([[1.0, 0.0], [0.0, 0.0]], dtype=np.complex128))
+
+    # Schedule
+    steps = np.linspace(0, 10, 101)
+    schedule = Schedule(steps, ["time"])
+
+    # Test 1: NONE without observables
+    evolution_result = cudaq.evolve(
+        hamiltonian,
+        dimensions,
+        schedule,
+        rho0,
+        store_intermediate_results=cudaq.IntermediateResultSave.NONE)
+
+    # NONE mode: only final state is saved, no intermediate states
+    assert len(evolution_result.intermediate_states()) == 1
+
+    # Test 2: NONE with observables
+    schedule.reset()
+    evolution_result = cudaq.evolve(
+        hamiltonian,
+        dimensions,
+        schedule,
+        rho0,
+        observables=[spin.y(0), spin.z(0)],
+        store_intermediate_results=cudaq.IntermediateResultSave.NONE)
+
+    # Verify final expectation value is reasonable
+    final_exp = evolution_result.expectation_values()
+    assert final_exp is not None
+
+    # Test 3: NONE with collapse_operators (tests the missing return bug)
+    schedule.reset()
+    evolution_result_decay = cudaq.evolve(
+        hamiltonian,
+        dimensions,
+        schedule,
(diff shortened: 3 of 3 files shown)


lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -1668,58 +1683,121 @@ struct U3ToRotations : public OpRewritePattern<quake::U3Op> {
...
+  /// - the target gates, i.e. the gates that the pattern creates. This must
+  ///   include all gates that may be created by any instance of the pattern.
+  static const std::map<StringRef, PatternMetadata> patternMetadata{
+      // TODO: "SToR1", "TToR1", "R1ToU3", "U3ToRotations" can be generalised to
+      // arbitrary number of controls, but we would need to reason over n
+      // HOp patterns
+      {"HToPhasedRx", makePatternMetadata<HToPhasedRx>("h", {"phased_rx"})},
+      {"CHToCX", makePatternMetadata<CHToCX>("h(1)", {"s", "h", "t", "x(1)"})},
+      // SOp patterns
+      {"SToPhasedRx", makePatternMetadata<SToPhasedRx>("s", {"phased_rx"})},
```
lmondada: As mentioned at the top, ideally this metadata would be parsed directly from the `struct` definitions.
1. I could derive the debug name by creating an empty MLIR context and a dummy instance of the pattern and reading the name from that. Is that a good idea?
2. I don't know if there is a way to infer the `Op` that a pattern matches?
3. To infer the `Op` that a rewrite introduces, there is a hacky solution, as I did in the test. However, this would not catch cases where the decomposition might changed based on parameters of the matched op. It is also quite verbose to create a "dummy module" on which the rewrite applies. Overall, this seems like a bad idea.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -1668,58 +1683,121 @@ struct U3ToRotations : public OpRewritePattern<quake::U3Op> {
...
+}
+const std::map<StringRef, PatternMetadata> &getAllPatternMetadata() {
+  /// Each entry documents:
+  /// - the debug name of the pattern (must match getDebugName() of the
+  /// pattern),
+  /// - the source gate, i.e. the gate that the pattern matches and decomposes,
+  /// - the target gates, i.e. the gates that the pattern creates. This must
+  ///   include all gates that may be created by any instance of the pattern.
+  static const std::map<StringRef, PatternMetadata> patternMetadata{
+      // TODO: "SToR1", "TToR1", "R1ToU3", "U3ToRotations" can be generalised to
```
lmondada: The pattern rewrites enumerated in the TODO also match on the `Op`s when they have an unbounded number of controls. However, reasoning about this is hard: we would need to introduce functional relationships between `n` the number of controls in the matched op and `m`, the number of controls on (one of) the ops in the output ops.

I've simplified this here by only advertising `n=0` in these metadata properties.

unittests/Optimizer/DecompositionPatternsTest.cpp
```diff
@@ -0,0 +1,312 @@
...
+    // All target gates should be non-empty
+    for (auto targetGate : targetGates) {
+      EXPECT_FALSE(targetGate.empty()) << "Pattern '" << patternName.str()
+                                       << "' has empty target gate in list";
+    }
+  }
+}
+
+// Test 3: Verify pattern decompositions produce only target gates
+TEST_F(DecompositionPatternsTest, DecompositionProducesOnlyTargetGates) {
```
lmondada, line 284: This test is quite verbose/ugly (see also `createTestModule`). Is there a better way to write it?

lmondada:
> /ok to test f7bd459d61ba6bfcd3b130b4e593125655928652

lib/Optimizer/Transforms/DecompositionPatterns.cpp
lmondada: this is outdated, the new version fixes most of these issues.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
lmondada: This was discussed and agreed in the specs proposal.

unittests/Optimizer/DecompositionPatternsTest.cpp
```diff
@@ -0,0 +1,331 @@
...
+  if (lastColon != llvm::StringRef::npos) {
+    debugName = debugName.substr(lastColon + 1);
+  }
+}
+
+} // namespace
+
+// Test 1: Verify the total number of registered decomposition patterns
+TEST_F(DecompositionPatternsTest, TotalPatternCount) {
+  auto patternNames = cudaq::DecompositionPatternType::RegistryType::entries();
```
sacpis: Wondering if we can fetch the entries just once and use it in all tests, as we are not changing anything in it?

unittests/Optimizer/DecompositionPatternsTest.cpp
```diff
@@ -0,0 +1,331 @@
...
+    // Collect all gates in the output
+    auto outputGates = collectGateTypesInModule(module);
+
+    auto isAllowedGate = [&](StringRef gate) {
+      // Split gate into prefix and number (e.g., "h(1)" -> "h", 1) using
+      // utility function
+      auto [gatePrefix, gateNum] = splitGateAndControls(gate);
+
+      for (auto targetGate : targetGates) {
+        auto [tPrefix, tNum] = splitGateAndControls(targetGate);
```
sacpis: Just a thought. What if we already store targetGates in a map with prefix as key and num as value? That way we don't have to iterate the list.

sacpis:
Seems like few tests are failing in CI.
https://github.com/NVIDIA/cuda-quantum/actions/runs/19477582015/job/55742702956?pr=3587#step:9:2070

sacpis:
Adding @khalatepradnya to take a look at it as well.

lmondada:
> Seems like few tests are failing in CI. https://github.com/NVIDIA/cuda-quantum/actions/runs/19477582015/job/55742702956?pr=3587#step:9:2070

Indeed! My guess is that this is due to a change in ordering of the patterns in the `RewritePatternSet`. I'll look into it.

EDIT: Resolved. That was indeed what was happening. To avoid future issues I have ordered the patterns lexicographically by their names.

unittests/Optimizer/DecompositionPatternsTest.cpp
lmondada: 👍 have implemented your suggestion.

unittests/Optimizer/DecompositionPatternsTest.cpp
lmondada: Yes we can, but in this case I don't think it's useful: `entries()` returns an iterator, which from the implementation seems very cheap (pretty much free) to construct.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -283,6 +308,36 @@ LogicalResult checkAndExtractControls(quake::OperatorInterface op,
...
+      std::unique_ptr<mlir::RewritePattern> pattern =                          \
+          RewritePattern::create<PATTERN>(context, benefit);                   \
+      return pattern;                                                          \
+    }                                                                          \
+  };                                                                           \
+  CUDAQ_REGISTER_TYPE(cudaq::DecompositionPatternType, PATTERN##Type, PATTERN)
+
+// TODO: The decomposition patterns "SToR1", "TToR1", "R1ToU3", "U3ToRotations"
+// can handle arbitrary number of controls, but currently metadata cannot
+// capture this. The pattern types therefore only advertise them for 0 controls.
```
bettinaheim, line 339: Why for 0 controls instead of for 1 control?

unittests/Optimizer/DecompositionPatternsTest.cpp
```diff
@@ -0,0 +1,344 @@
...
+  }
+  Value target = entry->getArgument(numControls);
+
+  // Create the gate operation based on gate name
+  Location loc = builder.getUnknownLoc();
+
+  Value pi_2 = cudaq::opt::factory::createFloatConstant(loc, builder, M_PI_2,
+                                                        builder.getF64Type());
+
+  if (gateName == "h") {
```
bettinaheim, line 115: Exp_pauli is treated differently than other decompositions right now, but we should probably reexamine that. It is one of the built-in gates, and correspondingly needs to be mapped...

lmondada: Having another look at this today, I realized it was fairly straightforward to support `exp_pauli` as well! I've expanded the test accordingly :)

lib/Optimizer/Transforms/DecompositionPatterns.cpp
lmondada: This is based on the uses of the patterns that I have seen: they are used to decompose `S`, `T` and `U3` to rotations. I couldn't check exhaustively, though. In case it turns out that these decompositions are also required for their singly-controlled versions (i.e. Control-S, Control-T and Control-U3), then we will have to register multiple versions of the pattern: one for 0 controls and one for 1 control.

Issue #2194, opened 2024-09-06
The quick start installation guide does not work for zsh (and likely other shells)


mitchdz:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The [quick start installation guide](https://nvidia.github.io/cuda-quantum/latest/using/quick_start.html#install-cuda-q) says to install and use cuda-quantum like so:

```bash
sudo -E bash install_cuda_quantum.$(uname -m) --accept
. /etc/profile
```

This can cause issues if you are not using bash, as once `install_cuda_quantum.$(uname -m)` finishes running, you are returned to the calling shell. In my case, I am using zsh. The issues occurs because `/opt/nvidia/cudaq/set_env.sh` uses `${BASH_SOURCE[0]}` which doesn't work well when using zsh.

### Steps to reproduce the bug

Run a shell other than bash (such as zsh) and then attempt to source the environment file.

```bash
$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
$ cat /etc/profile | tail -n 2
CUDAQ_INSTALL_PATH="/opt/nvidia/cudaq"
. "${CUDAQ_INSTALL_PATH}/set_env.sh"
$ . /etc/profile
$ echo $PATH
./bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

### Expected behavior

```bash
$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
$ cat /etc/profile | tail -n 2
CUDAQ_INSTALL_PATH="/opt/nvidia/cudaq"
. "${CUDAQ_INSTALL_PATH}/set_env.sh"
[...]

PR #3592, opened 2025-11-10
Support for zsh in cudaq_set_env.sh

### LINKED ISSUE #3476: setenv.sh only works in bash
### Required prerequisites

### Describe the bug

As the title says, this does not work in other shells.

Install cudaq, open zsh shell, run `source /opt/nvidia/cudaq/set_env.sh` and `echo $PATH`. `"./bin"` will have been prepended. 

### Expected behavior

It works as it does in bash in zsh.

Not a regression

### Environment

- **CUDA-Q version**: 0.12.0
- **Python version**: n/a
- **C++ compiler**: n/a

### Suggestions

I suggest something more portable such as `export CUDA_QUANTUM_PATH="$(realpath "$(dirname "$0")")"`

lmondada:
### Description

Support sourcing the `scripts/cudaq_set_env.sh` from non-bash shells by falling back to `$0` if `BASH_SOURCE[0]` is not defined.

Closes #3476

### DIFF
--- a/docs/sphinx/specification/cudaq/platform.rst
+++ b/docs/sphinx/specification/cudaq/platform.rst
@@ -12,8 +12,12 @@ information and enable asynchronous quantum kernel invocations.
 **[3]** The :code:`cudaq::quantum_platform` provides an API for querying the number of available 
 quantum processing units (QPUs), with each QPU assigned a logical integer index (:code:`{0,1,2,...}`). 
 
-**[4]** Programmers can specify the ID of the desired QPU and all subsequent CUDA-Q 
-kernel executions will target that QPU.
+**[4]** The properties of the QPUs on the platform are exposed by a collection
+of functions such as :code:`get_num_qubits`, :code:`is_simulator`, :code:`is_remote`, :code:`is_emulated`, etc.
+They take an optional :code:`qpu_id` argument (defaults to 0) to specify the QPU of interest.
+
+**[5]** The ID of the desired QPU can be specified as argument when invoking
+kernel functions such as :code:`cudaq::sample_async`, :code:`cudaq::run_async`, etc.
 
 The :code:`cudaq::quantum_platform`  should take the following structure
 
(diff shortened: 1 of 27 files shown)


lmondada:
I don't understand what/why CI is failing... any ideas?

sacpis:
> I don't understand what/why CI is failing... any ideas?

Seems like rerun will pass.

mitchdz:
This fixes https://github.com/NVIDIA/cuda-quantum/issues/2194

scripts/cudaq_set_env.sh
```diff
@@ -11,7 +11,7 @@
 
 # Prefix all paths such that running this script manually for a local installation
 # ensures that the local installation takes precedence over a system-wide installation.
-export CUDA_QUANTUM_PATH=`dirname "$(readlink -f "${BASH_SOURCE[0]}")"`
+export CUDA_QUANTUM_PATH=`dirname "$(readlink -f "${BASH_SOURCE[0]:-$0}")"`
```
sacpis, line 14: I do see `Bad Substitution error` when I run this in zsh

```
./scripts/cudaq_set_env.sh: 1: Bad substitution
```

So I tried the following code and it works both in bash and zsh

```
if [ -n "${BASH_VERSION-}" ]; then
        script_path="${BASH_SOURCE[0]}"
elif [ -n "${ZSH_VERSION-}" ]; then
        script_path="${(%):-%N}"
else
        script_path="$0"
fi

export CUDA_QUANTUM_PATH="$(dirname -- "$script_path")"
```

lmondada: Sorry for the late reply!

I checked again, the current syntax should work (works on bash and zsh, just tried it).

Could it be that by running it with `./scripts/cudaq_set_env.sh` you used `sh` or some other shell? This script only makes sense to be used as

```
source scripts/cudaq_set_env.sh
```

as it defines variables that must be set in the current shell. If you try that (or also `bash scripts/cudaq_set_env.sh` or `zsh scripts/cudaq_set_env.sh`) it should work.

Let me know!

sacpis: Thanks @lmondada. It works for me.

PR #3594, opened 2025-11-10
Fixing tests that accidentally resulted in a crash and incorrectly passing pipelines

bettinaheim:
This particular issue resulted in an LLVM assert being hit, and mlir tests hence silently crashing such that a good number of tests were simply skipped in our pipelines...


bettinaheim:
> /ok to test 8f772f0

Issue #3598, opened 2025-11-11
Disabling Anyon nightly integration test


sacpis:
Disabling Anyon nightly integration test as their server is non-functional at the time.
Here is the API we are using in the nightly integration test
https://api.anyon.cloud:5000/login

PR #3599, opened 2025-11-11
Disabling anyon nightly test

sacpis:
Disabling anyon nightly test. Please see issue #3598.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1021,6 +1021,7 @@ jobs:
           fetch-depth: 1
 
       - name: Install wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -16,7 +16,8 @@ on:
         type: choice
         options:
           - nightly
-          - anyon
+          # Issue 3598 (https://github.com/NVIDIA/cuda-quantum/issues/3598)
+          # - anyon
```
1tnguyen: I think we should keep this option available for manual execution, e.g., to check if the service is back at some point in the future.

sacpis: Fixed in 45963da.

PR #3610, opened 2025-11-13
Adding venv and fixing regex

sacpis:
Adding venv and fixing regex

### DIFF
--- a/docs/notebook_validation.py
+++ b/docs/notebook_validation.py
@@ -142,7 +142,9 @@ if __name__ == "__main__":
 
         ## `quantum_transformer`:
         ## See: https://github.com/NVIDIA/cuda-quantum/issues/2689
-        notebooks_skipped = ['quantum_transformer.ipynb', 'logical_aim_sqale.ipynb']
+        notebooks_skipped = [
+            'quantum_transformer.ipynb', 'logical_aim_sqale.ipynb'
+        ]
 
         for notebook_filename in notebook_filenames:
             base_name = os.path.basename(notebook_filename)


.github/workflows/docker_images.yml
```diff
@@ -841,6 +841,7 @@ jobs:
           # In containers without GPU support, UCX does not work properly since it is configured to work with GPU-support.
           # Hence, don't enforce UCX when running these tests.
           docker exec cuda-quantum bash -c "python3 -m pip install --break-system-packages pandas scipy seaborn h5py contfrac"
+          docker exec cuda-quantum bash -c "sudo apt install -y python3.12-venv"
```
1tnguyen: Q: Should we be specific about the Python version here? i.e., not `python3-venv` but specifically 3.12

docs/notebook_validation.py
```diff
@@ -25,13 +25,22 @@ def read_available_backends():
     return [backend.strip() for backend in available_backends]
 
 
+pattern = r'set_target\(\s*\\"([^"]+)\\"(?:\s*,\s*option\s*=\s*\\"([^"]+)\\")?'
```
1tnguyen: nit: I think a comment here about the matching criteria would be helpful, like a couple of examples of the intended matches.

docs/notebook_validation.py
```diff
@@ -25,13 +25,22 @@ def read_available_backends():
...
-        if match and (match.group(1) not in available_backends):
-            return False
+        if re.search(r'\s*\#', notebook_content):
+            continue
+
+        match = re.search(pattern, notebook_content)
+        if match:
+            target = match.group(1)
+            opt = match.group(2)
+            combined = f"{target}-{opt}" if opt else target
```
1tnguyen: I think this `{target}-{opt}` may not be correct in all cases, e.g., it would break for multiple target options, such as `mgpu,fp64`. 
Additionally, those combined targets, e.g., `nvidia-fp64`, are deprecated, so we shouldn't rely on them.

I suggest that we keep the target name as the matching criteria.

sacpis: Ah, good to know. Will rely only on

```
set_target(\"target\"...
set_target(\'target\'...
```

.github/workflows/docker_images.yml
sacpis: The error specifically complains about python3.12

```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.12-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: /tmp/cudaq_notebook_validation_venv/bin/python3

./validate_container.sh: line 363: /tmp/cudaq_notebook_validation_venv/bin/activate: No such file or directory

```

docs/notebook_validation.py
sacpis: Fixed in [61318d9](https://github.com/NVIDIA/cuda-quantum/pull/3610/commits/61318d92e8d19b444fc157f50574e4ba8b9165fa).

docs/notebook_validation.py
sacpis: Fixed in [61318d9](https://github.com/NVIDIA/cuda-quantum/pull/3610/commits/61318d92e8d19b444fc157f50574e4ba8b9165fa).

sacpis:
Note: Have also added support for single quotes in Fixed in [61318d9](https://github.com/NVIDIA/cuda-quantum/pull/3610/commits/61318d92e8d19b444fc157f50574e4ba8b9165fa).

.github/workflows/docker_images.yml
sacpis: Fixed in [895d674](https://github.com/NVIDIA/cuda-quantum/pull/3610/commits/895d674468dfc475e6f0891243f6746e8983cf3f).

1tnguyen (approved):
LGTM 👍 Thanks, @sacpis!

PR #3615, opened 2025-11-15
Enable anyon test

sacpis:
Enable anyon test. The server URL was down due to a power outage issue. It is back online now.

Fixes issue #3598.


1tnguyen (approved):
👍

PR #3298, opened 2025-08-04
Support dynamic quantum architecture for IQM QPUs

iqm-bhoffmann:
### Description
For IQM machines fetch the quantum architecture from server, derive the dynamic architecture from this and write it to an architecture file. This file is then passed in the lowering pipeline in place of the previously used static architecture files.

The output from the algorithm for finding usable qubits generates a list of qubits which are calibrated for mz, cz and prx. A mapping ensures that qubit names are enumerated in a linear manner and not fully calibrated qubits do not show in the architecture file. The architecture file is created in the system temp directory with a unique filename and deleted after the job has been sent to the IQM server.

#### Changes in function
* Removed parameter "--iqm-machine" which was used to specify the IQM quantum architecture at compile time.
* Removed static architecture files provided for Adonis, Apollo, and Aphrodite architectures.
* Removed machine architecture bindings from IQM tests.
* Implemented getting the authorization token from IQM_TOKEN environment variable. If this is not set the previous mechanism reading the token from a JSON file is used.
* Updated IQM JSON output of LLVM to latest syntax. "phased_rx" -> "prx", "measurement" -> "measure" in both emitter code and tests.

#### Test related changes
* Adapted IQM mock server to latest iqmclient API version (v28.0.0) and IQM server API changes.
* Implemented in IQM mock server the 20 qubit Apollo quantum architecture.
* Added support for mapping of qubits syntax in the job request to the mock server.
* In the mock server deliberately removed QB2 and QB3 from the list of PRX calibrated gates. This results in changes in the topology and forces the CUDA-Q code to work with an imperfect calibrated system. This tests the implemented algorithm with every testcase which uses more than 1 qubit.
* In mock server implemented a full checking of gates used in circuits versus the topology to verify that all 2 qubit gates are supported by the mock qpu.
* Pass the authorization token to IQM mock server with env-variable. This includes also a new testcase addressing the IQM_TOKEN environment variable.
* Save and restore environment variables for testcases modifying them. So the order of testcases does not matter anymore.
* Fixed evaluation of 2 qubit testcases which failed on some simulations.

#### Documentation
* Updated the IQM parts of the documentation and examples. Explained changes from previous version and removed references to the no longer needed machine names.
* Cleanup: improved doxygen headers and few other comments.
* Cleanup: Added copyright to touched files

#### Addressed issues
NVIDIA#1589, NVIDIA#865

PR #3619, opened 2025-11-17
Follow-up to PR #3298: Update the version of `iqm-client` in Python test environment

khalatepradnya:
Follow-up to PR #3298


sacpis:
Seems like we would need code owner's approval here.

PR #3608, opened 2025-11-12
Improve validate_pycudaq.sh script

huaweil-nv:
- Simplify cudaq installation using --find-links instead of pypi-server (install python 3.13 wheels failed with pypi-server)
- Add execution logging for all example and snippet runs
- Add API key checks for multiple quantum backends (fermioniq, qci, oqc, nvqc, pasqal, ionq)

PR #3621, opened 2025-11-17
Fix metapackages

sacpis:
Fix for metapackages:

- Put `find-links` in the proper place.

Related to: https://github.com/NVIDIA/cuda-quantum/pull/3608, which only works for newer cuda-q packages. In CI, we install a dummy (0.0.0) version. 

- Increase the `remote-mqpu` server start-up delay from 5 to 20 secs. This is consistent with the value used for mock servers and would be more stable on various platforms (slow HTTP server start-up).


1tnguyen:
Replaced by #3623. Keep the `test-publishing` branch for other debug activities.

PR #3622, opened 2025-11-17
Simplifying cudaq::state support.

schweitzpgi:
This PR greatly simplifies the use of cudaq::get_state by limiting its applicability strictly to local simulation execution contexts. Prior attempts were made to make cudaq::get_state generally available regardless of the target execution context. This level of support is being discontinued.

- Move the qvector ctor overloads to the cudaq::state class. This simplifies qvector by making it only possible to initialize a qvector with a cudaq::state object and dropping all the overloads.
- Fix tests.  Adapt the regression tests.
- Drop state with non-simulation targets.
- Add code to distinguish between floats and complex floats and get rid of a bunch of redundant code that wasn't needed. Also, update tests.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_draw.cpp
+++ b/python/runtime/cudaq/algorithms/py_draw.cpp
@@ -125,7 +125,9 @@ Returns:
   # Output
   #      ╭───╮╭──────────╮
   # q0 : ┤ h ├┤ ry(0.59) ├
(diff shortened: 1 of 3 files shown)


sacpis:
Seems like we have a few test failures.

schweitzpgi:
Abandoning this. This work will be merged as part of #3693

PR #3624, opened 2025-11-19
[sample] [C++] Updated kernels to not have user-specified measurement ops

khalatepradnya:
This PR refactors several C++ example and application files using `sample` API by removing explicit measurement instructions (`mz(...)` / `my(...)` / `mx(...)`) from quantum kernels and instead extracting measurement results using marginalization (`get_marginal`) or direct return values (`run`). 

This demonstrates best practices for using `sample` as we introduce scoping changes.

* Removed explicit measurement operations from quantum kernels in multiple files (e.g., `bernstein_vazirani.cpp`, `grover.cpp`, `ghz` examples, and noise modeling examples), relying on default measurement behavior or returning measurement results directly. 

* Updated main functions to extract and analyze measurement results using `get_marginal` or by using `run`, ensuring only relevant qubits are considered in the analysis. 

* Improved examples to show best practices for extracting and displaying measurement statistics.

* Identified examples with mid-circuit measurements that cannot (yet) be converted to `run` to emit `qir-adaptive` (Work on simulators)

- [ ] Fix the resource estimation tests


lib/Optimizer/Transforms/QuakeAddMetadata.cpp
```diff
@@ -99,6 +100,13 @@ struct QuakeFunctionAnalysis {
     LLVM_DEBUG(llvm::dbgs()
                << "Function to analyze: " << funcOp.getName() << '\n');
     QuakeMetadata data;
+
+    // Check for any measurements
+    funcOp->walk([&](quake::MeasurementInterface meas) {
```
khalatepradnya: Note: This will be removed. To be replaced with verifier logic on the kernel ....

khalatepradnya:
This PR served as means to demonstrate how many existing examples & tests would be affected by disallowing any measurement operations in sampled kernels.

PR #3625, opened 2025-11-19
Python bridge revision - part 2

bettinaheim:
Contains the bridge changes to introduce a proper value stack and revise the call visit.


bettinaheim:
> /ok to test f15e7cb

PR #3627, opened 2025-11-19
Deactivate testcase graph_coloring-1 for IQM backend

iqm-bhoffmann:
### Description

Testcase seems to be responsible for an increase of testtime by on hour on some architectures when executed for IQM QA.
Reported by @sacpis in comments to PR #3298. Please see: https://github.com/NVIDIA/cuda-quantum/pull/3298#issuecomment-3550908135

### DIFF
--- a/targettests/execution/graph_coloring-1.cpp
+++ b/targettests/execution/graph_coloring-1.cpp
@@ -9,7 +9,6 @@
 // REQUIRES: c++20
 // clang-format off
 // RUN: nvq++ %s -o %t --target infleqtion --emulate && %t | FileCheck %s
-// RUN: nvq++ %cpp_std --target iqm        --emulate %s -o %t && IQM_QPU_QA=%iqm_tests_dir/Crystal_20.txt %t | FileCheck %s
 // RUN: nvq++ %s -o %t --target quantinuum --emulate && %t | FileCheck %s
 // RUN: if %braket_avail; then nvq++ %s -o %t --target braket --emulate && %t | FileCheck %s; fi
 // RUN: if %qci_avail; then nvq++ %cpp_std --target qci --emulate %s -o %t && %t | FileCheck %s; fi


sacpis:
> /ok to test 53a60ed

sacpis (approved):
Thanks @iqm-bhoffmann.

sacpis:
We will get to know the test time from this PR itself.

sacpis:
`graph_coloring-1.cpp` has executed in 

`137.86s: CUDAQ-Target :: execution/graph_coloring-1.cpp`

sacpis:
Just tried with Crystal_5 arch, and we see this

```
+ IQM_QPU_QA=/workspaces/cuda-quantum/targettests/Target/IQM/Crystal_5.txt
+ /workspaces/cuda-quantum/build/targettests/execution/Output/graph_coloring-1.cpp.tmp
+ /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/graph_coloring-1.cpp
graph_coloring-1.cpp.tmp: /usr/local/llvm/include/llvm/ADT/SmallVector.h:294: T& llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::operator[](llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::size_type) [with T = quake::BorrowWireOp; <template-parameter-1-2> = void; llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::reference = quake::BorrowWireOp&; llvm::SmallVectorTemplateCommon<T, <template-parameter-1-2> >::size_type = long unsigned int]: Assertion `idx < size()' failed.
FileCheck error: '<stdin>' is empty.
FileCheck command line:  /usr/local/llvm/bin/FileCheck /workspaces/cuda-quantum/targettests/execution/graph_coloring-1.cpp
```

sacpis:
Seems like the issue is only with the arch Crystal_20. Crystal_54 arch works fine. Would you like to update this PR with Crystal_54 arch?

Crystal_54 arch takes around 500 seconds.

`587.05s: CUDAQ-Target :: execution/graph_coloring-1.cpp`

Or

We can debug the issue with the slowness and add this arch to this test later.

iqm-bhoffmann:
Hello @sacpis ,
I have several results and a conclusion.

1. As this algorithm uses 8 qubits +1 ancilla it will not work on a 5 qubit architecture. That this fails only with an assert without a clear error indication is certainly something which needs some attention. I think this condition should be checked where the algorithm and the architecture come first together. So the input to the transpilation.
2. Out quantum engineers explained to me that the algorithm is good for star architectures as it connects all qubits with almost all other qubits. But it is not suitable for our crystal architectures as it results in a huge number of swaps there. As result the job is very long and takes a long time to calculate. This is a systematic problem which cannot be resolved.
3. I ran the jobs from this algorithm on our Resonance machines for 20 and 54 qubits. But the length of the job exceeds the coherence times and so expectedly the test fails.
4. While testing yesterday I did not run the correct binary and so my results were all wrong. I now have execution times of 226 s - 262 s for 20 qubit and 153 s - 160 s for 54 qubit. This is running within the docker image on my local machine.

Conclusion: This test should not run on a superconducting crystal architecture. So deactivating it again for the IQM backend is the correct and final solution.

sacpis:
Perfect. Thank you very much @iqm-bhoffmann.
Adding @schweitzpgi fir viz.

PR #3632, opened 2025-11-24
Removing OQC env var

sacpis:
Removing OQC_PASSWORD, as we are now using OQC_AUTH_TOKEN.


sacpis:
> /ok to test 36cd812

PR #3633, opened 2025-11-25
Removing NVQC target and associated functions

sacpis:
Removing NVQC target and associated functions as we will be using [DGX cloud](https://nvidia.github.io/cuda-quantum/latest/using/install/local_installation.html#dgx-cloud).

This includes removing
1. Classes supporting NVCF funcitons
2. Serailiazed code execution
3. VQE optimization done for NVQC

Note: We would like to retain the code which converts kernel as a string. This is required for calling problematically generated or registered operations inside a kernel.

### DIFF
--- a/docs/sphinx/using/quick_start.rst
+++ b/docs/sphinx/using/quick_start.rst
@@ -16,10 +16,7 @@ found in our complete :doc:`Installation Guide <install/install>`.
 
 .. note::
 
-    If you are using CUDA 12.8 on Blackwell, you may need to install additional
-    dependencies to use the python wheels.
-
-    See :ref:`Blackwell CUDA Dependencies <blackwell-cuda-dependencies>` for more information.
+    CUDA-Q is currently supported on Linux only. On Windows, you can use `Windows Subsystem for Linux (WSL) <https://learn.microsoft.com/en-us/windows/wsl/>`_ to install CUDA-Q.
 
 Install CUDA-Q
 ----------------------------
@@ -35,7 +32,13 @@ Install CUDA-Q
       :start-after: (Begin complete install)
       :end-before: (End complete install)
 
-   Once you completed the installation, please follow the instructions
+.. note::
+
+    If you are using CUDA 12.8 on Blackwell, you may need to install additional
+    dependencies to use the python wheels.
+    See :ref:`Blackwell CUDA Dependencies <blackwell-cuda-dependencies>` for more information.
+
+Once you completed the installation, please follow the instructions
    :ref:`below <validate-installation>` to run your first CUDA-Q program!
 
 .. tab:: C++


sacpis:
> /ok to test e5a8d42

sacpis:
> /ok to test ad12622

sacpis:
> /ok to test f2be31b

python/runtime/cudaq/algorithms/py_optimizer.cpp
```diff
@@ -21,41 +21,42 @@
...
-    auto var_dict = get_serializable_var_dict();
-    ctx.scoped_var_dict = py::str(json.attr("dumps")(var_dict));
-    ctx.source_code = source_code;
-  } catch (py::error_already_set &e) {
-    throw std::runtime_error("Failed to serialized data: " +
-                             std::string(e.what()));
-  }
-  return ctx;
-}
+// static SerializedCodeExecutionContext
```
1tnguyen: Maybe we can just delete rather than comment out the code?

tools/cudaq-qpud/README.md
```diff
@@ -1,102 +1,22 @@
 <!-- markdownlint-disable MD013 -->
-# `remote-mqpu` / `nvqc` Debugging Tips
+# `remote-mqpu` Debugging Tips
 
 This file contains tips and tricks for when you are performing manual testing/
-debugging for `remote-mqpu` or `nvqc` targets. This file is primarily intended
+debugging for `remote-mqpu` targets. This file is primarily intended
```
1tnguyen: perhaps `targets` -> `target`?

runtime/common/SerializedCodeExecutionContext.h
1tnguyen, line 1: Q: Do we still need this class?

tools/cudaq-qpud/README.md
sacpis: Fixed in cb15a34.

python/runtime/cudaq/algorithms/py_optimizer.cpp
sacpis: Fixed in cb15a34.

runtime/common/SerializedCodeExecutionContext.h
sacpis: Now we don't as have removed the commented out code.

sacpis: Fixed in 013cefa.

python/runtime/cudaq/algorithms/py_optimizer.cpp
sacpis: We would like to retain the code which converts kernel as a string. This is required for calling problematically generated or registered operations inside a kernel.

@bettinaheim for viz.

Let me know if we need to add this code back.

1tnguyen (approved):
LGTM 👍

PR #3636, opened 2025-11-25
Fix some bugs in evolve_async related to storing intermediate results

huaweil-nv:
### Description

When checking the coverage of evolution.py (http://porting-server/cuda-quantum/20250924/cudaq/pycoverage/python--cudaq--dynamics--evolution.py.html#t570), it was found that this part of the test is missing.
Bug fixes in evolution.py:
- Fix missing return statement in evolve_single_async when collapse_operators is provided with store_intermediate_results=NONE mode
- Fix condition check: use 'store_intermediate_results != NONE' instead of boolean check on enum value

Test improvements:
- Add test_evolve_no_intermediate_results and test_evolve_async_no_intermediate_results tests

### DIFF
--- a/python/cudaq/dynamics/evolution.py
+++ b/python/cudaq/dynamics/evolution.py
@@ -544,7 +544,7 @@ def evolve_single_async(
         step_parameters, dt)
     if shots_count is None:
         shots_count = -1
-    if store_intermediate_results:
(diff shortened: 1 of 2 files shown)


1tnguyen:
> /ok to test 30a0e00e200aa2df1a36b72a5eb5fdf29e637506

PR #3646, opened 2025-12-03
Adding 0.13.0 to the releases

sacpis:
Adding 0.13.0 to the releases


docs/sphinx/releases.rst
khalatepradnya: The older release will no longer be visible on the top landing page but can be found under `tags` tab.

```suggestion
`Docker image <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/quantum/containers/cuda-quantum/tags>`__
```

Issue #3073, opened 2025-06-17
pyproject.toml license needs to a valid SPDX license expression


mitchdz:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

See https://github.com/NVIDIA/cuda-quantum/actions/runs/15632777958/job/44068729228#step:7:8149
```
  /tmp/pip-build-env-bvgfs9gp/overlay/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
  !!

          ********************************************************************************
          Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

          By 2026-Feb-18, you need to update your project and remove deprecated calls
          or your builds will no longer be supported.

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
    corresp(dist, value, root_dir)

```

### Steps to reproduce the bug

N/A

### Expected behavior

N/A

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

mitchdz:
This isn't an issue until February, but I would like to fix this after deployments are working well.

PR #3648, opened 2025-12-03
update to plain SPDX string for license

mitchdz:
Fixes https://github.com/NVIDIA/cuda-quantum/issues/3073


mitchdz:
wait for https://github.com/NVIDIA/cuda-quantum/actions/runs/19882314327 to finish before merging.

PR #3651, opened 2025-12-03
Fixes for alternate module resolution.

schweitzpgi:
Teaches kernel launching to understand how to deal with qualified symbols and the symbol lookup rules that are entailed from that.

### DIFF
--- a/python/cudaq/kernel/utils.py
+++ b/python/cudaq/kernel/utils.py
@@ -207,7 +207,6 @@ def recover_value_of(name, resMod):
     result = recover_value_of_or_none(name, resMod)
     if is_recovered_value_ok(result):
         return result
-    print(resMod)
     raise RuntimeError("'" + name + "' is not available in this scope.")
 
 


python/cudaq/kernel/utils.py
```diff
@@ -149,10 +203,11 @@ def is_recovered_value_ok(result):
     return False
 
 
-def recover_value_of(name):
-    result = recover_value_of_or_none(name)
+def recover_value_of(name, resMod):
+    result = recover_value_of_or_none(name, resMod)
     if is_recovered_value_ok(result):
         return result
+    print(resMod)
```
1tnguyen: Just checking whether this is a debug print.

1tnguyen (approved):
LGTM 👍

python/cudaq/kernel/utils.py
schweitzpgi: It is. Let me take it back out.

PR #3653, opened 2025-12-04
[features/py-redesign] Pack state pointer as a pointer, no copy

1tnguyen:
### Description

I think this is another merge/rebase issue as I remembered making this change before :) 

Also, update the error message check as I edited it in https://github.com/NVIDIA/cuda-quantum/commit/237aa9bf9f11c675c0f4dd14c5e941f72b376c18


schweitzpgi (approved):
Good catch.

PR #3655, opened 2025-12-04
Fixes for Python/C++ interop.

schweitzpgi:
Much of this no longer makes much sense, but port the parts of the tests that do make sense to the new Python refactoring.

Specifically verboten: It does not make sense to try and synthesize Python kernel decorators or look them up by non-unique short names in C++ code. The python interpreter knows what kernel decorators it is referencing and must do the resolution of symbols to bindings.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -252,7 +252,7 @@ class PyKernelDecorator(object):
         if handler.skip_compilation():
             return
 
-        # Otherwise, precompile the kernel to portable MLIR.
+        # Otherwise, `precompile` the kernel to portable MLIR.
         if self.qkeModule:
             raise RuntimeError(self.name + " was already compiled")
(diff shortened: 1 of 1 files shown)


python/cudaq/kernel/kernel_decorator.py
bmhowe23, line 1: It's not clear to me which change in this review caused this, so I'm just making the comment in the first file. But this PR (if merged) seems like it would break the libraries. Here is a run where this PR was attempted to be run through the CUDA-QX CI: https://github.com/NVIDIA/cudaqx/actions/runs/19942540084/job/57184566220?pr=377#step:13:119

schweitzpgi: No idea. This PR isn't based off main, of course.

bmhowe23: Ah, I didn't catch that. I verified that `main` works right now, but perhaps the feature branch already had issues.

PR #3657, opened 2025-12-05
[features/py-redesign] Fix for C++ mock server tests

1tnguyen:
### Description

- Use name check to prevent accidentally marking entry-point `funcOp` as `private`. Add a debug assert to make sure we can detect that the `funcOp` is not present.

- Re-add the logic to look up the return type from `funcOp` if it is not present as an attribute. This is the case for C++.

This PR would fix some mock server tests, e.g., `extra-payload-provider-tests`, `quantinuum-tests`, `qci-tests`, etc.


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -498,7 +498,7 @@ class BaseRemoteRESTQPU : public QPU {
       // Mark all newly merged kernels private.
       for (auto &op : moduleOp)
         if (auto f = dyn_cast<mlir::func::FuncOp>(op))
-          if (f != func)
+          if (f.getName() != func.getName())
```
schweitzpgi: Why are these suddenly different `FuncOp`s? The Module ought to be stable here.

runtime/common/RunTheKernel.cpp
```diff
@@ -52,12 +52,24 @@ static LayoutInfoType extractLayout(const std::string &kernelName,
   if (!fnOp)
     throw std::runtime_error("Could not find " + kernelName +
                              " function in the module.");
-  if (!fnOp->hasAttr(cudaq::runtime::enableCudaqRun))
-    throw std::runtime_error("Run kernel " + kernelName +
-                             " incorrectly annotated.");
-  auto arrAttr = cast<ArrayAttr>(fnOp->getAttr(cudaq::runtime::enableCudaqRun));
   // Extract layout information from the function's return type.
-  Type returnTy = cast<TypeAttr>(arrAttr[0]).getValue();
+  Type returnTy = [&]() {
```
schweitzpgi, line 56: This is patching over a bug. The enableCudaqRun attribute is _required_ to hold the return type now.

runtime/common/BaseRemoteRESTQPU.h
schweitzpgi: I merged in the changes in this file.

runtime/common/RunTheKernel.cpp
schweitzpgi, line 56: This part, I am still not sure about. The attribute is only added in one place and that is where we generate the runnable launch stack. So I don't see how this attribute is being avoided.

Do you remember which test this failed on?

schweitzpgi: I'll open an issue for this. There is some rework needed on the C++ side now.

schweitzpgi: https://github.com/NVIDIA/cuda-quantum/issues/3661

PR #3658, opened 2025-12-05
[features/py-redesign] Fix segfault in remote-mqpu: move the code block to proper scope

1tnguyen:
### Description

Make sure `*rawArgs` will not cause a segfault.


runtime/common/BaseRestRemoteClient.h
```diff
@@ -158,6 +155,8 @@ class BaseRemoteRestRuntimeClient : public RemoteRuntimeClient {
     if (rawArgs || args) {
       mlir::PassManager pm(&mlirContext);
       if (rawArgs && !rawArgs->empty()) {
+        // Merge other modules (e.g., if there are device kernel calls).
+        cudaq::detail::mergeAllCallableClosures(module, name, *rawArgs);
```
schweitzpgi: This looks suspicious. `module` is getting cloned into `moduleOp` in the code that is now being run prior to merging the callable closures. I think we need to leave the merging of callable closures where it is and just add the guard:
```c++
if (rawArgs && !rawArgs->empty())
  // merge
```

1tnguyen: Good point! Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3658/commits/8e3fbd56ffd2294f1bb8e97650a09fda4cb8fb79

PR #3659, opened 2025-12-05
[features/py-redesign] Fix `H2MoleculeTester`:  C++ with embedded Python interpreter

1tnguyen:
### Description

Ported some missing code from https://github.com/NVIDIA/cuda-quantum/pull/3507

### DIFF
--- a/unittests/integration/qubit_allocation.cpp
+++ b/unittests/integration/qubit_allocation.cpp
@@ -14,7 +14,7 @@
 #ifndef CUDAQ_BACKEND_STIM
 
 std::vector<cudaq::complex> randomState(int numQubits) {
-  std::vector<cudaq::complex> stateVec(2 * numQubits);
+  std::vector<cudaq::complex> stateVec(1 << numQubits);
   std::generate(stateVec.begin(), stateVec.end(), []() -> cudaq::complex {
     thread_local std::default_random_engine
         generator; // thread_local so we don't have to do any locking


python/runtime/interop/PythonCppInterop.cpp
```diff
@@ -119,7 +119,7 @@ __attribute__((visibility("default"))) void
 registerDeviceKernel(const std::string &module, const std::string &name,
                      const std::string &mangled) {
   auto key = module + "." + name;
-  deviceKernelMLIRMap.insert({key, getMLIRCodeAndName(name, mangled)});
+  deviceKernelMLIRMap[key] = getMLIRCodeAndName(name, mangled);
```
schweitzpgi, line 122: This is a semantics change. `insert` will not overwrite an entry in the map, but the new code will. I don't know which is the correct behavior here.

python/utils/LinkedLibraryHolder.cpp
```diff
@@ -115,7 +115,7 @@ void parseRuntimeTarget(const std::filesystem::path &cudaqLibPath,
     }
   }
   target.platformName = foundPlatformName.value_or("default");
-  target.simulatorName = foundSimulatorName.value_or("qpp");
+  target.simulatorName = foundSimulatorName.value_or("");
```
schweitzpgi, line 118: Is this legit? It effects everything in Python, no?

1tnguyen: We have the code block in line 430 to handle it. This is to make it consistent with how `nvq++` handles the default simulator: for targets that don't have a simulator defined (e.g., local emulation of some hardware targets), we need to use the 'default' simulator, not hardwiring it to `qpp`.

python/runtime/interop/PythonCppInterop.cpp
1tnguyen: The behavior in `main` line was updated in https://github.com/NVIDIA/cuda-quantum/pull/3507 to allow for override. This is to facilitate Python target swapping: as the target changes, some device kernels may need to be updated as well, e.g., each target defines its own implementation of some `device_call` libraries.

PR #3662, opened 2025-12-06
Expose resource estimates as dictionary

### LINKED ISSUE #3645: convert estimate_resources() output to json object
```
import cudaq
import re
import json

@cudaq.kernel
def mykernel():
    q = cudaq.qubit()

    rx(.0, q)
    ry(.0, q)
    rz(.0, q)
    h(q)
    x(q)
    y(q)
    z(q)
    s(q)
    t(q)

counts = cudaq.estimate_resources(mykernel)

for gate, val in counts.items():
      print (gate, val)
```

gives this error:
AttributeError: 'cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.Resources' object has no attribute 'items'

It would be good to get this as json object

sacpis:
Adding an api to expose estimate resources as a dictionary

Fixes #3645.

### DIFF
--- a/runtime/common/Resources.h
+++ b/runtime/common/Resources.h
@@ -39,7 +39,7 @@ public:
     /// @brief The number of controls
     std::size_t nControls;
 
-    Instruction(const std::string &n) : name(n) {}
+    Instruction(const std::string &n) : name(n), nControls(0) {}
 
     /// @brief The constructor
     Instruction(const std::string &n, const size_t c) : name(n), nControls(c) {}


sacpis:
Seems like it is not optimal. Let me rethink a bit about exposing estimate_resources as a dict and a string.

sacpis:
With [86ffdbd](https://github.com/NVIDIA/cuda-quantum/pull/3662/commits/86ffdbd3ced344ae9aeca76414d7946c711da7f7) we are traversing over the instructions just once. I think for building string we can make use of the map we have built.

runtime/common/Resources.cpp
```diff
@@ -67,6 +66,8 @@ void Resources::appendInstruction(const std::string &name,
   }
 
   totalGates += count;
+
+  gateCountsByName[name] += count;
```
1tnguyen: This appears to be different from what `dump` is currently doing. Specifically, it will take into account the controlled version of a gate when printing out the result.

For example, it looks like this `gateCountsByName` map would treat `x` and `cx` as the same gate, doesn't it?

sacpis: Thanks @1tnguyen for catching this one. Have fixed the above and added the required test in [8de8045](https://github.com/NVIDIA/cuda-quantum/pull/3662/commits/8de80454482609ed0eb3e6bda6076712f4de54fa) and [f26dc29](https://github.com/NVIDIA/cuda-quantum/pull/3662/commits/f26dc29fedea7fc2748952b7b362d44e313f3e8e).

runtime/common/Resources.cpp
```diff
@@ -46,12 +47,10 @@ std::size_t Resources::count_controls(const std::string &name,
 }
 
 std::size_t Resources::count(const std::string &name) const {
-  std::size_t result = 0;
```
1tnguyen, line 49: The CI test failure indicates that this API's `name` argument is expected to be the base name for both controlled and non-controlled versions.

I wonder whether we should take a simpler approach, similar to `dump`, for `gateCounts`: converting `instructions` into a map with the same `key` format as the `dump`.

sacpis: Just to understand a bit better. Do we just need to count the base gates (which will be the sum of base + controlled) in the output for estimate_resources?

sacpis: Reverted the code here. 

But wondering if we should
1. Store both the base gate and its controlled version in the map, if both exists.
2. Store keep both base and its control version, if only the control version exists.

Right?

If only the base exists then yes, we will have it in the map.

sacpis: Checking other frameworks for resources estimation.

1tnguyen: > Do we just need to count the base gates (which will be the sum of base + controlled) in the output for estimate_resources?

As for the output of `dump` and the proposed `to_dict`, I think it should distinguish controlled vs. non-controlled gates, as they are significantly different in terms of resource estimation. 

> But wondering if we should
> 
> Store both the base gate and its controlled version in the map, if both exists.
> Store keep both base and its control version, if only the control version exists.

The current tracking map, where the key is a (gate name, number of control qubits) pair, is the most generic form. Hence, keeping this single map would make it more consistent. I suppose here we're adding a helper API to translate it to a Python `dict` in addition to the `dump` to string API, so that users can programmatically manipulate it (e.g., iterating over it).

I'd think the `dict` output can be either (1) close to the underlying data (gate name + number of controls as keys) or (2) the same as the existing `dump` output (number of control bits is converted to `c` prefixes).  For example, if (2) is preferred, we can then translate the underlying map to the string-based map (number of controls -> letter `c` prefixes) to return it to users.

PR #3668, opened 2025-12-09
[features/python.redesign] Port some of the old logic to handle simulator precision

1tnguyen:
### Description

This is part of our previous logic in handling user-provided data from Python:
https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/qis/state.cpp#L131-L159

Some of the tests would fail if the default simulator is the single-precision `nvidia` simulator rather than `qpp` as in CI. This would fix the issue.


1tnguyen:
> Isn't `cudaq::complex` supposed to be single precision float if the simulator uses float though? Is it a bug that it is double?

Yes, it should be if the code is compiled with `nvq++` for the target. For the Python module, it is always compiled with no target-specific config, hence double-precision. We need to rely on runtime target information to handle the data precision in this case.

PR #3676, opened 2025-12-10
Create cudaq.trace_kernel()

bmhowe23:
This PR implements `cudaq.trace_kernel()` in Python. It returns a `Trace` object that can be iterated over (in Python) to see the gates and gate parameters that were applied in a traced kernel execution.

See `python/tests/backends/test_trace_kernel.py` for example use cases (and tests).

Related: https://github.com/NVIDIA/cuda-quantum/discussions/3534

### DIFF
--- a/docs/sphinx/using/quick_start.rst
+++ b/docs/sphinx/using/quick_start.rst
@@ -16,10 +16,7 @@ found in our complete :doc:`Installation Guide <install/install>`.
 
 .. note::
 
-    If you are using CUDA 12.8 on Blackwell, you may need to install additional
-    dependencies to use the python wheels.
-
-    See :ref:`Blackwell CUDA Dependencies <blackwell-cuda-dependencies>` for more information.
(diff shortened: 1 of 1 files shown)


python/runtime/cudaq/algorithms/py_trace_kernel.cpp
```diff
@@ -0,0 +1,99 @@
...
+      .def("append_instruction", &Trace::appendInstruction)
+      .def("get_num_qudits", &Trace::getNumQudits)
+      .def(
+          "__iter__",
+          [](const Trace &t) { return py::make_iterator(t.begin(), t.end()); },
+          py::keep_alive<0, 1>());
+
+  mod.def(
+      "trace_kernel",
+      [&](py::object kernel, py::args args) {
```
schweitzpgi, line 69: With new launching, it is necessary to resolve arguments before dropping to C++. This will require re-work.

bmhowe23: Good to know. I believe this is pretty similar to py_resource_count.cpp, so whatever needed to be done to that file should hopefully be able to be done here, too.

PR #3679, opened 2025-12-11
Add detector support to CUDA-Q

bmhowe23:
## Overview

This PR introduces support for defining detectors in CUDA-Q kernels, a feature for Quantum Error Correction (QEC), and adds functionality to export kernels to the Stim circuit format.

## Details
**1. Detectors Support**
* New Operation: Introduced `cudaq::detector` (C++) and `cudaq.detector` (Python) to define detectors. Detectors compare measurement results to check for errors (parity checks).
* IR & Dialect: Added `quake.detector` operation to the Quake dialect and its lowering to QIR (`__quantum__qis__detector`).
* Compiler: Added an `EraseDetectors` pass to remove detector operations when targeting backends that do not support them (e.g., standard hardware execution, or if explicitly undesired).
* Runtime:
  * Updated ExecutionManager and simulators to support tracing and handling detector operations.
  * Implemented `cudaq.detectors()` utility to extract detector metadata (measurement indices) from a kernel.

**2. Stim Integration**
* `cudaq.to_stim(kernel)`: Added a new Python utility that converts a CUDA-Q kernel into a Stim circuit representation.
  * Supports gates, measurements, noise models, and the new detector operations.
* Stim Simulator: Updated StimCircuitSimulator to support a "to_stim" execution mode for string generation.

**3. Tracing Improvements**
* Measurement Tracing: Updated BasicExecutionManager to explicitly append measurement instructions (mz) to the kernel trace even in tracer mode.
  * Side Effect: The cudaq.draw output now explicitly shows measurement operations, updating expectations in test_kernel_features.py and draw_test.cpp.

## Tests
* New Tests:
  * python/tests/backends/test_to_stim.py: Validates Stim circuit generation, including noise models and detectors.
  * python/tests/kernel/test_detectors.py: Verifies cudaq.detectors() analysis.
  * unittests/integration/detector_tester.cpp: C++ integration tests for detector IR and runtime behavior.
* Updated Tests: Adjusted existing visualization tests to account for the improved visibility of measurements in kernel traces.


lib/Optimizer/Transforms/EraseDetectors.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // %3 = cc.cast %2 : (!cc.ptr<!cc.array<i64 x 2>>) -> !cc.ptr<!cc.array<i64 x ?>>
+    // %4 = cc.cast %2 : (!cc.ptr<!cc.array<i64 x 2>>) -> !cc.ptr<i64>
+    // cc.store %c-1_i64, %4 : !cc.ptr<i64>
+    // %5 = cc.compute_ptr %2[1] : (!cc.ptr<!cc.array<i64 x 2>>) -> !cc.ptr<i64>
+    // cc.store %c-2_i64, %5 : !cc.ptr<i64>
+    // %6 = cc.stdvec_init %3, %c2_i64 : (!cc.ptr<!cc.array<i64 x ?>>, i64) -> !cc.stdvec<i64>
+    // "quake.detector"(%6) : (!cc.stdvec<i64>) -> ()
+    // clang-format on
+
+    // TODO - decide whether or not to keep this path.
```
bmhowe23, line 47: This is almost certainly too fragile or dangerous to keep it this way. If there is a standard MLIR way to do this, we should do it here.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1510,7 +1517,19 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
   }
-
+  void detector(const std::int64_t *indices, std::size_t num_indices) override {
+    flushGateQueue();
+    flushAnySamplingTasks();
+    if (isInTracerMode()) {
+      std::vector<double> params(num_indices);
+      for (std::size_t i = 0; i < num_indices; i++) {
+        params[i] = indices[i];
+      }
```
1tnguyen: I wonder if this would work :) 
```suggestion
      std::vector<double> params(indices, indices + num_indices);
```

python/runtime/cudaq/algorithms/py_to_stim.cpp
```diff
@@ -0,0 +1,77 @@
...
+        args = simplifiedValidateInputArguments(args);
+        std::unique_ptr<OpaqueArguments> argData(
+            toOpaqueArgs(args, kernelMod, kernelName));
+
+        auto ctx = std::make_unique<ExecutionContext>("to_stim", 1);
+        ctx->kernelName = kernelName;
+        if (noise_model.has_value()) {
+          if (platform.is_remote())
+            throw std::runtime_error(
+                "Noise model is not supported on remote platforms.");
```
1tnguyen: Perhaps this check is not needed, given that we have checked that the current target is `stim`.

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -351,6 +351,26 @@ ParseResult quake::ApplyNoiseOp::parse(OpAsmParser &parser,
...
+    if (isa<cudaq::cc::StdvecType>(measure.getType()))
+      numVectorsFound++;
+    else if (isa<IntegerType>(measure.getType()))
+      numIntegersFound++;
+  }
+  if (numVectorsFound > 1)
+    return emitOpError("can only have one std::vector<int64_t> argument");
+  if (numIntegersFound > 0 && numVectorsFound > 0)
+    return emitOpError("can only have one std::vector<int64_t> argument or one "
+                       "integer argument");
```
1tnguyen: Should it be `one std::vector<int64_t> argument or integer arguments`?

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1634,6 +1634,35 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+        }
+
+        if (auto stdvecTy = dyn_cast<cudaq::cc::StdvecType>(aTy)) {
+          if (isa<IntegerType>(stdvecTy.getElementType())) {
+            measures.push_back(a);
+            continue;
+          }
+        }
+
+        reportClangError(x, mangler, "detector argument types not supported: ");
```
1tnguyen, line 1659: Just want to check that we'd have some info after ':' in the error message.

runtime/nvqir/CircuitSimulator.h
bmhowe23: Updated in 6fe4ce2213

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
bmhowe23: Updated in 6fe4ce2213

python/runtime/cudaq/algorithms/py_to_stim.cpp
bmhowe23: Updated in 6fe4ce2213

schweitzpgi (changes requested):
First, nice work pulling together this prototype from end to end. Very cool.

After talking to the architects, I think we should hold off merging this for now until we have a consensus on the user facing aspects, etc.

bmhowe23:
Superseded by https://github.com/NVIDIA/cuda-quantum/pull/4566. Closing.

PR #3680, opened 2025-12-11
[features/python.redesign][WIP] Fixing Python backend tests

1tnguyen:
### Description

- Exporting the `AsyncSampleResult` and `AsyncObserveResult` types.

- Porting the `AsyncSampleResult` Json serialization features.

- Guard the temp. file deletion: these fixtures are session-level; multiple test files may be using the same fake config file, so just skip cleanup if the file is already deleted.

### DIFF
--- a/python/cudaq/runtime/sample.py
+++ b/python/cudaq/runtime/sample.py
@@ -232,7 +232,7 @@ def sample_async(decorator,
       explicit_measurements (Optional[bool]): A flag to indicate whether or not
           to concatenate measurements in execution order for the returned
           sample result.
-      qpu_id (Optional[int]): The optional identification for which QPU
+      `qpu_id` (Optional[int]): The optional identification for which QPU
           on the platform to target. Defaults to zero. Key-word only.
 
     Returns:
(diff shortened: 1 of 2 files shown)


python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -325,28 +325,6 @@ def kernel(thetas: list[float]):
     assert_close(expectation, 0.707)
 
 
-def test_capture_array():
```
1tnguyen, line 328: @schweitzpgi Just confirming if this complex array-capture feature is supported for hardware emulation?

schweitzpgi: It should be. The symbol `arr` ought to be lambda lifted & resolved at the call site. That resolution ought to then synthesize the value bound to `arr` as a constant `!cc.array` in the kernel...

python/tests/backends/test_Quantinuum_builder.py
```diff
@@ -50,7 +50,10 @@ def startUpMockServer():
 
     # Kill the server, remove the file
     p.terminate()
-    os.remove(credsName)
+    try:
+        os.remove(credsName)
```
schweitzpgi, line 54: heh

PR #3681, opened 2025-12-11
Pulling in changes from PR #3537

bettinaheim:
- I haven't properly merged the mlir test changes in https://github.com/NVIDIA/cuda-quantum/pull/3537 with the changes on the feature branch
- I have removed test_assignment (added in https://github.com/NVIDIA/cuda-quantum/pull/3537) from this PR (currently segfaults, not yet sure why)
- I undid the flattening of the symbol table in the feature branch for now (to be done as a separate change)
- There are a couple of FIXMEs related to code clean up in the ast_bridge.py and utils.py. I'll go and clean them up (before or after pulling this onto the feature branch it shouldn't matter)
- One of the new tests in test_kernel_features.py is still commented out; it gives a comprehensive error on main but should work with the changes on the feature branch, but currently segfaults

Tests I ran locally:
I ran all tests in python/tests/, excluding backends, domain and remote; 
- I have failures for tests that return lists from kernels; these tests run on the main branch, but there is still a (comprehensive) error given that this is not supported on the feature branch. Indeed, if I comment that check out, I see that something is segfaulting. Are we maybe still missing some changes from main that yet need to be ported?
- The only other failure is for test_internal_library_kernels (unknown function call); I haven't check yet if this one is passing on the feature branch or not
- All other tests (that I ran) pass

### DIFF
--- a/python/cudaq/kernel/utils.py
+++ b/python/cudaq/kernel/utils.py
@@ -591,7 +591,9 @@ def mlirTypeFromPyType(argType, ctx, **kwargs):
         argTypeToCompareTo = (kwargs['argTypeToCompareTo']
                               if 'argTypeToCompareTo' in kwargs else None)
         if argTypeToCompareTo is None:
-            eleTypes = [mlirTypeFromPyType(type(ele), ctx) for ele in argInstance]
+            eleTypes = [
+                mlirTypeFromPyType(type(ele), ctx) for ele in argInstance
+            ]
             tupleTy = mlirTryCreateStructType(eleTypes, context=ctx)
         else:
             tupleTy = argTypeToCompareTo
--- a/python/tests/kernel/test_direct_call_return_kernel.py
+++ b/python/tests/kernel/test_direct_call_return_kernel.py
@@ -373,6 +373,7 @@ def test_return_tuple_int_float():
     assert result == (-13, 42.3)
 
     with pytest.raises(RuntimeError) as e:
+
         @cudaq.kernel
         def simple_tuple_int_float_assign(
                 n: int, t: tuple[int, float]) -> tuple[int, float]:
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -940,6 +940,7 @@ def test_list_boundaries():
     assert '10001' in counts
 
     with pytest.raises(RuntimeError) as e:
+
         @cudaq.kernel
         def kernel15():
             qubits = cudaq.qvector(5)
@@ -952,6 +953,7 @@ def test_list_boundaries():
     assert "offending source -> range(1, 4, 0)" in str(e.value)
 
     with pytest.raises(RuntimeError) as e:
+
         @cudaq.kernel
         def kernel16(v: int):
             qubits = cudaq.qvector(5)
@@ -2345,6 +2347,7 @@ def test_issue_9():
 def test_issue_1641():
 
     with pytest.raises(RuntimeError) as error:
+
         @cudaq.kernel
         def less_arguments():
             q = cudaq.qubit()
@@ -2355,6 +2358,7 @@ def test_issue_1641():
     assert '(offending source -> rx(3.14))' in repr(error)
 
     with pytest.raises(RuntimeError) as error:
+
         @cudaq.kernel
         def wrong_arguments():
             q = cudaq.qubit()
@@ -2365,6 +2369,7 @@ def test_issue_1641():
     assert "(offending source -> rx('random_argument', q))" in repr(error)
 
     with pytest.raises(RuntimeError) as error:
+
         @cudaq.kernel
         def wrong_type():
             q = cudaq.qubit()
@@ -2374,6 +2379,7 @@ def test_issue_1641():
     assert 'invalid argument type for target operand' in repr(error)
 
     with pytest.raises(RuntimeError) as error:
+
         @cudaq.kernel
         def invalid_ctrl():
             q = cudaq.qubit()
@@ -2618,6 +2624,7 @@ def test_mid_circuit_measurements():
 def test_error_on_non_callable_type():
 
     with pytest.raises(RuntimeError) as e:
+
         @cudaq.kernel
         def kernel(op: cudaq.pauli_word):
             q = cudaq.qvector(2)
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -198,6 +198,7 @@ class PyStack(object):
             return len(self._frame.entries)
         return 0
 
+
 def recover_kernel_decorator(name):
     from .kernel_decorator import isa_kernel_decorator
     for frameinfo in inspect.stack():
@@ -2220,10 +2221,9 @@ class PyASTBridge(ast.NodeVisitor):
             if elemTy == self.getIntegerType(1):
                 elemTy = self.getIntegerType(8)
             ptrTy = cc.PointerType.get(self.getIntegerType(8))
-            resBuf = cc.StdvecDataOp(cc.PointerType.get(elemTy),
-                                        value).result
+            resBuf = cc.StdvecDataOp(cc.PointerType.get(elemTy), value).result
             eleSize = cc.SizeOfOp(self.getIntegerType(),
-                                    TypeAttr.get(elemTy)).result
+                                  TypeAttr.get(elemTy)).result
             dynSize = cc.StdvecSizeOp(self.getIntegerType(), value).result
             stackCopy = cc.AllocaOp(cc.PointerType.get(
                 cc.ArrayType.get(elemTy)),
@@ -2234,8 +2234,7 @@ class PyASTBridge(ast.NodeVisitor):
                 cc.CastOp(ptrTy, resBuf).result,
                 arith.MulIOp(dynSize, eleSize).result
             ])
-            return cc.StdvecInitOp(value.type, stackCopy,
-                                    length=dynSize).result
+            return cc.StdvecInitOp(value.type, stackCopy, length=dynSize).result
 
         def convertArguments(expectedArgTypes, values):
             assert len(expectedArgTypes) == len(values)
@@ -2356,7 +2355,7 @@ class PyASTBridge(ast.NodeVisitor):
             processQuantumOperation(opName, controls, targets, [], params,
                                     **kwargs)
 
-        def processDecorator(name, path = None):
+        def processDecorator(name, path=None):
             if path:
                 name = f"{path}.{name}"
                 decorator = resolve_qualified_symbol(name)
@@ -2369,9 +2368,9 @@ class PyASTBridge(ast.NodeVisitor):
                                              nvqppPrefix + decorator.uniqName)
                 funcTy = FunctionType(
                     TypeAttr(entryPoint.attributes['function_type']).value)
-                callableTy = cc.CallableType.get(self.ctx,
-                                                 funcTy.inputs[:decorator.firstLiftedPos],
-                                                 funcTy.results)
+                callableTy = cc.CallableType.get(
+                    self.ctx, funcTy.inputs[:decorator.firstLiftedPos],
+                    funcTy.results)
 
                 # callee will be a new BlockArgument
                 callee = cudaq_runtime.appendKernelArgument(
@@ -2390,7 +2389,8 @@ class PyASTBridge(ast.NodeVisitor):
                 self.emitFatalError(
                     f"`{symName}` object is not callable, found symbol of type {kernel.type}",
                     node)
-            functionTy = FunctionType(cc.CallableType.getFunctionType(kernel.type))
+            functionTy = FunctionType(
+                cc.CallableType.getFunctionType(kernel.type))
             nrArgs = len(functionTy.inputs)
             values = self.__groupValues(node.args, [(nrArgs, nrArgs)])
             values = convertArguments([t for t in functionTy.inputs], values)
@@ -2419,7 +2419,8 @@ class PyASTBridge(ast.NodeVisitor):
                     for idx, element in enumerate(call.results):
                         result = cc.InsertValueOp(
                             structTy, result, element,
-                            DenseI64ArrayAttr.get([idx], context=self.ctx)).result
+                            DenseI64ArrayAttr.get([idx],
+                                                  context=self.ctx)).result
             # The logic for calls that return values must
             # match the logic in `visit_Return`; anything
             # copied to the heap during return must be copied
@@ -2459,7 +2460,7 @@ class PyASTBridge(ast.NodeVisitor):
                             devKey = f"{module_name}.{'.'.join(moduleNames[1:])}"
                         except AttributeError:
                             continue
- 
+
                 # Handle registered C++ kernels
                 if cudaq_runtime.isRegisteredDeviceModule(devKey):
                     maybeKernelName = cudaq_runtime.checkRegisteredCppDeviceKernel(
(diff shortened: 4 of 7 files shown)


python/tests/kernel/test_kernel_features.py
```diff
@@ -1005,23 +1124,6 @@ def canCaptureList():
                       atol=1e-3)
 
 
-def test_capture_disallow_change_variable():
```
bettinaheim, line 1008: added to the (not yet migrated) test_assignments.py

python/cudaq/kernel/utils.py
```diff
@@ -128,7 +128,7 @@ def resolve_qualified_symbol(y):
     decorator name.
     """
     parts = y.split('.')
-    for i in range(len(parts), 0, -1):
+    for i in range(len(parts)):
```
schweitzpgi, line 131: The last part in the path should not be a python module. (It is because of a misspelling in our tests.)

python/tests/kernel/test_direct_call_return_kernel.py
```diff
@@ -203,7 +203,7 @@ def simple_list_bool_no_args() -> list[bool]:
     @cudaq.kernel
     def simple_list_bool(n: int, t: list[bool]) -> list[bool]:
         qubits = cudaq.qvector(n)
-        return t
+        return t.copy()
```
schweitzpgi, line 206: What's the point of the copy() method changes here?

python/tests/kernel/test_direct_call_return_kernel.py
```diff
@@ -594,7 +585,7 @@ class MyClass:
     @cudaq.kernel
     def test_return_dataclass(n: int, t: MyClass) -> MyClass:
         qubits = cudaq.qvector(n)
-        return t
+        return t.copy(deep=True)
```
schweitzpgi, line 588: Is there a corresponding change in the bridge somewhere that necessitates these copies and deep copies?

By virtue of the fact this is a kernel running on a device, everything on the kernel side is logically a deep copy already. All device code is run in a different memory address space, logically speaking. So it's sort of weird to have device code making extra copies of its copies to return via copies to the Python interpreter.

python/tests/kernel/test_kernel_features.py
```diff
@@ -2217,8 +2386,8 @@ def invalid_ctrl():
             rx.ctrl(np.pi, q)
 
         print(invalid_ctrl)
-    assert ('controlled operation requested without any control argument(s)'
-            in repr(error))
+    assert 'missing value' in repr(error)
```
schweitzpgi, line 2389: Isn't this diagnostic much less specific? Is that the intention?

python/tests/visualization/test_draw.py
```diff
@@ -138,11 +138,11 @@ def hw_kernel():
...
-     ╭───╮                                       ╭───╮╭─────╮╭───╮╭───╮
-q0 : ┤ h ├──●─────────────────────●──────────────┤ x ├┤ tdg ├┤ x ├┤ t ├
-     ╰───╯  │                     │              ╰─┬─╯╰─────╯╰─┬─╯├───┤
-q1 : ───────┼───────────●─────────┼───────────●────●───────────●──┤ t ├
-     ╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮ ╭───╮      ╰───╯
+     ╭───╮                                                        ╭───╮
+q0 : ┤ h ├──────────────●─────────────────────●────●───────────●──┤ t ├
+     ╰───╯              │                     │  ╭─┴─╮╭─────╮╭─┴─╮├───┤
+q1 : ───────●───────────┼─────────●───────────┼──┤ x ├┤ tdg ├┤ x ├┤ t ├
+     ╭───╮╭─┴─╮╭─────╮╭─┴─╮╭───╮╭─┴─╮╭─────╮╭─┴─╮├───┤╰┬───┬╯╰───╯╰───╯
```
schweitzpgi, line 145: This seems odd. Why is this a different layout?

schweitzpgi:
I don't understand the addition of `.copy()` method calls all over the tests. Is it just annotation? Or is it some sort of fix?

To get to the nuts and bolts of how this must work in a Python kernel:

- Every local variable in a kernel must be strongly typed. It cannot have one type at one point and another type at another point in the kernel. If any such code is seen, it is a semantics error and a diagnostic must be raised.
```python
@cudaq.kernel
def bug():
  i = 4
  i = ['now', 'a', 'list']   # ERROR!
```

- There are no nested scopes. All local variables are global to the entire kernel.
- All local variables must therefore be promoted to the entry block.
- Each symbol name is a unique variable. In Python, this is achieved by garbage collecting unreferenced chafe.
- In our kernel world, we accomplish the same thing by allowing a symbol (now a variable) to bind to exactly 1 strongly typed value.
- All values are "garbage collected" when the kernel returns to the caller.
- We do not simulate or use reference counting as it is not useful for our limited inventory of types and the restrictions of the device.
- Assignments of value types should work exactly like Python.
```python
  i = 4
  j = i
```
means the symbol `i` and the symbol `j` both are bound to the value 4 by copy (value semantics).

In the more nebulous domain:

- Assignments of reference types should not work like Python to be consistent with C++. That is,
```python
  i = [4, 5, 6]
  i = [8, 9]    # ERROR!
```
This case is binding the symbol `i` to both a "vector" of length 3 and a "vector" of length 2, which we don't support in C++ and treat as distinct types.

It's worth having a conversation on this case... Do we have any code examples that do this?

schweitzpgi:
I'm going to cut the line and merge this so we can keep making progress. Just noting that there may be some things to revisit here as well.

python/cudaq/kernel/utils.py
1tnguyen: I'm fixing an empty `modName` issue in https://github.com/NVIDIA/cuda-quantum/pull/3682/files

python/tests/kernel/test_direct_call_return_kernel.py
bettinaheim: It makes it explicit what the behavior is under the hood. We ("have to") make a couple of copies where python does not and passes by reference instead. To ensure that we don't have unexpected behaviors for users (code not behaving as python should), we force the copy that we have to make to be explicitly visible in the source code.  More details in the description of https://github.com/NVIDIA/cuda-quantum/pull/3537

python/tests/kernel/test_direct_call_return_kernel.py
bettinaheim: See response above.

python/tests/kernel/test_kernel_features.py
bettinaheim: It is a bit less specific because all function arguments validations are now handled by the same logic. So while we can know the data types, we cannot know it's a control argument.

python/tests/visualization/test_draw.py
bettinaheim: This is merely because the order or control qubits changed (order or controls is irrelevant); the new implementation that again handles all controls by the same logic happens to preserve the order of controls as written in the source code, whereas the old implementation reversed it.

bettinaheim:
> I don't understand the addition of `.copy()` method calls all over the tests. Is it just annotation? Or is it some sort of fix? 

It's kind of an annotation for the users sake indicating what is going on under the hood.

> 
> To get to the nuts and bolts of how this must work in a Python kernel:
> 
> * Every local variable in a kernel must be strongly typed. It cannot have one type at one point and another type at another point in the kernel. If any such code is seen, it is a semantics error and a diagnostic must be raised.
> 
> ```python
> @cudaq.kernel
> def bug():
>   i = 4
>   i = ['now', 'a', 'list']   # ERROR!
> ```
> 
> * There are no nested scopes. All local variables are global to the entire kernel.
> * All local variables must therefore be promoted to the entry block.

I need to come back to this with the change to flatten the symbol table. (Nothing "bad" happens as it is - the IR we generate is always valid and matches Python behavior - but we are not respecting Python scoping as written). We can be a bit more permissive than what you outline also when respecting Python scoping - we merely have to be strict when a variable is redefined in an inner scope (wrt MLIR scoping). I'll give it some more thought.

> * Each symbol name is a unique variable. In Python, this is achieved by garbage collecting unreferenced chafe.
> * In our kernel world, we accomplish the same thing by allowing a symbol (now a variable) to bind to exactly 1 strongly typed value.
> * All values are "garbage collected" when the kernel returns to the caller.
> * We do not simulate or use reference counting as it is not useful for our limited inventory of types and the restrictions of the device.
> * Assignments of value types should work exactly like Python.
> 
> ```python
>   i = 4
>   j = i
> ```
> 
> means the symbol `i` and the symbol `j` both are bound to the value 4 by copy (value semantics).

Yes.

> 
> In the more nebulous domain:
> 
> * Assignments of reference types should not work like Python to be consistent with C++. That is,

The goal that I am aiming for is that either it behaves like Python, or we give an explicit and comprehensive error that this is not allowed. I do want to fully get rid of cases that compile and run but don't behave as python would. 

> 
> ```python
>   i = [4, 5, 6]
>   i = [8, 9]    # ERROR!
> ```
> 
> This case is binding the symbol `i` to both a "vector" of length 3 and a "vector" of length 2, which we don't support in C++ and treat as distinct types.
> 
> It's worth having a conversation on this case... Do we have any code examples that do this?

The new assignment tests that are still missing contain a good number of tests along similar lines, though I may not have added one specifically related to lengths of lists

python/tests/kernel/test_direct_call_return_kernel.py
schweitzpgi: This is a user facing CUDA-Q design change that should have been gone through the architectural review process, I believe. @boschmitt @amccaskey 

We should be converting _local_ python bindings to independent variables. Making copies will just exaggerate the use of stack space and complicate compiler analyses needlessly.

In Python a local symbol is maintained in a dictionary as a reference to a value. If the value is a reference object, then this amounts to a double indirection. One cannot bind the same symbol to more than one value in the local scope. One can bind more than one symbol to the same reference object.

```python
   a = [1, 2]
   b = a
   a[0] = 4
   assert b[0] == 4  # True!
```

This functionality can be mimicked as classic variables without resorting to copies. Still, I'm not clear where the original Python front end proposal landed on this issue.

How to not use copies:

For value types, symbol binding and variables have a tight correspondence.
```python
  a = 4
  b = a
```
can be lowered as
```mlir
  %0 = alloca integer
  %1 = alloca integer
  store 4, %0
  // symbol 'a' is %0
  %2 = load %0
  store %2, %1
  // symbol 'b' is %1
```
and the observable semantics are the same.

For reference types
```python
  a = [0, 1]
  b = a
```
this can be lowered as follows
```mlir
  %0 - alloca list<2 x integer>
  // symbol a -> %0
  store 0, %0[0]
  store 1, %0[1]
  // symbol b -> %0
```
I believe this was the way CUDA-Q was operating.

This is a problem when it comes to runtime control flow issues.
```python
  a = [0, 1]
  if b:
    c = a
  else:
    c = [3, 4]
  # c is not statically known to be [0,1] or [3, 4] here
```

We can still solve this without making copies, and having natural looking python. This can be achieved by noting what was stated earlier: for a reference binding, the interpreter binds the symbol to a _reference_ meaning we have a double indirection.
```mlir
  %0 - alloca list<2 x integer>
  store 0, %0[0]
  store 1, %0[1]
  // %0 is the list object [0, 1]
  %1 = alloca ptr<list<...>>
  store %0, %1
  // symbol 'a' is %1
  %2 = alloca ptr<list<...>>
  // symbol 'b' is %2 (and undefined)
  %3 - alloca list<2 x integer>
  store 3, %3[0]
  store 4, %3[1]
  // %3 is the list object [3, 4]
  structured.if (%88) {
    store %0, %2
  } else {
    store %3, %2
  }
  // %2 is correctly either a reference to [0, 1] or [3, 4] here
```

The point being that all this can be done correctly by
- knowing the difference between value and reference types
- using a reference indirection correctly
- preserving the correspondence between symbol bindings in python to variables in the IR
- not changing or exposing properties of the implementation to the user-facing interface

schweitzpgi:
> > I don't understand the addition of `.copy()` method calls all over the tests. Is it just annotation? Or is it some sort of fix?
> 
> It's kind of an annotation for the users sake indicating what is going on under the hood.
> 
> > To get to the nuts and bolts of how this must work in a Python kernel:
> > 
> > * Every local variable in a kernel must be strongly typed. It cannot have one type at one point and another type at another point in the kernel. If any such code is seen, it is a semantics error and a diagnostic must be raised.
> > 
> > ```python
> > @cudaq.kernel
> > def bug():
> >   i = 4
> >   i = ['now', 'a', 'list']   # ERROR!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * There are no nested scopes. All local variables are global to the entire kernel.
> > * All local variables must therefore be promoted to the entry block.
> 
> I need to come back to this with the change to flatten the symbol table. (Nothing "bad" happens as it is - the IR we generate is always valid and matches Python behavior - but we are not respecting Python scoping as written). We can be a bit more permissive than what you outline also when respecting Python scoping - we merely have to be strict when a variable is redefined in an inner scope (wrt MLIR scoping). I'll give it some more thought.
> 
> > * Each symbol name is a unique variable. In Python, this is achieved by garbage collecting unreferenced chafe.
> > * In our kernel world, we accomplish the same thing by allowing a symbol (now a variable) to bind to exactly 1 strongly typed value.
> > * All values are "garbage collected" when the kernel returns to the caller.
> > * We do not simulate or use reference counting as it is not useful for our limited inventory of types and the restrictions of the device.
> > * Assignments of value types should work exactly like Python.
> > 
> > ```python
> >   i = 4
> >   j = i
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > means the symbol `i` and the symbol `j` both are bound to the value 4 by copy (value semantics).
> 
> Yes.
> 
> > In the more nebulous domain:
> > 
> > * Assignments of reference types should not work like Python to be consistent with C++. That is,
> 
> The goal that I am aiming for is that either it behaves like Python, or we give an explicit and comprehensive error that this is not allowed. I do want to fully get rid of cases that compile and run but don't behave as python would.
> 
> > ```python
> >   i = [4, 5, 6]
> >   i = [8, 9]    # ERROR!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This case is binding the symbol `i` to both a "vector" of length 3 and a "vector" of length 2, which we don't support in C++ and treat as distinct types.
> > It's worth having a conversation on this case... Do we have any code examples that do this?
> 
> The new assignment tests that are still missing contain a good number of tests along similar lines, though I may not have added one specifically related to lengths of lists

Thanks. It's good that these issues are getting thought about for sure!

See also https://github.com/NVIDIA/cuda-quantum/pull/3681#discussion_r2614857519. I think more conversations on what this ought to look like, particularly as presented to the user, would be a good thing.

python/tests/kernel/test_direct_call_return_kernel.py
schweitzpgi: I'll explicitly call out here the deal with "scope" wrt Python as well.

We are not concerned with any scope inside the kernel other than symbols at local scope. Symbols referenced at any other scope are lambda lifted and passed by value (deep copy).

By the nature of Python's symbol binding process, all values are "free" on the heap (and reference counted, etc.) There is no lexical scope and no defined lifetime. At some point, the interpreter will just garbage collect unreferenced objects on the heap.

In our device execution domain, we don't have a heap and certainly do not have garbage collection. We instead have variables and place values into those variables (stack slots).

By virtue of this scopeless, vague lifetime execution model in our host language, Python, we can simply promote local symbols to variables that dominate everything in the kernel (place them in the entry block) and we can also promote objects of reference type in the same way (construct them as early as possible, which isn't the same as always the entry block, but it may often be the case).

If we follow the same rules as the C++ limitations, the storage for the reference objects must be known at compile-time, so it is possible to allocate that storage in the entry block, even if the initialization of its element values is done along various control-flow paths.

PR #3683, opened 2025-12-12
Enabling most of the assignment tests

bettinaheim:
- capture test is still missing
- about 3 other tests currently fail; these cases are not supported in main currently, but produce a comprehensive error. We either need to support them, or add back the comprehensive error

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -3593,21 +3593,23 @@ class PyASTBridge(ast.NodeVisitor):
                 node)
 
         def process_void_list():
-            # NOTE: This does not actually create a valid value, and will fail is something
-            # tries to use the value that this was supposed to create later on.
-            # Keeping this to keep existing functionality, but this is a bit questionable.
-            # Aside from no list being produced, this should work regardless of what we
-            # iterate over or what expression we evaluate.
+            # NOTE: This does not actually create a valid value, and will fail
+            # if something tries to use the value that this was supposed to
+            # create later on. Keeping this to keep existing functionality, but
+            # this is a bit questionable. Aside from no list being produced,
+            # this should work regardless of what we iterate over or what
+            # expression we evaluate.
             self.emitWarning(
-                "produced elements in list comprehension contain None - expression will be evaluated but no list is generated",
-                node)
+                "produced elements in list comprehension contain None - "
+                "expression will be evaluated but no list is generated", node)
             forNode = ast.For()
             forNode.iter = node.generators[0].iter
             forNode.target = node.generators[0].target
             forNode.body = [node.elt]
             forNode.orelse = []
             forNode.lineno = node.lineno
-            # this loop could be marked as invariant if we didn't use `visit_For`
+            # This loop could be marked as invariant if we didn't use
+            # `visit_For`, but that would be premature optimization.
             self.visit_For(forNode)
 
         target_types = {}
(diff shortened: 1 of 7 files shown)


lib/Optimizer/Transforms/EraseVectorCopyCtor.cpp
```diff
@@ -59,8 +59,26 @@ class EraseVectorCopyCtorPattern : public OpRewritePattern<func::CallOp> {
     auto newStackSlot = casted.getValue().getDefiningOp<cudaq::cc::AllocaOp>();
     if (!newStackSlot)
       return failure();
-    rewriter.replaceOpWithNewOp<cudaq::cc::CastOp>(
-        newStackSlot, newStackSlot.getType(), analysis.copyFrom.getOperand(1));
+    auto source =
```
schweitzpgi, line 62: What's the IR pattern you are trying to match here? Do you have an example?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2221,10 +2221,12 @@ def copy_list_to_stack(value):
             if elemTy == self.getIntegerType(1):
                 elemTy = self.getIntegerType(8)
             ptrTy = cc.PointerType.get(self.getIntegerType(8))
-            resBuf = cc.StdvecDataOp(cc.PointerType.get(elemTy), value).result
+            ptrArrTy = cc.PointerType.get(cc.ArrayType.get(elemTy))
```
schweitzpgi, line 2224: Can `elemTy` already be an `ArrayType`?

schweitzpgi (commented):
Nice set of tests!

python/cudaq/kernel/ast_bridge.py
bettinaheim: No, it elemTy would be std::vectype for nested lists.

lib/Optimizer/Transforms/EraseVectorCopyCtor.cpp
bettinaheim: That's the segfault case that I shared in chat. The orginal IR prior to this pass looks something like this:
```
   %0 = cc.address_of @__nvqpp__mlirgen__get_list..0x778b00945190.rodata_1 : !cc.ptr<!cc.array<i64 x 1>>
    %1 = cc.cast %0 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i8>
    %2 = call @malloc(%c8_i64) : (i64) -> !cc.ptr<i8>
    call @llvm.memcpy.p0i8.p0i8.i64(%2, %1, %c8_i64, %false) : (!cc.ptr<i8>, !cc.ptr<i8>, i64, i1) -> ()
    %3 = cc.alloca !cc.array<i64 x 1>
    %4 = cc.cast %3 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i8>
    call @llvm.memcpy.p0i8.p0i8.i64(%4, %2, %c8_i64, %false) : (!cc.ptr<i8>, !cc.ptr<i8>, i64, i1) -> ()
    call @free(%2) : (!cc.ptr<i8>) -> ()
```
This causes problems with later use such as
```
    %5 = cc.cast %3 : (!cc.ptr<!cc.array<i64 x 1>>) -> !cc.ptr<i64>
    cc.store %c6_i64, %5 : !cc.ptr<i64>
```

Before this change, this would have been replaced all to be 
```
  store i64 6, i64* getelementptr inbounds ([1 x i64], [1 x i64]* @__nvqpp__mlirgen__get_list..0x7df1b096e5a0.rodata_1, i32 0, i32 0), align 8
```

What the change is intended to do is to check if the value we store in the malloc is a global constant (currently done by checking if it is a `cc::address_of`), and if it is, it removes all malloc, memcpy and free, but keeps the alloca and loads and stores the value in `address_of` on the stack.

PR #3685, opened 2025-12-16
Introducing any_qpu type to hold both qpus based on the QPU base class and trait based qpus.

Renaud-K:
In preparation for the runtime refactor and the transition of QPU based classes (qpu.h) to trait based classes. I propose the use of the any_qpu type in the quantum platform, which call hold both.  

1. Introducing a new directory where the new files are added. The idea is to make it easier to remove the old files once the migration is done.
2. The quantum_platforms holds the any_qpu type, instead of the QPU type. It turns out, that this was the only place where the change has to happen.
3. A non LLVM based type-erased registry is introduced to hold both types. The LLVM based registry only supports type which have a common base, however trait based class have no common base.
    - The registry is made to work with more than QPUs, so it is split in 2 files: type_erased_registry.h and any_qpu.h
    - The interface to the registry mimics that of the current LLVM based registry.
4. In the case of overloaded members, all overloads need to be present in a class for the look-up mechanism to work.
5.  For this registry to work across executables, shared libs and dynamically loaded libs, the trick it to explicitly instantiate only once the `get` function of the registry, in this aggregate, to insure that all binaries call a unique one. This is also the reason why there is LLVM_INSTANTIATE_REGISTRY.
6. There is an emerging task for the refactor to allow the use of the logger in the header files in a way that the fmtlib package does not bleed into user code. For now, I have split it into Logger.h and logger_lite.h, so basic logging can be done. This was a major pain point to debug the code.


python/runtime/utils/PyFermioniqRESTQPU.cpp
```diff
@@ -10,6 +10,7 @@
 #include "cudaq/Optimizer/InitAllDialects.h"
 #include "cudaq/platform/fermioniq/FermioniqBaseQPU.h"
 
+#include "version2/cudaq/qpus/any_qpu.h"
```
sacpis: A question. Can we rename any_qpu to qpu?

Renaud-K: By `QPU` type, we usually mean the type in `qpu.h`. (The file name is in lower case)
`any_qpu` encapsulates the `QPU` type and any new qpu type we introduce.
There would be a confusion if we called it qpu as well, even in lower case.

runtime/version2/cudaq/qpus/any_qpu.h
```diff
@@ -0,0 +1,262 @@
...
+      launchVQE;
+  std::function<KernelThunkResultType(
+      void *, const std::string &, KernelThunkType, void *, std::uint64_t,
+      std::uint64_t, const std::vector<void *> &)>
+      launchKernelThunk;
+  std::function<void(void *, const std::string &, const std::vector<void *> &)>
+      launchKernelArgs;
+  std::function<void(void *, const std::string &,
+                     SerializedCodeExecutionContext &)>
+      launchSerializedCodeExecution;
```
sacpis: We can remove this function as we are not supporting NVQC anymore.

runtime/cudaq/version2/cudaq/qpus/any_qpu.h
```diff
@@ -0,0 +1,262 @@
...
+    resetExecutionContext = [](void *i) {
+      static_cast<T *>(i)->resetExecutionContext();
+    };
+    setTargetBackend = [](void *i, const std::string &backend) {
+      static_cast<T *>(i)->setTargetBackend(backend);
+    };
+    launchVQE = [](void *i, const std::string &name, const void *kernelArgs,
+                   gradient *grad, const spin_op &H, optimizer &opt,
+                   const int n_params, const std::size_t shots) {
+      static_cast<T *>(i)->launchVQE(name, kernelArgs, grad, H, opt, n_params,
```
sacpis, line 121: I think VQE is going to be moved to CUDA-QX. @bmhowe23 would you please confirm?

runtime/version2/cudaq/qpus/any_qpu.h
```diff
@@ -0,0 +1,262 @@
...
+                                               resultSize, rawArgs);
+    };
+    launchKernelArgs = [](void *i, const std::string &name,
+                          const std::vector<void *> &rawArgs) {
+      static_cast<T *>(i)->launchKernel(name, rawArgs);
+    };
+    launchSerializedCodeExecution =
+        [](void *i, const std::string &name,
+           SerializedCodeExecutionContext &context) {
+          static_cast<T *>(i)->launchSerializedCodeExecution(name, context);
```
sacpis: We can remove launchSerializedCodeExecution (related to NVQC).

runtime/version2/cudaq/qpus/any_qpu.h
```diff
@@ -0,0 +1,262 @@
...
+  }
+  void launchKernel(const std::string &name,
+                    const std::vector<void *> &rawArgs) {
+    table().launchKernelArgs(m_instance.get(), name, rawArgs);
+  }
+  void launchSerializedCodeExecution(
+      const std::string &name,
+      SerializedCodeExecutionContext &serializeCodeExecutionObject) {
+    table().launchSerializedCodeExecution(m_instance.get(), name,
+                                          serializeCodeExecutionObject);
```
sacpis: Let's remove this one please.

runtime/version2/cudaq/qpus/any_qpu.h
```diff
@@ -0,0 +1,262 @@
...
+  }
+  void onRandomSeedSet(std::size_t seed) {
+    table().onRandomSeedSet(m_instance.get(), seed);
+  }
+
+private:
+  const details::qpu_dispatch_table &table() const {
+    assert(m_vtable && "any_qpu accessed without a dispatch table.");
```
sacpis: Instead of table() we can use m_vtable, right? I am wondering if we really need to check for m_vtable validity in every function.

runtime/version2/cudaq/utils/type_erased_registry.h
```diff
@@ -0,0 +1,51 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include <functional>
+#include <memory>
+#include <unordered_map>
+
+#pragma once
+
+template <typename any_type, typename dispatch_table> class TypeErasedRegistry {
```
sacpis: nit: class should be on a new line.

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: It is still used [here](https://github.com/Renaud-K/cuda-quantum/blob/1566ddf034075c29ee263fc474e944aef73e2ca9/python/runtime/cudaq/algorithms/py_optimizer.cpp#L210)

I am not sure how to remove it there.

runtime/version2/cudaq/utils/type_erased_registry.h
Renaud-K: This is WIP. I have not run clang-format yet.

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: Can you remove it from [there](https://github.com/Renaud-K/cuda-quantum/blob/1566ddf034075c29ee263fc474e944aef73e2ca9/python/runtime/cudaq/algorithms/py_optimizer.cpp#L210)?

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: It is inlined and the assert goes away in release builds. But yes, it can be removed.

lmondada (commented):
I like this proposal a lot!

The only drawback is that the any_qpu definition is very verbose: each method appears three times:
- as attribute in the `qpu_dispatch_table`,
- in the `build` method
- as a method on `any_qpu`

My C++ hackery skills are at a limit here. I don't see how this redundancy could be resolved with a macro. Any chance tablegen could help here? I am not familiar enough with what it can/cannot do.

Anyways, I think this is a good approach and I am in favour of merging this in!

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: It has been removed.

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: Removed

runtime/version2/cudaq/qpus/any_qpu.h
Renaud-K: Done.

Renaud-K:
Thank you for the support. I have not had a chance to give too much thoughts about the drawbacks mentioned [here](https://github.com/NVIDIA/cuda-quantum/pull/3685#pullrequestreview-3589078287). I am also not able to reply there. My focus on the past two days has been on making it work across executables, shared libs and dynamically loaded libs. The trick it to explicitly instantiate only once the `get` function of the registry, this aggregate to insure that all binaries call the same. This is the reason why there is `LLVM_INSTANTIATE_REGISTRY`.

Reflections would probably help best with this issue. Macros could help a little too I think. I am not sure I want to get too sophisticated this early in the process with tablegen.

Reducing the size of the interface will help too which should be where we are headed.

Renaud-K:
> /ok to test 32911eda7ae3dfd0f478e1863178711902168405

Renaud-K:
> /ok to test cc961141a12f6d4ae30c001fa32386ba7895f883

Renaud-K:
> /ok to test f2785df856088bbcc7beadbf9038441493768476

Renaud-K:
> /ok to test d007f5f65e4293af188768bb260d8aa3040da128

Renaud-K:
> /ok to test bb314c62a7576fa1d60a641d8e2950b2b2947fc3

Renaud-K:
> /ok to test 7f51047db9998e26515fd3987702db7ffa10e7d8

Renaud-K:
> /ok to test 4136d8fbfd3dbe268b98347243ad30f61807fd48

runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp
```diff
@@ -72,7 +80,9 @@ class DefaultQuantumPlatform : public cudaq::quantum_platform {
 public:
   DefaultQuantumPlatform() {
     // Populate the information and add the QPUs
-    platformQPUs.emplace_back(std::make_unique<DefaultQPU>());
+    CUDAQ_REGISTER_QPU_TYPE(DefaultQPU, default);
```
1tnguyen, line 81: Q: do we need `CUDAQ_REGISTER_QPU_TYPE` here as we have that in line 157?

runtime/cudaq/version2/cudaq/utils/type_erased_registry.h
```diff
@@ -0,0 +1,114 @@
+/*******************************************************************************
+ * Copyright (c) 2025 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include "common/logger_lite.h"
+#include "iostream"
```
1tnguyen: nit: should it be `<iostream>`?

Renaud-K: Fixed.

runtime/cudaq/platform/default/DefaultQuantumPlatform.cpp
Renaud-K: In the scenarios when the platform is created before main with for instance:
```
__attribute__((constructor)) void setTargetBackend() {
```
we do not control the order in which the constructors are called. So the registration of the QPUs may not have happened yet and the call to instantiating the default QPU can fail. 

The previous call did not have to worry about this because it could do a straight instantiation of it.
```
platformQPUs.emplace_back(std::make_unique<DefaultQPU>())
```
We need to rely on the registry for this case as well now.

I don't think there are scenarios where the DefaultQPU is needed before the platform, so line 157 could go. It is there just in case.

The MultiQPUQuantumPlatform does this strange check in its constructor:
```
  MultiQPUQuantumPlatform() {
    if (cudaq::registry::QPURegistry::get().is_registered("GPUEmulatedQPU")) {
       ...
    } else {
      CUDAQ_WARN(
          "MultiQPUQuantumPlatform: GPUEmulatedQPU type not registered.");
    }
```
And I am wondering if we should not have an explicit registration there as well. I am not sure how it has worked so far.

Renaud-K:
> /ok to test df4a77c701bfe7eb2be9b48e4c4cf7c35c57c944

Renaud-K:
> /ok to test 47771b1

Renaud-K:
> /ok to test 8ca28c591340cdcfc88cca496798a2c50032d0d3

Renaud-K:
> /ok to test 21ed0301750f76b7f32455f0071f5374d57bbcc0

Renaud-K:
> /ok to test a8f4f3eb7b5d6d3546fa170bc0099c9a9c13f0a9

PR #3688, opened 2025-12-16
Match python scoping rules when possible

bettinaheim:
The implementation of the PyScopedSymbolTable ensures that we are matching Python scoping rules for all variables that are represented as alloca at the beginning of the function. For everything else, where we cannot support python scoping rules, we give a comprehensive error.

Also included: updated the ast_list_comprehension tests (mostly just updating function signatures to account for captures). 

TODO:
- write more tests for scoping
- support inner functions (proper inner python scopes...) properly

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -59,7 +59,7 @@ ALLOWED_TYPES_IN_A_DATACLASS = [int, float, bool, cudaq_runtime.qview]
 
 class PyScopedSymbolTable(object):
 
-    # FIXME: make this similar to pystack to account for local functions
+    # FIXME: make this similar to `pystack` to account for local functions
     def __init__(self):
         self.symbolTable = {}
         self.scopes = deque()
@@ -1461,11 +1461,11 @@ class PyASTBridge(ast.NodeVisitor):
 
     def __validate_container_entry(self, mlirVal, pyVal):
         '''
-        Helper function that should be invoked for any elements that are stored in
-        tuple, dataclass, or list. Note that the `pyVal` argument is only used to
-        determine the root of `mlirVal` and as such could be either the Python
-        AST node matching the container item (`mlirVal`) or the AST node for the 
-        container itself.
+        Helper function that should be invoked for any elements that are stored
+        in tuple, `dataclass`, or list. Note that the `pyVal` argument is only
+        used to determine the root of `mlirVal` and as such could be either the
+        Python AST node matching the container item (`mlirVal`) or the AST node
+        for the container itself.
         '''
 
         rootVal = self.__get_root_value(pyVal)
@@ -1480,8 +1480,8 @@ class PyASTBridge(ast.NodeVisitor):
                     "cannot use `cudaq.State` as element in lists, tuples, "
                     "or dataclasses", self.currentNode)
             self.emitFatalError(
-                "lists, tuples, and dataclasses must not contain modifiable values",
-                self.currentNode)
+                "lists, tuples, and dataclasses must not "
+                "contain modifiable values", self.currentNode)
 
         if cc.StructType.isinstance(mlirVal.type):
             structName = cc.StructType.getName(mlirVal.type)
@@ -2416,7 +2416,7 @@ class PyASTBridge(ast.NodeVisitor):
                     self.ctx, funcTy.inputs[:decorator.firstLiftedPos],
                     funcTy.results)
 
-                # callee will be a new BlockArgument
+                # `callee` will be a new `BlockArgument`
                 callee = cudaq_runtime.appendKernelArgument(
                     self.kernelFuncOp, callableTy)
                 self.argTypes.append(callableTy)
@@ -2424,7 +2424,7 @@ class PyASTBridge(ast.NodeVisitor):
 
             return name if decorator else None
 
-        # FIXME: unify with processFunctionCall?
+        # FIXME: unify with `processFunctionCall`?
         def processDecoratorCall(symName):
             assert symName in self.symbolTable
             self.visit(ast.Name(symName))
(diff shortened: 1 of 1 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3878,9 +3910,6 @@ def visit_List(self, node):
         # Prevent the creation of empty lists, since we don't support inferring
         # their types. To do so, we would need to look forward to the first use
         # and determine the type based on that.
-        # FIXME: Lower a list to a `!cc.stdvec<T>` or a
```
bettinaheim, line 3881: @schweitzpgi This should **not** be done here. There is a distinction between creating a value and creating a variable. A value creates the stdvec object. Only when it is assigned to a variable do/should we create a stack slot **for the variable** and store the value in it. This is handled by visit_assign (I added a comment there regarding creating stack slots for vectors)

schweitzpgi: I think there is a misunderstanding. I am fully aware of the distinctions between values, variables, a list object, and a symbol binding.

schweitzpgi:
The title of this PR suggests some fundamental misunderstanding to me as Python functions do not have block scopes. Lowering Python to variable semantics for both value-type and reference-type objects, with or without bindings can be done in a straightforward way, which I think we're not quite arriving at yet.

In short, there are absolutely _no scopes_. We should not be simulating scopes because it is deeply incorrect. We should have no notion of scope in the Python bridge.

```python
def python(a, b, c):
   if a:
     local1 = b
     local3 = [a]
   else:
     local2 = c
     local3 = [a, a]
   # [*]
   ...
```

The symbols `local1`, `local2`, and `local3` have the exact same scope. That scope is "Local" scope in Python parlance. The interpreter knows about all these symbols and when executing the code below `# [*]` it will either lookup the value that the symbol is bound to or throw an exception that the symbol has not been bound.

`local3` is a symbol bound to a reference-type. A reference-type is _an object_ of non-specific scope. `local3` binds to the reference to the object. The construction of the list container will not have side-effects by the definition of CUDA-Q, so these containers can be hoisted for example.

I hope to make some slides that will clarify this and present to the team in January.

bettinaheim:
> The title of this PR suggests some fundamental misunderstanding to me as Python functions do not have block scopes. Lowering Python to variable semantics for both value-type and reference-type objects, with or without bindings can be done in a straightforward way, which I think we're not quite arriving at yet.
> 
> In short, there are absolutely _no scopes_. We should not be simulating scopes because it is deeply incorrect. We should have no notion of scope in the Python bridge.
> 
> ```python
> def python(a, b, c):
>    if a:
>      local1 = b
>      local3 = [a]
>    else:
>      local2 = c
>      local3 = [a, a]
>    # [*]
>    ...
> ```
> 
> The symbols `local1`, `local2`, and `local3` have the exact same scope. That scope is "Local" scope in Python parlance. The interpreter knows about all these symbols and when executing the code below `# [*]` it will either lookup the value that the symbol is bound to or throw an exception that the symbol has not been bound.
> 
> `local3` is a symbol bound to a reference-type. A reference-type is _an object_ of non-specific scope. `local3` binds to the reference to the object. The construction of the list container will not have side-effects by the definition of CUDA-Q, so these containers can be hoisted for example.
> 
> I hope to make some slides that will clarify this and present to the team in January.

Well, there are scopes, though we only need to worry about the one per function (the fixme in the table). However, the bridge connects python with MLIR, and while python does not have scopes, MLIR does, and until we eliminate any kind of local definition of a variable *in MLIR* we need to keep track of where these are. I my though is to get the compiler restructure fully fleshed out and confirm that the compiler can still deal with all needed code optimizations for qpus, before we add more alloca. Additionally, we do need to consider a bit more carefully what we want to do with quantum data types. In short, the full simplification of scoping that you are looking for is only possible when
1. we insert and deal with allocas for all classical data types
2. we have the proper type distinction for measurements
3. we make certain choices (that would be more consistent with Python but potentially suboptimal for optimization) for quantum data types. 

Hence, the symbol table is written to completely abstract this from the bridge code; we can easily choose which data types we impose additional restrictions on compared to python and which ones we don't. I believe this is a good thing to have that gives us all we need to easily tweak it in the future.

python/cudaq/kernel/ast_bridge.py
bettinaheim: Yes, I know you know. But now the code also makes this distinction (it really didn't before the bridge changes). :) 
The fixme was just in an odd place and this should be implemented in the symbol binding, not upon value creation.

schweitzpgi:
> > The title of this PR suggests some fundamental misunderstanding to me as Python functions do not have block scopes. Lowering Python to variable semantics for both value-type and reference-type objects, with or without bindings can be done in a straightforward way, which I think we're not quite arriving at yet.
> > In short, there are absolutely _no scopes_. We should not be simulating scopes because it is deeply incorrect. We should have no notion of scope in the Python bridge.
> > ```python
> > def python(a, b, c):
> >    if a:
> >      local1 = b
> >      local3 = [a]
> >    else:
> >      local2 = c
> >      local3 = [a, a]
> >    # [*]
> >    ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The symbols `local1`, `local2`, and `local3` have the exact same scope. That scope is "Local" scope in Python parlance. The interpreter knows about all these symbols and when executing the code below `# [*]` it will either lookup the value that the symbol is bound to or throw an exception that the symbol has not been bound.
> > `local3` is a symbol bound to a reference-type. A reference-type is _an object_ of non-specific scope. `local3` binds to the reference to the object. The construction of the list container will not have side-effects by the definition of CUDA-Q, so these containers can be hoisted for example.
> > I hope to make some slides that will clarify this and present to the team in January.
> 
> Well, there are scopes, though we only need to worry about the one per function (the fixme in the table). However, the bridge connects python with MLIR, and while python does not have scopes, MLIR does, and until we eliminate any kind of local definition of a variable _in MLIR_ we need to keep track of where these are.

No. MLIR does not have scopes per se within a `func.func`. `cc.scope` is an artificial IR construct to assist in lowering.

> I my though is to get the compiler restructure fully fleshed out and confirm that the compiler can still deal with all needed code optimizations for qpus, before we add more alloca. Additionally, we do need to consider a bit more carefully what we want to do with quantum data types. In short, the full simplification of scoping that you are looking for is only possible when
> 
> 1. we insert and deal with allocas for all classical data types
> 2. we have the proper type distinction for measurements
> 3. we make certain choices (that would be more consistent with Python but potentially suboptimal for optimization) for quantum data types.

The MLIR dialects already have the ability to allocate storage for classical data types and have distinct measurement types. These have existed for a long time now.
 
> Hence, the symbol table is written to completely abstract this from the bridge code; we can easily choose which data types we impose additional restrictions on compared to python and which ones we don't. I believe this is a good thing to have that gives us all we need to easily tweak it in the future.

I agree that this needs more conversation. Like I said, I'll try to pull together some visuals and lead a discussion when everyone is back. It's not simply a question of "do things like Python", but also how the CUDA-Q DSL design should behave. (Which is never _exactly_ like the host language.)

Have a good holiday.

PR #3691, opened 2025-12-17
[runtime] Track current QPU in ExecutionContext instead of platform

### LINKED ISSUE #3700: Race condition on quantum_platform::set_exec_ctx
### Required prerequisites

### Describe the bug

The test below current fails (sometimes) on `main`. The reason for this is that a global `executionContext` is used in `quantum_platform`, whereas the execution context should be thread-local

```cpp
#include "CUDAQTestUtils.h"
#include "common/ExecutionContext.h"
#include <atomic>
#include <thread>
#include <vector>

using namespace cudaq;

/// Test that execution contexts are thread-local.
/// Each thread sets its own context, verifies it can retrieve it,
/// and then resets it.
CUDAQ_TEST(ExecutionContextThreadTester, checkThreadLocalContext) {
  constexpr int numThreads = 4;

  std::atomic<int> successCount{0};
  std::vector<std::thread> threads;
  threads.reserve(numThreads);

  for (int i = 0; i < numThreads; ++i) {
    threads.emplace_back([&platform, &successCount, i]() {
      // Create a unique execution context for this thread.
      // Use qpuId to give each context a unique identifier.
      ExecutionContext ctx("sample", /*shots=*/100, /*qpu_id=*/0);
      ctx.batchIteration = static_cast<std::size_t>(i);

      // Set the context on the platform for this thread.
      platform.set_exec_ctx(&ctx);

      // Retrieve the context and verify it's the one we set.
      ExecutionContext *retrieved = platform.get_exec_ctx();
      if (retrieved == &ctx &&
          retrieved->batchIteration == static_cast<std::size_t>(i)) {
        successCount.fetch_add(1, std::memory_order_relaxed);
      }

      // Reset the execution context.
      platform.reset_exec_ctx();

      // After reset, context should be null.
      if (platform.get_exec_ctx() == nullptr) {
        successCount.fetch_add(1, std::memory_order_relaxed);
      }
    });
  }

  // Wait for all threads to complete.
  for (auto &t : threads) {
    t.join();
  }

  // Each thread should have 2 successful checks:
  // 1. Context was correctly set and retrieved.
  // 2. Context was null after reset.
  EXPECT_EQ(successCount.load(), numThreads * 2);
}
```

### Expected behavior

The test should pass

Not a regression

### Environment

- **CUDA-Q version**: `main`
- **Python version**: n/a
- **C++ compiler**: n/a
- **Operating system**: n/a

### Suggestions

_No response_

lmondada:
I have started centralising all thread-local state that the runtime requires into the `ExecutionContext` class. This consolidation will make it easier to reason about the runtime refactor. The aim is to remove the tight coupling that currently exists between the global states of the platform, the simulator and the execution context:
-  the platform, QPU, CircuitSimulator and ExecutionManager all keep a `ExecutionContext *` of the current context. This state should be thread-local and only tracked once.
- the platform keeps a global hash table mapping threads to current QPU IDs, but only keeps one global `ExecutionContext *`. Meanwhile, `BaseRemoteSimulatorQPU` keeps a single QPU ID but a hash table mapping threads to execution contexts.... Both QPU IDs and execution contexts should be thread-local, and should be tracked together. 
 
## Plan

Concretely, over the next three PRs, I propose:

1. Move tracking of "current QPU" from platform and QPU code into `ExecutionContext` (this PR)
2. Remove tracking of `ExecutionContext` in platform, QPU and CircuitSimulator code. Introduce a thread-local state for the current `ExecutionContext` (draft [here](https://github.com/lmondada/cuda-quantum/pull/1))
3. Move the thread-local `CircuitSimulator` instance in `runtime/nvqir/NVQIR.cpp` into `ExecutionContext`

 Note that this will temporarily increase the number of attributes tracked in the "god object" `ExecutionContext` further. This is required to centralise all global state in one place. The runtime refactor should then overtime be able to remove elements from that global state.

## This PR

- Adds a `qpuId` attribute to the `ExecutionContext` class.
- Removes the attributes `platformCurrentQPU`, `threadToQpuId` and `threadToQpuIdMutex` from the `quantum_platform` class. This is now tracked by the `qpuId` of the `ExecutionContext`.
- Modifies the `quantum_platform` API as follows:
   - `set_exec_ctx` and `reset_exec_ctx` always apply to the current thread. As a result, the `qpu_id` to which they apply can be inferred from the execution context and does not need to be specified. Note that currently the platform still keeps track of the execution context as a global. This is inconsistent and will be fixed in the next PR.
   - other methods such as `is_simulator`, `supports_conditional_feedback`, `set_noise` etc, which previously applied to the current QPU, now take an optional argument specifying which QPU to apply to.
   - `quantum_platform::executionContext` is changed to track a context per thread. This is temporary as this attribute will be removed altogether in the next PR.

fixes #3700

### DIFF
--- a/python/tests/backends/test_stim.py
+++ b/python/tests/backends/test_stim.py
@@ -16,10 +16,9 @@ import numpy as np
 
 @pytest.fixture(scope="session", autouse=True)
 def setTarget():
-    old_target = cudaq.get_target()
     cudaq.set_target('stim')
     yield
-    cudaq.set_target(old_target)
+    cudaq.reset_target()
 
 
 def test_stim_non_clifford():
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -952,9 +952,16 @@ public:
     nQubitsAllocated++;
     stateDimension = calculateStateDim(nQubitsAllocated);
 
-    if (!isInTracerMode())
+    if (!isInTracerMode()) {
       // Tell the subtype to grow the state representation
-      addQubitToState();
+      try {
+        addQubitToState();
+      } catch (...) {
+        nQubitsAllocated--;
+        stateDimension = previousStateDimension;
+        throw;
+      }
+    }
 
     // May be that the state grows enough that we
     // want to handle observation via sampling
@@ -1009,9 +1016,16 @@ public:
     nQubitsAllocated += count;
     stateDimension = calculateStateDim(nQubitsAllocated);
 
-    if (!isInTracerMode())
+    if (!isInTracerMode()) {
       // Tell the subtype to allocate more qubits
-      addQubitsToState(count, state);
+      try {
+        addQubitsToState(count, state);
+      } catch (...) {
+        nQubitsAllocated -= count;
+        stateDimension = previousStateDimension;
+        throw;
+      }
+    }
 
     // May be that the state grows enough that we
     // want to handle observation via sampling
@@ -1055,9 +1069,16 @@ public:
     nQubitsAllocated += count;
     stateDimension = calculateStateDim(nQubitsAllocated);
 
-    if (!isInTracerMode())
+    if (!isInTracerMode()) {
       // Tell the subtype to allocate more qubits
-      addQubitsToState(*state);
+      try {
+        addQubitsToState(*state);
+      } catch (...) {
+        nQubitsAllocated -= count;
+        stateDimension = previousStateDimension;
+        throw;
+      }
+    }
 
     // May be that the state grows enough that we
     // want to handle observation via sampling
--- a/runtime/nvqir/stim/StimCircuitSimulator.cpp
+++ b/runtime/nvqir/stim/StimCircuitSimulator.cpp
@@ -494,6 +494,9 @@ public:
         return qubits.empty();
       return true;
     }();
+    if (!sampleSim)
+      throw std::runtime_error("Stim simulator state is not initialized. "
+                               "Cannot sample from uninitialized state.");
     assert(shots <= sampleSim->batch_size);
     std::vector<std::uint32_t> stimTargetQubits(qubits.begin(), qubits.end());
     applyOpToSims("M", stimTargetQubits);


docs/sphinx/specification/cudaq/platform.rst
```diff
@@ -12,8 +12,13 @@ information and enable asynchronous quantum kernel invocations.
 **[3]** The :code:`cudaq::quantum_platform` provides an API for querying the number of available 
 quantum processing units (QPUs), with each QPU assigned a logical integer index (:code:`{0,1,2,...}`). 
 
-**[4]** Programmers can specify the ID of the desired QPU and all subsequent CUDA-Q 
-kernel executions will target that QPU.
```
lmondada, line 16: This description was actually out of date: most launches (`sample`, `run`, etc) would overwrite the current qpu ID with a new one (defaulting to 0).

This PR does not modify the current behaviour of CUDAQ, but rather I've taken the opportunity to update the docs.

runtime/cudaq/platform/quantum_platform.h
```diff
@@ -43,6 +43,23 @@ using KernelExecutionTask = std::function<sample_result()>;
...
+/// Will be removed when executionContext is eliminated.
+struct PerThreadExecCtx {
+  PerThreadExecCtx();
+  ~PerThreadExecCtx();
+  ExecutionContext *get() const;
+  void set(ExecutionContext *ctx);
+
+  struct Impl;
+  std::unique_ptr<Impl> impl;
+};
```
lmondada, line 57: I had to add this in temporarily as I was encountering race conditions. It seems CUDAQ was previously working fine when the `executionContext` in `quantum_platform` did not match the one in the QPUs. However this caused problems in my changes.

This will be removed in the next PR, which removes `executionContext` from the `quantum_platform` entirely.

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -49,20 +44,22 @@ quantum_platform *getQuantumPlatformInternal() {
...
+  auto &platformQPU = platformQPUs[qid];
   platformQPU->setNoiseModel(model);
 }
 
-const noise_model *quantum_platform::get_noise() {
-  if (executionContext)
-    return executionContext->noiseModel;
+const noise_model *quantum_platform::get_noise(QpuId qpu_id) {
+  if (auto *ctx = executionContext.get())
+    return ctx->noiseModel;
```
sacpis, line 55: Wondering if noiseModel should be associated with the executionContext or QPU?

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -95,131 +91,101 @@ void quantum_platform::validateQpuId(int qpuId) const {
...
 
 std::optional<QubitConnectivity> quantum_platform::connectivity() {
   return platformQPUs.front()->getConnectivity();
 }
 
-bool quantum_platform::is_simulator(const std::size_t qpu_id) const {
-  validateQpuId(qpu_id);
-  return platformQPUs[qpu_id]->isSimulator();
+bool quantum_platform::is_simulator(QpuId qpu_id) const {
+  std::size_t qid = resolveQpuId(qpu_id);
```
sacpis: I think we can avoid calling `resolveQpuId` for every function of this class, right? I think it would be better to just resolve qpu_id once (most prob in `set_exec_ctx`) and assign it to a local var, and use it across?

runtime/cudaq/platform/quantum_platform.h
```diff
@@ -230,6 +253,17 @@ class quantum_platform {
 private:
   // Helper to validate QPU Id
   void validateQpuId(int qpuId) const;
+
+  /// Resolve @p qpu_id to a QPU ID for this platform.
```
sacpis: What is @p?

runtime/cudaq/platform/quantum_platform.cpp
lmondada: I agree this will need to be cleaned up too. That will be for a separate PR.

runtime/cudaq/platform/quantum_platform.cpp
lmondada: The reason I've added the `qpu_id` argument to these methods is because I do not want the platform to have to track it. If we added a local var, then we'd be managing this state again (and in theory we'd have to deal with state across multiple threads, mutexes etc).

It's a very cheap operation anyway: if an integer is passed, it will just check it's valid. Would you want to be able to skip the validation step in some instances?

runtime/cudaq/platform/quantum_platform.h
lmondada: It's a [doxygen thing](https://www.doxygen.nl/manual/commands.html#cmdp): it means "the parameter called qpu_id".

Are we even using doxygen? Happy to remove it

runtime/cudaq/platform/quantum_platform.cpp
lmondada: Actually, after thinking about your comment, I have removed `resolveQpuId` and reverted the `qpu_id` argument as it was before: defaults to qpu 0 if none is specified.

Thanks for the input!

python/tests/kernel/test_run_kernel.py
```diff
@@ -34,8 +34,10 @@ def is_close_array(actual, expected):
 
 @pytest.fixture(autouse=True)
 def do_something():
+    cudaq.reset_target()
```
sacpis: why do we need to reset the target here?

python/tests/kernel/test_sample_kernel.py
```diff
@@ -17,8 +17,10 @@
 
 @pytest.fixture(autouse=True)
 def do_something():
+    cudaq.reset_target()
```
sacpis: same here.

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -81,144 +81,111 @@ quantum_platform::enqueueAsyncTask(const std::size_t qpu_id,
...
-  auto tid = std::hash<std::thread::id>{}(std::this_thread::get_id());
-  {
-    std::shared_lock lock(threadToQpuIdMutex);
-    auto iter = threadToQpuId.find(tid);
-    if (iter != threadToQpuId.end())
-      qpu_id = iter->second;
-  }
+                                    const std::vector<void *> &rawArgs,
+                                    std::size_t qpu_id) {
+  validateQpuId(qpu_id);
```
sacpis, line 202: Should we call `validateQpuId` only in `set_exec_ctx`, instead of in all functions?

lmondada: I've been copying the existing checks in `is_simulator`, `is_emulator` etc. 

I don't think there is any harm in doing bounds checking on the user input here, given that the user may pass a value that does not match the one in the current execution context (and I plan to remove `set_exec_ctx` in the next PR anyways)

python/tests/kernel/test_run_kernel.py
lmondada: I was encountering weird testing errors due to other tests not resetting the target properly.

I'm trying to remove these and instead make sure that the tests `test_state_mps`, `test_IQM` and `test_OQC` reset the target after the tests.

lmondada:
> Wondering if we should add tests for the changes in this PR?

I've added a test that was failing on main, and is fixed by this PR :) Other than that, this PR just reworks how existing things work.

python/tests/kernel/test_run_kernel.py
lmondada: confirmed! I was able to remove these superfluous `reset_target`s

PR #3698, opened 2025-12-22
[ci] Only run "Check links" on modified files

lmondada:
This PR changes CI checks as follows:
1. for checks on PRs, only check links in modified files.
2. to check for newly dead URLs that our existing documentation may point to, this CI check is scheduled to run nightly

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -5,7 +5,7 @@ header:
     software-name: CUDA-Q
 
     content: |
-      Copyright (c) 2023 - 2025 NVIDIA Corporation & Affiliates.
+      Copyright (c) 2023 - 2026 NVIDIA Corporation & Affiliates.
       All rights reserved.
 
       This source code and the accompanying materials are made available under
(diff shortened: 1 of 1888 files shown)


.github/workflows/repo_checks.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         with:
           use-verbose-mode: "yes"
           config-file: ".github/workflows/config/md_link_check_config.json"        
+          github-action-markdown-link-check: "yes"
```
sacpis: Don't we need `check-modified-files-only` instead?
https://github.com/gaurav-nelson/github-action-markdown-link-check (please check Custom variables table)

lmondada: oh sorry, of course! that was a bad copy paste :/

lmondada:
~I've taken the opportunity to also change the fork of the action that we are tracking, given the original one is now deprecated.~

EDIT: only whitelisted third party actions are allowed to be run, so reverted this to the original one.

sacpis:
> /ok to test 23a0389

khalatepradnya:
Thank you, @lmondada for this fix! I've run into random CI failures due links not working temporarily. I wonder though, if this would hide any genuine broken links? Can we perhaps enable the link checker on nightly / weekly basis?

lmondada:
> Can we perhaps enable the link checker on nightly / weekly basis?

Yes, good idea, will add that!

sacpis:
> /ok to test 305b035

bettinaheim (approved):

PR #3702, opened 2025-12-23
Renaming kernel_builder to kernel

### LINKED ISSUE #20: [RFC] Rename cudaq::kernel_builder -> cudaq::kernel
# Background 
Recently we update the Python bindings for `kernel_builder` to `Kernel` to be more Pythonic. I actually like this a little better. This type in C++ is meant to build Quake function representations, so it is in some sense a kernel builder. But ultimately it is a callable CUDA Quantum kernel, so I think it may be better to just call it `cudaq::kernel`.

sacpis:
Renaming cudaq::kernel_builder to cudaq::kernel.

Fixes #20.


docs/sphinx/api/languages/cpp_api.rst
```diff
@@ -278,13 +278,13 @@ Noise Modeling
 Kernel Builder
 ===============
```
khalatepradnya, line 279: Should this be still called `Kernel Builder`?

docs/sphinx/specification/cudaq/dynamic_kernels.rst
```diff
@@ -4,19 +4,19 @@ Just-in-Time Kernel Creation
 **[1]** CUDA-Q provides a set of programming abstractions for dynamically constructing
 quantum kernel code at runtime. 
 
-**[2]** The callable :code:`cudaq::kernel_builder` abstraction
+**[2]** The callable :code:`cudaq::kernel` abstraction
```
khalatepradnya, line 7: `kernel_builder` is clearer IMO.

python/cudaq/kernel/kernel.py
khalatepradnya, line 1: Q: Is this change for alignment with C++? `cudaq.kernel.kernel` seems less clear to me.

khalatepradnya (commented):
The code changes look fine to me. Isn't this a breaking change to the C++ API? @efratshabtai should weigh in.

python/cudaq/kernel/kernel.py
efratshabtai: What's the motivation for the rename? We do want to avoid breaking changes if possible

docs/sphinx/api/languages/cpp_api.rst
sacpis: Fixed in 2c2affb.

python/cudaq/kernel/kernel.py
sacpis: As per the issue filed

Recently we update the Python bindings for kernel_builder to Kernel to be more Pythonic. I actually like this a little better. This type in C++ is meant to build Quake function representations, so it is in some sense a kernel builder. But ultimately it is a callable CUDA Quantum kernel, so I think it may be better to just call it cudaq::kernel.

docs/sphinx/specification/cudaq/dynamic_kernels.rst
schweitzpgi: Agreed.

sacpis:
Closing this PR.

PR #3703, opened 2025-12-24
cudaq::realtime

caldwellshane:
Open sourcing the prototype for `cudaq::realtime` (fka `qclink`).

- [ ] Scrub instances of "qclink" and "nvqlink" strings in favor of "realtime" namespace

### DIFF
--- a/realtime/.clang-format
+++ b/realtime/.clang-format
@@ -1,0 +1,12 @@
+BasedOnStyle: LLVM
+AlwaysBreakTemplateDeclarations: Yes
+IncludeCategories:
+  - Regex:           '^<'
+    Priority:        4
+  - Regex:           '^"cudaq/'
+    Priority:        3
+  - Regex:           '^"(nvqlink|\.\.)/'
+    Priority:        2
+  - Regex:           '.*'
+    Priority:        1
+InsertNewlineAtEOF: Yes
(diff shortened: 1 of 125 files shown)


realtime/.clang-format
```diff
@@ -0,0 +1,12 @@
+BasedOnStyle: LLVM
```
schweitzpgi, line 1: Why should this code be different than the rest of the repo?

caldwellshane:
Closing this in favor of https://github.com/NVIDIA/cuda-quantum/tree/features/cudaq.realtime

realtime/.clang-format
caldwellshane: Shouldn't be, thanks for catching. Note this has moved to https://github.com/NVIDIA/cuda-quantum/tree/features/cudaq.realtime for further work

caldwellshane:
> It looks like this is a drop of an independent repo into a subdirectory, rather than a subdirectory of new functionality that is integrated into the existing repo. Is that really what we want?

Correct, and right it's a starting point not what we want. Work has moved to https://github.com/NVIDIA/cuda-quantum/tree/features/cudaq.realtime.

schweitzpgi:
There is already a clang format config at the top level. So it was odd to see another one.

This needs to be configured as an optional library into the top-level rather than a pure code dump of another repo's code. IMHO.

PR #3705, opened 2025-12-29
Kill qpud server thread on failure

lmondada:
I was encountering hanging tests for the `remote-mqpu` target. This was due to tests failing without proper cleanup (timeouts or segfaults), leaving the child thread hanging.

On linux,, setting `prctl(PR_SET_PDEATHSIG, SIGTERM);` ensures that the child is killed whenever the parent thread dies.

### DIFF
--- a/tools/cudaq-qpud/RestServerMain.cpp
+++ b/tools/cudaq-qpud/RestServerMain.cpp
@@ -11,6 +11,11 @@
 #include "common/RemoteKernelExecutor.h"
 #include "llvm/Support/CommandLine.h"
 
+#ifdef __linux__
+#include <signal.h>
+#include <sys/prctl.h>
+#endif
+
 // Declare CUDA-Q MPI API that we need since we cannot compile with cudaq.h
 // without RTTI (needed to link this tool against LLVMSupport).
 namespace cudaq {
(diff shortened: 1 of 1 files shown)


tools/cudaq-qpud/RestServerMain.cpp
```diff
@@ -43,6 +48,13 @@ static llvm::cl::opt<bool> printCudaProperties(
     llvm::cl::init(false));
 
 int main(int argc, char **argv) {
+#ifdef __linux__
+  // Request termination signal when parent process dies.
+  // This ensures cleanup when auto-launched by remote-mqpu tests,
+  // even if the parent is killed by SIGKILL (e.g., llvm-lit timeout).
+  prctl(PR_SET_PDEATHSIG, SIGTERM);
```
khalatepradnya, line 55: Q1: Why is this the right place for enabling the setting?

Q2: The title uses `[test`] prefix, but this change affects to `cudaq-qpud` tool in non-test, i.e. production environment as well, right?

khalatepradnya (commented):
This looks to be a good change to address the issue where CI seems to be "hanging" and eventually times out on certain remote target tests.
Thanks, @lmondada for the fix!

tools/cudaq-qpud/RestServerMain.cpp
lmondada: > Why is this the right place for enabling the setting?

This is a config flag that must be set somewhere in the child process. This main function is the entry point for this child process (see `runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp` lines 180-196 for the launch code in the parent thread), so seems like the natural place. Do you have a better suggestion?

> The title uses [test] prefix, but this change affects to cudaq-qpud tool in non-test, i.e. production environment as well, right?

Yes you are right. This handles a "smooth" shutdown whenever there is a runtime error and the remote-mqpu platform is being used on localhost. I've removed `[test]`.

To be clear, the remote-mqpu behavior remains unchanged in the absence of a segfault or other unexpected crash of the cudaq runtime.

khalatepradnya: That makes sense. Thanks!

PR #3709, opened 2026-01-02
Fix and issue for expectation value computations with noise on the tensornet-mps backend

huaweil-nv:
### Description

Fixed two bugs in simulator_mps.h:

1. prepareSpinOpTermData(): Loop iterated over empty `prods` vector instead of `ham`, causing empty term lists to be returned.

2. observe(): Missing `expVal += termExpVals[i]` to accumulate expectation values from each term.

These bugs caused observe() to return 0.0 expectation value when a noise model was applied to the MPS simulator.

Added regression tests in python/tests/kernel/test_state_mps.py

How to repro:
```
import cudaq
from cudaq import spin

# Set MPS backend
cudaq.set_target('tensornet-mps')

@cudaq.kernel
def kernel():
    q = cudaq.qubit()
    x(q)  # |0> -> |1>

# Observable: Z operator (|1> state should give <Z> = -1)
H = spin.z(0)

# Add noise
noise = cudaq.NoiseModel()
noise.add_all_qubit_channel("x", cudaq.BitFlipChannel(0.1))

# Test
result = cudaq.observe(kernel, H, noise_model=noise, num_trajectories=1000)

print(f"Result: {result.expectation()}")
print(f"Expected: ~-0.8")
print(f"Bug present: {abs(result.expectation()) < 0.01}")
```

### DIFF
--- a/runtime/cudaq.h
+++ b/runtime/cudaq.h
@@ -205,22 +205,12 @@ bool kernelHasConditionalFeedback(const std::string &kernelName);
 /// @brief Provide a hook to set the target backend.
 void set_target_backend(const char *backend);
 
-/// @brief Utility function for setting the shots on the platform
-[[deprecated("Specify the number of shots in the using the overloaded sample() "
-             "and observe() functions")]] void
-set_shots(const std::size_t nShots);
-
 /// @brief Set a custom noise model for simulation. The caller must also call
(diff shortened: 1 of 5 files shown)


1tnguyen (approved):
LGTM 👍

1tnguyen:
> /ok to test 9594533f1599828e9751585f345a36c37f5a4282

PR #3693, opened 2025-12-18
Complete revision of the way CUDA-Q embeds in Python

schweitzpgi:
This is a huge change that crosscuts almost every aspect of the system (except for the C++ frontend, maybe?).

The main drivers are:

- Replace all the random, ad hoc compiler pipelines being used in the Python compiler front-end with pipelines that the C++ compiler uses
- To achieve the previous bullet required
  - Changing pipelines that were run backwards or multiple times
  - Getting rid of ad hoc caches of information kept at all levels of the code
  - Deploying proper lambda lifting and beta reduction steps at definition and call, resp.
  - Radical brain surgery on Python kernel launchers (algorithms) so that they could properly resolve symbols in context
  - Fixing the misuse of MLIR symbols
  - Removing ad hoc, blind, and error prone pruning of MLIR code
  - Fixing bugs due to the conflation of kernel definition and kernel call semantics
  - Fixing tests with errors, bugs, etc.
- This includes folding (back) in changes to the Python ast_bridge (#3537)

Related follow-up issues:
- #3678
- #3696
- #3716
- #3727 
- #3754 (was #3729)
- #3732
- address newly added FIXME's in the Python bridge (although, it is not the intent of this feature branch to fix the Python bridge per se) - PR is up
- avoid API changes around get_state
- re-add support for state propagation (get_state followed by passing that state to subsequent kernels) for QPUs
- confirm that all realtime qec Python code still successfully compiles for hw backends (that support it)
- double check migration of the vqe APIs to libraries and any mitigation/deprecation we need to do
- functions defined within kernels are not processed properly

PR #3711, opened 2026-01-04
Enable building and running on macOS

taalexander:
This PR adds macOS support for CUDA-Q, addressing platform-specific differences in linking, symbol visibility, shell compatibility, and library handling. The changes *should* enable building and running CUDA-Q on macOS with Apple Silicon (arm64) and Intel (x86_64) architectures with the test suite passing (outside of several minor limitations noted below).

**This is a large PR** and full Python support also requires Python wheels and CI enablement. I have structured this PR such that it *should* not impact existing Linux builds. I recommend we treat this as phase one and merge after review/passing CI. We will then follow up with Python wheel and CI PRs to complete support.

**update**: This PR is now based on #3693 to prepare for it's imminent merger

PRs:
- [x] #3711
- [x] Python wheel support for macOS
- [x] CI support for macOS

I've tried my best to summarize the contents of the PR below:

### Platform Detection & Configuration
- Added macOS sysroot detection via `xcrun --show-sdk-path` for C++ stdlib headers
- Configured platform-appropriate linker flags (`--no-as-needed` for Linux, alternatives for macOS)
- Added `CUDAQ_LIBCXX_PATH` and `CUDAQ_SYSROOT_PATH` configuration for cross-platform header discovery
- Updated rpath handling: `@executable_path` on macOS vs `$ORIGIN` on Linux
- Changed `CMAKE_INSTALL_RPATH` to use semicolon separators on macOS

### Library Linking Changes
- Moved MLIR/LLVM dependencies from PUBLIC to PRIVATE in `cudaq-mlir-runtime` to reduce symbol visibility issues due to two-level namespace
- Updated plugin extension handling: `.dylib` on macOS vs `.so` on Linux
- Added platform-conditional linking for circular dependencies (`--start-group` not available on Apple ld)
- Force two-level namespace linking for `cudaq-common` to prevent symbol collisions with OpenSSL

### New `cudaq-utils` Library
Created a new low-level utilities library which resolves a circular dependency where `cudaq-operator` needs `complex_matrix` functions but is built before `libcudaq`

---

## 2. Two-Level Namespace Workarounds

macOS uses two-level namespace linking by default, where symbols are bound to specific libraries. This causes issues with LLVM/MLIR's static initializer pattern (PassRegistry, TargetRegistry, cl::Options).

### Workarounds Implemented

| Workaround | Location | Purpose |
|------------|----------|---------|
| `flat_namespace` linker flag | `CMakeLists.txt` | Global symbol visibility |
| `force_load` for LLVM CodeGen | `cmake/BuildHelpers.cmake` | Ensures static initializers run in correct library context |
| `add_lib_loading_macos_workaround` | `cmake/BuildHelpers.cmake` | Helper for Python extension targets to ensure proper library loading order |
| Symbol unexport list | `lib/Support/Config/CMakeLists.txt` | Hides LLVM/MLIR symbols from `CUDAQTargetConfigUtil` to prevent symbol collisions |
| Explicit `InitializeNativeTarget` | `CUDAQuantumExtension.cpp` | macOS-only target registration for Python extension to workaround issues where the targets were registered to the wrong registry copy |
| Execution manager override API | `execution_manager.h` | Allows explicit manager setting across library boundaries to manage behaviour with execution manager default symbol resolution |

### Future Removal Pathway

In later versions of clang the DYLIB linking issues have been fixed to ensure MLIR library links will all be rerouted against the dylibs that are built. We should consider moving to these single MLIR/LLVM dylibs at this point to avoid multiple linkage issues.

---

### Type Size Differences
- `unsigned long` is 8 bytes on macOS/arm64 but 4 bytes on some Linux systems so we switch usages to std::uint64_t
- Updated `CCTypes.cpp` to use explicit size types where needed

### Library Path Handling
- `TargetConfig.cpp`: Handle `.dylib` vs `.so` extensions
- `fixup-linkage.cpp`: Added handling for `define weak` linkage (macOS clang emits some functions with weak linkage)

### Shell Compatibility (POSIX)
Replaced bash commands with POSIX-compatible alternatives:
- `|&` → `2>&1 |` for stderr piping
- Fixed `mktemp` template usage (macOS requires `XXXXXX` suffix)
- Updated shebang and array handling in shell scripts

### Standard Library Differences
- `std::vector<bool>` has different internal layout between libc++ and libstdc++
- Added explicit `extern char **environ` declaration in `MQPUUtils.cpp` (POSIX requires explicit declaration on macOS)

### Other
- Updated Stim CMakeLists.txt to use platform-appropriate symbol hiding syntax

---

## 4. Third-Party Patches

Added patches in `tpls/customizations/` for compatibility:

| Patch | File | Purpose |
|-------|------|---------|
| LLVM idempotent option category | `llvm/idempotent_option_category.diff` | Makes `cl::OptionCategory` registration idempotent to handle multiple LLVM copies registering the same category (avoids assertion failures) |
| Pybind11 LTO flag fix | `pybind11/pybind11Common.cmake.diff` | Fixes [pybind/pybind11#5098](https://github.com/pybind/pybind11/issues/5098) - incorrect `-flto=` flag generation for Clang |

### Xtensor xio.hpp Workaround
- Removed `#include <xtensor/xio.hpp>` in `molecule.cpp` and replaced with manual printing
- Workaround for clang 17-18 template ambiguity with svector's `rebind_container` ([LLVM #91504](https://llvm.org/bugs/show_bug.cgi?id=91504))

---

## 5. (Pre-existing) Bug Fixes

*Note* most of these would have been caught by static code analysis. The majority of the bugs were likely a result of more aggressive allocator on OSX.

| File | Fix |
|------|-----|
| `RegToMem.cpp` | Added proper `WalkResult` return after `op->erase()` to prevent iterator invalidation |
| `LoopUnrollPatterns.inc` | Fixed iterator handling in loop unrolling pattern |
| `ResetBeforeReuse.cpp` | Fixed stale pointer bug caused by canonicalization in Quantinuum pass |
| `CombineMeasurements.cpp` | Clarified success return for erased unused measurements |
| `QuakeToLLVM.cpp` | Fixed variadic argument duplication in controlled rotation codegen - rotation parameters were passed twice to `invokeRotationWithControlQubits`, causing crashes on ARM64 Darwin (masked on x86_64 due to ABI differences in variadic float handling) |

---

## 6. Python Bindings

- Added `#ifdef __APPLE__` conditional for `InitializeNativeTarget` calls
- Updated CMake to use `add_lib_loading_macos_workaround` for Python extension targets
- Fixed Python virtual environment stdlib availability on macOS

---

## 7. Documentation & Developer Setup

- Updated `Dev_Setup.md` for developer environment setup
- Added `requirements-dev.txt` for Python development dependencies
- Updated `Building.md` with platform-specific notes
- Updated `scripts/install_toolchain.sh` and `scripts/install_prerequisites.sh` for macOS build instructions
- Updated `scripts/build_cudaq.sh` with improved macOS build support
- Changed OpenSSL build to use CMake on macOS to avoid pkg-config resolution issues with flat namespace

---

## 8. Test Updates

- Updated test `RUN` lines to use platform-appropriate flags (`calling_convention.cpp`, `infinite_loop.cpp`, kernel exec transform tests)
- Replaced `.so` with `%cudaq_plugin_ext` substitution for cross-platform tests
- Added `DISCOVERY_TIMEOUT 120` to backend unit tests (likely just required for my slow machine)
- Split `qvector_init_from_vector.cpp` to separate large array test (`qvector_init_large_array.cpp`) which is skipped on macOS due to stack size
- Fixed `cudaq-qpud` linking to enforce two-level namespace for braket backend tests

### Stack Size
macOS has a smaller default stack size (8MB) compared to Linux. Some tests with large stack allocations (e.g., large array initializations) may fail. The `qvector_init_large_array.cpp` test is currently skipped on macOS for this reason. Future work may address this via `ulimit -s` or code refactoring.

### flat_namespace
The `flat_namespace` linker flag can cause symbol collisions with system libraries. We should work toward removing this in a follow-up PR.

### LLVM cl::OptionCategory Duplicate Registration (Requires LLVM Patch)

Both `libcudaq` and `cudaq-mlir-runtime` link LLVM and therefore each contain their own copy of LLVM's `cl::OptionCategory` static globals. When both libraries are loaded, LLVM's default behavior asserts on duplicate category names.

The `idempotent_option_category.diff` patch makes registration idempotent, allowing the same category to be registered multiple times without assertion failures. This is a workaround—the proper fix would be to restructure the libraries so only one contains LLVM command-line infrastructure, but that requires more significant refactoring.

### C++ Exception Handling in JIT-compiled Code (macOS ARM64)

On macOS ARM64 (Apple Silicon), C++ exceptions thrown from JIT-compiled code 
cannot be caught by user code. The exception will terminate the program instead 
of unwinding to the catch block. This is could be due to improper exception handling
and for now we have xfailed `targettests/execution/estimate_resources_sample_in_choice.cpp`
which explicitly tests this capability.

---

## Testing

All tests passing on x86 with `ctest --output-on-failure` except one related to exception handling as detailed above and has been marked XFAIL.


sacpis:
> /ok to test 639dac7

sacpis:
> /ok to test 5965f42

sacpis:
> /ok to test d6f87f9

sacpis:
> /ok to test 5b5f4b3

runtime/cudaq/CMakeLists.txt
```diff
@@ -23,8 +23,6 @@ add_library(${LIBRARY_NAME}
                 qis/execution_manager.cpp
                 qis/remote_state.cpp
                 qis/state.cpp
-                utils/cudaq_utils.cpp
-                utils/matrix.cpp
```
bettinaheim: Instead of creating a separate utils library to break up dependencies, can you possibly move matrix.cpp into the cudaq-operator lib? I haven't checked if that causes another dependency issue, but pretty much everything already depends on the operators.

taalexander: Yes, I think this should work. It didn't seem to be being used by much else.

taalexander: Resolved in 8241423

taalexander:
> /ok to test 8241423

runtime/cudaq/CMakeLists.txt
taalexander: I forgot to move the header in the commit above. Included in - 913c245b

Building.md
```diff
@@ -63,6 +63,53 @@ detected running the command `nvidia-smi` in your development environment.
 [cutensor]: https://developer.nvidia.com/cutensor
 [nvidia_driver]: https://www.nvidia.com/download/index.aspx
 
+## Building on macOS
+
+CUDA-Q can be built on macOS for development purposes. Note that:
+
+- **CPU-only**: No CUDA/GPU support is available on macOS
+- **Apple Clang**: Uses the system compiler (no need to install GCC or LLVM separately)
```
schweitzpgi, line 71: Which Apple Clang? Or does this include porting all of our QIR code to a modern version of LLVM?

include/cudaq/Optimizer/Builder/Factory.h
```diff
@@ -44,6 +44,15 @@ namespace factory {
 constexpr const char targetTripleAttrName[] = "llvm.triple";
 constexpr const char targetDataLayoutAttrName[] = "llvm.data_layout";
 
+/// Padding size for std::vector<bool> host type representation.
+/// `libc++` (macOS) uses 24-byte layout, `libstdc++` (Linux) uses 40-byte
+/// layout. The padding is `sizeof(std::vector<bool>)` - `sizeof(void*)`.
+#ifdef _LIBCPP_VERSION
+constexpr std::size_t stdVecBoolPaddingSize = 16;
```
schweitzpgi: This can be written without reliance on magic macro definitions.

lib/Frontend/nvqpp/ConvertStmt.cpp
```diff
@@ -181,7 +184,7 @@ bool QuakeBridgeVisitor::TraverseCXXForRangeStmt(clang::CXXForRangeStmt *x,
             // shouldn't get here, but we can erase the call at minimum
             call->erase();
           }
-          return {i, {}, call.getOperand(1), call.getOperand(3)};
+          return {i, {}, initial, stepBy};
```
schweitzpgi, line 187: Nice catch. Is there a regression test for this bug fix?

lib/Optimizer/Builder/Factory.cpp
```diff
@@ -353,7 +353,7 @@ static cc::StructType stlHostVectorType(Type eleTy) {
   // std::vector<bool> is a different type than std::vector<T>.
   auto ptrTy = cc::PointerType::get(eleTy);
   auto i8Ty = IntegerType::get(ctx, 8);
-  auto padout = cc::ArrayType::get(ctx, i8Ty, 32);
+  auto padout = cc::ArrayType::get(ctx, i8Ty, factory::stdVecBoolPaddingSize);
```
schweitzpgi, line 361: See earlier comment.

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -484,10 +484,18 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   func.func private @__nvqpp_vector_bool_free_temporary_initlists(!cc.ptr<i8>) -> ()
 )#"},
 
-    // __nvqpp_vector_bool_to_initializer_list
+// __nvqpp_vector_bool_to_initializer_list
+// The array size must match factory::stdVecBoolPaddingSize (16 for libc++,
+// 32 for libstdc++) to match the host std::vector<bool> layout.
+#ifdef _LIBCPP_VERSION
```
schweitzpgi: We should be able to splice the string without resorting to macros.

lib/Optimizer/Dialect/CC/CCTypes.cpp
```diff
@@ -107,7 +107,7 @@ cc::StructType::getPreferredAlignment(const DataLayout &dataLayout,
 LogicalResult
 cc::StructType::verify(llvm::function_ref<mlir::InFlightDiagnostic()> emitError,
                        mlir::StringAttr, llvm::ArrayRef<mlir::Type> members,
-                       bool, bool, unsigned long, unsigned int) {
+                       bool, bool, std::uint64_t, unsigned int) {
```
schweitzpgi, line 110: Is there a bug here? i.e., does it really *have* to be 64 bits? What goes wrong is sizeof(long) is 4?

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -133,12 +133,17 @@ class ExtendQubitMeasurePattern : public OpRewritePattern<quake::MzOp> {
...
                                                  ValueRange{veq},
                                                  measure.getRegisterNameAttr());
-      else if (measure.use_empty())
+        return success();
+      }
+      if (measure.use_empty()) {
+        // Erasing an unused intermediate measurement is a valid rewrite action.
         rewriter.eraseOp(measure);
+        return success();
+      }
```
schweitzpgi, line 146: This looks like a serious bug. Regression test?

lib/Optimizer/Transforms/LoopUnrollPatterns.inc
```diff
@@ -147,8 +147,7 @@ struct UnrollCountedLoop : public OpRewritePattern<cudaq::cc::LoopOp> {
           auto termOpers = cont.getOperands();
           rewriter.setInsertionPoint(cont);
           rewriter.replaceOpWithNewOp<cf::BranchOp>(cont, contBlock, termOpers);
-        }
-        if (allowBreak) {
+        } else if (allowBreak) { // term is invalid after replace
```
schweitzpgi: ? Where is term replaced? I'm not seeing it.

Also, that's not true. `replaceOpWithNewOp` does not erase the Op. The Op remains in the IR until the rewriter reaches a point to commit all the changes. It's entirely possible the `replaceOpWithNewOp` will do nothing, in fact.

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
```diff
@@ -138,12 +154,14 @@ class ResetAfterMeasurePattern : public OpRewritePattern<quake::MzOp> {
...
             });
+        modified = true;
       } else {
         LLVM_DEBUG(llvm::dbgs() << "No next use\n");
       }
     }
 
-    return failure();
+    // Return success if we modified the IR, failure otherwise.
+    return modified ? success() : failure();
```
schweitzpgi: I think this could even be written
```suggestion
    return success(modified);
```

python/tests/mlir/target/mapping.py
```diff
@@ -6,7 +6,7 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-# RUN: CUDAQ_DUMP_JIT_IR=1 PYTHONPATH=../../.. python3 %s --target oqc --emulate |& FileCheck %s
+# RUN: CUDAQ_DUMP_JIT_IR=1 PYTHONPATH=../../.. python3 %s --target oqc --emulate 2>&1 | FileCheck %s
```
schweitzpgi, line 9: This might be a problem. The CI uses different shells, so these redirects were converted to `|&` to accommodate that. Not sure `2>&1` is ok.

runtime/cudaq/algorithms/unitary.h
```diff
@@ -10,7 +10,7 @@
 
 #include "common/Trace.h"
 #include "cudaq/algorithms/draw.h"
-#include "cudaq/utils/matrix.h"
+#include "cudaq/operators/matrix.h"
```
schweitzpgi, line 13: That's interesting.

runtime/cudaq/domains/chemistry/molecule.cpp
```diff
@@ -38,9 +37,18 @@ std::complex<double> &one_body_integrals::operator()(std::size_t p,
...
+  auto arr =
+      xt::adapt(data.get(), shape[0] * shape[1], xt::no_ownership(), shape);
+  std::cerr << "{";
+  for (std::size_t i = 0; i < shape[0]; ++i) {
+    std::cerr << (i > 0 ? " {" : "{");
+    for (std::size_t j = 0; j < shape[1]; ++j)
+      std::cerr << (j > 0 ? ", " : "") << arr(i, j);
+    std::cerr << "}" << (i + 1 < shape[0] ? ",\n" : "");
+  }
+  std::cerr << "}\n";
```
schweitzpgi: ? Is this because MacOS is missing a library? Is it the same output?

runtime/cudaq/platform/mqpu/helpers/MQPUUtils.cpp
```diff
@@ -25,6 +25,10 @@
 #include "cuda_runtime_api.h"
 #endif
 
+// On macOS, environ is not automatically declared; POSIX requires explicit
+// declaration
+extern char **environ;
```
schweitzpgi, line 31: Shouldn't we be including a header?

targettests/Remote-Sim/qvector_init_from_vector.cpp
```diff
@@ -17,16 +17,6 @@
 #include <cudaq.h>
 #include <iostream>
 
-__qpu__ void test_large_double_constant_array() {
```
schweitzpgi: It's not clear why tests are being removed here.

schweitzpgi (changes requested):
I didn't get all the way to the bottom, but wanted to send my initial comments...

This is huge!

runtime/cudaq/algorithms/unitary.h
schweitzpgi: I'm not clear on why this is being moved. Or did we have two copies of matrix? 🤷‍♂️

Building.md
taalexander: This is referring to the build compiler (in this case Apple clang 17). The vendored LLVM 16 is still built and used for all code-generation/jit execution of the code (ie., same mechanism how we can build cudaq with gcc).

We will need to certify/support the wheel on some combination of architecture (x86/arm), MacOS version (Tahoe), Apple clang (17) or xcode tools (26?) and python version (3.12/?)

taalexander:
Now building and test suite passing on Apple M4 Tahoe 26.0.1

lib/Optimizer/Transforms/ResetBeforeReuse.cpp
taalexander: Thanks. Included in #3755 (which we will rebase on after mergers go through).

taalexander:
I have pulled out the bugs that were identified and fixed into separate PRs #3748 #3752 #3755 #3761.

python/tests/mlir/target/mapping.py
taalexander: So far CI has proceeded Ok with this. I will reopen if it pops up again (if so the workaround might be duplicate tests conditioned by platform).

lib/Frontend/nvqpp/ConvertStmt.cpp
taalexander: I've tried to pull this bug out and reproduce it on my Linux machine with ASAN and was unable to. For now the regression test is running the test suite on MacOS (specifically test/AST-Quake/ranged_for.cpp). For that reason I think it should go in with the series of PRs, however, if you want me to pull it out as a separate PR to `main` without a test I can also do that.

lib/Optimizer/Transforms/CombineMeasurements.cpp
taalexander: Similar to above I couldn't reproduce this on Linux even with ASAN. However, this segfaults everytime on MacOS with `test/Transforms/combine_measurements.qke`. I'd also like to keep this fix in the PR and going forward the regression test will be running CI on MacOS (to be added in an upcoming PR).

lib/Optimizer/Dialect/CC/CCTypes.cpp
taalexander: The problem is [here](https://github.com/NVIDIA/cuda-quantum/blob/42d303c4d9d52e0e12a3aca271c84e7c3229f1cc/include/cudaq/Optimizer/Dialect/CC/CCTypes.td#L81) `StructType` is defined as `std::uint64_t`. Clang is giving a type mismatch on MacOS with `error: out-of-line definition of `verify` does not match any declaration ...`. 

The reason for this is that in the Mac SDK `std::uint64_t` is defined as `unsigned long long`.

```bash

cat /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/_types/_uint64_t.h

...

#ifndef _UINT64_T
#define _UINT64_T
typedef unsigned long long uint64_t;
#endif /* _UINT64_T */

```

lib/Optimizer/Transforms/LoopUnrollPatterns.inc
taalexander: I've moved this bug to it's own PR #3752. 

> ? Where is term replaced? I'm not seeing it.

Without the fix it would be on line 149. It's  a bit hard to see as the dynamic cast have different variable names but `cont` and `brk` would be the same operation.

> Also, that's not true. replaceOpWithNewOp does not erase the Op. The Op remains in the IR until the rewriter reaches a point to commit all the changes. It's entirely possible the replaceOpWithNewOp will do nothing, in fact.

That is also what I thought. However, with a bit of digging this is using the `GreedyPatternRewriteDiver` which ultimately resolves it's call to (as far as I can tell, I very much could be wrong here)

https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/mlir/lib/IR/PatternMatch.cpp#L270 

It seems like the resolution goes `GreedyPatternRewriteDriver -> PatternRewriter -> RewriterBase` which does the erase in `RewriterBase::replaceOp`. The method isn't overloaded along the way.

On the other hand `ConversionPatternRewriter` overrides the method to use the notify mechanism 

https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/mlir/lib/Transforms/Utils/DialectConversion.cpp#L1513-L1519

which I think would give the behaviour you/I would expect?

runtime/cudaq/algorithms/unitary.h
taalexander: The root problem here was that `cudaq-operator` needed the matrix functionality but the matrix functionality was linked into `cudaq-utils`. Therefore both `cudaq` and `cudaq-operator` were linking `cudaq-utils`. However, since `cudaq` *also* links `cudaq-operator` this was causing issues with the two-level namespace and duplicate symbols.

Initially, I tried to resolve this by creating a separate library for just the matrix symbols in utils. However, @bettinaheim pointed out that this library is basically only used by the operator library and we could just move it there https://github.com/NVIDIA/cuda-quantum/pull/3711#discussion_r2662154382

include/cudaq/Optimizer/Builder/Factory.h
taalexander: Great point. Fixed in 43d69a5459.

lib/Optimizer/Builder/Factory.cpp
taalexander: Fixed in 43d69a5459.

lib/Optimizer/Builder/Intrinsics.cpp
taalexander: I've done a different approach to try and resolve this in 6619dee

runtime/cudaq/platform/mqpu/helpers/MQPUUtils.cpp
taalexander: I'm certainly no expert on this but the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html) seems to just make this available as extern with no header? 

There is a header library which exists but it is mac specific and would require macros to inject (Which we could use although I think the current approach is better).

targettests/Remote-Sim/qvector_init_from_vector.cpp
taalexander: This is related to the known limitations in the PR message/documentation above:

```
Stack Size
macOS has a smaller default stack size (8MB) compared to Linux. Some tests with large stack allocations (e.g., large array initializations) may fail. The qvector_init_large_array.cpp test is currently skipped on macOS for this reason. Future work may address this via ulimit -s or code refactoring.
```

This test was failing on MacOS and as it was a segfault causing all other tests in the file to also fail. I move it to its own separate test `qvector_init_large_array.cpp` and then XFAILed that test on Darwin. This way we can keep coverage for the other tests on MacOS and also have XFAIL coverage for the unsupported case (large stack allocations) if we were to fix it.

schweitzpgi:
You're right. It's not defined in a standard header file. My bad.

runtime/cudaq/domains/chemistry/molecule.cpp
taalexander: > Is this because MacOS is missing a library? 

This is due to this bug https://github.com/llvm/llvm-project/issues/91504ma in Clang 17/18 (as the source compiler not the nvq compiler). 

> Is it the same output?

It actually *wasn't* identical to xtensor (getting it perfect was difficult and I gave up). However, your comment made me revisit this and see if we could figure out how to resolve the right template specialization. The fix is in a20b240. I think we could remove the macro and make it unconditional, however, it does only impact the macroed platforms as far as I can tell.

scripts/build_cudaq.sh
```diff
@@ -54,19 +58,35 @@ CUDAQ_INSTALL_PREFIX=${CUDAQ_INSTALL_PREFIX:-"$HOME/.cudaq"}
...
+  else
+    args_before_sep+=("$arg")
+  fi
+done
+set -- "${args_before_sep[@]}"
 
 # Run the script from the top-level of the repo
 working_dir=`pwd`
-this_file_dir=`dirname "$(readlink -f "${BASH_SOURCE[0]}")"`
+this_file_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
mitchdz: One small thing here, is that this is slightly different than the previous command, since the directory is resolved after the `cd`, symlinks will not resolve.

We can return the logical path (closest to readlink) with `pwd -P`

```suggestion
this_file_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
```

I think this will only cause an issue if someone does something like:
```
ln -s /home/user/work/cuda-quantum ~/cudaq
cd ~/cudaq
./scripts/build_cudaq.sh
```
But based on how we use this variable, it is probably fine without it.

scripts/build_cudaq.sh
```diff
@@ -175,11 +213,17 @@ if [ -z "$CUDAHOSTCXX" ] && [ -z "$CUDAFLAGS" ]; then
   fi
 fi
 
-# Determine OpenMP flags
-if [ -n "$(find "$LLVM_INSTALL_PREFIX" -name 'libomp.so')" ]; then
-  OMP_LIBRARY=${OMP_LIBRARY:-libomp}
-  OpenMP_libomp_LIBRARY=${OMP_LIBRARY#lib}
-  OpenMP_FLAGS="${OpenMP_FLAGS:-'-fopenmp'}"
+# Determine OpenMP flags (check for .so on Linux, .dylib on macOS)
+OpenMP_libomp_LIBRARY_PATH=$(find "$LLVM_INSTALL_PREFIX" -name 'libomp.so' -o -name 'libomp.dylib' 2>/dev/null | head -1)
```
mitchdz: Parenthesis when using `-or` with find.

```suggestion
OpenMP_libomp_LIBRARY_PATH=$(find "$LLVM_INSTALL_PREFIX" \( -name 'libomp.so' -o -name 'libomp.dylib' \) 2>/dev/null | head -1)
```

Because currently it is treated as:

`find ("$LLVM_INSTALL_PREFIX" AND -name 'libomp.so') OR (-name 'libomp.dylib')`

Good stackoverflow explaining it - https://stackoverflow.com/a/24339320

This can cause issues if you have libomp.dylib anywhere in the path _before_ libomp.so that is not part of the LLVM build (since `| head -1` is used here).

scripts/build_cudaq.sh
```diff
@@ -211,33 +255,39 @@ cmake_args="-G Ninja '"$repo_root"' \
...
-  -DCMAKE_COMPILE_WARNING_AS_ERROR=${CUDAQ_WERROR:-ON}"
+  -DCMAKE_COMPILE_WARNING_AS_ERROR=${CUDAQ_WERROR:-ON} \
+  $extra_cmake_args"
+
+# Add CUDA-specific flags only on non-macOS systems
+if [ "$(uname)" != "Darwin" ]; then
+  cmake_args="$cmake_args \
+  -DCMAKE_CUDA_COMPILER='"$cuda_driver"' \
+  -DCMAKE_CUDA_FLAGS='"$CUDAFLAGS"' \
+  -DCMAKE_CUDA_HOST_COMPILER='"${CUDAHOSTCXX:-$CXX}"'"
```
mitchdz, line 278: why are the double quotes wrapped in single quotes here?

scripts/install_prerequisites.sh
```diff
@@ -56,25 +55,30 @@ while getopts ":e:t:ml:-:" opt; do
...
-  CURL_INSTALL_PREFIX=${CURL_INSTALL_PREFIX:-/usr/local/curl}
-  AWS_INSTALL_PREFIX=${AWS_INSTALL_PREFIX:-/usr/local/aws}
-  CUQUANTUM_INSTALL_PREFIX=${CUQUANTUM_INSTALL_PREFIX:-/opt/nvidia/cuquantum}
-  CUTENSOR_INSTALL_PREFIX=${CUTENSOR_INSTALL_PREFIX:-/opt/nvidia/cutensor}
+  source "$(dirname "${BASH_SOURCE[0]}")/set_env_defaults.sh"
 fi
 
+echo $PATH
+echo $LLVM_INSTALL_PREFIX
+cmake --help
```
mitchdz: why is `cmake --help` here?

scripts/install_prerequisites.sh
```diff
@@ -56,25 +55,30 @@ while getopts ":e:t:ml:-:" opt; do
...
+
+# Create a temporary directory for building source packages
+PREREQS_BUILD_DIR=$(mktemp -d)
```
mitchdz, line 64: Since we have a `rm -rf` down below, I like to be a little extra cautious. Maybe check that the path exists properly.
```suggestion
PREREQS_BUILD_DIR=$(mktemp -d)
: "${PREREQS_BUILD_DIR:?ERROR mktemp failed}"
```

scripts/set_env_defaults.sh
```diff
@@ -0,0 +1,53 @@
...
+# The script respects existing environment variable values. If a variable
+# is already set, it won't be overridden.
+#
+# Platform-specific defaults:
+#   macOS:  Uses $HOME/.local/* for user-local installs
+#   Linux:  Uses /usr/local/* and /opt/* for system-wide installs
+
+if [ "$(uname)" = "Darwin" ]; then
+  # macOS: user-local installations (no sudo required)
+  mkdir -p $HOME/.local/bin
```
mitchdz: quote variables when mkdir as good practice (in case use has weird $HOME with a space in it)
```suggestion
  mkdir -p "$HOME/.local/bin"
```

scripts/set_env_defaults.sh
```diff
@@ -0,0 +1,53 @@
...
+  mkdir -p $HOME/.local/share/lib
```
mitchdz: same as above

scripts/build_cudaq.sh
taalexander: Will try it out.

scripts/build_cudaq.sh
taalexander: I think I was getting issues with bash expansion on MacOS's older bash but I will try to remove.

scripts/install_prerequisites.sh
taalexander: Woops, debug leftover I missed.

scripts/install_prerequisites.sh
taalexander: Thanks, makes sense.

scripts/build_cudaq.sh
taalexander: Investigated this and this was required as paths with spaces were playing poorly with the ` echo "$cmake_args" | xargs cmake` invocation.

scripts/set_env_defaults.sh
taalexander: Addressed in 1e240890ed

scripts/set_env_defaults.sh
taalexander: Addressed in 1e240890ed

scripts/install_prerequisites.sh
taalexander: Addressed in 1e240890ed

scripts/install_prerequisites.sh
taalexander: Addressed in 1e240890ed

scripts/build_cudaq.sh
taalexander: Addressed in 1e240890ed

scripts/build_cudaq.sh
taalexander: Addressed in 1e240890ed

scripts/build_cudaq.sh
mitchdz: that makes sense, I figured it was something like that. Just wanted to be sure!

lib/Optimizer/Dialect/CC/CCTypes.cpp
taalexander: Closing as addressed for now to ensure we can get this in and stop tracking main. I am happy to go back and address this in a follow up PR.

lib/Optimizer/Transforms/CombineMeasurements.cpp
taalexander: Closing as addressed for now to ensure we can get this in and stop tracking main. I am happy to go back and address this in a follow up PR.

1tnguyen (approved):
👍

PR #3712, opened 2026-01-05
Remove unnecessary restriction during evolution of custom operators

1tnguyen:
### Description

Elementary ops can be defined on multiple degrees of freedom. Also, add a test case for this: explicitly define a coupling operator as a multi-degree custom operator.


sacpis (approved):
LGTM. Thanks @1tnguyen.

PR #3713, opened 2026-01-05
Fix issue during migration to host in tensornet backend

huaweil-nv:
### Description

Fix a bug in `toHostImpl` where memcpy uses `sizeof(ScalarType)` instead of `sizeof(std::complex<ScalarType>)`, causing only half of the state vector data to be copied to the host buffer.

### How to repro
```
// Compile: nvq++ --target tensornet-mps repro.cpp -o repro
#include <cudaq.h>
#include <iostream>
#include <vector>
#include <complex>
#include <cmath>

__qpu__ void bell() {
  cudaq::qubit q, r;
  h(q);
  cx(q, r);
}

int main() {
  auto state = cudaq::get_state(bell);
  
  // Initialize buffer with sentinel values
  std::vector<std::complex<double>> buffer(4, {-999.0, -999.0});
  
  // Copy state to host
  state.to_host(buffer.data(), 4);
  
  // Check results - elements 2 and 3 still contain sentinel values!
  for (int i = 0; i < 4; i++)
    std::cout << "buffer[" << i << "] = " << buffer[i] << "\n";
  
  return 0;
}
```
Expected output:
```
buffer[0] = (0.707107,0)
buffer[1] = (0,0)
buffer[2] = (0,0)
buffer[3] = (0.707107,0)
```
Actual output (before fix):
```
buffer[0] = (0.707107,0)
buffer[1] = (0,0)
buffer[2] = (-999,-999)   ← NOT COPIED!
buffer[3] = (-999,-999)   ← NOT COPIED!
```

### Problem
In both `mps_simulation_state.inc` and `tn_simulation_state.inc`, the `toHostImpl` function calculates the memcpy byte count incorrectly:
```
// Before (incorrect) - line 691 in mps_simulation_state.inc
std::memcpy(clientAllocatedData, stateVec.data(),
            numElements * sizeof(ScalarType));  // Only copies half the data!
```
Since stateVec contains std::complex<ScalarType> elements

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -5,7 +5,7 @@ header:
     software-name: CUDA-Q
 
     content: |
-      Copyright (c) 2023 - 2025 NVIDIA Corporation & Affiliates.
+      Copyright (c) 2023 - 2026 NVIDIA Corporation & Affiliates.
       All rights reserved.
 
       This source code and the accompanying materials are made available under
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # ============================================================================ #
-# Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                   #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
 # All rights reserved.                                                         #
 #                                                                              #
 # This source code and the accompanying materials are made available under     #
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 Apache License Version 2.0
 
-Copyright (c) 2023 - 2025 NVIDIA Corporation & Affiliates
+Copyright (c) 2023 - 2026 NVIDIA Corporation & Affiliates
 All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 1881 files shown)


sacpis:
> /ok to test d138ebd

unittests/integration/mps_to_host_tester.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+  // Second element would be wrong if only half was copied
+  EXPECT_NEAR(expected, hostBuffer[1].real(), 1e-6);
+  EXPECT_NEAR(0.0, hostBuffer[1].imag(), 1e-6);
+}
+
+#endif // CUDAQ_BACKEND_TENSORNET_MPS
+
+// Test for fp32 precision backend (tensornet-mps-fp32)
+#ifdef CUDAQ_SIMULATION_SCALAR_FP32
```
1tnguyen, line 150: I wonder if this `#ifdef CUDAQ_SIMULATION_SCALAR_FP32` should be within the `CUDAQ_BACKEND_TENSORNET_MPS` `#ifdef`.
i.e., we want to test this `fp32` for MPS, not other simulator backends?

unittests/integration/mps_to_host_tester.cpp
1tnguyen, line 1: nit: could we rename the file to `mps_to_host_tester.cpp`?

unittests/integration/mps_to_host_tester.cpp
huaweil-nv: fixed in new commit, thanks for the review.

unittests/integration/mps_to_host_tester.cpp
huaweil-nv: fixed in new commit.

1tnguyen (approved):
LGTM 👍 Thanks, @huaweil-nv

1tnguyen:
> /ok to test 0f132aef22662ea38cf333a08fdccd8a4efae596

PR #3715, opened 2026-01-06
Moving function definitions of DeviceCodeRegistry.cpp from cudaq.h to DeviceCodeRegistry.h

Renaud-K:
It is helpful to have the functions defined in DeviceCodeRegistery.cpp be defined in their own header: DeviceCodeRegistry.h.

DeviceCodeRegistry is isolated into its own module which makes it easier to use with the runtime refactor.

### DIFF
--- a/runtime/common/DeviceCodeRegistry.h
+++ b/runtime/common/DeviceCodeRegistry.h
@@ -7,9 +7,12 @@
  ******************************************************************************/
 
 #pragma once
+#include <cstdint>
+#include <optional>
 #include <string>
 
-namespace cudaq::registry {
+namespace cudaq {
+namespace registry {
 extern "C" {
 void __cudaq_deviceCodeHolderAdd(const char *, const char *);
 void cudaqRegisterKernelName(const char *);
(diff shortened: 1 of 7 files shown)


runtime/common/DeviceCodeRegistry.h
```diff
@@ -1,15 +1,18 @@
-/****************************************************************-*- C++ -*-****
- * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+/*******************************************************************************
```
schweitzpgi: Do not remove the emacs tags. Thanks.

runtime/common/DeviceCodeRegistry.h
```diff
@@ -1,15 +1,18 @@
...
+ * Copyright (c) 2025 NVIDIA Corporation & Affiliates.                         *
```
schweitzpgi: 2026? :)

Renaud-K:
> /ok to test 8bb682720904a24ac6c8a575dea116310554b108

runtime/common/DeviceCodeRegistry.h
Renaud-K: Done. Not to sure how that happened but good to know about emacs tags.

PR #3721, opened 2026-01-07
calling clearkernelregistries before reset_target

sacpis:
calling clearkernelregistries before reset_target


sacpis:
Closing this PR and opening a new one as seems like I don't have permission to push to this PR. Strange.

PR #3722, opened 2026-01-08
[TESTING] Check whether a disabled test is still valid

1tnguyen:


1tnguyen:
It doesn't work yet. Would need the Python redesign.

PR #3723, opened 2026-01-08
adding log level and -s for pytest

sacpis:
adding log level and -s for pytest


1tnguyen:
Finished debugging. Closing.

PR #3724, opened 2026-01-08
Fix for seg fault

sacpis:
Fix for seg fault in photonics and evolve tests.


schweitzpgi (approved):
Hope this puts it over the top.

PR #3730, opened 2026-01-08
Unify Python and C++ MLIR initialization

Renaud-K:
The C++ runtime links with libcudaq-mlir-runtime.so for its MLIR dependencies.
```
add_library(cudaq-mlir-runtime
  SHARED
    ArgumentConversion.cpp
    CodeGenConfig.cpp
    Environment.cpp
    JIT.cpp
    Logger.cpp
    RuntimeMLIR.cpp
    RunTheKernel.cpp
)
```

The python MLIR extension already contains the MLIR dependencies and links in these files explicitly
```
declare_mlir_python_extension(CUDAQuantumPythonSources.Extension
...
    ../runtime/utils/PyRemoteRESTQPU.cpp
    ../runtime/utils/PyFermioniqRESTQPU.cpp
    ../runtime/utils/PyRemoteSimulatorQPU.cpp
    ../runtime/utils/PyRestRemoteClient.cpp
    ../../runtime/common/ArgumentConversion.cpp
    ../../runtime/common/CodeGenConfig.cpp
    ../../runtime/common/RunTheKernel.cpp
```

It brings in the MLIR initialization code through `PyRemoteRESTQPU.cpp` which essentially does a cut-and-paste of `RuntimeMLIR.cpp` and includes `RuntimeMLIRCommonImpl.h` which is hacky.

Under these changes, RuntimeMLIR.cpp is added to the list of common files and differences between C++ and Python are handled with 2 new files: `RuntimeCppMLIR.cpp` and `RuntimePyMLIR.cpp`

This allows us to call the same initialization function for all QPUs, which is a first step towards moving compilation out of the QPU. The next step will be to remove the need for virtual functions which will then allow us to have generic code into the BaseRemoteRESTQPU class, that can then be moved out.

After the python refactor, there are 2 ways of getting the MLIR context. Either we get a brand new one and we are responsible for its destruction or we use a scratch one. These changes formalize a bit more this difference with 2 new calls: `cudaq::getOwningMLIRContext` and `cudaq::getMLIRContext`. And since `extractQuakeCodeAndContext` is always getting a new context but leaving it to program logic to clean it up, I thought it was time for the clean-up (`cleanupContext`) to be cleaned up. Since we are already creating an std::unique_ptr<MLIRContext> we can simply return it and let it be automatically cleaned up when lowerQuakeCode is done with it.

### DIFF
--- a/docs/sphinx/applications/python/bernstein_vazirani.ipynb
+++ b/docs/sphinx/applications/python/bernstein_vazirani.ipynb
@@ -9,11 +9,11 @@
     "\n",
     "The Bernstein-Vazirani (BV) algorithm demonstrates an exponential speedup over classical methods for the particular task: \n",
     "\n",
-    "Suppose we have an oracle that implements the function $f: \\{0,1\\}^n \\longrightarrow \\{0,1\\}$. \n",
+    "Suppose we have an oracle that implements the function $f\\colon \\{0,1\\}^n \\longrightarrow \\{0,1\\}$. \n",
     "\n",
     "$f(x)$ is promised to be the dot product between $x$ and a secret $n$-bit binary string $s$:\n",
     "\n",
-    "$f(x) = x \\cdot s (mod \\space 2) =  (x_1s_1 + x_2s_2 + ... + x_ns_n) (mod \\space 2)$.\n",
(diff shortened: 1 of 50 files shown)


Renaud-K:
> /ok to test 15445d1e3783c1c40cabb63db793b5790823b86d

Renaud-K:
> /ok to test 9fd74a4

PR #3736, opened 2026-01-09
Track current ExecutionContext as a thread local variable

lmondada:
This PR's main change is to remove the tracking of the current execution context from the various components of the runtime (qpu, quantum_platform, simulators and execution managers) and to store it instead in a single place, as a thread-local variable in the new `runtime/common/ExecutionContext.cpp` file.

## Uniform handling of ExecutionContext

Having had issues to track where state is set and reset (e.g. exceptions not cleaning up the state, or context set in some components of the runtime but not all), I have homogenised how the execution context is handled across the runtime. Instead of manually setting and resetting context in the various components, I suggest using a Python style `with` construction:
```python
ctx = ExecutionContext("sample")
with ctx:
    kernel()
```
The intent of this is to make it clear that all code run within such a block can expect the execution context to be set, whereas for all code outside of it, no context exists. It also separates syntactically code run "on the host" and code run "on (a simulation of the) device".

The syntax above is valid in Python. In C++, use `quantum_platform::with_execution_context`, which can be used as follows
```cpp
cudaq::ExecutionContext ctx("sample")
cudaq::get_platform().with_execution_context(ctx, [&]() {
    kernel();
});
```

## Under the hood: Context Lifecycle

At a high level, context setting and resetting was (most of the time) handled as follows:

```
setExecutionContext(ctx)  →  [execute kernel]  →  resetExecutionContext()
```
These set and reset calls were propagated across the various components of the runtime (QPU, simulators, ExecutionManagers, platforms etc). However, these functions had much more responsibilities than just context setting:
- in simulators and execution managers, resetContext was used to "flush" the state and update the context; calling it twice resulted in errors/undefined behaviour
- on the other hand, when exceptions happened or the run was aborted otherwise, resetContext was often not called at all (and could not be called, as it would attempt to finalise the context state). As a result, invalidated context pointers were not always reset
- Some QPUs forwarded set and reset calls to execution managers and/or simulators, but others didn't. This meant that at any one point, various context pointers across the runtime may be pointing at different states. Similarly, execution contexts on different threads were not always tracked separately.
- for `get_state`, `CircuitSimulator::resetExecutionContext` is in charge of moving the `simulationState` into the context after the execution. However, it is only called after DefaultExecutionManager::resetExecutionContext, which is in charge of deallocating qubits. Thus qubits must be marked for deallocation in the simulator but their actual deallocation must be deferred for the simulationState to remain valid long enough
- execution modes such as `tracer` and `resource-count` would often overwrite the execution context global but never reset it back; restoring it actually caused further post-processing to break
- Execution managers had hooks `handleExecutionContextChanged` and `handleExecutionContextEnded` which were triggered precisely at set and reset time.

To resolve and unify this, I suggest the following context lifecycle:

```
configureExecutionContext(ctx)  →  setExecutionContext(ctx)  →  beginExecution()
    →  [execute kernel]  →
finalizeExecutionContext(ctx)  →  endExecution()  →  resetExecutionContext()
```

I have adjusted all runtime components to be using these method names consistently.

These functions work in pairs and have well-defined responsibilities:

### 1. configureExecutionContext(ctx) & finalizeExecutionContext(ctx)
Propagated across the runtime components (QPUs, simulators, execution managers, platforms).

These methods should be (morally) const on the components of the runtime (qpu, simulators etc). Their purpose is to mutate `ctx` for the upcoming execution, resp. flush all data to it after execution. I have not always marked them `const`, as they may mutate internal state (such as queued gates etc).

`configureExecutionContext` is called before setting the context and `finalizeExecutionContext` is called before de-initializing the runtime components.

### 2. beginExecution and endExecution
Propagated across the runtime components (QPUs, simulators, execution managers, platforms).

These methods are used by the components of the runtime to initialise/deinitialise resources for the execution of the kernel. When `beginExecution` is called, the execution context has already been set, so that the component can be initialised accordingly.

### 3. setExecutionContext & resetExecutionContext
Does what it says: sets the thread-local variable to point to the current context. The context remains owned by the calling scope. These are NOT propagated to any runtime components. Further, there should not be any reason to call these functions directly anymore: use `with_execution_context` (see above) to guarantee that the runtime is always left in a consistent state.

### DIFF
--- a/runtime/nvqir/NVQIR.cpp
+++ b/runtime/nvqir/NVQIR.cpp
@@ -698,14 +698,11 @@ static std::vector<std::size_t> safeArrayToVectorSizeT(Array *arr) {
 // user and about the kernel when executed.
 void __quantum__qis__trap(std::int64_t code) {
   if (code == 0) {
-    CUDAQ_WARN("could not autogenerate the adjoint of a kernel");
-    throw std::runtime_error("could not autogenerate the adjoint of a kernel");
+    CUDAQ_ERROR("could not autogenerate the adjoint of a kernel");
   } else if (code == 1) {
-    CUDAQ_WARN("unsupported return type from entry-point kernel");
-    throw std::runtime_error("unsupported return type from entry-point kernel");
+    CUDAQ_ERROR("unsupported return type from entry-point kernel");
   } else {
-    CUDAQ_WARN("code generation failure for target");
-    throw std::runtime_error("code generation failure for target");
+    CUDAQ_ERROR("code generation failure for target");
   }
 }
 
--- a/runtime/common/Logger.cpp
+++ b/runtime/common/Logger.cpp
@@ -96,6 +96,7 @@ namespace details {
 void trace(const std::string_view msg) { spdlog::trace(msg); }
 void info(const std::string_view msg) { spdlog::info(msg); }
 void warn(const std::string_view msg) { spdlog::warn(msg); }
+void error(const std::string_view msg) { spdlog::error(msg); }
 void debug(const std::string_view msg) {
 #ifdef CUDAQ_DEBUG
   spdlog::debug(msg);
--- a/runtime/common/Logger.h
+++ b/runtime/common/Logger.h
@@ -20,12 +20,13 @@ bool isTimingTagEnabled(int tag);
 namespace details {
 // This enum must match spdlog::level enums. This is checked via static_assert
 // in Logger.cpp.
-enum class LogLevel { trace, debug, info, warn };
+enum class LogLevel { trace, debug, info, warn, error };
 bool should_log(const LogLevel logLevel);
 void trace(const std::string_view msg);
 void info(const std::string_view msg);
 void debug(const std::string_view msg);
 void warn(const std::string_view msg);
+void error(const std::string_view msg);
 std::string pathToFileName(const std::string_view fullFilePath);
 } // namespace details
 
@@ -62,6 +63,7 @@ std::string pathToFileName(const std::string_view fullFilePath);
 
 CUDAQ_LOGGER_DEDUCTION_STRUCT(info);
 CUDAQ_LOGGER_DEDUCTION_STRUCT(warn);
+CUDAQ_LOGGER_DEDUCTION_STRUCT(error);
 
 #ifdef CUDAQ_DEBUG
 CUDAQ_LOGGER_DEDUCTION_STRUCT(debug);
@@ -272,6 +274,12 @@ public:
 
 // The following macros avoid the unnecessary processing cost of argument
 // evaluation and string formation until after the log level check is done.
+#define CUDAQ_ERROR(...)                                                       \
+  do {                                                                         \
+    ::cudaq::error(__VA_ARGS__);                                               \
+    throw std::runtime_error(__VA_ARGS__);                                     \
+  } while (false)
+
 #define CUDAQ_WARN(...)                                                        \
   do {                                                                         \
     if (::cudaq::details::should_log(::cudaq::details::LogLevel::warn)) {      \
--- a/runtime/cudaq/platform/quantum_platform.h
+++ b/runtime/cudaq/platform/quantum_platform.h
@@ -85,7 +85,7 @@ public:
     detail::setExecutionContext(&ctx);
     beginExecution();
 
-    auto cleanup = [&]() {
+    auto cleanup = [this, &ctx, &outerContext]() {
       detail::try_finally(
           [this, &ctx] {
             finalizeExecutionContext(ctx);


runtime/cudaq/qis/managers/photonics/PhotonicsExecutionManager.cpp
```diff
@@ -154,69 +154,69 @@ class PhotonicsExecutionManager : public cudaq::BasicExecutionManager {
...
+      for (auto &q : sampleQudits) {
+        const auto measurement_tuple = qpp::measure(
+            state, qpp::cmat::Identity(q.levels, q.levels), {q.id},
+            /*qudit dimension=*/q.levels, /*destructive measmt=*/false);
+        const auto measurement_result = std::get<qpp::RES>(measurement_tuple);
+        const auto &post_meas_states = std::get<qpp::ST>(measurement_tuple);
+        const auto &collapsed_state = post_meas_states[measurement_result];
+        state = Eigen::Map<const qpp::ket>(collapsed_state.data(),
+                                           collapsed_state.size());
       }
```
lmondada, line 206: This looks like a big diff: it is in fact the same old code, just unindented (removing the outer if)

runtime/nvqir/cutensornet/simulator_cutensornet.inc
```diff
@@ -273,9 +276,8 @@ void SimulatorTensorNetBase<ScalarType>::applyNoise(
 
   // Apply all prior gates before applying noise.
   std::vector<int32_t> qubits{targets.begin(), targets.end()};
-  CUDAQ_INFO(
-      "[SimulatorTensorNetBase] Applying kraus channel {} on qubits: {}",
-      cudaq::get_noise_model_type_name(channel.noise_type), qubits);
+  CUDAQ_INFO("[SimulatorTensorNetBase] Applying kraus channel {} on qubits: {}",
+             cudaq::get_noise_model_type_name(channel.noise_type), qubits);
```
lmondada: sorry for the noisy diff here. No changes were made, but it seems formatting on `.inc` files is not enforced.

I can revert this if desired but it is probably useful to have this formatted correctly for the future.

runtime/nvqir/NVQIR.cpp
```diff
@@ -284,26 +286,6 @@ void __quantum__rt__finalize() {
...
-    nvqir::getCircuitSimulatorInternal()->setExecutionContext(ctx);
-  }
-}
-
-/// @brief Reset the Execution Context
-void __quantum__rt__resetExecutionContext() {
-  ScopedTraceWithContext("NVQIR::resetExecutionContext");
-  CUDAQ_INFO("Resetting execution context.");
-  nvqir::getCircuitSimulatorInternal()->resetExecutionContext();
-}
```
lmondada, line 305: I haven't found a use case for these and I don't think we should encourage setting the context manually. Happy to revert if I have missed the reason for their existence.

runtime/nvqir/cutensornet/simulator_cutensornet.inc
```diff
@@ -30,18 +30,19 @@ SimulatorTensorNetBase<ScalarType>::SimulatorTensorNetBase()
...
-  bool deterministic = cudaq::getEnvBool("CUDAQ_TENSORNET_FIND_DETERMINISTIC",
-                                         false);
+  bool deterministic =
+      cudaq::getEnvBool("CUDAQ_TENSORNET_FIND_DETERMINISTIC", false);
   if (deterministic) {
     CUDAQ_INFO("Enabling deterministic tensornet with 1 pathfinding thread");
-    setenv("CUTENSORNET_CONTRACTION_OPTIMIZER_CONFIG_HYPER_NUM_THREADS",
-           "1",1);
+    setenv("CUTENSORNET_CONTRACTION_OPTIMIZER_CONFIG_HYPER_NUM_THREADS", "1",
+           1);
```
lmondada, line 45: sorry for the noisy diff here. No changes were made, but it seems formatting on `.inc` files is not enforced.

I can revert this if desired but it is probably useful to have this formatted correctly for the future.

sacpis:
@lmondada In the following flow

```
configureExecutionContext(ctx)  →  setExecutionContext(ctx)  →  beginExecution()
    →  [execute kernel]  →
finalizeExecutionContext(ctx)  →  endExecution()  →  resetExecutionContext()
```

Shouldn't finalizeExecutionContext happen after endExecution?

lmondada:
haha, I agree with you @sacpis that it would be more symmetric and aesthetically pleasing that way - and trust me, I've tried :P

The issue is that `finalizeExecutionContext` currently requires the components to still be initialized. Think for instance of the queue of gates in simulators: if you freed that memory (in `endExecution`) you wouldn't be able to flush the gates and apply them in `finalizeExecutionContext`.

In the end, I see no harm in the ordering as I suggested it, so I've left it as it is. Maybe we'll be able to clean this up as part of the refactor.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1042,15 +1035,18 @@ class CircuitSimulatorBase : public CircuitSimulator {
     if (!state)
       return allocateQubits(count);
 
-    if (!isInTracerMode() && count != state->getNumQubits())
+    // TODO: remove execution context from CircuitSimulator
+    auto executionContext = cudaq::getExecutionContext();
```
sacpis: I am a bit confused as why do we have `auto executionContext = cudaq::getExecutionContext();` in every method?

lmondada: Before, `executionContext` was a member attribute of `CircuitSimulator`. Now it has to be accessed through `cudaq::getExecutionContext`.

Adding this statement at the beginning of the methods means the rest of the method body does not have to be changed. The better solution would be to rewrite them to not depend on `executionContext` altogether (hence the TODOs). That will be part of the refactor.

schweitzpgi:
I'm confused by this. The execution context is "live" on the host side during a particular kernel launch. Why would we want different threads to be launching different kernels in different contexts concurrently?

Is this related to some concurrency proposal?

lmondada:
I'm marking this as ready to be reviewed. There is one test that still fails that I need to track down: an exception is (correctly) being thrown, but the LIT test fails as the error does not make it into stderr. I can't figure out why and how come this PR changed this behaviour.

For this purpose I have a commit currently pushed called "revertme" -- ignore it please.

runtime/common/ExecutionContext.cpp
```diff
@@ -0,0 +1,50 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "ExecutionContext.h"
+
+namespace {
+/// @brief Thread-local storage for the current execution context.
+thread_local cudaq::ExecutionContext *currentExecutionContext;
```
Renaud-K: Do we not need to initialize to nullptr?

runtime/common/ExecutionContext.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace cudaq {
+
+ExecutionContext *getExecutionContext() { return currentExecutionContext; }
+
+bool isInTracerMode() {
+  return currentExecutionContext && currentExecutionContext->name == "tracer";
+}
+
+bool isInBatchMode() {
+  if (!currentExecutionContext)
```
Renaud-K: nit: Can we implement the logic like the other functions?

utils/mock_qpu/__init__.py
```diff
@@ -6,6 +6,8 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
+from .preallocated_qubits_context import PreallocatedQubitsContext
```
Renaud-K, line 9: Do we need this? I saw a lot of the mock servers doing their own import.

runtime/cudaq/platform/qpu.h
```diff
@@ -160,10 +157,25 @@ class QPU : public registry::RegisteredType<QPU> {
   virtual void
   enqueue(QuantumTask &task) = 0; //{ execution_queue->enqueue(task); }
 
-  /// Set the execution context, meant for subtype specification
-  virtual void setExecutionContext(ExecutionContext *context) = 0;
-  /// Reset the execution context, meant for subtype specification
-  virtual void resetExecutionContext() = 0;
+  /// @brief Configure the execution context for this QPU.
+  virtual void configureExecutionContext(ExecutionContext &context) const {}
```
Renaud-K, line 165: I wonder if these all need to be virtual and if code could not be moved into them.

Renaud-K (approved):
Thank you for changes. It is a step in the right direction. I have a few comments but I am happy with it as is, as it would enable me to move `lowerQuakeCode` tomorrow.

runtime/common/ExecutionContext.cpp
lmondada: yes, sorry for the inconsistency!

utils/mock_qpu/__init__.py
lmondada: no, not strictly speaking, but we'd have to amend all the imports in the mock servers to
```python
from ..preallocated_qubits_context import PreallocatedQubitsContext
```

which is more verbose and repetitive. I'll leave this as it is.

runtime/cudaq/platform/qpu.h
lmondada: Unfortunately right now there are slight variations between the QPUs (e.g. what they do with ExecutionManagers)... I wanted to unify this too, but it will have to wait for a future PR.

runtime/cudaq/utils/cudaq_utils.h
```diff
@@ -7,13 +7,17 @@
  ******************************************************************************/
 
 #pragma once
+#include "common/Logger.h"
```
Renaud-K, line 10: Could you find an alternate way of issuing the exception mechanism than the logger? The issue is that until then `cudaq_utils.h` was a header only utility. It only included std C++ headers. It now adds a dependency to `libcudaq-logger.so` which is a bit unintuitive. Users of header libraries do not expect to have to link with shared libraries. It is not listed in any cmake file, so there is not a way of saying that this is an INTERFACE library with a dependency to other libraries. And I do not believe that cmake accepts dependencies on INTERFACE libraries that are not INTERFACE libraries themselves.

You were picking up an implementation because Logger.cpp was explicitly compiled in cudaq and cudaq-common and there was a big chance that binaries link with one of these.  Now, the issue is more apparent. I can fix it by adding `cudaq-logger` to some 10 new places but it would better if the file could remain a std C++ only implementation.

PR #3742, opened 2026-01-12
Proper scoping for the symbol table

bettinaheim:
Python adds a scope for inner functions. This change addresses a fixme to properly reflect that in the PyScopedSymbolTable. 
Adds a bunch of tests for inner functions, and adds back the tests for captures.
Also cleans up other fixme's I'd added to the bridge on the feature branch.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_run.cpp
+++ b/python/runtime/cudaq/algorithms/py_run.cpp
@@ -12,6 +12,7 @@
 #include "runtime/cudaq/platform/py_alt_launch_kernel.h"
 #include "utils/OpaqueArguments.h"
 #include "mlir/Bindings/Python/PybindAdaptors.h"
+#include "mlir/ExecutionEngine/ExecutionEngine.h"
 #include <future>
 #include <pybind11/complex.h>
 #include <pybind11/functional.h>
@@ -78,13 +79,13 @@ pyRunTheKernel(const std::string &name, quantum_platform &platform,
                                "`list` of `dataclass`/`tuple` from "
                                "entry-point kernels.");
   }
+
   auto results = details::runTheKernel(
       [&]() mutable {
         [[maybe_unused]] auto result =
             clean_launch_module(name, mod, retTy, opaques);
       },
       platform, name, name, shots_count, qpu_id, mod.getOperation());
-
   return results;
 }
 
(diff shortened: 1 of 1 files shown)


python/tests/mlir/ast_decrementing_range.py
bettinaheim, line 1: When I ran the MLIR tests locally, this test failed on my branch. I don't think any of my changes should touch that, so not sure what is up with that.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -5092,6 +5221,17 @@ def visit_AugAssign(self, node):
...
 
+        # NOTE: `AugAssign` must fail if the target is a variable
+        # defined in a parent scope. This behavior is different
+        # for standard assignments to variables, where a new
+        # variable is created in the current scope that shadows the
+        # one in the parent scope. 
+        if (isinstance(node.target, ast.Name) and
+            self.symbolTable.isFromParentScope(node.target.id)):
+            # matching Python error message here
+            self.emitFatalError(f"cannot access local variable '{node.target.id}'" +
```
bettinaheim: This is the error that python gives when you do something like this:
```
i = 5
def f():
  i += 1
f()
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4688,27 +4814,32 @@ def visit_If(self, node):
...
                     cc.ContinueOp([])
                 self.popIfStmtBlockStack()
-                self.symbolTable.popScope()
+                self.symbolTable.endBlock()
 
     def visit_Return(self, node):
 
+        if self.symbolTable.isInnerScope:
+            self.emitFatalError(
+                "functions defined within quantum kernels must not contain return statement",
```
bettinaheim, line 4853: we simply assumed that before but didn't properly enforce it

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1734,26 +1876,13 @@ def functor(qubits):
...
-                self.visit(
-                    node.body)  # only one statement in a python lambda :(
-            cc.ReturnOp([])
-        self.pushValue(createLambda.result)
-        return
+        # Python lambdas can only have a single statement.
+        # Here we will enhance our language by processing a single Tuple
+        # statement as a set of statements for each element of the tuple
+        statements = node.body.elts if isinstance(node.body, ast.Tuple) else [node.body]
+        lambdaFct = self.__createFunctionWithinKernel(node.args.args, statements)
```
bettinaheim: helper function to avoid code duplication

bettinaheim:
> /ok to test 6fb99a3

PR #3747, opened 2026-01-13
[test] Unify mock QPU launch

lmondada:
I would like to add some shared Python logic between the mock QPUs used in testing.

As they were currently launched directly with
```bash
python utils/mock_qpu/<provider>/__init__.py
```
they were not part of a same package and thus code sharing was (unnecessarily) difficult.

I suggest having a server start up script located at `utils/start_mock_qpu.py` that can be used to launch all QPUs through a unified script, passing the backend name as argument.

drive-by: removed the recursive imports in `utils/mock_qpu/__init__.py` as they were all shadowing each other (all mock qpus export more or less the same symbols)

todo: there is a lot more code that could be shared between mock qpus, but I am leaving this for another time.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_state.cpp
+++ b/python/runtime/cudaq/algorithms/py_state.cpp
@@ -8,6 +8,7 @@
 #include "py_state.h"
 #include "LinkedLibraryHolder.h"
 #include "common/ArgumentWrapper.h"
+#include "common/FmtCore.h"
 #include "common/Logger.h"
 #include "cudaq/algorithms/get_state.h"
 #include "runtime/cudaq/platform/py_alt_launch_kernel.h"
--- a/runtime/common/CodeGenConfig.cpp
+++ b/runtime/common/CodeGenConfig.cpp
@@ -7,7 +7,9 @@
  ******************************************************************************/
 
 #include "CodeGenConfig.h"
+#include "FmtCore.h"
 #include "Logger.h"
+#include <stdexcept>
 
 namespace {
 
--- a/runtime/common/Logger.cpp
+++ b/runtime/common/Logger.cpp
@@ -7,8 +7,9 @@
  ******************************************************************************/
 
 #include "Logger.h"
+#include "FmtCore.h"
 #include "Timing.h"
-#include "spdlog/sinks/stdout_color_sinks.h"
+#include "fmt/args.h"
 #include <filesystem>
 #include <set>
 #include <spdlog/cfg/env.h>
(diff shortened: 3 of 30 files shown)


lmondada:
We can hold off merging this in until https://github.com/NVIDIA/cuda-quantum/pull/3693 has been merged (assuming that happens today)

docs/sphinx/using/extending/backend.rst
```diff
@@ -307,7 +307,7 @@ Create unit tests for your server helper:
     #!/bin/bash
     
     # Start the mock server
-    python3 -m utils.mock_qpu.<provider_name> @PORT@ &
+    python3 utils.start_mock_qpu.py <provider_name> &
```
taalexander: Why do we use script mode here, I imagine the script relies on other modules in the package and could be somewhat brittle with imports. 

I imagine this path (`utils`) is relatively brittle as it assumes our cwd is in `python/` or we have it on the path? Would it be better to use the full path (I understand this is just updating existing docs).

utils/start_mock_qpu.py
```diff
@@ -0,0 +1,61 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
```
taalexander, line 9: It may be good practice to add a shebang at the top of the script, 
```suggestion
#!/usr/bin/env python3
# ============================================================================ #
# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
# All rights reserved.                                                         #
#                                                                              #
# This source code and the accompanying materials are made available under     #
# the terms of the Apache License 2.0 which accompanies this distribution.     #
# ============================================================================ #
```

utils/start_mock_qpu.py
```diff
@@ -0,0 +1,61 @@
...
```
taalexander, line 3: Should this live in the `mock_qpu` module?

utils/start_mock_qpu.py
```diff
@@ -0,0 +1,61 @@
...
+from mock_qpu import get_backend_port, all_backend_names
+import uvicorn
+
+
+def main():
+    if len(sys.argv) < 2:
+        print("Please specify a backend: ", ", ".join(all_backend_names()))
+        sys.exit(1)
+
+    backend = sys.argv[1].lower()
```
taalexander: It may be overkill, but I always prefer to start scripts from argparse as they have a habit of growing over time 🤣.

docs/sphinx/using/extending/backend.rst
lmondada: You're right that my edit made no sense: I'm calling it as a script but still using the module "." notation. I meant the following:
```bash
python3 utils/start_mock_qpu.py <provider_name> &
```
I've fixed this.

Re this being brittle and relying on other imports:
- the script imports modules from within the `mock-qpu` folder, that sits in the same folder as the script. Other than that, it just depends on cudaq, (along with other "standard" python packages, which are assumed to be installed)
- here `utils` refers to a directory at the root of the repo, so this seems like an okay way to invoke it? It's confusing, because there are `utils` directories all over the place... What do you think?

I'm a bit fuzzy on the difference between script mode and running this with `-m` within a package. What do you reckon is at risk of breaking? Is it that it may not find the `mock-qpu` package?

utils/start_mock_qpu.py
lmondada: for sure, thanks!

utils/start_mock_qpu.py
lmondada: it could... Do you think that would be better? Part of my issue here is that it's not clear what is a package and what isn't, and hence what should be within and what should be outside.

I have chosen to interpret `mock-qpu` as a "package" and `start_mock_qpu.py` as a stand-alone script that calls stuff within that package. What do you think?

utils/start_mock_qpu.py
lmondada: hmm fair point. No harm in that and we get a nicer help :)

docs/sphinx/using/extending/backend.rst
taalexander: I think the risk was whether `cudaq` is assumed to be installed or not and this was a piece of it. I had missed that `utils/` was at root and therefore agree it could be just a script.

utils/start_mock_qpu.py
taalexander: Sure, that makes sense to me.

utils/mock_qpu/qci/__init__.py
```diff
@@ -1,5 +1,5 @@
 # ============================================================================ #
-# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates and Contributors.  #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
```
khalatepradnya: This was added by one of the external contributors, please retain it.

khalatepradnya (commented):
Thank you for the changes!

PR #3748, opened 2026-01-14
Fix use-after-free in RegToMem cleanup walk

taalexander:
### Description
Fixes a use-after-free bug in the RegToMem pass first uncovered in #3711 where erasing operations during cleanup continued to access the dangling op pointer. The fix returns WalkResult::skip() immediately after op->erase() to prevent accessing freed memory. This bug immediately resulted in crashes on macOS likely due to it's more aggressive allocator.

To reproduce without MacOS the regression test (regtomem-2.qke) runs with MALLOC_PERTURB_=165 which scribbles freed memory with 165, causing the unfixed code to crash in my Linux devcontainer (which the fix fixes).

A proactive way we could avoid this sort of bugs in the future would be to have a build with `-fsanitize=address` (and likely the other fsanitize modes) to catch use-after-free and other memory errors across the codebase. This should be combined with use of a static analyzer like `scan-build`

### DIFF
--- a/docker/release/cudaq.Dockerfile
+++ b/docker/release/cudaq.Dockerfile
@@ -61,8 +61,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && python3 -m pip install --no-cache-dir --break-system-packages numpy scipy \
     && ln -s /bin/python3 /bin/python
 RUN apt-get update && apt-get install -y --no-install-recommends gcc g++ python3-dev \
-    # Ref: https://github.com/qutip/qutip/issues/2412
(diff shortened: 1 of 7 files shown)


schweitzpgi:
Should this be going into the features/python.redesign.0 branch instead of main?

taalexander:
> Thanks for the bug fix. Should we make an issue out of your suggestions to improve testing? (It's an excellent idea.)

Sure, will do. 

> Should this be going into the features/python.redesign.0 branch instead of main?

I think this can go into main. It should stand independent of your feature branch. However, if you think it might impact merger of your branch I'm ok to put it in there. Whatever you would prefer.

PR #3749, opened 2026-01-14
Fix a bug in overlap computation for tensornet-mps backend

huaweil-nv:
### Description

This PR fixes a bug in the MPS simulator's `computeOverlap()` function where the fp32 precision branch was incorrectly discarding the imaginary part of the overlap computation.

Problem:

In `mps_simulation_state.inc:240`, the fp32 branch only read the real part of the complex overlap value:

```
// Before (BUG)
overlap = static_cast<double>(cuCrealf(overlapValue));
```
This caused incorrect results when the inner product <ψ₁|ψ₂> has a non-zero imaginary component. For example:
|ψ₁⟩ = H|0⟩ = (|0⟩ + |1⟩)/√2
|ψ₂⟩ = SH|0⟩ = (|0⟩ + i|1⟩)/√2
⟨ψ₁|ψ₂⟩ = (1 + i)/2
Expected |⟨ψ₁|ψ₂⟩| = √2/2 ≈ 0.7071
Bug result: 0.5 (only real part was used)
The fp64 branch was correctly reading both real and imaginary parts.

How to repro:
```
import numpy as np
import cudaq

# Use fp32 precision MPS simulator
cudaq.set_target('tensornet-mps', option='fp32')

@cudaq.kernel
def kernel_h():
    q = cudaq.qubit()
    h(q)  # |psi1> = (|0> + |1>)/sqrt(2)

@cudaq.kernel
def kernel_sh():
    q = cudaq.qubit()
    h(q)
    s(q)  # |psi2> = (|0> + i|1>)/sqrt(2)

state1 = cudaq.get_state(kernel_h)
state2 = cudaq.get_state(kernel_sh)
overlap = state1.overlap(state2)

# <psi1|psi2> = (1 + i)/2, so |<psi1|psi2>| = sqrt(2)/2 ≈ 0.7071
expected = np.sqrt(2) / 2

print(f"Overlap: {abs(overlap):.4f}")
print(f"Expected: {expected:.4f}")
print(f"Match: {np.isclose(abs(overlap), expected, atol=1e-3)}")

# Before fix: prints 0.5 (WRONG - only real part)
# After fix:  prints 0.7071 (CORRECT)
```

### DIFF
--- a/python/tests/kernel/test_state_mps.py
+++ b/python/tests/kernel/test_state_mps.py
@@ -254,19 +254,8 @@ def test_mps_overlap_complex_inner_product():
     cudaq.__clearKernelRegistries()
     cudaq.set_target('tensornet-mps', option='fp32')
 
-    @cudaq.kernel
-    def kernel_h_fp32():
-        q = cudaq.qubit()
-        h(q)
-
-    @cudaq.kernel
-    def kernel_sh_fp32():
-        q = cudaq.qubit()
-        h(q)
-        s(q)
-
-    state1_fp32 = cudaq.get_state(kernel_h_fp32)
-    state2_fp32 = cudaq.get_state(kernel_sh_fp32)
+    state1_fp32 = cudaq.get_state(kernel_h)
+    state2_fp32 = cudaq.get_state(kernel_sh)
     overlap_fp32 = state1_fp32.overlap(state2_fp32)
 
     # fp32 should also produce correct result (with slightly looser tolerance)


python/tests/kernel/test_state_mps.py
```diff
@@ -208,6 +208,73 @@ def ansatz(theta: float):
...
+    @cudaq.kernel
+    def kernel_h_fp32():
+        q = cudaq.qubit()
+        h(q)
+
+    @cudaq.kernel
+    def kernel_sh_fp32():
+        q = cudaq.qubit()
+        h(q)
+        s(q)
```
1tnguyen: nit: I think we can reuse the `kernel_h` and `kernel_sh` kernels in the test below.

1tnguyen:
> /ok to test 0483fe1ccf4b249c94f05900bbadbd27a7c18f2a

python/tests/kernel/test_state_mps.py
huaweil-nv: fixed in new commit.

1tnguyen:
> /ok to test 743b0ea24895aa9ac15ab48513ca3bc0d457338e

1tnguyen:
Merging. `visualization.ipynb` failure in `CI / Create Docker images / Validation` is a known issue (pending resolution), not related to this PR.

PR #3751, opened 2026-01-14
[ci] Only download specified commit in manylinux dockerfile

lmondada:
We do not need to download the entire git history in `docker/build/devdeps.manylinux.Dockerfile`, we only need one specific commit.

The hope is that this ywill reduce the size of the cached images.


lmondada:
I gave up after spending some time understanding what the `--filter=tree:0` argument in git clone does. It downloads the history tree but only lazily fetches the files/diffs when required.

Switching to picking the exact commit has pros and cons:
- pro: just picking a commit will still save you a bit of data as you don't store the history at all, but not all that much: you are just skipping the list of commit hashes
- con: you have to recompute the history for every git fetch, which is very slow. Also, it seems that if you compute a diff and the parent commits aren't downloaded, then the result you get isn't reliable

We do some weird cherry-picking in scripts/build_llvm.sh so I think that the additional time and overall trickiness of getting the right commits downloaded is not worth the hassle.

PR #3755, opened 2026-01-14
Fix use of dropped pointer in ResetBeforeReuse pass

taalexander:
…ons which could be deleted before use. This fix works by computing on demand instead of caching. This segfaulted 100% of the time on MacOS. Takes special scenario to fail on Linux

### Description

This is a fix for a bug encountered in the MacOS support branch (#3711). 

The `RegUseTracker` was caching pointers at construction time which could interact poorly with passes in-between like canonicalization which remove operations. This is not easy to reproduce on Linux but can be done by using an ASAN build (plus MALLOC_PERTURB_ for good measure) in a semi-reliable manner.

```
cd build
cmake .. \
  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer" \
  -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address"

ninja cudaq-opt test_quantinuum

echo "=== With FIX - 50 iterations ===" && count=0; for i in {1..50}; do MALLOC_PERTURB_=255 ASAN_OPTIONS="quarantine_size_mb=0:detect_stack_use_after_return=1" ./unittests/backends/quantinuum/test_quantinuum --gtest_filter="*QuantinuumNGTester.checkControlledRotations*" 2>&1 | tail -1 | grep -q PASSED && ((count++)); done; echo "$count passed out of 50"
```

for me this segfaults ~20% of the time with the fix correcting this. 
I had tried creating custom tests/passpipelines to more reliably reproduce on Linux but was not able to.

The fix resolves this by avoiding caching and recomputing on demand. ~~I have some performance concern about dropping caching which I will look into.~~

Edit: I've added back a safe caching mechanism that should avoid the performance hit.

### DIFF
--- a/include/cudaq/Optimizer/CodeGen/Passes.td
+++ b/include/cudaq/Optimizer/CodeGen/Passes.td
@@ -131,6 +131,14 @@ def QIRToQIRProfile : Pass<"convert-to-qir-profile"> {
   let constructor = "cudaq::opt::createQIRToQIRProfilePass(\"qir-base\")";
 }
 
+def QirInsertArrayRecord : Pass<"qir-insert-array-record", "mlir::ModuleOp"> {
+  let summary = "Analyze result recording calls and insert new instructions for QIR";
+  let description = [{
+    This pass runs after QIR prep and conversion passes, then inserts array 
+    recording instruction based on the number of result recording calls.
+  }];
+}
+
 def RemoveMeasurements : Pass<"remove-measurements"> {
   let summary =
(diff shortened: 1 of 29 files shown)


1tnguyen (approved):
LGTM 👍

PR #3759, opened 2026-01-15
Bump qutip>5

mitchdz:
Bump qutip. It was held back for older python versions that we no longer support. This is causing issues in CI, and should be bumped anyways.


1tnguyen (approved):
👍

sacpis (approved):

PR #3761, opened 2026-01-15
Fix for bug in QuakeToLLVM where controlled locations were lowered with arguments duplicated in the call

taalexander:
### Description

Originally was discovered in #3711. This was a latent bug that was functioning on x86 as the ABI stores variadic float/integer arguments in separate registers. At runtime the variadic arguments are resolved by type and the duplicate floats are therefore ignored without issue. On MacOS ARM64 the ABI unifies handling of variadic arguments placing them all on the stack. These duplicated arguments corrupt the subsequent arguments resulting in a segfault.

I've added a lit test to verify we're generating the expected signature without the duplicated values.

### DIFF
--- a/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
+++ b/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
@@ -147,8 +147,7 @@ struct UnrollCountedLoop : public OpRewritePattern<cudaq::cc::LoopOp> {
           auto termOpers = cont.getOperands();
           rewriter.setInsertionPoint(cont);
           rewriter.replaceOpWithNewOp<cf::BranchOp>(cont, contBlock, termOpers);
-        }
-        if (allowBreak) {
+        } else if (allowBreak) {
           if (auto brk = dyn_cast<cudaq::cc::BreakOp>(term)) {
             auto termOpers = brk.getOperands();
             rewriter.setInsertionPoint(brk);


lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1008,7 +1009,8 @@ class OneTargetThreeParamRewrite : public ConvertOpToLLVMPattern<OP> {
         cudaq::opt::factory::genLlvmI64Constant(loc, rewriter, numControls));
     funcArgs.push_back(isArrayAndLengthArr);
     funcArgs.push_back(ctrlOpPointer);
-    funcArgs.append(instOperands.begin(), instOperands.end());
+    // Skip instOperands[0..2] since they were already added.
+    funcArgs.append(instOperands.begin() + 3, instOperands.end());
```
Renaud-K: Could we use here some of the operands as defined in `QuakeOps.td` to avoid the hardcoded value.
```
  let arguments = !con(extraArgs, (ins
    UnitAttr:$is_adj,
    Variadic<AnyFloat>:$parameters,
    Variadic<AnyQType>:$controls,
    Variadic<AnyQTargetType>:$targets,
    OptionalAttr<DenseBoolArrayAttr>:$negated_qubit_controls
  ));
...
`

taalexander: Addressed in 96d43d7

Renaud-K: Looks good. Thank you.

PR #3764, opened 2026-01-15
Isolate Logger.h from fmtlib headers

Renaud-K:
With these changes, including Logger.h will no longer include "FmtCore.h" which contains the fmtlib headers. 
This enables us to include Logger.h in header files without bleeding the fmtlib headers into user code.

Logger.h only includes the CUDAQ_ logging capability.
FmtCore.h has not changed and includes all of the fmtlib headers for `fmt::format` and all other formatting capabilities.

So, when Logger.h used to be included , now both `Logger.h` and `FmtCore.h` may have to be included.

In order to use the fmtlib capabilities, Logger.h packs all the argument into an std::array<std::variant<T...>, N>, that is implicitly converted to an std::span<std::variant<T...>>, which calls a non-templated function in Logger.cpp that takes care of the unpacking.

### DIFF
--- a/runtime/common/Logger.h
+++ b/runtime/common/Logger.h
@@ -42,29 +42,23 @@ std::string pathToFileName(const std::string_view fullFilePath);
 namespace details {
 
 //
-// Type traits to check if T is in the variant
+// Concepts to check if T is in the variant
 //
 template <typename T, typename... Types>
-struct is_one_of
-    : std::bool_constant<(
-          (std::is_same_v<T, Types> ||
-           std::is_same_v<std::reference_wrapper<const T>, Types>) ||
-          ...)> {};
+concept one_of = ((std::same_as<T, Types> ||
+                   std::same_as<std::reference_wrapper<const T>, Types>) ||
+                  ...);
 
 template <typename T, typename Variant>
 struct is_variant_member;
 
 template <typename T, typename... Types>
-struct is_variant_member<T, std::variant<Types...>> : is_one_of<T, Types...> {};
+struct is_variant_member<T, std::variant<Types...>>
+    : std::bool_constant<one_of<T, Types...>> {};
 
 template <typename T, typename Variant>
-inline constexpr bool is_variant_member_v =
-    is_variant_member<T, Variant>::value;
-
-// Check all types in a parameter pack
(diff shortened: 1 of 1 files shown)


Renaud-K:
> /ok to test b44556cd57ee86d092d3179a7462fbcecfb1afd8

runtime/common/Logger.h
```diff
@@ -32,6 +39,94 @@ void warn(const std::string_view msg);
...
+                  const std::span<fmt_arg> &arr);
+
+//
+// Functions substitutes for fmt::format and fmt::print
+//
+template <typename... Args>
+std::string format(const std::string_view message, Args &&...args) {
+  static_assert(
+      all_in_variant_v<fmt_arg::storage_t, Args...>,
+      "One or more argument types are not valid variant alternatives. ");
```
lmondada: You could use the shiny new C++20 concepts:

```suggestion
template <typename... Args>
  requires all_in_variant_v<fmt_arg::storage_t, Args...>
std::string format(const std::string_view message, Args &&...args) {
```

runtime/common/Logger.h
```diff
@@ -32,6 +39,94 @@ void warn(const std::string_view msg);
...
+  auto array = std::array<fmt_arg, sizeof...(Args)>{
+      fmt_arg(std::forward<Args>(args))...};
+  return format_packed(message, array);
+}
+
+template <typename... Args>
+void print(const std::string_view message, Args &&...args) {
+  static_assert(
+      all_in_variant_v<fmt_arg::storage_t, Args...>,
+      "One or more argument types are not valid variant alternatives. ");
```
lmondada: and here:

```suggestion
template <typename... Args>
  requires all_in_variant_v<fmt_arg::storage_t, Args...>
void print(const std::string_view message, Args &&...args) {
```

lmondada (approved):
Thanks, very clean and very useful (have run into this issue myself when trying to use `Logger.h` in new places)!

Renaud-K:
> /ok to test 546d163c7a3e421958750306073034a320ff23dd

runtime/common/Logger.h
Renaud-K: I guess, I could now yes.

PR #3768, opened 2026-01-20
Fix density matrix indexing bug for dynamics backend

huaweil-nv:
### Description

This PR fixes a critical bug in `CuDensityMatState::operator()` where density matrix element access used incorrect dimension for bounds checking and linear index calculation.

### Problem
The `operator()` function used the total dimension (`dim*dim`) instead of single-side dimension (`dim`) when accessing density matrix elements:
```
// Before (buggy)
if (indices[0] >= dimension || indices[1] >= dimension)  // dimension = 16 for 4x4 matrix
    throw std::runtime_error("CuDensityMatState indices out of range");
return extractValue(indices[0] * dimension + indices[1]);  // Wrong linear index
```
For a 2-qubit system (4×4 density matrix with dimension=16):
Valid indices should be 0, 1, 2, 3
rho[1,1] computed index 1*16+1=17, causing CUDA memory access error
rho[0,4] passed bounds check (4 < 16) but accessed wrong memory location
Only rho[0,0] worked by coincidence (0*16+0 == 0*4+0)

How to repro:
```
import cudaq
import numpy as np
import cupy as cp
from cudaq import spin

cudaq.set_target("dynamics")

psi0 = cudaq.State.from_data(cp.array([1.0, 0.0, 0.0, 0.0], dtype=cp.complex128))
hamiltonian = 0.0 * spin.z(0) + 0.0 * spin.z(1)
schedule = Schedule(np.linspace(0.0, 0.1, 2), ["t"])

result = cudaq.evolve(
    hamiltonian, {0: 2, 1: 2}, schedule, psi0,
    collapse_operators=[spin.z(0)],
    store_intermediate_results=cudaq.IntermediateResultSave.ALL,
)

rho = result.final_state()
print(rho[0, 0])  # Works
print(rho[1, 1])  # Before: CUDA crash | After: 0j ✓
```

### DIFF
--- a/python/tests/dynamics/test_evolve_dynamics.py
+++ b/python/tests/dynamics/test_evolve_dynamics.py
@@ -154,6 +154,87 @@ def test_batching_bugs():
         assert len(evolution_result.intermediate_states()) == len(steps)
 
 
+def test_density_matrix_indexing():
+    """
+    Test that density matrix element access uses correct indexing.
+    
+    This is a regression test for a bug where the operator() function
+    used the total dimension (dim*dim) instead of single-side dimension (dim)
+    for bounds checking and linear index calculation.
+    
+    For a 2-qubit system (4x4 density matrix with 16 total elements):
+    - Valid indices should be 0, 1, 2, 3
+    - Accessing rho[i, j] should compute linear index as i * 4 + j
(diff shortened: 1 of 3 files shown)


sacpis:
> /ok to test da49c7a

Issue #3642, opened 2025-12-01
Remove fmtheaders from CUDA-Q internal headers


sacpis:
### Required prerequisites

- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

From @amccaskey
The big issue here is that Logger.h is a common header for downstream libraries (so they can get cudaq::info()) and it has FmtCore.h as an include. The changes here do not change that. The real fix is to remove any fmtlib headers from our own headers and hide them in cpp implementation code. I'm not sure how one would do that and keep cudaq::info() variadic though.

Please see [comment](https://github.com/NVIDIA/cuda-quantum/pull/3014#issuecomment-2950937339) on the PR #3014.

sacpis:
Assigning it to @Renaud-K as he has already worked on it in PR #3764. @Renaud-K Can you please confirm and close this issue?

Renaud-K:
#3764 takes care of the biggest issue whereby `FmtCore.h` was included in `Logger.h`, but there are still a few places where `FmtCore.h`  or `fmt/....h`  are directly included in headers files. So, these would probably have to be cleaned up as well before #3014 can proceed.

sacpis:
> [#3764](https://github.com/NVIDIA/cuda-quantum/pull/3764) takes care of the biggest issue whereby `FmtCore.h` was included in `Logger.h`, but there are still a few places where `FmtCore.h` or `fmt/....h` are directly included in headers files. So, these would probably have to be cleaned up as well before [#3014](https://github.com/NVIDIA/cuda-quantum/pull/3014) can proceed.

Thanks @Renaud-K.

Renaud-K:
Fixed with https://github.com/NVIDIA/cuda-quantum/pull/3770

PR #3770, opened 2026-01-20
Removing fmtlib headers from all headers

Renaud-K:
Under these changes:
1. The formatting utilities from Logger.h were moved into a separate `cudaq_fmt.h` header file to 
    - better distinguish formatting from logging. 
    - make the formatting utility available publicly as it is used in multiple headers
2. All header files were grepped for includes of `FmtCore.h` or `fmt/xxxxx.h` and replaced with `cudaq_fmt.h` when appropriate.

This should get us closer to fixing https://github.com/NVIDIA/cuda-quantum/issues/3642

### DIFF
--- a/runtime/common/BraketExecutor.h
+++ b/runtime/common/BraketExecutor.h
@@ -12,7 +12,6 @@
 #include "common/Executor.h"
 #include "common/Logger.h"
 #include "common/SampleResult.h"
-#include "common/cudaq_fmt.h"
 #include "cudaq.h"
 #include <aws/braket/BraketClient.h>
 #include <aws/core/Aws.h>
--- a/runtime/common/JsonConvert.h
+++ b/runtime/common/JsonConvert.h
@@ -9,7 +9,6 @@
 #pragma once
 #include "GPUInfo.h"
 #include "common/ExecutionContext.h"
-#include "common/cudaq_fmt.h"
 #include "cudaq/Support/Version.h"
 #include "cudaq/gradients.h"
 #include "cudaq/optimizers.h"


Renaud-K:
> /ok to test 2d12d11cc59bf4931fb1f64753883231a81c3d82

runtime/common/BraketExecutor.h
```diff
@@ -10,9 +10,9 @@
 
 #include "common/BraketServerHelper.h"
 #include "common/Executor.h"
-#include "common/FmtCore.h"
```
1tnguyen, line 13: I wonder for those header files that include `common/FmtCore.h` but don't have any direct `fmt::` usage, whether we should remove the include and let the compiler figure out where it is actually needed.

runtime/common/cudaq_fmt.h
```diff
@@ -0,0 +1,124 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
```
sacpis, line 2: nit: Should we use the camelCase for this lib? Somthing like `cudaqFmt.h`? or to align with other libs `CudaqFmt.h`?

sacpis (approved):
LGTM. Thanks @Renaud-K.

runtime/common/cudaq_fmt.h
Renaud-K: We seem to abide to a snake case convention for public cudaq interfaces. Moving forward with the refactoring, I think we will see more snake case usage. But at the same time we should be able to replace `fmt::format` with `std::format` when we upgrade to gcc13, so I would not sweat this too much.

sacpis: Thank you @Renaud-K for the information.

runtime/common/BraketExecutor.h
Renaud-K: Yes, I will remove them. They are not needed.

Renaud-K:
> /ok to test 45903cb32eba8552fc848bfec5ad0e44553fa748

PR #3771, opened 2026-01-21
Fix segmentation fault in distributed batched state evolution with `store_intermediate_results`

huaweil-nv:
### Description

This PR fixes a **segmentation fault** that occurs when running batched quantum state evolution with `store_intermediate_results=IntermediateResultSave.ALL` in a multi-GPU MPI environment. 

## Problem Description

When users run batched `cudaq.evolve()` with multiple GPUs and request intermediate results, the program crashes with a segmentation fault:

```python
# This crashes on multi-GPU systems
results = cudaq.evolve(
    hamiltonian,
    dimensions,
    schedule,
    [state1, state2, state3, state4],  # Batched states
    store_intermediate_results=cudaq.IntermediateResultSave.ALL,  # Triggers the bug
    ...
)
```

**Error output:**

```
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PROCESS ID 12345 RUNNING AT hostname
=   EXIT CODE: 11 (Segmentation fault)
===================================================================================
```

### Prerequisites

| Requirement | Description |
|-------------|-------------|
| **Hardware** | Multi-GPU system (2+ NVIDIA GPUs, e.g., 2x A100) |
| **Software** | CUDA-Q with dynamics target (`cudaq.has_target('dynamics')`) |
| **MPI** | MPI runtime with GPU support (e.g., OpenMPI) |
| **cuDensityMat plugin** | Installed (typically via `CUDAQ_DYNAMICS_MPI_COMM_LIB`) |

### Conditions to Trigger the Bug

All **three** conditions must be met simultaneously:

1. **Multi-GPU MPI execution**: `mpirun -np N` where `N > 1`
2. **Batched initial states**: Pass a list of states to `cudaq.evolve()` (e.g., `[state1, state2, state3, state4]`)
3. **Store intermediate results**: `store_intermediate_results=cudaq.IntermediateResultSave.ALL`

**Note:** If any one condition is missing, the bug does NOT occur:

- Single GPU → works fine (no distribution needed)
- Single initial state → works fine (no batching)
- `store_intermediate_results=NONE` or `LAST_ONLY` → works fine (doesn't call `splitBatchedState`)

### Minimal Reproducer

```python
# reproducer.py
import cudaq
import cupy as cp
import numpy as np
from cudaq import spin, Schedule, RungeKuttaIntegrator

cudaq.mpi.initialize()
cudaq.set_target('dynamics')

# Simple single-qubit Hamiltonian
dimensions = {0: 2}
schedule = Schedule(np.linspace(0, 1, 11), ['time'])

# Create 4 distinct initial states (batched)
initial_states = []
for i in range(4):
    theta = i * np.pi / 8
    state_data = cp.array([np.cos(theta), np.sin(theta)], dtype=cp.complex128)
    initial_states.append(cudaq.State.from_data(state_data))

# This crashes before the fix (Segmentation fault or incorrect results)
results = cudaq.evolve(
    hamiltonian,
    dimensions,
    schedule,
    initial_states,  # batched states
    store_intermediate_results=cudaq.IntermediateResultSave.ALL,  # triggers splitBatchedState
    integrator=RungeKuttaIntegrator())

# Verify results (won't reach here before fix)
for i, result in enumerate(results):
    states = result.intermediate_states()
    print(f"Rank {cudaq.mpi.rank()}: State {i} has {len(states)} intermediate states")

cudaq.mpi.finalize()
```
### Run Command

```bash
# Set MPI comm library if needed
export CUDAQ_DYNAMICS_MPI_COMM_LIB=/path/to/libcudaq_distributed_interface_mpi.so

# Run with 2 GPUs
mpirun -np 2 python reproducer.py
```

### Expected Behavior (Before Fix)

```
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   EXIT CODE: 11 (Segmentation fault)
===================================================================================
```

### Expected Behavior (After Fix)

```
Rank 0: State 0 has 11 intermediate states
Rank 0: State 1 has 11 intermediate states
Rank 1: State 2 has 11 intermediate states
Rank 1: State 3 has 11 intermediate states
```

## Root Cause Analysis

The bug is in the **existing** distributed batched state implementation. The code path for distributing batched states across GPUs exists (`distributeBatchedStateData`), but the reverse operation (`splitBatchedState`) was never updated to handle distributed states correctly.

### Bug 1: `CuDensityMatState::splitBatchedState` (C++ layer)

**Original buggy code (line 785-793):**

```cpp
const int64_t stateSize = batchedState.dimension / batchedState.batchSize;
// ...
for (int i = 0; i < batchedState.batchSize; ++i) {
    // Read from ptr + i * stateSize
}
```

**Problem:** In distributed mode:

- `batchedState.dimension` = local buffer size (e.g., 4 for 2 states on this rank)
- `batchedState.batchSize` = total batch size (e.g., 4 states across all ranks)
- `stateSize = 4 / 4 = 1` ❌ (should be 2)
- Loop tries to read 4 states from a buffer that only holds 2 → **segmentation fault**

### Bug 2: `cudm_solver.py` (Python layer)

**Original buggy code:**

```python
split_states = bindings.splitBatchedState(state)
for i in range(batch_size):  # Assumes len(split_states) == batch_size
    intermediate_states[i].append(split_states[i])  # IndexError in distributed mode
```

**Problem:** In distributed mode, `splitBatchedState` only returns the local subset of states, not all `batch_size` states. Indexing with `batch_size` causes an `IndexError`.


python/tests/parallel/test_mpi_dynamics.py
```diff
@@ -65,6 +67,268 @@ def testMpiRun():
         store_intermediate_results=cudaq.IntermediateResultSave.NONE,
         integrator=RungeKuttaIntegrator())
 
+    # Verify result exists
+    assert evolution_result is not None
+
+
+@skipIfUnsupported
+def testMpiBatchedStatesStoreAll():
```
schweitzpgi, line 75: Do all of these tests fail if run against the current tip of `main`?

schweitzpgi:
<s>This looks like it should be merged into the python rewrite branch to me.</s>

Good to go!

python/tests/parallel/test_mpi_dynamics.py
```diff
@@ -15,8 +15,9 @@
     reason="dynamics backend not available or not a multi-GPU machine")
 
 
-@pytest.fixture(autouse=True)
-def do_something():
+@pytest.fixture(scope="module", autouse=True)
```
sacpis, line 18: Can we keep the scope as function which is the default one so that we can use this setup for each test?

sacpis (approved):
LGTM. Thanks @huaweil-nv. May be @1tnguyen can take another look at it.

1tnguyen:
> /ok to test 1c11fa1de758e0b351a40280453a74dde28df649

python/tests/parallel/test_mpi_dynamics.py
huaweil-nv: Yes, the failure modes are:
1. Segmentation fault (most common) - caused by out-of-bounds memory access in splitBatchedState
2. Incorrect state shapes - states returned with wrong dimensions (e.g., (1,) instead of (2,))
3. IndexError in Python layer - when cudm_solver.py tries to store results at wrong indices

python/tests/parallel/test_mpi_dynamics.py
huaweil-nv: We changed to scope="module" because MPI standard requires that MPI_Init is called only once and MPI_Finalize cannot be followed by another MPI_Init.
With the original scope="function", the first test would call initialize()/finalize(), and subsequent tests would fail when trying to initialize() again after finalization.
 Since the original file had only one test, this wasn't an issue before. But with multiple tests added, we need module-level scope.

1tnguyen:
> /ok to test 2cc2d70f79e7a66da44b31060e8ac08b985b0805

Issue #3757, opened 2026-01-15
Add AddressSanitizer (ASAN) CI Testing


taalexander:
### Required prerequisites

- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

As noted in #3748 and #3711 we have caught quite a few latent bugs (Also #3752 and #3755) in development of MacOS support that would likely have been caught at development time with address sanitizer builds (or maybe static analysis). This would enable catching many memory errors like buffer overflows, use-after-free, and memory leaks early in development.

The current CI pipeline (`ci.yml`, `test_in_devenv.yml`) builds and tests CUDA-Q with standard Debug/Release configurations using clang16, gcc11, and gcc12 toolchains. However, there is no sanitizer-enabled build configuration.

I propose:
1. we add ASAN as a cmake option
2. We add a new CI job for this

As ASAN builds can have a relatively large slowdown I hope we can do it in such a way that they are pipelined early and don't add to the total wall-clock build time. We might want to limit to only one or two compilers/platform targets to avoid duplicating the consumption of too many CI resources.

lmondada:
#3772 added support for sanitised builds in CMake. I suggest we proceed as follows:
1. fix memory leaks in the test suites
2. once we have all tests passing, we'll have a better estimate of CI execution time. We can then decide how to integrate ASan in CI, i.e. which settings to use (ASan vs UBSan, ...), which tests to run on all PRs and which to run nightly, etc

PR #3772, opened 2026-01-21
[build] Optional sanitizers build flags in build script

lmondada:
This PR adds support for an `-s` flag in the `scripts/build_cudaq.sh` script that activates compiler options that check for memory leaks, use-after-frees etc.

Specifically, it activates:
```
  # -fsanitize=address: Detects use-after-free, buffer overflows, stack overflows
  # -fsanitize=undefined: Detects undefined behavior (null ptr deref, signed overflow, etc.)
  # -fno-sanitize=vptr: Disable vptr sanitizer (requires RTTI, but our LLVM is built without RTTI)
  # -fno-omit-frame-pointer: Preserves frame pointers for better stack traces
  # -fno-optimize-sibling-calls: Disables tail call optimization for accurate stack traces
  # -fsanitize-address-use-after-scope: Detects use-after-scope bugs
```

In the process of testing this, I have tracked down and fixed a couple of issues that these options were flagging at compile time (see the individual commits).

More are brought up when these sanitized builds are used to run the test suite. These will have to be resolved before sanitized builds can be run in CI.

There is a memory leak that I was unable to track down and that was causing the quake tool (`tools/cudaq-quake/cudaq-quake.cpp`) to fail. This is also run at CUDAQ compile time, so I had to suppress the error in order to successfully build cudaq.

### DIFF
--- a/tools/cudaq-quake/cudaq-quake.cpp
+++ b/tools/cudaq-quake/cudaq-quake.cpp
@@ -40,7 +40,8 @@
 
 // LeakSanitizer suppression, currently necessary to successfully compile with
 // ASan.
-// TODO: find cause of leak and fix it.
+// TODO: find cause of leak and fix it (see
+// https://github.com/NVIDIA/cuda-quantum/issues/3780).
 #if defined(__SANITIZE_ADDRESS__) ||                                           \
     defined(__has_feature) && __has_feature(address_sanitizer)
 extern "C" const char *__lsan_default_suppressions() {


schweitzpgi:
> There is a memory leak that I was unable to track down and that was causing the quake tool (tools/cudaq-quake/cudaq-quake.cpp) to fail. This is also run at CUDAQ compile time, so I had to suppress the error in order to successfully build cudaq.

Please file an issue to track this with all the details on how to repo, etc., if you haven't done so already. Thanx, Luca!

taalexander:
Awesome :tada:. I think if we add a CI job for this it will close #3757 ?

scripts/build_cudaq.sh
```diff
@@ -193,6 +197,36 @@ else
...
+  fi
+  SANITIZER_LINK_FLAGS="-fsanitize=address,undefined"
+  SANITIZER_FLAGS="\
+    -DCMAKE_C_FLAGS_INIT='$SANITIZER_COMPILE_FLAGS' \
+    -DCMAKE_CXX_FLAGS_INIT='$SANITIZER_COMPILE_FLAGS' \
+    -DCMAKE_EXE_LINKER_FLAGS_INIT='$SANITIZER_LINK_FLAGS' \
+    -DCMAKE_SHARED_LINKER_FLAGS_INIT='$SANITIZER_LINK_FLAGS' \
+    -DCMAKE_MODULE_LINKER_FLAGS_INIT='$SANITIZER_LINK_FLAGS'"
+  echo "  Sanitizer compile flags: $SANITIZER_COMPILE_FLAGS"
+  echo "  Sanitizer link flags: $SANITIZER_LINK_FLAGS"
```
taalexander: I wonder if it is more accessible if we enable this as a Cmake option (which could still be triggered via build_cudaq.sh). 

My take would be that we should minimize what gets added to build_cudaq.sh as this could lead to it ultimately not being possible to build without it (eg. with pure Cmake) if we continue to add complexity to it. I am new to the project though and keen to hear others opinions.

schweitzpgi: I, for one, would welcome moving configuration settings into the cmake files (it's what cmake is for) and reducing scripts to as small a nub as possible.

lmondada: you are right, will do!

lmondada:
>Awesome 🎉. I think if we add a CI job for this it will close https://github.com/NVIDIA/cuda-quantum/issues/3757 ?

Yes, but I think we ought to do two things first:
- most of the value from an ASan build comes from using it at runtime. I suspect the proportion of bugs that get flagged at compile time is tiny in comparison. We need to start running tests with ASan enabled and either fix them or track the bugs and add temporary exceptions for them.
- speed up CI. CI is way too slow already, I am not keen on adding a very slow build process on top of that. That's not even talking of running the tests... Google tells me ASan slows down runtime by at least a factor 2x

lmondada:
> Please file an issue to track this with all the details on how to repo, etc., if you haven't done so already. Thanx, Luca!

Done! https://github.com/NVIDIA/cuda-quantum/issues/3780

taalexander:
I agree it may be too slow to include now. IREE includes some documentation [here](https://iree.dev/developers/debugging/sanitizers/). In their build architecture they separate the compiler/runtime builds but mention the ASAN (not TSAN) builds are worth it such that [they run everytime](https://iree.dev/developers/general/github-actions/#other-workflows) whereas others are scheduled. 

I've been working on a CI comparison for my macos work to better understand the differences. I'll share this at somepoint today.

taalexander (approved):
LGTM. Thank you for addressing comments.

lmondada:
next up: #3782

PR #3773, opened 2026-01-21
Reducing cut-and-paste on the backend unit tests with cmake function

Renaud-K:
The plan is to remove `setTargetBackend` from the platform public interface to enable trait based development. This will require to split the backend tests for each target, because the target will have to be defined at compile time. Typically, there are only 2 targets: an emulation target and a mock server. 

As a first step, this change is meant to reduce the amount of cut-and-paste in cmake to make it easier to split the tests.  
A function is introduced to create the test executable.

In a separate PR, it is expected that the same function will be called on the split tests.

### DIFF
--- a/unittests/backends/anyon/AnyonTester.cpp
+++ b/unittests/backends/anyon/AnyonTester.cpp
@@ -12,7 +12,6 @@
 #include <fstream>
 #include <gtest/gtest.h>
 #include <regex>
-#include <filesystem>
 
 // port number and localhost connect to mock_qpu backend server within the
 // container (mock_qpu/anyon).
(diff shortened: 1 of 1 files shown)


Renaud-K:
> /ok to test 646a866cb1250a2ef7847461e950470ddeed247c

Renaud-K:
> /ok to test 6d9416e0c6a4e690ee9c6ea86d4adb7cb9d9b57c

lmondada:
I'm taking the liberty to merge this in already, before CI re-runs are required

schweitzpgi:
These changes broke cmake. I'll file an issue.

```console
-- Registering CustomPassPlugin as a pass plugin (static build: OFF)
CMake Error at unittests/backends/CMakeLists.txt:22 (define_property):
  define_property not given a BRIEF_DOCS <brief-doc> argument.


```

PR #3775, opened 2026-01-21
Parameter fix for `exp_pauli` decomposition

khalatepradnya:
This PR updates the decomposition of `exp_pauli` instruction by correcting the angle passed to `rz` gate, aligning with the mathematical identity `Rz(theta) = exp(-i*theta/2 Z)`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,7 @@ option(CUDAQ_SKIP_MPI "Do not build with MPI, even when it is available." OFF)
 option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Frontend." OFF)
 option(CUDAQ_ENABLE_CC "Enable CUDA-Q code coverage generation." OFF)
 option(CUDAQ_REQUIRE_OPENMP "Fail the build if OpenMP is not found." OFF)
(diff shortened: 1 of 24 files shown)


python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -133,11 +133,14 @@ def ansatz(theta: float):
         0) * spin.y(1) + .21829 * spin.z(0) - 6.125 * spin.z(1)
 
     # Run the observe task on quantinuum synchronously
-    res = cudaq.observe(ansatz, hamiltonian, .59, shots_count=100000)
+    res = cudaq.observe(ansatz, hamiltonian, .59 * -0.5, shots_count=100000)
```
1tnguyen, line 136: Just a question: Is it intentional that we change the runtime parameter (0.59 -> -0.5 * 0.59) in one test and change the kernel (`exp_pauli(theta / -2.0, ...`) in the other test?

khalatepradnya: yes, I had tried `.59 * 0.5` in one test and then your suggestion of `-theta / 2` in the other test. Both are equivalent, so thought of keeping it to showcase how the argument flows. I can make it uniform for consistency. :)

PR #3777, opened 2026-01-22
Introduce a new attribute to track qubits generated by compiler passes

khalatepradnya:


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -72,7 +72,8 @@ def quake_AllocaOp : QuakeOp<"alloca", [MemoryEffects<[MemAlloc, MemWrite]>]> {
   }];
 
   let arguments = (ins
-    Optional<AnySignlessInteger>:$size
+    Optional<AnySignlessInteger>:$size,
+    UnitAttr:$compiler_generated
```
schweitzpgi, line 76:
```suggestion
    OptionalAttr<UnitAttr>:$compiler_generated
```

schweitzpgi: If you add this here, you will have to add it to the syntax, add it to the roundtrip test, etc.

Note: you can add attributes to Ops _without_ adding them formally. Every Op has an generic attribute dictionary.

Also note: this is probably of really limited use. The compiler can and will fuse `quake.alloca` Ops, which will almost certainly make this annotation highly suspect at best if not lost altogether.

Can you explain the actual application here? Is this to paper over the issue of throwing away named measurement registers?

lib/Optimizer/Transforms/MultiControlDecomposition.cpp
```diff
@@ -99,8 +99,11 @@ ArrayRef<Value> Decomposer::getAncillas(Location loc, std::size_t numAncillas) {
   OpBuilder::InsertionGuard g(builder);
   builder.setInsertionPointToStart(entryBlock);
   // If we don't have enough ancillas, allocate some more.
-  for (size_t i = allocatedAncillas.size(); i < numAncillas; ++i)
-    allocatedAncillas.push_back(builder.create<quake::AllocaOp>(loc));
+  for (size_t i = allocatedAncillas.size(); i < numAncillas; ++i) {
+    auto alloc = builder.create<quake::AllocaOp>(loc);
+    alloc.setCompilerGeneratedAttr(builder.getUnitAttr());
+    allocatedAncillas.push_back(alloc);
```
schweitzpgi, line 105: Yeah, if you add it formally, we'll need the whole 9 yards. Including a builder, etc.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
khalatepradnya: Thank you for the feedback! I was seeking direction check on this... The intention was to track compiler generated qubits and eventually filer them out. 
Currently if a kernel does not use named sub-registers, the final result may have compiler generated ancilla qubits, for example, see [this](https://github.com/NVIDIA/cuda-quantum/blob/main/unittests/backends/quantinuum/QuantinuumTester.cpp#L298).

khalatepradnya:
This is too outdated.

PR #3779, opened 2026-01-22
Reduce data type conversion overhead in Torch integrator implementation 

1tnguyen:
### Description

In the RHS compute implementation of Torch integrator, we use `dlpack` to convert different representations of the underlying device data pointer.

This conversion is relatively quick, ~50us, but this RHS computation is on the hot path, e.g., a Dopri8 integrator might need 14 interpolation stages per integration step. This overhead is more noticeable on smaller problem sizes, where the actual compute time of cudensitymat is small.

This PR removes the `dlpack` conversion and replaces it with a thin `CuDensityMatState` wrapping mechanism for unowned device pointers. In this case, `CuDensityMatState` won't clean up the device pointer.

Note: this wrap/unwrap action is contained within the `compute_rhs` function; hence, no dangling reference is possible (we don't use the wrapped `CuDensityMatState` outside the function scope).

Other minor fixes:

- Move the `test_density_matrix_indexing` test into a common test case so that we can test it for all integrator types.

- Fix an issue in the `dynamics` target config, causing it to report wrong precision setting.

### DIFF
--- a/python/tests/kernel/test_state_kernel.py
+++ b/python/tests/kernel/test_state_kernel.py
@@ -62,7 +62,8 @@ def test_state_vector_simple():
         # if not np.isclose(got_vector[i], got_vector_b[i]):
         print(f"want = {want_state[i]}")
         print(f"got = {got_vector[i]}")
-    assert np.allclose(want_state, np.array(got_state))
+    assert np.allclose(np.array(cudaq.StateMemoryView(want_state)),
+                       np.array(got_state))
     cudaq.reset_target()
 
 


python/tests/dynamics/test_evolve_dynamics.py
```diff
@@ -154,85 +154,13 @@ def test_batching_bugs():
         assert len(evolution_result.intermediate_states()) == len(steps)
 
 
-def test_density_matrix_indexing():
```
1tnguyen, line 157: **Note**: I moved this test to `system_models.py` so that we can test all integrators.

Issue #3784, opened 2026-01-22
Python run code caching performance tweak caused sigsegv failures


schweitzpgi:
Had to revert #3754 it for now.

atgeller:
Seems like this was successfully fixed by #3953

PR #3786, opened 2026-01-22
Reintroduce `run` perf cache + fix for shutdown segfault

1tnguyen:
### Description

Segfault signature is pointing to the python garbage collector deleting a `PyPassManager`.

Resolved https://github.com/NVIDIA/cuda-quantum/issues/3784


1tnguyen:
Need to investigate further. There seems to be multiple failure modes.

Issue #3785, opened 2026-01-22
Broken cmake


schweitzpgi:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

PR #3773 was recently merged and breaks cmake from the command line. The following is the log excerpt.

```console
-- Registering CustomPassPlugin as a pass plugin (static build: OFF)
CMake Error at unittests/backends/CMakeLists.txt:22 (define_property):
  define_property not given a BRIEF_DOCS <brief-doc> argument.


```

### Steps to reproduce the bug

```console
cmake ../cuda-quantum/ -DCMAKE_INSTALL_PREFIX=<dir> -DCMAKE_BUILD_TYPE=Debug -DLLVM_DIR=${CLANG_HERE}/lib/cmake/llvm -DMLIR_DIR=${CLANG_HERE}/lib/cmake/mlir -DLLVM_EXTERNAL_LIT=$HOME/cuda-quantum/tpls/llvm/build/bin/llvm-lit -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DZLIB_USE_STATIC_LIBS=FALSE -DBLAS_LIBRARIES=<dir>/libblas.a -DCUDAQ_ENABLE_PYTHON=On -DCUDAQ_ENABLE_STATIC_LINKING=OFF -DOPENSSL_ROOT_DIR=<dir> -DZLIB_ROOT=<dir> -DCUDAQ_ENABLE_BRAKET_BACKEND=OFF
```

### Expected behavior

cmake should not fail

### Is this a regression? If it is, put the last known working version (or commit) here.

8e3ac696bc086d27a8f3cc158a45cc8966e4f1b4

### Environment

- **CUDA-Q version**: tip
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

An incorrect change that works around the problem.
```console
[...]

PR #3791, opened 2026-01-23
Python/C++ interoperability tools

schweitzpgi:
This replaces the old interop interface with one that hides more of the implementation details about the compiler and runtime libraries from the user.

This is not quite as dynamic as native Python as kernel decorators may be resolved in a python interpreter context *before* they are used. This allows kernel decorator code to be pickled before use by C++ kernel code, which in turn allows the C++ code to largely be unaware that the kernel it is calling isn't just another C++ kernel. The one caveat to this is that a qkernel object holding a pointer to the entry point function of a kernel decorator must be annotated so that the runtime layer can distinguish it and know it need not try to find a host-side entry point in the C++ runtime dictionary.

### DIFF
--- a/python/tests/interop/quantum_lib/quantum_lib.h
+++ b/python/tests/interop/quantum_lib/quantum_lib.h
@@ -22,11 +22,10 @@ void another(cudaq::qview<> qubits, std::size_t);
 
 void uccsd(cudaq::qview<> qubits, std::size_t);
 
-__qpu__ void sit_and_spin_test(qkernel<void()> &&);
-__qpu__ void marshal_test(qkernel<void(std::size_t)> &&, std::size_t);
-__qpu__ void plug_and_chug_test(qkernel<void()> &&);
-__qpu__ void brain_bend_test(qkernel<void(cudaq::qvector<> &)> &&);
-__qpu__ void
-most_curious_test(qkernel<void(cudaq::qvector<> &, std::size_t)> &&);
+void sit_and_spin_test(qkernel<void()> &&);
+void marshal_test(qkernel<void(std::size_t)> &&, std::size_t);
+void plug_and_chug_test(qkernel<void()> &&);
+void brain_bend_test(qkernel<void(cudaq::qvector<> &)> &&);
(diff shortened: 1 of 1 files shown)


unittests/backends/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ set(default_backend_unittest_libs
   cudaq-platform-default 
   gtest_main)
 
-define_property(DIRECTORY PROPERTY BACKEND_UNITTEST_LIBS INHERITED)
+define_property(DIRECTORY PROPERTY BACKEND_UNITTEST_LIBS INHERITED BRIEF_DOCS "whatever" FULL_DOCS "man")
```
schweitzpgi, line 22: See #3785

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -902,6 +901,31 @@ py::object cudaq::marshal_and_launch_module(const std::string &name,
                               reinterpret_cast<char *>(args.getArgs().back()));
 }
 
+// NB: `cachedEngine` is actually of type `mlir:;ExecutionEngine**`.
```
1tnguyen: nit: typo
```suggestion
// NB: `cachedEngine` is actually of type `mlir::ExecutionEngine**`.
```

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -653,6 +653,25 @@ def __call__(self, *args):
             self.uniqName, specialized_module, mlirTy, *processedArgs)
         return result
 
+    def beta_reduction(self, execEngine, *args):
+        """
+        Perform beta reduction on this kernel decorator in the current calling
+        context. We are primary concerned with resolving the lambda lifted
+        arguments, but the formal arguments may be supplied as well.
```
1tnguyen, line 660: Question: Does "formal arguments _may be_ supplied as well" mean that we can pass an empty `args` list here even for kernels that take some arguments?

schweitzpgi: That's the eventual intention, but I haven't got that part working yet.

PR #3794, opened 2026-01-26
Handling distributed batched results for dynamics backend

1tnguyen:
### Description

This is the corresponding C++ changes of https://github.com/NVIDIA/cuda-quantum/pull/3771 (refer to this PR for more information). An equivalent C++ test suite is added as well.

Some code refactoring:

- Move some MPI helper from `CuDensityMatUtils.cpp` to `CuDensityMatContext.cpp` (adding a couple more helpers).

- Add a validation check for (`batch_size % num_ranks`) + update the documentation. Add a test case for this in both C++ and Python.

- Fix an issue with the state-to-array usage since the Python update: need to convert the state to a view.

### DIFF
--- a/docker/build/devdeps.manylinux.Dockerfile
+++ b/docker/build/devdeps.manylinux.Dockerfile
@@ -48,7 +48,8 @@ ENV LLVM_INSTALL_PREFIX=/usr/local/llvm
 RUN if [ "${toolchain#gcc}" != "$toolchain" ]; then \
         gcc_version=`echo $toolchain | grep -o '[0-9]*'` && \
         if [ -z "$(which gcc 2> /dev/null | grep $gcc_version)" ]; then \
-            dnf install -y --nobest --setopt=install_weak_deps=False gcc-toolset-$gcc_version && \
+            # Using releasever=8.9: boost packages missing from 8.10 mirrors for aarch64
+            dnf install -y --nobest --setopt=install_weak_deps=False --releasever=8.9 gcc-toolset-$gcc_version && \
             enable_script=`find / -path '*gcc*' -path '*'$gcc_version'*' -name enable` && . "$enable_script"; \
         fi && \
         CC="$(which gcc)" && CXX="$(which g++)"; \
@@ -68,7 +69,8 @@ ENV CXX="$LLVM_INSTALL_PREFIX/bootstrap/cxx"
 # Build pybind11 - 
 # we should be able to use the same pybind version independent on what Python version we generate bindings for.
 ENV PYBIND11_INSTALL_PREFIX=/usr/local/pybind11
-RUN dnf install -y --nobest --setopt=install_weak_deps=False \
+# Using releasever=8.9: cmake packages missing from 8.10 mirrors for aarch64
+RUN dnf install -y --nobest --setopt=install_weak_deps=False --releasever=8.9\
         ninja-build cmake python3-devel \
     && mkdir /pybind11-project && cd /pybind11-project && git init \
     && git remote add origin https://github.com/pybind/pybind11 \
@@ -101,7 +103,8 @@ ENV CUDA_VERSION=${cuda_version}
 RUN arch_folder=$([ "$(uname -m)" == "aarch64" ] && echo sbsa || echo x86_64) \
     && dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/$distro/$arch_folder/cuda-$distro.repo \
     && dnf clean expire-cache \
-    && dnf install -y --nobest --setopt=install_weak_deps=False wget \
+    # Using releasever=8.9: cmake packages missing from 8.10 mirrors for aarch64
+    && dnf install -y --nobest --setopt=install_weak_deps=False --releasever=8.9 wget \
         cuda-compiler-$(echo ${CUDA_VERSION} | tr . -) \
         cuda-cudart-devel-$(echo ${CUDA_VERSION} | tr . -) \
         libcublas-devel-$(echo ${CUDA_VERSION} | tr . -) \


runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -354,6 +366,17 @@ evolveBatchedImpl(const std::vector<int64_t> dims, const schedule &schedule,
     expectations.emplace_back(CuDensityMatExpectation(handle, cudmObsOp));
   }
 
+  // Helper to compute the state idx within the batch for distributed mode
+  const auto getDistributedGlobalIdx = [](int localIdx, int batchSize) {
+    // MPI rank info is static during the program execution
+    static const auto mpiNumRanks =
+        dynamics::Context::getCurrentContext()->getNumRanks();
+    static const auto mpiRank =
+        dynamics::Context::getCurrentContext()->getRank();
```
huaweil-nv: Nit: This may be totally fine in practice if MPI is always initialized before this code path is hit and the context never changes. That said, I don’t see a strong benefit to caching rank/num_ranks as function-local static (the lookup is cheap), and the static values can become stale if the first call happens before MPI is initialized or if the context is recreated. I’d prefer querying getNumRanks()/getRank() each time (or passing them in) to avoid subtle timing/lifetime issues.

python/tests/parallel/test_mpi_dynamics.py
```diff
@@ -322,14 +322,54 @@ def testMpiTwoQubitBatched():
...
 
+@skipIfUnsupported
+def testMpiBatchedStatesInvalidBatchSize():
+    """
+    Test invalid batch size for distributed batched evolution. This should raise a runtime error when the batch size
+    is not evenly divisible by the number of MPI ranks.
+    """
+    import cupy as cp
+
+    rank = cudaq.mpi.rank()
```
sacpis: nit: seems like we can remove this line as rank is unused in the test.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -374,9 +397,24 @@ evolveBatchedImpl(const std::vector<int64_t> dims, const schedule &schedule,
...
+        assert(states.size() <= batchSize);
+
+        const auto numLocalStates = states.size();
+        if (numLocalStates == batchSize) {
+          // Non-distributed mode: all states are local
+          for (int i = 0; i < batchSize; ++i) {
+            intermediateStates[i].emplace_back(states[i]);
+          }
+        } else {
+          for (int i = 0; i < states.size(); ++i) {
```
sacpis: nit: can use `numLocalStates`.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -387,7 +425,12 @@ evolveBatchedImpl(const std::vector<int64_t> dims, const schedule &schedule,
 
   if (storeIntermediateResults == cudaq::IntermediateResultSave::All) {
     std::vector<evolve_result> results;
+    // Note: In distributed mode, each rank only has local states.
+    // Hence, only return results only for the states this rank holds.
```
sacpis: nit: `Hence, return results only for the states this rank holds.`

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -397,19 +440,36 @@ evolveBatchedImpl(const std::vector<int64_t> dims, const schedule &schedule,
...
+
+      if (numLocalStates == batchSize) {
+        // Non-distributed mode: all states are local
+        for (int i = 0; i < batchSize; ++i) {
+          results.emplace_back(
+              evolve_result({cudaq::state(states[i])}, expectationVals[i]));
+        }
+      } else {
+        // Distributed mode: each rank contains only a subset of batch data.
+        for (int i = 0; i < states.size(); ++i) {
```
sacpis: nit: can use `numLocalStates`.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
```diff
@@ -420,10 +480,22 @@ evolveBatchedImpl(const std::vector<int64_t> dims, const schedule &schedule,
...
+      }
+    } else {
+      // Distributed mode: each rank contains only a subset of batch data.
+      for (int i = 0; i < states.size(); ++i) {
+        const auto globalIdx = getDistributedGlobalIdx(i, batchSize);
+        if (globalIdx < batchSize) {
+          results.emplace_back(
+              evolve_result({cudaq::state(states[i])}, expVals[globalIdx]));
+        }
+      }
```
sacpis: Can we put this functionality in a helper function. Seems like it is called on line 452 as well?

unittests/mqpu/dynamics_mpi_batching_tester.cpp
```diff
@@ -0,0 +1,541 @@
...
+};
+
+::testing::Environment *const dynamics_test_env =
+    AddGlobalTestEnvironment(new TestEnvironment);
+
+TEST(DynamicsBatchingMpi, checkSaveModes) {
+  EXPECT_TRUE(cudaq::mpi::is_initialized());
+  EXPECT_GT(cudaq::mpi::num_ranks(), 1);
+
+  const auto rank = cudaq::mpi::rank();
```
sacpis: Let's remove `const auto rank = cudaq::mpi::rank();` from all tests below as rank is an unused variable.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
1tnguyen: Sounds good. I've removed `static` in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

python/tests/parallel/test_mpi_dynamics.py
1tnguyen: Removed in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
1tnguyen: Fixed in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
1tnguyen: Fixed in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
1tnguyen: Fixed in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

runtime/nvqir/cudensitymat/CuDensityMatEvolution.cpp
1tnguyen: Sure, we can. I've refactored it in 41ebb7d51facdf22beca8e5e0a94229c0375db7d.

unittests/mqpu/dynamics_mpi_batching_tester.cpp
1tnguyen: I've removed these unused variables, except in one test case, where I want to log whether test conditions are actually running or not (useful for manually running the tests outside `ctest`).

PR #2549, opened 2025-01-27
Add pre-commit

nvidia-dobri:
### Description
Adds `pre-commit` to the repository on an opt-in basis. Requires a one-time `pre-commit install` from either inside the dev container or a user setup that already has `pre-commit` installed.

PR #3796, opened 2026-01-26
Migrate static checks to precommit

### LINKED ISSUE #3760: Migrate Static Checks to pre-commit
### Required prerequisites

### Describe the feature

I had a discussion with @mitchdz about replacing the current `repo_checks.yml` GitHub Actions workflow with a [pre-commit](https://pre-commit.com/) based approach. This enables developers to run the same checks locally before pushing, reducing CI churn from formatting/linting failures while maintaining CI enforcement.

The current static checks workflow has a few issues:

- CI churn: Developers push code, wait for CI, discover formatting issues, push again
- No local validation: Developers can't easily run the exact same checks locally
- Complex workflow: Custom shell scripting for file filtering and check execution

I propose we:
- migrate the workflow to `pre-commit` (reusing existing scripts where possible by invoking from pre-commit). 
- migrate the existing static checks workflow to instead use pre-commit (single source of truth for static checks invocation)
- update developer documentation for how to use locally.

From a local development workflow perspective this would still be *opt-in* so for most nothing would change from the usage today.

taalexander:
### Description

This PR adds support for pre-commit for running CI. Closes #3760.  It is a duplicate of #2549 (unfortunately I did not know about this before implementing). 

This implementation provides a duplicate CI pipeline for A/B testing (we should remove the existing pipeline before committing). It also provides a default installation for the devcontainer so that it may easily be used offline with `pre-commit run --all-files`

New pre-commit run (1m 23s): https://github.com/NVIDIA/cuda-quantum/actions/runs/21368765205
Old static checks run (2m 10s): https://github.com/NVIDIA/cuda-quantum/actions/runs/21367779792

### DIFF
--- a/pyproject.toml.cu13
+++ b/pyproject.toml.cu13
@@ -18,19 +18,20 @@ requires-python = ">=3.11"
 license = { file="LICENSE" }
 dependencies = [
   'astpretty ~= 3.0',
-  'custatevec-cu13 ~= 1.10',
-  'cutensornet-cu13 ~= 2.9',
-  'cudensitymat-cu13 ~= 0.3',  
   'numpy >= 1.24',
   'scipy >= 1.10.1',
   'requests >= 2.31',
-  'nvidia-cublas ~= 13.0',
-  'nvidia-curand ~= 10.4',
-  'nvidia-cusparse ~= 12.6',
-  'nvidia-cuda-runtime ~= 13.0',
-  'nvidia-cusolver ~= 12.0',
-  'nvidia-cuda-nvrtc ~= 13.0',
-  'cupy-cuda13x ~= 13.6.0'
+  # CUDA dependencies - excluded on macOS (CPU-only support)
+  'custatevec-cu13 ~= 1.10; sys_platform != "darwin"',
+  'cutensornet-cu13 ~= 2.9; sys_platform != "darwin"',
+  'cudensitymat-cu13 ~= 0.3; sys_platform != "darwin"',
+  'nvidia-cublas ~= 13.0; sys_platform != "darwin"',
+  'nvidia-curand ~= 10.4; sys_platform != "darwin"',
+  'nvidia-cusparse ~= 12.6; sys_platform != "darwin"',
+  'nvidia-cuda-runtime ~= 13.0; sys_platform != "darwin"',
+  'nvidia-cusolver ~= 12.0; sys_platform != "darwin"',
+  'nvidia-cuda-nvrtc ~= 13.0; sys_platform != "darwin"',
+  'cupy-cuda13x ~= 13.6.0; sys_platform != "darwin"',
 ]
 classifiers = [
     'Intended Audience :: Science/Research',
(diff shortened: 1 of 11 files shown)


Developing.md
```diff
@@ -78,28 +78,161 @@ comments as well as a comment at the top of the file to indicating its purpose.
...
+locally that run in CI, catching issues before you push.
+
+**Benefits:**
+
+- Catch formatting and linting issues locally before CI runs
+- Same checks run locally and in CI (guaranteed consistency)
+- Automatic formatting fixes where possible
+- Fast incremental checks (only checks changed files)
+
+#### Prerequisites
```
mitchdz, line 94: For this section, maybe combine each platform into one install step? e.g.

for macos
```
# aspell for spell checking
# node for link validation
# go for license header validation
brew install aspell node go
```

mitchdz (approved):
Looks great to me. The above comment is more of a nit. Thanks @taalexander

Developing.md
taalexander: Added.

taalexander:
Thanks for the review @mitchdz. I've addressed and also replaced the existing static check CI job with the new one now. Should be good to merge pending CI passes.

.github/pre-commit/spellcheck_config.yml
```diff
@@ -1,9 +1,16 @@
+# Pyspelling configuration for CUDA-Q
+# Note: Source exclusion patterns (e.g., !tpls/**) match the exclusions
```
bettinaheim, line 2: I am assuming you are defining these exclusions here because that makes it easier to run spellchecks locally? 
For the pipelines the pattern so far is to have the pipeline filter out folders rather than the config, but I can go either way.

.github/workflows/config/spelling_allowlist.txt
```diff
@@ -149,6 +150,7 @@ ansatz
 ansatzes
 archiver
 arity
+aspell
```
bettinaheim: The rule we want to follow for what goes into the spelling allow list is: "If you can find it in a (US Eng.) dictionary, it's ok to add it here". If it is not a valid word then likely 
- it is coding related (command, package name, code, ...), in which case you should use backticks (`) to mark it as such, which will exclude it from spellcheck (and if it doesn't check the spellcheck config regex), or
- it is part of a pattern (e.g. link, filename, ...) that is not spoken language, in which case you should update the config to filter that pattern.

I am a bit confused why you needed to add so many command related non-English words here for this PR. It looks like maybe the code blocks in the markdown files aren't properly filtered. Can you update the filters with a suitable regex pattern instead?

taalexander: I added the readme in this PR and I think it is grabbing these for some reason. I'll update it to reflect your intent.

.github/pre-commit/spellcheck_config.yml
taalexander: Yes, now the exclusions are available locally (this way the usage is just `pre-commit ...`). I can move these files to a more central location so they are outside of .github.

taalexander: I've now moved all pre-commit/static check configurations to `.github/precommit` so that they reflect the intention of code owners.

.github/workflows/config/spelling_allowlist.txt
taalexander: I've now updated the spelling list to reflect this intent and fixed up everything else.

.github/pre-commit/spellcheck_config.yml
bettinaheim: Thanks!

bettinaheim (approved):

PR #3799, opened 2026-01-26
pin AlmaLinux mirrors to 8.9

mitchdz:
This helps alleviate an issue where boost libraries were not found for the arm64 images, causing failures when building the wheels[0].

This should ideally not be permanent, and Almalinux should fix the mirrors.

[0] - https://github.com/NVIDIA/cuda-quantum/actions/runs/21368765817/job/61512365389#step:10:595

### DIFF
--- a/docker/build/devdeps.manylinux.Dockerfile
+++ b/docker/build/devdeps.manylinux.Dockerfile
@@ -69,7 +69,8 @@ ENV CXX="$LLVM_INSTALL_PREFIX/bootstrap/cxx"
 # Build pybind11 - 
 # we should be able to use the same pybind version independent on what Python version we generate bindings for.
 ENV PYBIND11_INSTALL_PREFIX=/usr/local/pybind11
-RUN dnf install -y --nobest --setopt=install_weak_deps=False \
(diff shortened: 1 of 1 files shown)


docker/build/devdeps.manylinux.Dockerfile
```diff
@@ -68,7 +69,8 @@ ENV CXX="$LLVM_INSTALL_PREFIX/bootstrap/cxx"
 # Build pybind11 - 
 # we should be able to use the same pybind version independent on what Python version we generate bindings for.
 ENV PYBIND11_INSTALL_PREFIX=/usr/local/pybind11
-RUN dnf install -y --nobest --setopt=install_weak_deps=False \
+# Using releasever=8.9: cmake packages missing from 8.10 mirrors for aarch64
+RUN dnf install -y --nobest --setopt=install_weak_deps=False --releasever=8.9\
```
sacpis, line 73: nit: Do we need a space here `--releasever=8.9 \`?

mitchdz: because the lines below have spaces, no - but it probably should just for consistency. I say we ignore it so we don't wait a whole another CI run :)

PR #3800, opened 2026-01-27
[C++] Type distinction for measurement result

khalatepradnya:
This PR introduces comprehensive support for the `measure_result` type throughout the CUDA-Q C++ frontend, enabling deferred discrimination of measurement results. 

- New standalone `cudaq::measure_result` class with `std::int64_t value` and `std::int64_t unique_id` fields (16 bytes).
- Kernels with `measure_result` in their signature are classified as device-only.
- `quake.discriminate` is inserted only at points where a classical value is needed: if/ternary conditions, bool/int/double casts, comparisons, arithmetic, explicit `to_bool_vector()` / `to_integer()` calls and return from entry-point kernels.

These changes collectively ensure that quantum measurement results are treated as first-class citizens in the type system and IR lowering.

#### Breaking Changes
- `auto kernel() { return mz(q); }` now returns `measure_result`, making the kernel device-only.
- `std::vector<bool> v = mz(qubits)` no longer compiles; use `cudaq::to_bool_vector(mz(qubits))`.
- `int x = mz(q)` requires `static_cast<int>(mz(q))`; `double x = mz(q)` requires `static_cast<double>(mz(q))`.
- Entry-point kernels cannot return or accept as parameters `measure_result` or `std::vector<measure_result>`.
- `cudaq::measure_result` has no default constructor; every instance must come from `mz`/`mx`/`my`.
- `cudaq::measure_result` is not assignable (copy and move `operator=` are deleted).

*Note*: Python front-end changes will be covered in a separate PR.


test/AST-Quake/qir_profiles.cpp
```diff
@@ -168,6 +168,7 @@ struct comprehensive {
 // ADAPT:         tail call void @__quantum__qis__z__body(%Qubit* nonnull inttoptr (i64 4 to %Qubit*))
 // ADAPT:         tail call void @__quantum__qis__z__body(%Qubit* nonnull inttoptr (i64 5 to %Qubit*))
 // ADAPT:         tail call void @__quantum__qis__z__body(%Qubit* nonnull inttoptr (i64 6 to %Qubit*))
+// ADAPT:         tail call void @__quantum__qis__cz__body(%Qubit* nonnull inttoptr (i64 2 to %Qubit*), %Qubit* nonnull inttoptr (i64 4 to %Qubit*))
```
khalatepradnya, line 178: (Note to self) Check this!

lib/Frontend/nvqpp/ASTBridge.cpp
```diff
@@ -53,7 +53,7 @@ listReachableFunctions(clang::CallGraphNode *cgn) {
 // Does `ty` refer to a Quake quantum type? This also checks custom recursive
 // types. It does not check builtin recursive types; e.g., `!llvm.ptr<T>`.
 static bool isQubitType(Type ty) {
-  if (quake::isQuakeType(ty))
+  if (quake::isQuantumType(ty))
```
schweitzpgi: I think there is a discussion to be had here.

`!quake.measure` values are semantically defined to be abstract values that live the entirety of their lives on the device. It doesn't make sense for these to "escape" into C++ code as they are pure abstractions. What does make sense is for the QPU device, which itself is an abstraction over hardware, simulators, etc., to be able to figure out on its own accord how to map "a measurement" to an integer value. An integer value is something that classical code understands and can use.

This change is, I believe, here to allow entry-point kernels to pass around `!quake.measure` arguments and results from and to classical C++ code.

We want this the opposite way. `!quake.measure` values are nonsensical in classical compute code.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -636,36 +642,70 @@ bool QuakeBridgeVisitor::VisitCastExpr(clang::CastExpr *x) {
...
     castToTy = popType();
     if (isa<IntegerType>(castToTy) && isa<IntegerType>(sub.getType())) {
       auto locSub = toLocation(x->getSubExpr());
       bool result = intToIntCast(locSub, sub);
       assert(result && "integer conversion failed");
       return result;
     }
+    auto i1Type = builder.getI1Type();
+    // Handle conversion of `measure_result`
+    if (isa<quake::MeasureType>(sub.getType())) {
```
```suggestion
    auto measTy = quake::MeasureType::get(builder.getContext());
    if (sub.getType() == measTy) {
```
more efficient.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -636,36 +642,70 @@ bool QuakeBridgeVisitor::VisitCastExpr(clang::CastExpr *x) {
...
+      if (isa<FloatType>(castToTy))
+        return pushValue(builder.create<cudaq::cc::CastOp>(
+            loc, castToTy, i1Val, cudaq::cc::CastOpMode::Unsigned));
+      // Otherwise, just return the `i1` value
+      return pushValue(i1Val);
+    }
+
+    // Handle conversion of `std::vector<measure_result>` to `std::vector<bool>`
+    if (auto vecTy = dyn_cast<cc::StdvecType>(sub.getType()))
+      if (isa<quake::MeasureType>(vecTy.getElementType()))
```
schweitzpgi: replace with an `==` test as above.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -636,36 +642,70 @@ bool QuakeBridgeVisitor::VisitCastExpr(clang::CastExpr *x) {
...
+        return pushValue(builder.create<quake::DiscriminateOp>(
+            loc, cc::StdvecType::get(i1Type), sub));
+
+    // Handle pointer to `measure_result` (from vector element access)
+    if (auto ptrTy = dyn_cast<cc::PointerType>(sub.getType()))
+      if (isa<quake::MeasureType>(ptrTy.getElementType())) {
```
schweitzpgi: and here.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1430,6 +1477,21 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
     TODO_loc(loc, "unhandled std::vector<bool> member function, " + funcName);
   }
 
+  // Handle `measure_result` member functions
+  if (isInClassInNamespace(func, "measure_result", "cudaq")) {
+    if (func->isOverloadedOperator()) {
+      auto overloadedOperator = func->getOverloadedOperator();
+      if (isAssignmentOperator(overloadedOperator)) {
```
schweitzpgi: I'm confused. Why are we storing pure abstractions to memory? That doesn't make sense to me.

runtime/cudaq/qis/measure_result.h
```diff
@@ -0,0 +1,69 @@
...
+  friend bool operator==(const measure_result &m, bool b) {
+    return static_cast<bool>(m) == b;
+  }
+  friend bool operator==(bool b, const measure_result &m) {
+    return b == static_cast<bool>(m);
+  }
+
+  friend bool operator!=(const measure_result &m1, const measure_result &m2) {
+    return static_cast<bool>(m1) != static_cast<bool>(m2);
+  }
```
sacpis, line 69: In operator== we compare both result and uniqueId. Seems like it is inconsistent with !=? So for something like measure_result(1, 0) and measure_result(1, 1) both == and != will return false.

sacpis (commented):
Overall LGTM. Thanks @khalatepradnya. Left a comment.

lib/Frontend/nvqpp/ASTBridge.cpp
khalatepradnya: I agree that `!quake.measure` values are device-side abstractions and must not escape into classical C++ code. The host should only see that converted result.
At this line we only use `quake::isQuantumType(ty)`, which intentionally excludes `MeasureType` (see `include/cudaq/Optimizer/Dialect/Quake/QuakeTypes.h: "NB: this intentionally excludes MeasureType"`). This was already present.
So `!quake.measure` is not treated as a “quantum type” for the purposes of this check. Entry-point kernels cannot take `measure_result` (or `std::vector<measure_result>`) as an argument, that is enforced separately by `hasMeasureResultArgs` below.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Not sure I understand. The host only ever sees the converted type (`cudaq::measure_result`) built from the payload received from the device.

test/AST-Quake/qir_profiles.cpp
khalatepradnya: Confirmed that this `cz` call is present in the output of `main` as well.

runtime/cudaq/qis/measure_result.h
```diff
@@ -0,0 +1,68 @@
...
+  explicit measure_result(int val) : value(val) {}
+  explicit measure_result(int val, int id) : value(val), unique_id(id) {}
+
+  // Operator overloads for conversions and comparisons
+#ifdef CUDAQ_LIBRARY_MODE
+  operator bool() const { return __nvqpp__MeasureResultBoolConversion(value); }
+#else
+  operator bool() const { return value == 1; }
+#endif
+  explicit operator int() const { return value; }
```
sacpis: Is `explicit int()` a breaking change? Seems like any implicit conversion to int will break at compile time?Should we put a comment here?

sacpis (approved):
LGTM. Thanks @khalatepradnya. Left one comment.

runtime/cudaq/qis/measure_result.h
khalatepradnya: Yes, it is. I've added in the description under "Breaking Changes",
 `int x = mz(q) requires static_cast<int>(mz(q)).`

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -99,6 +99,13 @@ maybeUnpackOperands(OpBuilder &builder, Location loc, ValueRange operands,
   return std::make_pair(targets, SmallVector<Value>{});
 }
 
+static Value emitDiscriminate(OpBuilder &builder, Location loc, Value val) {
+  auto measTy = quake::MeasureType::get(builder.getContext());
+  if (val.getType() == measTy)
+    return builder.create<quake::DiscriminateOp>(loc, builder.getI1Type(), val);
+  return val;
```
schweitzpgi: Why do we need this conditional? We shouldn't be building a `discriminate` if the argument isn't a measurement.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -636,36 +643,71 @@ bool QuakeBridgeVisitor::VisitCastExpr(clang::CastExpr *x) {
         builder.create<cudaq::cc::CastOp>(loc, castToTy, popValue(), mode));
   }
   case clang::CastKind::CK_IntegralToFloating: {
+    auto value = popValue();
+    value = emitDiscriminate(builder, loc, value);
```
schweitzpgi: I need to see the AST. This looks suspicious to me. There ought to be an explicit call to the method, not just a cast expression, no?

schweitzpgi (commented):
Let me try these changes and look at some of the input. This suggests something that doesn't seem to align with what I was expecting. But I've been surprised by the clang ASTs before. heh

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: OK, looking at the AST, I don't think these conversions are being handled in the right place.

```c++
__qpu__ bool lookatme() {
  cudaq::qubit q;
  bool thisisit = mz(q);
  return thisisit;
}
```
produces an AST (partially copied) of
```console
  | |     `-ImplicitCastExpr 0x739083027ca8 <col:19, col:23> 'bool' <UserDefinedConversion>
  | |       `-CXXMemberCallExpr 0x739083027c88 <col:19, col:23> 'bool'
  | |         `-MemberExpr 0x739083027c58 <col:19, col:23> '<bound member function type>' .operator bool 0x73908394d1c0
  | |           `-ImplicitCastExpr 0x739083027c40 <col:19, col:23> 'const cudaq::measure_result' lvalue <NoOp>
  | |             `-MaterializeTemporaryExpr 0x739083027c28 <col:19, col:23> 'measure_result':'cudaq::measure_result' lvalue
  | |               `-CallExpr 0x739083027c00 <col:19, col:23> 'measure_result':'cudaq::measure_result' adl
  | |                 |-ImplicitCastExpr 0x739083027be8 <col:19> 'measure_result (*)(qubit &)' <FunctionToPointerDecay>
  | |                 | `-DeclRefExpr 0x739083027bc8 <col:19> 'measure_result (qubit &)' lvalue Function 0x739083838f60 'mz' 'measure_result (qubit &)'
  | |                 `-DeclRefExpr 0x7390830277e0 <col:22> 'cudaq::qubit':'cudaq::qudit<2>' lvalue Var 0x739083027630 'q' 'cudaq::qubit':'cudaq::qudit<2>'
```

The `CallExpr` is lowered to the `mz` op and should return a value of `!quake.measure`. That SSA value flows as an argument to the `CXXMemberCallExpr`. We must process that `CXXMemberCallExpr` along with the prior visited `MemberExpr` to grok that this is the `operator bool()` function being called. That call will lower to a `quake.discriminate`.

We don't want to simply guess that an `ImplicitCastExpr` might have been wrapping a subtree like this along ad hoc paths in the AST traversal however. That approach will fall apart if other ASTs with similar casts but different semantics come to play in the future.

schweitzpgi (changes requested):
We're making good progress here. Need to tweak the bridge changes a bit more.

lib/Frontend/nvqpp/ASTBridge.cpp
```diff
@@ -640,6 +660,7 @@ void ASTBridgeAction::ASTBridgeConsumer::HandleTranslationUnit(
       // Flag func as a quantum kernel.
       func->setAttr(kernelAttrName, unitAttr);
       if ((!hasAnyQubitTypes(func.getFunctionType())) &&
+          (!hasAnyMeasureResultTypes(func.getFunctionType())) &&
```
schweitzpgi: Let's merge the `hasAnyQubit` and `hasAnyMeasureResult` functionality. We can rename the test to `hasAllQuakeTypes` to be clear to future readers.

A single traversal of the signature is sufficient to detect if any of the input types or result types are quake types (which must include measurement).

lib/Frontend/nvqpp/ConvertDecl.cpp
```diff
@@ -744,6 +747,11 @@ bool QuakeBridgeVisitor::VisitVarDecl(clang::VarDecl *x) {
       // and if so, find the mz and tag it with the variable name
       auto elementType = vecType.getElementType();
 
+      if (auto meas = initVec.getDefiningOp<quake::MeasurementInterface>()) {
+        meas.setRegisterName(builder.getStringAttr(x->getName()));
```
schweitzpgi, line 763: Aside: are we going to get rid of this naming by variable feature?

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: We should not need this. See comment on line 647.

lib/Frontend/nvqpp/ConvertExpr.cpp
schweitzpgi: OK. I'll explain by example.

This quake is valid.
```mlir
  %1 = quake.mz %0 : (!quake.ref) -> !quake.measure
  %2 = quake.discriminate %1 : (!quake.measure) -> i1
  %3 = cc.alloca i8
  %4 = cc.cast unsigned %2 : (i1) -> i8
  cc.store %4. %3 : !cc.ptr<i8>
```

But this quake is *invalid nonsense*.
```mlir
  %1 = quake.mz %0 : (!quake.ref) -> !quake.measure
  %3 = cc.alloca !quake.measure
  cc.store %1. %3 : !cc.ptr<!quake.measure>
```
We know this because we're using `cc.alloca`, which allocate storage for classical values, to allocate "space" for a quantum dialect abstraction.  

If this is lowering to the first case, all is good. If it is lowering to the latter, it'll need to be fixed.

Does that help?

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: I removed the call to `emitDiscriminate`.

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -77,6 +79,10 @@ Value cudaq::cc::getByteSizeOfType(OpBuilder &builder, Location loc, Type ty,
                 // we're assuming pointers are 64 bits.
                 return {8};
               })
+          .Case([](quake::MeasureType) -> std::optional<std::int32_t> {
+            // Size of `measure_result` {int value, int unique_id} = 8 bytes
+            return {8};
```
atgeller: Is this always supposed to be 8 bytes, or could it someday be more depending on what the backend stores in the measure_result?

atgeller (approved):
LGTM, thanks @khalatepradnya! Left a comment.

lib/Optimizer/Dialect/CC/CCOps.cpp
khalatepradnya: It's supposed to be 16 bytes, I've updated it.
Any additional info from backend, in the future, may be retrieved using the unique identifier.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -2124,16 +2197,43 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+                                                 cudaqConvertToInteger,
+                                                 ValueRange{arg})
+                           .getResult(0));
+    }
+
+    if (funcName == "to_bool_vector") {
+      // `args[0]` is `!cc.stdvec<!quake.measure>`
+      auto arg = args[0];
+      // Insert discriminate if needed
+      if (auto vecTy = dyn_cast<cc::StdvecType>(arg.getType())) {
```
schweitzpgi: The C++ would not be type correct if `cudaq::to_bool_vector` was passed anything other than an argument of type `std::vector<cudaq::measurement_type>`, right? So this check isn't needed. Indeed, if the argument has some other type, it is an indication that something has gone completely wrong with the bridge. We could have an assertion here or even throw an internal compiler error and stop.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -3251,6 +3351,20 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
     return pushValue(builder.create<cc::LoadOp>(loc, copyObj));
   }
 
+  // Handle `measure_result` copy and move constructors
+  if ((ctor->isCopyConstructor() || ctor->isMoveConstructor()) &&
+      isInClassInNamespace(ctor, "measure_result", "cudaq")) {
+    // The source is a pointer to measure_result (!cc.ptr<!quake.measure>)
+    // Just load and return the value
```
schweitzpgi: This comment specifies something invalid.

In the IR, we cannot have pointers to abstract quake types. We cannot load the abstraction and we cannot store it. (Think of a measurement type as some volatile stream of random data of infinite size. There is no way to allocate a finite amount of space for it, and even if you could, as soon as you tried to load or store it, the data would change and be something else.)

lib/Frontend/nvqpp/ConvertType.cpp
```diff
@@ -124,13 +124,24 @@ static bool isFunctionCallable(Type t) {
   return false;
 }
 
+static bool isMeasureResultType(Type t) { return isa<quake::MeasureType>(t); }
+
+static bool isMeasureResultSequenceType(Type t) {
```
schweitzpgi: Seems overly specific. Why "result"? Also overly granular. The uses below look like they test both `measure` and `vector<measure>`. This could be folded into a single function.

lib/Frontend/nvqpp/ConvertType.cpp
```diff
@@ -139,7 +150,8 @@ static bool isKernelResultType(Type t) {
 static bool isKernelArgumentType(Type t) {
   return isArithmeticType(t) || isComposedArithmeticType(t) ||
          quake::isQuantumReferenceType(t) || isKernelCallable(t) ||
-         isFunctionCallable(t) ||
+         isFunctionCallable(t) || isMeasureResultType(t) ||
+         isMeasureResultSequenceType(t) ||
```
schweitzpgi: So the comment above. Here, we're testing _argument_ types, but the two functions use `result` as part of the name which is misleading.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -136,6 +136,29 @@ struct QIRAPITypeConverter : public TypeConverter {
...
+    });
+    addConversion([&](cudaq::cc::ArrayType ty) {
+      auto newEleTy = convertType(ty.getElementType());
+      auto size = ty.getSize();
+      if (size)
+        return cudaq::cc::ArrayType::get(ty.getContext(), newEleTy, size);
+      return cudaq::cc::ArrayType::get(newEleTy);
+    });
+    addConversion([&](cudaq::cc::StructType ty) -> Type {
+      SmallVector<Type> members;
```
schweitzpgi, line 170: Structs may be opaque and in that case conversion isn't required.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -136,6 +136,29 @@ struct QIRAPITypeConverter : public TypeConverter {
...
+      for (auto memTy : ty.getMembers())
+        members.push_back(convertType(memTy));
+      if (ty.getName())
+        return cudaq::cc::StructType::get(ty.getContext(), ty.getName(),
+                                          members, ty.getBitSize(),
+                                          ty.getAlignment(), ty.getPacked());
+      return cudaq::cc::StructType::get(ty.getContext(), members,
+                                        ty.getBitSize(), ty.getAlignment(),
+                                        ty.getPacked());
```
schweitzpgi:
```suggestion
      return cudaq::cc::StructType::get(ty.getContext(), ty.getNameAttr(), members,
                                        /*opaque=*/false, ty.getPacked(), ty.getBitSize(), ty.getAlignment()
                                        );
```
I think this should work and used the base constructor to simplify the code.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -1709,6 +1733,24 @@ struct OpInterfacePattern : public OpConversionPattern<OP> {
   matchAndRewrite(OP op, typename Base::OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     auto newResultTy = getTypeConverter()->convertType(op.getType());
+
+    // Check if result type changed
```
schweitzpgi: What is this for? The IR here must be strongly type correct. So the types must be the same, they cannot "change".

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -1722,6 +1764,9 @@ using UndefOpPattern = OpInterfacePattern<cudaq::cc::UndefOp>;
 using PoisonOpPattern = OpInterfacePattern<cudaq::cc::PoisonOp>;
 using CastOpPattern = OpInterfacePattern<cudaq::cc::CastOp>;
 using SelectOpPattern = OpInterfacePattern<arith::SelectOp>;
+using ComputePtrOpPattern = OpInterfacePattern<cudaq::cc::ComputePtrOp>;
+using StdvecInitOpPattern = OpInterfacePattern<cudaq::cc::StdvecInitOp>;
+using StdvecDataOpPattern = OpInterfacePattern<cudaq::cc::StdvecDataOp>;
```
schweitzpgi: I think we're using the wrong idiom here in retrospect. A normal `stdvec<T>` value is a memory buffer, but a `stdvec<measure>` never was. It was always expanded and converted to `stdvec<bool>` immediately.

Again, loading and storing `quake.anytype` from a stack location makes no sense.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -1754,6 +1799,18 @@ struct StoreOpPattern : public OpConversionPattern<cudaq::cc::StoreOp> {
   }
 };
 
+struct LogOutputOpPattern : public OpConversionPattern<cudaq::cc::LogOutputOp> {
```
schweitzpgi: Can't this one be another OpInterfacePattern instance?

khalatepradnya:
- Depends on https://github.com/NVIDIA/cuda-quantum/pull/4208

lib/Frontend/nvqpp/ConvertDecl.cpp
khalatepradnya: My understanding is that yes, it will be removed in the future.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Fixed in 4aac0bf9c006e111d3b19fc4a76ace9cc31de4ee

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
khalatepradnya: Removed.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
khalatepradnya: Updated in 4aac0bf9c006e111d3b19fc4a76ace9cc31de4ee

sacpis (approved):

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -658,14 +665,45 @@ bool QuakeBridgeVisitor::VisitCastExpr(clang::CastExpr *x) {
...
+    if (isa<quake::MeasurementsType>(sub.getType()))
+      return pushValue(builder.create<quake::DiscriminateOp>(
+          loc, cc::StdvecType::get(i1Type), sub));
+
+    // Handle pointer to `measure_result` (from vector element access)
+    if (auto ptrTy = dyn_cast<cc::PointerType>(sub.getType()))
+      if (ptrTy.getElementType() == measTy) {
+        auto loaded = builder.create<cc::LoadOp>(loc, sub);
+        return pushValue(
+            builder.create<quake::DiscriminateOp>(loc, i1Type, loaded));
```
schweitzpgi: Does this case actually happen still? We can't load and store quantum values of type measure to classical memory spaces.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1520,10 +1566,41 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
   auto funcArity = func->getNumParams();
   SmallVector<Value> args = lastValues(funcArity);
   if (isa<clang::CXXMethodDecl>(func)) {
-    [[maybe_unused]] auto thisPtrValue = popValue();
+    auto thisPtrValue = popValue();
+
+    // Handle `measure_result` conversion operators
+    if (isa<clang::CXXConversionDecl>(func) &&
+        isInClassInNamespace(func, "measure_result", "cudaq")) {
+      return pushValue(thisPtrValue);
```
schweitzpgi, line 1567: Not sure what's happening here. In C++, there are objects of type `measure_result` of course. In quake, what the bridge is generating here, there are no objects at all. So the `this` pointer is meaningless.

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -3264,6 +3360,17 @@ bool QuakeBridgeVisitor::VisitCXXConstructExpr(clang::CXXConstructExpr *x) {
     return pushValue(builder.create<cc::LoadOp>(loc, copyObj));
   }
 
+  // Handle `measure_result` copy and move constructors.
+  // The source is always a `!quake.measure` SSA value
+  if ((ctor->isCopyConstructor() || ctor->isMoveConstructor()) &&
+      isInClassInNamespace(ctor, "measure_result", "cudaq")) {
+    assert(x->getNumArgs() == 1);
```
schweitzpgi, line 3362: debug code?

lib/Frontend/nvqpp/ConvertStmt.cpp
```diff
@@ -12,6 +12,7 @@
 #include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
 #include "llvm/Support/Debug.h"
 #include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinOps.h"
```
schweitzpgi: What builtin op is in the new code below?

lib/Optimizer/Builder/Marshal.cpp
```diff
@@ -790,6 +790,11 @@ std::pair<bool, func::FuncOp> cudaq::opt::marshal::lookupHostEntryPointFunc(
     // No host entry point needed.
     return {false, func::FuncOp{}};
   }
+  // Device-only kernels (those with quantum types or `measure_result` in their
+  // signature) have no host-side entry point, so skip them.
+  if (!funcOp->hasAttr(cudaq::entryPointAttrName)) {
+    return {false, func::FuncOp{}};
+  }
```
schweitzpgi: No curly braces.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: No, it is dead code, will remove.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Agreed, there's no object in quake. The comment was misleading, I've updated it.
Without this, a kernel like:
```
auto m = mz(q);
if (m) { ... } 
```
crashes since the bridge tries to emit a member call on `!quake.measure`, which has no callable.

lib/Frontend/nvqpp/ConvertExpr.cpp
khalatepradnya: Invariant guards, not debug code. There are two to verify the preconditions for safe SSA forwarding: single argument, and source is `!quake.measure`. Intent is to prevent silently creating incorrect IR. Happy to convert to diagnostic errors if you prefer.

schweitzpgi: Is it even possible to have a `measure_result` copy ctor that does not have exactly 1 argument? This test seems to be verifying a constant condition.

clang++ itself is also running here, so it is going to flag completely broken input on its own.
So something like
```c++
cudaq::measure_result mr2{mr1, "I'm horribly broken!"};
```
is not going to compile.

schweitzpgi (approved):

Issue #3806, opened 2026-01-27
[run] Incorrect results on emulated target


khalatepradnya:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following kernel works correctly with simulator targets but gives incorrect results on emulated (hardware) targets.

```
#include <cudaq.h>
#include <iostream>

struct kernel_with_conditional {
  std::vector<bool> operator()() __qpu__ {
    std::vector<bool> var(2);
    cudaq::qubit q0, q1;
    var[0] = mz(q0);
    var[1] = mz(q1);
    return var;
  }
};

int main() {
  auto results = cudaq::run(10, kernel_with_conditional{});
  std::cout << "First result: " << results[0][0] << " " << results[0][1]
            << "\n";
  return 0;
}
```

### Steps to reproduce the bug

```
# nvq++ --target qci --emulate ./issue.cpp && ./a.out
First result: 0 1
```

### Expected behavior

Expected result is `0 0`

Correct output with simulator target.

```
# nvq++ ./issue.cpp && ./a.out
First result: 0 0
```

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: `main` as of commit 6ab71f9e73b9a2920b9c83639956e2ecbe16a83e
- **Python version**: N/A
- **C++ compiler**: 
- **Operating system**: WSL2

PR #3807, opened 2026-01-27
[Bug-fix] [run] Correctly handle output logging for `std::vector<bool>` to use `i8` pointer

khalatepradnya:
This PR addresses issue #3806 

* Updated the rewriter pattern in `ReturnToOutputLog.cpp` to store `std::vector<bool>` elements as `i8` (8-bit integers) instead of `i1` (1-bit), and to properly cast back to `i1` when reading values for logging.


khalatepradnya:
Covered by https://github.com/NVIDIA/cuda-quantum/pull/3868

PR #3808, opened 2026-01-27
 Adding PTSBE data structures

sacpis:
Adding the following data structures along with unittests
	- KrausSelection
	- KrausTrajectory
	- PTSBEResult
	- ShotAllocationStrategy
	- TrajectoryMetadata

### DIFF
--- a/unittests/ptsbe/ShotAllocationTester.cpp
+++ b/unittests/ptsbe/ShotAllocationTester.cpp
@@ -163,16 +163,6 @@ CUDAQ_TEST(ShotAllocationTest, EmptyTrajectoryList) {
   EXPECT_ANY_THROW({ allocateShots(trajectories, 1000, strategy); });
 }
 
-CUDAQ_TEST(ShotAllocationTest, ZeroShotsRequested) {
-  std::vector<KrausTrajectory> trajectories = {makeTrajectory(0, 0.5),
-                                               makeTrajectory(1, 0.5)};
-
-  ShotAllocationStrategy strategy(ShotAllocationStrategy::Type::PROPORTIONAL);
-
-  // Zero shots should throw an exception
-  EXPECT_THROW(allocateShots(trajectories, 0, strategy), std::invalid_argument);
-}
-
 CUDAQ_TEST(ShotAllocationTest, SingleShotDistribution) {
   std::vector<KrausTrajectory> trajectories = {makeTrajectory(0, 0.5),
                                                makeTrajectory(1, 0.5)};


runtime/cudaq/ptsbe/KrausSelection.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace cudaq {
+
+/// @brief Type-safe index for Kraus operators in a noise channel
+/// IDENTITY (0) represents no error, indices 1+ represent actual error
+/// operators
+enum class KrausOperatorIndex : std::size_t {
+  IDENTITY = 0,
+  // Values 1+ represent actual error operators from the noise channel
+};
```
taalexander:
```suggestion
enum class KrausOperatorType : std::size_t {
  IDENTITY = 0,
  // Values 1+ represent actual error operators from the noise channel
};
```

maybe?

runtime/cudaq/ptsbe/KrausSelection.h
```diff
@@ -0,0 +1,65 @@
...
+
+/// @brief Represents the choice of a specific Kraus operator at one noise point
+struct KrausSelection {
+  /// @brief Unique position in the circuit's total ordering of noise-capable
+  /// operations Each operation that can have noise applied is assigned a
+  /// sequential position (0, 1, 2, ...)
+  std::size_t circuit_location = 0;
+
+  /// @brief Qubits affected by this noise operation (controls + targets)
+  std::vector<std::size_t> qubits;
```
taalexander, line 33: A requirement I hadn't thought about - do we support simulation of qudits today? I have seen some mention of them throughout the codebase but it seems secondary. This might be an explicit limitation (which seems reasonable to me).

runtime/cudaq/ptsbe/KrausTrajectory.h
```diff
@@ -0,0 +1,67 @@
...
+
+/// @brief Represents one complete path through the space of possible noise
+/// realizations
+struct KrausTrajectory {
+  /// @brief Unique identifier for this trajectory
+  std::size_t trajectory_id = 0;
+
+  /// @brief Complete specification of which Kraus operators to apply at each
+  /// noise point This tracks only the injected noise operators
+  std::vector<KrausSelection> kraus_selections;
```
taalexander, line 39: Can we assume these to be ordered by circuit index?

runtime/cudaq/ptsbe/KrausTrajectory.h
```diff
@@ -0,0 +1,67 @@
...
+
+  /// @brief Computed probability of this trajectory occurring
+  /// This is the product of individual Kraus operator probabilities
+  double probability = 0.0;
+
+  /// @brief Number of measurement shots allocated to this trajectory
+  std::size_t num_shots = 0;
```
taalexander, line 46: Does this conflate the trajectory sampling step with the shot sampling step? Eg., I think we first build up our trajectories and there respective probabilities before sampling a number of shots from them.

Maybe we expect to build up the trajectory in a factory like manner, step by step?

runtime/cudaq/ptsbe/KrausTrajectory.h
```diff
@@ -0,0 +1,67 @@
...
+
+  /// @brief The measurement results for this specific trajectory
+  std::optional<std::map<std::string, std::size_t>> measurement_counts;
```
taalexander, line 49: Now seeing the measurement counts it seems like the this is the overall container for the trajectory for input/intermediate/output state? 

Another approach to consider might be to have a `KrausTrajectory`, `KrausTrajectorySample` and `KrausTrajectoryResult` which would respectively be used at each stage?

runtime/cudaq/ptsbe/PTSBEResult.h
```diff
@@ -0,0 +1,54 @@
...
+/// @brief Complete result from PreTrajectorySamplingEngine
+template <typename ScalarType>
+struct PTSBEResult {
+  /// @brief Task lists - one complete trajectory per inner vector
+  /// Each inner vector contains GateApplicationTasks (original gates + noise
+  /// operations)
+  std::vector<std::vector<nvqir::GateApplicationTask<ScalarType>>> task_lists;
+
+  /// @brief Trajectory metadata
+  std::vector<cudaq::KrausTrajectory> trajectories;
```
taalexander, line 32: Can you remind me how the noise location in the trajectory maps to the gate operation in the task list? Does the noise location correspond to the (integer) location in the task list or the location of the n'th noise event? I think the former is easier from a mapping perspective but it might not line up with the current definition above (`KrausSelection`)

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
```diff
@@ -0,0 +1,148 @@
...
+  if (total_shots == 0) {
+    // Zero shots requested
+    for (auto &traj : trajectories) {
+      traj.num_shots = 0;
+    }
+    return;
+  }
+
+  switch (strategy.type) {
+  case ShotAllocationStrategy::Type::PROPORTIONAL: {
```
taalexander: Should break these out as functions.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    if (total_probability <= 0.0) {
+      throw std::invalid_argument(
+          "Total probability must be positive for proportional allocation");
+    }
+
+    for (auto &traj : trajectories) {
+      traj.num_shots = static_cast<std::size_t>(
+          (traj.probability / total_probability) * total_shots);
+    }
```
taalexander: How does this handle scenario where number of trajectories is greater than number of shots?

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
```diff
@@ -0,0 +1,148 @@
...
+    break;
+  }
+
+  case ShotAllocationStrategy::Type::UNIFORM: {
+    // Equal shots per trajectory
+    std::size_t shots_per_traj = total_shots / trajectories.size();
```
taalexander: Same question about trajectories greater than shots.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  if (allocated < total_shots && !trajectories.empty()) {
+    // Distribute remaining shots evenly among trajectories
+    std::size_t remaining = total_shots - allocated;
+    std::size_t idx = 0;
+    while (remaining > 0 && idx < trajectories.size()) {
+      trajectories[idx].num_shots++;
+      remaining--;
+      idx++;
+    }
```
taalexander: Doing this one-by-one is inefficient (although ideally there should only ever be a few remaining at best).

runtime/cudaq/ptsbe/ShotAllocationStrategy.h
```diff
@@ -0,0 +1,66 @@
...
+
+/// @brief Strategy for allocating shots across selected trajectories
+/// After trajectories are selected, this determines how many shots each
+/// trajectory receives.
+struct ShotAllocationStrategy {
+  enum class Type {
+    PROPORTIONAL,    // Shots proportional to trajectory probability
+    UNIFORM,         // Equal shots per trajectory
+    LOW_WEIGHT_BIAS, // Bias toward low-weight error trajectories
+    HIGH_WEIGHT_BIAS // Bias toward high-weight error trajectories
```
taalexander, line 29: Nit - Follow up PR for user-specified strategy via lambda?

runtime/cudaq/ptsbe/ShotAllocationStrategy.h
```diff
@@ -0,0 +1,66 @@
...
+/// @param strategy Shot allocation strategy
+void allocateShots(std::span<cudaq::KrausTrajectory> trajectories,
+                   std::size_t total_shots,
+                   const ShotAllocationStrategy &strategy);
+
+/// @brief Count non-identity errors in a trajectory (error weight)
+/// @param trajectory The trajectory to analyze
+/// @return Number of non-identity Kraus operators (error count)
+[[nodiscard]] constexpr std::size_t
+countErrors(const cudaq::KrausTrajectory &trajectory) {
```
taalexander: Should this be attached to the trajectory?

runtime/cudaq/ptsbe/KrausTrajectory.h
```diff
@@ -0,0 +1,67 @@
...
+        probability(prob), num_shots(shots) {}
+
+  /// @brief Equality comparison for testing
+  /// @param other KrausTrajectory to compare with
+  /// @return true if trajectory_id, selections, probability, and num_shots
+  /// match
+  constexpr bool operator==(const KrausTrajectory &other) const {
+    return trajectory_id == other.trajectory_id &&
+           kraus_selections == other.kraus_selections &&
+           std::abs(probability - other.probability) < 1e-9 &&
```
taalexander: Maybe we should pull 1e-9 out into a constant.

runtime/cudaq/ptsbe/KrausSelection.h
sacpis: I think we can support qudits (depends on the tasks priorities) or most prob in the next version. @1tnguyen Would like to know your thoughts here as well.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
sacpis: Wondering what would be the other way than Round robin approach to distribute the remaining shots as uniformly as possible?

runtime/cudaq/ptsbe/KrausSelection.h
taalexander: I think it gets quite confusing to support them (including the noise operators themselves). It was just a thought as I went through.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
taalexander: Something like:
- Take the ceiling of the remaining shots / num_trajectories. 
- Add that amount to each trajectory so long as remaining > 0. 
- If less than zero add remaining and exit.

runtime/cudaq/ptsbe/KrausSelection.h
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/KrausTrajectory.h
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/ShotAllocationStrategy.h
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/KrausTrajectory.h
sacpis: Addressed in the recent commit.

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
sacpis: Addressed in 3208d9f.

runtime/cudaq/ptsbe/KrausSelection.h
sacpis: Got it. Thanks @taalexander.

runtime/cudaq/ptsbe/ShotAllocationStrategy.h
sacpis: Its a nice idea. Can we handle customization in the next version though?

runtime/cudaq/ptsbe/KrausSelection.h
1tnguyen: We had `qudit` support at the front-end level, but none of the simulator backend support it (all simulators currently assume qubits). The only exception is the photonics backend, which uses qudits and its own custom gate set.

runtime/cudaq/ptsbe/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
```
1tnguyen: nit: year in these new files.

sacpis: Addressed in fcc1c52

runtime/cudaq/ptsbe/ShotAllocationStrategy.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  if (total_probability <= 0.0) {
+    throw std::invalid_argument(
+        "Total probability must be positive for proportional allocation");
+  }
+
+  for (auto &traj : trajectories) {
+    traj.num_shots = static_cast<std::size_t>(
+        (traj.probability / total_probability) * total_shots);
+  }
```
1tnguyen, line 49: Does this guarantee that the sum of all `traj.num_shots` == `total_shots`?

1tnguyen: nvm, I saw the `correctRoundingErrors` function below.

runtime/cudaq/ptsbe/KrausTrajectory.h
sacpis: Seems like yes. I think here we can use the builder pattern to construct the kraustrajectory keeping allocateShots optional.

runtime/cudaq/ptsbe/KrausTrajectory.h
sacpis: > Now seeing the measurement counts it seems like the this is the overall container for the trajectory for input/intermediate/output state?
> 
> Another approach to consider might be to have a `KrausTrajectory`, `KrausTrajectorySample` and `KrausTrajectoryResult` which would respectively be used at each stage?
Seems like creating separate classes will create more types, more objects. Thinking if we really need these many classes. Would like to have your and other thoughts on it.. One more thing that comes to mind is with the addition of different algos (observe, run etc), we may have to add those many times number of classes.

PR #3809, opened 2026-01-28
Add `distributed-device-call` pass to the default simulation pipeline

1tnguyen:
### Description

We accidentally dropped this pass; hence, `device_call` is not properly lowered to a pure function call.


bmhowe23 (approved):
Thanks, Thien!

Issue #3818, opened 2026-01-29
`ApplyOpSpecialization` doesn't support `cc.loop`s with multiple arguments


atgeller:
Currently, `ApplyOpSpecialization` seems to assume that each loop will have exactly one argument, which is the stepped loop iteration variable. There are a few failure modes when this assumption gets broken:

A segfault can occur if the first argument is not the iteration variable being stepped, because the value in the step block will be accessed, but may not be set. I was finding this in the following test case:
```
unsigned long j = num_qubits;
for (unsigned long i = 0; i < num_qubits; i++) {
  x<cudaq::ctrl>(qubits[i], qubits[j]);
  j--;
}
```

If a variable is threaded through two loops by `memtoreg`, then it will not properly be handled when the loops are reversed, so A) the second loop will still refer to the result of the first loop, even after it's moved before the first loop, causing invalid error; and B) the argument will still run forward through the loop. For example:
```
unsigned long j = 0;
for (unsigned long i = 0; i < num_qubits/2; i++) {
  x<cudaq::ctrl>(qubits[i], qubits[j]);
  j++;
}
for (unsigned long i = num_qubits/2; i < num_qubits; i++) {
  x<cudaq::ctrl>(qubits[i], qubits[j]);
  j++;
}
```

schweitzpgi:
I don't think this is really a problem for `ApplyOpSpecialization` to deal with.

We should have a distinct pass that merges induction variables. Especially if the inductions are all linear expressions. This would reduce the loop nest's bookkeeping overhead at the expense of computing small linear expressions per iteration. In the case submitted in #4897 , that expression is simply the identity function.

PR #3819, opened 2026-01-29
Improve error handling for loops in ApplyOpSpecialization

atgeller:
### Description

This is a bandaid for #3818 that avoids transforming a `quake.apply<adj>` in `ApplyOpSpecialization` instead of failing in a more ugly way if a loop in the adjoint function has multiple arguments.

E.g.:
```
unsigned long j = num_qubits;
for (unsigned long i = 0; i < num_qubits; i++) {
  x<cudaq::ctrl>(qubits[i], qubits[j]);
  j--;
}
```
Would result in a runtime error `could not autogenerate the adjoint of a kernel`

### DIFF
--- a/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
+++ b/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
@@ -282,11 +282,19 @@ static bool regionHasUnstructuredControlFlow(Region &region) {
         return true; // TODO: Currently, cloneReversedLoop requires that the
                      // first operand is the induction variable
                      // See https://github.com/NVIDIA/cuda-quantum/issues/3818
-      for (size_t i = 0; i < loop.getNumResults(); i++)
-        if (!loop.getResult(i).getUses().empty())
+      for (size_t i = 0; i < loop.getNumResults(); i++) {
+        if (!loop.getResult(i).getUses().empty()) {
+          auto res = loop.getResult(i);
+          auto users = SmallVector<Operation *>(res.getUsers().begin(),
+                                                res.getUsers().end());
+          if (users.size() == 1 && users[0]->hasTrait<OpTrait::IsTerminator>())
+            continue;  // Exception, threading variables through nested loops is
+                       // acceptable
           return true; // TODO: Threading variables through loops as
                        // arguments/returns is not handled properly
                        // See https://github.com/NVIDIA/cuda-quantum/issues/3818
+        }
+      }
     }
     for (auto &reg : op.getRegions())
       if (regionHasUnstructuredControlFlow(reg))


atgeller:
> /ok to test c0bee12

targettests/execution/adjoint.cpp
```diff
@@ -0,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// RUN: nvq++ --enable-mlir %s -o %t && %t
+// XFAIL: *
```
Renaud-K, line 10: Is there an issue tracking this? Could be good the comment it here.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -692,8 +692,8 @@ class ApplySpecializationPass
   /// For each Op in \p invertedOps, visit them in reverse order and move each
   /// to just in front of \p term (the end of the function). This reversal of
   /// the order of quantum operations is done recursively.
-  static void reverseTheOpsInTheBlock(Location loc, Operation *term,
-                                      SmallVector<Operation *> &&invertedOps) {
+  void reverseTheOpsInTheBlock(Location loc, Operation *term,
```
Renaud-K: Is this in an anonymous workspace? Is it why you are removing static?

test/Transforms/apply_op_specialization.qke
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    func.func @test_caller() {
+        %c8_i64 = arith.constant 8 : i64
+        %0 = quake.alloca !quake.veq<8>
+        %1 = cc.instantiate_callable @test_threaded_val() : () -> !cc.callable<(!quake.veq<8>, i64) -> ()>
+        quake.apply<adj> %1 %0, %c8_i64 : (!quake.veq<8>, i64) -> ()
+        return
+    }
+}
```
Renaud-K: I was asked in a prior cudaq review to end files with an empty line. I still have to find out about the rationale. I thought I would point it out here.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
atgeller: It's defined within the class for the pass. I'm removing `static` to have access to `signalPassFailure()` when an issue is encountered. The other option, which may be more proper/idiomatic, would be to replace the `void` return with a `LogicalResult`.

targettests/execution/adjoint.cpp
atgeller: Added a comment.

test/Transforms/apply_op_specialization.qke
atgeller: Fixed.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
Renaud-K: I see. It is a static member. Then I think that if it is easy, it is best to keep it static and return a value like you said. Somebody went through the trouble of making it static, clearly indicating that it does not depend on any member, so yes, it would be better.

atgeller:
> /ok to test 0f2f17c

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
atgeller: I've addressed this by moving the checking logic earlier to `regionHasUnstructuredControlFlow`.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -275,6 +275,17 @@ static bool regionHasUnstructuredControlFlow(Region &region) {
...
+      auto contOp =
+          cast<cudaq::cc::ContinueOp>(loop.getStepBlock()->getTerminator());
+      if (!contOp.getOperand(0).getDefiningOp())
+        return true; // TODO: Currently, cloneReversedLoop requires that the
+                     // first operand is the induction variable
+      for (size_t i = 0; i < loop.getNumResults(); i++)
+        if (!loop.getResult(i).getUses().empty())
+          return true; // TODO: Threading variables through loops as
+                       // arguments/returns is not handled properly
+    }
```
khalatepradnya, line 298: The PR description has this error message -
```
error: ApplyOpSpecialization does not currently support loops with multiple arguments. This happens as a result of some existing value being assigned to in a loop.
```
Where is it coming from? Can we make it actionable?

Does it work if for instance the example below does not use the additional variable (`j`)? If so, we can add to the error message - "consider re-writing the kernel to not use single loop variable"?

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -275,6 +275,17 @@ static bool regionHasUnstructuredControlFlow(Region &region) {
     if (!isa<cudaq::cc::IfOp>(op) && !cudaq::opt::isaMonotonicLoop(&op) &&
         op.getNumRegions() > 1)
       return true; // Op has multiple regions but is not a known Op.
+    if (auto loop = dyn_cast<cudaq::cc::LoopOp>(op)) {
+      auto contOp =
+          cast<cudaq::cc::ContinueOp>(loop.getStepBlock()->getTerminator());
+      if (!contOp.getOperand(0).getDefiningOp())
+        return true; // TODO: Currently, cloneReversedLoop requires that the
```
khalatepradnya, line 282: nit: It might help to add the issue number here, so when we revisit, we know where to look.

python/tests/kernel/test_kernel_features.py
```diff
@@ -162,6 +162,37 @@ def kernel(withAdj: bool):
     #assert len(counts) == 1 and '00000000' in counts
 
 
+# TODO: Update when `ApplyOpSpecialization` can handle multi-argument loops
+# See: https://github.com/NVIDIA/cuda-quantum/issues/3818
+@pytest.mark.xfail(raises=RuntimeError)
+def test_adjoint_bug():
+    num_electrons = 2
+    num_qubits = 8
```
khalatepradnya: Q: Curious to know if this failure is seen only at higher number of qubits?

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
atgeller: Ah, that error message is outdated, I will update the PR description. I changed how the error is handled to be more in line with the rest of the pass, where it simply refuses to transform the `quake.apply`, which will eventually result in a runtime error `could not autogenerate the adjoint of a kernel`.

atgeller: Unfortunately that does make the error message less helpful, but it keeps the pass a little cleaner and matches the existing error handling.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
atgeller: Done.

python/tests/kernel/test_kernel_features.py
atgeller: Looks like it works with 2 qubits but not with 4.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
schweitzpgi: Loop normalization should normalize the induction to look like `i = i + e` rather than `i = e + i`, so this sounds correct. Are there inductions that are still `i = e + i`?

schweitzpgi:
> /ok to test b94a3c2

atgeller:
> /ok to test 4f68537

atgeller:
> /ok to test d5c8c00

atgeller:
This is blocked on #3989 to enable the tests to pass.

atgeller:
> /ok to test d7350b3

atgeller:
> /ok to test fa9b102

atgeller:
> /ok to test cee99b6

atgeller:
> /ok to test 5fa4710

Issue #3762, opened 2026-01-15
`cudaq::{control, adjoint}` not supported in NVQIR-based simulations


lmondada:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following kernel definitions
```c++
void cx1() __qpu__ {
  cudaq::qvector q(2);
  cudaq::control([&]() { x(q[1]); }, q[0]);
}

void cx2() __qpu__ {
  cudaq::qvector q(2);
  x<cudaq::ctrl>(q[0], q[1]);
}
```

should be equivalent. However, the first one is not executed correctly when simulating locally (see "Steps to reproduce the bug" below). This is because `cudaq::control` (and in fact, most of the functionality implemented in `runtime/cudaq/qis/qubit_qis.h`) is implemented using `ExecutionManager`s, which NVQIR bypasses entirely (calling directly into the `Simulator` instances).

### Steps to reproduce the bug

Using the kernel definitions above and the following `main`:

```c++
#include <cudaq.h>
#include <cudaq/algorithms/get_state.h>
#include <cudaq/qis/qubit_qis.h>
#include <iostream>

int main() {
  std::cout << "using cudaq::control: " << std::endl;
  auto state = cudaq::get_state(cx1);
  state.dump();

  std::cout << "using x<cudaq::ctrl>: " << std::endl;
  state = cudaq::get_state(cx2);
  state.dump();
  return 0;
}
```

shows the issue:

```bash
$ nvq++ build/test.cpp && ./a.out
using cudaq::control: 
(0,0)
(0,0)
(1,0)
(0,0)
using x<cudaq::ctrl>: 
(1,0)
(0,0)
(0,0)
(0,0)
```

The first state should match the second.

### Expected behavior

The first state should match the second.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: main
- **Python version**: n/a
[...]

PR #3823, opened 2026-01-30
Ptsbe simulator interface

taalexander:
### Description
This PR adds the core interface for sampling with PTSBE and integration proposal using concepts. It integrates with the core data structures introduced in https://github.com/NVIDIA/cuda-quantum/pull/3808. 

The rough pathway is:
- In sample detect if ptsbe is selected and is a simulator. If so route call to `runSamplePTSBE` implementation
- `runSamplePTSBE` runs a trace and gets the "Circuit trace" (this will of course have downstream issues with MCM/dynamic circuits which we will have to think about.
- Trace is unioned with PTS and passed as input to concept method with fallback to generic implementation (neither of which are implemented yet).

The next steps would be to added conversion from trace to gate task list internally in the simulator method to allow generic execution/batched execution (no mechanism yet exists to convert from trace to the actual gate, will do in follow up).

### DIFF
--- a/runtime/cudaq/algorithms/sample.h
+++ b/runtime/cudaq/algorithms/sample.h
@@ -289,7 +289,13 @@ sample_result sample(const sample_options &options, QuantumKernel &&kernel,
   sample_result ret;
   // TODO: Only have added PTSBE to this core sample method right now.
   // If we agree on the approach I will add to others.
-  if (options.use_ptsbe && platform.is_simulator()) {
+  if (options.use_ptsbe) {
+    if (options.noise.empty())
+      throw std::runtime_error(
+          "PTSBE requires a noise model to be set. Please provide a noise "
+          "model in sample_options.");
+    if (!platform.is_simulator())
+      throw std::runtime_error("PTSBE is only supported on simulators.");
     ret = ptsbe::runSamplingPTSBE(
         [&]() mutable { kernel(std::forward<Args>(args)...); }, platform,
         kernelName, shots);


runtime/cudaq/algorithms/sample.h
```diff
@@ -282,10 +286,21 @@ sample_result sample(const sample_options &options, QuantumKernel &&kernel,
...
   platform.set_noise(&options.noise);
-  auto ret = details::runSampling(
-                 [&]() mutable { kernel(std::forward<Args>(args)...); },
-                 platform, kernelName, shots, options.explicit_measurements)
-                 .value();
+
+  sample_result ret;
+  // TODO: Only have added PTSBE to this core sample method right now.
+  // If we agree on the approach I will add to others.
+  if (options.use_ptsbe && platform.is_simulator()) {
```
sacpis: nit: and `noise model is available` condition?

runtime/cudaq/ptsbe/PTSBESample.h
```diff
@@ -0,0 +1,231 @@
...
+/// measurement outcomes.
+///
+/// Detection uses two mechanisms:
+/// 1. MLIR-compiled kernels: Check registered quake code for
+///    qubitMeasurementFeedback attribute
+/// 2. Library mode: Check registerNames populated during tracing when
+///    __nvqpp__MeasureResultBoolConversion is called
+///
+/// @param kernelName Name of the kernel (for MLIR lookup)
+/// @param ctx ExecutionContext populated after tracing (for library mode)
```
sacpis, line 52: Should we remove `(for library mode)`?

runtime/cudaq/ptsbe/PTSBESample.h
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+/// @brief Extract measurement qubits from kernel trace
+///
+/// For PTSBE POC, assumes terminal measurements on all qubits in the circuit.
+/// Future work may extract explicit measurement operations from the trace.
+///
+/// @param trace Captured kernel trace
+/// @return Vector of qubit indices [0, 1, ..., numQudits-1]
```
sacpis: Sorry, a bit confused here. The function name seems to extract `Qubits` but the doc and the implementation seems to be work with `Qudits`. Should we change the function name in that case?

runtime/cudaq/algorithms/sample.h
taalexander: I'm actually wondering what we should do here? Technically without a noise model I think PTSBE would still work - it just would be a single noiseless trajectory from which we sample shots? I could see the options being :

1. Return samples just down the PTSBE path
2. Fall back to normal sample path
3. Error

I tend to think we should go with 1. as it would work with:
1. The case where the noise model doesn't have any noise
2. A future point in time where we annotate noise in the kernel itself rather than an out of band context object.

runtime/cudaq/ptsbe/PTSBESample.h
taalexander: Hmm, what do you think? I think the current sampling implementation supports library mode - https://github.com/NVIDIA/cuda-quantum/blob/69ddae9f5ac30ec1b3119152c6c87eb840eb5b53/runtime/cudaq/algorithms/sample.h#L64-L90

runtime/cudaq/ptsbe/PTSBESample.h
taalexander: Woops, this is a byproduct of Trace being defined in terms of qudits. Replacing.

taalexander: Addressed in e3be4376

runtime/cudaq/algorithms/sample.h
sacpis: > I'm actually wondering what we should do here? Technically without a noise model I think PTSBE would still work - it just would be a single noiseless trajectory from which we sample shots? I could see the options being :
> 
> 1. Return samples just down the PTSBE path
> 2. Fall back to normal sample path
> 3. Error
> 
> I tend to think we should go with 1. as it would work with:
> 
> 1. The case where the noise model doesn't have any noise
> 2. A future point in time where we annotate noise in the kernel itself rather than an out of band context object.

I am more leaning towards option 3 (throwing error).

>The case where the noise model doesn't have any noise
I am wondering why should we have a noise model without any noise?

runtime/cudaq/ptsbe/PTSBESample.h
sacpis: @khalatepradnya Would you please confirm if we are going to support library mode in the future?

khalatepradnya: Short answer - yes.

sacpis: @taalexander Let's keep the comment.

runtime/cudaq/algorithms/sample.h
taalexander: I am ok with this for now if it's what you want. I think we will have to walk this back if/when we want to support noise in-kernel (which it sounds like is coming with DEMs).

taalexander: Addressed in 53d758f90e

runtime/cudaq/algorithms/sample.h
```diff
@@ -42,13 +43,12 @@ runSampling(KernelFunctor &&wrappedKernel, quantum_platform &platform,
             std::size_t qpu_id = 0, details::future *futureResult = nullptr,
             std::size_t batchIteration = 0, std::size_t totalBatchIters = 0) {
 
-  auto hasConditionalFeebdback =
-      cudaq::kernelHasConditionalFeedback(kernelName);
+  auto hasConditionalFeedback = cudaq::kernelHasConditionalFeedback(kernelName);
```
lmondada, line 46: This makes me think: if you need to make any changes to the codebase that aren't directly related to the PTSB feature, consider making a separate PR into main, so that we can keep the delta between this feature branch and `main` as small as necessary.

Obviously that comes with a higher workload and you're quite tight on time, so up to you to make a call when/if that's worth the additional effort. Just keep it in mind please :)

runtime/cudaq/ptsbe/PTSBEInterface.h
```diff
@@ -0,0 +1,87 @@
...
+///
+/// @param simulator Circuit simulator instance
+/// @param batch Batch specification
+/// @return Aggregated execution result
+/// @throws std::runtime_error Not yet implemented
+template <typename ScalarType>
+cudaq::sample_result
+executePTSBE(nvqir::CircuitSimulatorBase<ScalarType> &simulator,
+             const PTSBatch &batch) {
+  throw std::runtime_error("executePTSBE: Not implemented");
```
lmondada, line 58: Hmm, the overall design you're going with is quite nice and straight-forward, but unfortunately it looks like you'll have to bypass NVQIR entirely here (I assume you'll be calling the simulator instance directly based on the traced calls)...

Things to watch out for: parts of the code base were written with the expectation that QIR calls would always flow through an `ExecutionManager`. This is already causing bugs, see https://github.com/NVIDIA/cuda-quantum/issues/3762 (and maybe others that I haven't found yet). You would run into the same issue.

Another potential pitfall is that a lot of the `CircuitSimulator` backends expect the execution context to be set in a certain way, so you'll probably want to set it accordingly, too. Relatedly, watch out for initialization/deinitialization of `CircuitSimulator` instances that is implemented within `setExecutionContext` and `resetExecutionContext`. Some simulators won't execute any gates until it gets "flushed" within the reset context method. The tricky part here is that different simulators behave differently, so you'll have to either explicitly restrict the simulators you support or test them well.

You can pretty much ignore/hack around these particular issues as I am already working on proper fixes for them as part of the refactor. However, this also means that we'll inevitably end up with merge conflicts. Might be good for you to keep a close eye on refactors in main, so we can resolve (and maybe even co-design) some of the changes to the runtime.

lmondada: actually, I just came back from my coffee break and realised something!

You _might_ be able to define a separate C++ lambda for each trajectory that you want to sample from. It would capture the trace and loop over the recorded gates, executing them and injecting noise where required. You could pass that directly to `runSampling` and handle it like any other sample job. This is the closest we can currently get to generating kernels at kernel compile time (which is what we eventually want to be doing).

This would reuse more of the existing code and reduce the number of edge cases that you'll have to handle manually.

taalexander: > Hmm, the overall design you're going with is quite nice and straight-forward, but unfortunately it looks like you'll have to bypass NVQIR entirely here (I assume you'll be calling the simulator instance directly based on the traced calls)...

Yes, this is the intended execution pathway right now. It would be technically possible to inject ourself at a lower level than this in a similar manner to the gate queue to build up the trajectory. This would however be much more invasive than the existing approach requiring intercepts (or a completely separate wrapper class for the base simulator which intercepts and caches all calls) to be instantiated to effectively perform this tracing from the PoV of the runtime and than be dispatched internally when the appropriate output method if called. If you think this is required or better we can certainly consider this.

> You might be able to define a separate C++ lambda for each trajectory that you want to sample from. It would capture the trace and loop over the recorded gates, executing them and injecting noise where required. You could pass that directly to runSampling and handle it like any other sample job. This is the closest we can currently get to generating kernels at kernel compile time (which is what we eventually want to be doing).

Thanks for the feedback on this. We had considered doing something like this with a custom noise model. The problem is that we need to group batches of trajectories for execution on the mgpu simulator. The current way this is handled relies on a *single* trajectory being executed N unique times with a sampled noise model such that it has the same structure. I think the approach you suggest also would make this difficult to tie together the individual calls to form a batch across circuits (hence the executePTSBE method consumes a batch of pre-sampled trajectories).

runtime/cudaq/algorithms/sample.h
taalexander: Yes, will consider it. My follow up PR - https://github.com/NVIDIA/cuda-quantum/pull/3832 reshuffles some base class components in the simulator and it may be worth it there.

runtime/cudaq/ptsbe/PTSBEInterface.h
lmondada: Ah yes, you're right that I did not consider that you'd have to batch multiple trajectories together. Then ignore my second comment.

PR #3824, opened 2026-01-30
Reenable `state` support for remote REST QPU platforms

1tnguyen:
### Description

Re-enable state argument support for remote REST QPU:

- Restore `replace-state-with-kernel` and `state-prep` passes. Add back the corresponding Python and C++ tests.

Note: For `state-prep` pass only support state amplitudes provided as arrays, not a `state` constructed outside the kernel with an amplitude array (i.e., state constructed by a different simulator). An error will be raised accordingly.

- Modify the Python and C++ bridges to directly handle `qvector` allocation with amplitude arrays. We no longer want to wrap it via a `state` construction (which needs to use the local simulator). 

Original implementation of these passes is in #1891 and #2291.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -134,7 +134,8 @@ except Exception:
 from .display import display_trace
 from .kernel.kernel_decorator import kernel, PyKernelDecorator
 from .kernel.kernel_builder import (make_kernel, QuakeValue, PyKernel)
-from .kernel.ast_bridge import (globalAstRegistry, globalRegisteredOperations)
+from .kernel.ast_bridge import (globalAstRegistry, globalRegisteredOperations,
+                                PyASTBridge)
 from .runtime.sample import sample
 from .runtime.sample import sample_async, AsyncSampleResult
 from .runtime.observe import observe
--- a/python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
+++ b/python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
@@ -39,6 +39,13 @@ except ImportError:
 
 try:
     from torchdiffeq import odeint
+    from torchdiffeq._impl.dopri5 import Dopri5Solver
+    from torchdiffeq._impl.dopri8 import Dopri8Solver
+    from torchdiffeq._impl.bosh3 import Bosh3Solver
+    from torchdiffeq._impl.adaptive_heun import AdaptiveHeunSolver
+    from torchdiffeq._impl.fehlberg2 import Fehlberg2
+    from torchdiffeq._impl.rk_common import _RungeKuttaState
+    from torchdiffeq._impl.misc import _rms_norm
 except ImportError:
     has_torchdiffeq = False
 
@@ -96,9 +103,16 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[cudaq_runtime.State]):
         self.order = None
         self.is_density_state = None
         self.batchSize = None
+        self._dimensions_list = None
+        self._solver_instance = None
 
     def compute_rhs(self, t, vec):
-        t_scalar = t.item()
+        if torch.is_tensor(t):
+            t_scalar = t.item()
+            if isinstance(t_scalar, complex):
+                t_scalar = t_scalar.real
+        else:
+            t_scalar = float(t.real) if isinstance(t, complex) else float(t)
         # Note: this RHS compute is on the hot path of the integrator;
         # hence, we minimize overhead as much as possible.
         # In particular, avoid data conversion between different frameworks.
@@ -106,25 +120,60 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[cudaq_runtime.State]):
         # Get device pointer of the input torch tensor
         device_ptr = vec.data_ptr()
         size = vec.numel()
+
+        if self._dimensions_list is None:
+            self._dimensions_list = list(self.dimensions)
+
         # Wrap the device pointer as a `cudaq::state` (no copy)
         temp_state = bindings.initializeState(device_ptr, size,
-                                              list(self.dimensions),
+                                              self._dimensions_list,
                                               self.batchSize)
         # Pre-allocate output tensor (torch tensor)
         result_vec = torch.zeros_like(vec)
         # Wrap the output tensor device pointer as a `cudaq::state` (no copy)
         result_state = bindings.initializeState(result_vec.data_ptr(), size,
-                                                list(self.dimensions),
+                                                self._dimensions_list,
                                                 self.batchSize)
         # Compute the RHS into the output state
         self.stepper.compute_inplace(temp_state, t_scalar, result_state)
         return result_vec
+
+    def _create_wrapped_rhs_func(self):
+        # Wrapper that adds the required callback methods
+        class RHSFuncWrapper:
+
+            def __init__(self, integrator):
+                self.integrator = integrator
+                self.callback_step = lambda *args, **kwargs: None
+                self.callback_accept_step = lambda *args, **kwargs: None
+                self.callback_reject_step = lambda *args, **kwargs: None
+
+            def __call__(self, t, y, perturb=None):
+                return self.integrator.compute_rhs(t, y)
+
+        return RHSFuncWrapper(self)
 
     def __post_init__(self):
         self.n_steps = self.integrator_options.get('nsteps', 10)
         self.atol = self.integrator_options.get('atol', self.atol)
         self.rtol = self.integrator_options.get('rtol', self.rtol)
         self.order = self.integrator_options.get('order', None)
+
+    def _is_adaptive_solver(self):
+        adaptive_solvers = [
+            'dopri5', 'dopri8', 'bosh3', 'adaptive_heun', 'fehlberg2'
+        ]
+        return self.solver in adaptive_solvers
+
+    def _get_solver_class(self):
+        solver_map = {
+            'dopri5': Dopri5Solver,
+            'dopri8': Dopri8Solver,
+            'bosh3': Bosh3Solver,
+            'adaptive_heun': AdaptiveHeunSolver,
+            'fehlberg2': Fehlberg2,
+        }
+        return solver_map.get(self.solver)
 
     def integrate(self, t):
         if self.stepper is None:
@@ -164,26 +213,55 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[cudaq_runtime.State]):
         y0_cupy = to_cupy_array(self.state)
         y0 = torch.from_dlpack(y0_cupy)
 
-        # time span
-        t_span = torch.tensor([self.t, t], device='cuda', dtype=torch.float64)
+        if self._is_adaptive_solver():
+            solver_class = self._get_solver_class()
 
-        # solve ODE using TorchDiffEq
-        solution = odeint(self.compute_rhs,
-                          y0,
-                          t_span,
-                          method=self.solver,
-                          rtol=self.rtol,
-                          atol=self.atol)
+            if self._solver_instance is None:
+                wrapped_func = self._create_wrapped_rhs_func()
+                self._solver_instance = solver_class(func=wrapped_func,
+                                                     y0=y0,
+                                                     rtol=self.rtol,
+                                                     atol=self.atol,
+                                                     norm=_rms_norm)
+            else:
+                self._solver_instance.y0 = y0
+                if hasattr(self._solver_instance, 'rk_state'):
+                    self._solver_instance.first_step = self._solver_instance.rk_state.dt
 
-        # solution at final time
-        y_t = solution[-1]
+            # Initialize solver state
+            t_init = torch.tensor([self.t, t],
+                                  device='cuda',
+                                  dtype=torch.float64)
+            self._solver_instance._before_integrate(t_init)
+
+            t_target = torch.tensor(t, device='cuda', dtype=torch.float64)
+            y_t = self._solver_instance._advance(t_target)
+        else:
+            # time span
+            t_span = torch.tensor([self.t, t],
+                                  device='cuda',
+                                  dtype=torch.float64)
+
+            # solve ODE using TorchDiffEq
+            solution = odeint(self.compute_rhs,
+                              y0,
+                              t_span,
+                              method=self.solver,
+                              rtol=self.rtol,
+                              atol=self.atol)
+
+            # solution at final time
+            y_t = solution[-1]
 
         # convert the solution back to CuPy array
         y_t_cupy = cp.from_dlpack(y_t)
 
+        if self._dimensions_list is None:
+            self._dimensions_list = list(self.dimensions)
+
         # Keep results in GPU memory
         self.state = cudaq_runtime.State.from_data(y_t_cupy)
-        self.state = bindings.initializeState(self.state, list(self.dimensions),
+        self.state = bindings.initializeState(self.state, self._dimensions_list,
                                               self.is_density_state,
                                               self.batchSize)
         self.t = t
@@ -191,6 +269,7 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[cudaq_runtime.State]):
     def set_state(self, state: cudaq_runtime.State, t: float = 0.0):
         super().set_state(state, t)
         self.batchSize = bindings.getBatchSize(state)
+        self._solver_instance = None
 
 
 class CUDATorchDiffEqRK4Integrator(CUDATorchDiffEqIntegrator):
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -54,77 +54,203 @@ ALLOWED_TYPES_IN_A_DATACLASS = [int, float, bool, cudaq_runtime.qview]
 
 class PyScopedSymbolTable(object):
 
-    # FIXME: make this similar to `pystack` to account for local functions
-    def __init__(self):
-        self.symbolTable = {}
-        self.scopes = deque()
-        self.scopeID = 0
+    class Scope(object):
 
-    def pushScope(self):
-        self.scopes.append(self.scopeID)
-        self.scopeID += 1
+        def __init__(self, scope_root, parent=None):
+            assert isinstance(scope_root, Block) and hasattr(
+                scope_root, 'owner')
+            self.root = scope_root
+            self.parent = parent
+            self._blockID = -1
+            self._parentBlocks = deque()
+            self._symbols = {}
 
-    def popScope(self):
-        self.scopes.pop()
+        def beginBlock(self):
+            self._blockID += 1
+            self._parentBlocks.append(self._blockID)
 
-    @property
-    def depth(self):
-        return len(self.scopes)
+        def endBlock(self):
+            assert self._parentBlocks
+            self._parentBlocks.pop()
 
-    def isDefined(self, symbol):
-        return symbol in self.symbolTable
+        def isDefined(self, symbol):
+            return symbol in self._symbols
 
-    def add(self, symbol, value, level=0):
-        """
-        Add a symbol to the scoped symbol table at any scope level.
-        """
-        self.symbolTable[symbol] = (value, level)
+        def tryGet(self, symbol):
+            """
+            Returns the value of the given symbol in this scope,
+            as well as a boolean indicating whether the value is
+            valid in the current scope.
+            """
+            if not symbol in self._symbols:
+                return None, False
 
-    def __contains__(self, symbol):
-        if not symbol in self.symbolTable:
+            # We need to make sure that the symbol is not only defined,
+            # but also accessible at this location in the MLIR we generate.
+            # To check for this, the symbol table keeps track of which block
+            # a symbol is defined in. Because/if some variables are stored
+            # as values in the symbol table, a value defined in an inner
+            # block may not be accessible in the outer block (MLIR fails
+            # with "operand does not dominate this use"). We hence fail
+            # with a comprehensive error if a symbol is defined according
+            # to Python scoping rules, but not valid to use at the current
+            # location in the generated MLIR.
+            value, sid = self._symbols[symbol]
+            return value, sid in self._parentBlocks
+
+        def addOrUpdate(self, symbol, value):
+            assert self._parentBlocks
+            if hasattr(value, 'owner'):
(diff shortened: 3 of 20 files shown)


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -405,9 +405,30 @@ void cudaq::packArgs(OpaqueArguments &argData, py::list args,
...
-            });
+            auto *stateArg = arg.cast<state *>();
+
+            if (stateArg == nullptr)
+              throw std::runtime_error("Null cudaq::state* argument passed.");
+            auto simState = cudaq::state_helper::getSimulationState(
+                const_cast<cudaq::state *>(stateArg));
+            if (!simState)
+              throw std::runtime_error("Error: Unable to retrieve simulation "
+                                       "state from cudaq::state.");
```
sacpis: nit: We can add `The state object may be invalid or corrupted`?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+    // the position where the lth and (l + 1)th bit strings g[l] and g[l+1] of
+    // the binary reflected Gray code differ.
+    auto position = std::log2(code[i] ^ code[(i + 1) % code.size()]);
+    // N.B: The algorithm expects the least significant bit (LSb) on the left
+    //
+    //  lsb -v
+    //       001
+    //         ^- msb
+    //
+    // Meaning that the bitstring 001 represents the number four instead of one.
```
sacpis, line 55: nit: Should we add a unit test for qubit ordering to ensure that this convention is maintained correctly? We have had issue in the past with respect to qubit ordering.

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+    // We have the converted position: 2, which is what we need.
+    indices.emplace_back(numBits - position - 1);
+  }
+  return indices;
+}
+
+std::vector<double> convertAngles(const std::span<double> alphas) {
+  // Implements Eq. (3) from https://arxiv.org/pdf/quant-ph/0407010.pdf
+  //
+  // N.B: The paper does fails to explicitly define what is the dot operator in
```
sacpis: nit: The paper fails to

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+  // the exponent of -1. Ref. 3 solves the mystery: its the bitwise inner
```
sacpis: Sorry Ref. 3 from where? Is it from the `State preparation` section of the paper?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+      numerator +=
+          std::pow(std::abs(data[(2 * j - 1) * (1 << (k - 1)) + l - 1]), 2);
+    }
+
+    double denominator = 0;
+    for (std::size_t l = 1; l <= (1ULL << k); ++l) {
+      denominator += std::pow(std::abs(data[(j - 1) * (1 << k) + l - 1]), 2);
+    }
+
+    if (denominator == 0.0) {
```
sacpis, line 139: I think we should never enter in this loop, right? If we do, then can we throw a runtime_error, `divide by 0`?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+      assert(numerator == 0.0 &&
+             "If the denominator is zero, the numerator must also be zero.");
+      angles.push_back(0.0);
+      continue;
+    }
+    angles.push_back(2.0 * std::asin(std::sqrt(numerator / denominator)));
```
sacpis: If numerator / denominator > 1.0 due to floating-point errors, std::asin will return NaN.

We can do

```
double ratio = numerator / denominator;
ratio = std::min(1.0, std::max(0.0, ratio));
angles.push_back(2.0 * std::asin(std::sqrt(ratio)));
```

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+  std::unordered_map<std::size_t, mlir::Value> qubitRefs =
+      std::unordered_map<std::size_t, mlir::Value>();
+};
+
+class StateDecomposer {
+public:
+  StateDecomposer(StateGateBuilder &b, std::span<std::complex<double>> a,
+                  double t)
+      : builder(b), amplitudes(a), numQubits(log2(a.size())),
```
sacpis, line 202: If a.size() is not a power of 2, log2 returns a non-integer. Should we validate that?

```
a,size() == 0|| (a.size() & (a.size() - 1)) != 0
```

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+    // Apply uniformly controlled z-rotations, the construction in Eq. (4).
+    for (std::size_t j = 1; j <= numQubits; ++j) {
+      auto k = numQubits - j + 1;
+      auto numControls = j - 1;
+      auto target = j - 1;
+      auto alphaZk = cudaq::details::getAlphaZ(phases, numQubits, k);
+      if (alphaZk.empty())
+        continue;
+      applyRotation<quake::RzOp>(alphaZk, numControls, target);
+    }
```
sacpis: Seems like this code is repeated (line 222-228) apart from `alphaYk` and `alphaZk`. Should we create a function instead?

lib/Optimizer/Transforms/StatePreparation.cpp
```diff
@@ -0,0 +1,406 @@
...
+        data = cast.getValue();
+
+      if (auto addr = data.getDefiningOp<cudaq::cc::AddressOfOp>()) {
+        auto globalName = addr.getGlobalName();
+        auto module = init->getParentOfType<mlir::ModuleOp>();
+        auto symbol = module.lookupSymbol(globalName);
+        if (auto global = dyn_cast<cudaq::cc::GlobalOp>(symbol)) {
+
+          // Read state initialization data from the global array.
+          auto vec = cudaq::opt::factory::readGlobalConstantArray(global);
```
sacpis, line 369: Should we validate vec size to be a power of 2?

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -172,19 +174,17 @@ class PyQPUState : public QPUState {
...
-  auto kernelName = kernel.attr("uniqName").cast<std::string>();
-  auto kernelMod = kernel.attr("qkeModule").cast<MlirModule>();
+state pyGetStateQPU(const std::string &kernelName, MlirModule kernelMod,
+                    py::args args) {
+  auto moduleOp = unwrap(kernelMod);
+  std::string mlirCode;
+  llvm::raw_string_ostream outStr(mlirCode);
+  mlir::OpPrintingFlags opf;
+  opf.enableDebugInfo(/*enable=*/true, /*pretty=*/false);
+  moduleOp.print(outStr, opf);
```
sacpis, line 184: Let's put the serialization code in a try/catch block. In case it fails.

runtime/common/ArgumentConversion.cpp
```diff
@@ -348,6 +348,9 @@ static Value genConstant(OpBuilder &builder, const cudaq::state *v,
   auto loc = builder.getUnknownLoc();
   auto simState =
       cudaq::state_helper::getSimulationState(const_cast<cudaq::state *>(v));
+  if (!simState)
+    throw std::runtime_error(
+        "Error: Unable to retrieve simulation state from cudaq::state.");
```
sacpis: nit: let's add `The state may be invalid or uninitialized.`?

runtime/common/ArgumentConversion.cpp
```diff
@@ -359,8 +362,144 @@ static Value genConstant(OpBuilder &builder, const cudaq::state *v,
...
+
+  if (simState->getKernelInfo().has_value()) {
+    auto [calleeName, code, calleeArgs] = simState->getKernelInfo().value();
+    std::string calleeKernelName =
+        cudaq::runtime::cudaqGenPrefixName + calleeName;
+
+    auto ctx = builder.getContext();
+    auto loc = builder.getUnknownLoc();
+
+    assert(!code.empty() && "Quake code not found for callee");
```
sacpis: Seems like `assert` is removed in release builds. Let's replace all asserts with `if .. throw std::runtime_error` logic.

runtime/common/ArgumentConversion.cpp
```diff
@@ -359,8 +362,144 @@ static Value genConstant(OpBuilder &builder, const cudaq::state *v,
...
+    auto initKernelName = cudaq::runtime::cudaqGenPrefixName + initName;
+    auto numQubitsKernelName =
+        cudaq::runtime::cudaqGenPrefixName + numQubitsName;
+
+    // Create `callee.init_N` and `callee.num_qubits_N` used to replace
+    // `quake.materialize_state` in ReplaceStateWithKernel pass
+    if (!converter.isRegisteredKernel(initName) ||
+        !converter.isRegisteredKernel(numQubitsName)) {
+      createInitFunc(builder, substMod, calleeFunc, initKernelName);
+      createNumQubitsFunc(builder, substMod, calleeFunc, numQubitsKernelName);
```
sacpis: Should we just create one when that one is missing? Wondering why do we need to create both if any one is missing?

```
if (!converter.isRegisteredKernel(initName))
    createInitFunc(builder, substMod, calleeFunc, initKernelName);
if (!converter.isRegisteredKernel(numQubitsName))
    createNumQubitsFunc(builder, substMod, calleeFunc, numQubitsKernelName);
```

runtime/cudaq/algorithms/get_state.h
```diff
@@ -130,13 +130,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
   // Store kernel name and arguments for quantum states.
   return state(
       new QPUState(details::getKernelName(std::forward<QuantumKernel>(kernel)),
+                   cudaq::get_quake_by_name(details::getKernelName(
+                       std::forward<QuantumKernel>(kernel))),
                    std::forward<Args>(args)...));
```
sacpis: Instead of calling `getKernelName` twice, we can cache it outside.

```
auto kernelName = details::getKernelName(std::forward<QuantumKernel>(kernel));
return state(
    new QPUState(kernelName,
                 cudaq::get_quake_by_name(kernelName),
                 std::forward<Args>(args)...));
```

runtime/cudaq/algorithms/get_state.h
```diff
@@ -130,13 +130,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
...
 
 #elif defined(CUDAQ_QUANTUM_DEVICE)
   // Kernel builder is MLIR-based kernel.
   if constexpr (has_name<QuantumKernel>::value)
     return state(new QPUState(
         details::getKernelName(std::forward<QuantumKernel>(kernel)),
+        cudaq::get_quake_by_name(
+            details::getKernelName(std::forward<QuantumKernel>(kernel))),
         std::forward<Args>(args)...));
```
sacpis: Same here.

runtime/cudaq/algorithms/get_state.h
```diff
@@ -130,13 +130,17 @@ auto get_state(QuantumKernel &&kernel, Args &&...args) {
   // Store kernel name and arguments for quantum states.
   return state(
       new QPUState(details::getKernelName(std::forward<QuantumKernel>(kernel)),
+                   cudaq::get_quake_by_name(details::getKernelName(
```
sacpis: Seems like `get_quake_by_name` can return empty strings for kernels without registered MLIR (from line 116).

Let's pull out quakeCode as well and validate it?

```
auto quakeCode = cudaq::get_quake_by_name(kernelName);
if (quakeCode.empty())
    throw std::runtime_error(...);
```

runtime/cudaq/algorithms/get_state.h
sacpis, line 129: This (condition on line 129 and line 137) can be simplified as

```
#elif defined(CUDAQ_QUANTUM_DEVICE)
  #ifdef CUDAQ_LIBRARY_MODE
    if constexpr (!has_name<QuantumKernel>::value)
      throw std::runtime_error(
          "could not create state in library mode for quantum devices");
  #endif
  
  auto kernelName = details::getKernelName(std::forward<QuantumKernel>(kernel));
  auto quakeCode = cudaq::get_quake_by_name(kernelName);
  if (quakeCode.empty())
    throw std::runtime_error(
        "Cannot create state: MLIR code not found for kernel " + kernelName);
  return state(new QPUState(kernelName, quakeCode, std::forward<Args>(args)...));
#endif
```

runtime/cudaq/qis/qvector.h
```diff
@@ -37,6 +37,54 @@ class qvector {
...
   /// @endcond
 
+  /// @brief Construct a `qvector` from an input state vector.
+  /// The number of qubits is determined by the size of the input vector.
+  /// If `validate` is set, it will check the norm of input state vector.
+  explicit qvector(const std::vector<complex> &vector, bool validate)
+      : qudits(std::log2(vector.size())) {
+    if (Levels == 2) {
+      auto numElements = std::log2(vector.size());
+      if (std::floor(numElements) != numElements)
```
sacpis: This comparison might fail due to floating point errors with `std::log2(vector.size())` call as it might return 2.9999999999 or 3.0000000001.
I think we can check the power of 2 using

```
vector.size() == 0 || (vector.size() & (vector.size() - 1)) != 0)
```

runtime/cudaq/qis/qvector.h
```diff
@@ -37,6 +37,54 @@ class qvector {
   qvector() : qudits(1) {}
   /// @endcond
 
+  /// @brief Construct a `qvector` from an input state vector.
+  /// The number of qubits is determined by the size of the input vector.
+  /// If `validate` is set, it will check the norm of input state vector.
+  explicit qvector(const std::vector<complex> &vector, bool validate)
+      : qudits(std::log2(vector.size())) {
+    if (Levels == 2) {
```
sacpis, line 45: Just to confirm. We are not supporting higher level systems, right?

runtime/cudaq/qis/qvector.h
```diff
@@ -37,6 +37,54 @@ class qvector {
...
+    getExecutionManager()->initializeState(targets, vector.data(), precision);
+  }
+  qvector(const std::vector<complex> &vector)
+      : qvector(vector, /*validate=*/false){};
+
+  qvector(const std::vector<double> &vector)
+      : qvector(std::vector<complex>{vector.begin(), vector.end()}) {}
+  qvector(std::vector<double> &&vector)
+      : qvector(std::vector<complex>{vector.begin(), vector.end()}) {}
+  qvector(const std::initializer_list<double> &list)
```
sacpis, line 76: Seems like the move constructor does not actually move. It just takes the rvalue reference but copies elements into new vector anyway. 
I think we can remove it if we don't need it or implement the move as below

```
qvector(std::vector<double> &&vector) {
    std::vector<complex> complexVec;
    complexVec.reserve(vector.size());
    for (auto &val : vector)
        complexVec.emplace_back(val, 0.0);
    *this = qvector(std::move(complexVec));
}
```

runtime/cudaq/qis/qvector.h
```diff
@@ -37,6 +37,54 @@ class qvector {
...
+    for (auto &q : qudits)
+      targets.emplace_back(QuditInfo{Levels, q.id()});
+
+    auto precision = std::is_same_v<complex::value_type, float>
+                         ? simulation_precision::fp32
+                         : simulation_precision::fp64;
+    getExecutionManager()->initializeState(targets, vector.data(), precision);
+  }
+  qvector(const std::vector<complex> &vector)
+      : qvector(vector, /*validate=*/false){};
```
sacpis, line 70: Curious, why have we disabled validation be default?

targettests/execution/state_preparation_vector.cpp
```diff
@@ -54,8 +64,7 @@ __qpu__ void test_real_constant_array() {
 }
 
 __qpu__ void test_real_constant_array_floating_point() {
-  cudaq::qvector v(
-      cudaq::state{std::vector<cudaq::real>({M_SQRT1_2, M_SQRT1_2, 0., 0.})});
+  cudaq::qvector v(std::vector<cudaq::real>({M_SQRT1_2, M_SQRT1_2, 0., 0.}));
```
sacpis, line 67: Should we add tests for states with negative, imaginary etc phases?

sacpis (approved):
Overall LGTM. Thanks @1tnguyen. Left a few comments.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: This is tested in `targettests/execution/state_preparation_vector_sizes.cpp`, where we checked every amplitude location up to 5 qubits (vector size = 32).

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: I think the algorithm is expected to handle this condition (zero denominator) as a zero-rotation angle.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: I addressed the comment [below](https://github.com/NVIDIA/cuda-quantum/pull/3824/changes#r2752072231), which added a validation check at a higher-level, before we start the decomposition.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: Indeed, I think we should. Added in https://github.com/NVIDIA/cuda-quantum/pull/3824/changes/1c7556e548d9508996ba5ff5cb15b62839e7d17c

python/runtime/cudaq/algorithms/py_state.cpp
1tnguyen: This IR printing procedure is not expected to fail, given that we have a valid MLIR module.

runtime/cudaq/algorithms/get_state.h
1tnguyen: I don't think that the suggested refactor is equivalent to the code. Specifically, kernel builder kernels would have MLIR code regardless of compilation mode, and thus can still be supported.

runtime/cudaq/qis/qvector.h
1tnguyen: Actually, we do, e.g., 

https://github.com/NVIDIA/cuda-quantum/blob/745f0577e5a3f3848b50b478367db3cd554a200f/unittests/photonics/PhotonicsTester.cpp#L32

runtime/cudaq/qis/qvector.h
1tnguyen: I'm not 100% sure, but from a high-level standpoint, this norm check procedure here is a simple one (CPU-based, no multi-threading, etc.); thus may not be suitable as a default checker. This data will eventually be sent on to the simulator, which might have a more efficient checking mechanism if it needs to validate the state vector.

runtime/cudaq/qis/qvector.h
1tnguyen: This `qvector` class is a representation of the qubit register, _not_ a state vector, hence we don't store a `complexVec` in this class. For simulation, the data is passed (as a data pointer + size) to the simulator to create its internal data structure.

targettests/execution/state_preparation_vector.cpp
1tnguyen: I added a test for a negative amplitude value in https://github.com/NVIDIA/cuda-quantum/pull/3824/changes/1c7556e548d9508996ba5ff5cb15b62839e7d17c. 

A complex value (with imaginary part) amplitude test would be quite tricky for hardware QPUs, where we can only sample the bitstrings.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: I need to read the paper to figure out that it's referring to the ref [3] in the above paper. I updated the comment with the explicit arxiv link for that ref [3], along with the equation number.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
1tnguyen: Fixed the error message.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: Fixed as suggested.

lib/Optimizer/Transforms/StatePreparation.cpp
1tnguyen: Good point. Refactored this as a helper.

runtime/common/ArgumentConversion.cpp
1tnguyen: Fixed the error message.

runtime/common/ArgumentConversion.cpp
1tnguyen: Definitely. Split the two conditions accordingly.

runtime/common/ArgumentConversion.cpp
1tnguyen: Fixed as suggested.

runtime/cudaq/algorithms/get_state.h
1tnguyen: Sure, fixed accordingly.

runtime/cudaq/algorithms/get_state.h
1tnguyen: Fixed as well.

runtime/cudaq/algorithms/get_state.h
1tnguyen: Here, we don't pass `false` to `get_quake_by_name`, so it would throw if it cannot find the quake code.

runtime/cudaq/qis/qvector.h
1tnguyen: Fixed as suggested.

PR #3827, opened 2026-01-30
Renaming the devdeps ext to have a separate image name

bettinaheim:
This is so that we can make the devdeps image internal. The extended image that serves as the base image for the dev container remains public and is now called devcontainer.


bettinaheim:
Pulling this into main, since it is quick to revert and easier to test the full flows there.

PR #3829, opened 2026-01-30
Clean up target args processing

lmondada:


python/cudaq/__init__.py
```diff
@@ -300,25 +300,44 @@ def __clearKernelRegistries():
...
+# TODO: remove this
+elif any(
+        w in ''.join(sys.argv) for w in
+    ['-target', '--target-option', '--emulate', '--cudaq-full-stack-trace']):
+    import warnings
+    warnings.warn(
+        "Will now parse command line arguments. This will be removed in a future "
+        "release, call cudaq.parse_args() explicitly to parse arguments.",
+        DeprecationWarning)
+    parse_args()
```
lmondada, line 343: reminder to myself to open an issue about this before merging in

lmondada:
These changes uncovered a bug whereby errors where not thrown properly when `--cudaq-full-stack-trace` was set. Added a test case to cover that and fixed the issue in `python/cudaq/kernel/utils.py`.

mitchdz:
FYI the new test is failing CI - https://github.com/NVIDIA/cuda-quantum/actions/runs/21677749926/job/62505072295?pr=3829#step:9:2152

python/cudaq/kernel/utils.py
```diff
@@ -248,6 +249,25 @@ def recover_value_of(name, resMod):
     raise RuntimeError("'" + name + "' is not available in this scope.")
 
 
+@contextmanager
+def set_tracebacklimit(limit=None):
+    """
+    Set the traceback limit for the duration of the context.
+    
+    Restores the original traceback limit after the context is exited.
```
khalatepradnya: nit: Use backticks here (since it refers to Python module) and drop from the allowed spelling list.

```suggestion
    Set the `traceback` limit for the duration of the context.
    
    Restores the original `traceback` limit after the context is exited.
```

python/cudaq/__init__.py
```diff
@@ -301,25 +301,45 @@ def __clearKernelRegistries():
...
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('--target', '-target', type=str, dest='target')
+    parser.add_argument('--target-option', type=str, dest='option')
+    parser.add_argument('--emulate', action='store_true', dest='emulate')
+    parser.add_argument('--cudaq-full-stack-trace',
+                        action='store_true',
+                        dest='full_stack_trace')
+
+    # Parse only known arguments to avoid errors from unrecognized options
+    args, _ = parser.parse_known_args(args)
```
khalatepradnya, line 321: Q: What about target-specific options such as machine name? 
For example, `--target quantinuum --quantinuum-machine foo`

lmondada:
>FYI the new test is failing CI

Thanks @mitchdz for the heads up. It turns out piping from the python script to `FileCheck` was not working in the LIT test? I've resolved this by using
```
# RUN: PYTHONPATH=../../ python3 %s --cudaq-full-stack-trace 2> %t; cat %t | FileCheck %s -check-prefix=FAIL
```
i.e. I store the stderr output in a temporary file and then read from that

python/cudaq/kernel/utils.py
lmondada: good idea, will do!

python/cudaq/__init__.py
lmondada: Looking at the old code, it looks like these were not supported? I've grepped through the codebase, I can't find an example of a python script taking target-specific options, other than through the `--target-option` flag. Do you have an example use case in mind?

khalatepradnya: I think you are right about it not being currently supported, and it seems that it's not what we want to support here.
I have one test which can potentially use `--target-option` but, I can update it after this PR goes in.

python/cudaq/__init__.py
```diff
@@ -301,25 +301,45 @@ def __clearKernelRegistries():
...
+
+    args = vars(args)  # convert to dict
+    args.pop('full_stack_trace', None)
+
+    cudaq_runtime.initialize_cudaq(**args)
+
+
+if __name__ == '__main__':
+    parse_args()
+# TODO: remove this
```
khalatepradnya: Is the following comment about an issue pertaining to this TODO? If so, I recommend adding the issue link here in the comment.

lmondada: I wanted to make sure we are fine with this approach before creating the issue. Will do so now and update the TODO to include the issue number :)

python/cudaq/__init__.py
lmondada: Done: https://github.com/NVIDIA/cuda-quantum/issues/3863

python/cudaq/__init__.py
```diff
@@ -301,25 +301,45 @@ def __clearKernelRegistries():
...
+
+def parse_args(args: Sequence[str] | None = None):
+    """
+    Parse command line arguments and initialize the CUDA-Q environment.
+    """
+    import argparse
+
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('--target', '-target', type=str, dest='target')
+    parser.add_argument('--target-option', type=str, dest='option')
```
khalatepradnya, line 314: One more clarifying question - the `target-option` is the target-specific config argument, right? And each target can have multiple arguments. The current behavior supports only one option, right? But, what about the target option's key? Like `quantinuum-machine=foo`.
(I think that is the current behavior and this PR retains it - which is okay).

lmondada: Yes, your understanding is correct. I have left the current behaviour unchanged, but I must admit I am not quite sure where this is used and supported.

I can see in the docs that it is used for the nvidia target
```
python3 program.py [...] --target nvidia --target-option fp32
```

but I am not sure if this is still supported.

PR #3830, opened 2026-01-30
Ergonomic changes for deployment

bettinaheim:
Deployment running here: https://github.com/NVIDIA/cuda-quantum/actions/runs/21524179333

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-.github @bettinaheim @1tnguyen @bmhowe23
+.github @bettinaheim @1tnguyen @bmhowe23 @mitchdz
 .github/workflows/config @bettinaheim @1tnguyen @bmhowe23 @schweitzpgi


.github/workflows/deployments.yml
```diff
@@ -398,6 +398,90 @@ jobs:
         run: |
           echo '${{ steps.read_json.outputs.result }}'
 
+  stitch_devdeps:
+    name: Stitch devdeps (CUDA ${{ matrix.cuda }}) images
+    needs: ext_config
```
mitchdz, line 410: This should also need `digest_config` job for the `key="amd64-cu${cuda}-dev-image"` lookup

mitchdz: You can see an example of the output of that from a recent workflow run, e.g. https://github.com/NVIDIA/cuda-quantum/actions/runs/21461041298/job/61821855321#step:3:8

.github/workflows/deployments.yml
```diff
@@ -398,9 +405,94 @@ jobs:
...
+            key="${arch}-cu${cuda}-ext"
+            image=$(echo "$EXT_JSON" | jq -r ".image_hash[\"$key\"]")
+            refs="$refs $image"
+          done
+
+          registry=$(echo "$image" | cut -d'@' -f1)
+
+          # Retrieve image tag from EXT_JSON, just look at amd64 and filter it out
+          key="amd64-cu${cuda}-dev-image"
+          image_tag=$(echo "$EXT_JSON" | jq -r --arg k "$key" '.image_tag[$k]')
```
mitchdz, line 465: and then after including `needs: digest_config`
```suggestion
          image_tag=$(echo "$DIGESTS_JSON" | jq -r --arg k "$key" '.image_tag[$k]')
```

It does feel weird here, as I'm using the dev image to get the image tag for the devdeps.

mitchdz: what I was trying to look for here was 
```
"amd64-cu13.0-dev-image": "ghcr.io/nvidia/cuda-quantum-dev:ext-amd64-cu13.0-gcc11-main",
````
Which exists in the DIGESTS_JSON. I don't think I can dynamically find that `ext-amd64-cu13.0-gcc11-main` tag (which I used to construct `ext-cu13.0-gcc11-main`) in `EXT_JSON` at the moment. Here's a list of the current values that EXT_JSON prints:

https://github.com/NVIDIA/cuda-quantum/actions/runs/21461041298/job/61821944303#step:7:20

Comment (bettinaheim) [reply .github/workflows/deployments.yml:465]: Ah, I see the problem. What I wanted to do with this is precisely to decouple the stitching of the devdeps from the full images so that I can produce the devdeps image on a nightly basis without building everything else. I have some calls now, but will get back to looking at how I work around that.

Comment (mitchdz) [reply .github/workflows/deployments.yml:465]: we can use the build_cache:

```
    "amd64-cu13.0-ext": "type=registry,ref=ghcr.io/nvidia/buildcache-cuda-quantum-devdeps-ext-cu13-0-gcc11-amd64:main",
```

It _is_ a bit messy to convert it, but can definitely do it.

```
          # Retrieve build_cache value from EXT_JSON for amd64-cu${cuda}-ext
          key="amd64-cu${cuda}-ext"
          build_cache_value=$(echo "$EXT_JSON" | jq -r --arg k "$key" '.build_cache[$k]')

          # Extract the image reference from "type=registry,ref=..." format
          # Example input: type=registry,ref=ghcr.io/nvidia/buildcache-cuda-quantum-devdeps-ext-cu12-6-gcc11-amd64:main
          build_cache_ref=$(echo "$build_cache_value" | sed 's/^type=registry,ref=//')

          # hardcode registry base
          registry_base="ghcr.io/nvidia/cuda-quantum-devdeps"

          # Extract the tag components after "buildcache-cuda-quantum-devdeps-"
          # Result: ext-cu12-6-gcc11-amd64:main
          tag_parts=$(echo "$build_cache_ref" | sed 's|.*buildcache-cuda-quantum-devdeps-||')

          # Remove architecture suffix (-amd64 or -arm64) before the colon
          # Result: ext-cu12-6-gcc11:main
          tag_without_arch=$(echo "$tag_parts" | sed 's/-\(amd64\|arm64\):/:/')

          # Convert CUDA version format from cu12-6 to cu12.6 (dash to dot)
          # Result: ext-cu12.6-gcc11:main
          tag_converted=$(echo "$tag_without_arch" | sed 's/\(cu[0-9]*\)-\([0-9]*\)/\1.\2/')

          # Move branch name into the tag by replacing colon with dash
          # Result: ext-cu12.6-gcc11-main
          tag_final=$(echo "$tag_converted" | tr ':' '-')

          # Construct final stitched tag
          # Result: ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu12.6-gcc11-main
          stitched_tag="${registry_base}:${tag_final}"

```

bettinaheim: @mitchdz I think everything looks good now. Here is the deployment - can you double check that it looks right to you, too? https://github.com/NVIDIA/cuda-quantum/actions/runs/21524179333

PR #3832, opened 2026-01-31
Add ptsbe generic implementation and early testing with 

taalexander:
### Description

This PR builds on #3823 and enables a generic PTSBE execution path that works with any `CircuitSimulatorBase` simulator. Simulators can opt-in to optimized batched execution by implementing `sampleWithPTSBE()`, or automatically use the generic fallback.

While we don't yet have the PTS construction it is otherwise mostly working with batch execution + mock trajectories (some details like the available noise operators need to be fleshed out).

## API Changes

To support external PTSBE execution, these `CircuitSimulatorBase` members were moved from `protected` to `public`:
- `GateApplicationTask` struct
- `applyGate(const GateApplicationTask&)`
- `setToZeroState()` (moved to base `CircuitSimulator` interface)

We also added a new base gate `id` (identity) to the gate registry. If you prefer we could have worked around with `rz(0)` but this makes the mapping more simple.

## Example

```cpp
// Any simulator works via generic fallback
QppCircuitSimulator<qpp::ket> sim;
PTSBatch batch = /* ... */;

// Returns per-trajectory results
auto results = executePTSBE(sim, batch);

// Aggregate for final counts
auto aggregated = aggregateResults(results);

// Custom simulators can opt-in for optimized execution
class MyGPUSimulator : public CircuitSimulatorBase<double> {
public:
  std::vector<sample_result> sampleWithPTSBE(const PTSBatch& batch) {
    // Batched GPU execution...
  }
};

MyGPUSimulator gpuSim;
auto gpuResults = executePTSBE(gpuSim, batch); // Routes to custom impl

```

### DIFF
--- a/targettests/ptsbe/sample_intercept.cpp
+++ b/targettests/ptsbe/sample_intercept.cpp
@@ -33,7 +33,7 @@ int main() {
 
   // Verify batch has captured trace
   std::size_t count = 0;
-  for (const auto &inst : batch.kernel_trace) {
+  for (const auto &inst : batch.kernelTrace) {
     (void)inst;
     ++count;
   }
@@ -43,13 +43,13 @@ int main() {
     return 1;
   }
 
-  if (batch.measure_qubits.size() != 2) {
+  if (batch.measureQubits.size() != 2) {
     printf("FAIL: Expected 2 measure qubits, got %zu\n",
-           batch.measure_qubits.size());
+           batch.measureQubits.size());
     return 1;
   }
 
   printf("PASS: PTSBatch captured with %zu instructions and %zu measure qubits\n",
-         count, batch.measure_qubits.size());
+         count, batch.measureQubits.size());
   return 0;
 }


runtime/cudaq/ptsbe/PTSBEInterface.h
```diff
@@ -45,43 +55,274 @@ concept PTSBECapable = requires(SimulatorType &sim, const PTSBatch &batch) {
...
+/// @param sel KrausSelection specifying the noise operation
+/// @return GateApplicationTask ready for simulator execution
+///
+/// TODO: Currently uses op_name string lookup as a workaround. When
+/// KrausOperatorType is expanded to include named error types (X_ERROR,
+/// Y_ERROR, Z_ERROR, etc.), this should map directly from enum to gate.
+template <typename ScalarType>
+typename nvqir::CircuitSimulatorBase<ScalarType>::GateApplicationTask
+krausSelectionToTask(const cudaq::KrausSelection &sel) {
+  using TaskType =
```
sacpis, line 130: I think we can make `TaskType` global.

1tnguyen (approved):
🚀

runtime/cudaq/ptsbe/PTSBEInterface.h
```diff
@@ -45,43 +55,274 @@ concept PTSBECapable = requires(SimulatorType &sim, const PTSBatch &batch) {
...
+/// @param simulator Circuit simulator instance
+/// @param batch PTSBE specification
+/// @return Per-trajectory sample results
+/// @throws std::runtime_error if gate conversion fails
+template <typename ScalarType>
+std::vector<cudaq::sample_result>
+executePTSBEGeneric(nvqir::CircuitSimulatorBase<ScalarType> &simulator,
+                    const PTSBatch &batch) {
+  if (batch.trajectories.empty())
+    return {};
```
lmondada, line 250: ah, saw this PR too late, see my comments on #3823

taalexander: Replied in that thread.

PR #3834, opened 2026-01-31
Moving the logger into its own library

Renaud-K:
We have several libraries explicitly  compiling Logger.cpp. 
Example:
```add_library(cudaq-mlir-runtime
  SHARED
    ArgumentConversion.cpp
    CodeGenConfig.cpp
    Environment.cpp
    JIT.cpp
    Logger.cpp      <-------------------   Prevents cudaq-mlir-runtime from being moved
    RuntimeMLIR.cpp
    RuntimeCppMLIR.cpp
    RunTheKernel.cpp
)
```

As we are looking at moving libraries out of common and inside their own directory, it becomes apparent that we need to modularize the logger first and reuse it as a library. 

cudaq-logger is now a shared library that wraps the static `libspdlog.a` 

Together with this change a new include directory tree is introduced to separate public headers from private headers. The logger's interface is at `#include "cudaq/runtime/logger/logger.h"`.

For instance, the next public header would be #include "cudaq/runtime/compiler/compiler.h"

### DIFF
--- a/docker/build/assets.Dockerfile
+++ b/docker/build/assets.Dockerfile
@@ -216,7 +216,6 @@ RUN cd /cuda-quantum && \
 # - linux_x86_64, etc.
 # If input is linux_<ARCH>, then choose manylinux_2_28_<ARCH> output
 RUN echo "Patching up wheel using auditwheel..." && \
-    cd /cuda-quantum && \
     ## [>CUDAQuantumWheel]
     CUDAQ_WHEEL="$(find . -name 'cuda_quantum*.whl')" && \
     MANYLINUX_PLATFORM="$(echo ${CUDAQ_WHEEL} | grep -o '[a-z]*linux_[^\.]*' | sed -re 's/^linux_/manylinux_2_28_/')" && \


Renaud-K:
> /ok to test 57cfeba68af84a4c535f35e709fd05eb0c056bee

Renaud-K:
> /ok to test ba97a090783e1e1fbc0d9cab2352f007affbb34d

Renaud-K:
> /ok to test 4f42685b416de27712f14698a523be3aca22a25a

Renaud-K:
> /ok to test 7270b2b2e392dc9e17540b16f2d8733c49b868b3

Renaud-K:
> /ok to test 04bb6739aad7108d895316d2b344b7bcf7e1bd4f

Renaud-K:
> /ok to test faa78a3ece03b3532e30df57218a5aeb5a19bde1

runtime/common/CMakeLists.txt
```diff
@@ -46,7 +45,8 @@ target_link_libraries(${LIBRARY_NAME}
   PUBLIC
     cudaq-operator
   PRIVATE
-    spdlog::spdlog
+    cudaq-logger
```
1tnguyen, line 48: Q: Is there a reason we don't put `cudaq-logger` in `PUBLIC` (same as `cudaq-operator`)?
Also, the comment "# Link privately to all dependencies" seems to be incorrect :)

Renaud-K: I think that it is cleaner to be explicit about dependencies. If a new module includes `logger.h` then it is reasonable to require this module to link with cudaq-logger.
If it were public, then cudaq-common would bring headers that are not part of its interface. I am not sure why we would do that. 

I expect that cudaq-common will be further broken down and the new libraries will be explicitly listed where they are needed. Then we can remove cudaq-common where it is no longer needed. It could be that some modules link with cudaq-common just for the logger.

Why is `cudaq-operator` public? I am not sure. Maybe it is a shortcut to avoid repeating it where it is needed. 

I can fix the comment. I will have to rerun the CI anyway.

Renaud-K:
> /ok to test 22b25ef381654e0f408d0ee7af6c8e9fd85dd641

Renaud-K:
> /ok to test ae31e6d35e765a35c65afc6745e256cf15775796

Renaud-K:
> /ok to test  395f885

Renaud-K:
> /ok to test 395f8857b5e72661ff21364a46c96a149429a435

Renaud-K:
> /ok to test 8f08099434365bca0d3a739cd0dc16309ef24512

docker/build/assets.Dockerfile
```diff
@@ -216,6 +216,7 @@ RUN cd /cuda-quantum && \
 # - linux_x86_64, etc.
 # If input is linux_<ARCH>, then choose manylinux_2_28_<ARCH> output
 RUN echo "Patching up wheel using auditwheel..." && \
+    cd /cuda-quantum && \
```
mitchdz: Reminder to remove this!

Renaud-K: Thanks. I undid the change. The file is no longer listed.

Renaud-K:
> /ok to test 217ad0796ee28b07ba12e9bfc266d745a8a69ee4

PR #3839, opened 2026-02-02
Add application validation for macOS cpp/python

taalexander:
## Description
- Add application validation tests for macOS C++/Python builds
- Update notebook validation script to support macOS-specific execution
- Fix notebook compatibility issues for running on macOS ARM64

### DIFF
--- a/docs/notebook_validation.py
+++ b/docs/notebook_validation.py
@@ -46,7 +46,7 @@ GPU_REQUIRED_NOTEBOOKS = [
     'afqmc.ipynb',  # AFQMC algorithm, times out on CPU
     'digitized_counterdiabatic_qaoa.ipynb',  # QAOA optimization, times out on CPU
     'qm_mm_pe.ipynb',  # VQE+SCF with many iterations, times out on CPU
-    'vqe_advanced.ipynb',  # VQE optimization with mqpu, requires multi-GPU
+    'vqe_advanced.ipynb',  # VQE optimization with `mqpu`, requires multi-GPU
     'edge_detection.ipynb',  # Requires CuPy
     'entanglement_acc_hamiltonian_simulation.ipynb',  # Requires CuPy
     'skqd.ipynb',  # Requires CuPy
--- a/docs/sphinx/applications/python/krylov.ipynb
+++ b/docs/sphinx/applications/python/krylov.ipynb
@@ -71,7 +71,11 @@
     "import scipy\n",
     "\n",
     "# Single-node, single gpu\n",
-    "cudaq.set_target(\"nvidia\")\n"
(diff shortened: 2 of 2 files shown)


docs/sphinx/applications/python/cost_minimization.ipynb
```diff
@@ -14,14 +14,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
lmondada: really super nit: put this back to `1`, given that the following cells have been executed in order.

lmondada: applies also further below. If I were you I'd just be lazy and use `git reset HEAD^` followed by `git add -p`, to only commit the parts you care about.

docs/notebook_validation.py
```diff
@@ -40,45 +40,88 @@ def read_available_backends():
 # `set_target('abc', option = 'xyz')`
 pattern = r"set_target\(\s*(\\?['\"])([^'\"]+)\1(?:\s*,\s*option\s*=\s*(\\?['\"])([^'\"]+)\3)?\)"
 
+# Platform-dependent skip list: notebooks that require GPU/CUDA libraries
+# These cannot run on CPU-only systems even with fallback logic
+GPU_REQUIRED_NOTEBOOKS = [
+    'adapt_vqe.ipynb',  # Uses state-based cudaq.observe() (GPU-only feature)
+    'adapt_qaoa.ipynb',  # Uses state-based cudaq.observe() (GPU-only feature)
+    'afqmc.ipynb',  # AFQMC algorithm, times out on CPU
+    'krylov.ipynb',  # Kernel crash on CPU, needs GPU
```
lmondada: It's actually super useful to know these don't run on CPU... If you ask me this isn't all on purpose and would make for great test cases that we should (at least in part) fix. Could you create an issue so we can track these?

lmondada (approved):

docs/notebook_validation.py
taalexander: Yes, can do. I had most of these wired up for cpu support on macos. Unfortunately, I threw away most of the information here and I did it back at the start of January so I don't recall the details on most of these...

docs/sphinx/applications/python/cost_minimization.ipynb
taalexander: Resolved.

docs/notebook_validation.py
taalexander: I now have all notebooks running. Found two relatively critical bugs that likely would have needed to be fixed before release (I don't think these were limited to macos) #3860 #3861.

PR #3844, opened 2026-02-02
Fix perf issue with adaptive solvers

sacpis:
Optimizing adaptive solvers by maintaining solver state between integration calls. It preserves
- Adaptive step size information
- RK coefficients and interpolation data
- Internal solver buffers

Also, cached dimensions list conversion, which was called on every RHS evaluation. This PR also reduces tensor creation from 3x to 1x per continuation step.

There is notable 35 - 45% (1.45x to 1.79x) speedup for dynamics simulation.

### DIFF
--- a/python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
+++ b/python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
@@ -223,25 +223,16 @@ class CUDATorchDiffEqIntegrator(BaseIntegrator[cudaq_runtime.State]):
                                                      rtol=self.rtol,
                                                      atol=self.atol,
                                                      norm=_rms_norm)
-                # Initialize solver state
-                t_init = torch.tensor([self.t, t],
-                                      device='cuda',
-                                      dtype=torch.float64)
-                self._solver_instance._before_integrate(t_init)
             else:
                 self._solver_instance.y0 = y0
                 if hasattr(self._solver_instance, 'rk_state'):
-                    t_current = torch.tensor(self.t,
-                                             device='cuda',
-                                             dtype=torch.float64)
-                    f1 = self._solver_instance.func(t_current, y0)
-                    self._solver_instance.rk_state = _RungeKuttaState(
-                        y1=y0,
-                        f1=f1,
-                        t0=t_current,
-                        t1=t_current,
-                        dt=self._solver_instance.rk_state.dt,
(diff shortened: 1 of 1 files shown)


1tnguyen:
Thanks, @sacpis.

Could we please add all those adaptive solver classes that we have a custom wrapper for in the `all_integrator_classes` list in `test_evolve_dynamics_torch_integrators.py`?

sacpis:
> Thanks, @sacpis.
> 
> Could we please add all those adaptive solver classes that we have a custom wrapper for in the `all_integrator_classes` list in `test_evolve_dynamics_torch_integrators.py`?

Added but don't you think it will take additional time for testing?

1tnguyen:
> Added but don't you think it will take additional time for testing?

I think it's worthwhile in this case, as we now have custom interaction with `torchdiffeq` for these solvers. 
With the tests, we can detect any potential breaking changes in `torchdiffeq` in the future.

We can trim down the list based on common integration classes, e.g., just Dopri5 to represent all Dopri solvers.

python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
```diff
@@ -164,33 +213,72 @@ def integrate(self, t):
...
+                                             device='cuda',
+                                             dtype=torch.float64)
+                    f1 = self._solver_instance.func(t_current, y0)
+                    self._solver_instance.rk_state = _RungeKuttaState(
+                        y1=y0,
+                        f1=f1,
+                        t0=t_current,
+                        t1=t_current,
+                        dt=self._solver_instance.rk_state.dt,
+                        interp_coeff=[y0] * 5)
```
1tnguyen: I wonder why we need this step. It looks like `_before_integrate` should have set this up for us.

sacpis: The [y0] * 5 is used as a placeholder/initialization value because no actual step has been taken yet (t0 == t1). The real interp_coeff will be computed by `_interp_fit` after the first actual RK step.

sacpis: Thanks @1tnguyen. Fixed in [15e142b](https://github.com/NVIDIA/cuda-quantum/pull/3844/commits/15e142b19c6bc81372dbc4ec3b4b5b626eea2869).

PR #3845, opened 2026-02-03
Removing blank after backslash, causing the library path to not be provided to auditwheel.

Renaud-K:


Renaud-K:
> /ok to test 7b11dd26fe5b94d9dd4460b33a63867d3254b33e

mitchdz:
Thanks for this, this is a pretty nasty hidden bug. For more info if you are confused, refer to https://www.gnu.org/software/bash/manual/bash.html#Escape-Character

PR #3848, opened 2026-02-03
[PTSBE] Adding sampling strategies

sacpis:
Adding the following sampling strategies for PTSBE

- Probabilistic Sampling
- Exhaustive Sampling
- Ordered Sampling
- Conditional Sampling

### DIFF
--- a/runtime/cudaq/ptsbe/PTSBESampleIntegration.h
+++ b/runtime/cudaq/ptsbe/PTSBESampleIntegration.h
@@ -105,9 +105,7 @@ runSamplingPTSBE(KernelFunctor &&wrappedKernel, quantum_platform &platform,
 
   // Stage 0: Capture trace via ExecutionContext("tracer")
   ExecutionContext traceCtx("tracer");
-  platform.set_exec_ctx(&traceCtx);
-  wrappedKernel();
-  platform.reset_exec_ctx();
+  platform.with_execution_context(traceCtx, [&]() { wrappedKernel(); });
 
   // Stage 1: Validate kernel eligibility (no dynamic circuits)
   validatePTSBEKernel(kernelName, traceCtx);
@@ -139,9 +137,8 @@ template <typename QuantumKernel, typename... Args>
 PTSBatch capturePTSBatch(QuantumKernel &&kernel, Args &&...args) {
   ExecutionContext traceCtx("tracer");
   auto &platform = get_platform();
-  platform.set_exec_ctx(&traceCtx);
-  kernel(std::forward<Args>(args)...);
-  platform.reset_exec_ctx();
+  platform.with_execution_context(
+      traceCtx, [&]() { kernel(std::forward<Args>(args)...); });
 
   throwIfMidCircuitMeasurements(traceCtx);
 
--- a/runtime/cudaq/ptsbe/PTSBESampler.cpp
+++ b/runtime/cudaq/ptsbe/PTSBESampler.cpp
@@ -29,10 +29,10 @@ template <typename ScalarType>
 std::vector<cudaq::sample_result>
 samplePTSBEGeneric(nvqir::CircuitSimulatorBase<ScalarType> &simulator,
                    const PTSBatch &batch) {
-  if (!simulator.getExecutionContext())
+  if (!cudaq::getExecutionContext())
     throw std::runtime_error(
         "samplePTSBEGeneric requires ExecutionContext to be set. "
-        "Call simulator.setExecutionContext() before invoking.");
+        "Use cudaq::detail::setExecutionContext() before invoking.");
 
   if (batch.trajectories.empty())
     return {};
@@ -116,7 +116,8 @@ samplePTSBEWithLifecycle(const PTSBatch &batch,
   auto *sim = nvqir::getCircuitSimulatorInternal();
 
   cudaq::ExecutionContext ctx(contextType, batch.totalShots());
-  sim->setExecutionContext(&ctx);
+  cudaq::detail::setExecutionContext(&ctx);
+  sim->configureExecutionContext(ctx);
   sim->allocateQubits(batch.kernelTrace.getNumQudits());
 
   auto results = samplePTSBE(batch);
@@ -124,7 +125,8 @@ samplePTSBEWithLifecycle(const PTSBatch &batch,
   std::vector<std::size_t> qubitIds(batch.kernelTrace.getNumQudits());
   std::iota(qubitIds.begin(), qubitIds.end(), 0);
   sim->deallocateQubits(qubitIds);
-  sim->resetExecutionContext();
+  sim->finalizeExecutionContext(ctx);
+  cudaq::detail::resetExecutionContext();
 
   return results;
 }
--- a/unittests/ptsbe/ExecutePTSBETester.cpp
+++ b/unittests/ptsbe/ExecutePTSBETester.cpp
@@ -28,7 +28,8 @@ template <typename SimulatorType>
 std::vector<cudaq::sample_result> runPTSBETest(SimulatorType &sim,
                                                const PTSBatch &batch) {
   cudaq::ExecutionContext ctx("sample", batch.totalShots());
-  sim.setExecutionContext(&ctx);
+  cudaq::detail::setExecutionContext(&ctx);
+  sim.configureExecutionContext(ctx);
   sim.allocateQubits(batch.kernelTrace.getNumQudits());
 
   std::vector<cudaq::sample_result> results;
@@ -41,7 +42,8 @@ std::vector<cudaq::sample_result> runPTSBETest(SimulatorType &sim,
   std::vector<std::size_t> qubitIds(batch.kernelTrace.getNumQudits());
   std::iota(qubitIds.begin(), qubitIds.end(), 0);
   sim.deallocateQubits(qubitIds);
-  sim.resetExecutionContext();
+  sim.finalizeExecutionContext(ctx);
+  cudaq::detail::resetExecutionContext();
   return results;
 }
 


runtime/cudaq/ptsbe/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@
 # Source files
 set(PTSBE_SOURCES
     ShotAllocationStrategy.cpp
+    ProbabilisticSamplingStrategy.cpp
+    ExhaustiveSamplingStrategy.cpp
+    OrderedSamplingStrategy.cpp
+    ConditionalSamplingStrategy.cpp
```
taalexander: These might be able to be consolidated to a single translation unit (or migrated to a separate directory if we wish to keep separate).

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  std::size_t total_trajectories = 1;
+  std::vector<std::size_t> operator_counts;
+  operator_counts.reserve(noise_points.size());
+
+  for (const auto &noise_point : noise_points) {
+    std::size_t count = noise_point.kraus_operators.size();
+    operator_counts.push_back(count);
+    total_trajectories *= count;
+  }
```
taalexander: What is the intention behind total_trajectories? It seems like this might be the combinatoric number of possible trajectories? This will grow very very fast.

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  for (std::size_t candidate_id = 0;
+       candidate_id < total_trajectories && results.size() < max_trajectories;
+       ++candidate_id) {
+
+    std::vector<KrausSelection> selections;
+    selections.reserve(noise_points.size());
+    double probability = 1.0;
+
+    for (std::size_t i = 0; i < noise_points.size(); ++i) {
```
taalexander, line 46: Can you please go over how this works? It seems like we are grabbing the Kraus operator type to be from 0..number of noise points. I think we need to sample from the finite number of Kraus operators supported per-noise operator? Where do we make the concrete noise selection (eg. convert it to unitary noise). 

I could be missing this in the implementation right now so please correct me if I'm wrong.

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
```diff
@@ -0,0 +1,83 @@
...
+                          .setSelections(std::move(selections))
+                          .setProbability(probability)
+                          .build();
+
+    if (predicate_(trajectory)) {
+      results.push_back(std::move(trajectory));
+      trajectory_id++;
+    }
+
+    for (std::size_t i = 0; i < indices.size(); ++i) {
```
taalexander, line 64: What does this do? Is this how we increment the current noise selection? It seems like it might be incrementing all noise operators for all noise points each round (I could be wrong here, it's a little hard to read).

runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.h
```diff
@@ -0,0 +1,55 @@
...
+ ******************************************************************************/
+
+#pragma once
+
+#include "PTSSamplingStrategy.h"
+#include <functional>
+
+namespace cudaq::ptsbe {
+
+/// @brief Predicate function type for filtering trajectories
```
taalexander, line 18: I worry that since we use lexicographic order we are going to bias the noise model to trajectories with errors on the first operations. I imagine predicated probabilistic would be most interesting but I could be wrong here.

runtime/cudaq/ptsbe/strategies/ProbabilisticSamplingStrategy.cpp
```diff
@@ -0,0 +1,70 @@
...
+    return results;
+  }
+
+  std::set<std::vector<std::size_t>> seen_patterns;
+
+  std::size_t trajectory_id = 0;
+  std::size_t max_attempts = max_trajectories * 100;
+  std::size_t attempts = 0;
+
+  while (results.size() < max_trajectories && attempts < max_attempts) {
```
taalexander, line 44: Minor optimization but we should be able to compute the total number of possible trajectories and limit if we have reached that many trajectories (in the exhaustive case).

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
```diff
@@ -0,0 +1,92 @@
...
+  std::vector<double> probabilities;
+
+  /// @brief Check if this is a unitary mixture channel
+  /// @return true if all probabilities sum to ~1.0
+  [[nodiscard]] bool isUnitaryMixture() const {
+    double sum = 0.0;
+    for (auto p : probabilities)
+      sum += p;
+    return std::abs(sum - 1.0) < 1e-9;
+  }
```
taalexander: This alone isn't sufficient. We need to verify our Kraus operators are all unitary operations themselves. There are methods in NoiseModel.cpp for this. Are we able to reuse?

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/// @brief Base class for trajectory sampling strategies
+/// The sampling strategy receives processed noise information from the engine
+/// and returns a list of unique trajectories to execute.
+class PTSSamplingStrategy {
```
taalexander, line 76: Do we require all of the constructors/assignments below?

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
sacpis: Let me add safeguards here.

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
sacpis: Added algorithm in the doc. Please let me know if you have any questions.

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
sacpis: Addressed in the latest commits.

runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.h
sacpis: Addressed in the latest commits.

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
sacpis: Had to expose those methods in NoiseModel (were static before). Addressed in the latest commits.

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
sacpis: For now, we can keep it. I think I might need it for PTSSamplingEngine.

runtime/cudaq/ptsbe/ConditionalSamplingStrategy.cpp
sacpis: Thanks for catching this one. Seems like this was an accidental addition. Removed it.

runtime/cudaq/ptsbe/CMakeLists.txt
sacpis: As it is already in the ptsbe directory, do we need to create a sub-directory for strategy (maybe)?

runtime/cudaq/ptsbe/strategies/ProbabilisticSamplingStrategy.cpp
sacpis: Addressed in [d789bba](https://github.com/NVIDIA/cuda-quantum/pull/3848/commits/d789bba310ce543e4ffe6313af14b58b42b5470f)

runtime/cudaq/ptsbe/CMakeLists.txt
sacpis: Addressed in 137c73e

runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.h
```diff
@@ -0,0 +1,82 @@
...
+
+  /// @brief Clone this strategy
+  /// @return Unique pointer to a copy
+  [[nodiscard]] std::unique_ptr<PTSSamplingStrategy> clone() const override {
+    return std::make_unique<ConditionalSamplingStrategy>(*this);
+  }
+
+private:
+  TrajectoryPredicate predicate_;
+  mutable std::mt19937_64 rng_;
```
1tnguyen, line 84: We should figure out how to seed this random engine with CUDAQ random seed (if set).

runtime/cudaq/ptsbe/strategies/OrderedSamplingStrategy.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  if (noise_points.empty()) {
+    return results;
+  }
+
+  std::size_t total_trajectories = 1;
+  std::vector<std::size_t> operator_counts;
+  operator_counts.reserve(noise_points.size());
+
+  for (const auto &noise_point : noise_points) {
```
1tnguyen, line 33: nit: this code here looks quite similar to the exhaustive sampling method.

runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.h
sacpis: Fixed in [dbc4f9b](https://github.com/NVIDIA/cuda-quantum/pull/3848/commits/dbc4f9b5eded6436b230a72ca1b3dea55927381c).

runtime/cudaq/ptsbe/strategies/OrderedSamplingStrategy.cpp
sacpis: Fixed in [dbc4f9b](https://github.com/NVIDIA/cuda-quantum/pull/3848/commits/dbc4f9b5eded6436b230a72ca1b3dea55927381c).

runtime/cudaq/ptsbe/strategies/ProbabilisticSamplingStrategy.cpp
```diff
@@ -27,8 +27,14 @@ ProbabilisticSamplingStrategy::generateTrajectories(
...
 
+  // Total possible unique trajectories with overflow protection
+  // If we want more trajectories than exist, cap at total to avoid infinite
+  // sampling
+  std::size_t total_possible = computeTotalTrajectories(noise_points);
+  std::size_t actual_target = std::min(max_trajectories, total_possible);
+
   std::size_t trajectory_id = 0;
-  std::size_t max_attempts = max_trajectories * 100;
+  std::size_t max_attempts = actual_target * 10;
```
taalexander: Not blocking. Could move `10` out to constant or parameter.

runtime/cudaq/ptsbe/PTSSamplingStrategy.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+  if (!validated_result.has_value())
+    return false;
+
+  const auto &validated_probs = validated_result->first;
+
+  for (std::size_t i = 0; i < probabilities.size(); ++i) {
+    if (std::abs(probabilities[i] - validated_probs[i]) > tolerance)
+      return false;
+  }
```
taalexander: I think we can drop this. We can also drop the initial probability check at the top if we want to force the trace preserving property (likely a good idea for now) to re-scale all probabilities to 1. (computeUnitaryMixture already does this, which likely means the noise model already does this).

runtime/cudaq/ptsbe/strategies/ProbabilisticSamplingStrategy.cpp
sacpis: Addressed in [bf566e8](https://github.com/NVIDIA/cuda-quantum/pull/3848/commits/bf566e8f99e88f9bf588958854472e623e12e651).

runtime/cudaq/ptsbe/PTSSamplingStrategy.cpp
sacpis: [bf566e8](https://github.com/NVIDIA/cuda-quantum/pull/3848/commits/bf566e8f99e88f9bf588958854472e623e12e651)

taalexander (approved):

PR #3851, opened 2026-02-03
[PTSBE] Add support for async/broadcasting

taalexander:
### Description
Adds PTSBE sample support to also support async/argument broadcasting. Refactors the implementation to move most implementations to cpp files and updates the calling conventions. etc to handle context in a more centralized fashion.

### DIFF
--- a/.github/pre-commit/spelling_allowlist_cxx.txt
+++ b/.github/pre-commit/spelling_allowlist_cxx.txt
@@ -1,8 +1,5 @@
-async
 clang
-deallocating
 gcc
-lifecycle
 nullary
 pre
 typedef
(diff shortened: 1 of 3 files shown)


.github/pre-commit/spelling_allowlist_cxx.txt
```diff
@@ -1,5 +1,8 @@
+async
 clang
+deallocating
 gcc
+lifecycle
```
sacpis: I think we should use backticks (``) around these words in the comments, instead of adding to the aloowlist.

taalexander: I'll hyphenate de-allocating/life-cycle and back-tick `async`.

PR #3860, opened 2026-02-04
Fix `cudaq.control(...)` type specialization bug

taalexander:
### Description

Fix a bug where controlled kernel calls failed with type mismatch errors when the kernel had both float and list[complex] parameters and accessed .real on complex values. Required to run the Krylov notebook in #3839.

The bug occurred during constant propagation in ApplySpecialization when veq argument types were specialized (un-relaxed from !quake.veq<?> to !quake.veq<N>) but inner func.call operations still expected the dynamic type, causing "'func.call' op operand type mismatch" errors.

The fix adds hasCallWithDynamicVeq() to check if a function has any inner calls expecting !quake.veq<?>, and skips veq type specialization in those cases to avoid the type mismatch.

### DIFF
--- a/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
+++ b/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
@@ -76,23 +76,19 @@ using ApplyOpAnalysisInfo = DenseMap<Operation *, ApplyVariants>;
 ///    accept the concrete veq size, propagating type info deeper for better
 ///    optimization but increasing code size.
 static bool hasCallWithDynamicVeq(func::FuncOp func, ModuleOp module) {
-  bool found = false;
-  func.walk([&](func::CallOp callOp) {
-    if (found)
-      return;
+  auto result = func.walk([&](func::CallOp callOp) {
     auto callee = module.lookupSymbol<func::FuncOp>(callOp.getCallee());
     if (!callee)
-      return;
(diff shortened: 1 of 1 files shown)


lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -63,6 +63,38 @@ struct ApplyVariants {
...
+    if (found)
+      return;
+    auto callee = module.lookupSymbol<func::FuncOp>(callOp.getCallee());
+    if (!callee)
+      return;
+    for (auto inputTy : callee.getFunctionType().getInputs()) {
+      if (auto veqTy = dyn_cast<quake::VeqType>(inputTy)) {
+        if (!veqTy.hasSpecifiedSize()) {
+          found = true;
+          return;
```
lmondada: I think you can return
```
```
and then you wouldn't need the `if (found) return` clause at the beginning.

I'm assuming you're in the same case as e.g. https://github.com/NVIDIA/cuda-quantum/blob/7fa01169b5ebd08a083227fe06de04da24c956aa/lib/Optimizer/CodeGen/VerifyNVQIRCalls.cpp#L61

taalexander: Fixed in 58e63cc

PR #3866, opened 2026-02-04
Bump cuquantum version

sacpis:
Bumping 
    cuquantum version to 26.1.0
    cudensitymat version to 0.4.0
    cuStateVc to 1.12.0
    cuTensorNet to 2.11.0


runtime/nvqir/cudensitymat/mpi_support.cpp
```diff
@@ -108,6 +109,10 @@ int cudensitymatMpiTestRequest(cudensitymatDistributedRequest_t request,
   return getMpiPluginInterface()->TestRequest(request, completed);
 }
 
+int cudensitymatMpiGroupStart() { return 0; }
```
1tnguyen: Could you please add a comment here that these 2 are only for NCCL?

sacpis: Fixed in [8b49935](https://github.com/NVIDIA/cuda-quantum/pull/3866/commits/8b49935d11bbedf148cd773d1420d7e22ec34764).

sacpis:
Please hold on...

sacpis:
Confirmed that it builds and installs on my machine.

PR #3870, opened 2026-02-05
[PTSBE] Adding NoiseExtractor to extract noise from a circuit trace

sacpis:
Adding the `NoiseExtractor` component for PTSBE, which extracts and validates noise sites from circuit traces given a noise model.

### DIFF
--- a/unittests/ptsbe/NoiseExtractorTester.cpp
+++ b/unittests/ptsbe/NoiseExtractorTester.cpp
@@ -24,15 +24,6 @@ cudaq::Trace createSimpleCircuit() {
   return trace;
 }
 
-cudaq::Trace createParameterizedCircuit() {
-  cudaq::Trace trace;
-  trace.appendInstruction("rx", {0.5}, {}, {cudaq::QuditInfo(2, 0)});
-  trace.appendInstruction("ry", {1.0}, {}, {cudaq::QuditInfo(2, 1)});
-  trace.appendInstruction("z", {}, {cudaq::QuditInfo(2, 0)},
-                          {cudaq::QuditInfo(2, 1)});
-  return trace;
-}
-
 TEST(NoiseExtractorTest, EmptyCircuit) {
   cudaq::Trace empty_trace;
   cudaq::noise_model noise_model;
@@ -256,39 +247,6 @@ TEST(NoiseExtractorTest, GracefulValidation_ValidChannels) {
   EXPECT_TRUE(result.all_unitary_mixtures);
 }
 
-TEST(NoiseExtractorTest, ProbabilitiesSumToOne) {
-  auto trace = createSimpleCircuit();
-  cudaq::noise_model noise_model;
-  noise_model.add_channel("h", {0}, cudaq::depolarization_channel(0.01));
-
-  auto result = extractNoiseSites(trace, noise_model);
-
-  ASSERT_EQ(result.noise_sites.size(), 1);
-
-  const auto &probs = result.noise_sites[0].channel.probabilities;
-  double sum = 0.0;
-  for (auto p : probs) {
-    sum += p;
-  }
-
-  EXPECT_NEAR(sum, 1.0, 1e-6);
-}
-
-TEST(NoiseExtractorTest, AllProbabilitiesNonNegative) {
-  auto trace = createSimpleCircuit();
-  cudaq::noise_model noise_model;
-  noise_model.add_channel("h", {0}, cudaq::depolarization_channel(0.05));
-
-  auto result = extractNoiseSites(trace, noise_model);
-
-  ASSERT_EQ(result.noise_sites.size(), 1);
-
-  for (auto p : result.noise_sites[0].channel.probabilities) {
-    EXPECT_GE(p, 0.0);
-    EXPECT_LE(p, 1.0);
-  }
-}
-
 TEST(NoiseExtractorTest, SingleQubitGateTracksQubit) {
   auto trace = createSimpleCircuit();
   cudaq::noise_model noise_model;
@@ -320,38 +278,25 @@ TEST(NoiseExtractorTest, TwoQubitGateTracksQubits) {
   EXPECT_TRUE(has_qubit_0 || has_qubit_1);
 }
 
-TEST(NoiseExtractorTest, ToleranceParameter) {
-  auto trace = createSimpleCircuit();
-  cudaq::noise_model noise_model;
-  noise_model.add_channel("h", {0}, cudaq::depolarization_channel(0.01));
-
-  auto result1 = extractNoiseSites(trace, noise_model, true);
-  auto result2 = extractNoiseSites(trace, noise_model, true);
-
-  EXPECT_EQ(result1.noise_sites.size(), result2.noise_sites.size());
-  EXPECT_TRUE(result1.all_unitary_mixtures);
-  EXPECT_TRUE(result2.all_unitary_mixtures);
-}
-
 TEST(NoiseExtractorTest, LargeCircuit) {
-  cudaq::Trace trace;
+  constexpr std::size_t NUM_QUBITS = 10;
+  constexpr std::size_t NUM_GATES = 100;
 
-  for (int i = 0; i < 100; ++i) {
-    trace.appendInstruction(
-        "h", {}, {}, {cudaq::QuditInfo(2, static_cast<std::size_t>(i % 10))});
+  cudaq::Trace trace;
+  for (std::size_t i = 0; i < NUM_GATES; ++i) {
+    trace.appendInstruction("h", {}, {}, {cudaq::QuditInfo(2, i % NUM_QUBITS)});
   }
 
   cudaq::noise_model noise_model;
-
-  for (std::size_t q = 0; q < 10; ++q) {
+  for (std::size_t q = 0; q < NUM_QUBITS; ++q) {
     noise_model.add_channel("h", {q}, cudaq::depolarization_channel(0.01));
   }
 
   auto result = extractNoiseSites(trace, noise_model);
 
-  EXPECT_EQ(result.total_instructions, 100);
-  EXPECT_EQ(result.noise_sites.size(), 100);
-  EXPECT_EQ(result.noisy_instructions, 100);
+  EXPECT_EQ(result.total_instructions, NUM_GATES);
+  EXPECT_EQ(result.noise_sites.size(), NUM_GATES);
+  EXPECT_EQ(result.noisy_instructions, NUM_GATES);
   EXPECT_TRUE(result.all_unitary_mixtures);
 }
 


runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+ ******************************************************************************/
+
+#include "NoiseExtractor.h"
+#include <algorithm>
+#include <sstream>
+
+namespace cudaq::ptsbe {
+
+/// @brief Context information about an instruction being processed
+struct InstructionContext {
```
taalexander: To a large degree this information is similar to what is contained in Trace.  IMO, we need to decide on what is the "representation" we are working with through the PTSBE section to avoid multiple duplicates.

Eg.:
- Should this replace the trace (for example when we do trace to task conversion) and be what we return to the user?
- Would it be just as easy to wrap the trace instruction in a tuple `instr_ctx = (idx, traceinstr)` to avoid duplications, etc.
- If so we might also need to include the parameters?

Let's have a quick thought about this in the context of:
- What goes back to the user
- How do we evolve to MCM at the very least

I don't think this is blocking at the moment.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+  return ctx;
+}
+
+/// @brief Convert Kraus operators from cudaq::complex to std::complex<double>
+///
+/// @param channel Noise channel with Kraus operators
+/// @return Vector of Kraus operator matrices in std::complex<double> format
+static std::vector<std::vector<std::complex<double>>>
+convertKrausOperators(const cudaq::kraus_channel &channel) {
```
taalexander: Related to the question above about return types. If we are consuming a raw Kraus channel matrix form we do not have a string pseudonym to pass back to the user so the trajectories must be in the form of their underlying matrices. However, the kraus channel class has a `noise_model_type` enum and all string conversions which would be very useful.

I wonder if we should carry them around in the form of a kraus_channel since it is already a container available to us.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+/// @brief Compute probabilities and normalized operators for a unitary mixture
+///
+/// Uses computeUnitaryMixture to validate and normalize Kraus operators.
+/// Returns nullopt if the channel is not a valid unitary mixture.
+///
+/// @param kraus_operators Raw Kraus operator matrices
+/// @param tolerance Numerical tolerance for validation
+/// @return Pair of (probabilities, normalized_operators) or nullopt
+static std::optional<std::pair<std::vector<double>,
+                               std::vector<std::vector<std::complex<double>>>>>
```
taalexander: Seems like the roll of this is very similar to convertKrausOperators above (assuming we have a unitary mixture).

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+      noise_point.kraus_operators = std::move(unitary_result->second);
+    } else {
+      // Not a valid unitary mixture
+      is_valid_unitary_mixture = false;
+
+      // Assign uniform probabilities as fallback (for non-validating mode)
+      noise_point.probabilities.resize(noise_point.kraus_operators.size(),
+                                       1.0 /
+                                           noise_point.kraus_operators.size());
+    }
```
taalexander: IMO, this should be an error for PTSBE and it's current implementation.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+    noise_point.kraus_operators = channel.unitary_ops;
+    noise_point.probabilities = channel.probabilities;
+  } else {
+    // Convert Kraus operators and compute unitary mixture
+    noise_point.kraus_operators = convertKrausOperators(channel);
+
+    auto unitary_result =
+        computeUnitaryMixtureData(noise_point.kraus_operators, tolerance);
+
+    if (unitary_result.has_value()) {
```
taalexander: Are you able to just call `generateUnitaryParameters` on the noise channel and avoid managing this yourself? (before the `if` predicate abovek)

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+createNoisePointFromChannel(const cudaq::kraus_channel &channel,
+                            const InstructionContext &ctx, double tolerance) {
+  NoisePoint noise_point;
+  noise_point.circuit_location = ctx.index;
+  noise_point.op_name = ctx.name;
+
+  noise_point.qubits = ctx.target_qubits;
+  noise_point.qubits.insert(noise_point.qubits.end(),
+                            ctx.control_qubits.begin(),
+                            ctx.control_qubits.end());
```
taalexander: I am increasingly thinking we should pass the `kraus_channel` itself into noise_point once it is validated. It is already a public type for the user, contains all of the information we require and has python bindings.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+    if (validate_unitary_mixture) {
+      std::ostringstream msg;
+      msg << "Noise channel validation failed for gate '" << ctx.name
+          << "' at instruction " << ctx.index
+          << ". Channel does not satisfy unitary mixture properties.";
+      throw std::invalid_argument(msg.str());
+    }
+    return false;
+  }
+  return true;
```
taalexander: It seems like if we're limiting to unitary noise channels this should be caught earlier and is redundant.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
taalexander: Update - see comment below about using `kraus_channel` as the core type.

unittests/ptsbe/NoiseExtractorTester.cpp
```diff
@@ -0,0 +1,615 @@
...
+
+  auto result = extractNoiseSites(trace, noise_model);
+
+  ASSERT_EQ(result.noise_sites.size(), 1);
+
+  for (auto p : result.noise_sites[0].probabilities) {
+    EXPECT_GE(p, 0.0);
+    EXPECT_LE(p, 1.0);
+  }
+}
```
taalexander: If we used Kraus channel directly this would be guaranteed by our earlier unitary testing.

unittests/ptsbe/NoiseExtractorTester.cpp
```diff
@@ -0,0 +1,615 @@
...
+  auto trace = createSimpleCircuit();
+  cudaq::noise_model noise_model;
+  noise_model.add_channel("h", {0}, cudaq::depolarization_channel(0.01));
+
+  auto result1 = extractNoiseSites(trace, noise_model, true, 1e-6);
+  auto result2 = extractNoiseSites(trace, noise_model, true, 1e-9);
+
+  EXPECT_EQ(result1.noise_sites.size(), result2.noise_sites.size());
+  EXPECT_TRUE(result1.all_unitary_mixtures);
+  EXPECT_TRUE(result2.all_unitary_mixtures);
```
taalexander: What is this for? Possibly a duplicate of testing in Kraus channel if we use.

unittests/ptsbe/NoiseExtractorTester.cpp
```diff
@@ -0,0 +1,615 @@
...
+  for (std::size_t q = 0; q < 10; ++q) {
+    noise_model.add_channel("h", {q}, cudaq::depolarization_channel(0.01));
+  }
+
+  auto result = extractNoiseSites(trace, noise_model);
+
+  EXPECT_EQ(result.total_instructions, 100);
+  EXPECT_EQ(result.noise_sites.size(), 100);
+  EXPECT_EQ(result.noisy_instructions, 100);
+  EXPECT_TRUE(result.all_unitary_mixtures);
```
taalexander: This is a bit confusing could use some constants  to define number of qubits/gates and the relationship.

unittests/ptsbe/NoiseExtractorTester.cpp
```diff
@@ -0,0 +1,615 @@
...
+  cudaq::noise_model noise_model;
+
+  noise_model.add_channel("h", {0}, cudaq::bit_flip_channel(0.01));
+  noise_model.add_channel("h", {0}, cudaq::phase_flip_channel(0.01));
+
+  auto result = extractNoiseSites(trace, noise_model);
+
+  EXPECT_GE(result.noise_sites.size(), 1);
+  EXPECT_TRUE(result.all_unitary_mixtures);
+}
```
taalexander: What is the role of these "modular design" tests? They seem quite similar if not redundant to earlier tests above.

I think the same holds for integration tests below as well.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+      // Track unitary mixture status
+      if (!is_valid_unitary) {
+        result.all_unitary_mixtures = false;
+        if (validate_unitary_mixture) {
+          throwUnitaryMixtureError(ctx);
+        }
+      }
+
+      // Validation of the constructed NoisePoint
+      bool point_valid = validateNoisePoint(
```
taalexander: I think this is redundant.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+          noise_point, ctx, validate_unitary_mixture, tolerance);
+      if (!point_valid) {
+        result.all_unitary_mixtures = false;
+      }
+
+      result.noise_sites.push_back(std::move(noise_point));
+      result.noisy_instructions++;
```
taalexander: This is updating for each channel and not once-per instruction. I think this will overcount.

unittests/ptsbe/NoiseExtractorTester.cpp
```diff
@@ -0,0 +1,615 @@
...
+  trace.appendInstruction("mz", {}, {}, {{0, 2}});
+
+  // Measure qubit 1
+  trace.appendInstruction("mz", {}, {}, {{1, 2}});
+
+  return trace;
+}
+
+// A circuit with parameterized gates
+cudaq::Trace createParameterizedCircuit() {
```
taalexander: Is this used?

runtime/cudaq/ptsbe/NoiseExtractor.cpp
```diff
@@ -0,0 +1,237 @@
...
+      noise_point.probabilities = std::move(unitary_result->first);
+      noise_point.kraus_operators = std::move(unitary_result->second);
+    } else {
+      // Not a valid unitary mixture
+      is_valid_unitary_mixture = false;
+
+      // Assign uniform probabilities as fallback (for non-validating mode)
+      noise_point.probabilities.resize(noise_point.kraus_operators.size(),
+                                       1.0 /
+                                           noise_point.kraus_operators.size());
```
taalexander: What happens if the number of operators is 0? Division by 0?

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: >What goes back to the user

User gets noiseExtractionResult with vector<NoisePoint>. Each NoisePoint contains the validated kraus_channel. The channel already has all the needed info (ops, probabilities, type).

>How do we evolve to MCM at the very least

I think MCM support will require conditional branches in trace. NoiseExtractor will have to handle branching logic. I think we can extend trace processing design later when we introduce MCM. 

Would you like to add/update anything here?

sacpis: >To a large degree this information is similar to what is contained in Trace.

The InstructionContext was created for internal processing. I think we can remove this struct and wrap existing trace instruction in a pair of std::pair<size_t, const Trace::Instruction&>.

sacpis: > Would it be just as easy to wrap the trace instruction in a tuple instr_ctx = (idx, traceinstr) to avoid duplications, etc.

Makes sense.

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Ah, this is added when I had a giant function which I broke into several functions. Let me check if these tests are repeated,

runtime/cudaq/ptsbe/NoiseExtractor.cpp
taalexander: Thanks, I think I have spent a little while thinking about what will go back to the user today. I should have a PR this evening or tomorrow morning to show the end-to-end flow

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Actually forgot to remove this with the test using it. Removed in the latest commit.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: We no longer have a fallback that resizes probabilities from kraus_operators or does 1.0 / size().

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Done. We use kraus_channel as the core type. NoisePoint holds a cudaq::kraus_channel channel.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Fixed.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Done.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Fixed.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Fixed. We now count once per instruction. An instruction_has_noise flag is set inside the channel loop, and result.noisy_instructions++ runs only once per instruction.

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Removed irrelevant tests.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Addressed. We no longer have convertKrausOperators or computeUnitaryMixtureData. We use kraus_channel as the single source of truth.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Done. We now pass the validated kraus_channel into the noise point.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Agreed. We only validate once now.

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Addressed. We store the kraus_channel on NoisePoint and removed the ProbabilitiesSumToOne and AllProbabilitiesNonNegative tests as redundant with unitary validation.

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Removed the ToleranceParameter test.

unittests/ptsbe/NoiseExtractorTester.cpp
sacpis: Addressed.

runtime/cudaq/ptsbe/NoiseExtractor.cpp
sacpis: Thank you.

PR #3874, opened 2026-02-05
[nvq++] Remove kernel bodies in fixup-linkage

lmondada:
Currently, in MLIR mode kernel bodies are compiled by clang but never called: the MLIR-compiled kernels always shadow the LLVM-compiled ones. This PR removes these kernel bodies in the LLVM IR, when in MLIR mode. This will speed-up compilation and (in MLIR mode) will allow us to remove definitions of C++ types and functions that are only used within kernels (e.g. `cudaq::qvector`).

This PR is made of 2 commits, which can be reviewed independently:
- the first commit rewrites the fixup-linkage script using LLVM instead of ad-hoc string/regexp parsing of `.ll` files. The rewritten script is **equivalent** to the original
- the second commit uses the parsed LLVM module to replace the body of each kernel function with a body made of a single `unreachable` op. If this op is ever reached, the program would crash, but because the program control flow is always redirected to the MLIR kernels, this can never happen.

### DIFF
--- a/tools/fixup-linkage/fixup-linkage.cpp
+++ b/tools/fixup-linkage/fixup-linkage.cpp
@@ -6,12 +6,19 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
-/// The fixup-linkage tool is used to rewrite the LLVM IR produced by clang for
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
👍

PR #3876, opened 2026-02-05
Removing cudaq-mlir-runtime dependency on cudaq

Renaud-K:
The dependency was broken by splitting implementation across module boundaries for 2 files: `JIT.cpp` and `RunTheKernel.cpp`

1. JIT.cpp
`invokeWrappedKernel` in `JIT.cpp` was doing 2 things. It was creating and launching a wrapped kernel. This was broken up by
   - creating the kernel, in `cudaq-mlir-runtime` which needs MLIR but returns a kernel as a lambda with no MLIR dependencies. By creating the kernel on this side of the interface we no longer need to provide platform and execution context, which eliminates the `cudaq` dependency.
   - invoking the wrapped kernel with the provided lambda on the `RestRemoveServer.cpp` side. This side provides the platform and execution context.   

2. RunTheKernel.cpp
RunTheKernel is an algorithm implementation that is shared by python and C++. This is a good thing and we should be consolidating more algorithm similarly moving forward. Curiously, however, this file is located in `cudaq-mlir-runtime`. This is so it can call `getLayoutInfo` which is MLIR dependent. So the fix is to keep `getLayoutInfo` on the `cudaq-mlir-runtime` side and move the rest of RunTheKernel.cpp to the side of algorithms. 
For C++, `run` calls `getLayoutInfo` in templated code of `run.h`, so the dependency with `cudaq-mlir-runtime` shifts from build time to user compile time. 
For Python, the python mlir extension already has a dependency with MLIR, so `getLayoutInfo` can be called from `run_py`.

### DIFF
--- a/runtime/common/LayoutInfo.cpp
+++ b/runtime/common/LayoutInfo.cpp
@@ -6,32 +6,19 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
+#include "LayoutInfo.h"
 #include "RuntimeMLIR.h"
-#include "common/ExecutionContext.h"
-#include "common/RecordLogParser.h"
-#include "cudaq.h"
-#include "cudaq/Optimizer/Builder/Factory.h"
+#include "common/DeviceCodeRegistry.h"
 #include "cudaq/Optimizer/Builder/Runtime.h"
-#include "cudaq/Optimizer/CodeGen/Passes.h"
 #include "cudaq/Optimizer/CodeGen/QIROpaqueStructTypes.h"
-#include "cudaq/Optimizer/InitAllDialects.h"
-#include "cudaq/algorithms/run.h"
-#include "cudaq/simulators.h"
-#include "nvqir/CircuitSimulator.h"
+#include "cudaq/runtime/logger/logger.h"
 #include "llvm/IR/DataLayout.h"
-#include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
-#include "mlir/IR/DialectRegistry.h"
-#include "mlir/IR/MLIRContext.h"
 #include "mlir/IR/Types.h"
 #include "mlir/Parser/Parser.h"
-#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
 #include "mlir/Target/LLVMIR/TypeToLLVM.h"
-#include <utility>
 
 using namespace mlir;
-
-#include "LayoutInfo.h"
 
 namespace {
 cudaq::LayoutInfoType extractLayout(const std::string &kernelName,
--- a/runtime/cudaq/algorithms/run.cpp
+++ b/runtime/cudaq/algorithms/run.cpp
@@ -5,6 +5,7 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
+
 #include "run.h"
 #include "common/LayoutInfo.h"
 #include "common/RecordLogParser.h"


Renaud-K:
> /ok to test eae7e784ee345b4d6d8ddde01235f2e1657652f6

runtime/common/CMakeLists.txt
```diff
@@ -149,7 +149,6 @@ target_link_libraries(cudaq-mlir-runtime
     MLIRLLVMCommonConversion
     MLIRLLVMToLLVMIRTranslation
   PRIVATE
-    cudaq
```
sacpis, line 152: @Renaud-K Is this PR building?

runtime/common/JIT.h
```diff
@@ -7,27 +7,23 @@
...
-// Optionally, the JIT'ed kernel can be executed a number of
-// times along with a post-execution callback. For example, sample a dynamic
-// kernel.
-std::unique_ptr<llvm::orc::LLJIT>
-invokeWrappedKernel(std::string_view llvmIr, const std::string &kernelName,
-                    void *args, std::uint64_t argsSize,
-                    ExecutionContext &executionContext,
-                    std::size_t numTimes = 1,
-                    std::function<void(std::size_t)> postExecCallback = {});
+std::tuple<std::unique_ptr<llvm::orc::LLJIT>, std::function<void()>>
```
lmondada, line 26: This return type and its meaning might be easier to parse if you introduced a type alias for it.

I can't think of a good name for this type though, so maybe this is not a good idea.

runtime/common/CMakeLists.txt
sacpis: No worries. Had some bad symbols in the build directory on my machine. Works now.

runtime/common/JIT.h
lmondada: I've now seen how you use this later down. The only reason you return the LLJIT object is to keep it alive, right?

I would personally find something like the following nice and clean:

```cpp
// Using a helper alias to keep the return type readable
using JITFunction = std::unique_ptr<
    std::function<void()>, 
    std::function<void(std::function<void()>*)>
>;

JITFunction createWrappedKernel(...);
```

and then in the `.cpp` you'd construct such a pointer with e.g.

```c++
template<typename Callable>
JITFunction make_jit_function(std::unique_ptr<llvm::orc::LLJIT> jit, Callable&& func) {
    // 1. Move the callable into a heap-allocated std::function
    auto* raw_ptr = new std::function<void()>(std::forward<Callable>(func));

    // 2. Construct the unique_ptr. 
    // The deleter lambda captures the JIT unique_ptr by value.
    return JITFunction(
        raw_ptr,
        [keep_alive = std::move(jit)](std::function<void()>* p) {
            delete p; 
            // 'keep_alive' (LLJIT) is destroyed here after the function is deleted.
        }
    );
}
```

i.e you capture the LLJIT object in the custom deleter of `std::unique_ptr`.

But you might find that over the top. You can also keep your current solution :)

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
```diff
@@ -345,27 +367,30 @@ class RemoteRestRuntimeServer : public cudaq::RemoteRuntimeServer {
           // deleted.
           clearRegOpsAndDestroyJIT(llvmJit);
```
lmondada, line 368: if you were to use the custom deleter approach I mentioned above, you'd replace this with 

```c++
cudaq::getExecutionManager()->clearRegisteredOperations();
```

and you could delete `clearRegOpsAndDestroyJIT`

lmondada (approved):
🎉 thanks, nice and clean, I like it 😊

runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp
Renaud-K: I realize that the way the current code manages the LLJIT object is not great but changing that was not really the goal of the PR. The bigger issue is that having to bring this object to the calling side means adding a dependency with `cudaq-mlir-runtime`. As we are looking at removing this dependency from QPUs, my hope is that we will find a way of managing this lifetime differently. This will have to be revisited at that point.

runtime/common/JIT.h
Renaud-K: By making the deleter a callback, we would remove the dependency on `cudaq-mlir-runtime` but that was going to be a different task. Yes, we will need a mechanism like this one when we do that.

runtime/common/LayoutInfo.cpp
```diff
@@ -27,15 +27,14 @@
...
 #include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
 #include "mlir/Target/LLVMIR/TypeToLLVM.h"
-#include <string>
 #include <utility>
-#include <vector>
 
 using namespace mlir;
 
-using LayoutInfoType = std::pair<std::size_t, std::vector<std::size_t>>;
+#include "LayoutInfo.h"
```
khalatepradnya: nit: Should this include file be at the top?

runtime/common/LayoutInfo.h
```diff
@@ -0,0 +1,19 @@
...
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace cudaq {
+
+using LayoutInfoType = std::pair<std::size_t, std::vector<std::size_t>>;
```
khalatepradnya, line 15: maybe this should be in `cudaq::details` namespace?

runtime/cudaq/algorithms/run.cpp
```diff
@@ -0,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#include "run.h"
```
khalatepradnya, line 9: nit: Blank line between copyright banner and first include

python/extension/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ declare_mlir_python_extension(CUDAQuantumPythonSources.Extension
     ../runtime/cudaq/algorithms/py_sample_async.cpp
     ../runtime/cudaq/algorithms/py_resource_count.cpp
     ../runtime/cudaq/algorithms/py_run.cpp
+    ../../runtime/cudaq/algorithms/run.cpp      # Common Python and C++ implementation of run
```
khalatepradnya, line 62: nit: For consistency, move this to ~line#91

khalatepradnya (commented):
One comment on namespace and a few nits, otherwise LGTM.

runtime/common/JIT.h
Renaud-K: I also was not sure about the lifetime of objects captured in lambdas and how the lifetime of a lambda is defined. I was going to have experiment with it. Thank you for understanding and letting me proceed.

python/extension/CMakeLists.txt
Renaud-K: I am not sure what it would be consistent with. This is the first common C++/Python algorithm file that we list. I thought that we would list all the algorithm files together.

runtime/common/LayoutInfo.h
Renaud-K: Good question. It touches on the bigger issue of how we design our APIs.  As I am moving things around, I have been asking myself these questions a lot lately. At this point my thinking is that we need to distinguish between

- user API,  
- public module APIs 
- private APIs. 

The user API is what we allow users to use. We have to be careful about providing some backward compatibility and deprecation schedules. It also uses the snake_case style. (Because another question I have been asking is whether to change the case for some of these files)

public module APIs are public in the cmake sense. They need to be discoverable by the build system and `nvq++` but are not for users to use. The problem is that since they need to be discoverable for `nvq++` there is not a way to enforce users will not use them. I think we need to come up with some rules or conventions to distinguish them. These APIs could use the CamelCase style, simply because that is probably what most of them currently and would be less work.

private APIs can only be used within a module. We can enforce this by using them only through relative paths. Though I am not sure with TMP if the file path is relative to the current file or the included file. Some experimentation is needed there.

`getLayoutInfo` is a public module API. If we were to move all module APIs under details, we would be moving too much. So, it should not be under `details`. Maybe it can be under `compiler`. But so far we have not introduced a namespace for it. I would like us too though.

So, I would like us to have these discussions and formalize our APIs. This needs to come in the form of a proposal. I can start having some conversations and gather maybe some existing doc. We can discuss it more on Monday.

runtime/common/LayoutInfo.cpp
Renaud-K: Yes, it should. And I can remove a lot of unused headers as well.

runtime/cudaq/algorithms/run.cpp
Renaud-K: Will do.

Renaud-K:
> /ok to test 121401d0098789a2fc41826eac269f11a34b1d5b

python/extension/CMakeLists.txt
Renaud-K: I think  you meant to move it with the other `../../runtime` files. I realize this is not as visually appealing but I opted for listing the algorithm together. My hope is that all algorithm can be consolidated in a common file in a similar manner. It is an important precedent to set.

Renaud-K:
Adding @1tnguyen  and @boschmitt as an FYI.

PR #3878, opened 2026-02-06
[PTSBE] Trajectory Deduplication

sacpis:
Adding trajectory deduplication: hash trajectories by content, merge duplicates, and track multiplicity so downstream shot allocation and result aggregation can weight correctly.

### DIFF
--- a/unittests/ptsbe/ExecutePTSBETester.cpp
+++ b/unittests/ptsbe/ExecutePTSBETester.cpp
@@ -62,7 +62,7 @@ CUDAQ_TEST(ExecutePTSBETest, ThrowsWithoutExecutionContext) {
   batch.trajectories.push_back(traj);
 
   // No setExecutionContext() - should throw
-  EXPECT_THROW(samplePTSBEGeneric(sim, batch), std::runtime_error);
+  EXPECT_THROW(samplePTSBEGeneric(sim, batch), std::exception);
 }
 
 /// Single trajectory Hadamard circuit: execute H|0> and expect 50/50
--- a/unittests/ptsbe/MergeTasksWithTrajectoryTester.cpp
+++ b/unittests/ptsbe/MergeTasksWithTrajectoryTester.cpp
@@ -132,7 +132,7 @@ CUDAQ_TEST(MergeTasksWithTrajectoryTest, InvalidCircuitLocationThrows) {
   KrausTrajectory trajectory(0, selections, 0.1, 10);
 
   EXPECT_THROW(mergeTasksWithTrajectory<double>(baseTasks, trajectory),
-               std::runtime_error);
+               std::exception);
 }
 
 /// Verify noise at last valid gate index works
--- a/unittests/ptsbe/NoiseExtractorTester.cpp
+++ b/unittests/ptsbe/NoiseExtractorTester.cpp
@@ -144,7 +144,7 @@ TEST(NoiseExtractorTest, AmplitudeDampingIsNotUnitaryMixture) {
         auto result = extractNoiseSites(trace, noise_model, true);
         (void)result;
       },
-      std::invalid_argument);
+      std::exception);
 }
 
 TEST(NoiseExtractorTest, AmplitudeDampingWithValidationDisabled) {
@@ -158,7 +158,7 @@ TEST(NoiseExtractorTest, AmplitudeDampingWithValidationDisabled) {
         auto result = extractNoiseSites(trace, noise_model, false);
         (void)result;
       },
-      std::invalid_argument);
+      std::exception);
 }
 
 TEST(NoiseExtractorTest, MixedUnitaryAndNonUnitary) {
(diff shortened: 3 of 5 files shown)


runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <vector>
+
+namespace cudaq::ptsbe {
+
+namespace {
+
+// Referred from `runtime/cudaq/operators/helpers.cpp`
+inline void hashCombine(std::size_t &seed, std::size_t value) {
+  seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
```
taalexander, line 23: I wish this exact hashing logic wasn't replicated in N places through the codebase but I don't think now is the time to resolve it.

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+std::size_t hashKrausSelection(const cudaq::KrausSelection &sel) {
+  std::size_t h = std::hash<std::size_t>{}(sel.circuit_location);
+  for (std::size_t q : sel.qubits)
+    hashCombine(h, std::hash<std::size_t>{}(q));
+  hashCombine(h, std::hash<std::string>{}(sel.op_name));
```
taalexander: Small optimization but I think the op_name is potentially redundant (if it's referring to a known operator at that location and this is just a human-readable name)?

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
```diff
@@ -0,0 +1,74 @@
...
+    hashCombine(h, hashKrausSelection(sel));
+  return h;
+}
+
+std::vector<cudaq::KrausTrajectory>
+deduplicateTrajectories(std::span<const cudaq::KrausTrajectory> trajectories) {
+  if (trajectories.empty())
+    return {};
+
+  std::unordered_map<std::size_t, std::vector<std::size_t>> hash_to_indices;
```
taalexander: Minor - would an option would be to implement `struct std::hash<KrausTrajectory>` to use the KrausTrajectory directly as the hash?

runtime/cudaq/ptsbe/KrausTrajectory.h
```diff
@@ -69,13 +73,13 @@ struct KrausTrajectory {
...
-  /// @return true if trajectory_id, selections, probability, and num_shots
-  /// match
+  /// @return true if trajectory_id, selections, probability, num_shots, and
+  /// multiplicity match
   constexpr bool operator==(const KrausTrajectory &other) const {
     return trajectory_id == other.trajectory_id &&
            kraus_selections == other.kraus_selections &&
            std::abs(probability - other.probability) < PROBABILITY_EPSILON &&
-           num_shots == other.num_shots;
+           num_shots == other.num_shots && multiplicity == other.multiplicity;
```
taalexander, line 82: Just something to keep in mind. We might need to take care adding multiplicity here as it might be changing as we iterate through trajectories de-duplicating below.

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
```diff
@@ -0,0 +1,74 @@
...
+  for (std::size_t q : sel.qubits)
+    hashCombine(h, std::hash<std::size_t>{}(q));
+  hashCombine(h, std::hash<std::string>{}(sel.op_name));
+  hashCombine(h, static_cast<std::size_t>(sel.kraus_operator_index));
+  return h;
+}
+
+} // namespace
+
+std::size_t hashTrajectoryContent(const cudaq::KrausTrajectory &trajectory) {
```
taalexander, line 51: What happens if the probabilities of the trajectories are different? I don't think this should happen but is it considered an error if it does?

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
sacpis: Thanks, will keep in mind to refactor in order to use it at other places in CUDA-Q.

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
sacpis: Addressed in [59e2529](https://github.com/NVIDIA/cuda-quantum/pull/3878/commits/59e2529973494ab97469823ff00960af82030289).

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
sacpis: Addressed in [59e2529](https://github.com/NVIDIA/cuda-quantum/pull/3878/commits/59e2529973494ab97469823ff00960af82030289).

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
sacpis: It is not treated as an error. Deduplication now only keys on kraus_selection. When 2 trajectories have the same content but diff prob, they are merged. The first occurrence's prob (and id, num_shots) is kept and multiplicity is incremented. The second trajectory's prob is ignored.

runtime/cudaq/ptsbe/KrausTrajectory.h
sacpis: Now in trajectory dedup logic we are adding the incoming trajectory's multiplicity when merging and creating the representative.

PR #3879, opened 2026-02-06
[python] Fix type mismatch in struct vector member marshaling

huaweil-nv:
### Summary
Fix a bug in handleStructMemberVariable (py_alt_launch_kernel.cpp) where list[int] and list[bool] members inside a dataclass struct were incorrectly marshaled when passed to a CUDA-Q kernel.
The appendVectorValue template lambda had its type parameter T deduced but never used — it always created std::vector<double> and cast elements with v.cast<double>(). This caused non-float list members to be stored as doubles. When the kernel read the data back using the expected integer type, it interpreted the IEEE 754 bit pattern as an integer, producing garbage values.
For example, passing list[int] with value [7] resulted in the kernel reading 4619567317775286272 (0x401C000000000000, the double representation of 7.0).
### Root Cause
```
// Before (buggy): T is deduced but ignored
auto appendVectorValue = []<typename T>(..., T) {
    std::vector<double> *values = new std::vector<double>(asList.size());  // always double
    (*values)[i++] = v.cast<double>();                                      // always double
};
```
When T=double (the common list[float] case), behavior is unchanged.

### DIFF
--- a/runtime/cudaq/platform/quantum_platform.cpp
+++ b/runtime/cudaq/platform/quantum_platform.cpp
@@ -97,6 +97,28 @@ void quantum_platform::validateQpuId(std::size_t qpuId) const {
   }
 }
 
+// [remove at]: runtime refactor release
+// Deprecated: Use with_execution_context instead.
+void quantum_platform::set_exec_ctx(ExecutionContext *ctx) {
+  configureExecutionContext(*ctx);
+  detail::setExecutionContext(ctx);
+  beginExecution();
+}
+
+// [remove at]: runtime refactor release
+// Deprecated: Use with_execution_context instead.
+void quantum_platform::reset_exec_ctx() {
+  auto *ctx = getExecutionContext();
+  if (ctx == nullptr)
+    return;
+
+  detail::try_finally([this, ctx] { finalizeExecutionContext(*ctx); },
+                      [this] {
+                        endExecution();
+                        detail::resetExecutionContext();
+                      });
+}
+
 // Specify the execution context for this platform.
 // This delegates to the targeted QPU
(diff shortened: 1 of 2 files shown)


1tnguyen:
> /ok to test b46d0bf2962a0061e8bee9183a020772c343ced8

huaweil-nv:
Oooh, this fix caused a test to fail. Maybe it’s not the right fix — I’ll check it out.

huaweil-nv:
The existing test11 in test_list_of_dataclass_update_failures was wrapped in pytest.raises(AssertionError) because the struct argument conversion was known to be broken — there's a FIXME comment right there saying so. Now that the fix is in, the conversion works and the test no longer raises, which actually broke CI. 
```
____________________ test_list_of_dataclass_update_failures ____________________

    def test_list_of_dataclass_update_failures():
    
        @dataclass(slots=True)
        class MyTuple:
            l1: list[int]
            l2: list[int]
    
        # FIXME: The argument conversion from host to device is not correct
        # currently. Either add back the error message the main line gives, or
        # remove from here and uncomment test that this works above.
>       with pytest.raises(AssertionError) as e:
E       Failed: DID NOT RAISE <class 'AssertionError'>

python/tests/kernel/test_assignments.py:850: Failed
----------------------------- Captured stdout call -----------------------------
[1, 2, 1, 1, 1, 1, 1, 3]
```
So I just unwrapped it and let it assert the correct results in commit https://github.com/NVIDIA/cuda-quantum/pull/3879/changes/f515e88cee8de3b549346cb0878f5f1083cfeae1, which is exactly what the FIXME comment says to do when the bug gets fixed.

sacpis:
> /ok to test 48ab130

PR #3882, opened 2026-02-06
[python] Replace bindings for QIS types with stubs

lmondada:
In Python, the types `qubit`, `qvector` and `qview` are defined as bindings to the equivalent C++ types. These bindings are not useful, as these are only "placeholder" types that can only be used within kernels: they will never be called from within Python, but rather be parsed by our MLIR compiler and lowered away.

I suggest replacing those definitions with dummy Python classes. The class method stubs are not necessary, but I've added them as they are useful for Python static analysis tools (type checking, autocomplete etc).

This will allow us to remove the C++ implementations of those types in a future PR.

### DIFF
--- a/python/cudaq/kernel_types.py
+++ b/python/cudaq/kernel_types.py
@@ -61,7 +61,7 @@ class qubit(KernelType):
         ...
 
 
-class qview:
+class qview(KernelType):
     """A non-owning view on a register of qubits."""
 
     def size(self) -> int:
--- a/python/tests/kernel/test_kernel_qview.py
+++ b/python/tests/kernel/test_kernel_qview.py
@@ -40,4 +40,4 @@ def test_qview_non_zero_length():
 def test_qview_init_outside_of_kernel():
     # cannot instantiate a qview outside a kernel
     with pytest.raises(KernelTypeError):
-        cudaq.qview(2)
+        cudaq.qview()


python/runtime/cudaq/qis/py_pauli_word.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace cudaq {
+
+void bindPauliWord(py::module &mod) {
+
+  py::class_<pauli_word>(mod, "pauli_word",
+                         "The `pauli_word` is a thin wrapper on a Pauli tensor "
+                         "product string, e.g. `XXYZ` on 4 qubits.")
+      .def(py::init<>())
+      .def(py::init<const std::string>());
```
lmondada, line 21: This was all that was left over of `py_qubit_qis.cpp`, so I moved it to a file with a dedicated name.

sacpis:
Would you please elaborate a little bit more about why we are getting rid of bindings? Are we going to use these data types outside the kernel? I am just trying to see what do we gain by keeping the data types definition separate.

lmondada:
> Would you please elaborate a little bit more about why we are getting rid of bindings?

The bindings are currently never used. They _cannot_ be used because these types **only** make sense within a kernel, but on the other hand, all code within a kernel is never executed. It is parsed by the AST bridge and then transformed into MLIR.

By keeping these unused bindings, we are forcing ourselves to also keep the C++ implementations of these types around, which are themselves never used in MLIR mode (for exactly the same reason).

I am proposing to replace the unused bindings with the minimal amount of code that's actually useful: just the type signatures, so that type checkers, linters etc know which methods are available (and implemented by the MLIR compiler).

> Are we going to use these data types outside the kernel?

No - it doesn't make sense to allocate or handle qubits outside of qpu kernels.

PR #3884, opened 2026-02-06
[sample] Emit warning on named measurement ops in sampling context

khalatepradnya:
This PR introduces a deprecation warning for using named measurement registers (sub-registers) in sampling mode.

### DIFF
--- a/python/tests/dynamics/test_evolve_simulators.py
+++ b/python/tests/dynamics/test_evolve_simulators.py
@@ -249,7 +249,7 @@ def test_evolve(init_state):
     ]
     np.testing.assert_allclose(results_with_shots,
                                expected_result_ideal,
-                               atol=0.1)
+                               atol=0.15)
 
 
 def test_evolve_async():
@@ -329,7 +329,7 @@ def test_evolve_async():
     ]
     np.testing.assert_allclose(results_with_shots,
                                expected_result_ideal,
-                               atol=0.1)
+                               atol=0.15)
 
 
 @pytest.mark.skip(reason="Skipping test due to issue #3678")


khalatepradnya:
@efratshabtai for viz.

targettests/execution/sample_named_reg.cpp
```diff
@@ -0,0 +1,28 @@
...
+  }
+};
+
+int main() {
+  auto counts = cudaq::sample(test_kernel{});
+  counts.dump();
+  return 0;
+}
+
+// CHECK: WARNING: Kernel 'test_kernel' uses named measurement results
```
Renaud-K: Can we have a test that checks for the other warning and one for the exception if reachable? Or if the python flow covers them, can we make it clear which ones?

khalatepradnya: yes, added one in [873b492](https://github.com/NVIDIA/cuda-quantum/pull/3884/commits/873b4926c5296a91549580acb6fb67c6f64faa52).

The two `RUN` checks in this test both the warnings - one each.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -586,8 +587,14 @@ class BaseRemoteRESTQPU : public QPU {
...
-          executionContext->hasConditionalsOnMeasureResults = true;
-          break;
+          throw std::runtime_error(
+              "`cudaq::sample` and `cudaq::sample_async` no longer support "
+              "kernels "
+              "that branch on measurement results. Kernel '" +
+              kernelName +
+              "' uses conditional feedback. Use `cudaq::run` or "
+              "`cudaq::run_async` "
+              "instead. See CUDA-Q documentation for migration guide.");
```
lmondada, line 597: Is there a link that you can add to this help message? Could be useful

runtime/nvqir/CircuitSimulator.h
```diff
@@ -673,6 +674,19 @@ class CircuitSimulatorBase : public CircuitSimulator {
     // Ask the subtype to sample the current state
     auto execResult = sample(sampleQubits, getNumShotsToExec());
 
+    // Warn if there are named measurement registers beyond `__global__`
+    if (!executionContext->warnedNamedMeasurements &&
+        registerNameToMeasuredQubit.size() > 1) {
+      executionContext->warnedNamedMeasurements = true;
+      std::cerr
```
lmondada, line 681: You could use CUDAQ_WARN, given that `logger.h` is already included in this file.

efratshabtai (approved):
Only reviewed error / warning messages

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -646,6 +653,31 @@ class BaseRemoteRESTQPU : public QPU {
...
+            });
+            if (hasNamedMeasurements) {
+              executionContext->warnedNamedMeasurements = true;
+              std::cerr
+                  << "WARNING: Kernel \"" << kernelName
+                  << "\" uses named measurement results "
+                  << "but is invoked in sampling mode. Support for "
+                  << "sub-registers in `sample_result` is deprecated and will "
+                  << "be removed in a future release. Use `run` to retrieve "
+                  << "individual measurement results." << std::endl;
```
lmondada, line 677: You could use CUDAQ_WARN, given that `logger.h` is already included in this file.

runtime/common/ExecutionContext.h
```diff
@@ -135,6 +135,10 @@ class ExecutionContext {
   /// order.
   bool explicitMeasurements = false;
 
+  /// @brief Flag to indicate that a warning about named measurement registers
+  /// in sampling context has already been emitted.
+  bool warnedNamedMeasurements = false;
```
lmondada, line 140: Instead of tracking this in the `ExecutionContext`, consider using `std::call_once`, i.e. something like

```c++
#include <mutex>

void warnNamedMeasurements()
{
    // only print warning once
    static std::once_flag flag;
    std::call_once(flag, [](){ CUDAQ_WARN("..."); });
}
```

Then you could call this function multiple times without having to track a boolean in the execution context.

lmondada: actually re-reading this example: we don't really care about thread safety, so you could also just use `static bool flag` 😅 in either case, I think it would be better if we could avoid putting this in the execution context.

runtime/common/BaseRemoteRESTQPU.h
khalatepradnya: No the docs aren't ready yet, but we can update once link is available.

runtime/nvqir/CircuitSimulator.h
khalatepradnya: hmm, isn't that opt-in? If user turns off the logger, they wouldn't see the message at all? In this case, we want the warning to be issued no matter the log level...

runtime/common/BaseRemoteRESTQPU.h
khalatepradnya: ditto

runtime/common/ExecutionContext.h
khalatepradnya: I like the idea of not tracking in execution context, but in this case the warning must be issued per invocation of kernel with named measurements in sampling context.
For example, if we have two kernels where waning should be issued, then it will show up only for the first one, right?
```
cudaq::sample(kernelA).dump(); // shows warning
cudaq::sample(kernelB).dump(); // won't show warning even if it should
```
Can we accommodate that?

Issue #3880, opened 2026-02-06
IQM unit tests silently skipped in CI due to broken pip install


taalexander:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The iqm-client pip install in test_in_devenv.yml has been silently failing as the image enforces PEP 668. As a result, iqm-client is never installed, and ctest auto-skips iqm-tests and iqm-dqa-tests on every CI run.

https://github.com/NVIDIA/cuda-quantum/blob/dde0113945dd40c20e757f9d47d78e674723cae1/.github/workflows/test_in_devenv.yml#L94

From a recent upstream merge queue run:
```
python3 -m pip install iqm-client==28.0.0

error: externally-managed-environment
× This environment is externally managed
```

When the install is fixed (via --break-system-packages), 3 of the 8 IQM tests fail:

- iqm_IQMTester.invalidTokenFromEnvVariable
- iqm_IQMTester.iqmServerUrlEnvOverride
- iqm_IQMTester.tokenFilePathEnvOverride

### Steps to reproduce the bug

Run CI on main and look at logs.

### Expected behavior

Run CI on main and look at the logs

### Is this a regression? If it is, put the last known working version (or commit) here.

Seems to be last version before Ubuntu image migration

### Environment

Docker image: Ubuntu 24.04

### Suggestions

- Add `--break-system-packages`
- Fix the bugs in tests

khalatepradnya:
@Renaud-K for viz. This is the same issue we discussed last week.

Renaud-K:
I suspect that there will be other failures after this one is fixed. This for instance:
```
CUDAQ_TEST(IQMTester, invalidTokenFromEnvVariable) {
  char *token = getenv("IQM_TOKEN");

  EXPECT_THAT([]() { setenv("IQM_TOKEN", "invalid-invalid-invalid", true); },
              testing::ThrowsMessage<std::runtime_error>(
[...]

PR #3885, opened 2026-02-06
run IQM tests; fix lambda expression

mitchdz:
Fixes https://github.com/NVIDIA/cuda-quantum/issues/3880.

This PR addresses two related issues uncovered after upgrading to `ubuntu:24.04`.

First, pip now requires `--break-system-packages` for system-wide installs. Because the test scripts were running without `set -e`, the resulting pip install failure was silently ignored, causing the affected tests to be skipped unintentionally.

Second, the IQM tests themselves were ineffective: the lambda expressions only called `setenv()`, which does not throw. As a result, the expected exceptions were never triggered. The tests now explicitly initialize the IQM backend after setting environment variables.


taalexander (approved):
LGTM, thank you.

PR #3886, opened 2026-02-06
Add more flexibility for python kernels

atgeller:
### Description

- Moves around some checks in `ArgumentConversion.cpp` to allow compiling python kernels without fully synthesized arguments.
- Adds dummy arguments that are ignored during argument synthesis to allow compiling a python kernel that takes arguments but also needs to have calls to other python kernels resolved.

Includes tests to ensure that the changes to compiling python kernels does not break in the following circumstances:
- When a python kernel has a callable argument that needs to be resolved (e.g., calls another python kernel).
- When a python kernel returns a value to a C++ kernel

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -445,10 +445,7 @@ class PyKernelDecorator(object):
     def formal_arity(self):
         return len(self.arg_types())
 
-    def handle_call_arguments(self,
-                              *args,
-                              ignoreReturnType=False,
-                              allow_no_args=False):
+    def handle_call_arguments(self, *args, allow_no_args=False):
         """
         Resolve all the arguments at the call site for this decorator.
         """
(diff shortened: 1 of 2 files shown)


atgeller:
The issue in `CPPPyKernelDecorator` with the lifetime of the returned JITed execution engine not having a long enough lifetime should be addressed by #3908, so the ad-hoc solution here can be replaced when that is merged in.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -297,7 +297,8 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
       throw std::runtime_error("cache must not be populated");
     *cache = jit;
 
-    std::string entryName = resultTy ? name + ".thunk" : fullName;
+    // std::string entryName = resultTy ? name + ".thunk" : fullName;
```
Renaud-K: With https://github.com/NVIDIA/cuda-quantum/pull/3933, you will be able to simplify this by calling straight to jit->run(entryName). I might have introduced `lookupRawName OrFail` just for this case.

atgeller: Excellent! I'll wait for #3933 and then make that change.

atgeller:
This is waiting for #3987 to go through so we have a proper way of passing device kernels from python to C++.

atgeller:
> /ok to test e147c2d

atgeller:
> /ok to test ed2c167

PR #3890, opened 2026-02-07
Adding probability validation to noise channels

### LINKED ISSUE #3663: No error message raised when required for cudaq.apply_noise
### Required prerequisites

### Describe the bug

There's no error message raised if sum(p1,p2,p3)>1 when using cudaq.apply_noise(cudaq.Pauli1, p1, p2, p3, qvector[0]). I assumed that we could apply multiple errors at once but this isn't allowed.

For example, if I write:

cudaq.apply_noise(cudaq.Pauli1, 1, 0.3, 0.05, qvector[0])

the kernel will crash without an error message to guide the user in noticing where an error has been made.

#Short example code to replicate the bug:
```
import cudaq

#setting the target to the density matrix simulator
cudaq.set_target('density-matrix-cpu')

qubit_count = 3

#defining the quantum kernel function
@cudaq.kernel
def kernel(qubit_count: int):
    qvector = cudaq.qvector(qubit_count) #allocating the qubits

    cudaq.apply_noise(cudaq.Pauli1, 1, 0.3, 0.05, qvector[0])

    for i in range(qubit_count):
        h(qvector[i])
        h(qvector[i])

    #measure in Z-basis
    mz(qvector)

print(cudaq.draw(kernel, qubit_count)) #draw the circuit
```
### Expected behavior

With the above provided code, I expect for the kernel to crash, with no error message detailing why this happened.

Not a regression

### Environment

- **CUDA-Q version**: 0.12.0.post1
- **Python version**: 3.13.2
- **C++ compiler**: /
- **Operating system**: Fedora Linux

### Suggestions

I would like for there either to be a limit on each p_i to be max 1 (so multiple errors could be introduced in 1 step) or for an error message to be raised when sum_i(p_i)>1.

sacpis:
Adding probability validation to noise channels along with tests.

Fixes #3663

### DIFF
--- a/python/tests/builder/test_NoiseModel.py
+++ b/python/tests/builder/test_NoiseModel.py
@@ -1087,15 +1087,6 @@ def test_noise_validation_probability_check(target: str):
     with pytest.raises(RuntimeError, match=INVALID_PROBABILITY_MSG):
         cudaq.BitFlipChannel(1.5)
 
-    @cudaq.kernel
-    def kernel_invalid():
-        q = cudaq.qvector(1)
-        cudaq.apply_noise(cudaq.XError, -0.1, q[0])
-        mz(q)
-
-    with pytest.raises(RuntimeError, match=INVALID_PROBABILITY_MSG):
-        cudaq.sample(kernel_invalid, noise_model=cudaq.NoiseModel())
-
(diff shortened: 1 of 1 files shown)


sacpis:
Seems like we will have to validate probability params in Python before execution. So invalid values raise runtime error instead of aborting when C++ validates during JIT.

PR #3895, opened 2026-02-09
[python] Simplify PyKernelDecorator __init__

lmondada:
In the process of familiarizing myself with the code, I reorganized the logic a bit.

I find the refactored version much easier to read and parse. I've been quite careful to keep it equivalent to the old code (go through the branches one by one, you should see that it does the same thing).

I am opening this PR as I think it is a net positive so we might as well  merge it in, but feel free to disagree. If you think this is an unnecessary change, I'll happily close the PR.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -153,7 +153,6 @@ class PyKernelDecorator(object):
             self._parse_signature_from_python()
             self.pre_compile()
 
-
     def __del__(self):
         # explicitly call `del` on the MLIR `ModuleOp` wrappers.
         if self.qkeModule:


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -86,178 +87,80 @@ def __init__(self,
...
-                         self.kernelFunction.__name__)
-            self.location = ((inspect.getfile(self.kernelFunction),
-                              inspect.getsourcelines(self.kernelFunction)[1])
-                             if self.kernelFunction is not None else ('', 0))
 
+        if funcSrc is not None:
+            emitWarning(
+                "Passing 'funcSrc' to PyKernelDecorator is deprecated. Pass a string to `function` instead."
+            )
+            function = funcSrc
```
lmondada, line 92: The previous code checked whether `function` was a string, but then ignored its contents if it was. As a result, a dummy string value had to be passed to the constructor in `PyKernelDecorator.from_json`. The function source code was then expected to be found in `funcSrc`.

I see two ways of cleaning this up:
1. Use `function` for this purpose, i.e. `function` can either be a Python function OR a string, in which case it is treated as source code. This is what I've implemented here.
2. Make the `function` argument optional, and make `function` and `funcSrc` mutually exclusive (i.e. error out if both are provided).

Let me know if you prefer 2.

1tnguyen: 👍

PR #3897, opened 2026-02-09
Python 3.14 package support / CI

taalexander:
### Description
I was working on my machine and noticed there is not yet support for "Pi"thon (3.14) 😄 . Consider this a first draft to see what is missing.


mitchdz:
Make sure to run these changes through deployments once ready.

taalexander:
Known 3.14 blockers:
- CuPy has not yet released support for 3.14. Based on this comment https://github.com/cupy/cupy/issues/9604#issuecomment-3860865060 we're about two weeks out. 
- Cuquantum support is expected in May.

taalexander:
Closing as blocked for a couple reasons listed above.

acosmicflamingo:
Oof, what bad luck that CuPy actually released 14.0.0 with Python 3.14 support less than a week after @taalexander  checked (and technically was on [pypi](https://pypi.org/project/cupy/14.0.0/) February 16th). Although, cuQuantum looks like it doesn't support Python 3.14 yet. 

Judging by this [PR](url) and looking at the pypi URLs for the cuquantum packages, they all look to have noarch behavior, so it's not that cuda-quantum needs cuquantum to add Python 3.14 support. I suppose that means there are no more Python 3.14 blockers? If so, I could certainly take a crack at it.

PR #2946, opened 2025-05-22
[nfc] Unify the internal CUDA-Q namespces

khalatepradnya:
This pull request standardizes the namespace usage in the codebase by replacing occurrences of `cudaq::details` with `cudaq::detail`. Also moves `cudaq::__internal__` into single `cudaq::detail` namespace.

PR #3898, opened 2026-02-09
C++ API doc

Renaud-K:
This is a proposal to formalize the coding style of our C++ APIs. (Which can later be consolidated with python).
The main ideas is that we define 3 types of interfaces:

- user API
- internal public APIs
- private APIs

To clearly distinguish between the user API and internal APIs. I propose headers under a different root directory.
User API is under `cudaq/`
Internal APIs would be under `cudaq_dev/`, `cudaq_internals/` or `cudaq_<module_name>/` . TBD.

Another significant change is that module APIs are all under a <module_name> namespace.
Example: `cudaq::compiler::lowerQuake`

### DIFF
--- a/CppAPICodingStyle.md
+++ b/CppAPICodingStyle.md
@@ -6,14 +6,12 @@ toolchain.
 
 ## 1. Goals
 
-1. Make it obvious which APIs are supported for **end users** versus **internal
-development**.
-2. Preserve long-term stability for the **User API** (backward compatibility +
- deprecation schedule).
-3. Allow **Hardware vendors** and **core developers** to access internal
-interfaces without accidentally turning them into user-facing commitments.
-4. Ensure `cudaq` can discover headers needed to compile user code (including
- internal headers that user headers must include).
+1. Provide definitions for the different levels of APIs developers or end users
+may encounter.
+2. Provide definitions for developer categories and how each category may
+interact with the above APIs.
+3. Make it obvious which APIs are supported for end users versus internal
+development.
 
 ---
 
@@ -21,8 +19,9 @@ interfaces without accidentally turning them into user-facing commitments.
 
 ### 2.1 Users (e.g, Quantum algorithm developers)
 
+- Do not contribute to changes to the `cudaq` codebase.
 - Use only the **User API** shipped with the product.
-- Compile with `cudaq` and link to shipped runtime libraries.
+- Compile and link to shipped runtime libraries with `nvq++`
 - Must not rely on internal headers/namespaces.
 
 ### 2.2 Library developers (e.g., `cudaqx`)
@@ -33,17 +32,17 @@ their own CMake project.
 
 ### 2.3 Hardware vendors and core developers
 
+- Do contribute to changes to the `cudaq` codebase.
 - May use:
   - **User API**
   - **Internal public module APIs**
   - **Module private APIs**
-- Contribute changes to the codebase and/or add backends, modules, extensions.
 
 ---
 
 ## 3. API Layers and Rules
 
-We define a three API layers as illustrated below:
+We define three API layers as illustrated below:
 
 1. **User API**
 2. **Internal public module APIs**
@@ -67,7 +66,7 @@ We define a three API layers as illustrated below:
 ├────────────────────────────────────────────────────────────────────┤
 │ Audience:   Hardware vendors + core developers                     │
 │            (NOT for users / external libs to depend on)            │
-│ Headers:    "cudaq_internals/<module>/<hdr>.h"  (or cudaq_dev/...) │
+│ Headers:    "cudaq_internal/<module>/<hdr>.h"  (or cudaq_dev/...)  │
 │ Namespace:  cudaq::<module>::...  (module lowercase)               │
 │             cudaq::<module>::detail = NON-public                   │
 │ Naming:     CamelCase (or consistent module convention)            │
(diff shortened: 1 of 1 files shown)


CppAPICodingStyle.md
```diff
@@ -0,0 +1,236 @@
...
+  - Example: `#include "cudaq/algorithms/run.h"`
+- File naming:
+  - lowercase file names
+
+#### 3.1.3 Namespaces
+
+- All user-visible declarations live in `namespace cudaq { ... }`.
+- Nested namespaces are considered public **except**:
+  - `cudaq::details`
+  - `cudaq::__internal__`
```
lmondada: I agree with an earlier point that @khalatepradnya made: these private namespaces should be consolidated into one. See also https://github.com/NVIDIA/cuda-quantum/pull/2946

CppAPICodingStyle.md
```diff
@@ -0,0 +1,236 @@
...
+
+---
+
+### 3.2 Internal Public Module APIs (Exported, not user-supported)
+
+#### 3.2.1 Definition
+
+Internal public module APIs are:
+
+- **not supported for end users**, but
```
lmondada, line 141: An alternative to consider: instead of allowing breaking changes with no notifications downstream, we could treat each module independently with its own version number. At any breaking change of the public API of a module, we would update the version and hike the dependency upstream as required. The expectation would be that these version numbers increase much faster than the more "stable" cudaq user-facing library (e.g. weekly, if not daily) .

In this way, if `cudaqx` or other projects must depend on one of these modules, they could pin this version. It would also facilitate shipping binaries and only rebuilding what's required.

This is what we used to do in Rust-land and worked very well. However, it assumes we have a good build system and a distribution system for pre-compiled binaries. This comes for free in e.g. Python and Rust, not sure how easy it is to achieve in C++. If we go down this route then publishing a new version of a module must be as easy as changing the version number in one file, with CI handling the rest.

lmondada (commented):
Thank you so much for taking the time to write this. I fully agree that something like this is needed and would organise the code base much more clearly.

We should also think a bit about how we want to package these modules to make getting started and building CUDAQ and downstream easier. I've made a first comment about this, below.

CppAPICodingStyle.md
Renaud-K: I could make the change. I believe the runtime is using `details` a lot more than `detail`, so this could be a lot of work. But officially, we could just make it `detail`.

CppAPICodingStyle.md
```diff
@@ -0,0 +1,236 @@
...
+  - **Module private APIs**
+- Contribute changes to the codebase and/or add backends, modules, extensions.
+
+---
+
+## 3. API Layers and Rules
+
+We define three API layers:
+
+1. **User API** (public, supported, stable)
```
sacpis: Would you please tell the difference between public, supported, and stable?

CppAPICodingStyle.md
```diff
@@ -0,0 +1,236 @@
...
+2. **Internal public module APIs** (not user-supported, but exported for internal development)
```
sacpis: Would we expose such APIs to end users? Do you have an example here?

CppAPICodingStyle.md
```diff
@@ -0,0 +1,236 @@
...
+
+These APIs often must be shipped because **user headers may include them**, and nvq++ must be able to find them.
+
+#### 3.2.2 Header location and include prefix (recommendation)
+
+Internal public headers must **not** live under the `cudaq/` include root to avoid confusion with the User API.
+
+**Proposed convention:**
+
+- `#include "cudaq_internals/<module_name>/<header_name>.h"`
```
sacpis: Like this one.

CppAPICodingStyle.md
sacpis: No worries. I see it below.

CppAPICodingStyle.md
Renaud-K: Doesn't 3.1 develop these ideas enough?

CppAPICodingStyle.md
Renaud-K: If users look into one of our headers they could see such a path. It is unavoidable and is the reason why we need to better distinguish public and internal APIs. We cannot prevent them from including such headers as well, but this would be an unsupported usage.

CppAPICodingStyle.md
sacpis: Adding a bit about supported and stable would help there.

sacpis:
In this PR's description if we could have a kind of graph for these APIs, that would really help with quickly understanding the proposed structure.

CppAPICodingStyle.md
Renaud-K: Fixed.

CppAPICodingStyle.md
Renaud-K: This is next level though. Also the doc does not cover what the user API and how it should be transitioned. It just addresses issues that are blocking and does not put any restriction on how versioning should be implemented.

CppAPICodingStyle.md
Renaud-K: I am not sure how to make this concise and clearer. So, I have just removed it. It was just intended to summarize what was to come. Maybe it was just adding confusion.

Renaud-K:
> In this PR's description if we could have a kind of graph for these APIs, that would really help with quickly understanding the proposed structure.

There are several ways this could be implemented and this document does not intend to restrict any implementations but only to provide an initial framework. Anything covering the physical structure of these API would belong in a design doc. And I know that some people have an opinion there. 

I have added a graph to show how APIs would be conceptually layered. I hope this will help with your understanding.

CppAPICodingStyle.md
lmondada: agreed!

CppAPICodingStyle.md
Renaud-K: That was changed.

Renaud-K:
> /ok to test 8b4b2ad890cec01d18c6b26b81cd78fe789a545e

mitchdz (approved):

PR #3906, opened 2026-02-10
Fix deprecations warnings in CI

sacpis:
Fix multiple deprecation warnings in CI.
https://github.com/NVIDIA/cuda-quantum/actions/runs/21833766797/job/63037798714#step:10:3969


python/tests/operator/test_matrix_op.py
```diff
@@ -752,10 +752,8 @@ def check_composite(op_create, matrix_create, coeff_val):
 
     check_composite(zero, zero_matrix, 0)
     check_composite(identity, identity_matrix, 1)
-    with pytest.warns(DeprecationWarning):
-        create(5)
-    with pytest.warns(DeprecationWarning):
-        annihilate(5)
+    boson.create(5)
+    boson.annihilate(5)
```
1tnguyen, line 756: I wonder why we need to modify this test: seems like it's testing the deprecation warnings.

sacpis: Do we need to check for deprecated warnings anymore as we have fixed and have started using the new API?

1tnguyen: If we don't need to check deprecation warnings, I guess we can just remove these 2 lines instead, since they are just creating the operator but not using the results.

1tnguyen (approved):
LGTM 👍

PR #3908, opened 2026-02-10
Resolving the dependency of `cudaq-python-interop` lib

1tnguyen:
### Description

There is a sort of circular dependency in `cudaq-python-interop`.

It now has explicit dependencies on MLIR (via `ExecutionEngine`), which is available in `CUDAQuantumPythonSources.Extension`.
However, `cudaq-python-interop` is a dependency of `CUDAQuantumPythonSources.Extension`.

Hence, breaking this loop at `marshal_and_retain_module` in `py_alt_launch_kernel.cpp`, where we have access to the existing `JITExecutionCache` utility. The cache key is returned for clean-up purposes.

### DIFF
--- a/lib/Optimizer/Transforms/SROA.cpp
+++ b/lib/Optimizer/Transforms/SROA.cpp
@@ -50,8 +50,18 @@ public:
         if (ptrOp.getNumIndices() == 1 && ptrOp.getConstantIndex(0))
           continue;
       if (auto cast = dyn_cast<cudaq::cc::CastOp>(user))
-        if (castMatches(allocOp.getElementType(), cast.getResult().getType()))
+        if (castMatches(allocOp.getElementType(), cast.getResult().getType())) {
+          for (auto *cast_user : cast->getUsers()) {
+            if (!cast_user)
+              continue;
+            if (auto load = dyn_cast<cudaq::cc::LoadOp>(cast_user))
+              continue;
+            if (auto store = dyn_cast<cudaq::cc::StoreOp>(cast_user))
+              continue;
+            return failure();
+          }
           continue;
+        }
       if (auto load = dyn_cast<cudaq::cc::LoadOp>(user))
         continue;
       return failure();
--- a/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
+++ b/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
@@ -394,6 +394,22 @@ def test_run():
             non_zero_count += 1
     assert non_zero_count > 0
 
+    @cudaq.kernel
+    def kernel_with_conditional() -> list[bool]:
+        var = [True, True]
+        q0 = cudaq.qubit()
+        q1 = cudaq.qubit()
+        var[0] = mz(q0)
+        var[1] = mz(q1)
+        return var
+
+    results = cudaq.run(kernel_with_conditional, shots_count=2)
+    assert len(results) == 2
+    for res in results:
+        assert len(res) == 2
+        assert res[0] is False
+        assert res[1] is False
+
 
 def test_quantinuum_state_preparation():
 
--- a/targettests/execution/issue_3806.cpp
+++ b/targettests/execution/issue_3806.cpp
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// clang-format off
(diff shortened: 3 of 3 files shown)


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -963,7 +977,19 @@ static void *marshal_and_retain_module(const std::string &name,
...
   clone.erase();
-  return funcPtr;
+  assert(cachedEngine && *cachedEngine &&
+         "Expected the cached engine pointer to be set by "
+         "streamlinedSpecializeModule.");
+  // Use address as the hash key to cache the JITted engine, and store the
+  // engine pointer in the cache
+  getJITCache().cache(reinterpret_cast<std::size_t>(cachedEngine),
+                      cachedEnginePtrStorage);
+  return std::make_pair(funcPtr, reinterpret_cast<std::size_t>(cachedEngine));
```
sacpis: I think here cachedEngine is the address of a local variable which can be reused across calls. Basically this can cause key collision and engine leaks. So we can use unique id for each call instead of a stack address. Like a counter.

```
static std::atomic<std::size_t> nextcacheKey{1};
std::size_t key = nextCacheKey++;
getJITCache().cache(key, cachedEnginePtrStorage);
return std::make_pair(funcPtr, key);
```

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -59,6 +59,17 @@ struct PyStateVectorData {
   cudaq::simulation_precision precision = cudaq::simulation_precision::fp32;
   std::string kernelName;
 };
+
+cudaq::JITExecutionCache &getJITCache() {
+  // Runtime JIT cache for storage of JIT execution engines for Python-launched
+  // kernels. This is needed mainly for interop with C++, whereby we want to
+  // JIT-compile the Python kernel and then call it from C++.
+  static std::unique_ptr<cudaq::JITExecutionCache> jitCache;
```
sacpis, line 65: Seems like as we are now using getJITCache(), the global variable on line [50](https://github.com/1tnguyen/cuda-quantum/blob/f60513190e5cf5cad308dfb994021bc55745fbad/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp#L50) and its initialization on line [1181](https://github.com/1tnguyen/cuda-quantum/blob/f60513190e5cf5cad308dfb994021bc55745fbad/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp#L1181) can be removed.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -963,7 +977,19 @@ static void *marshal_and_retain_module(const std::string &name,
   void *funcPtr = cudaq::streamlinedSpecializeModule(name, clone, rawArgs,
                                                      resTy, cachedEngine);
   clone.erase();
-  return funcPtr;
+  assert(cachedEngine && *cachedEngine &&
```
sacpis: I think assert gets removed in the release build, right?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
1tnguyen: This is a bug :) 

The intention here is to use the memory address of the actual `ExecutionEngine` object that we allocated (with `new`) as the key. Hence, this should be unique until we clean up that object in memory.  It should be `ExecutionEngine*`. Will fix it accordingly.

1tnguyen: Fixed in ed3328d4a1985faae7fbc7a824fced74ec08cb45

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
1tnguyen: Indeed. The other one is actually unused, hence removed in ed3328d4a1985faae7fbc7a824fced74ec08cb45.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
1tnguyen: Good point. I made it a runtime exception instead in ed3328d4a1985faae7fbc7a824fced74ec08cb45.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -59,6 +57,17 @@ struct PyStateVectorData {
   cudaq::simulation_precision precision = cudaq::simulation_precision::fp32;
   std::string kernelName;
 };
+
+cudaq::JITExecutionCache &getJITCache() {
```
Renaud-K, line 61: Any reason why this not in `JITExecutationCache.cpp`? 
I can move it when I merge https://github.com/NVIDIA/cuda-quantum/pull/3933.

PR #3909, opened 2026-02-10
Fix density matrix overlap dimension

huaweil-nv:
### Description

- Fix a bug in `CuDensityMatState::overlap()` where the density matrix branch incorrectly used `dimension` (total element count, N²) as the Eigen matrix side length, creating an N² × N² matrix and attempting to copy N⁴ elements. This would cause excessive memory allocation or crashes. The fix uses `sqrt(dimension)` as the matrix side length, consistent with `dump()`, `getTensor()`, and `operator()` in the same file.
- Add regression tests for density matrix overlap covering self-overlap, orthogonal states, and partial overlap.
- Include `singleStateDimension` in the move constructor and move assignment operator for completeness. This is a defensive change with no impact on current usage, as batched states with non-zero `singleStateDimension` are always heap-allocated and managed via `unique_ptr`.

### Details

In `CuDensityMatState`, the `dimension` member stores the **total number of elements**: N for state vectors, N² for density matrices. The overlap method's density matrix branch directly used `dimension` as both the row and column count of the Eigen matrix:

```
// Before (buggy):
Eigen::MatrixXcd state(dimension, dimension);       // N² × N² matrix
const auto size = dimension * dimension;             // copies N⁴ elements
```


1tnguyen (approved):
LGTM 👍

1tnguyen:
> /ok to test 4b6a6ca1eb8e32f43e68a26ab81accce559f6734

runtime/nvqir/cudensitymat/CuDensityMatState.cpp
```diff
@@ -55,13 +55,16 @@ CuDensityMatState::overlap(const cudaq::SimulationState &other) {
   }
 
   // FIXME: implement this in GPU memory
-  Eigen::MatrixXcd state(dimension, dimension);
-  const auto size = dimension * dimension;
+  // For density matrices, `dimension` is the total number of elements (N^2).
+  // The matrix side length is sqrt(dimension).
+  const std::size_t matDim = static_cast<std::size_t>(std::sqrt(dimension));
+  Eigen::MatrixXcd state(matDim, matDim);
+  const auto size = dimension;
```
sacpis: Wondering why do we need to store dimension in the size variable? Can we use the dimension directly?

huaweil-nv:
> /ok to test

huaweil-nv:
> /ok to test 83e5f39

runtime/nvqir/cudensitymat/CuDensityMatState.cpp
huaweil-nv: Yes, fixed in new commit.

huaweil-nv:
> /ok to test 9b42918

khalatepradnya:
> /ok to test 573a323

PR #3913, opened 2026-02-10
Fix missing runtime dependency for check-targets

### LINKED ISSUE #1571: make check-targets doesn't have all dependencies in the CMakeLists.txt files
### Required prerequisites

### Describe the bug

When switching between different branches, the command
```
% make check-targets
```
does not build all of the runtime libraries and as a result a large number of tests will fail. One possible failure mode is a spew of undefined symbols, where the symbols are defined in the last branch that was built.

```
% git checkout experimental/stateHandling
% cd build; make install
% cd ..; git checkout main
% cd build; make check-targets
```

### Expected behavior

The above sequence of commands should not produce linker failures with undefined symbols. The runtime libraries should be rebuilt.

Not a regression

### Environment

- **CUDA Quantum version**: tip
- **Python version**: irrelevant
- **C++ compiler**: clang++ 16

### Suggestions

_No response_

Siddharth-06:
What:
- Ensure check-targets depends on the CUDA-Q runtime target.

Why:
- Prevents stale runtime libraries and undefined symbol errors when switching branches and reusing a build directory, as described in #1571.

How:
- Add an explicit dependency on the cudaq-runtime target to CUDAQ_TEST_DEPENDS while preserving existing fine-grained runtime dependencies.

Notes:
- Local reproduction on macOS was limited by external toolchain requirements (LLVM 16 / ZLIB). This is a build-graph dependency fix and does not alter runtime behavior; CI on supported platforms should validate the change.

Fixes #1571


schweitzpgi:
> /ok to test 7eaff2c

schweitzpgi:
LGTM. Thanks!

targettests/CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ if (NOT CUDAQ_DISABLE_CPP_FRONTEND)
     cudaq-quake
     fixup-linkage
     nvq++
+    cudaq-runtime
```
Renaud-K, line 47: Where is `cudaq-runtime` defined? I could not find a definition for it.

taalexander:
> /ok to test 96e82a8

schweitzpgi:
Does not fix the issue.

PR #3916, opened 2026-02-10
Upgrade assets build runner for better performance

mitchdz:
The current assets build takes ~1h 15m to complete ([example run](https://github.com/NVIDIA/cuda-quantum/actions/runs/21874755484/job/63144837346)￼).

With the new runners, the same build completes in ~41 minutes ([example run](https://github.com/NVIDIA/cuda-quantum/actions/runs/21879464248/job/63168279666?pr=3916)￼).

Note: After merging, we should monitor runner utilization. If this CPU group becomes oversubscribed, queue times could offset the performance gains. This change is straightforward to revert if needed.


1tnguyen (approved):
🚀

PR #3917, opened 2026-02-10
Change device key lookup behavior in python bridge

atgeller:
### Description

The updates to the python bridge changed `resolveQualifiedName` in two subtle ways:
1. The matching behavior was changed - from `endsWith` to strict equality
2. Instead of returning on the first successful match, the last match was returned.

This introduced an `Unknown Function Call` error for aliased kernel names in submodules (e.g., `cudaq_solvers.stateprep.ceo`, which is imported from and therefore an alias for `cudaq_solvers.pycudaqx_solvers_the_suffix_matters_cudaq_solvers.stateprep.ceo`). To resolve this, I've changed the returned device key to get the full module path, rather than the aliased module path.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -2695,7 +2695,10 @@ class PyASTBridge(ast.NodeVisitor):
                     obj = frame.f_locals.get(moduleNames[0])
                 elif moduleNames[0] in frame.f_globals:
                     obj = frame.f_globals.get(moduleNames[0])
-                if obj:
+                if obj is not None:
+                    # In case a module has been imported multiple times, grab the latest
+                    if isinstance(obj, list):
+                        obj = obj[-1]
                     devKey = checkModule(obj, moduleNames)
                     if devKey:
                         return devKey, pyVal.attr


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3501,17 +3505,24 @@ def check_vector_init():
...
+                                    break
+                                elif not otherFuncName:
+                                    otherFuncName = f"{devKey}.{name}" if devKey else name
+                                    maybeKernelName = cudaq_runtime.checkRegisteredCppDeviceKernel(
+                                        self.module, otherFuncName)
+                                    if maybeKernelName != None:
+                                        self.emitFatalError(
+                                            "calling cudaq.control or cudaq.adjoint on "
+                                            "a kernel defined in C++ is not currently "
+                                            "supported", node)
```
sacpis: Seems like we have a bug here. When `processDecorator` call succeeds, we break and keep that name. So we are good here. When `processDecorator` call does not succeeds, we keep overwriting the fallback each time. Only the last devkey's fallback survives. 

I think we can change it to

```
for devKey in devKeys:
    otherFuncName = processDecorator(name, path=devKey)

    if not otherFuncName:
        otherFuncName = f"{devKey}.{name}" if devKey else name

    if otherFuncName:
        maybeKernelName = cudaq_runtime.checkRegisteredCppDeviceKernel(
            self.module, otherFuncName)
        if maybeKernelName is not None:
            self.emitFatalError(
                "calling cudaq.control or cudaq.adjoint on "
                "a kernel defined in C++ is not currently "
                "supported", node)
        break
```

sacpis (commented):
Left a comment.

python/cudaq/kernel/ast_bridge.py
atgeller: Ah, yes, looks like the logic got flipped. Thanks for the fix!

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2668,8 +2667,8 @@ def resolveQualifiedName(pyVal):
                         obj = module
                         for part in moduleNames[1:]:
                             obj = getattr(obj, part)
-                        devKey = f"{module_name}.{'.'.join(moduleNames[1:])}" if len(
-                            moduleNames) > 1 else module_name
+                        getattr(obj, pyVal.attr)
+                        devKey = obj.__name__
                     except AttributeError:
                         continue
             return devKey, pyVal.attr
```
sacpis: Should we return from inside the for loop? Or do we still want the last one to be returned?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2668,8 +2667,8 @@ def resolveQualifiedName(pyVal):
                         obj = module
                         for part in moduleNames[1:]:
                             obj = getattr(obj, part)
-                        devKey = f"{module_name}.{'.'.join(moduleNames[1:])}" if len(
-                            moduleNames) > 1 else module_name
+                        getattr(obj, pyVal.attr)
```
1tnguyen: Could we add a comment here to document the lookup logic, e.g., how it handles registered C++ kernels under a Python submodule, etc.?

1tnguyen (approved):
👍

python/cudaq/kernel/ast_bridge.py
atgeller: Makes sense to me, added.

python/cudaq/kernel/ast_bridge.py
atgeller: Done.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2661,18 +2661,50 @@ def resolveQualifiedName(pyVal):
...
+            # in case it was aliased (e.g., `import mod1 as mod2`)
+            obj = None
+            for frameinfo in inspect.stack():
+                frame = frameinfo.frame
+                if moduleNames[0] in frame.f_locals:
+                    obj = frame.f_locals.get(moduleNames[0])
+                elif moduleNames[0] in frame.f_globals:
+                    obj = frame.f_globals.get(moduleNames[0])
+                if obj is not None:
+                    # In case a module has been imported multiple times, grab the latest
```
schweitzpgi, line 2699: nit: yapf doesn't format to 80 columns properly.

PR #3918, opened 2026-02-11
Use nvidia proxy cache

mitchdz:
This change uses the [nv-gha-runners/setup-proxy-cache](https://github.com/nv-gha-runners/setup-proxy-cache)￼ workflow to enable the internal APT proxy on our runners.

The pip and conda proxies are enabled by default; APT must be explicitly opted into, which is what this PR configures.


1tnguyen (approved):
👍

PR #3922, opened 2026-02-11
[PTSBE] Use `dynamic_cast` for `sampleWithPTSBE` opt-in

1tnguyen:
### Description

As the NVQIR simulator is retrieved as a base class pointer in the runtime, it seems like dynamic casting is the solution for simulator implementations to provide `sampleWithPTSBE`.


sacpis (approved):
LGTM. Thanks @1tnguyen.

1tnguyen:
> 1. Should BatchSimulator be a subclass of simulator? On one hand I don't love multiple inheritance... On the other hand it would enable us to move the methods I took from protected to public back to protected in the simulator base class.
> 2. What are your thoughts on making the generic implementation the default for BatchSimulator and PTSBE explicit opt-in? It's a bit more boiler plate but would allow simulator to choose not to support PTSBE?

Since most simulators can run the generic implementation of PTSBE out of the box, I think making this `BatchSimulator` (with a default implementation) the base class for most simulators may introduce boilerplate code and require simulators to inherit from it (thus requiring them to know about PTSBE).  

About the opt-in mechanism for PTSBE, I guess simulators that support arbitrary gates (including custom unitary gates, provided as a matrix) should be able to support PTSBE fully. Maybe we could have that flag (e.g., as a getter method to indicate it can support arbitrary gates) at the base `CircuitSimulator` for subclasses to override, letting us know that it may not be able to support PTSBE in all scenarios, such as an arbitrary mixture of unitary channel provided by users.

PR #3928, opened 2026-02-11
Add operation names to noise model

taalexander:
### Description
Adds operation names to the cpp noise model. This is a helper function which for PTSBE allows human readable operation names as well as compatibility with simulators which apply noise/gates by name (eg., Stim). This should be API backward compatible. 

Pulled out from https://github.com/taalexander/cuda-quantum/pull/5


1tnguyen (approved):
LGTM 👍

PR #3930, opened 2026-02-11
[python] Move kernel signature inference out of MLIR compilation

### LINKED ISSUE #2895: [python] cudaq.sample API assumes broadcast on nested list arguments
Should `cudaq.sample` be changed to accept an explicit broadcast parameter?

Details:

`cudaq.sample` checks if the first argument is a 2-d array and broadcasts by passing elements of the array to the kernel.

This leads to confusion for a non-broadcasting sample that expects a nested list, for example:

```
@cudaq.kernel
def test_list_list(a: list[list[int]]):
    qc = cudaq.qvector(3)
    j = 0
    for i in a:
        if i[0] == 1:
            x(qc[j])
        j += 1
    mz(qc)

print(cudaq.sample(test_list_list, [[0,1],[1,0],[1,0]]))
```

As a workaround, we can make kernel take a non-list first parameter:

```python
@cudaq.kernel
def test_list_list(n: int, a: list[list[int]]):
    qc = cudaq.qvector(3)
    j = 0
    for i in a:
        if i[0] == 1:
            x(qc[j])
        j += 1
    mz(qc)

print(cudaq.sample(test_list_list, 2, [[0,1],[1,0],[1,0]]))
```

Related: https://github.com/NVIDIA/cuda-quantum/issues/1809
Related: https://github.com/NVIDIA/cuda-quantum/issues/2627

lmondada:
I've broken up this PR into two commits that can be reviewed independently:

The first commit contains "clean-ups". It doesn't introduce any new functionality nor should it change how anything is currently working. It just removes function arguments that aren't used, renames a couple of things and should make the code more readable. I'm including this here because it made my actual work much easier. These breaking changes are acceptable as they are within the `python/cudaq/kernel` folder, which is not publicly exposed (discussed with Bettina).

The second commit is the real contribution:
- The main goal is to parse the kernel signature _before_ running the MLIR compilation. This is required so that we can delay MLIR compilation until the first kernel invocation (in a future PR).
- To this end, an AST visitor in `python/cudaq/kernel/analysis.py` is extended to parse the function annotations. This can be run before MLIR compilation
- A type `KernelSignature` is introduced to consolidate all kernel type information that was scattered across many attributes in one place.
- `KernelSignature` now only stores the kernel types as MLIR types (before, both Python and MLIR types were stored). This makes it easier to handle as there is a single source of truth (and Python types can be recovered using `mlirTypeToPyType`)

EDIT: Added a third commit that fixes issue #2895

I've noted that a lot of this Python code uses CamelCase, even though our style guide ([Google style guide](https://github.com/NVIDIA/cuda-quantum/issues/2895)) recommends snake_case. I'm using snake_case where new names/variables are introduced, but limited the number of existing names that had to be changed.

closes #2895

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -16,6 +16,8 @@ import sys
 import cudaq
 from cudaq import spin
 
+from test_helpers import h2_hamiltonian_4q
+
 
 @pytest.fixture(autouse=True)
 def do_something():
@@ -349,15 +351,7 @@ def test_observe():
 
 def test_pauli_word_input():
 
-    h2_data = [
-        3, 1, 1, 3, 0.0454063, 0, 2, 0, 0, 0, 0.17028, 0, 0, 0, 2, 0, -0.220041,
-        -0, 1, 3, 3, 1, 0.0454063, 0, 0, 0, 0, 0, -0.106477, 0, 0, 2, 0, 0,
-        0.17028, 0, 0, 0, 0, 2, -0.220041, -0, 3, 3, 1, 1, -0.0454063, -0, 2, 2,
-        0, 0, 0.168336, 0, 2, 0, 2, 0, 0.1202, 0, 0, 2, 0, 2, 0.1202, 0, 2, 0,
-        0, 2, 0.165607, 0, 0, 2, 2, 0, 0.165607, 0, 0, 0, 2, 2, 0.174073, 0, 1,
-        1, 3, 3, -0.0454063, -0, 15
-    ]
-    h = cudaq.SpinOperator(h2_data, 4)
+    h = h2_hamiltonian_4q()
 
     @cudaq.kernel
     def kernel(theta: float, var: cudaq.pauli_word):
@@ -397,15 +391,7 @@ def test_pauli_word_input():
 
 
 def test_exp_pauli():
-    h2_data = [
-        3, 1, 1, 3, 0.0454063, 0, 2, 0, 0, 0, 0.17028, 0, 0, 0, 2, 0, -0.220041,
-        -0, 1, 3, 3, 1, 0.0454063, 0, 0, 0, 0, 0, -0.106477, 0, 0, 2, 0, 0,
-        0.17028, 0, 0, 0, 0, 2, -0.220041, -0, 3, 3, 1, 1, -0.0454063, -0, 2, 2,
-        0, 0, 0.168336, 0, 2, 0, 2, 0, 0.1202, 0, 0, 2, 0, 2, 0.1202, 0, 2, 0,
-        0, 2, 0.165607, 0, 0, 2, 2, 0, 0.165607, 0, 0, 0, 2, 2, 0.174073, 0, 1,
-        1, 3, 3, -0.0454063, -0, 15
-    ]
-    h = cudaq.SpinOperator(h2_data, 4)
+    h = h2_hamiltonian_4q()
 
     @cudaq.kernel
     def kernel(theta: float):
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -1,4 +1,7 @@
 on:
+  # Daily rebuild on main to ensure devdeps rebuilt as needed
+  schedule:
+    - cron: '0 4 * * *'  # aim to be ready by 9am CET/CEST
   workflow_dispatch:
     inputs:
       cache_base:
--- a/.github/workflows/dev_environment_macos.yml
+++ b/.github/workflows/dev_environment_macos.yml
@@ -50,7 +50,8 @@ jobs:
       platform_tag: ${{ steps.build_info.outputs.platform_tag }}
       owner: ${{ steps.build_info.outputs.owner }}
       pr_number: ${{ steps.build_info.outputs.pr_number }}
-      image_name: ${{ steps.build_info.outputs.image_name }}
+      scripts_hash: ${{ steps.build_info.outputs.scripts_hash }}
+      image_name_base: ${{ steps.build_info.outputs.image_name_base }}
 
     environment:
       name: ${{ inputs.environment || 'default' }}
@@ -89,28 +90,28 @@ jobs:
           # when the copied files change), macOS uses ORAS artifact caching with no
           # content-based invalidation. This hash ensures cache is rebuilt when
           # build scripts change. Update this list if new files affect the build.
-          deps_hash=$(cat \
+          scripts_hash=$(cat \
             scripts/install_prerequisites.sh \
             scripts/build_llvm.sh \
             scripts/set_env_defaults.sh \
             | sha256sum | cut -c1-8)
-          echo "deps_hash=$deps_hash"
+          echo "scripts_hash=$scripts_hash" >> $GITHUB_OUTPUT
 
           # Registry and image name
-          image_name=ghcr.io/${repo_owner,,}/${{ vars.packages_prefix }}cuda-quantum-macos-devdeps
+          image_name_base=ghcr.io/${repo_owner,,}/${{ vars.packages_prefix }}cuda-quantum-macos-devdeps
           if [ -n "$pr_number" ]; then
-            image_name=${image_name}-ci
+            image_name_base=${image_name_base}-ci
           fi
 
           registry_cache_base=$(echo ${{ inputs.registry_cache_from || github.event.pull_request.base.ref || 'main' }} | tr / -)
-          image_name="${image_name}:${platform_tag}-${registry_cache_base}-${deps_hash}"
-          echo "image_name=$image_name" >> $GITHUB_OUTPUT
+          image_name_base="${image_name_base}:${platform_tag}-${registry_cache_base}-"
+          echo "image_name_base=$image_name_base" >> $GITHUB_OUTPUT
 
   build:
     name: Build macOS Cache
     needs: metadata
     runs-on: ${{ needs.metadata.outputs.runner }}
-    timeout-minutes: 180
+    timeout-minutes: 240
     permissions:
       contents: read
       packages: write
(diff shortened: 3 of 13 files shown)


lmondada:
EDIT: done

~Note: I've already rebased on top of https://github.com/NVIDIA/cuda-quantum/pull/3932 but am waiting for that to get merged in. In the meantime ignore commit `cec7875` as it has already been reviewed. Will do a final rebase on `main` when that's possible.~

python/cudaq/kernel/ast_bridge.py
```diff
@@ -378,7 +378,13 @@ class keeps track of a MLIR Value stack that is pushed to and popped from
     MLIR code.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self,
+                 knownResultType=None,
```
khalatepradnya: Curious as to why only `knownResultType` is positional argument... (yet to look at the rest of the changes)

python/cudaq/kernel/kernel_builder.py
```diff
@@ -1402,22 +1402,6 @@ def resolve_callable_arg(self, insPt, target):
...
-        measurement (:class:`QuakeValue`): The handle to the single qubit 
-            measurement instruction.
-        function (Callable): The function to conditionally apply to the 
-            :class:`Kernel`.
-
-        Raises:
-        RuntimeError: No longer supported.
-        """
-        emitFatalError(
-            "`c_if` is no longer supported. Use kernel mode with `run` API.")
```
khalatepradnya, line 1420: Okay to remove this. The intention to keep was so that it shows the error message if anyone tries to use it.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -450,24 +456,22 @@ def handle_call_arguments(self, *args):
...
-                    qkeModule = Module.parse(maybeCode,
-                                             context=self.qkeModule.context)
-                    processedArgs.append(
-                        LinkedKernelCapture(linkedKernel, qkeModule))
-                else:
-                    a_value = recover_value_of(a, None)
-                    self.process_argument(processedArgs, i, a_value,
-                                          callingModule)
+        for arg in self.signature.captured_args:
+            # get the value associated with the variable named "a" in the
```
khalatepradnya: This comment should be updated since "a" was removed, no?

python/cudaq/kernel/kernel_builder.py
lmondada: You're right, I'll add this back.

python/cudaq/kernel/ast_bridge.py
lmondada: In the second commit, this is converted into a `signature` argument, which is a required argument -- hence positional.

python/cudaq/kernel/kernel_builder.py
lmondada: done :)

python/cudaq/kernel/kernel_signature.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @staticmethod
+    def parse_from_mlir(mlir_module: mlir.Module,
+                        kernel_name: str,
+                        compiled: bool = True) -> "KernelSignature":
+        """
+        Parse the signature of a CUDA-Q kernel from an MLIR module.
+        
+        If `compiled` is `True`, the kernel has been compiled and the captured
+        arguments list is assumed to be empty.
```
khalatepradnya: Q: Can you help me understand what's the difference between scenarios when `captured_args` is empty v/s None ?

python/tests/kernel/test_kernel_signature.py
```diff
@@ -0,0 +1,192 @@
...
+    # This should raise an error during pre_compile
+    with pytest.raises(RuntimeError,
+                       match="must have argument type annotations"):
+        cudaq.kernel(my_kernel)
+
+
+# leave for gdb debugging
+if __name__ == "__main__":
+    loc = os.path.abspath(__file__)
+    pytest.main([loc, "-rP"])
```
khalatepradnya, line 376: Nice set of tests! 👍
Can you add cases for complex types, dataclasses, callables, np arrays, nested arguments, aggregate return types etc. perhaps?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -77,18 +79,10 @@ def __init__(self,
...
                  module=None,
                  kernelName=None,
-                 funcSrc=None,
                  signature=None,
                  location=None,
                  overrideGlobalScopedVars=None,
-                 decorator=None,
-                 fromBuilder=False):
-
-        if funcSrc is not None:
```
taalexander, line 87: This has gone through deprecation period?

python/cudaq/kernel/analysis.py
```diff
@@ -8,27 +8,40 @@
...
     """
-    This visitor will visit the function definition and report 
-    true if that function has a return statement.
+    This visitor will visit the function definition of `kernel_name` and report 
+    type annotations and whether the function has a return statement.
     """
 
-    def __init__(self):
-        self.hasReturnNode = False
+    arg_annotations: list[(str, Type)] = []
```
taalexander: Needs to be initialized in __init__ as right now this list will be shared by all instances at the class level (it is overwritten further down but this is fragile especially if function def is not found?).

python/tests/kernel/test_kernel_features.py
```diff
@@ -814,22 +814,39 @@ def kernel(n: int, s: list[list[str]]):
...
+    def kernel(l: list[list[int]]):
+        q = cudaq.qvector(2)
+        for inner in l:
+            for i in inner:
+                x(q[i])
+
+    # passing a list of lists works
+    counts = cudaq.sample(kernel, [[0, 1]])
+    # passing a 3-nested list works (broadcasting)
+    counts = cudaq.sample(kernel, [[[0, 1]], [[0, 1]]])
```
khalatepradnya: While I understand the issue this is trying to fix, this syntax is confusing IMO.
May I suggest moving out this fix into its own PR and updating docs with the clarification of the syntax? (May need arch team inputs).

python/cudaq/kernel/kernel_signature.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if visitor.return_annotation is None and visitor.has_return_statement:
+            raise KernelSignatureError(
+                f"CUDA-Q kernel {kernel_name} has return statement but no return type annotation"
+            )
+
+        arg_types = [
+            _mlir_type_from_annotation(annotation)
+            for _name, annotation in visitor.arg_annotations
+        ]
```
taalexander, line 107: Are these types from the MLIR context? What happens on reset?

khalatepradnya (commented):
Suggest moving commit# 3 in a separate PR.

taalexander (commented):
At a high-level this looks good to me. As I am unfamiliar I will leave to others to approve.

python/cudaq/kernel/kernel_decorator.py
lmondada: good catch, thank you! Removed

python/cudaq/kernel/kernel_signature.py
lmondada: Yes, that's an important point. I've made it clearer.

- I've added the following comment on the attribute `captured_args`:
```python
    # The list of captured arguments of a kernel is only known after the kernel
    # has been compiled. We thus distinguish between two false-y values:
    # - `captured_args = None` when the capture list is unknown (for kernels
    #    before compilation)
    # - `captured_args = []` when the capture list is known and empty (for
    #    kernels after compilation)
```
- I've removed the argument `compiled=True` in `parse_from_mlir` and instead always assume it's true (the False case wasn't used and doesn't make much sense). I've expanded the docs to:
```python
    @staticmethod
    def parse_from_mlir(mlir_module: mlir.Module,
                        kernel_name: str) -> "KernelSignature":
        """
        Parse the signature of a CUDA-Q kernel from an MLIR module.
        
        The kernel is assumed to have no captured arguments, i.e. all arguments
        of the MLIR `FuncOp` are treated as direct arguments to the kernel and
        `captured_args` is set to `[]`.
        """
```
- I've expanded the docs of `parse_from_ast` to
```python
    @staticmethod
    def parse_from_ast(ast_module: ast.Module,
                       kernel_name: str) -> "KernelSignature":
        """
        Parse the signature of a CUDA-Q kernel from a Python AST.
        
        This does not perform any capture analysis (typically done during MLIR
        compilation) and thus does not populate `captured_args`, the list of
        arguments captured by the kernel.
        """
```

I hope this clears it up!

python/tests/kernel/test_kernel_signature.py
lmondada: Sure thing, I've now added a whole bunch of additional tests covering these cases :)

python/tests/kernel/test_kernel_features.py
lmondada: Ok sure, will do. Sorry for mixing this in.

lmondada: I have reverted this change and kept the previous behaviour, where anything other than a flat list is not considered a broadcast. I'll create a separate PR to discuss this.

python/cudaq/kernel/kernel_decorator.py
lmondada: I discussed with Bettina. I had previously left this in out of caution, but Bettina confirmed that this is not a public API and has gone through plenty of breaking changes since the last release -- hence no need to keep this around.

python/cudaq/kernel/analysis.py
lmondada: ah, those Python footguns! My bad, thanks for spotting it.

python/cudaq/kernel/kernel_signature.py
lmondada: EDIT: This is fine, because we do not reset the context, we just clear live operations. Types remain alive.

~so... good point. Right now this is maybe fine, because we compile the MLIR right after type inference and context is only ever cleared after compilation.~

~That being said, I don't know what happens to these types if someone accesses them later (say when a second kernel wants to call into the first, after it has been compiled). This will become even more of a problem if we defer compilation until invocation, in which case we won't be able to guarantee that there hasn't been another kernel compilation in-between that has cleared the context.~

~I need to understand how MLIR contexts work and we might need a bigger discussion on this. If it is true that clearing the context invalidates MLIR types, then current `main` already contains this problem, as type signatures may be accessed long after compilation...~

PR #3932, opened 2026-02-11
[python] Remove unused CapturedDataStorage

lmondada:
I've split this into a separate PR because I am not sure whether this should be removed.

I can see in the codebase that `CapturedDataStorage` is not used anymore. This PR would remove it.

Is it correct that this is no longer required, or is there a dependency I am missing? I have checked that it is not used in cudaqx either.


sacpis (approved):
LGTM. Thanks @lmondada.

1tnguyen (approved):
LGTM 👍

PR #3933, opened 2026-02-11
Type erased jit engine

Renaud-K:
With this change, the mlir::ExecutionEngine is now wrapped in a type-erased class that is:
- Removing the MLIR dependency for the clients of the class at link time. The client's shared library does not need to link with MLIR.
- Managing memory internally which greatly simplifies memory management for the jit engines.
- The python flow can now hold an optional JitEngine for caching and caching is simplified throughout.
- Caching can now more easily be extended, since it just needs hold one of these objects.
 
```
class JitEngine {
public:
  JitEngine(std::unique_ptr<mlir::ExecutionEngine>);
  void run(const std::string &kernelName) const;
  void (*lookupRawNameOrFail(const std::string &kernelName) const)();

private:
  class Impl;
  std::shared_ptr<Impl> impl;
};
```

This is the last out hold-out preventing QPUs and compilers to be decoupled.

### DIFF
--- a/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
+++ b/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
@@ -965,7 +965,7 @@ marshal_and_retain_module(const std::string &name, MlirModule module,
                                                      resTy, cachedEngine);
   clone.erase();
   // `streamlinedSpecializeModule` should always set the cached engine pointer
-  if (cachedEngine)
+  if (!cachedEngine)
     throw std::runtime_error("Failed to retrieve the JIT engine pointer when "
                              "specializing the module.");
   // Use address of the allocated `ExecutionEngine` as the hash key to cache the


Renaud-K:
> /ok to test e5e29449ab51893f4ca1f5d34c60be18d5c8bba0

python/runtime/interop/PythonCppInterop.cpp
```diff
@@ -12,11 +12,7 @@
 #include "mlir/ExecutionEngine/ExecutionEngine.h"
 
 cudaq::python::CppPyKernelDecorator::~CppPyKernelDecorator() {
```
lmondada, line 13: Alternatively just write `~CppPyKernelDecorator() = default;` in the header?

runtime/common/ExecutionContext.h
```diff
@@ -157,10 +158,8 @@ class ExecutionContext {
   bool allowJitEngineCaching = false;
 
   /// For performance, a launcher may cache the JIT execution engine and use it
-  /// for multiple discrete calls. This is actually a pointer to a
-  /// `mlir::ExecutionEngine` object, but we hide that because of problems with
-  /// the structure and organization of the runtime libraries.
-  void *jitEng = nullptr;
+  /// for multiple discrete calls.
+  std::optional<JitEngine> jitEng = std::nullopt;
```
lmondada: I've learnt that whatever we put in here automatically lands in the public docs [here](https://nvidia.github.io/cuda-quantum/latest/api/languages/cpp_api.html#_CPPv4N5cudaq16ExecutionContextE). I expect this will also cause you docs build failures.

Consider putting it after the `@cond HIDDEN_MEMBERS` below to avoid this issue. My hope is that eventually ExecutionContext won't be on the front page of our public APIs, but we are not there yet :)

runtime/common/JIT.cpp
```diff
@@ -144,3 +147,39 @@ cudaq::createWrappedKernel(std::string_view irString,
...
+
+namespace cudaq {
+class JitEngine::Impl {
+public:
+  Impl(std::unique_ptr<mlir::ExecutionEngine> jitEngine)
+      : jitEngine(std::move(jitEngine)) {}
+  void run(const std::string &kernelName) const {
+    auto funcPtr = lookupRawNameOrFail(
+        std::string(cudaq::runtime::cudaqGenPrefixName) + kernelName);
+    reinterpret_cast<void (*)()>(*funcPtr)();
```
lmondada: I don't think you need this type casting, given that you have already cast the returned type of `lookupRawNameOrFail`.

python/runtime/interop/PythonCppInterop.h
```diff
@@ -34,8 +36,7 @@ class CppPyKernelDecorator {
   T getEntryPointFunction(As... as) {
     // Perform beta reduction on the kernel decorator.
     void *p =
-        kernel
-            .attr("beta_reduction")(&execution_engine, std::forward<As>(as)...)
+        kernel.attr("beta_reduction")(&jit_engine, std::forward<As>(as)...)
```
atgeller: For the python interface, it is desirable for the JIT engine to be able to outlive the lifetime of the `CppPyKernelDecorator` it belongs to, is that possible with this approach?

Renaud-K: Well, this is related to https://github.com/NVIDIA/cuda-quantum/pull/3908 from @1tnguyen . This PR re-introduces the `JITExecutionEngine` and collides with this one. I was not aware of that change and mine was just meant to be an NFC. I am now in the process of merging these changes to adapt to the `JITExecutionEngine`.

python/runtime/interop/PythonCppInterop.cpp
Renaud-K: I know. There was so much code that I felt like I needed to comment why it was removed. :)

python/runtime/interop/PythonCppInterop.h
Renaud-K: But, if I look closer, I don't think we take advantage of the `JITExecutionEngine` since we delete the entry in the decorator's destructor, which is equivalent to keeping it local to the decorator.
```
cudaq::python::CppPyKernelDecorator::~CppPyKernelDecorator() {
  if (cachedEngineKey.has_value()) {
    kernel.attr("delete_cache_execution_engine")(cachedEngineKey.value());
  }
}
```

And I guess, it will not hurt to extend my changes to the `JITExecutionEngine` for future development.

runtime/common/ExecutionContext.h
Renaud-K: Done. Thank you.

python/runtime/interop/PythonCppInterop.cpp
Renaud-K: https://github.com/NVIDIA/cuda-quantum/pull/3908 is re-introducing a destructor and I am merging in these changes.

runtime/common/JIT.cpp
Renaud-K: Removed it.

python/runtime/interop/PythonCppInterop.h
atgeller: Sounds good. Thanks!

PR #3938, opened 2026-02-12
[Python] Unify `StateMemoryView` and `State`

1tnguyen:
### Description

Having a separate `StateMemoryView` class for the Python buffer interface of `State` would be a breaking change. Hence, merging the buffer interface back to the `State` binding.

### DIFF
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -1,4 +1,7 @@
 on:
+  # Daily rebuild on main to ensure devdeps rebuilt as needed
+  schedule:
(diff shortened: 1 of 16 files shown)


sacpis:
Seems like we will need to update `python/tests/parallel/test_mpi_dynamics.py` with this change as well.

1tnguyen:
> Seems like we will need to update `python/tests/parallel/test_mpi_dynamics.py` with this change as well.

Indeed. Somehow, I lost that file when staging the changes. Fixed in 2f6f4a3029f60838387d9a1c1659acf1cb846648.

schweitzpgi:
Thanks, Thien.

PR #3939, opened 2026-02-12
[PTSBE] Fix PTSBESampler

sacpis:
Use op_names as vector instead of pointer

Fixes: 
- Op names
https://github.com/NVIDIA/cuda-quantum/actions/runs/21951226632/job/63403864188#step:12:8734

- Op names in merge tests and exception type
https://github.com/NVIDIA/cuda-quantum/actions/runs/21953508304/job/63433299570?pr=3939#step:12:24879


1tnguyen (approved):
👍

PR #3940, opened 2026-02-12
[python] Use separate context for kernel builder

lmondada:
This PR fixes the following crash

```python
import cudaq

kernel = cudaq.make_kernel()

@cudaq.kernel()
def kernel2():
    pass

kernel.qalloc(1) # this (or any other MLIR modification) crashes
```

Error: `RuntimeError: the operation has been invalidated`

The reason for this is that kernel2 was getting compiled using `compile_to_mlir` in `python/cudaq/kernel/ast_bridge.py`. This calls `context._clear_live_operations()` at the end of the function. However, the context is global, thus it also clears the context for the kernel builder. This invalidates the insertion point stored in the builder, hence any further modifications of the kernel builder fails.

This PR proposes to use a separate context for the kernel builder. When a kernel builder gets invoked and needs to be jitted, we then output the module as string and reload it into the global context. This will come at some performance cost, but only affects kernels defined using the builder and only at the first invocation (as the compiled MLIR module is cached).

### DIFF
--- a/python/cudaq/kernel/kernel_builder.py
+++ b/python/cudaq/kernel/kernel_builder.py
@@ -32,10 +32,10 @@ from .common.givens import givens_builder
 from .kernel_decorator import isa_kernel_decorator
 from .quake_value import QuakeValue
 from .utils import (emitFatalError, emitWarning, nvqppPrefix, getMLIRContext,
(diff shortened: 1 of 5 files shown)


lmondada:
We'll have to discuss a plan to fix this. For now I've disabled using the kernel builder with kernels that have deferred compilation.

PR #3942, opened 2026-02-12
[PTSBE] Integration tests

sacpis:
Adding a dedicated Python test package for PTSBE (probabilistic trajectory-based sampling) under python/tests/ptsbe.

### DIFF
--- a/python/tests/ptsbe/test_noise_analytical.py
+++ b/python/tests/ptsbe/test_noise_analytical.py
@@ -39,24 +39,15 @@ def cnot_echo():
     mz(q)
 
 
-def _total_shots(result):
-    return sum(result.count(bs) for bs in result)
-
-
-def _probability(result, bitstring):
-    total = _total_shots(result)
-    return result.count(bitstring) / total if total else 0.0
-
-
 def test_check_bit_flip_type():
     cudaq.set_random_seed(42)
     noise = cudaq.NoiseModel()
     noise.add_channel("x", [0], cudaq.BitFlipChannel(0.1))
     result = cudaq.ptsbe.sample(x_op, noise_model=noise, shots_count=2000)
-    assert _total_shots(result) == 2000
+    assert sum(result.count(bs) for bs in result) == 2000
     assert len(result) == 2
-    p0 = _probability(result, "0")
-    p1 = _probability(result, "1")
+    p0 = result.probability("0")
+    p1 = result.probability("1")
     assert abs(p0 - 0.1) <= 0.1
     assert abs(p1 - 0.9) <= 0.1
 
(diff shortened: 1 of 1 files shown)


python/tests/ptsbe/async_broadcast.py
```diff
@@ -0,0 +1,46 @@
...
+    shots = 200
+    future = cudaq.ptsbe.sample_async(bell,
+                                      noise_model=noise,
+                                      shots_count=shots)
+    result = future.get()
+    assert isinstance(result, cudaq.SampleResult)
+    total = sum(result.count(bs) for bs in result)
+    assert total == shots
+    bell_counts = result.count("00") + result.count("11")
+    assert bell_counts > shots * 0.8
```
1tnguyen: I think we need to revise the condition to check for noisy results, not just bell states, e.g., `bell_counts < shots`.

python/tests/ptsbe/async_broadcast.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_ptsbe_broadcast():
+    noise = cudaq.NoiseModel()
+    noise.add_all_qubit_channel("ry", cudaq.DepolarizationChannel(0.01))
```
1tnguyen, line 31: Can we also add a bit-flip channel test, whereby we can validate the noisy results?

e.g., 10% bit flip (0.1), then we can check the correct bit probability ~ 0.9 and the flipped ~ 0.1.

python/tests/ptsbe/async_broadcast.py
sacpis: Fixed in latest commits.

python/tests/ptsbe/async_broadcast.py
sacpis: Fixed in latest commits.

python/tests/ptsbe/test_noise_analytical.py
```diff
@@ -0,0 +1,100 @@
...
+    mz(q)
+
+
+def _total_shots(result):
+    return sum(result.count(bs) for bs in result)
+
+
+def _probability(result, bitstring):
+    total = _total_shots(result)
+    return result.count(bitstring) / total if total else 0.0
```
1tnguyen: I thought we already had a `probability()` method on the sample result to compute this.

sacpis: Fixed in [d9cebdf](https://github.com/NVIDIA/cuda-quantum/pull/3942/commits/d9cebdfcb096c6dd3c4b59c131cd773d6991c611).

PR #3944, opened 2026-02-13
[PTSBE] Integration tests with edge cases

sacpis:
Addressing feedback received from @taalexander for PR #3942. Adding more edge cases for the integration tests.


python/tests/ptsbe/conftest.py
```diff
@@ -31,5 +22,31 @@ def density_matrix_target():
...
+
+
+@pytest.fixture
+def phase_flip_kernel_fixture():
+    return phase_flip_kernel
+
+
+@pytest.fixture
+def cnot_echo_kernel():
+    return cnot_echo
```
taalexander: Could the kernels just be defined within and drop `.kernels`? If that doesn't work having them as fixtures might be redundant.

python/tests/ptsbe/test_errors.py
```diff
@@ -0,0 +1,40 @@
...
+def test_ptsbe_broadcast_empty_args_returns_empty_list():
+    noise = cudaq.NoiseModel()
+    noise.add_all_qubit_channel("ry", cudaq.DepolarizationChannel(0.01))
+    results = cudaq.ptsbe.sample(
+        rotation_kernel,
+        [],
+        noise_model=noise,
+        shots_count=10,
+    )
+    assert results == []
```
taalexander: Can the kernels be consumed as fixtures? Similarily, noise models could be used in this way but not required.

taalexander (approved):
Thanks, should we split the existing tests in `test_ptsbe` out to the matching test file you created deleting tests if there are duplicates?

python/tests/ptsbe/conftest.py
sacpis: Fixed in a0fd4e2.

python/tests/ptsbe/test_errors.py
sacpis: Fixed in a0fd4e2.

sacpis:
> Thanks, should we split the existing tests in `test_ptsbe` out to the matching test file you created deleting tests if there are duplicates?

Fixed in [2c8821a](https://github.com/NVIDIA/cuda-quantum/pull/3944/commits/2c8821a3c4e30f699ee9f65628e7fad137fff912).

python/tests/ptsbe/test_ptsbe.py
```diff
@@ -5,33 +5,9 @@
 # This source code and the accompanying materials are made available under     #
```
taalexander: There's only one test here at this point. We should likely just remove this module.

python/tests/ptsbe/test_execution_data.py
```diff
@@ -0,0 +1,185 @@
...
+def bell():
+    q = cudaq.qvector(2)
+    h(q[0])
+    x.ctrl(q[0], q[1])
+    mz(q)
+
+
+@pytest.fixture
+def bell_kernel():
+    return bell
```
taalexander: The same kernels are now duplicated in most tests? Why not just put them in conftest?

taalexander (changes requested):
A few small changes left.

python/tests/ptsbe/test_ptsbe.py
sacpis: Fixed in f27ae82.

python/tests/ptsbe/test_execution_data.py
sacpis: Fixed in f27ae82.

PR #3948, opened 2026-02-13
Defer Python kernel compilation until invocation

lmondada:
This PR introduces "deferred compilation" in Python. By default, kernels are no longer compiled to MLIR (aka pre-compiled) at kernel definition time, but only when the kernel is invoked for the first time. As before, the MLIR module is then cached for further invocation. For example,

```python
import cudaq

@cudaq.kernel
def foo():
    pass
```

would not trigger any compilation, until `foo()` is called for the first time. This enables:
1. Faster package load times: `import cudaq` takes ~4s now, versus `~8s` before, as the kernels included with `cudaq` are no longer compiled when the package is loaded. This will also apply to any other library that provides kernel definitions.
2. Kernels can be defined out-of-order: previously, calling kernel B from kernel A would error if B was defined after A. This is now supported (see tests).
3. Captured variables that aren't defined at kernel definition time but are defined at invocation time are now supported.

Compilation can be forced (thus recovering the old behaviour) using the `defer_compilation=False` flag. The following:

```python
@cudaq.kernel(defer_compilation=False)
def foo():
    pass
```

triggers compilation immediately.

## Limitations
To limit the scope of this PR, I introduced two limitations (see tests for examples):
1. Recursive kernel calls are still not supported. The way kernels are captured and lifted into arguments would have to handle recursive calls specially.
2. A kernel builder A cannot call another kernel B using `apply_call(B)` if B wasn't compiled beforehand. In that case, the user will receive a clear error suggesting to either use `defer_compilation=False` or calling `B.compile()` directly, e.g.
```python
@cudaq.kernel(defer_compilation=True)
def notPrecompiledKernel():
    pass

kernel = cudaq.make_kernel()
kernel.apply_call(notPrecompiledKernel)  # fails. User must set `defer_compilation=False`
                                         # or call `notPrecompiledKernel.compile()`
```

### DIFF
--- a/python/tests/builder/test_kernel_builder.py
+++ b/python/tests/builder/test_kernel_builder.py
@@ -1274,8 +1274,7 @@ def test_call_kernel_deferred_compilation():
         kernel.apply_call(notPrecompiledKernel)
 
     assert "must be compiled to be used in the kernel builder" in str(e.value)
-    assert "Call `notPrecompiledKernel.compile()` before initialzing" in str(
-        e.value)
+    assert "notPrecompiledKernel" in str(e.value)
     assert "or deactivate deferred compilation" in str(e.value)
 
 
--- a/python/tests/kernel/test_assignments.py
+++ b/python/tests/kernel/test_assignments.py
@@ -333,26 +333,27 @@ def test_dataclass_update_failures():
         controls: cudaq.qview
         target: cudaq.qubit
 
-    # We do not currently allow any kind of updates to
-    # quantum structs.
-    @cudaq.kernel
-    def test1(t: MyQTuple, controls: cudaq.qview):
-        t.controls = controls
-
     with pytest.raises(RuntimeError) as e:
-        test1.compile()
 
+        # We do not currently allow any kind of updates to
+        # quantum structs.
+        @cudaq.kernel
+        def test1(t: MyQTuple, controls: cudaq.qview):
+            t.controls = controls
+
+        print(test1)
     assert 'accessing attribute of quantum tuple or dataclass does not produce a modifiable value' in str(
         e.value)
     assert '(offending source -> t.controls)' in str(e.value)
 
-    @cudaq.kernel
-    def test2(arg: MyQTuple, controls: cudaq.qview):
-        t = arg.copy()
-        t.controls = controls
-
     with pytest.raises(RuntimeError) as e:
-        test2.compile()
+
+        @cudaq.kernel
+        def test2(arg: MyQTuple, controls: cudaq.qview):
+            t = arg.copy()
+            t.controls = controls
+
+        print(test2)
     assert 'copy is not supported' in str(e.value)
     assert '(offending source -> arg.copy())' in str(e.value)
 
@@ -361,54 +362,54 @@ def test_dataclass_update_failures():
         angle: float
         idx: int
 
-    @cudaq.kernel
-    def update_tuple1(t: MyTuple):
-        t.angle = 5.
-
-    @cudaq.kernel
-    def test3() -> MyTuple:
-        t = MyTuple(0., 0)
-        update_tuple1(t)
-        return t
-
     with pytest.raises(RuntimeError) as e:
-        test3.compile()
 
+        @cudaq.kernel
+        def update_tuple1(t: MyTuple):
+            t.angle = 5.
+
+        @cudaq.kernel
+        def test3() -> MyTuple:
+            t = MyTuple(0., 0)
+            update_tuple1(t)
+            return t
+
+        print(test3)
     assert 'value cannot be modified - use `.copy(deep)` to create a new value that can be modified' in str(
         e.value)
     assert '(offending source -> t.angle)' in str(e.value)
 
-    @cudaq.kernel
-    def update_tuple2(t: MyTuple):
-        t.angle += 5.
-
-    @cudaq.kernel
-    def test4() -> MyTuple:
-        t = MyTuple(0., 0)
-        update_tuple2(t)
-        return t
-
     with pytest.raises(RuntimeError) as e:
-        test4.compile()
 
+        @cudaq.kernel
+        def update_tuple2(t: MyTuple):
+            t.angle += 5.
+
+        @cudaq.kernel
+        def test4() -> MyTuple:
+            t = MyTuple(0., 0)
+            update_tuple2(t)
+            return t
+
+        print(test4)
     assert 'value cannot be modified - use `.copy(deep)` to create a new value that can be modified' in str(
         e.value)
     assert '(offending source -> t.angle)' in str(e.value)
 
-    @cudaq.kernel
-    def update_tuple3(arg: MyTuple):
-        t = arg
-        t.angle = 5.
-
-    @cudaq.kernel
-    def test5() -> MyTuple:
-        t = MyTuple(0., 0)
-        update_tuple3(t)
-        return t
-
     with pytest.raises(RuntimeError) as e:
-        test5.compile()
 
+        @cudaq.kernel
+        def update_tuple3(arg: MyTuple):
+            t = arg
+            t.angle = 5.
+
+        @cudaq.kernel
+        def test5() -> MyTuple:
+            t = MyTuple(0., 0)
+            update_tuple3(t)
+            return t
+
+        print(test5())
     assert 'cannot assign dataclass passed as function argument to a local variable' in str(
         e.value)
     assert 'use `.copy(deep)` to create a new value that can be assigned' in str(
@@ -420,47 +421,47 @@ def test_dataclass_update_failures():
         val: MyTuple
         num: int
 
-    @cudaq.kernel
-    def test6() -> NumberedMyTuple:
-        t = MyTuple(0.5, 1)
-        return NumberedMyTuple(t, 0)
-
     with pytest.raises(RuntimeError) as e:
-        test6.compile()
 
+        @cudaq.kernel
+        def test6() -> NumberedMyTuple:
+            t = MyTuple(0.5, 1)
+            return NumberedMyTuple(t, 0)
+
+        test6()
     assert ('only dataclass literals may be used as items in other '
             'container values' in str(e.value))
     assert 'use `.copy(deep)` to create a new MyTuple' in str(e.value)
 
-    @cudaq.kernel
-    def test7(cond: bool) -> tuple[MyTuple, MyTuple]:
-        t1 = MyTuple(1, 1)
-        t2 = t1
-        if cond:
-            t3 = MyTuple(2, 2)
-            t1 = t3
-            t3.angle = 5
-        return (t1, t2)
-
     with pytest.raises(RuntimeError) as e:
-        test7.compile()
 
+        @cudaq.kernel
+        def test7(cond: bool) -> tuple[MyTuple, MyTuple]:
+            t1 = MyTuple(1, 1)
+            t2 = t1
+            if cond:
+                t3 = MyTuple(2, 2)
+                t1 = t3
+                t3.angle = 5
+            return (t1, t2)
+
+        test7(True)
     assert 'only literals can be assigned to variables defined in parent scope' in str(
         e.value)
     assert '(offending source -> t1 = t3)' in str(e.value)
 
-    @cudaq.kernel
-    def test8(cond: bool) -> MyTuple:
-        t1 = [MyTuple(1, 1)]
-        if cond:
-            t3 = MyTuple(2, 2)
-            t1[0] = t3
-            t3.angle = 5
-        return t1
-
     with pytest.raises(RuntimeError) as e:
-        test8(True)
 
+        @cudaq.kernel
+        def test8(cond: bool) -> MyTuple:
+            t1 = [MyTuple(1, 1)]
+            if cond:
+                t3 = MyTuple(2, 2)
+                t1[0] = t3
+                t3.angle = 5
+            return t1
+
+        test8(True)
     assert 'only dataclass literals may be used as items in other container values' in str(
         e.value)
     assert 'use `.copy(deep)` to create a new MyTuple' in str(e.value)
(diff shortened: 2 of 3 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -2571,21 +2571,6 @@ def processDecoratorCall(symName):
             # of eliminating unnecessary copies.
             return self.__migrateLists(result, copy_list_to_stack)
 
-        def processFunctionCall(kernel):
```
lmondada, line 2579: This was not used anywhere

python/tests/interop/test_interop.py
```diff
@@ -169,9 +169,9 @@ def callUCCSD(setControl: bool):
                 x(c)
             cudaq.control(cudaq_test_cpp_algo.qstd.uccsd, c, q, 2)
 
-        counts = cudaq.sample(callQftAndAnother, False)
+        counts = cudaq.sample(callUCCSD, False)
```
lmondada, line 172: This was throwing an error for the wrong reason!

python/tests/mlir/ast_attributes.py
```diff
@@ -66,7 +66,7 @@ def kernel3():
 def test_attribute_failures():
     try:
 
-        @cudaq.kernel
+        @cudaq.kernel(defer_compilation=False)
```
lmondada: In all these tests, `defer_compilation` was deactivated so that the `print` statements that follow print the MLIR correctly. This is because the `__str__` definition of `PyKernelDecorator` was changed to print the uncompiled kernel if it has not been compiled yet.

IMO it would be odd for a print statement to trigger compilation, but it would be another way of resolving this.

python/tests/kernel/test_assignments.py
```diff
@@ -333,27 +333,26 @@ class MyQTuple:
         controls: cudaq.qview
         target: cudaq.qubit
 
-    with pytest.raises(RuntimeError) as e:
```
lmondada, line 336: The diff in the tests ended up being quite verbose. Nothing has changed, other than the fact that compilation errors are only triggered when either `kernel.compile()` is called explicitly, or the kernel is defined with `defer_compilation=False`.

One notable exception is `test_var_capture`, where a case that was previously unsupported (capturing a variable in a function before defining it) is now accepted (because compilation happens later).

schweitzpgi:
> Captured variables that aren't defined at kernel definition time but are defined at invocation time are now supported.

This was already true. Symbols from EGB scopes were lambda lifted to be arguments to the kernel. This change allows EGB scoped variables to be _undefined_ at the point the kernel is defined. Since the kernel's code is only generated later, the type(s) of these symbol(s) isn't required and no prior declaration is needed.

schweitzpgi:
> Recursive kernel calls are still not supported. The way kernels are captured and lifted into arguments would have to handle recursive calls specially.

I'm not clear on why this is any different than any other kernel call.
```python
@cudaq.kernel
def foo():
  foo()
```
should record the call to itself as a lambda lifted callable argument when we build the code for `foo` in the AST bridge step. This is required to come first since we'd have no idea what symbols appear in the body of `foo` if we tried to resolve those symbols before building the kernel code.

Given that, we then must resolve all the lambda lifted arguments. In this example, that's the symbol `foo`. That symbol must resolve to the kernel decorator we're building which is bound to the same symbol. heh. But we just built that code, so there is nothing to do.

Even if we inject several layers of calls ($\ge 1$) in the middle, `foo` is `foo` and the recursion naturally terminates here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -421,6 +422,7 @@ def node_error(msg):
         self.isSubscriptRoot = False
         self.verbose = verbose
         self.currentNode = None
+        self.parentVariables = parentVariables or {}
```
schweitzpgi, line 430: Why are we bringing back snapshots of the Python interpreter frames?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -33,6 +34,42 @@
...
+
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
+        self._ensure_compiled()
+        return method(self, *args, **kwargs)
+
+    return wrapper
+
+
+def ensure_not_recursive(method):
```
schweitzpgi, line 51: Why is this needed?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -89,11 +128,12 @@ def __init__(self,
         self.kernelModuleName = None
         self.name = kernelName
         self.verbose = verbose
-        # The `qkeModule` will be the quake target independent ModuleOp
-        self.qkeModule = None
-        # The `nvqModule` will be (if present) the default simulation ModuleOp
-        self.nvqModule = None
+        # Caches the `qkeModule` property once compiled
+        self._cached_qkeModule = None
```
schweitzpgi, line 132: Why the attribute name change?

python/tests/interop/test_interop.py
schweitzpgi: There are quite a few tests like that. :)

python/tests/kernel/test_kernel_features.py
```diff
@@ -571,41 +571,39 @@ def test2(myList: List[int]):
...
+    @cudaq.kernel
+    def kernel(params: List[float]):
+        params.append(1.0)
 
-        @cudaq.kernel
-        def kernel():
-            l = [i for i in range(10)]
-            l.append(11)
+    with pytest.raises(RuntimeError):
+        kernel.compile()
```
schweitzpgi: It's really unfortunate to have to bring back this `.compile()` method crutch.

python/tests/mlir/ast_attributes.py
schweitzpgi: I'm sort of thinking `defer_compilation` should default to `False` given all the rework of tests needed here. Changing it seems very heavy handed.

We only really need deferred compilation in oddball corner cases like the notebooks, which are written inside-out, right?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -174,40 +225,53 @@ def captured_variables(self):
...
-        if self.qkeModule:
-            raise RuntimeError(self.name + " was already compiled")
-        self.uniqueId = id(self)
-        self.uniqName = self.name + ".." + hex(self.uniqueId)
-        self.qkeModule = compile_to_mlir(id(self),
-                                         self.astModule,
-                                         self.signature,
-                                         verbose=self.verbose,
-                                         location=self.location,
-                                         parentVariables=self.globalScopedVars,
```
schweitzpgi, line 198: Some of these things are simply bit rot that needs to be thrown away.

We should _never_ keep maps of random bonus symbols and maybe what they may have binded to at some arbitrary point in time. When building the code or running the kernel, we must resolve symbols properly. Keeping any of these on-the-side maps is not needed and extremely error prone.

schweitzpgi: That said, `globalScopedVars` is a bit of an oddball case. It is also tracking registered types, if I remember correctly. I don't think we _should_ have special on-the-side dictionaries here to maybe-kinda track these variables or the registered types, but getting rid of this is still a TODO item.

lmondada:
> I'm not clear on why this is any different than any other kernel call.
> [...]
> Given that, we then must resolve all the lambda lifted arguments. In this example, that's the symbol foo. That symbol must resolve to the kernel decorator we're building which is bound to the same symbol. heh. But we just built that code, so there is nothing to do.

First of all, recursion is currently not supported in `main`. This is because the symbol `foo` within the kernel definition:
```python
@cudaq.kernel
def foo():
    foo()
```
is undefined at the definition time of `foo` (which is when it currently gets AOT compiled). Similarly, foo calling bar calling foo (or any other recursive setup) would not compile, given that one of the kernels would have to reference a not-yet-defined kernel.

---

That being said, I agree that with this code change, supporting recursion becomes easier. The execution you outline is how it _should_ (and will) work, but the reason it currently does not work is that compilation and execution proceeds as follows:

1. Compile `foo`.
2. Find a reference to `foo` within the body. Lambda lift that into an argument. **Now `foo` takes a (captured) callable argument.**
3. Compiling completes successfully
4. Kernel `foo` is invoked, so the captured arguments must be resolved. Captured `foo` is resolved to the same decorator as being invoked.
5. Now we need to (recursively) resolve the captured arguments of captured `foo`. Here `foo` gets resolved again to the decorator... this is how we get stuck in a loop of repeatedly capturing foo.

We need to add a conditional in the resolution of captured arguments that checks whether `foo` has been resolved before (and keep a list of all resolved callables somewhere), so that we can resolve it without recursing. I've already agreed with Bettina that I will work on a PR for this next.

lmondada:
> > Captured variables that aren't defined at kernel definition time but are defined at invocation time are now supported.
>
> This was already true. Symbols from EGB scopes were lambda lifted to be arguments to the kernel. This change allows EGB scoped variables to be undefined at the point the kernel is defined. Since the kernel's code is only generated later, the type(s) of these symbol(s) isn't required and no prior declaration is needed.

What I meant is that the following previously threw an error

```python
@cudaq.kernel
def bar(i: int):
    q = cudaq.qvector(n)

n = 3
    
bar()
```

python/cudaq/kernel/ast_bridge.py
lmondada: This is because of the bug https://github.com/NVIDIA/cuda-quantum/issues/3965, where captured arguments are getting resolved in the calling scope instead of the scope that the kernel was defined in.

Though your comment makes me think that this change doesn't really address the root issue. I might try to create another PR that fixes https://github.com/NVIDIA/cuda-quantum/issues/3965. If we then rebase on that, this change might become redundant.

python/cudaq/kernel/kernel_decorator.py
lmondada: See my comment [below](https://github.com/NVIDIA/cuda-quantum/pull/3948#issuecomment-3926127724). I intend to create a follow-up PR that handles recursion properly.

python/cudaq/kernel/kernel_decorator.py
lmondada: The change from `qkeModule` to `_cached_qkeModule` is not visible from the outside, because of the
```python
@property
def qkeModule(self): ...
```
method that I added.

As a result, accessing `decorator.qkeModule` on kernels that have been AOT compiled will work just as it did before. However, instead of having `qkeModule` be set to `None` for kernels that have not been compiled yet, accessing `decorator.qkeModule` now triggers a compilation before returning the compiled module. The compiled module is stored in `_cached_qkeModule` for future accesses of `self.qkeModule`.

python/tests/kernel/test_kernel_features.py
lmondada: This is for testing ONLY. For the user, whenever `kernel` is being invoked, the compilation will happen transparently and should be invisible.

The reason we need `.compile` here in testing is because these tests don't invoke the kernel but instead do one of two things:
- they **print out** the kernel. Printing an uncompiled kernel does not force it to compile. We could change the printing logic to trigger a compilation, but I thought that would be strange.
- they check for errors thrown at compilation time. It is a natural (and unavoidable) consequence of deferred compilation that errors that were previously thrown at kernel definition time now no longer get thrown until first invocation (or explicit compilation).

Do you agree?

python/tests/mlir/ast_attributes.py
lmondada: I had discussed this with Bettina, she is fine with having it `True` by default.

See also my [comment further down](https://github.com/NVIDIA/cuda-quantum/pull/3948#discussion_r2826909980) about why this is an artifact of our testing and why end users would rarely need to force compilation like this.

Currently, there is exactly ONE case where a user would need this: it's if they choose to use their kernel within another kernel builder. This is only temporary. I will make a PR that implements lambda lifting for kernel builders, which will resolve this for that case, too. At that point, there should never be a case where forcing compilation is required.

python/cudaq/kernel/ast_bridge.py
lmondada: Update: I have made a new PR that solves #3965. It also reverts the addition of `parentVariables`.

-> https://github.com/NVIDIA/cuda-quantum/pull/4005

That PR is based on top of this one. I suggest we leave this PR as is and address this issue in that one. Ok?

python/cudaq/kernel/kernel_decorator.py
lmondada: Following these comments, I have gone ahead and resolved this "the proper way" in https://github.com/NVIDIA/cuda-quantum/pull/4005. It gets rid of `parentVariables` and resolves captured variables properly.

I suggest we review that PR once this one has been merged in, and defer this fix until then.

python/cudaq/kernel/ast_bridge.py
bettinaheim: Yep, saw that too - dropped out with another PR and wasn't deleted.

python/tests/kernel/test_assignments.py
bettinaheim: The reverse test change was done when pulling in the python compiler refactor - effectively, you can manually revert the change to have the failure expected during compilation and put it back to having it be expected during invocation like it was before.

python/cudaq/kernel/ast_bridge.py
bettinaheim: Agreed - good to get rid of it again, so I'll ignore this addition for this PR and look at 4005 to add a cleaner solution.

python/cudaq/kernel/kernel_decorator.py
bettinaheim: Same here - I'll ignore this for this PR with the expectation that the subsequent PR for supporting recursion revises this.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -276,19 +338,29 @@ def merge_quake_source(self, quakeText):
 
     def __str__(self):
```
bettinaheim, line 339: To avoid a breaking change, I'd suggest to have print actually access the qkeModule (i.e. trigger the AOT compilation if it wasn't built yet).

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -445,6 +517,7 @@ def convertStringsToPauli(self, arg):
     def formal_arity(self):
         return len(self.arg_types())
 
+    @ensure_compiled
```
bettinaheim, line 512: I like the use of a decorator for common validation (not sure about perf in general, but shouldn't be relevant here)

python/tests/builder/test_kernel_builder.py
```diff
@@ -1133,7 +1133,7 @@ def test_fermionic_swap_op():
 
 def test_call_kernel_expressions():
 
-    @cudaq.kernel()
+    @cudaq.kernel(defer_compilation=False)
```
bettinaheim, line 1136: This is needed so that the builder kernel can call it?

python/tests/builder/test_kernel_builder.py
```diff
@@ -1259,6 +1259,23 @@ def kernelThatTakesIntAndListListFloat(qubits: cudaq.qview, qbit: int,
...
+    # invalidate the references to MLIR objects that the kernel builder holds.
+    @cudaq.kernel(defer_compilation=True)
+    def notPrecompiledKernel():
+        pass
+
+    kernel = cudaq.make_kernel()
+    with pytest.raises(RuntimeError) as e:
+        kernel.apply_call(notPrecompiledKernel)
+
+    assert "must be compiled to be used in the kernel builder" in str(e.value)
```
bettinaheim, line 1276: Give a more concrete message for what the user needs to do. E.g. "... Use notPrecompiledKernel.compile() to compile it before creating the calling kernel using the builder."

python/tests/kernel/test_kernel_features.py
bettinaheim: Agreed that this is a testing artifact.

python/tests/mlir/ast_break.py
```diff
@@ -13,7 +13,7 @@
 
 def test_break():
 
-    @cudaq.kernel(verbose=False)
+    @cudaq.kernel(defer_compilation=False)
```
bettinaheim: Not needed if we keep the behavior of print as it is. Does print support options (e.g. for formatting) that we could exploit?

python/tests/kernel/test_deferred_compilation.py
bettinaheim, line 1: I like the tests.

schweitzpgi:
<s>Is this a new PR? I thought I already reviewed something like this and left comments.</s>
Nevermind. The GUI appears to be confused as I see my comments are still here.

python/cudaq/kernel/kernel_decorator.py
lmondada: ok, will revert this here and in all the IR tests.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: I see. I'll close #3965 as that is not a defect. Resolving the arguments _where they are used in the call_ is the correct Pythonic interpretation. We should not regress proper behavior that was just fixed to behave like something broken and strange.

python/tests/builder/test_kernel_builder.py
lmondada: yes. The `apply_call` method of the kernel builder will throw an error if the kernel being invoked has not been compiled. Eric has proposed a fix for this, which will involve lambda lifting the invoked kernel into a kernel argument that can be compiled separately afterwards. This is tracked in jira by ticket CUDAQ-496.

python/tests/builder/test_kernel_builder.py
lmondada: The full error is in kernel_builder.py, line 1354:

```python
                emitFatalError(
                    f"Kernel '{name}' must be compiled to be used in the kernel builder. "
                    f"Call `{name}.compile()` before initializing the kernel builder, "
                    f"or deactivate deferred compilation:\n\n"
                    f"    @cudaq.kernel(defer_compilation=False)\n"
                    f"    def {name}(...): ...\n")
```

so already along the lines of what you suggest. I will add an extra `assert` checking for another bit of this error string.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: Sorry, reading the example in #3965, I think the issue is much, much narrower than the title mislead me to believe.

- We absolutely, positively must resolve arguments at the call sites. That's just the way Python works.
- Now that we're deferring compilation, we may have bitten off more than we wanted to. A reference to another callable in E(nclosing) scope isn't really even seen until later.
- Obviously G(lobal) and B(uiltin) scopes are ubiquitous and need not be changed.

lmondada:
@schweitzpgi https://github.com/NVIDIA/cuda-quantum/pull/4005 resolves the issue you bring up.

It was already an issue before the changes that are proposed here: whilst compilation was working fine, at call time the captured arguments could not be resolved. So, either way, we do not get around keeping a reference to the frame defining the kernel, independently of when we compile it.

It's thus unrelated to the changes here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -426,6 +427,7 @@ def node_error(msg):
...
```
schweitzpgi, line 430: As #4005 fixes the resolution of lifted arguments bug, why do we need this on-the-side dictionary?

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2533,6 +2535,11 @@ def processDecorator(name, path=None):
                 decorator = resolve_qualified_symbol(name)
             else:
                 decorator = recover_kernel_decorator(name)
+                if decorator is None and name in self.parentVariables:
```
schweitzpgi, line 2538: Seems like we know more than we're letting on here.

If the symbol `name` is an actual argument, we have to resolve it at the call site and in the context of the call site. If it was a lifted argument (which can be determined without any new map being added), we need to lookup the symbol in the definition frame context (which is what I understand !4005 does).

python/cudaq/kernel/kernel_decorator.py
schweitzpgi: Yes. My question was in the theoretical sense here. Glad we all agree that there is no theoretical issue, just bugs in the gently meandering implementation ...

python/tests/kernel/test_assignments.py
schweitzpgi: I'm not a huge fan of reverting all the test changes. None of the existing tests _require_ delayed compilation, so why reintroduce that?

In fact, we now have deferred compilation as an option. It would behoove the project to have many kernels that defer compilation and many that do not defer compilation so that both code paths are adequately tested.

python/tests/kernel/test_kernel_features.py
schweitzpgi: I think printing the kernel is now compiling it.

I think `compile` is misleading. What this is really doing is building the IR, if it doesn't yet exist. Maybe we should call it `build_ir` or `instantiate_ir` for clarity?

"Compiling" is a multistep process and highly ambiguous here.

python/cudaq/kernel/ast_bridge.py
lmondada: #4005 is a PR on top of this one (see earlier comment https://github.com/NVIDIA/cuda-quantum/pull/3948#discussion_r2828695771), so it can only be merged in after this one. It entirely reverts this change, as well as solving the whole bug.

I have to leave this in temporarily, as this bug causes some kernel names don't resolve properly.

python/cudaq/kernel/ast_bridge.py
lmondada: same comment as before. As I've admitted, `parentVariables` is just a hack to avoid a pre-existing problem. #4005 fixes the root problem and should be merged in next.

lmondada:
> Where did we land wrt setting the default of deferred to False? Since it still defaults to True I can infer the answer, but I missed the explanation.

Here are the pros and cons of having deferred compilation on by default:
- pro: speeds up loading of scripts, libraries etc that define kernels but might not use them immediately
- pro: allows out-of-order definition, which is the correct Python semantics.
- pro: easier support for recursion (it works naturally when deferring compilation, given that the kernel reference exists by the time it is getting compiled)
- con: does not work with kernel builder. This is temporary and will be fixed by lambda lifting.

The only disadvantage is temporary and will go away. Being closer to Python semantics is a serious pro. 

> I'm not a huge fan of reverting all the test changes. None of the existing tests require delayed compilation, so why reintroduce that?

I've reverted all the test changes.

> I think printing the kernel is now compiling it.

Yes, I had made that change on Bettina's request. As a result of this, most tests could be reverted to their original. Thanks for spotting that, it's addressed in the latest commit.

> I think compile is misleading. What this is really doing is building the IR, if it doesn't yet exist. Maybe we should call it build_ir or instantiate_ir for clarity?

Previously, this was done in a function named `pre_compile`, but a `compile` function existed alongside it that was a no-op. I've consolidated both into one but agree that `compile` is ambiguous. I ask for permission to resolve this in a separate PR as getting rid of `compile` altogether will change a lot of unrelated code (see e.g. the uses in `python/tests/kernel/test_kernel_features.py`)

> We want to have tests that use deferred and not deferred code paths and validation that both cases work correctly

I have reverted more tests to `defer_compilation=False`, so there should be a good mix of both now. If you want to go beyond this, we could parametrise many tests in `python/tests/kernel/test_assignments.py` and elsewhere on whether they should defer compilation or not. We could then run all tests twice, once for each setting. However, this gets expensive quickly. Can we explore the options in a separate PR?

> Deferring the instantiation of the IR should be completely orthogonal to tracking contexts to resolving arguments.

Correct, this PR does not affect how contexts are resolved. Delaying compilation, however, surfaced more of the existing bugs around symbol resolution in the incorrect scope, as compilation was no longer happening in the scope of definition.

PR #3950, opened 2026-02-13
Fix macos devdeps cache invalidation on Github actions runner update

taalexander:
### Description
Github actions runner update upgraded Cmake. As this was not tracked in the cache key no rebuild was triggered. The devdeps were fetch and then when an incremental build of Python was attempted all of LLVM was built instead of just the MLIR python bindings. 

The fix here is to track the runner version in the cache key. As these update every week or two I added a cron job to potentially rebuild the devdeps (if an upgrade is triggered) in the morning before EU is online.


mitchdz (approved):
Let's make sure this passes CI first, but LGTM otherwise.

PR #3952, opened 2026-02-13
Hololink bridge v3

cketcham2333:
### Description

Updates libcudaq-realtime to work with HoloScan Sensor Bridge (HSB) (aka "Hololink").  Adds the bridge tool, which listens to inbound messages over HSB from an FPGA and transmits responses over HSB back to the FPGA.  Adds the FPGA stimulus tool to send data to the FPGA playback RAM over a control plane, initiate playback, and then retrieve data captured in the capture RAM.  Also adds an FPGA emulator.


.github/pre-commit/spelling_allowlist.txt
```diff
@@ -60,6 +64,7 @@ JSON
 Jupyter
 JupyterLab
 Kagome
+Ketcham
```
1tnguyen: Q: Could we figure out a way to bypass this word in the actual files rather than adding it to the `allowlist`?

bmhowe23: > Q: Could we figure out a way to bypass this word in the actual files rather than adding it to the `allowlist`?

+1

.github/pre-commit/spelling_allowlist.txt
```diff
@@ -76,6 +81,8 @@ Max-Cut
 Miniconda
 MyST
 NGC
+NIC
+NICs
```
bmhowe23: I think that if you add NIC that NICs is not required.

.github/pre-commit/spelling_allowlist.txt
```diff
@@ -195,7 +214,9 @@ coupler
 couplers
 cuQuantum
 cuTensor
+cudaHostAlloc
```
bmhowe23: If you escape function names with backticks, the spell checker allows them. This is the preferred way rather than adding code artifacts like function names to the list.

.github/pre-commit/spelling_allowlist.txt
```diff
@@ -244,14 +265,19 @@ extensibility
 extremal
 fermionic
 fermioniq
+fmt
```
bmhowe23: Same as above

realtime/include/cudaq/nvqlink/daemon/dispatcher/dispatch_kernel.cuh
```diff
@@ -35,6 +35,10 @@ namespace cudaq::nvqlink {
 inline void launch_dispatch_kernel_regular_inline(
     volatile std::uint64_t* rx_flags,
     volatile std::uint64_t* tx_flags,
+    std::uint8_t* rx_data,
+    std::uint8_t* tx_data,
+    std::size_t rx_stride_sz,
+    std::size_t tx_stride_sz,
```
bmhowe23, line 41: @caldwellshane note the breaking changes in this file.

.github/pre-commit/spelling_allowlist.txt
bmhowe23: Resolved in 6a8799fdcc

.github/pre-commit/spelling_allowlist.txt
bmhowe23: Resolved in 6a8799fdcc

.github/pre-commit/spelling_allowlist.txt
bmhowe23: Resolved in 6a8799fdcc

bmhowe23:
I resolved some of the spelling issues in 6a8799fdcc with the intent of not adding so many "not a word" words into the allow list. However, since the CI checks all the files that have changed since `main` rather than all the files that have changed since `features/cudaq.realtime`, there are still some failures. I think we should not block the merge for these spell check issues while this is still a feature branch.

.github/pre-commit/spelling_allowlist.txt
bmhowe23: Resolved in 6a8799fdcc

1tnguyen:
> I resolved some of the spelling issues in [6a8799f](https://github.com/NVIDIA/cuda-quantum/commit/6a8799fdccff1e61e13c3378e31674fe88f41ad7) with the intent of not adding so many "not a word" words into the allow list. However, since the CI checks all the files that have changed since `main` rather than all the files that have changed since `features/cudaq.realtime`, there are still some failures. I think we should not block the merge for these spell check issues while this is still a feature branch.

Agree. The two HTML files, where the spell check failures are from, are interim documents. We would probably convert them to RST/Markdown documents so that special words can be bypassed more effectively.

PR #3953, opened 2026-02-13
Re-enable run caching with new updates

atgeller:
### Description

Reenables the run caching in python, hoping that the changes in #3933 fixed the issue with Python garbage collecting being a little too aggressive and causing seg faults (see #3784).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -88,8 +88,8 @@ jobs:
       with:
         category: "/language:${{matrix.language}}"
 
-  # The CodeQL action status is not reported in merge queues. See https://github.com/github/codeql-action/issues/1537.
-  # This is a workaround to check the status of the CodeQL analysis in the PR but not in the merge queue. See https://github.com/orgs/community/discussions/46757#discussioncomment-7768838
+  # CodeQL status is not reported in merge queues. See https://github.com/github/codeql-action/issues/1537.
+  # Workaround: verify CodeQL passed via API keyed by pull_request.number so the PR gets a required status.
   check_codeql_status:
       name: Check CodeQL Status
       needs: analyze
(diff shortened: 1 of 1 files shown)


atgeller:
> /ok to test 7f644ea

atgeller:
> /ok to test 89520d3

atgeller:
> /ok to test 74423bb

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -119,13 +118,25 @@ static std::vector<py::object> pyRun(const std::string &shortName,
...
+
   auto span = pyRunTheKernel(shortName, platform, mod, retTy, shots_count,
                              qpu_id, opaques);
   auto results = pyReadResults(span, mod, shots_count, shortName);
 
   if (noise_model.has_value())
     platform.reset_noise();
 
+  if (auto *execCtx = cudaq::getExecutionContext();
+      execCtx && execCtx->jitEng) {
```
sacpis: Seems like both getExecutionContext() calls returns nullptr. Seems like we will need to fix it? @atgeller Would you please check this?

atgeller: @sacpis what test were you seeing this in?

schweitzpgi (approved):
Thanks, Adam. Let's get this merged so we can get a better analysis on Python performance.

python/runtime/cudaq/algorithms/py_run.cpp
sacpis: It is `python/tests/kernel/test_run_kernel.py`.

atgeller: Looks like the execution wasn't being set for `run` in python. I've added an execution context to hopefully address that.

atgeller:
> /ok to test 428996d

python/cudaq/runtime/run.py
```diff
@@ -61,9 +61,12 @@ def run(decorator, *args, shots_count=100, noise_model=None, qpu_id=0):
 
     specMod, processedArgs = decorator.handle_call_arguments(*args)
     retTy = decorator.get_none_type()
-    return cudaq_runtime.run_impl(decorator.uniqName + ".run", specMod, retTy,
-                                  shots_count, noise_model, qpu_id,
-                                  *processedArgs)
+    ctx = cudaq_runtime.ExecutionContext("run", shots_count)
+    ctx.kernelName = decorator.uniqName
+    with ctx:
```
lmondada: The execution context you are setting here will be ignored, because it gets overwritten later down the call stack:

`cudaq_runtime.run_impl` calls `pyRunTheKernel` calls `details::runTheKernel`.

This last one sets the execution context (see https://github.com/NVIDIA/cuda-quantum/blob/ce995fb3e8a786840a9a204475e233349c2b7cb9/runtime/cudaq/algorithms/run.cpp#L37), which will shadow the execution context you are trying to set here.

python/runtime/cudaq/algorithms/py_run.cpp
```diff
@@ -119,13 +118,25 @@ static std::vector<py::object> pyRun(const std::string &shortName,
   auto retTy = unwrap(returnTy);
   auto fnOp = getFuncOpAndCheckResult(mod, shortName);
   auto opaques = marshal_arguments_for_module_launch(mod, runtimeArgs, fnOp);
+
+  // Enable kernel caching over this launch.
+  if (auto *execCtx = cudaq::getExecutionContext())
```
lmondada: given that `python/cudaq/runtime/run.py` is the only place that calls this `run_impl` function, I would suggest just passing a new `allowJitEngineCaching` bool down the stack to runTheKernel, so that it can set it in its context accordingly.

Would that work, or do you need to cache the JIT outside of the runTheKernel execution?

python/runtime/cudaq/algorithms/py_run.cpp
lmondada: See my other comments. The execution context is set later on in `runTheKernel`. Setting a separate context outside of it is useless as it will be shadowed.

You must either pass `allowJitEngineCaching` as a boolean down the stack to set it in the context in `runTheKernel`, or you have to move all the context-setting logic out of runTheKernel and into here.

atgeller:
> /ok to test 830ca61

python/runtime/cudaq/algorithms/py_run.cpp
atgeller: I don't see a reason why we would need to cache it at a higher point in the stack. I took your approach of passing the bool down to `runTheKernel`, which seems to be working. Thanks!

atgeller:
> /ok to test c4f6288

atgeller:
Thanks @sacpis for the catch, and @lmondada for the help! I've verified locally that it is actually caching now.

PR #3954, opened 2026-02-13
Fixing iqm tests

Renaud-K:
Let me preface by saying that the issues with the test only show with the compiler refactor of https://github.com/NVIDIA/cuda-quantum/pull/3941. I might merge it with https://github.com/NVIDIA/cuda-quantum/pull/3941 and we can do some post-commit review.

With the compiler being its own object with a local lifetime, it can be calling the ServerHelper multiple times for its setup in `updatePassPipeline`. In this flow, we were asking to re-open or create the dqa file but error if the file already existed. This is what `O_CREAT | O_EXCL` does.  Removing `O_EXCL` allows to open the file an not error if it already existed.

The other issues where in test `dynamicQuantumArchitectureFile` which sets and unsets `IQM_SAVE_QPU_QA` and `IQM_QPU_QA`. The issue, here, is that both were only read once at start-up, way before there are actually set in the test. They therefore had no effect and no file was saved. The test does not check that a file is saved and was passing. With the refactor however, `IQM_QPU_QA` is read and found to not exist.

I don't think these environment variables were meant and designed to be used as global variables to be read from run to run. (I hope not :) ) and so I don't think we should be expecting consistent behavior using them that way. 

There is an additional twist whereby `IQM_QPU_QA` is read when the ServerHelper is inititalized and as it turns out we have a second initialization of the ServerHelper in `Future.cpp` when we process the `GET` message.
```
 sample_result future::get() {
...
  auto serverHelper = registry::get<ServerHelper>(qpuName);
  serverHelper->initialize(serverConfig);
```
So `IQM_QPU_QA` was read at the end of the second sample call but too late to have any effect. 
It is thanks to this second initialization that the `IQM_TOKEN` tests work but I feel that, this is probably a fluke.

I will focus initially on `dynamicQuantumArchitectureFile` and try set a precedent for creating tests that need to work sequentially under different environment settings, while still reusing the server invocation script `IQMStartServerAndTest.sh`. (We do not need more cut-and-paste versions of it)

So the test has been pulled out of `IQMTester.cpp` and moved into its own file `IQMDqaSaveTester.cpp`, where it is broken down into 3 tests. These tests are invoked with a script `TestDqaSaveLoad.sh` that will set the environment prior to launching tests one-by-one with `--gtest_filter`.

### DIFF
--- a/unittests/backends/iqm/CMakeLists.txt
+++ b/unittests/backends/iqm/CMakeLists.txt
@@ -17,6 +17,7 @@ add_backend_unittest_executable(test_iqm
   LINK_LIBS ${backend_unittest_libs} gmock_main
 )
 add_test(NAME iqm-tests COMMAND bash IQMStartServerAndTest.sh test_iqm WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+set_property(TEST iqm-tests PROPERTY SKIP_RETURN_CODE 77)
 
 ## IQM dynamic architecture file save/load
 add_backend_unittest_executable(test_iqm_save_load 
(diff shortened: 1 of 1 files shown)


Renaud-K:
> /ok to test 677ce5f3e102c1d27e4f94d0583a70deea3d7332

Renaud-K:
> /ok to test 606c437c5304dc2b62b292027e4bb19ed3fb6000

PR #3955, opened 2026-02-14
Fix wheel packaging macos

taalexander:
### Description
The macOS wheel was silently broken. It passed CI but failed when installed in a clean environment because the validation had gaps that masked packaging bugs. Identified this when trying to install the CI artifacts locally.

On the build side several CMake files used $ORIGIN for rpaths instead of macOS's @loader_path, so plugin dylibs couldn't find libraries. This allowed us removal of `--ignore-missing` from delocate so a broken wheel now fails the build instead of slipping through.

On the validation side we unset a bunch of variables that were unfortunately allowing our wheel to pass tests even thought it wasn't distributable so the wheel can't accidentally load libraries from the build tree. Also added running all tests from a temp directory so nothing in the repo tree can leak in via the working directory. Finally, I fixed a symlink-based test staging so all snippets/examples are actually discovered (turns out these were silently skipped). This now required skips for gpu and remote-backend tests. Going forward I'll move these to the macos publishing workflow but for now they are here (left a note in the code to follow up about).

Update - I've grabbed the artifacts and tested they are working locally on my machine and can run applications as expected with these fixes.

### DIFF
--- a/scripts/validate_pycudaq.sh
+++ b/scripts/validate_pycudaq.sh
@@ -88,13 +88,20 @@ OPTIND=$__optind__
 # Without this, DYLD_LIBRARY_PATH from a prior build step can cause the
 # wheel to load libraries from _skbuild/ instead of its own bundled copies,
 # masking packaging bugs.
-unset DYLD_LIBRARY_PATH
-unset DYLD_FALLBACK_LIBRARY_PATH
-unset LD_LIBRARY_PATH
-unset CUDAQ_INSTALL_PREFIX
-unset CUDA_QUANTUM_PATH
-unset PYTHONPATH
-echo "Environment sanitized (unset DYLD_LIBRARY_PATH, LD_LIBRARY_PATH, CUDAQ_INSTALL_PREFIX, CUDA_QUANTUM_PATH, PYTHONPATH)"
+SANITIZE_VARS="
+DYLD_LIBRARY_PATH
+DYLD_FALLBACK_LIBRARY_PATH
+LD_LIBRARY_PATH
+CUDAQ_INSTALL_PREFIX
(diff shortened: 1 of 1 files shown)


scripts/validate_pycudaq.sh
```diff
@@ -68,6 +83,19 @@ while getopts ":c:f:i:p:qv:" opt; do
...
+# Without this, DYLD_LIBRARY_PATH from a prior build step can cause the
+# wheel to load libraries from _skbuild/ instead of its own bundled copies,
+# masking packaging bugs.
+unset DYLD_LIBRARY_PATH
+unset DYLD_FALLBACK_LIBRARY_PATH
+unset LD_LIBRARY_PATH
+unset CUDAQ_INSTALL_PREFIX
+unset CUDA_QUANTUM_PATH
+unset PYTHONPATH
+echo "Environment sanitized (unset DYLD_LIBRARY_PATH, LD_LIBRARY_PATH, CUDAQ_INSTALL_PREFIX, CUDA_QUANTUM_PATH, PYTHONPATH)"
```
mitchdz: might be nice here to add all the paths in a list so the `echo` never gets out of date:
```suggestion
SANITIZE_VARS="
DYLD_LIBRARY_PATH
DYLD_FALLBACK_LIBRARY_PATH
LD_LIBRARY_PATH
CUDAQ_INSTALL_PREFIX
CUDA_QUANTUM_PATH
PYTHONPATH
"

for var in $SANITIZE_VARS; do
  unset "$var"
done

echo "Environment sanitized (unset: $SANITIZE_VARS)"
```

taalexander: Addressed in 888c6dce

PR #3960, opened 2026-02-16
Enable ccache for CI builds with GHCR caching

taalexander:
### Description
CI builds currently recompile everything from scratch on every run, even when most source files haven't changed. This PR adds ccache across all CI workflows so that unchanged translation units are served from cache, which should cut build times substantially for most development work.

Cache data is stored in GHCR using ORAS. When a build starts, it looks for a cache in priority order: first the current PR's previous cache, then the PR's target branch, then main. After building, the updated cache is pushed back if the miss rate exceeded 10%. When a PR is closed, its cache artifacts are cleaned up automatically.

The ccache logic is implemented as three reusable composite actions (ccache-setup, ccache-restore, ccache-push) shared across all workflows. For macOS jobs, ccache runs on the host. For Linux jobs, cache data is injected into Docker builds via a BuildKit build-context and extracted afterward for persistence. This should seamlessly fallback to a local cache on the local devcontainer (which conveniently should also implement ccache support where it previously wasn't being used).

ci.yml/macos_ci.yml times (need to run):

Before:
- Linux: 1h 18m 39s
- MacOS: 1h 4m 8s

After (hot-cache):
- Linux:
- MacOS: 

Publishing: (Need to run)

Deployments: (Need to run)

The build cudaq/wheel stages are down to 4-5 mins instead of ~15-20 they were before. Now the big blocker is test times (I've been thinking about some inefficiencies in parallel tests in #4111).


taalexander:
I have migrated getVersion/getFullRepositoryVersion to be injected via cpp to avoid cache miss cascades on each commit. 

With this we have high proportion of cache hits:

```
Prepare all required actions
Getting action download info
Run ./.github/actions/ccache-push
Run cache_path="/Users/runner/.ccache"
=== ccache stats ===
Cacheable calls:    452 / 452 (100.0%)
  Hits:             448 / 452 (99.12%)
    Direct:         447 / 448 (99.78%)
    Preprocessed:     1 / 448 ( 0.22%)
  Misses:             4 / 452 ( 0.88%)
Local storage:
  Cache size (GiB): 0.3 / 5.0 ( 5.83%)
  Hits:             448 / 452 (99.12%)
  Misses:             4 / 452 ( 0.88%)
ccache miss rate: 0% (4 misses / 452 total)
Miss rate below 10%; skipping push.
```

and great build time improvements.

taalexander:
> /ok to test 4bd692b

taalexander:
> /ok to test 4bb5633

PR #3962, opened 2026-02-16
[orca] Execution context and lifecycle simplification

khalatepradnya:
* No need to override execution context methods used by local simulators since `orca` is a remote target.

* Follow-up to the changes in PR https://github.com/NVIDIA/cuda-quantum/pull/3736 which caused the nightly integration failures seen here: https://github.com/NVIDIA/cuda-quantum/actions/runs/22029239324/job/63651273975

```
 Sampling detected on a kernel with no qubits. Your kernel must have qubits to sample it.
```

* Manually verified that the `orca.cpp` and `orca.py` tests are working.


sacpis (approved):
LGTM. Thanks @khalatepradnya.

nit: commit message needs to be fixed to `Do not need` :)

lmondada:
Thank you 🙏

PR #3964, opened 2026-02-16
[python] Discriminate broadcast vs list kernel arguments

### LINKED ISSUE #2895: [python] cudaq.sample API assumes broadcast on nested list arguments

lmondada:
The logic deciding whether a kernel call `kernel(*args)` is in fact a broadcast (i.e. a batched kernel call with different arguments given by the elements of the lists passed as arguments) is currently very simple:
- if the first argument is a list AND
- if the kernel does NOT expect a non-nested list as first argument,

then the call is taken to be a broadcast: each list argument is unpacked, and passed one-by-one to the kernel.

This wrongly detects some cases. For instance, the following is wrongly detected as a broadcast:

```python
@cudaq.kernel
def kernel(l: list[list[int]]):
    pass

cudaq.sample(kernel, [[0, 1]])
```

whereas on the other hand this is wrongly identified as not-a-broadcast:

```python
@cudaq.kernel
def kernel(l: list[int]):
    pass

cudaq.sample(kernel, [[0, 1]])
```

As a result, both fail with a "Invalid runtime argument type" error.

---

This PR suggests a simple criterion to check whether a call is a broadcast: it computes the depth of nested lists in the passed arguments $n_{args}$ and compares it to the depth of nested lists in the expected type $n_{exp}$. If $n_{args} == n_{exp} + 1$, then the invocation is determined to be a broadcast, otherwise it is not.

This should be much more reliable and I cannot think of a case that would be identified wrongly. Do you see disadvantages to this proposal?

closes #2895

### DIFF
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -800,20 +800,37 @@ def test_list_list_string_argument_error():
 
 def test_broadcast():
 
+    @cudaq.kernel
+    def kernel(l: list[list[int]]):
+        q = cudaq.qvector(2)
+        for inner in l:
+            for i in inner:
+                x(q[i])
+
+    # passing a list of lists works
+    counts = cudaq.sample(kernel, [[0, 1]])
+    # passing a 3-nested list works (broadcasting)
+    counts = cudaq.sample(kernel, [[[0, 1]], [[0, 1]]])
+    assert len(counts) == 2
+    # passing a singly-nested list fails
     with pytest.raises(RuntimeError) as e:
-
-        @cudaq.kernel
-        def kernel(l: list[list[int]]):
-            q = cudaq.qvector(2)
-            for inner in l:
-                for i in inner:
-                    x(q[i])
-
-        #FIXME: update broadcast detection logic to allow this case.
-        # https://github.com/NVIDIA/cuda-quantum/issues/2895
-        counts = cudaq.sample(kernel, [[0, 1]])
+        counts = cudaq.sample(kernel, [0, 1])
     assert 'Invalid runtime argument type. Argument of type list[int] was provided' in repr(
         e)
+
+
+def test_broadcast_np_array():
+
+    @cudaq.kernel
+    def kernel(l: list[float]):
+        q = cudaq.qvector(2)
+
+    # passing a mix of list and np.array broadcasts properly
+    counts = cudaq.sample(kernel, [np.array([0, 1]), np.array([0, 1])])
+    assert len(counts) == 2
+    # passing a 2d np.array broadcasts properly
+    counts = cudaq.sample(kernel, np.array([[0, 1], [0, 1]]))
+    assert len(counts) == 2
 
 
 def test_list_creation_with_cast():
--- a/python/cudaq/runtime/utils.py
+++ b/python/cudaq/runtime/utils.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 from cudaq.kernel.kernel_builder import PyKernel
 from cudaq.kernel.kernel_decorator import isa_kernel_decorator
 from cudaq.kernel.utils import mlirTypeToPyType
+from cudaq.mlir.ir import Type as MlirType
 from cudaq.mlir._mlir_libs._quakeDialects import cudaq_runtime
 from cudaq.mlir.dialects import cc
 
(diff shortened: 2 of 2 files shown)


sacpis (approved):
LGTM. Thanks @lmondada.

schweitzpgi:
Should we restrict broadcast to device-only kernel calls where a qubit target argument is "magically expanded" by passing a qvector actual?

```python
  qs = cudaq.qvector(5)
  # expand the next line to 5 independent calls
  my_device_only_kernel(qs)
```

That would be consistent with what broadcast means as far as CUDA-Q intrinsics.

Do we have cases where we're expecting a list of integers to expand into multiple calls in Python? (That sounds suspect.)

python/tests/kernel/test_kernel_features.py
```diff
@@ -800,22 +800,39 @@ def kernel(n: int, s: list[list[str]]):
...
+        q = cudaq.qvector(2)
+        for inner in l:
+            for i in inner:
+                x(q[i])
+
+    # passing a list of lists works
+    counts = cudaq.sample(kernel, [[0, 1]])
+    # passing a 3-nested list works (broadcasting)
+    counts = cudaq.sample(kernel, [[[0, 1]], [[0, 1]]])
+    assert len(counts) == 2
```
khalatepradnya, line 814: Am I correct to understand that in the first case you get single counts dictionary, and in the second case you get a list of counts, corresponding to each argument in the list?
Can you expand this test to clarify that?

khalatepradnya (approved):
The proposed fix LGTM. Thanks, @lmondada !
I think the docs should be updated to reflect this interpretation of broadcast (can be a follow-up task), and it would be good to get confirmation from @boschmitt as well.

schweitzpgi:
More specifically, this won't compile in C++ and I am at a bit of a loss why it should work for Python.
```python
    # passing a 3-nested list works (broadcasting)
    counts = cudaq.sample(kernel, [[[0, 1]], [[0, 1]]])
```

In C++, if we have
```c++
__qpu__ void kernel(std::vector<std::vector<int>> arg) ...
...
}
...
  std::vector<std::vector<std::vector<int>>> type_error_thing = ...;
  cudaq::sample(kernel, type_error_thing);  // C++ error: "no dice, bucko!"
```
@amccaskey

lmondada:
Yes, the result of this PR would be a strictly more permissive policy around broadcasting arguments. That is required if we want the following to be valid

```python
@cudaq.kernel
def kernel(l: list[int]):
    pass

cudaq.sample(kernel, [[0, 1], [1, 2]])
```

That the above fails I find inconsistent with the expected behaviour of broadcasting -- that's my subjective opinion. I could not find any documentation (public or otherwise) regarding broadcasting, please point me to it if I have missed it.

1. Do you think the above example merits getting supported or not?
2. I can adjust this PR to reject whatever cases we do not wish to support (such as unsupported nested arguments).
3. I can make the same changes to C++ to maintain consistency.
4. Crucially, this proposal (and any restriction of it) does NOT alter the behavior of anything currently valid. 

@schweitzpgi I did not expect this to be controversial (just tried to solve a failing test). If you consider this a useful contribution, I can iterate on it as you see fit. Otherwise I have no problems shelving it.

khalatepradnya:
> Yes, the result of this PR would be a strictly more permissive policy around broadcasting arguments. That is required if we want the following to be valid
> 
> ```python
> @cudaq.kernel
> def kernel(l: list[int]):
>     pass
> 
> cudaq.sample(kernel, [[0, 1], [1, 2]])
> ```
> 
> That the above fails I find inconsistent with the expected behaviour of broadcasting -- that's my subjective opinion. I could not find any documentation (public or otherwise) regarding broadcasting, please point me to it if I have missed it.
> 
> 1. Do you think the above example merits getting supported or not?
> 2. I can adjust this PR to reject whatever cases we do not wish to support (such as unsupported nested arguments).
> 3. I can make the same changes to C++ to maintain consistency.
> 4. Crucially, this proposal (and any restriction of it) does NOT alter the behavior of anything currently valid.
> 
> @schweitzpgi I did not expect this to be controversial (just tried to solve a failing test). If you consider this a useful contribution, I can iterate on it as you see fit. Otherwise I have no problems shelving it.

I think we need to fix the bug described here - https://github.com/NVIDIA/cuda-quantum/issues/2895. It's a valid kernel with valid arguments that are incorrectly interpreted.
We can disallow automatic broadcast on kernels (and continue allowing on quantum operations), and follow-up with a spec proposal for broadcast separately.

lmondada:
There's currently too much going on for the upcoming release and this is not a priority. Might revisit later.

Issue #3973, opened 2026-02-17
[CI] - Remote-Sim/observe_async.cpp test failure bind: Address already in use


mitchdz:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

https://github.com/NVIDIA/cuda-quantum/actions/runs/22005332183/job/63588601783?pr=3804#step:12:26327

```
#77 29.26 FAIL: CUDAQ-Target :: Remote-Sim/observe_async.cpp (54 of 327)
#77 29.26 ******************** TEST 'CUDAQ-Target :: Remote-Sim/observe_async.cpp' FAILED ********************
#77 29.26 Exit Code: 134
#77 29.26 
#77 29.26 Command Output (stderr):
#77 29.26 --
#77 29.26 RUN: at line 12: nvq++ --target remote-mqpu --remote-mqpu-auto-launch 3 /cuda-quantum/targettests/Remote-Sim/observe_async.cpp -o /cuda-quantum/build/targettests/Remote-Sim/Output/observe_async.cpp.tmp && /cuda-quantum/build/targettests/Remote-Sim/Output/observe_async.cpp.tmp
#77 29.26 + nvq++ --target remote-mqpu --remote-mqpu-auto-launch 3 /cuda-quantum/targettests/Remote-Sim/observe_async.cpp -o /cuda-quantum/build/targettests/Remote-Sim/Output/observe_async.cpp.tmp
#77 29.26 + /cuda-quantum/build/targettests/Remote-Sim/Output/observe_async.cpp.tmp
#77 29.26 libc++abi: terminating due to uncaught exception of type std::__1::system_error: bind: Address already in use
#77 29.26 libc++abi: terminating
[...]

PR #3974, opened 2026-02-17
Fix "Address already in use" in remote-mqpu auto-launch (observe_async test)

mitchdz:
Fixes https://github.com/NVIDIA/cuda-quantum/issues/3973

Tests that use `--remote-mqpu-auto-launch` (e.g. `Remote-Sim/observe_async.cpp`) can fail with:
```
libc++abi: terminating due to uncaught exception of type std::__1::system_error: bind: Address already in use
```
The runtime picks a port by binding in `portAvailable()` and then starts cudaq-qpud on that same port. After the probe socket is closed, the port can stay in TIME_WAIT, so when the server binds it gets "Address already in use."

Also, set `servAddr.sin_port` from host byte order to network byte order.


1tnguyen (approved):
LGTM 👍

PR #3975, opened 2026-02-17
[Testing] build cuda-quantum-src image

mitchdz:
# Add Docker image with third-party library (tpls) source

Adds a `cuda-quantum-src` image that extends the CUDA-Q base image with the source code of the open-source libraries used by CUDA-Q (tpls). The image is built in CI during deployments and published to the same registry with matching tags.


mitchdz:
> /ok to test ad0039c

mitchdz:
> /ok to test e8f8795

PR #3978, opened 2026-02-17
[CI] enable fail-fast for merge-group runs

mitchdz:
## Summary
Enable `fail-fast: true` for matrix jobs when the workflow is triggered by a merge group (merge queue). For normal PR/push runs, matrix jobs keep `fail-fast: false` so the full matrix still runs for debugging.

## Motivation
When a merge-group run has a failing job, we want the run to fail quickly so the merge group is marked failed and the queue can move on to the next group. With fail-fast enabled only for merge-group:
- The first failing matrix combination cancels the rest of that job’s matrix (saves time and runners).
- The workflow run fails as soon as that job fails, so the merge group check fails and the next groups can start sooner.
- Regular PR and push runs are unchanged and still run the full matrix for debugging.

## Changes
- Set `fail-fast: ${{ github.event_name == 'merge_group' }}` on all matrix jobs: `devdeps`, `wheeldeps`, `source_build`, `build_and_test`, `gen_code_coverage`, `docker_image`, `python_wheels`, and `binaries`.


taalexander (approved):
lgtm, makes sense.

1tnguyen (approved):
👍

1tnguyen (approved):

PR #3982, opened 2026-02-17
[fermioniq] [python] Override the `launchModule` method 

khalatepradnya:
* This PR refactors the kernel launching logic in `FermioniqBaseQPU` to improve code reuse and handling of execution contexts for `observe` operations.

* Fixes the nightly integration test failure seen here: https://github.com/NVIDIA/cuda-quantum/actions/runs/22085270871/job/63818586242

* Manually verified by running the test locally:
```
# python3 ../docs/sphinx/targets/python/fermioniq_observables.py
Energy is -1.7487948611472164
```

* Builds on top of changes in PR https://github.com/NVIDIA/cuda-quantum/pull/3986

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -88,8 +88,8 @@ jobs:
       with:
         category: "/language:${{matrix.language}}"
 
-  # The CodeQL action status is not reported in merge queues. See https://github.com/github/codeql-action/issues/1537.
-  # This is a workaround to check the status of the CodeQL analysis in the PR but not in the merge queue. See https://github.com/orgs/community/discussions/46757#discussioncomment-7768838
+  # CodeQL status is not reported in merge queues. See https://github.com/github/codeql-action/issues/1537.
+  # Workaround: verify CodeQL passed via API keyed by pull_request.number so the PR gets a required status.
   check_codeql_status:
       name: Check CodeQL Status
       needs: analyze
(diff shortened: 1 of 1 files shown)


runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
```diff
@@ -61,58 +87,44 @@ class FermioniqBaseQPU : public BaseRemoteRESTQPU {
...
+      cudaq::detail::setExecutionContext(&defaultContext);
+    }
 
     Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
                       noiseModel, emulate);
-    auto codes = rawArgs.empty()
-                     ? compiler.lowerQuakeCode(kernelName, args, {})
-                     : compiler.lowerQuakeCode(kernelName, nullptr, rawArgs);
-    if (codes.size() != 1) {
+    auto codes = lower(compiler);
```
Renaud-K: If what we need here is the compilation to happen with the defaultContext, then I would much rather have it passed to the compiler as an argument. Then the compiler does not rely on the global state. It only calls `getExecutionContext` in one place, so it would be easy to change to an argument. The only reason why it still calls `getExecutionContext()` is because the code was freshly moved and there is less merge conflict by moving the code verbatim.  It can change now.

khalatepradnya: Thanks for the suggestion. I agree that passing the execution context explicitly to the `Compiler` would be cleaner than manipulating global state.
However, that change would require me to refactor `lowerQuakeCodePart2` and update all callers, right? More than this one place... Would you be okay with keeping the current approach for this PR and addressing the Compiler API change as a follow-up?

Renaud-K: I would not be okay with it. It is everything we are trying to walk away from. Here is the PR that does it: https://github.com/NVIDIA/cuda-quantum/pull/3986

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
```diff
@@ -61,58 +87,44 @@ class FermioniqBaseQPU : public BaseRemoteRESTQPU {
...
-    auto *originalContext = executionContext;
-    if (executionContext->name == "observe")
-      executionContext = &defaultContext;
+    cudaq::ExecutionContext *originalContext = nullptr;
+    if (executionContext->name == "observe") {
+      originalContext = executionContext;
+      cudaq::detail::setExecutionContext(&defaultContext);
+    }
 
     Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
```
Renaud-K, line 95: Did the compiler refactor introduce this issue? If so, could you add a test for it?

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
khalatepradnya: Okay, I can build on top of this PR.

runtime/cudaq/platform/fermioniq/FermioniqBaseQPU.h
khalatepradnya: There exists a test that caught this - https://github.com/NVIDIA/cuda-quantum/blob/main/docs/sphinx/targets/python/fermioniq_observables.py.
However, it does not run in per-PR CI. See PR description for the link to CI failure.
Unfortunately, the test infrastructure doesn't cover this target properly (yet), there isn't mock server or emulation mode. I can open up a follow-up task to address this gap since I think it is a large work item.

Renaud-K: Sounds reasonable. Thank you.

khalatepradnya:
> /ok to test 8870003

khalatepradnya:
> /ok to test 3cd0aba

PR #3986, opened 2026-02-17
Passing the execution context as an argument to the compiler.

Renaud-K:
To avoid having to modify and restore global state when the compiler needs a different execution context, these changes pass it as an argument.

### DIFF
--- a/runtime/common/Compiler.cpp
+++ b/runtime/common/Compiler.cpp
@@ -224,11 +224,9 @@ Compiler::Compiler(ServerHelper *serverHelper,
 Compiler::~Compiler() = default;
 
 std::vector<cudaq::KernelExecution> Compiler::lowerQuakeCodePart2(
-    const std::string &kernelName, void *kernelArgs,
-    const std::vector<void *> &rawArgs, mlir::ModuleOp m_module,
-    mlir::MLIRContext *contextPtr, void *updatedArgs) {
-  auto executionContext = cudaq::getExecutionContext();
-
+    ExecutionContext *executionContext, const std::string &kernelName,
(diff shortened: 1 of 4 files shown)


Renaud-K:
> /ok to test 1c9f8bfc7f6f2e9a119c5930586e26fa673a3554

runtime/common/Compiler.cpp
```diff
@@ -224,11 +224,9 @@ Compiler::Compiler(ServerHelper *serverHelper,
...
 
 std::vector<cudaq::KernelExecution> Compiler::lowerQuakeCodePart2(
-    const std::string &kernelName, void *kernelArgs,
-    const std::vector<void *> &rawArgs, mlir::ModuleOp m_module,
-    mlir::MLIRContext *contextPtr, void *updatedArgs) {
-  auto executionContext = cudaq::getExecutionContext();
-
+    ExecutionContext *executionContext, const std::string &kernelName,
+    void *kernelArgs, const std::vector<void *> &rawArgs,
+    mlir::ModuleOp m_module, mlir::MLIRContext *contextPtr, void *updatedArgs) {
```
schweitzpgi, line 229: I know this is pre-existing, but...

nit: why the weird `m_` prefix here, I wonder?

schweitzpgi:
Looks great, Renaud.

runtime/common/Compiler.cpp
Renaud-K: I noticed them in other places as well. Maybe it started as `m_` for member and then took on a new life. `m_` for mlir? I can clean them up when I next have the opportunity.

PR #3987, opened 2026-02-17
Introduce new way of getting `qkernels` from python kernels in C++

atgeller:
### Description

Adds a clear distinction between specializing a python module to a pure device kernel (via a new `CppPyKernelDecorator::getPureDeviceKernel` function), which is meant to be called from other kernels, and an entry point (via a pre-existing `CppPyKernelDecorator::getEntryPointFunction` function), which is meant to be called from host code.


runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -265,6 +272,9 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
     if (!funcOp)
       throw std::runtime_error("no kernel named " + name + " found in module");
 
+    auto isPureDevice = funcOp->hasAttr("cudaq_puredevice");
+    llvm::outs() << "Is pure device? " << isPureDevice << "\n";
```
schweitzpgi: Looks like debug code.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -280,15 +290,17 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
     CUDAQ_INFO("Run Argument Synth.\n");
     if (enablePythonCodegenDump)
       module.dump();
-    specializeKernel(name, module, rawArgs, resultTy, enablePythonCodegenDump);
+    specializeKernel(name, module, rawArgs, resultTy, enablePythonCodegenDump,
+                     !isPureDevice);
```
schweitzpgi: Should this follow the argument `isEntryPoint` instead of an attribute stuck in the IR?

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Whoops, yes, that was not supposed to be included.

python/runtime/interop/PythonCppInterop.h
```diff
@@ -31,24 +31,45 @@ class CppPyKernelDecorator {
 
   ~CppPyKernelDecorator();
 
+  /// Fully compiles this python kernel, returning a `qkernel` that can
+  /// be directly invoked by host code. Do not pass the returned `qkernel`
+  /// into other kernels, as this will lead to bad things.
   template <typename T, typename... As>
     requires QKernelType<T>
   T getEntryPointFunction(As... as) {
+    auto p = getKernelHelper(true, as...);
```
schweitzpgi:
```suggestion
    auto p = getKernelHelper(/*isEntryPoint=*/true, as...);
```

python/runtime/interop/PythonCppInterop.h
```diff
@@ -31,24 +31,45 @@ class CppPyKernelDecorator {
...
+  }
+
+  /// Fully compiles this python kernel, returning a `qkernel` that can
+  /// be indirectly invoked by other kernels. Only pass the returned
+  /// `qkernel` into other kernels, calling it directly from host code
+  /// will lead to bad things.
+  template <typename T, typename... As>
+    requires QKernelType<T>
+  T getDirectKernelCall(As... as) {
+    auto p = getKernelHelper(false, as...);
```
schweitzpgi:
```suggestion
    auto p = getKernelHelper(/*isEntryPoint=*/false, as...);
```

atgeller:
> /ok to test 6721e60

PR #3988, opened 2026-02-17
[python] [testing] Restore the logic for `start_server` in individual mock QPUs

khalatepradnya:
* Follow-up to the changes in PR https://github.com/NVIDIA/cuda-quantum/pull/3747.

* The test mock servers are used in both C++ and Python testing workflows, but they are invoked differently. This PR restores the utility function to each mock QPU backend to start its server, so that Python import path works.

* Also, make the CI check strict so that these tests are no longer silently skipped.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -514,7 +514,7 @@ class PyKernelDecorator(object):
             self.uniqName, specialized_module, mlirTy, *processedArgs)
         return result
 
-    def beta_reduction(self, *args):
+    def beta_reduction(self, isEntryPoint, *args):
         """
         Perform beta reduction on this kernel decorator in the current calling
         context. We are primary concerned with resolving the lambda lifted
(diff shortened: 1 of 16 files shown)


utils/mock_qpu/oqc/__init__.py
```diff
@@ -191,3 +191,8 @@ async def qetQpu(authentication_token: str = Header(...)):
     }
 
     return JSONResponse(content=data)
+
+
+def startServer(port):
```
amccaskey, line 196: just an alternative - instead of adding startServer to all the mock qpus, you can get the uvicorn app directly like this https://github.com/amccaskey/cuda-quantum/blob/llvm22-migration-v2/python/tests/backends/test_Quantinuum_kernel.py#L16 (see lines 16-23 there).

khalatepradnya: Thanks for sharing alternative idea! I see that it keeps the mock package cleaner and moves everything the test needs inside the file. However, I like keeping the test infrastructure separate so the individual test can simply call them, no? Perhaps I'm missing something?

mitchdz (approved):

PR #3989, opened 2026-02-17
Reset simulator on trap

atgeller:
### Description

I noticed an issue in Python where, if a kernel traps, it causes a runtime error which can be nicely caught, handled, and recovered from. However, the simulator state is not cleared, so any subsequent use of the simulator results in incorrect results. I was unable to reproduce this behavior in C++.

This PR addresses this issue by explicitly resetting the simulator on a trap, and adds a test that simulation after a trap no longer leads to incorrect results.

### DIFF
--- a/docs/sphinx/examples/cpp/sample_to_run_migration.cpp
+++ b/docs/sphinx/examples/cpp/sample_to_run_migration.cpp
@@ -1,0 +1,139 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include <algorithm>
+#include <cassert>
+#include <cstdio>
+#include <cudaq.h>
+#include <map>
+#include <string>
+
+// [Begin Sample_Works]
+__qpu__ void bell() {
+  cudaq::qvector q(2);
+  h(q[0]);
+  x<cudaq::ctrl>(q[0], q[1]);
+}
+
+__qpu__ void reset_pattern() {
+  cudaq::qubit q;
+  h(q);
+  mz(q);
+  reset(q);
+  x(q);
+}
+// [End Sample_Works]
+
+// [Begin Example1]
+struct simple_conditional {
+  auto operator()() __qpu__ {
+    cudaq::qvector q(2);
+    h(q[0]);
+    auto r = mz(q[0]);
+    if (r) {
+      x(q[1]);
+    }
+    return mz(q[1]);
+  }
+};
+// [End Example1]
+
+// [Begin Example2]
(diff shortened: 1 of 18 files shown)


runtime/nvqir/NVQIR.cpp
```diff
@@ -697,6 +697,7 @@ static std::vector<std::size_t> safeArrayToVectorSizeT(Array *arr) {
 // kernel, which calls this function. The trap should explain the issue to the
 // user and about the kernel when executed.
 void __quantum__qis__trap(std::int64_t code) {
+  nvqir::getCircuitSimulatorInternal()->endExecution();
```
sacpis: It looks good, just one thing here. Can we check if this cleans up execution manager's internal qudit tracking?

atgeller: I'll double check tomorrow, but I'm pretty sure it does because the bug I was seeing was that previously allocated qubits were appearing, which is no longer the case with this addition.

atgeller: Ah the execution manager, no it looks like it doesn't. I'll add an explicit call there.

atgeller:
> /ok to test e649a05

runtime/nvqir/NVQIR.cpp
```diff
@@ -697,6 +697,11 @@ static std::vector<std::size_t> safeArrayToVectorSizeT(Array *arr) {
...
+  if (auto *ctx = cudaq::getExecutionContext())
+    if (auto *manager = ctx->executionManager)
+      manager->endExecution();
```
sacpis: Seems like we are calling `simulator::endExecution()` in `endExecution()` which ultimately calls [CircuitSimlatorBase::endExecution()](https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp#L174). It it being called twice now? Would freeing the device memory twice cause a crash or will it be a no-op?

Wondering if we can skip one endExecution, using something like

```
if (auto *ctx = cudaq::getExecutionContext())
    if (auto *manager = ctx->executionManager) {
      manager->endExecution();
      goto finish;
    }
    nvqir::getCircuitSimulatorInternal()->endExecution();
    finish:
```

sacpis: Would you please check if we should call `resetExecutionContext`? Seems like `endExecution` never clears the execution context. Or may be I am missing something on my side.

sacpis (commented):
Left a few comments. Found something else related to qubits and trap. Will get back here in some time.

runtime/nvqir/NVQIR.cpp
sacpis: So `endExecution()` returns qubit indices from `contextQuditIdsForDeletion`. That list is filled up when qubits are freed (using destructor). Will there be an issue when we have mid-execution trap and some live qubits haven't been freed yet, so their indices are not returned? I haven't tested it but would the next kernel allocate qubits with wrong indices.

sacpis (approved):
LGTM. Thanks @atgeller.

PR #3990, opened 2026-02-17
[PTSBE] Apply noise style api

sacpis:
PTSBE now supports inline cudaq.apply_noise() in kernels in addition to gate-based noise from a NoiseModel. The noise_model argument to cudaq.ptsbe.sample() is optional. Both styles can be used together.

Usage
- In kernel, use `cudaq.apply_noise(cudaq.DepolarizationChannel, 0.05, q[0])`
- With model, use noise.add_channel("h", [0], cudaq.DepolarizationChannel(0.05)) then cudaq.ptsbe.sample(kernel, noise_model=noise, ...)
- Both can be used in the same run.

### DIFF
--- a/python/tests/ptsbe/test_errors.py
+++ b/python/tests/ptsbe/test_errors.py
@@ -80,4 +80,3 @@ def test_ptsbe_sample_non_integer_shots_raises(depol_noise, bell_kernel):
 def test_mcm_kernel_rejected(depol_noise, mcm_kernel_fixture):
     with pytest.raises(RuntimeError, match="conditional feedback|measurement"):
         cudaq.ptsbe.sample(mcm_kernel_fixture, noise_model=depol_noise)
-


1tnguyen (approved):
LGTM 👍

runtime/common/Trace.h
```diff
@@ -9,35 +9,53 @@
...
+#include <optional>
 #include <string>
 #include <vector>
 
 namespace cudaq {
 
 struct QuditInfo;
 
+/// @brief Name used in the trace for apply_noise (inline noise) instructions.
+constexpr std::string_view TRACE_APPLY_NOISE_NAME = "apply_noise";
```
taalexander, line 29: Not blocking but I think it would be best if we are introducing a new type to the instruction trace we should add a type enum (can default to GATE for backwards compatiability) and corresponding field for the Trace object to allow clean differentiation betwee gate/noise. See how I did this in the PTSBE trace implementation.

taalexander (approved):
One comment in-line that is non-blocking and could be cleaned up going forward. I also want to understand what happens to the other users of Trace now that noise instructions may be present? Will they function or gracefully throw an error?

sacpis:
> One comment in-line that is non-blocking and could be cleaned up going forward. I also want to understand what happens to the other users of Trace now that noise instructions may be present? Will they function or gracefully throw an error?

All current Trace users either skip or explicitly handle Noise. I don't see any code path which assumes every instruction is a gate (let me know if you see any, I may have overlooked). If any user in the future treats every instruction as a gate, they would get an explicit error for a Noise instruction. So failures are graceful rather than silent.

runtime/common/Trace.h
sacpis: Good one. Addressed in d5b4b8a.

PR #3993, opened 2026-02-18
Make CUDA-Q installer non-root installable

### LINKED ISSUE #1075: Support installing CUDA Quantum in a local directory
CUDA Quantum currently finds some of its dependencies (cuQuantum, cuTensor) via the rpaths configured during build.
These paths are absolute, making the installation non-relocatable. As a consequence, installing pre-built binaries requires admin privileges to install the dependencies in global locations. 

It would be nice if it was possible to install CUDA Quantum in a user directory without requiring building CUDA Quantum from source. 
Specifically, the cuda_quantum_installer should take the installation location as argument. It would also be nice if the installer checked 
the system pre-requisites before proceeding with the installation.

taalexander:
### Description
Closes #1075. I ran into a series of issues with the macOS installer when attempting to locally install the Github artifact. Ultimately the core issue was that the dependencies were being installed in a variety of locations many of which were not macOS friendly. Additionally, the installer was picking up the Cmake installer paths directly which was fine for the docker environment where the package was built in root but not in the case of macOS where the `runner` user is used to build cudaq and build the wheel.

The fix I propose after discussing with @mitchdz is to consolidate all cudaq dependencies in the installer into the `cudaq` directory. Then make this directory optionally installable to a non-root locations through `--installpath <path>` argument. By default it will continue to be installed to the existing install location. Documentation is updated to reflect this. The only (small) downside to this approach is that on the offchance users are linking or using the cuquantum/llvm packages shipped with cudaq (which is not recommended) they will have to update their paths.

I also fixed a bug in set_env that was causing it to pollute environments by appending to the path multiple times (everytime the shell was sourced).

Change list:
- Merged installer layout: LLVM tools, clang resource dir, and CUDA libs are staged directly into the CUDA-Q prefix (bin/, lib/, include/) rather than separate subdirectories. nvq++ detects the merged layout at runtime and falls back to the build-tree LLVM path when running from a dev build.
- `--installpath` support: The installer now accepts `--installpath <dir>` for non-root installs to custom locations. Shell profiles (.bashrc, .zshrc) are updated automatically.
Idempotent set_env.sh: Sourcing the environment script multiple times no longer duplicates
- `PATH/DYLD_LIBRARY_PATH` entries. Uses a portable printenv/export helper that works in both bash and zsh.
CI validation: macOS and Linux workflows updated to validate custom install paths instead of only testing the default /opt/nvidia/cudaq.

Checks to verify:
- [x] Passes CI
- [x] Passes deployments
- [x] Passes publishing

### DIFF
--- a/python/runtime/common/py_ExecutionContext.cpp
+++ b/python/runtime/common/py_ExecutionContext.cpp
@@ -77,7 +77,8 @@ void bindExecutionContext(py::module &mod) {
       .def("__exit__", [](cudaq::ExecutionContext &ctx, py::object type,
                           py::object value, py::object traceback) {
         if (type.is_none()) {
-          // No exception, so we finalize the context and reset it
+          // Normal exit: finalize results, clean up the simulator,
+          // and reset the context (guaranteed even if finalize throws).
           auto &platform = cudaq::get_platform();
           detail::try_finally(
               [&] {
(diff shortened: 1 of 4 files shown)


taalexander:
Passed deployments https://github.com/NVIDIA/cuda-quantum/actions/runs/22159027027

mitchdz (approved):
This looks great. I haven't had time to thoroughly play with the results but I couldn't visually find any issues. Let's get this in.

PR #3994, opened 2026-02-18
Support AOT compilation with device call declarations for remote backends

1tnguyen:
### Description

For remote backends, there might be a usage scenario whereby `device_call` refers to only function declarations, without definitions. 
The device calls will be resolved on the remote backends, while locally, these function definitions don't need to be present.

A dummy implementation for these device call declarations is needed so that the AOT compilation can proceed. During remote execution, i.e., JIT, we submit the declarations in the IR, with no implementation anyway.

This PR:

- Support injection of trap (unreachable) function definitions when compiling for remote hardware. This would allow AOT compilation for undefined device functions. 

- This support is only active for remote hardware. For local emulation or simulation, the device call function implementations need to be present.

- Add support for `quake` as a codegen option.

- Add a mock `quake`-based target for testing purposes.

### DIFF
--- a/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
+++ b/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
@@ -69,38 +69,40 @@ public:
       // provider's runtime library.
 
       // (1) Add a trap implementation for this device function declaration.
-      auto insPt = rewriter.saveInsertionPoint();
-      // Add an entry block
-      auto &entryBlock = *devFunc.addEntryBlock();
-      rewriter.setInsertionPointToStart(&entryBlock);
-      // Create a call to the trap intrinsic.
-      // Error code 2 is used to indicate illegal execution of unreachable code.
-      Value errorCodeTwo =
-          rewriter.create<arith::ConstantIntOp>(devcall.getLoc(), 2, 64);
-      rewriter.create<func::CallOp>(devcall.getLoc(), TypeRange{},
-                                    cudaq::opt::QISTrap,
-                                    ValueRange{errorCodeTwo});
-      // For return (after the trap), load from nullptr to create return value
-      // of the same type as the device function, i.e., `return *(T*)nullptr;`
-      // for return type `T`.
-      // Note: this will never be executed because of the trap above. It's only
-      // to create a valid IR with the correct return type for the function.
-      SmallVector<Value> trapResults;
-      for (Type resTy : devFunc.getFunctionType().getResults()) {
-        auto nullPtr = rewriter.create<arith::ConstantOp>(
-            devcall.getLoc(),
-            rewriter.getZeroAttr(rewriter.getIntegerType(64)));
-        auto ptrTy = cudaq::cc::PointerType::get(resTy);
-        auto castedNullPtr = rewriter.create<cudaq::cc::CastOp>(
-            devcall.getLoc(), ptrTy, nullPtr);
-        auto loadedVal =
-            rewriter.create<cudaq::cc::LoadOp>(devcall.getLoc(), castedNullPtr);
-        trapResults.push_back(loadedVal);
+      {
+        OpBuilder::InsertionGuard guard(rewriter);
+        // Add an entry block
+        auto &entryBlock = *devFunc.addEntryBlock();
+        rewriter.setInsertionPointToStart(&entryBlock);
+        // Create a call to the trap intrinsic.
+        // Error code 2 is used to indicate illegal execution of unreachable
+        // code.
+        Value errorCodeTwo =
+            rewriter.create<arith::ConstantIntOp>(devcall.getLoc(), 2, 64);
+        rewriter.create<func::CallOp>(devcall.getLoc(), TypeRange{},
+                                      cudaq::opt::QISTrap,
+                                      ValueRange{errorCodeTwo});
+        // For return (after the trap), load from nullptr to create return value
+        // of the same type as the device function, i.e., `return *(T*)nullptr;`
+        // for return type `T`.
+        // Note: this will never be executed because of the trap above. It's
+        // only to create a valid IR with the correct return type for the
+        // function.
+        SmallVector<Value> trapResults;
+        for (Type resTy : devFunc.getFunctionType().getResults()) {
+          auto nullPtr = rewriter.create<arith::ConstantOp>(
+              devcall.getLoc(),
+              rewriter.getZeroAttr(rewriter.getIntegerType(64)));
+          auto ptrTy = cudaq::cc::PointerType::get(resTy);
+          auto castedNullPtr = rewriter.create<cudaq::cc::CastOp>(
+              devcall.getLoc(), ptrTy, nullPtr);
+          auto loadedVal = rewriter.create<cudaq::cc::LoadOp>(devcall.getLoc(),
+                                                              castedNullPtr);
+          trapResults.push_back(loadedVal);
+        }
+
+        rewriter.create<func::ReturnOp>(devcall.getLoc(), trapResults);
       }
-
-      rewriter.create<func::ReturnOp>(devcall.getLoc(), trapResults);
-      rewriter.restoreInsertionPoint(insPt);
-
       // (2) Set this trap function as private and weak_odr linkage, to allow
       // multiple definitions across translation units without linker errors.
       // For example, compiling for a remote hardware provider with the actual
@@ -114,6 +116,11 @@ public:
 
       // (3) Replace the device call with a no-inline call to prevent inlining
       // of the trap function.
+      // We use a no-inline call here to ensure that the call to the device
+      // function is preserved as a call in the IR (even in the presence of the
+      // trap implementation). If the actual implementation is provided at link
+      // time, it will be used instead of the trap implementation due to the
+      // weak_odr linkage.
       rewriter.replaceOpWithNewOp<cudaq::cc::NoInlineCallOp>(
           devcall, devFunc.getFunctionType().getResults(), devFuncName,
           devcall.getArgs());
--- a/runtime/common/CodeGenConfig.cpp
+++ b/runtime/common/CodeGenConfig.cpp
@@ -53,7 +53,7 @@ cudaq::parseCodeGenTranslation(const std::string &codegenTranslation) {
   auto [codeGenName, codeGenVersion, codeGenOptions] =
       parseCodeGenTranslationString(codegenTranslation);
 
-  if (codeGenName == "quake") {
+  if (codeGenName == "nop") {
     CodeGenConfig config = {.profile = codeGenName,
                             .isQIRProfile = false,
                             .isAdaptiveProfile = true,
--- a/runtime/internal/compiler/Compiler.cpp
+++ b/runtime/internal/compiler/Compiler.cpp
@@ -175,13 +175,13 @@ Compiler::Compiler(cudaq::ServerHelper *serverHelper,
 
     // 5. Apply the target-agnostic finalization passes. This lowers the IR to
     // CFG form.
-    // If this is not emulation, and the codegen translation is quake, then we
-    // want to keep device calls as-is, to be submitted to the server for
-    // lowering and execution.
+    // If this is not emulation, and the codegen translation is nop (dumping
+    // CUDA-Q MLIR), then we want to keep device calls as-is, to be submitted to
+    // the server for lowering and execution.
     passPipelineConfig +=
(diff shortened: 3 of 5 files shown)


unittests/backends/quake_backend/test_app.cpp
```diff
@@ -0,0 +1,129 @@
+
```
schweitzpgi:
```suggestion
```

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -2210,10 +2210,80 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+
+      auto [symbol, devFuncTy] = [&]() {
+        if (funcConst) {
+          // We have a function declaration, so we can look up the function type directly.
+          auto symbol = funcConst.getValue();
+          auto devFunc = module.lookupSymbol<func::FuncOp>(symbol);
+          devFunc->setAttr(cudaq::deviceCallAttrName, builder.getUnitAttr());
+          return std::make_pair(symbol, cast<FunctionType>(funcConst.getType()));
+        }
+        StringRef funcNameStr =
```
schweitzpgi: I'm clearly missing something here.

It looks like we're going down the road of trying to change the semantics of C++ here to implement something in a way that breaks C++'s sense of being a strongly typed language. The big question is then: why?

Why not compile this as normal C++ code and require that the remote callback functions simply be declared like normal C++ functions? If they are made weak symbols, that should get around any host-side linking issues, no?

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -100,6 +159,23 @@ class ResolveDevicePtrOpPat
...
 
+// Remove auto-generated device call declarations that are not resolved.
+// We have already replaced all resolved auto-generated device calls with traps.
+class RemoveAutoGenDeviceCallDecl : public OpRewritePattern<func::FuncOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(func::FuncOp func,
+                                PatternRewriter &rewriter) const override {
+    if (func->getAttr(cudaq::autoDeviceCallAttrName) && func.isDeclaration()) {
```
schweitzpgi: I don't think we need to write this pattern. The SymbolDCE pass should do this for us for free.

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -52,6 +53,64 @@ class QIRVendorDeviceCallPat
     hash.final(result);
     std::uint32_t callbackCode = result.low();
 
+    if (devFunc->getAttr(cudaq::autoDeviceCallAttrName) &&
+        devFunc.isDeclaration()) {
+      // This is an auto-generated device call, e.g., arbitrary functions
```
schweitzpgi: I like this idea and its better than a weak symbol (which would also trap but with an annoying sigsegv).

So we might declare these remote-side functions as real C++ (so they are properly type checked) but with default implementations of generating traps in the IR. Hmm...

lib/Optimizer/Transforms/Pipelines.cpp
```diff
@@ -144,8 +152,10 @@ void cudaq::opt::createTargetFinalizePipeline(OpPassManager &pm) {
   pm.addPass(createSymbolDCEPass());
 }
 
-static void createJITTargetFinalizePipeline(OpPassManager &pm) {
-  pm.addPass(cudaq::opt::createDistributedDeviceCall());
+static void createJITTargetFinalizePipeline(
+    OpPassManager &pm, const TargetFinalizationJitPipelineOptions &options) {
+  if (options.lowerDeviceCalls)
+    pm.addPass(cudaq::opt::createDistributedDeviceCall());
```
schweitzpgi, line 157: Are we needing to run everything in this pass (all that it entails) or are we really just wanting to massage the host-is-trap and remote-is-a-call distinctions?

The IR could look like (pseudo-code here)
```mlir
func.func @callback(...) attributes { "cudaq-remote-implementation" } {
  cc.trap
  cc.return
}

....
  cc.device_call @callback()
```
and a rather simple pass run when generating the transport layer code to erase the bodies of any function tagged with `"cudaq-remote-implementation"`.

schweitzpgi:
It might be useful to add a new Op for this, of course. The semantics of this, if I am understanding it, are really not the same as `device_call` after all.

amccaskey:
The original intent for `device_call` was that external partners could deploy their own device libraries via standard headers and library implementations. This should be sufficient for remote integrations as well - just in that case the library can sit on the server and the string can be inferred on the there as well (they can even mark the functions `extern "C"` to avoid name mangling). Let me know if I'm missing some detail here. Why use the string here?

1tnguyen:
> The original intent for `device_call` was that external partners could deploy their own device libraries via standard headers and library implementations. This should be sufficient for remote integrations as well - just in that case the library can sit on the server and the string can be inferred on the there as well (they can even mark the functions `extern "C"` to avoid name mangling). Let me know if I'm missing some detail here. Why use the string here?

The main idea of the string-based support is flexibility for remote providers. For example, we don't require headers for AOT compilation; the backend may provide an endpoint to query the list of supported functions for JIT-time validation.
If this is undesirable, we can limit this flexibility and require the declarations in a header file.

lib/Frontend/nvqpp/ConvertExpr.cpp
1tnguyen: As discussed, this function-name-only approach is abandoned. This PR has a more focused scope of supporting remote callback functions that are declared but may not be defined.

include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -422,6 +422,11 @@ def DistributedDeviceCall : Pass<"distributed-device-call", "mlir::ModuleOp"> {
   }];
 
   let dependentDialects = ["mlir::LLVM::LLVMDialect"];
+
+  let options = [
+    Option<"insertTrapImplementation", "insert-trap", "bool", /*default=*/"false",
+      "Enable pass to insert a default implementation (runtime trap) for device call (default=false). The default implementation has weak linkage, i.e., allowed to be overridden at link time.">
```
schweitzpgi: nit: 80 columns?

runtime/common/Compiler.cpp
```diff
@@ -176,7 +176,14 @@ Compiler::Compiler(ServerHelper *serverHelper,
 
     // 5. Apply the target-agnostic finalization passes. This lowers the IR to
     // CFG form.
-    passPipelineConfig += ",jit-finalize-pipeline";
+    // If this is not emulation, and the codegen translation is quake, then we
+    // want to keep device calls as-is, to be submitted to the server for
+    // lowering and execution.
+    passPipelineConfig +=
+        ",jit-finalize-pipeline{lower-device-calls=" +
+        std::string{(codegenTranslation == "quake" && !emulate) ? "false"
```
schweitzpgi: "quake" isn't really any sort of transport layer. Can we call this something like "weak_host_symbols"?

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -100,6 +174,26 @@ class ResolveDevicePtrOpPat
   }
 };
 
+// Remove device call declarations that we have already inserted trap
+// implementations for, to avoid duplicate declarations of the same device
+// function.
+class RemoveDuplicateDeviceCallDeclaration
```
schweitzpgi: Instead of removing these, why aren't they rewritten in place? You could just add a body to the declaration `func::FuncOp`, right?

runtime/common/Compiler.cpp
1tnguyen: This logic here is actually unrelated to the support for weak host symbols during AOT compilation.
Rather, it's meant to preserve the `cc.device_call` in the MLIR payload for job submission, not converting it to a standard call op when `quake` is selected. The remote backend would consume the `quake` code; hence, it understands both `quake` and `cc` dialects.

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: I have simplified this in https://github.com/NVIDIA/cuda-quantum/pull/3994/changes/4f073d456d3cb69b1b9a4369621d04f1e3d0a392.

include/cudaq/Optimizer/Transforms/Passes.td
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/3994/changes/4f073d456d3cb69b1b9a4369621d04f1e3d0a392

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: Good point! Fixed as suggested in https://github.com/NVIDIA/cuda-quantum/pull/3994/changes/4f073d456d3cb69b1b9a4369621d04f1e3d0a392

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -52,6 +58,69 @@ class QIRVendorDeviceCallPat
...
+      // remote hardware providers), we want to insert a trap implementation for
+      // any unresolved device function (declaration only), so that we can
+      // perform AOT compilation without needing the actual device function
+      // definitions. This trap function will never be executed as the remote
+      // JIT pipeline would not be using the `device_call` functions anyway.
+      // Rather, these functions will only be resolved at runtime by the remote
+      // provider's runtime library.
+
+      // (1) Add a trap implementation for this device function declaration.
+      auto insPt = rewriter.saveInsertionPoint();
```
schweitzpgi: This is fine, but you can also use the `InsertionGuard` idiom for this sort of thing.

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -52,6 +58,69 @@ class QIRVendorDeviceCallPat
...
+      // device call library linkage (even though unused) should not cause any
+      // problems.
+      devFunc.setPrivate();
+      auto weakOdrLinkage = mlir::LLVM::linkage::Linkage::WeakODR;
+      auto linkage =
+          mlir::LLVM::LinkageAttr::get(rewriter.getContext(), weakOdrLinkage);
+      devFunc->setAttr("llvm.linkage", linkage);
+
+      // (3) Replace the device call with a no-inline call to prevent inlining
+      // of the trap function.
```
schweitzpgi, line 118: Maybe add a few more words on why we don't want it inlined here in the comment.

runtime/common/CodeGenConfig.cpp
```diff
@@ -53,6 +53,15 @@ cudaq::parseCodeGenTranslation(const std::string &codegenTranslation) {
   auto [codeGenName, codeGenVersion, codeGenOptions] =
       parseCodeGenTranslationString(codegenTranslation);
 
+  if (codeGenName == "quake") {
```
schweitzpgi: I still find this a bit jarring since quake code really doesn't "lower to" quake code. I think we need better terminology here...

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: Good point! Fixed (used the `InsertionGuard` idiom) in bacafb9a7d33f0f1e07f62a2d1fcabb902385d72

lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: Fixed: added comments in [bacafb9](https://github.com/NVIDIA/cuda-quantum/pull/3994/commits/bacafb9a7d33f0f1e07f62a2d1fcabb902385d72)

runtime/common/CodeGenConfig.cpp
1tnguyen: Agreed! I've updated the terminology: using `nop`, as suggested, instead.

1tnguyen:
> This all looks good.
> 
> I just think that "quake lowering to quake" sounds like an oxymoron and will be confusing to n00bs. This is more of a "NOP backend". It's not even lowering to CC, right?

Agreed. I've updated the code to use "`nop`" instead.

PR #3996, opened 2026-02-18
[ci] Fix Mock QPU tests and ensure CI runs them

lmondada:
Spotted by @amccaskey: our python backend tests that relied on mock QPUs where broken and not being run since https://github.com/NVIDIA/cuda-quantum/pull/3747.

This wasn't caught because the `ImportError` that was raised was silenced and the tests were subsequently skipped. This PR fixes this.

I have also added a `MOCK_NO_SKIP` env variable that can be used as a flag by CI to disallow test skipping on import failure, so that such an issue does not get missed again.

### DIFF
--- a/.github/workflows/test_in_devenv.yml
+++ b/.github/workflows/test_in_devenv.yml
@@ -250,7 +250,7 @@ jobs:
               exit 1
             fi
             for backendTest in python/tests/backends/*.py; do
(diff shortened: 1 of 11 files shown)


lmondada:
Ah, I'm now seeing this was already resolved by @khalatepradnya in https://github.com/NVIDIA/cuda-quantum/pull/3988. Never mind!

PR #4000, opened 2026-02-18
[Python] Measurement result type

khalatepradnya:
* Python frontend support for the `measure_result` type, enabling deferred discrimination of measurement results. 
* Companion to the C++ frontend changes in #3800.

#### Breaking Changes
- `mz()` now returns `measure_result` (an opaque handle), not `bool`.
  Discrimination to a classical value happens only when needed (boolean context, comparison, return as `bool`, etc.). Kernels that relied on `mz()` producing `i1` immediately will see different IR.
- Entry-point kernels cannot return `measure_result` or containers thereof (`list[measure_result]`, `tuple[measure_result, ...]`,   dataclasses with `measure_result` fields). Use them as device-only kernels called from other kernels that return `bool`.


python/cudaq/kernel/utils.py
```diff
@@ -37,6 +40,46 @@
 # Keep a global registry of all registered custom operations.
 globalRegisteredOperations = {}
 
+
+def check_no_measure_result_return(fn, *, context="direct invocation"):
+    """Raises if function's return annotation contains measure_result."""
```
sacpis, line 45: A question. Will a function's return annotation ever change? In that case, would we want to store this in a cache with the `fn` being the key? Because for a kernel with say, 1000 shots, this will be called those many times.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -1855,13 +1907,19 @@ def storedAsValue(val):
...
+            def _containsMeasureType(ty):
+                if quake.MeasureType.isinstance(ty):
+                    return True
+                if quake.MeasurementsType.isinstance(ty):
+                    return True
+                if cc.StructType.isinstance(ty):
+                    return any(
+                        _containsMeasureType(t)
+                        for t in cc.StructType.getTypes(ty))
+                return False
```
sacpis: Should we make this function static or put it at module level?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -841,13 +849,30 @@ cudaq::marshal_arguments_for_module_launch(ModuleOp mod, py::args runtimeArgs,
   return args;
 }
 
+static bool containsMeasureResult(Type ty) {
+  if (isa<quake::MeasureType>(ty) || isa<quake::MeasurementsType>(ty))
+    return true;
+  if (auto vecTy = dyn_cast<cudaq::cc::StdvecType>(ty))
+    if (isa<quake::MeasureType>(vecTy.getElementType()))
+      return true;
```
sacpis: Can we confirm if this branch is executed?

sacpis (commented):

python/cudaq/kernel/utils.py
khalatepradnya: This check is invoked only once per user-facing API call, not per shot.

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Sure, fixed in [7b02410](https://github.com/NVIDIA/cuda-quantum/pull/4000/commits/7b02410dd8c9f0f81daa1a80237ec16eb844c341)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
khalatepradnya: Good catch! No it is relic of the past (changes prior to PR #4208 ). Removed.

python/runtime/cudaq/qis/py_measure_result.cpp
```diff
@@ -0,0 +1,47 @@
...
+                    "The integer measurement value (0 or 1).")
+      .def("__bool__",
+           [](const measure_result &self) { return static_cast<bool>(self); })
+      .def("__int__", [](const measure_result &self) { return self.value; })
+      .def("__eq__", [](const measure_result &self,
+                        const measure_result &other) { return self == other; })
+      .def("__eq__",
+           [](const measure_result &self, bool other) {
+             return static_cast<bool>(self) == other;
+           })
```
sacpis, line 34: Seems like we will need to add `__hash__` method. In Python 3, any class that defines `__eq__` without `__hash__` becomes unhasable.

We can add something like

```
.def("__hash__", [](const measure_result &self) {
    return std::hash<int>{}(self.value);         
})
```

python/tests/kernel/test_measure_result.py
```diff
@@ -0,0 +1,224 @@
...
+        x(q[0])
+        r0 = mz(q[0])
+        r1 = mz(q[1])
+        t = (r0, r1)
+        return t
+
+    @cudaq.kernel
+    def entry_point() -> tuple[bool, bool]:
+        q = cudaq.qvector(2)
+        r = device_kernel(q)
```
sacpis, line 187: Can we add a test to check for `==` operator?

```
if r[0] == r[1]:
    do_something...
```

Here both will have value as 1 but their unique ids will be different, no?

sacpis (commented):
Thanks @khalatepradnya. Left 2 comments.

python/runtime/cudaq/qis/py_measure_result.cpp
khalatepradnya: Added.

python/tests/kernel/test_measure_result.py
khalatepradnya: Good point. Note that unique_id is not assigned by the default simulator, so they will be identical.

python/runtime/cudaq/qis/py_measure_result.cpp
```diff
@@ -0,0 +1,53 @@
...
+              "The integer measurement value (0 or 1).")
+      .def("__bool__",
+           [](const measure_result &self) { return static_cast<bool>(self); })
+      .def("__int__", [](const measure_result &self) { return self.value; })
+      .def("__eq__", [](const measure_result &self,
+                        const measure_result &other) { return self == other; })
+      .def("__eq__",
+           [](const measure_result &self, bool other) {
+             return static_cast<bool>(self) == other;
+           })
```
sacpis, line 33: Just a last issue with these bool overloads of `__eq__` and `__ne__` is that `__eq__/__ne__(measure_result, bool)` returns static_cast<bool>(self) == other, but `__hash__` uses value and unique_id. So `m == True ` can hold while the objects hashes are not equal. In Python, I think we should have the same hash values for similar objects, right?

Here, I think we can drop these bool overloads, which will make Python to use `__bool__/__init__` for comparison. This will be only triggered when we get `bool(m) == True`. Then `__hash__` with unique_id and value won't be an issue.

sacpis (approved):
Overall LGTM. Thanks @khalatepradnya. Just have one last concern.

khalatepradnya:
Closing owing to change in direction...

PR #4001, opened 2026-02-18
Add measurement noise support and unify PTSBE trace representation

taalexander:
### Description
Add measurement noise to the PTSBE pipeline. Previously, the tracer discarded mz() calls and the execution path assumed terminal measurement of all qubits with no noise. This meant noise channels registered on "mz" were silently ignored which is not the case on the non-ptsbe path.

The core change introduces `buildPTSBETrace()`, which converts the raw `cudaq::Trace` into a single interleaved `PTSBETrace` containing Gate, Noise, and Measurement entries with all channels already resolved. This replaces the prior design where `PTSBatch` carried a raw `kernelTrace` alongside a separate noise_sites vector, and downstream code had to coordinate between them through fragile index-based lookups. The unified trace becomes the sole input to noise extraction, trajectory merging, and execution, eliminating an entire class of index-mismatch bugs.

Measurements are now recorded in the kernel trace during tracing (both `BasicExecutionManager` for library-mode kernels and `CircuitSimulator` for MLIR-compiled kernels). When a kernel omits explicit `mz()` calls, `buildPTSBETrace` generates per-qubit implicit measurements matching standard `cudaq::sample() `behavior, with per-qubit noise model queries so that channels registered via `add_channel("mz", {q}, ...)` are matched correctly.

### DIFF
--- a/runtime/cudaq/algorithms/draw.cpp
+++ b/runtime/cudaq/algorithms/draw.cpp
@@ -411,6 +411,11 @@ std::vector<Layer> layers_from_trace(const Trace &trace) {
 
   std::size_t ref = 0;
   for (const auto &inst : trace) {
+    if (inst.type != cudaq::TraceInstructionType::Gate) {
+      ref += 1;
+      continue;
+    }
+
     std::vector<Diagram::Wire> wires = convertToIDs(inst.targets);
     const auto minmax_wires = std::minmax_element(begin(wires), end(wires));
     auto min_dwire = *minmax_wires.first;
@@ -448,6 +453,9 @@ boxes_from_trace(const Trace &trace) {
 
   // same iteration order as in layers_from_trace
   for (const auto &inst : trace) {
+    if (inst.type != cudaq::TraceInstructionType::Gate)
+      continue;
+
     std::vector<Diagram::Wire> wires = convertToIDs(inst.targets);
     std::sort(wires.begin(), wires.end());
 


unittests/ptsbe/ExecutePTSBETester.cpp
```diff
@@ -51,17 +51,13 @@ std::vector<cudaq::sample_result> runPTSBETest(SimulatorType &sim,
 CUDAQ_TEST(ExecutePTSBETest, ThrowsWithoutExecutionContext) {
   QppSimulator sim;
 
-  Trace trace;
-  trace.appendInstruction("h", {}, {}, {QuditInfo(2, 0)});
-
   PTSBatch batch;
-  batch.kernelTrace = trace;
+  batch.trace = {{ptsbe::TraceInstructionType::Gate, "h", {0}, {}, {}}};
```
sacpis: nit: should we pull it our as a global trace as its is being used elsewhere in the test? Same for x.

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
```diff
@@ -24,7 +24,8 @@ namespace cudaq::ptsbe {
 /// Stores the validated kraus_channel which contains the Kraus operators and
 /// their probabilities for that noise channel.
 struct NoisePoint {
-  /// @brief Location in the circuit (gate index)
+  /// @brief Index into a noise location in thePTSBE instruction sequence.
```
sacpis: nit: missing space `thePTSBE`

unittests/ptsbe/ExecutePTSBETester.cpp
```diff
@@ -182,7 +153,7 @@ CUDAQ_TEST(ExecutePTSBETest, EmptyInputsReturnEmpty) {
   // Test 2: Empty measureQubits
   {
     PTSBatch batch;
-    batch.kernelTrace = trace;
+    batch.trace = {{ptsbe::TraceInstructionType::Gate, "h", {0}, {}, {}}};
```
sacpis: same here

runtime/cudaq/ptsbe/NoiseExtractor.h
```diff
@@ -8,16 +8,32 @@
...
+/// Converts QuditInfo targets/controls to plain qubit indices. All instruction
+/// types are preserved. Gate and measurement entries pass through. Noise
+/// entries (from apply_noise) have their channels resolved via the noise model.
+/// The resulting vector defines the unified index space for circuit_location
+/// referenced by NoisePoint.
+///
+/// @param trace Raw circuit trace (may contain Gate, Noise, and Measurement)
+/// @param noise_model Noise model used to resolve inline apply_noise channels
+/// @return PTSBETrace with resolved channels for Noise entries
+[[nodiscard]] PTSBETrace buildPTSBETrace(const cudaq::Trace &trace,
```
sacpis: Should the `buildPTSBETrace` declaration go in `PTSBESample.h` as it is defined in `PTSBESample.cpp`?

taalexander: Woops, let me align these.

taalexander: Fixed in f067d2215a

unittests/ptsbe/ExecutePTSBETester.cpp
taalexander: Fixed in f067d2215a

runtime/cudaq/ptsbe/PTSSamplingStrategy.h
taalexander: Fixed in f067d2215a

unittests/ptsbe/ExecutePTSBETester.cpp
taalexander: Fixed in f067d2215a

Issue #3965, opened 2026-02-16
[python] [bug] When invoked, kernel resolves captured arguments in calling scope


lmondada:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The following MWE shows a kernel defined in a separate scope (in this case, within a function, but could be a separate module). The captured kernel should be resolved in the definition scope of the kernel, but instead it is using the calling scope, causing a `RuntimeError: 'inner' is not available in this scope` error message.

### Steps to reproduce the bug

```python
import cudaq

def another_module():

    @cudaq.kernel
    def inner(q: cudaq.qubit):
        x(q)

    @cudaq.kernel
    def outer():
        q = cudaq.qubit()
        inner(q)

    return {'thekernel': outer}

m = another_module()
thekernel = m['thekernel']
thekernel()
```

### Expected behavior

The above should work.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**:

schweitzpgi:
This is not a bug. Arguments are intentionally resolved at the point of a call in order to obey Python semantics. Any other interpretation (like previous embeddings of CUDA-Q in Python) are simply incorrect. We should not revert to incorrect behavior.

schweitzpgi:
I was thrown by the title of this, sorry.

I agree that callable symbols in E(nclosing) scope at the definition point ought to be resolved properly.

Feel free to re-open this issue but limit it to that issue. When I first read this, it read like: "lets unwind all of the last 5 months of work because we like our original bugs."

schweitzpgi:
This example exposes the fallacy of the #3948 solution. I'll comment there...

lmondada:
[...]

PR #4005, opened 2026-02-19
Captured args are resolved in kernel definition scope

lmondada:
Python follows lexical scoping, which means that captured variables in a function `foo` must resolve in the scope where `foo` is defined, and not in the caller's scope. This is different than arguments passed to a function, which are (obviously) resolved in the caller's scope. For instance:

```python
# Module mod_a.py

a = 3
def foo():
    print(a)

# Module mod_b.py
from mod_a import foo

a = "something else"
foo()
```

should print `3`.

This PR adjusts the behaviour of captured arguments in kernels to behave similarly, by keeping a reference to the frame that the kernel was defined in.

See tests and https://github.com/NVIDIA/cuda-quantum/issues/3965.

closes https://github.com/NVIDIA/cuda-quantum/issues/3965

### DIFF
--- a/python/cudaq/runtime/ptsbe.py
+++ b/python/cudaq/runtime/ptsbe.py
@@ -103,19 +103,19 @@ def sample(kernel,
         argSets = __createArgumentSet(*args)
         results = []
         for argSet in argSets:
-            specMod, processedArgs = decorator.handle_call_arguments(*argSet)
+            processedArgs, module = decorator.prepare_call(*argSet)
             retTy = decorator.get_none_type()
             result = cudaq_runtime.ptsbe.sample_impl(
-                decorator.uniqName, specMod, retTy, shots_count, noise_model,
+                decorator.uniqName, module, retTy, shots_count, noise_model,
                 max_trajectories, sampling_strategy, shot_allocation,
                 return_execution_data, *processedArgs)
             results.append(result)
         return results
 
-    specMod, processedArgs = decorator.handle_call_arguments(*args)
+    processedArgs, module = decorator.prepare_call(*args)
     retTy = decorator.get_none_type()
 
-    return cudaq_runtime.ptsbe.sample_impl(decorator.uniqName, specMod, retTy,
+    return cudaq_runtime.ptsbe.sample_impl(decorator.uniqName, module, retTy,
                                            shots_count, noise_model,
                                            max_trajectories, sampling_strategy,
                                            shot_allocation,
@@ -160,14 +160,14 @@ def sample_async(kernel,
     if noise_model is None:
         noise_model = cudaq_runtime.NoiseModel()
 
-    specMod, processedArgs = decorator.handle_call_arguments(*args)
+    processedArgs, module = decorator.prepare_call(*args)
     retTy = decorator.get_none_type()
 
     impl = cudaq_runtime.ptsbe.sample_async_impl(
-        decorator.uniqName, specMod, retTy, shots_count, noise_model,
+        decorator.uniqName, module, retTy, shots_count, noise_model,
         max_trajectories, sampling_strategy, shot_allocation,
         return_execution_data, *processedArgs)
 
-    result = AsyncSampleResult(impl, specMod)
+    result = AsyncSampleResult(impl, module)
     result._noise_model = noise_model  # prevent GC until .get() is called
     return result


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -536,16 +519,37 @@ def handle_call_arguments(self, *args):
...
+        Resolve the arguments passed to the decorator at call site.
+        """
+        return [
+            self.process_argument(arg, arg_type)
+            for arg, arg_type in zip(args, self.arg_types())
+        ]
+
+    def prepare_call(self, *args):
+        """
+        Process call site arguments, capture lifted arguments and retrieve
```
bettinaheim, line 548: nit: worth capturing in the doc comment that this creates a clone of the qkeModule.

python/tests/kernel/test_deferred_compilation.py
```diff
@@ -221,6 +221,157 @@ def bar():
...
+
+
+def _make_caller_callee_pair():
+    """Helper: define a caller/callee pair inside a nested scope and return them."""
+
+    @cudaq.kernel
+    def inner(q: cudaq.qubit):
+        x(q)
+
+    @cudaq.kernel
```
bettinaheim, line 236: nit: Edge case that is not properly dealt with:
using the `@kernel` attribute in an inner function currently I believe will be accepted by the bridge (the attribute is simply ignored). I believe the logic to get the correct frame relies on functions defined within kernels not having that attribute. That's fine - but if it is a quick thing to do it would be nice to make the bridge give an error if you do this:
```
    @cudaq.kernel
    def kernel(q: cudaq.qubit):
        @cudaq.kernel # error?
        def inner_fct():
            ...
```

python/cudaq/kernel/ast_bridge.py
```diff
@@ -3909,9 +3886,11 @@ def get_item_type(pyval):
                 return cc.StdvecType.get(base_elTy)
             elif isinstance(pyval, ast.Call):
                 if isinstance(pyval.func, ast.Name):
+                    from .kernel_decorator import isa_kernel_decorator
                     # supported for calls but not here: 'range', 'enumerate'
-                    decorator = recover_kernel_decorator(pyval.func.id)
-                    if decorator:
+                    decorator = recover_value_of_or_none(
```
schweitzpgi, line 3918: Why are you getting rid of the encapsulation of this functionality to find the decorator?

schweitzpgi:
What would be helpful for reviews is to avoid gratuitous symbol name changes in the PR and focus on functional changes or (exclusively or) perfecting the spelling of symbols in the code in separate PRs. That will let me focus on how you are changing things rather than spending bandwidth tripping over spelling changes. Thanx.

python/cudaq/kernel/kernel_builder.py
```diff
@@ -1374,11 +1373,6 @@ def resolve_callable_arg(self, insPt, target):
         fn = recover_func_op(self.module, fulluniq)
 
         # build the closure to capture the lifted `args`
-        thisPyMod = recover_calling_module()
```
schweitzpgi, line 1377: Need to discuss this offline.

python/cudaq/kernel/ast_bridge.py
lmondada: `recover_kernel_decorator` does the same thing as `recover_value_of_or_none`. I have just consolidated and removed one of the two redundant function definitions.

lmondada:
> What would be helpful for reviews is to avoid gratuitous symbol name changes in the PR and focus on functional changes or (exclusively or) perfecting the spelling of symbols in the code in separate PRs.

It's tricky. I 100% agree name changes can be confusing and I always try to keep them to a minimum. At the same time there are a lot of function names that don't do what they say. I only changed names where I had to modify code anyway. I'll aim to be even more selective next time.

schweitzpgi:
> > What would be helpful for reviews is to avoid gratuitous symbol name changes in the PR and focus on functional changes or (exclusively or) perfecting the spelling of symbols in the code in separate PRs.
> 
> It's tricky. I 100% agree name changes can be confusing and I always try to keep them to a minimum. At the same time there are a lot of function names that don't do what they say. I only changed names where I had to modify code anyway. I'll aim to be even more selective next time.

Thank you! I appreciate the need for both.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2530,18 +2509,16 @@ def processQuakeCtor(opName,
...
-            else:
-                decorator = recover_kernel_decorator(name)
-                if decorator is None and name in self.parentVariables:
-                    from .kernel_decorator import isa_kernel_decorator
-                    var = self.parentVariables[name]
-                    if isa_kernel_decorator(var):
-                        decorator = var
 
-            if decorator and not name in self.symbolTable:
+            decorator = recover_value_of_or_none(name, self.defFrame)
```
atgeller, line 2540: We'll probably need to make similar changes to https://github.com/NVIDIA/cuda-quantum/blob/63535ff3f58a621d820694b08290873673ec30f2/python/cudaq/kernel/ast_bridge.py#L2639

python/cudaq/kernel/kernel_decorator.py
lmondada: done

python/tests/kernel/test_deferred_compilation.py
lmondada: I've resolved this, too.

python/cudaq/kernel/ast_bridge.py
lmondada: yes, good catch. I will get this in first, and then create a follow up task to sort out any similar bugs in the AST bridge.

PR #4008, opened 2026-02-19
Removing cuda-mlir-runtime dependency on nvqir.

Renaud-K:
There was coupling between the compiler `cuda-mlir-runtime` and the resource count simulator `nvqir` because the compiler was trying to directly update the resource count of the simulator. With these changes the resource counts are passed as a return value of the compiler and the resource count simulator is updated in the QPU.

The `KernelExecution` has now an additional `Resources` field and certainly shows that it is being overloaded as a return value but this will have to be addressed by other refactors when compilation is broken up into different policies which can then have different return types. 

`nvqir` was publicly linking* with `cuda-common` and other libraries and  `cuda-mlir-runtime` was publicly linking `nvqir`. Removing the dependency on `nvqir` means that existing dependencies have to be made explicit. So `cuda-mlir-runtime` still depends on `cudaq-common` for resouce counts, device code registry and other utilities and the dependency between QPUs and `nvqir` is now apparent.

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -32,7 +32,7 @@
 namespace nvqir {
 // QIR helper to retrieve the output log.
 std::string_view getQirOutputLog();
-cudaq::Resources *getResourceCounts();
+void setResourceCounts(cudaq::Resources &&);
 bool isUsingResourceCounterSimulator();
 } // namespace nvqir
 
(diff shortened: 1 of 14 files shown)


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -32,7 +32,7 @@
 namespace nvqir {
 // QIR helper to retrieve the output log.
 std::string_view getQirOutputLog();
-cudaq::Resources *getResourceCounts();
+void setResourceCounts(cudaq::Resources &&);
```
atgeller, line 35: It would be nice to eventually replace this with a nicer way to pass information to the simulator/other backend, which is out of scope for this PR as we discussed, but may be worth making an issue to remember to come back to this someday when we have some time.

Renaud-K: I have put some thoughts into this but I don't see how we can avoid a `setResourceCounts` method somewhere. It would be nice if the simulator weren't a global object. One of the goal of the refactor is to scope the simulators within the QPU in which case it becomes a local object. Then `setResourceCounts` would directly call a member function.

PR #4010, opened 2026-02-20
[Realtime] Replace HTML docs with markdown

1tnguyen:
### Description

- Replace the HTML docs with markdown files as they are more amendable to manual editing (e.g., spell checker fixes)

- Fix spelling and linting errors in these documents. This should give this feature branch a clean CI state.


sacpis:
Shouldn't we keep both the extensions? As we have it in CUDA-Q.

1tnguyen:
> Shouldn't we keep both the extensions? As we have it in CUDA-Q.

Yes, HTML is supported, but these HTML documents are rendered from a Markdown source, so we replace them with the source. This would make incremental updates easier to review (e.g., changes to code blocks would be very hard to see in HTML docs)

sacpis:
> > Shouldn't we keep both the extensions? As we have it in CUDA-Q.
> 
> Yes, HTML is supported, but these HTML documents are rendered from a Markdown source, so we replace them with the source. This would make incremental updates easier to review (e.g., changes to code blocks would be very hard to see in HTML docs)

I concur. But html docs would be nice from the user's perspective.

PR #4011, opened 2026-02-20
[PTSBE] Fixing proportional shot allocation

sacpis:
Fixing proportional shot allocation with multinomial sampling, applied to all three weighted strategies. 
- Eliminates the silent zero-shot bug where low-probability trajectories received no shots due to integer truncation
- The total shot count is now guaranteed to be exact by construction


sacpis:
Tracked in PR #4012.

PR #4015, opened 2026-02-20
Fixing pr_cleanup job

sacpis:
Fixing pr_cleanup job never running on schedule due to overly strict midnight time check.


mitchdz (approved):
LGTM, thanks @sacpis

PR #4018, opened 2026-02-20
[sample] [C++] Suppress warning when noise model is empty

khalatepradnya:
When using default `sample_options` noise model is empty, yet the simulator tries to apply since it's not null. This PR suppresses that warning message.

* Example 
```
#include <cudaq.h>

__qpu__ void bell_pair() {
  cudaq::qvector q(2);
  h(q[0]);
  x<cudaq::ctrl>(q[0], q[1]);
}

int main() {
  cudaq::sample_options options;
  options.shots = 10;
  auto counts = cudaq::sample(options, bell_pair);
  counts.dump();
  return 0;
}
```

* Before
```
# nvq++ sample_options.cpp && ./a.out
[2026-02-20 23:10:24.198] [warning] [CircuitSimulator.h:785] Applying noise is not supported on custatevec-fp32 simulator.
[2026-02-20 23:10:24.199] [warning] [CircuitSimulator.h:785] Applying noise is not supported on custatevec-fp32 simulator.
{ 00:5 11:5 }
```

* After
```
# nvq++ sample_options.cpp && ./a.out
{ 00:6 11:4 }
```

This would have shown up any time using `explicit_measurements` option in C++.

### DIFF
--- a/docs/sphinx/snippets/python/using/backends/dynamics.py
+++ b/docs/sphinx/snippets/python/using/backends/dynamics.py
@@ -56,20 +56,23 @@ evolution_result = cudaq.evolve(
     store_intermediate_results=cudaq.IntermediateResultSave.ALL)
 #[End Evolve]
 
-#[Begin Plot]
-get_result = lambda idx, res: [
-    exp_vals[idx].expectation() for exp_vals in res.expectation_values()
-]
+try:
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
👍

PR #4020, opened 2026-02-21
Fix matplotlib ImportError

sacpis:
Putting matplotlib import in a try/except block.

Fixes: https://github.com/NVIDIA/cuda-quantum/actions/runs/22245309957/job/64361991617#step:7:20687

### DIFF
--- a/docs/sphinx/snippets/python/using/backends/dynamics.py
+++ b/docs/sphinx/snippets/python/using/backends/dynamics.py
@@ -56,12 +56,12 @@ evolution_result = cudaq.evolve(
     store_intermediate_results=cudaq.IntermediateResultSave.ALL)
 #[End Evolve]
 
-#[Begin Plot]
-get_result = lambda idx, res: [
-    exp_vals[idx].expectation() for exp_vals in res.expectation_values()
-]
-
 try:
+    #[Begin Plot]
+    get_result = lambda idx, res: [
+        exp_vals[idx].expectation() for exp_vals in res.expectation_values()
+    ]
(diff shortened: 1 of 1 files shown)


docs/sphinx/snippets/python/using/backends/dynamics.py
```diff
@@ -61,14 +61,17 @@
...
 
-import matplotlib.pyplot as plt
-
-plt.plot(steps, get_result(0, evolution_result))
-plt.plot(steps, get_result(1, evolution_result))
-plt.plot(steps, get_result(2, evolution_result))
-plt.ylabel("Expectation value")
-plt.xlabel("Time")
-plt.legend(("Sigma-X", "Sigma-Y", "Sigma-Z"))
+try:
```
1tnguyen, line 59: Could we move the entire `#[Begin Plot]` .. `#[End Plot]` block inside the `try` so that we only render that block for the documentation and not show the try..except?

e.g., 

```
try:
   #[Begin Plot]
   get_result = lambda idx, res: [
      p_vals[idx].expectation() for exp_vals in res.expectation_values()
   ]
   import matplotlib.pyplot as plt
   ....
   
   #[End Plot]
 except ImportError:
    pass  
```

sacpis: Fixed in [820a23a](https://github.com/NVIDIA/cuda-quantum/pull/4020/commits/820a23a6df9dc57cc802313dcde31f27070e29ef).

PR #4022, opened 2026-02-22
[PTSBE] Sampling fixes and improvements from verification testing

taalexander:
### Description
These are a result of the sampling and performance issues I identified during my verification testing (see the included notebook for benchmarking based on a generalization of linear XEB for noisy simulator characterization).

Changes:
- Corrected the Monte Carlo sampling algorithm which was using a biased sampler and producing incorrect results.
- `mergeTasksWithTrajectory` now skips identity (non-error) Kraus operators by default in the generic per-trajectory path. An includeIdentity parameter allows the batched simulator to opt in to uniform gate structure, which cusvsim requires.
    - Updated mgpu commit to eecd21f to pick up this change.
- `samplePTSBEWithLifecycle` now guarantees context teardown (finalize, reset, deallocate) even when the simulator throws, preventing "Context already set" errors on subsequent calls.
- Added `CUDAQ_PTSBE_FORCE_GENERIC` env var which when set, bypasses the BatchSimulator dispatch and forces the generic per-trajectory code path. This has proven useful as there are performance issues with the mgpu implementation
- Added PTSBE logging. This was helpful for initial debugging.

### DIFF
--- a/runtime/cudaq/ptsbe/PTSBESample.h
+++ b/runtime/cudaq/ptsbe/PTSBESample.h
@@ -206,7 +206,6 @@ sample_result runSamplingPTSBE(KernelFunctor &&wrappedKernel,
   const auto &noiseModel = noisePtr ? *noisePtr : kEmptyNoiseModel;
 
   // Stage 0: Capture trace via ExecutionContext("tracer")
-  cudaq::info("[ptsbe] Capturing circuit trace for kernel '{}'", kernelName);
   ExecutionContext traceCtx("tracer");
   platform.with_execution_context(traceCtx, [&]() { wrappedKernel(); });
   cleanupTracerQubits(traceCtx.kernelTrace);
@@ -230,7 +229,6 @@ sample_result runSamplingPTSBE(KernelFunctor &&wrappedKernel,
               batch.totalShots(), batch.trajectories.size());
 
   // Stage 4: Execute PTSBE with life-cycle management
-  cudaq::info("[ptsbe] Executing batched simulation");
   auto perTrajectoryResults = samplePTSBEWithLifecycle(batch);
 
   // Stage 5: Aggregate per-trajectory results
--- a/runtime/cudaq/ptsbe/PTSBESampler.cpp
+++ b/runtime/cudaq/ptsbe/PTSBESampler.cpp
@@ -166,6 +166,8 @@ template <typename ScalarType>
 std::vector<cudaq::sample_result>
 samplePTSBEGeneric(nvqir::CircuitSimulatorBase<ScalarType> &simulator,
                    const PTSBatch &batch) {
+  ScopedTraceWithContext("ptsbe::samplePTSBEGeneric",
+                         batch.trajectories.size());
   if (!cudaq::getExecutionContext())
     throw std::runtime_error(
         "samplePTSBEGeneric requires ExecutionContext to be set. "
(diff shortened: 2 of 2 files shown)


runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.cpp
```diff
@@ -40,57 +40,58 @@ ConditionalSamplingStrategy::generateTrajectories(
...
+      probability *= noise_points[i].channel.probabilities[idx];
     }
 
-    if (seen_patterns.insert(pattern).second) {
+    auto it = pattern_to_index.find(pattern);
+    if (it != pattern_to_index.end()) {
+      results[it->second].multiplicity++;
+    } else if (rejected_patterns.contains(pattern)) {
+      // Already tested and failed predicate; skip.
+    } else {
```
1tnguyen: nit: Could we combine this empty `else if` with the below `else`, e.g.,  using `else if (!rejected_patterns.contains(pattern))` instead?

1tnguyen (approved):
LGTM 💯

runtime/cudaq/ptsbe/strategies/ConditionalSamplingStrategy.cpp
taalexander: Thanks, that is some slop I didn't catch...

runtime/common/NoiseModel.h
```diff
@@ -334,6 +334,13 @@ class kraus_channel {
     generateUnitaryParameters_fp64(ops, this->unitary_ops, this->probabilities);
   }
 
+  /// @brief Check whether the operator at the given index is an identity.
+  bool is_identity_op(std::size_t index) const {
+    if (index >= op_names.size())
+      return false;
+    return op_names[index] == "id" || op_names[index] == "ii";
```
sacpis, line 341: Wondering if we should also cover "I" or "identity" or "iz"?

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
```diff
@@ -26,7 +26,9 @@ std::size_t hashKrausSelection(const cudaq::KrausSelection &sel) {
   std::size_t h = std::hash<std::size_t>{}(sel.circuit_location);
   for (std::size_t q : sel.qubits)
     hashCombine(h, std::hash<std::size_t>{}(q));
-  hashCombine(h, static_cast<std::size_t>(sel.kraus_operator_index));
+  hashCombine(h, std::hash<std::string>{}(sel.op_name));
+  hashCombine(h, sel.kraus_operator_index);
+  hashCombine(h, std::hash<bool>{}(sel.is_error));
```
sacpis, line 31: nit: seems like `is_error` is redundant here as it is derived from `kraus_op_inedx` and `channel` which are already hashed?

runtime/common/NoiseModel.h
taalexander: I could add `I` but `iz` would be interpreted as I x Z pauli.

runtime/cudaq/ptsbe/TrajectoryDeduplication.cpp
taalexander: I agree it is but I was adding for consistency going forward.

PR #4023, opened 2026-02-23
Introduces Pasqal CUDA-Q QRMI integration

awennersteen:
## Summary

  This MR introduces CUDA-Q a connection for Pasqal to the Quantum Resource Management Interface.
  It re-uses the existing pasqal target, enabling use of qrmi when that is available on a supported cluster, and unifies the two targets into a common `PasqalExecutor`.

  - QRMI repo: https://github.com/qiskit-community/qrmi
In QRMI this PR implements some QRMI side features  for this PR https://github.com/qiskit-community/qrmi/pull/55. It is needed because we use a different endpoint for cuda-q than for our own Pulser job format.

This PR (in cuda-q) also needs to switch to using libqrmi.so directly and not building QRMI. That also needs enabling a quick artifact publishing step on QRMI side.

Intended to be used together with the SPANK plugins integrating it into a Slurm cluster.
- QRMI Spank plugins for Slurm integration: https://github.com/qiskit-community/spank-plugins

Checklist:
- [x] I have added tests to cover my changes.
- [x] I have updated the documentation accordingly.


sacpis:
> /ok to test e44c21e

sacpis:
> /ok to test a8c7424

docs/sphinx/using/backends/hardware/neutralatom.rst
```diff
@@ -180,6 +180,20 @@ Alternatively, users can set the following environment variables directly.
   export PASQAL_AUTH_TOKEN=<>
   export PASQAL_PROJECT_ID=<>
 
+Pasqal via QRMI
+```````````````
+
+CUDA-Q also provides a ``pasqal`` target for routing Pasqal
+jobs through the vendor agnostic Quantum Resource Management Interface (QRMI).
+This target enables integration with resource managers like Slurm for scheduling.
+Select Pasqal backend with the ``--qpu`` ``sbatch`` option let QRMI handle submission.
```
sacpis: nit: option and let

docs/sphinx/using/backends/hardware/neutralatom.rst
```diff
@@ -180,6 +180,20 @@ Alternatively, users can set the following environment variables directly.
...
+
+
+For this route, credentials and project id are read by QRMI using either
+``~/.pasqal/config``or other methods supported by your clusters QRMI setup.
```
sacpis: nit: ``~/.pasqal/config`` or

docs/sphinx/using/examples/hardware_providers.rst
```diff
@@ -161,6 +161,7 @@ Pasqal
 ==================================
 
 The following code illustrates how to run kernels on Pasqal's backends.
+For QRMI-routed Pasqal jobs, specify `pasqal` as the target, the `machine` argument will be supplied by QRMI at runtime.
```
sacpis: nit: ``pasqal``

runtime/cudaq/platform/pasqal/PasqalUtils.h
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  return ExecutionResult(counts);
+}
+
+inline ExecutionResult
+parseExecutionResultFromTaskResult(const std::string &taskResultJson) {
+  auto payload = nlohmann::json::parse(taskResultJson);
+
+  // Fix this outside CUDA-Q. Why are these payloads different?
```
sacpis: Does it need attention?

docs/sphinx/using/backends/hardware/neutralatom.rst
awennersteen: done

docs/sphinx/using/backends/hardware/neutralatom.rst
awennersteen: done

docs/sphinx/using/examples/hardware_providers.rst
awennersteen: with double `` you mean?
If so, done

runtime/cudaq/platform/pasqal/PasqalUtils.h
awennersteen: Fixed, thanks

awennersteen:
I pushed a change that forces the QRMI to be off for now.

This is a useful test to verify this works. As the old path should remain functional with its associated tests. 

- Locally, this build succeeded and the existing tests work, as well as manual end-to-end
- They also work building with QRMI enabled, along with manual end-to-end with and without QRMI
    - (i.e. I can still use the existing direct CUDA-Q -> Pasqal Cloud integration)

khalatepradnya:
> /ok to test 3531112

CMakeLists.txt
```diff
@@ -103,6 +103,18 @@ if (NOT DEFINED CUDAQ_ENABLE_PASQAL_BACKEND)
...
+  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+   set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR ON CACHE BOOL "Enable building the Pasqal QRMI connector.")
+ else()
+   set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR OFF CACHE BOOL "Enable building the Pasqal QRMI connector.")
+ endif()
+
+  # Keep the Pasqal QRMI connector opt-in until `libqrmi.so` is distributed.
+  set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR OFF CACHE BOOL "Enable building the Pasqal QRMI connector." FORCE)
+endif()
+
```
khalatepradnya, line 117: Can you help me understand the intent of this? 
Do we want to set it on only on Linux if build variable is present? Something like -
```
if (NOT DEFINED CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR)
  set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR OFF CACHE BOOL "Enable building the Pasqal QRMI connector.")
elseif (CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR AND NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
  message(WARNING "CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR=ON is supported only on Linux; forcing OFF.")
  set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR OFF CACHE BOOL "Enable building the Pasqal QRMI connector." FORCE)
endif()
```

awennersteen: the desired behaviour in the future would be:
* Build with QRMI by default on Linux, but enable turning off
* Never build with QRMI on Non-Linux (It's not supported upstream in the overall project)
* Right now, we only support building from source - to change tomorrow hopefully. So right now I force unset variables to mean not install, to leave a single line change to re-enable build.

awennersteen: But your proposal on
```
CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR AND NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
makes sense, and I see that my line 111 should say "Disable building the QRMI..." and not "enable".

khalatepradnya: Thanks for the explanation!
You reckon by tomorrow the `libqrmi.so` library will be available? So we can simply all the build logic... See `AWSSDK_ROOT` in CUDA-Q repo for example.

runtime/cudaq/platform/pasqal/CMakeLists.txt
```diff
@@ -7,33 +7,147 @@
...
+  )
+endforeach()
     
 install(TARGETS ${LIBRARY_NAME} DESTINATION lib)
+install(TARGETS ${SERVERHELPER_LIBRARY_NAME} DESTINATION lib)
 
 add_target_config(pasqal)
+
+if(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  macro(_pasqal_download_or_fail url out_path label)
```
khalatepradnya: Do we want to make this a pre-requisite? 
See https://github.com/NVIDIA/cuda-quantum/pull/2412 for potential idea.

runtime/cudaq/platform/pasqal/QrmiUtils.h
```diff
@@ -0,0 +1,176 @@
...
+      std::chrono::seconds pollInterval = std::chrono::seconds(1)) {
+    QrmiTaskStatus status = QRMI_TASK_STATUS_QUEUED;
+    while (true) {
+      ensureSuccess(
+          qrmi_resource_task_status(resource_.get(), taskId.c_str(), &status),
+          "qrmi_resource_task_status()");
+      if (!taskInProgress(status))
+        return status;
+      std::this_thread::sleep_for(pollInterval);
+    }
```
khalatepradnya, line 122: Should this have a timeout parameter?

runtime/cudaq/platform/pasqal/QrmiUtils.h
```diff
@@ -0,0 +1,176 @@
...
+  if (rc != QRMI_RETURN_CODE_SUCCESS)
+    throw std::runtime_error(std::string(context) + ": " + lastError());
+}
+
+inline bool taskInProgress(QrmiTaskStatus status);
+inline std::string taskStatusAsString(QrmiTaskStatus status);
+
+class ResourceSession {
+public:
+  ResourceSession( // TODO: I think resource type should not be needed by
```
khalatepradnya, line 57: Do we expect this to be handled / cleaned up soon?

runtime/cudaq/platform/pasqal/PasqalQrmiServerHelper.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+std::vector<ServerMessage>
+PasqalQrmiServerHelper::createJobs(std::vector<KernelExecution> &circuitCodes,
+                                   std::size_t shots,
+                                   const std::string &backendName) {
+  std::vector<ServerMessage> tasks;
+  for (auto &circuitCode : circuitCodes) {
+    ServerMessage message;
+    message["machine"] = backendName;
+    message["job_runs"] = static_cast<std::int32_t>(shots);
```
sacpis: Wondering as shots is `std::size_t`, would casting it to `int32_t` without a range check cause an issue?

runtime/cudaq/platform/pasqal/QrmiUtils.h
awennersteen: I've added a timeout similar to runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp, but didn't include the watchdog thread from there right now.

runtime/cudaq/platform/pasqal/PasqalQrmiServerHelper.cpp
awennersteen: range check added to avoid thinking about it.

runtime/cudaq/platform/pasqal/QrmiUtils.h
awennersteen: removed TODO, dealing with it in the appropriate repo. 
It would need investigation and some re-design so it's not imminent, and it may be that the current design is correct.

runtime/cudaq/platform/pasqal/CMakeLists.txt
awennersteen: EDIT: Scratch the below, I think I understand better now.  let me do it


It's a small library, and in Rust. I think it's best to just fetch the pre-built `libqrmi.so` on the fly.
The AWS_SDK solution seems overkill for now.

Should QRMI, as a vendor agonstic project, gain sufficient adoption that multiple vendors will want to pull it in to CUDA-Q we should look at such options.

mitchdz: Yes please, +1 this idea. This makes auditing third party source installed code much easier.

CMakeLists.txt
awennersteen: The final release is not ready yet. more likely that it will be ready tomorrow at this point, unfortunately.

awennersteen:
last commit is bad - do not test.

Branch state ok, QRMI built very soon. Will ping for reviews once OK to test

awennersteen:
This should work, we've released QRMI as `libqrmi.so` and `qrmi.h`. Locally I have successfully built this code. so should be OK to tets

khalatepradnya:
> /ok to test 1dc70b9

.github/workflows/integration_tests.yml
```diff
@@ -632,11 +632,24 @@ jobs:
...
                   else
-                    echo ":x: Test failed (failed to execute): $filename" >> $GITHUB_STEP_SUMMARY
+                    echo ":x: Test failed (failed to execute): $filename (direct execution)" >> $GITHUB_STEP_SUMMARY
                     test_err_sum=$((test_err_sum+1))
                   fi
+                  if [[ "$filename" == *.py ]]; then
+                    PASQAL_MACHINE_TARGET=qrmi \
+                    SLURM_JOB_QPU_RESOURCES=EMU_FREE \
+                    EMU_FREE_QRMI_PASQAL_CLOUD_PROJECT_ID=$PASQAL_PROJECT_ID \
+                    python3 "$run_file" 1> /dev/null
```
khalatepradnya: What is `$run_file`?
```suggestion
                    python3 "$filename" 1> /dev/null
```

CMakeLists.txt
```diff
@@ -103,6 +103,15 @@ if (NOT DEFINED CUDAQ_ENABLE_PASQAL_BACKEND)
...
+# Enable Pasqal QRMI connector by default only on Linux builds.
+if (NOT DEFINED CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR)
+  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+   set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR ON CACHE BOOL "Enable building the Pasqal QRMI connector.")
+ else()
+   set(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR OFF CACHE BOOL "Disable building the Pasqal QRMI connector.")
+ endif()
+endif()
+
 # Enable Quantum Circuits, Inc. (QCI) target by default.
```
khalatepradnya, line 118: Also add in this file, around line ~221-
```
if(NOT QRMI_INSTALL_PREFIX AND CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR)
  set(QRMI_INSTALL_PREFIX "$ENV{QRMI_INSTALL_PREFIX}" CACHE PATH "Path to QRMI installation")
endif()
```
and remove from the inner CMake?

runtime/cudaq/platform/pasqal/CMakeLists.txt
```diff
@@ -7,33 +7,102 @@
...
-target_include_directories(${LIBRARY_NAME} PRIVATE .
-  PUBLIC 
-    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/runtime>
-    $<INSTALL_INTERFACE:include>
+add_library(${SERVERHELPER_LIBRARY_NAME}
+  SHARED
+    PasqalServerHelper.cpp
+    PasqalQrmiServerHelper.cpp
+    PasqalExecutor.cpp
 )
```
khalatepradnya, line 29: Optionally add the `PasqalQrmiServerHelper.cpp` file? Something like -

```suggestion
set(_serverhelper_sources
  PasqalServerHelper.cpp
  PasqalExecutor.cpp
)
if(CUDAQ_ENABLE_PASQAL_QRMI_CONNECTOR AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
  list(APPEND _serverhelper_sources PasqalQrmiServerHelper.cpp)
endif()

add_library(${SERVERHELPER_LIBRARY_NAME}
  SHARED
    ${_serverhelper_sources}
)
```

runtime/cudaq/platform/pasqal/CMakeLists.txt
```diff
@@ -7,33 +7,102 @@
...
+
+  find_path(QRMI_INCLUDE_DIR qrmi.h
+            HINTS ${_qrmi_include_hints})
+  find_library(QRMI_LIBRARY NAMES qrmi
+               HINTS ${_qrmi_library_hints})
+
+  if(NOT QRMI_INCLUDE_DIR OR NOT QRMI_LIBRARY)
+    message(FATAL_ERROR
+      "Pasqal QRMI connector requested but qrmi.h/libqrmi was not found. "
+      "Install QRMI and set QRMI_INSTALL_PREFIX=<prefix> if needed.")
```
khalatepradnya: Do we want to make this fatal error or allow to proceed with warning and degraded experience?

runtime/cudaq/platform/pasqal/CMakeLists.txt
```diff
@@ -7,33 +7,102 @@
...
+
+  message(STATUS "Enabling Pasqal QRMI mode support in pasqal target.")
+  target_include_directories(${SERVERHELPER_LIBRARY_NAME} PRIVATE ${QRMI_INCLUDE_DIR})
+  target_link_libraries(${SERVERHELPER_LIBRARY_NAME} PRIVATE ${QRMI_LIBRARY})
+  target_compile_definitions(${SERVERHELPER_LIBRARY_NAME}
+                             PRIVATE CUDAQ_PASQAL_QRMI_ENABLED)
+
+  get_filename_component(QRMI_LIBRARY_DIR "${QRMI_LIBRARY}" DIRECTORY)
+  set_target_properties(${SERVERHELPER_LIBRARY_NAME} PROPERTIES
+    BUILD_RPATH "${QRMI_LIBRARY_DIR}"
```
khalatepradnya: Don't we need to set `INSTALL_RPATH ` as well?

runtime/cudaq/platform/pasqal/PasqalExecutor.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+/*
+ * Unified Pasqal executor:
+ * - direct mode delegates to normal CUDA-Q REST Executor flow
+ * - machine=qrmi mode uses QRMI task lifecycle directly
+ *
+ * Why:
+ *
+ * The standard remote flow for backends like `pasqal` is implemented in the
+ * shared `runtime/common` Executor + Future stack. Tt uses RestClient to
```
khalatepradnya: nit: `Tt` --> `It`

runtime/cudaq/platform/pasqal/PasqalExecutor.cpp
```diff
@@ -0,0 +1,136 @@
...
+          // Today we only submit a single circuit per job.
+
+          for (auto &job : jobs) {
+            std::string sequence = job.at("sequence").dump();
+            auto jobRuns = job.at("job_runs").get<std::int32_t>();
+
+            QrmiPayload payload{};
+            payload.tag = QRMI_PAYLOAD_PASQAL_CLOUD;
+            payload.PASQAL_CLOUD.sequence =
+                const_cast<char *>(sequence.c_str());
```
khalatepradnya: Is this safe? May be we should use `std::vector<char>` instead and use it's `data`?

runtime/cudaq/platform/pasqal/QrmiUtils.h
```diff
@@ -0,0 +1,188 @@
...
+          "qrmi_resource_task_start() returned a null task id.");
+
+    StringPtr taskId(taskIdRaw);
+    return std::string(taskId.get());
+  }
+
+  QrmiTaskStatus waitForTaskTerminalStatus(
+      const std::string &taskId,
+      std::chrono::seconds pollInterval = std::chrono::seconds(1),
+      std::chrono::seconds timeout = std::chrono::seconds(60 * 60 * 24 * 30)) {
```
khalatepradnya, line 103: This is 30 days? Want to confirm that's the intended behavior...

scripts/install_prerequisites.sh
```diff
@@ -446,6 +446,35 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
...
+    tar -xzf "${qrmi_archive}"
+    cp "${qrmi_unpack_dir}/qrmi.h" "$qrmi_header"
+    cp "${qrmi_unpack_dir}/libqrmi.so" "$qrmi_library"
+
+    popd
+    remove_temp_installs
+  else
+    echo "QRMI already installed in $QRMI_INSTALL_PREFIX."
+  fi
+fi
```
khalatepradnya, line 498: We should add QRMI to the `NOTICE` file 
(https://github.com/NVIDIA/cuda-quantum/blob/main/NOTICE)
@mitchdz for viz.

scripts/install_prerequisites.sh
```diff
@@ -446,6 +446,35 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
   fi
 fi
 
+# [QRMI] Needed for the Pasqal QRMI connector
+if [ -n "$QRMI_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep qrmi)" ]; then
```
khalatepradnya: Check here for Linux? `[ "$(uname)" = "Linux" ]`

runtime/cudaq/platform/pasqal/QrmiUtils.h
awennersteen: Same default as used by the CUDA-Q rest server.

https://github.com/NVIDIA/cuda-quantum/blob/main/runtime/cudaq/platform/default/rest_server/helpers/RestRemoteServer.cpp#L599

It could be shorter, but to avoid timeout when calling cloud resources (it is supported), it will have to be in the order of days.

runtime/cudaq/platform/pasqal/PasqalExecutor.cpp
awennersteen: fixed, thanks

scripts/install_prerequisites.sh
awennersteen: yes, good point. done

It was already guarded in the set_env_defaults.sh but let's add both

runtime/cudaq/platform/pasqal/PasqalExecutor.cpp
awennersteen: Replaced with
```cpp
std::vector<char> sequenceBuffer(sequence.begin(), sequence.end());
sequenceBuffer.push_back('\0');

QrmiPayload payload{};
payload.tag = QRMI_PAYLOAD_PASQAL_CLOUD;
payload.PASQAL_CLOUD.sequence = sequenceBuffer.data();
```

runtime/cudaq/platform/pasqal/CMakeLists.txt
awennersteen: Honestly, I'm not the best at CMake. Youre probably right.  It now says
```cmake
get_filename_component(QRMI_LIBRARY_DIR "${QRMI_LIBRARY}" DIRECTORY)
set_property(TARGET ${SERVERHELPER_LIBRARY_NAME} APPEND PROPERTY
        BUILD_RPATH "${QRMI_LIBRARY_DIR}")
set_property(TARGET ${SERVERHELPER_LIBRARY_NAME} APPEND PROPERTY
        INSTALL_RPATH "${QRMI_LIBRARY_DIR}")
```
also taking care to append to rather than replace

runtime/cudaq/platform/pasqal/CMakeLists.txt
awennersteen: thanks, good idea
```cmake
message(WARNING
      "Pasqal QRMI connector requested but qrmi.h/libqrmi was not found. "
      "Install QRMI and set QRMI_INSTALL_PREFIX=<prefix> if needed; "
      "building pasqal target without QRMI mode support.")
  else()
```

.github/workflows/integration_tests.yml
awennersteen: :facepalm: thanks

CMakeLists.txt
awennersteen: Done

scripts/install_prerequisites.sh
awennersteen: added
```
----------------------------------------------------------------

Quantum Resource Management Interface (QRMI) - Apache License 2.0
<https://github.com/qiskit-community/QRMI>


License at <https://github.com/qiskit-community/qrmi/blob/main/LICENSE.txt>

----------------------------------------------------------------
```

awennersteen: Corrected in followup to above

khalatepradnya:
> /ok to test 9434906

mitchdz:
> /ok to test 9434906

khalatepradnya:
> /ok to test d3928b0

khalatepradnya:
@awennersteen I see that now the installer builds are failing...

awennersteen:
Should be fixed @khalatepradnya 

Looks like my local build with GCC does not trigger the warning about covered default in switch statements... But the NVQ++ in the pipeline does...


I also added extra guards to avoid surprises in integration tests. We'll release a more widely built version by the next cuda-q release.

khalatepradnya:
> /ok to test 579ff58

runtime/cudaq/platform/pasqal/CMakeLists.txt
```diff
@@ -7,33 +7,108 @@
...
+      target_include_directories(${SERVERHELPER_LIBRARY_NAME} PRIVATE ${QRMI_INCLUDE_DIR})
+      target_link_libraries(${SERVERHELPER_LIBRARY_NAME} PRIVATE ${QRMI_LIBRARY})
+      target_compile_definitions(${SERVERHELPER_LIBRARY_NAME}
+                                PRIVATE CUDAQ_PASQAL_QRMI_ENABLED)
+
+      get_filename_component(QRMI_LIBRARY_DIR "${QRMI_LIBRARY}" DIRECTORY)
+      set_property(TARGET ${SERVERHELPER_LIBRARY_NAME} APPEND PROPERTY
+        BUILD_RPATH "${QRMI_LIBRARY_DIR}")
+      set_property(TARGET ${SERVERHELPER_LIBRARY_NAME} APPEND PROPERTY
+        INSTALL_RPATH "${QRMI_LIBRARY_DIR}")
```
khalatepradnya, line 111: Is this installer path relocatable?
@taalexander for guidance.

scripts/install_prerequisites.sh
```diff
@@ -446,6 +446,37 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
...
+    QRMI_RELEASE_REPO=${QRMI_RELEASE_REPO:-qiskit-community/qrmi}
+    QRMI_RELEASE_TAG=${QRMI_RELEASE_TAG:-v0.12.0}
+    QRMI_RELEASE_VERSION=${QRMI_RELEASE_TAG#v}
+    qrmi_release_base="https://github.com/${QRMI_RELEASE_REPO}/releases/download/${QRMI_RELEASE_TAG}"
+    qrmi_archive="libqrmi-${QRMI_RELEASE_VERSION}-el8-x86_64.tar.gz" # Note: el8 build works on Ubuntu 18.04+.
+    qrmi_unpack_dir="libqrmi-${QRMI_RELEASE_VERSION}"
+
+    mkdir -p "$QRMI_INSTALL_PREFIX/include" "$QRMI_INSTALL_PREFIX/lib64"
+    wget "${qrmi_release_base}/${qrmi_archive}" -O "${qrmi_archive}"
+    tar -xzf "${qrmi_archive}"
```
khalatepradnya, line 486: We should have some verification here.
@mitchdz for guidance.

runtime/cudaq/platform/pasqal/CMakeLists.txt
awennersteen: Is this related to the build failure that there can not be gcc-built code in the nvq++binaries?
at this point we're beyond my capabilities. 

A backup solution that to us is acceptable is to not have qrmi in the built cuda-q.
We're expecting customers to build cuda-q from scratch on their HPC center anyways, so having it available for custom builds is acceptable.

Unfortunately, QRMI at core is Rust, and building rust from NVQ++ is experimental at best.

I cannot make code changes today - but feel free to push this change if you have the time.


The origin for these lines is that in the Python code, there is a (I believe) unnecessary hardcoded check that the target "<target>-serverhelper*.so" is locatable. Despite in our implementation we use a custom executor and not the rest api's serverhelper.

scripts/install_prerequisites.sh
awennersteen: A backup solution that to us is acceptable is to not have qrmi in the built cuda-q.
We're expecting customers to build cuda-q from scratch on their HPC center anyways, so having it available for custom builds is acceptable.

Unfortunately, QRMI at core is Rust, and it's probably best to not introduce the Rust compiler chain into CUDA-Q...

I cannot make code changes today - but feel free to push this change if you have the time.

runtime/cudaq/platform/pasqal/CMakeLists.txt
taalexander: I agree that the current `INSTALL_RPATH` is not relocatable with the installer as it's hardcoding the absolute path from the build machine. This on it's own is resolvable (install `libqrmi` into the `lib/` folder during install and than use `$ORIGIN/@loader_path` as the rpath). However, I think @awennersteen has highlighted a larger issue that QRMI is pulling in GCC runtime dependencies.

Keeping in mind I'm quite new to the project and don't know how all of the plugins are being built today. The options that I can see are:
1. Do what Aleksander suggests and don't include by default with the installer only including with optional source builds?
2. Load the dynamic library at runtime so that they are not linked into nvq++ at build-time (a decent amount of work)

taalexander: Doing a bit of investigating for option 2. the server helpers (including pasqal's) are already being loaded dynamically https://github.com/NVIDIA/cuda-quantum/blob/54972bea26b967dffc1684c2461eddcd57c69f1f/python/utils/LinkedLibraryHolder.cpp#L452-L454 so there is some precedent for doing this for plugins.

I think maybe doing the dynamic load in `runtime/cudaq/platform/pasqal/QrmiUtils.h` would be relatively drop-in (combined with packaging fixes). That being said I think option 1. is most expedient.

awennersteen: Id be up for the dynamic load. But derisking and giving some breathing room by initially going for option 1 makes sense to me.

khalatepradnya: Sounds good!
Let's go for source-only (optional with flag) build in this PR. I will make the necessary changes.

scripts/install_prerequisites.sh
khalatepradnya: Added checksum.

khalatepradnya:
> /ok to test c5c295b

khalatepradnya:
> /ok to test 4b540a8

runtime/cudaq/platform/pasqal/QrmiUtils.h
khalatepradnya: Not sure if this comment implies that the `TODO` should have been removed, but it's still there...

runtime/cudaq/platform/pasqal/pasqal.yml
```diff
@@ -12,7 +12,7 @@ config:
   # Tell DefaultQuantumPlatform what QPU subtype to use
   platform-qpu: pasqal
   # Add the pasqal-qpu library to the link list
-  link-libs: ["-lcudaq-pasqal-qpu"]
+  link-libs: ["-lcudaq-pasqal-qpu", "-lcudaq-serverhelper-pasqal"]
```
khalatepradnya, line 15: I had to add this without which the C++ workflow was broken.

khalatepradnya (approved):
I manually tested `pasqal.py` and `pasqal.cpp` on this branch, with and without QRMI installation. I was also able to run the newly added integration test. Note that this doesn't really test QRMI functionality but ensures that the APIs are working correctly.
These changes LGTM for source-only build.

khalatepradnya:
> /ok to test 0958183

bettinaheim:
Pulling this in, leaving any inclusion by default in the distributed packages as follow up as per the offline thread.

PR #4024, opened 2026-02-23
[python] Use pybind for kernel argument type casting

lmondada:
Previously, kernel arguments passed to `PyKernelDecorator` were passed as-is to the C++ bindings and then cast to the appropriate types within C++. This is hard to maintain as inspecting python objects from C++ is cumbersome, especially if we want to keep some of the python "duck typing" (e.g. handling both np.array and lists seamlessly).

This PR cleans this up as follows:
1. Define in Python a clear list of kernel argument types that are supported as

### DIFF
--- a/python/cudaq/kernel/kernel_args.py
+++ b/python/cudaq/kernel/kernel_args.py
@@ -1,0 +1,175 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.    #
+# ============================================================================ #
+"""
+Typed argument wrappers for CUDA-Q kernel arguments.
+
+Each wrapper represents a Python value that can be passed to a kernel and
+corresponds to a supported MLIR type. These types will be marshaled into
+OpaqueArguments by the C++ type caster in `python/utils/OpaqueArgumentsCaster.h`.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Union
+
+
+@dataclass(slots=True)
+class Float64Arg:
+    """A float value destined for an MLIR Float64Type parameter."""
+    value: float
+
+
+@dataclass(slots=True)
+class Float32Arg:
+    """A float value destined for an MLIR Float32Type parameter."""
+    value: float
+
+
+@dataclass(slots=True)
+class IntArg:
+    """An integer value destined for an MLIR IntegerType (width != 1)."""
+    value: int
+
+
+@dataclass(slots=True)
+class BoolArg:
+    """A boolean value destined for an MLIR IntegerType(1)."""
+    value: bool
+
+
+@dataclass(slots=True)
+class ComplexF64Arg:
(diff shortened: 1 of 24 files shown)


python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -166,15 +149,15 @@ class PyQPUState : public QPUState {
 /// @brief Run `cudaq::get_state` for qpu targets on the provided
 /// kernel and args
 state pyGetStateQPU(const std::string &kernelName, MlirModule kernelMod,
-                    py::args args) {
+                    OpaqueArguments args) {
   auto moduleOp = unwrap(kernelMod);
+  setDataLayout(kernelMod);
```
lmondada, line 154: `setDataLayout` was previously called in `toOpaqueArgs`, which was called from this function.

python/cudaq/kernel/kernel_args.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@dataclass(slots=True)
+class Float32Arg:
+    """A float value destined for an MLIR Float32Type parameter."""
+    value: float
+
+
+@dataclass(slots=True)
+class IntArg:
```
lmondada, line 35: Check the code: is there support for multiple integer widths?

python/utils/OpaqueArgumentsCaster.h
```diff
@@ -0,0 +1,185 @@
...
+        state *copyState = new state(*stateArg);
+        argData.emplace_back(
+            copyState, [](void *ptr) { delete static_cast<state *>(ptr); });
+      } else {
+        argData.emplace_back(stateArg, [](void *) {});
+      }
+
+    } else if (cls == "StructArg") {
+      auto mlirTy = obj.attr("mlir_type");
+      auto mlirMod = obj.attr("module");
```
lmondada, line 71: Why not tag the inner members too?

python/cudaq/kernel/kernel_args.py
```diff
@@ -0,0 +1,175 @@
...
+OpaqueArguments by the C++ type caster in `python/utils/OpaqueArgumentsCaster.h`.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Union
+
+
+@dataclass(slots=True)
```
bettinaheim, line 22: As briefly discussed offline, there is a mismatch here between how the outermost type is handled (wrapped/tagged with these dataclasses) vs how inner elements are handled (currently not tagged - composite types should be dealt with recursively). We can either get rid of the tags, or also make use of them for inner elements. I'll leave it up to your judgement on whether it is worth keeping the tags, but if you do, I think it deserves a bit more explanation in the docs comment of the file what these classes are for (and what they are not for).

lmondada:
This isn't ready to be merged. As mentioned in the comments, we should either move ALL the conversion logic to Python (including the nested arguments within `VecArg`s and `StructArg`s, or none at all.

This would also be an opportunity to consolidate all conversion logic in one place.

I am thinking the right way to do this would be to
1. have the various `KernelArg` types be C++ types with bindings into Python. The conversion then happens transparently and we never need to go grab python attributes from C++.
2. clearly distinguish between conversion (for free) and the logic that packs the arguments into `OpaqueArguments` into its own thing (factored out). Right now both are happening in one step, it feels a bit muddled and hard to read.

PR #4025, opened 2026-02-23
[macOS] Add macOS support for publishing workflow (wheels/installer)

taalexander:
### Description
This PR adds support for macOS to the publishing workflow for wheels/installer. As we do not have any internal deployments to run this workflow calls the `ci_macos.yml` workflow to prepare artifacts (instead of doing this through deployments). Then fetches these artifacts from the GH api.

### DIFF
--- a/scripts/validate_pycudaq.sh
+++ b/scripts/validate_pycudaq.sh
@@ -236,8 +236,8 @@ if $is_macos; then
                 echo "Installing wheel: $wheel_file"
                 pip install --force-reinstall "$wheel_file"
             else
-                echo "No wheel found in ${extra_packages}, installing from PyPI"
-                pip install --upgrade cudaq==${cudaq_version}
+                echo -e "\e[01;31mNo wheel or metapackage found in ${extra_packages}. Refusing to install from PyPI when -i is set.\e[0m" >&2
+                (return 0 2>/dev/null) && return 100 || exit 100
             fi
         fi
     else


.github/workflows/publishing.yml
```diff
@@ -1227,9 +1301,82 @@ jobs:
             exit $status_sum
           fi
 
+  metapackage_validation_macos:
```
bettinaheim, line 1304: nit: Not for this PR, but at some point it may be nice to unify this a bit more with the jobs for the other platforms to reduce code duplication.

scripts/validate_pycudaq.sh
```diff
@@ -221,15 +221,24 @@ if $is_macos; then
...
         else
-            echo "No wheel found in ${extra_packages}, installing from PyPI"
-            pip install --upgrade cudaq==${cudaq_version}
+            wheel_file=$(ls "${extra_packages}"/cuda_quantum*.whl 2>/dev/null | head -1)
+            if [ -n "$wheel_file" ]; then
+                echo "Installing wheel: $wheel_file"
+                pip install --force-reinstall "$wheel_file"
+            else
+                echo "No wheel found in ${extra_packages}, installing from PyPI"
+                pip install --upgrade cudaq==${cudaq_version}
```
bettinaheim: What was the reason for adding the fallback to install from pip here? 
This script is used by QA, and it seems like this might just open the possibility that we accidentally validate something else than what we think we validate.

bettinaheim:
@taalexander Forgot one thing: Can you please make sure that the MacOS wheels and installer are included in the artifacts of the draft release that is created at the end of the publishing workflow?

scripts/validate_pycudaq.sh
taalexander: This is a carryover from main https://github.com/taalexander/cuda-quantum/blob/505b69bd0831580c1a7e70ac976c226b770da187/scripts/validate_pycudaq.sh#L238-L241. 

Now why is it on main? I'm not sure. I've removed the fallback when `-i`.

.github/workflows/publishing.yml
taalexander: Agreed!

PR #4031, opened 2026-02-23
Support creating qubit lists in Python list comprehension

### LINKED ISSUE #4009: list comprehension with x.ctrl for multi-control
### Required prerequisites

### Describe the feature

These ways in the example for multi-control works:

```
import cudaq

cudaq.set_target('nvidia')

@cudaq.kernel
def third_example(N:int, theta:list[float]):
    qubit=cudaq.qvector(N)

    h(qubit)

    for i in range(0,N//2):
        ry(theta[i],qubit[i])
    
    x.ctrl([qubit[0],qubit[1]],qubit[2]) #ccx
    x.ctrl([qubit[0],qubit[1],qubit[2]],qubit[3]) #cccx
    x.ctrl(qubit[0:3],qubit[3]) #cccx using Python slicing syntax
    
    controls = qubit.front(N - 1)
    target = qubit.back()
    x.ctrl(controls, target)
    
    #cont=[0,1,2]
    #x.ctrl([qubit[i] for i in cont], qubit[3]) #cccx using list comprehension

    mz(qubit)

params=[0.15,1.5]

print(cudaq.draw(third_example, 4, params))
```

However, if I uncomment
```
#cont=[0,1,2]
#x.ctrl([qubit[i] for i in cont], qubit[3]) #cccx using list comprehension
```
and using list comprehension, it gives error. It appears that list comprehension in unallowed.

sacpis:
Extending the bridge to support qubit-ref list comprehensions (e.g. x.ctrl([qs[i] for i in indices], target)) for all iterable forms:

- [qs[i] for i in [0, 1, 2]]
- [qs[i] for i in range(3)]
- cont = [0, 1, 2]; [qs[i] for i in cont]
- def kernel(qs, indices: list[int])

Fixes #4009.

```
cont=[0,1,2]
x.ctrl([qubit[i] for i in cont], qubit[3]) #cccx
```

### DIFF
--- a/python/tests/mlir/ast_list_comprehension.py
+++ b/python/tests/mlir/ast_list_comprehension.py
@@ -1110,32 +1110,26 @@ def test_list_comprehension_qubit_refs():
 # CHECK: return
 # CHECK-LABEL: func.func @__nvqpp__mlirgen__kernel2..
 # CHECK: quake.extract_ref
-# CHECK: quake.concat
 # CHECK: cc.loop
 # CHECK: quake.concat
 # CHECK: return
 # CHECK-LABEL: func.func @__nvqpp__mlirgen__kernel3..
 # CHECK: quake.extract_ref
-# CHECK: quake.concat
 # CHECK: cc.loop
 # CHECK: quake.concat
 # CHECK: return
 # CHECK-LABEL: func.func @__nvqpp__mlirgen__kernel4..
 # CHECK: quake.extract_ref
-# CHECK: quake.concat
 # CHECK: cc.loop
 # CHECK: quake.concat
 # CHECK: return
 # CHECK-LABEL: func.func @__nvqpp__mlirgen__kernel5..
 # CHECK: quake.extract_ref
-# CHECK: quake.concat
 # CHECK: cc.loop
 # CHECK: quake.concat
 # CHECK: return
 # CHECK-LABEL: func.func @__nvqpp__mlirgen__kernel6..
 # CHECK: cc.stdvec_size
-# CHECK: quake.extract_ref
-# CHECK: quake.concat
 # CHECK: cc.loop
 # CHECK: quake.concat
 # CHECK: return


python/cudaq/kernel/ast_bridge.py
```diff
@@ -426,6 +426,7 @@ def node_error(msg):
         self.isSubscriptRoot = False
         self.verbose = verbose
         self.currentNode = None
+        self.pyConstantVars = {}
```
schweitzpgi: Can't one write recursive list comprehensions? This single variable is held for the entirety of the bridge's lifetime.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4033,6 +4063,35 @@ def get_item_type(pyval):
         if listElemTy is None:
             return
 
+        if quake.RefType.isinstance(listElemTy):
```
schweitzpgi, line 4037: It seems like this should generate IR to walk a vector of values (which may be unavailable until runtime) and then use those values to select qubit references and concat them into a veq aggregate.

python/cudaq/kernel/ast_bridge.py
schweitzpgi: Aren't list comprehensions allowed to be recursive and dynamic? I think this misses those possibilities as we have this 1 attribute for the entire bridge that's apparently tracking constants (not vars) and not Quake values.

python/cudaq/kernel/ast_bridge.py
sacpis: Yes. you are right. The problem here is that quake.ConcatOp takes a fixed, statically known list of operands. There is no dynamic concat in the quake dialect (please correct me if this is not the case). 

Building a runtime-indexed veq from scattered refs would require a new quake op. Or we can use the unroll trick that only works when the trip count is a compile-time constant. And we are using it here.

python/cudaq/kernel/ast_bridge.py
sacpis: Makes sense to allow list comprehension to be recursive and dynamic. I was looking at Marwa's case only. But will try to extend it.

sacpis: Thanks Eric. Addressed in [3e79f90](https://github.com/NVIDIA/cuda-quantum/pull/4031/commits/3e79f900436f2bda6e9d487dd793cd66117bde14).

python/cudaq/kernel/ast_bridge.py
sacpis: @1tnguyen pointed me to the recursion (`veq = Concat(veq, ref)`). Thank you @1tnguyen. Addressed in [3e79f90](https://github.com/NVIDIA/cuda-quantum/pull/4031/commits/3e79f900436f2bda6e9d487dd793cd66117bde14).

python/cudaq/kernel/ast_bridge.py
sacpis: Addressed in [3e79f90](https://github.com/NVIDIA/cuda-quantum/pull/4031/commits/3e79f900436f2bda6e9d487dd793cd66117bde14).

python/cudaq/kernel/ast_bridge.py
```diff
@@ -4033,6 +4034,55 @@ def get_item_type(pyval):
...
+                    cc.PointerType.get(iterTy), iterable, [c0],
+                    DenseI32ArrayAttr.get([kDynamicPtrIndex], context=self.ctx))
+                first_idx = cc.LoadOp(first_addr).result
+                self.symbolTable.beginBlock()
+                self.__deconstructAssignment(node.generators[0].target,
+                                             first_idx)
+                self.visit(node.elt)
+                first_ref = self.popValue()
+                self.symbolTable.endBlock()
+                init_veq = quake.ConcatOp(veqTy, [first_ref]).result
```
1tnguyen: Q: Can `ConcatOp` handle an empty list (i.e., create an empty `Veq` that we then append to)?
That'd be safer, as we can handle the case where an empty list is provided.

sacpis: ConcatOp with zero operands is erased by lowering. We can't use it as an empty initial accumulator. Alternative is to use `quake.AllocaOp` with initial size 0, then loop through all elements.

sacpis: Addressed in [52c5879](https://github.com/NVIDIA/cuda-quantum/pull/4031/commits/52c58791aa0067410344a21080d6b4211d7170d7).

python/tests/mlir/ast_list_comprehension.py
khalatepradnya, line 1163: Q: Why is this invalid?

python/tests/mlir/ast_list_comprehension.py
```diff
@@ -1027,6 +1027,114 @@ def kernel7(adj: list[bool]):
...
+    # keep after assert, such that we have no output if assert fails
+    print(kernel1)
+
+    @cudaq.kernel
+    def kernel2():
+        qs = cudaq.qvector(2)
+        x(qs[0])
+        x(qs[1])
+        target = cudaq.qubit()
+        x.ctrl([qs[i] for i in [0, 1]], target)
```
khalatepradnya, line 1051: Q: Do we also support condition inside the expression like - `x.ctrl([q[i] for i in qs if i%2 == 0], target)`

sacpis: Is there a gate where we set the control on the even (or odd or conditional) indices?

sacpis: Got it.

sacpis: Right now, we don't support filter. Would you mind filing an issue in case we need to support it?

python/tests/mlir/ast_list_comprehension.py
sacpis: This is not a part of this PR. Will clean up in a separate PR. Mind filing an issue?

sacpis: Just to answer you original question, we have implemented this feature for qubit-ref (quake.RefType).  In your example, mz(...) is an `ast.Call` to a measurement gate, which takes measurement path.

PR #4037, opened 2026-02-24
Fix bug where cleanup code was cleaned up

atgeller:
### Description

Looks like, during cleanup, the pybind bound function `cudaq-runtime.delete_cache_execution_engine` is deleted (and thus set to `None`) before it is called when cleaning up some decorators. I just added a check to not perform clean up in this case.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -551,7 +551,7 @@ class PyKernelDecorator(object):
         """
         # Make sure this hasn't already been cleaned up as we're winding down
         if (cudaq_runtime is not None and
-            cudaq_runtime.delete_cache_execution_engine is not None):
+                cudaq_runtime.delete_cache_execution_engine is not None):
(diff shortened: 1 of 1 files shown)


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -549,7 +549,9 @@ def delete_cache_execution_engine(self, key):
         """
         Delete the `ExecutionEngine` cache given by a cache key.
         """
-        cudaq_runtime.delete_cache_execution_engine(key)
+        # Make sure this hasn't already been cleaned up as we're winding down
+        if cudaq_runtime.delete_cache_execution_engine is not None:
```
sacpis: nit: `cudaq_runtime is not None and ...`

atgeller: Fixed

sacpis: Seems like the fix is not pushed yet?

schweitzpgi (approved):

PR #4039, opened 2026-02-24
Fix invalid instruction issues

atgeller:
### Description

This fixes two issues causing QIR validation errors.
1. The `noiseModel` used to be improperly ignored, which was good because it didn't work with emulated targets, but it became used properly, which caused issues. Now it is ignored intentionally and explicitly (for the short term).
2. There was a typo in `optimizeLLVM` which led to incorrect LLVM being generated.


atgeller:
Thanks @schweitzpgi and @Renaud-K for the help tracking down these errors!

PR #3957, opened 2026-02-14
[testing] Fix run tests for QIR adaptive profile 

khalatepradnya:
* Apply the workaround for issue https://github.com/NVIDIA/cuda-quantum/issues/3708.
* Re-enable tests on `quantinuum` and `qci` backends.

Issue #3708, opened 2025-12-30
[qir-adaptive] Invalid instruction `alloca`


khalatepradnya:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

A kernel which has conditional on measurement result and returns that value (or composite of it) produces an `alloca` instruction which is invalid QIR.

### Steps to reproduce the bug

Following kernel works with simulator (`qir-full`), but fails with targets emitting `qir-adaptive`.

```
#include <cudaq.h>
#include <iostream>

struct kernel_with_conditional {
  std::vector<bool> operator()() __qpu__ {
    cudaq::qubit q0, q1;
    auto m0 = mz(q0);
    if (m0) {
      x(q1);
    }
    return {m0, mz(q1)};
  }
};

int main() {
  auto results = cudaq::run(10, kernel_with_conditional{});
  std::cout << "First result: " << results[0][0] << " " << results[0][1]
            << "\n";
  return 0;
}
```

```
$ nvq++ ./min_repro.cpp && ./a.out
First result: 0 0

$ nvq++ --target qci --emulate ./min_repro.cpp && ./a.out
QIR verification error - invalid instruction found:   %2 = alloca i16, align 2 (adaptive profile)
terminate called after throwing an instance of 'std::runtime_error'
  what():  Could not successfully translate to qir-adaptive:1.0:int_computations,float_computations,output_log.
Aborted (core dumped)
```

Note that if we just return single measurement result (`mz(q1)`) then it works.
```
struct kernel_with_conditional {
  bool operator()() __qpu__ {
    cudaq::qubit q0, q1;
    auto m0 = mz(q0);
    if (m0) {
      x(q1);
    }
    return mz(q1);
  }
};
```

### Expected behavior

Kernel compiles and runs successfully.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

[...]

PR #4048, opened 2026-02-26
[testing] Apply the work-around for issue#3708 to an example in docs

khalatepradnya:
* Follow-up to PR https://github.com/NVIDIA/cuda-quantum/pull/3957
* Same work-around for issue https://github.com/NVIDIA/cuda-quantum/issues/3708 should be applied to this example
* Note that the probabilities depend on the shots count, higher is better, but to keep the test fast, we limit the number of shots

### DIFF
--- a/docs/sphinx/targets/cpp/qci.cpp
+++ b/docs/sphinx/targets/cpp/qci.cpp
@@ -67,9 +67,9 @@ int main() {
     return static_cast<double>(ones[q]) / num_shots;
   };
 
-  // mz[0] and mz[1] are Bell measurement outcomes, so each is ~50%.
-  // mz[2] is the teleported qubit. Its P(1) is determined by the
-  // prepared state: P(mz[2]=1) ~4.6% for the angles above.
+  // `mz[0]` and `mz[1]` are Bell measurement outcomes, so each is ~50%.
+  // Probability of measuring`mz[2]` in 1 is determined by the prepared state,
+  // which is ~4.6% for the angles above.
   std::cout << "Results over " << num_shots << " shots:\n";
   for (std::size_t q = 0; q < 3; ++q)
     std::cout << "  mz[" << q << "] = 1:  " << ones[q] << " / " << num_shots


docs/sphinx/targets/cpp/qci.cpp
```diff
@@ -34,25 +37,43 @@ struct teleportation {
...
+    for (std::size_t q = 0; q < 3; ++q)
+      ones[q] += static_cast<std::size_t>(shot[q]);
+
+  auto freq = [&](std::size_t q) {
+    return static_cast<double>(ones[q]) / num_shots;
+  };
+
+  // `mz[0]` and `mz[1]` are Bell measurement outcomes, so each is ~50%.
+  // Probability of measuring`mz[2]` in 1 is determined by the prepared state,
+  // which is ~4.6% for the angles above.
```
sacpis, line 72: Is this comment accurate? I am getting the following output

```
Results over 25 shots:
  mz[0] = 1:  16 / 25  (64%)
  mz[1] = 1:  14 / 25  (56%)
  mz[2] = 1:  0 / 25  (0%)
```

khalatepradnya: Good Q! One of the reasons why this test was left out when I worked on PR #3957 , I kept getting all 0s and thought it was wrong.
Since for 10 shots, on average you'll get one 0.46 times, 0 is the most likely outcome.
Can you try with a greater number of shots? Say 100 or 1000? There's a comment saying the probability is closer if you use more shots. We use small number of shots to keep the test fast.

sacpis: Yes, for more number of shots (1000), it shows mz[2] to be 4%.

```
Results over 1000 shots:
  mz[0] = 1:  473 / 1000  (47.3%)
  mz[1] = 1:  489 / 1000  (48.9%)
  mz[2] = 1:  40 / 1000  (4%)
```

and for 100 shots as well

```
Results over 100 shots:
  mz[0] = 1:  48 / 100  (48%)
  mz[1] = 1:  50 / 100  (50%)
  mz[2] = 1:  4 / 100  (4%)
```

PR #4052, opened 2026-02-26
[python] Unify specialize and launch in ModuleLauncher

lmondada:
This PR is the start of a line of work aiming to unify the "specialize kernel" and "launch kernel" execution paths for kernels defined (and executed from) Python.

I have started by creating a `CompiledKernel` class that can be shared by the two execution paths at the bottom of their respective call stacks (see diagram). As we progress, we should be able to progressively unify the two call stacks into one.

```mermaid
flowchart TB
    subgraph unify["This PR"]
        module["ModuleLauncher::compileModule"]
        module --> compiled[CompiledKernel]
    end
    qpu["QPU::specializeModule"]
    qpu2["QPU::launchModule"]
    platform["quantum_platform::specializeModule"]
    platform2["quantum_platform::launchModule"]
    strm["cudaq::streamlinedSpecializeModule"]
    strm2["cudaq::streamlinedLaunchModule"]
    pyLaunchModule ~~~ strm
    pyLaunchModule["cudaq::pyLaunchModule"]
    clm["cudaq::clean\_launch\_module"]
    marshal["marshal\_and\_retain\_module"]
    marshal2["marshal\_and\_launch\_module"]

    subgraph cls["PyKernelDecorator"]
        python["beta_reduction"]
        python2["\_\_call\_\_"]
    end

    python --> marshal --> strm --> platform --> qpu --> module
    python2 --> marshal2 --> clm --> pyLaunchModule -->strm2 --> platform2 --> qpu2 --> module
    
    platform2 --> otherqpu
    subgraph other[other QPUs]
        otherqpu[...]
    end
```

### DIFF
--- a/docs/sphinx/targets/python/quantum_machines_all_gates.py
+++ b/docs/sphinx/targets/python/quantum_machines_all_gates.py
@@ -1,4 +1,5 @@
 import cudaq
+import math
 
 # Test all gates on the quantum machines target.
 # The default executor is mock, use executor name to run on another backend (real or simulator).
--- a/docs/sphinx/using/applications.rst
+++ b/docs/sphinx/using/applications.rst
@@ -291,7 +291,7 @@ This page contains a number of different applications implemented using CUDA-Q. 
         <img src="../_static/app_title_images/quantum_pagerank_preview.png" alt="Quantum Pagerank Preview" class="notebook-image">
     </div> 
 
-    <div class="notebook-entry" data-tags="chemistry,noisless">
+    <div class="notebook-entry" data-tags="chemistry,noiseless">
         <a href="../applications/python/uccsd_wf_ansatz.html" class="notebook-title">UCCSD Wavefunction Ansatz</a>
         <div class="notebook-content">
             Learn how to implement the UCCSD wavefunction ansatz using CUDA-Q.
@@ -307,7 +307,7 @@ This page contains a number of different applications implemented using CUDA-Q. 
         <img src="../_static/app_title_images/mps_encoding.png" alt="MPS Encoding" class="notebook-image">
     </div>
 
-    <div class="notebook-entry" data-tags="chemistry,noisless">
+    <div class="notebook-entry" data-tags="chemistry,noiseless">
         <a href="../applications/python/qm_mm_pe.html" class="notebook-title">QM/MM simulation: VQE within a Polarizable Embedded Framework.</a>
         <div class="notebook-content">
             Learn how to implement QM/MM with PE framework using CUDA-Q.


runtime/common/JIT.cpp
```diff
@@ -365,4 +366,32 @@ void (*JitEngine::lookupRawNameOrFail(const std::string &kernelName) const)() {
   return impl->lookupRawNameOrFail(kernelName);
 }
 
+CompiledKernel::CompiledKernel(JitEngine engine, std::string kernelName,
+                               bool hasResult)
+    : engine(std::move(engine)), name(std::move(kernelName)),
```
Renaud-K: I would not move here. Value semantic is fine for a shared_ptr. The resource count goes up when the engine argument is passed by value and then down when the constructor returns.

runtime/common/JIT.cpp
```diff
@@ -7,6 +7,7 @@
  ******************************************************************************/
 
 #include "JIT.h"
+#include "ThunkInterface.h"
```
Renaud-K: This is also included in JIT.h. Would you consider a forward declaration there instead?

Renaud-K:
Thank you for simplify this part. Any consolidation/simplification of the code is a big plus.

runtime/common/JIT.cpp
lmondada: Good point, I've removed the `move`.

runtime/common/JIT.cpp
lmondada: Removed in the new version (see my comment below).

lmondada:
Thanks @Renaud-K for the early review!

I have settled for a design of `CompiledKernel` that does not depend on MLIR and can just live in its own file and get compiled in the `cudaq-common` library. The drawback of this is that we need to keep track of more data (e.g. the JITEngine AND the function pointer), but I think it is well worth it to remove the dependency.

To keep it easy to construct instances of this type, I've created a factory function that needs MLIR and lives in `JIT.cpp`. I'm making the class constructor private to make it impossible to create invalid class instances. The only way to create CompiledKernel instances is thus through the factory, which is a friend of the class.

My hope is that over time we can start using this class in other parts of the runtime, and merge execution paths between C++ and Python.

runtime/common/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(COMMON_RUNTIME_SRC
   SampleResult.cpp
   ServerHelper.cpp
   Trace.cpp
+  CompiledKernel.cpp
```
Renaud-K, line 29: Why would CopimledKernel be in cudaq-common? cudaq-mlir-runtime depends on cudaq-common and now cuda-common depends on cudaq-mlir-runtime. cudaq-common is a low level library which so far add little dependency on anything else. It should stay that way.

runtime/common/CompiledKernel.h
```diff
@@ -0,0 +1,64 @@
...
+#include <vector>
+
+namespace cudaq {
+
+class JitEngine;
+
+/// A unique_ptr with a plain function pointer as destructor, allowing
+/// type-erased ownership of forward-declared (incomplete) types.
+template <typename T>
+using OpaquePtr = std::unique_ptr<T, void (*)(T *)>;
```
Renaud-K: A unique pointer and an opaque pointer are 2 very different things. A unique pointer carries ownership information. An opaque pointer just reduces compile-time dependencies but not link time.
JitEngine is already opaque that sense. I don't see what it brings to wrap it more.

lmondada: So what I want is:
- ownership
- removing the dependency on "JIT.h", as that comes with an MLIR dependency (this is what I mean by "opaque")

I cannot use a "normal" `std::unique_ptr` as it does not work with the forward-declared `JITEngine` type (it must link in the destructor). This modified `std::unique_ptr` type erases the destructor, and thus it works. I could also use a `shared_ptr` or a custom opaque type, if you preferred. If I used a raw pointer or reference, I would not be able to destroy the JITEngine at destruction time.

What would you suggest I use?

runtime/common/CompiledKernel.h
```diff
@@ -0,0 +1,64 @@
...
+private:
+  CompiledKernel(OpaquePtr<JitEngine> engine, std::string kernelName,
+                 void (*entryPoint)(), bool hasResult);
+
+  // Use the following factory function in JIT.h to construct CompiledKernels.
+  friend CompiledKernel createCompiledKernel(JitEngine engine,
+                                             std::string kernelName,
+                                             bool hasResult);
+
+  OpaquePtr<JitEngine> engine;
```
Renaud-K: JitEngine uses reference counting. It is meant to be stored by value. The lifetime of the pointer is not guaranteed.

lmondada: See above: I want to avoid storing `JITEngine` by value because that would add a dependency on "JIT.h" and thus on MLIR.

The type name "OpaquePtr"' might be what's confusing. This is an owning type and thus the lifetime of the `JitEngine` is guaranteed to outlive CompiledKernel.

Maybe a better alternative would be to move JitEngine out of `cudaq-mlir-runtime`? In that case we could indeed just store it by value.

runtime/common/CompiledKernel.h
Renaud-K: Removing the header dependency only removes the compile-time dependency not the link time dependency. JIT.h has already been reduced to opaque pointers and erased types. It does not include any MLIR stuff. shared_ptr unlike unique_ptr does not require a destructor and JITEngine already uses a shared_ptr .  JITEngine is meant to be used with value semantic. It already brings the isolation you are seeking.

lmondada:
@Renaud-K You were right that `JitEngine` can be stored by value. I am not sure what compilation error I had stumbled across, but I must have mis-interpreted it.

Thanks a lot for the prototyping!

I've made that change in the latest commit. Does this go in the right direction? I would look into merging JitEngine and CompiledKernel in a separate PR, as that would involve changes to QPUs etc

Renaud-K (approved):

Renaud-K:
I am comfortable with the CompiledKernel changes, however the QPU.cpp changes maybe should be reviewed by @schweitzpgi .

Note that I would still like to break-up `cudaq-common` because it has a dependency with the `cudaq-operator`, which the compiler does not need. So, what would be `cudaq-device-code`, would have:
	• DeviceCodeRegistry.h/.cpp
	• CodeGenConfig{.h,.cpp}
	• Resources{.h,.cpp}
	• ThunkInterface.h
	• Kernel_utils.h
        • CompiledKernel{.h,.cpp}

That will also be a task for the next cycle.

lmondada:
@schweitzpgi Hey Eric, I'm starting to remove duplication between the `specializeModule` and `launchModule` paths.

After talking to Renaud, the `CompiledKernel` type this PR introduces seems like a good way to do that. It stores the function pointer and JitEngine in one place (and hence manages their lifetimes together), without any MLIR dependency. In a future PR I can replace the raw function pointers returned along the `specializeModule` path with the `CompiledKernel` type.

I think there is also an overlap with your "universal cache" idea. In the future, the compiled kernel type could be expanded to keep a hash of all launch arguments and be the main data type that we cache. Of course, more thinking will be required on this.

In terms of observable functionality, this PR is a no-op. The code that has been de-duplicated was identical across the two execution paths.

runtime/common/CompiledKernel.cpp
```diff
@@ -0,0 +1,35 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
```
schweitzpgi: nit: I don't think you can claim copyright back into the past on something new, like this file.
```suggestion
 * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
```

runtime/common/CompiledKernel.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "CompiledKernel.h"
+
+namespace cudaq {
```
schweitzpgi:
```suggestion
```
We should move to the LLVM style so as to better catch bugs, etc.

runtime/common/CompiledKernel.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+CompiledKernel::CompiledKernel(JitEngine engine, std::string kernelName,
```
schweitzpgi:
```suggestion
cudaq::CompiledKernel::CompiledKernel(JitEngine engine, std::string kernelName,
```

runtime/common/CompiledKernel.cpp
```diff
@@ -0,0 +1,35 @@
...
+      hasResult(hasResult) {}
+
+KernelThunkResultType
+CompiledKernel::execute(const std::vector<void *> &rawArgs) const {
+  auto funcPtr = getEntryPoint();
+  if (hasResult) {
+    void *buff = const_cast<void *>(rawArgs.back());
+    return reinterpret_cast<KernelThunkResultType (*)(void *, bool)>(funcPtr)(
+        buff, /*client_server=*/false);
+  } else {
```
schweitzpgi: nit: else after return.

https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

runtime/common/CompiledKernel.cpp
```diff
@@ -0,0 +1,35 @@
...
+    reinterpret_cast<void (*)()>(funcPtr)();
+    return {nullptr, 0};
+  }
+}
+
+void (*CompiledKernel::getEntryPoint() const)() { return entryPoint; }
+
+const JitEngine CompiledKernel::getEngine() const { return engine; }
```
schweitzpgi: Why `const` on the return value? It's rather useless, eh?

```c++
const int f() { return 0; }
int i = f();
```
is perfectly valid C++.

runtime/common/CompiledKernel.h
```diff
@@ -0,0 +1,55 @@
...
+public:
+  /// @brief Execute the JIT-ed kernel.
+  ///
+  /// If the kernel has a return type, the caller must have appended a result
+  /// buffer as the last element of \p rawArgs.
+  KernelThunkResultType execute(const std::vector<void *> &rawArgs) const;
+
+  // TODO: remove these two methods once the CompiledKernel is returned to
+  // Python.
+  void (*getEntryPoint() const)();
```
schweitzpgi, line 45: I'm dubious of this. CUDA-Q does not define entry point kernels as always having `void(void)` signatures. So this seems like a cul-de-sac.

runtime/common/JIT.cpp
```diff
@@ -325,6 +326,16 @@ cudaq::JitEngine cudaq::createQIRJITEngine(mlir::ModuleOp &moduleOp,
   return JitEngine(std::move(jitOrError.get()));
 }
 
+cudaq::CompiledKernel cudaq::createCompiledKernel(JitEngine engine,
+                                                  std::string kernelName,
+                                                  bool hasResult) {
+  std::string fullName = cudaq::runtime::cudaqGenPrefixName + kernelName;
+  std::string entryName = hasResult ? kernelName + ".thunk" : fullName;
```
schweitzpgi: This code looks like the efficiency hack used in Python. i.e., unrelated to the C++ implementation. But it is appearing here in what looks like common code?

schweitzpgi:
> @schweitzpgi Hey Eric, I'm starting to remove duplication between the `specializeModule` and `launchModule` paths.
> 
> After talking to Renaud, the `CompiledKernel` type this PR introduces seems like a good way to do that. It stores the function pointer and JitEngine in one place (and hence manages their lifetimes together), without any MLIR dependency. In a future PR I can replace the raw function pointers returned along the `specializeModule` path with the `CompiledKernel` type.
> 
> I think there is also an overlap with your "universal cache" idea. In the future, the compiled kernel type could be expanded to keep a hash of all launch arguments and be the main data type that we cache. Of course, more thinking will be required on this.
> 
> In terms of observable functionality, this PR is a no-op. The code that has been de-duplicated was identical across the two execution paths.

I'm trying to play catchup here, so I appreciate your patience.

- I don't understand the need to divorce "Compiled Kernels" from MLIR. In my mind, a CUDA-Q kernel *is* Quake *is* MLIR. The daylight between these things should be zero.
- The notion that a kernel can be reduced to a function pointer is just plain wrong. (Yeah, I know the specialization path used that, but that was a pre-existing artifact of the prototype and not good practice.) A kernel is quite literally a collection of functions and trying to be reductive and pretend it is a _single_ function pointer is only going to lead to future problems. At any rate, this is precisely why Python now threads _Modules_ through the runtime rather than cavalierly and incorrectly assuming it can lop off functions and data structures however it might fancy any time it deems it cool to do so. In short, we should keep the entire module of code and select the correct function for the circumstances. That is _not_ true, even now...

Thanks for combining the mostly duplicative paths. That seems like a good thing for sure.

Renaud-K:
>I don't understand the need to divorce "Compiled Kernels" from MLIR.

We have moved the compiler `cudaq-runtime-mlir` out of the QPUs and we will eventually want to remove the dependency all QPUs have with it. The QPUs should not have to link with `cudaq-runtime-mlir`. In modularizing the runtime, the QPU and compiler would be 2 independent entities that share information in a manner that does not couple them. This can be achieved with lower level libraries. In many ways, we already do that in the remote QPU case, where we really only pass the QPU a string representation of the kernel and other STL based containers. In the emulation case however, the QPU currently does the name lookup in the jitted code. This action creates a dependency. We would like to move to an exchange format between compiler and QPU which does not. So, just like an std::string, an std::function is low level and decouples the two. This is the motivation.

runtime/common/CompiledKernel.cpp
lmondada: Understood, I'm removing the `namespace {}` block.

runtime/common/CompiledKernel.cpp
lmondada: Fixed, thanks.

runtime/common/CompiledKernel.cpp
lmondada: 👍

runtime/common/CompiledKernel.cpp
lmondada: Yes, my bad! I was returning a const reference in a previous version and this slipped through when I changed it :)

runtime/common/CompiledKernel.h
lmondada: Yes, we need better support for handling function signatures. Currently the function pointers get cast to a different type when a KernelThunkResultType is expected to be returned. This only supports Python, where void(void) is used most of the time. 

As we discussed, the plan is to move towards handling function signatures better within `CompiledKernel` in the future.

runtime/common/JIT.cpp
lmondada: I think we addressed this in our discussion. I am starting with using this in Python and will expand to C++ later.

lmondada:
> /ok to test 2fab48a

lmondada:
> /ok to test 09aaf0f

runtime/common/CompiledKernel.h
```diff
@@ -0,0 +1,60 @@
...
```
schweitzpgi, line 45: Let's document that this entry point assumes that all arguments (if any) must have been eliminated via argument synthesis in order for this call to be made correctly.

runtime/common/JIT.cpp
```diff
@@ -332,6 +333,22 @@ cudaq::JitEngine cudaq::createQIRJITEngine(ModuleOp &moduleOp,
   return JitEngine(std::move(jitOrError.get()));
 }
 
+cudaq::CompiledKernel cudaq::createCompiledKernel(JitEngine engine,
+                                                  std::string kernelName,
+                                                  bool hasResult,
+                                                  bool hasVariationalArgs) {
```
schweitzpgi: nit: I'm not sure about the adjective `variational` here. Should probably just call them `runtime` or `dynamic` arguments or even `isParametric` to use more common computer science terminology.

runtime/common/CompiledKernel.h
lmondada: Good point, will do!

lmondada: ...unless the compiled kernel is parametric, in which case the entryPoint expects a buffer pointer as argument. Will document that.

runtime/common/JIT.cpp
lmondada: I agree that this will clarify things. I will take your `isParametric` suggestion.

lmondada: I could also invert the boolean argument and call it `isFullySpecialized`. Might be closer to the existing terminology in the codebase.

lmondada:
> /ok to test 1eb4e92

lmondada:
> /ok to test 5706578

PR #4054, opened 2026-02-26
Introduce python mechanism to explicitly persist JITEngines across launches

atgeller:
This PR improves benchmarking by providing a mechanism to persist the JITEngine across multiple launches, allowing properly benchmarking just on the execution time.

1. Introduces a global flag, `persist_jit_execution_context`, and global saved JITEngine
2. Introduces a `with reuse_compiler_artifacts` in python that will enable `persist_jit_execution_context`
3. When `persist_jit_execution_context` is set, the cached JIT from the execution context will be saved into the global JIT cache slot
4. When entering a new execution context, we check for `persist_jit_execution_context`, and if true we set the save the JITEngine from the execution context to the global JITEngine slot

Now, an example benchmarking case could look like
```[python]
with reuse_compiler_artifacts():
  # Warm up run
  cudaq.sample(kernel, args..., shots_count=1)
  # Real benchmarking
  time = benchmark(cudaq.sample(kernel, args..., shots_count=...)
  ```

### DIFF
--- a/.github/pre-commit/spelling_allowlist.txt
+++ b/.github/pre-commit/spelling_allowlist.txt
@@ -281,6 +281,7 @@ macOS
 makefiles
 merchantability
 mps
+multinomial
 multithreaded
 mutex
 namespace
@@ -319,6 +320,7 @@ qumode
 qumodes
 reStructuredText
 reconfigurable
+reproducibility
 runtime
 runtimes
 rvalue
--- a/.github/pre-commit/spelling_allowlist_cxx.txt
+++ b/.github/pre-commit/spelling_allowlist_cxx.txt
@@ -1,4 +1,5 @@
 clang
 gcc
 nullary
+pre
 typedef
--- a/.github/workflows/config/gitlab_commits.txt
+++ b/.github/workflows/config/gitlab_commits.txt
@@ -1,2 +1,2 @@
 nvidia-mgpu-repo: cuda-quantum/cuquantum-mgpu.git
-nvidia-mgpu-commit: e41bb18a5e5ff5be44090dfc4cb6c59fb511e9c7
+nvidia-mgpu-commit: 69fa05df00a069c3bc5c040294678014a396fa68
--- a/docs/sphinx/api/languages/cpp_api.rst
+++ b/docs/sphinx/api/languages/cpp_api.rst
@@ -120,6 +120,10 @@ Common
 .. doxygenclass:: cudaq::sample_result
     :members:
 
+.. _classcudaq_1_1ptsbe_1_1sample__result:
+
+**cudaq::`ptsbe`::sample_result** — Result type returned by ``ptsbe::sample()``, extending `cudaq::sample_result` with optional execution data (trace and per-trajectory info). See ``PTSBESampleResult.h`` and ``PTSBEExecutionData.h``.
+
 .. doxygenstruct:: cudaq::sample_options
     :members:
 
--- a/docs/sphinx/examples/python/ptsbe_end_to_end_workflow.ipynb
+++ b/docs/sphinx/examples/python/ptsbe_end_to_end_workflow.ipynb
@@ -1,0 +1,217 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## PTSBE end-to-end workflow\n",
+        "\n",
+        "**PTSBE** (Pre-Trajectory Sampling with Batch Execution) is a method for sampling from **noisy** quantum circuits efficiently. Instead of simulating the full density matrix and sampling once per shot, PTSBE:\n",
+        "\n",
+        "1. **Traces** the kernel to get the gate sequence and qubit layout.\n",
+        "2. **Extracts noise sites** by matching the noise model to the trace (each noisy gate becomes a *noise site* with a set of Kraus outcomes, e.g. I, X, Y, Z for depolarization).\n",
+        "3. **Generates trajectories** — each trajectory is one possible *realization* of noise (one Kraus outcome per site). A **sampling strategy** decides which trajectories to use (e.g. **Exhaustive**: all combinations, or **Probabilistic**: sample by probability).\n",
+        "4. **Allocates shots** across trajectories (e.g. **proportional** to trajectory probability, or **uniform**).\n",
+        "5. **Runs batches** — for each trajectory, the circuit is run as a **noiseless** circuit with that trajectory's outcomes applied; results are collected.\n",
+        "6. **Aggregates** all per-trajectory counts into a single `SampleResult`.\n",
+        "\n",
+        "You get the same statistics as density-matrix sampling (in the limit of many shots), but with the ability to batch many shots per trajectory and to control cost via the number of trajectories. This notebook runs the full workflow with a single API call: `cudaq.ptsbe.sample()`."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "### 1. Set up the environment\n",
+        "\n",
+        "Use the density-matrix simulator target (required for PTSBE). Set a random seed for reproducibility."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "import cudaq\n",
+        "\n",
+        "cudaq.set_target(\"density-matrix-cpu\")\n",
+        "cudaq.set_random_seed(42)"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "### 2. Define the circuit and noise model\n",
+        "\n",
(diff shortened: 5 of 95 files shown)

### CODE REVIEW

File: python/runtime/common/py_ExecutionContext.cpp [hunk 1]
```diff
@@ -24,6 +24,8 @@ void clearQirOutputLog();
 
 namespace cudaq {
 
+class PersistentCache {};
```
schweitzpgi: Does this really need to be in the cudaq namespace?

atgeller: Not at all. I'll move it.

runtime/common/ExecutionContext.cpp
```diff
@@ -40,9 +42,28 @@ bool isLastBatch() {
 std::size_t getCurrentQpuId() {
   return currentExecutionContext ? currentExecutionContext->qpuId : 0;
 }
+
 void detail::setExecutionContext(ExecutionContext *ctx) {
   currentExecutionContext = ctx;
+
+  if (currentExecutionContext && persist_cache
+      && jitEng.has_value())
+    currentExecutionContext->jitEng.emplace(jitEng.value());
```
Renaud-K: `emplace` is nice for constructors that take many arguments, making sure the constructor is called in-place. But here, this will just call the copy constructor so you may as well make it clear that it is a copy. `...jitEng = jitEng;`

atgeller: Done.

atgeller: Done.

python/runtime/common/py_ExecutionContext.cpp
```diff
@@ -22,6 +22,8 @@ std::string_view getQirOutputLog();
 void clearQirOutputLog();
 } // namespace nvqir
 
+class PersistentCache {};
```
schweitzpgi:
```suggestion
namespace {
class PersistentCache {};
}
```

schweitzpgi (approved):
LGTM

runtime/common/ExecutionContext.cpp
```diff
@@ -40,9 +42,27 @@ bool isLastBatch() {
 std::size_t getCurrentQpuId() {
   return currentExecutionContext ? currentExecutionContext->qpuId : 0;
 }
+
 void detail::setExecutionContext(ExecutionContext *ctx) {
   currentExecutionContext = ctx;
+
+  if (currentExecutionContext && persist_cache && jitEng.has_value())
+    currentExecutionContext->jitEng = jitEng.value();
```
Renaud-K, line 50: We are making an assumption here that we are launching the exact same kernel. If this is the level of support that we need and is sufficient for the release that's fine but what happens if it is is not the same kernel? Shouldn't we have some basic checks?

atgeller: The assumption is that this is only used for benchmarking and the user knows what they are doing. If it's not exactly the same kernel, with exactly the same arguments, and in exactly the same launch mode, things will blow up. We should make sure this is clear in `with reuse_compiler_artifact`. We have an idea for the longer-term to provide a built-in benchmarking API that will prevent handing the user a foot gun like we are now.

python/tests/kernel/test_reuse_compiler.py
```diff
@@ -0,0 +1,42 @@
...
+    cudaq.sample(simple, num_qubits, shots_count=1)
+    captured = capfd.readouterr()
+    assert "Using cached JIT engine" not in captured.out
+    with cudaq.cudaq_runtime.reuse_compiler_artifacts():
+        cudaq.sample(simple, num_qubits, shots_count=1)
+        captured = capfd.readouterr()
+        assert "Using cached JIT engine" not in captured.out
+        cudaq.sample(simple, num_qubits, shots_count=1)
+        captured = capfd.readouterr()
+        assert "Using cached JIT engine" in captured.out
```
Renaud-K: Could the test also check that a run of a a different kernel, after the context manager, work?

runtime/common/ExecutionContext.cpp
Renaud-K: Yes it would blow up. The long term solution should not rely on the execution context. It is a God object that we are looking at dismantling.

python/tests/kernel/test_reuse_compiler.py
atgeller: Good idea, I'll add a check.

python/runtime/common/py_ExecutionContext.cpp
```diff
@@ -128,5 +132,16 @@ void bindExecutionContext(py::module &mod) {
             const std::size_t bufferSize = parser.getBufferSize();
             std::memcpy(info.ptr, origBuffer, bufferSize);
           });
+
+  py::class_<PersistentCache>(mod, "reuse_compiler_artifacts")
```
lmondada: I think it's important to add a docstring for `reuse_compiler_artifacts` that clearly states

> Within a context, CUDAQ will blindly reuse compiled objects. It is up to the user to ensure that there are never two distinct computations launched within a single context.

python/runtime/common/py_ExecutionContext.cpp
```diff
@@ -22,6 +22,10 @@ std::string_view getQirOutputLog();
 void clearQirOutputLog();
 } // namespace nvqir
 
+namespace {
+class PersistentCache {};
```
lmondada: Can we call this `PersistentJitEngineContext` or something like that?

I think "cache" is ambiguous (what are we caching?) and also to a certain extent misleading: usually "caching" is harmless (i.e. always a good thing if one can spare the memory for it), whereas what we are dealing with here is an override mechanism that should only be used by users that know what they are doing.

runtime/common/ExecutionContext.cpp
```diff
@@ -11,6 +11,8 @@
 namespace {
 /// @brief Thread-local storage for the current execution context.
 thread_local cudaq::ExecutionContext *currentExecutionContext = nullptr;
+thread_local bool persist_cache = false;
```
lmondada: Here as well, I think `persistJitEngine` would be more descriptive.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -170,6 +170,8 @@ static std::optional<cudaq::JitEngine> alreadyBuiltJITCode() {
   auto *currentExecCtx = cudaq::getExecutionContext();
   if (!currentExecCtx || !currentExecCtx->allowJitEngineCaching)
     return std::nullopt;
+  if (currentExecCtx->jitEng)
+    CUDAQ_INFO("Using cached JIT engine for " + currentExecCtx->kernelName);
```
lmondada: This log is also a good opportunity to alert the user that this might not be doing what they think. E.g. output something along the lines of

```suggestion
    CUDAQ_INFO("Loading previously compiled JIT engine for {}. This will "
               "re-run the previous job, discarding any changes to the kernel, "
               "arguments or launch configuration.",
               currentExecCtx->kernelName);
```

python/tests/kernel/test_reuse_compiler.py
```diff
@@ -0,0 +1,39 @@
...
+    res = cudaq.sample(simple, 2, shots_count=1)
+    assert (res.count("11") == 1)
+    res = cudaq.sample(simple, 3, shots_count=1)
+    assert (res.count("111") == 1)
+    with cudaq.cudaq_runtime.reuse_compiler_artifacts():
+        res = cudaq.sample(simple, 4, shots_count=1)
+        assert (res.count("1111") == 1)
+        # Abuse the foot gun to make sure the cached kernel is rerun
+        # (and therefore the number of qubits is the same)
+        res = cudaq.sample(simple, 5, shots_count=1)
```
bettinaheim, line 36: This is a good start. Let's discuss follow up work that actually gives us proper caches (i.e. doesn't put all of the responsibility on the user to ensure that the right jit engine is used).

schweitzpgi:
@lmondada Please don't forget to unblock this PR as we want it to merge today.

python/tests/kernel/test_reuse_compiler.py
schweitzpgi: FWIW, this doesn't change the behavior from previous releases that used `.synthesize()` to separately perform argument synthesis and get raw pointers to host code blocks. (It doesn't expose raw pointers, which is a big improvement.) This gives the same behavior, just uses the normal launch path, and the normal launch path is what we truly want to benchmark.

lmondada (commented):
Just submitting a second review to ensure it's not blocked by my comments above.

python/runtime/common/py_ExecutionContext.cpp
atgeller: Done.

python/runtime/common/py_ExecutionContext.cpp
atgeller: Good point, I changed the naming to avoid the word "cache"

runtime/common/ExecutionContext.cpp
atgeller: Agreed

lmondada (approved):
Thanks Adam!

PR #4057, opened 2026-02-27
[PTSBE] Warn experimental ptsbe trajectory data

taalexander:
### Description
Add warning on accessing of PTSBE experimental/trajectory data about experimental nature.


sacpis (approved):
LGTM

PR #4058, opened 2026-02-27
[core] Add new ops to quake for value semantics support.

schweitzpgi:
Continue the work on supporting the "value semantics" form of the Quake dialect so as to enable moving some optimization passes to use the explicit dependence form of the IR.

  - adds syntactic sugar ops for array-like syntax
  - adds call_by_ref Op as syntactic sugar


schweitzpgi:
> /ok to test 9d9b46932

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -32,8 +73,46 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+        if (isUseConvertible(users))
+          continue;
+        return false;
+      }
+      return true;
+    };
+    if (isa<quake::VeqType>(allocOp.getType()) && !usesAreConvertible())
+      return failure();
+
+    if (auto stqTy = dyn_cast<quake::StruqType>(allocOp.getType())) {
```
atgeller: Do we need an `usesAreConvertible` check somewhere in here?

schweitzpgi: I don't think so. A value of `struq` type is of very limited utility. (Basically, you can create one and pass it to another kernel via a `call`.)

At any rate, we're just breaking up a `struq` allocation into a member-by-member set of allocations. Those members must be `ref` or `veq` types. For the former, we're done. For the latter, we recurse and `usesAreConvertible()` is called at that point.

schweitzpgi:
Noting that the `Remote-Sim/phase-folding/aliasing.cpp` test is failing on misc. jobs. Of course, it does not fail locally. 🤦‍♂️

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -32,8 +73,46 @@ class AllocaPat : public OpRewritePattern<quake::AllocaOp> {
...
+      SmallVector<Value> memAllocs;
+      for (auto memTy : stqTy.getMembers())
+        memAllocs.emplace_back(
+            rewriter.create<quake::AllocaOp>(loc, memTy).getResult());
+      // 2. Create a value of the original struq type using quake.make_struq.
+      auto aggregate =
+          rewriter.create<quake::MakeStruqOp>(loc, stqTy, memAllocs);
+      // 3. Walk all the uses. If they are quake.get_member operations, replace
+      // them with direct uses.
+      for (auto *user : allocOp->getUsers())
```
sacpis: Should we copy `allocaOp-getUsers()` to a variable like users as it is being modified on line 108 (`rewriter.replaceOp(getMem, memAllocs[index]);`)?

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -664,6 +664,35 @@ void quake::WrapOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+// CastAsRefOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult quake::CastAsRefOp::verify() {
+  std::size_t arity = 1;
+  if (auto ty = dyn_cast<CableType>(getValue().getType()))
+    arity = ty.getSize();
+
+  if (getValue().getType() != getOutVal().getType())
+    return failure();
```
sacpis: nit: Should we put some descriptive message instead of returning it empty? Same as we did it in LeftTeeOp::verify [here](https://github.com/NVIDIA/cuda-quantum/pull/4058/changes#diff-71646259467aaf9c1b5f5c6a6df3baaab524e1cac33f1ac4167b63c5d7aef70eR773).

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -664,6 +664,35 @@ void quake::WrapOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+
+  // Make sure the ref output has the same arity.
+  std::size_t coarity = 0;
+  auto outTy = getOutRef().getType();
+  if (auto ty = dyn_cast<StruqType>(outTy)) {
+    if (!ty.hasSpecifiedSize())
+      return failure();
```
sacpis: same here.

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -664,6 +664,35 @@ void quake::WrapOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    coarity = ty.getArity().value();
+  } else if (auto ty = dyn_cast<VeqType>(outTy)) {
+    if (!ty.hasSpecifiedSize())
+      return failure();
```
sacpis: same here.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -131,126 +246,37 @@ class FactorQuantumAllocationsPass
...
-      if (d.getReference().getDefiningOp<quake::InitializeStateOp>())
-        return true;
-      if (auto ty = dyn_cast<quake::VeqType>(d.getReference().getType()))
-        return !ty.hasSpecifiedSize();
-      return true;
-    });
-    if (failed(applyPartialConversion(func.getOperation(), target,
-                                      std::move(patterns)))) {
-      signalPassFailure();
+    if (failed(applyPatternsAndFoldGreedily(func, std::move(patterns))))
```
sacpis: Should we call `signalPassFailure();` before returning so that the pipeline knows about this failure?

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -131,126 +246,37 @@ class FactorQuantumAllocationsPass
...
-            continue;
-          return false;
-        }
-        return true;
-      }();
-      if (usesAreConvertible)
-        allocations.push_back(alloc);
-    });
-    if (allocations.empty())
+    if (failed(applyPatternsAndFoldGreedily(func, std::move(patterns))))
```
sacpis: Should we call `signalPassFailure();` before returning so that the pipeline knows about this failure?

lib/Optimizer/Dialect/Quake/QuakeTypes.cpp
```diff
@@ -80,6 +80,34 @@ Type quake::StruqType::parse(AsmParser &parser) {
...
+  for (auto ty : getMembers())
+    if (auto veqTy = llvm::dyn_cast<quake::VeqType>(ty))
+      if (!veqTy.hasSpecifiedSize())
+        return false;
+  return true;
+}
+
+std::optional<std::size_t> quake::StruqType::getArity() const {
+  if (getMembers().empty())
+    return std::nullopt;
```
sacpis: I think we should return `{0}` from here. The reason being we are calling this `getArity()` function from CastAsRefOp::verify(). If we return nullopt, then calling .value() on it would throw an error.

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -692,6 +733,88 @@ def quake_TerminateCableOp : QuakeOp<"terminate_cable"> {
...
+    position all shifted down by 1.
+
+    This syntactic sugar is provided to capture a common array-like idiom and
+    give the illusion that cables can be used like an imperative array.
+
+    Example:
+    ```mlir
+      // create a cable with 3 wires: 0, 1, and 2
+      %20 = quake.null_cable !quake.cable<3>
+      %21:2 = quake.left_tee %20[1] : (!cable<3>) -> (!wire, !cable<2>)
```
sacpis: nit: `(!quake.cable<3>) -> (!quake.wire, !quake.cable<2>)`

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -692,6 +733,88 @@ def quake_TerminateCableOp : QuakeOp<"terminate_cable"> {
...
+      // %21#0 contains wire 1 previous at position 1 in %20
```
sacpis: nit: `previously`

sacpis (commented):
Overall LGTM. Thanks @schweitzpgi. Left few comments.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi: No. This change was because I don't want the pass to fail at all.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi: I'll add a variable, but just FYI `replaceOp` doesn't delete IR out from underneath itself. It's phased. The patterns create new IR and do bookkeeping on what needs to get updated. And once all the updates have been made, another pass is made to commit all those updates.

sacpis: Thanks @schweitzpgi for letting me know.

lib/Optimizer/Dialect/Quake/QuakeTypes.cpp
schweitzpgi: I wrote it this way so I could write a simple "does this struq have arity >= 1" very simply. I will refactor this bit for clarity, etc.

PR #4061, opened 2026-02-28
[core] Consolidate all the various QIR verifiers into 1 library.

schweitzpgi:
QIR verification needs to happen at the very tail end of code generation to determine if the original source code actually conforms with the miscellaneous limitations and restrictions of the QIR specification.

This PR collects all the various random places that these verifications are done and consolidates them into one library.

### DIFF
--- a/include/cudaq/Verifier/QIRSpec.h
+++ b/include/cudaq/Verifier/QIRSpec.h
@@ -10,7 +10,10 @@
 
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Support/LogicalResult.h"
-#include "mlir/Target/LLVMIR/Export.h"
+
+namespace llvm {
+class Module;
+}
 
 namespace cudaq::verifier {
 
--- a/lib/Optimizer/CodeGen/Pipelines.cpp
+++ b/lib/Optimizer/CodeGen/Pipelines.cpp
@@ -22,10 +22,6 @@ struct TargetCodegenPipelineOptions
       *this, "loops-may-have-break",
       llvm::cl::desc("Enable break statements in loops."),
       llvm::cl::init(true)};
-  PassOptions::Option<bool> appendDeprecatedVerifier{
-      *this, "append-verifier",
-      llvm::cl::desc("Append the QIR verifier pipeline."),
-      llvm::cl::init(false)};
   PassOptions::Option<std::string> target{
       *this, "convert-to", llvm::cl::desc("Conversion target specifier."),
       llvm::cl::init("")};
--- a/lib/Verifier/QIRSpec.cpp
+++ b/lib/Verifier/QIRSpec.cpp
@@ -12,6 +12,7 @@
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Instructions.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Target/LLVMIR/Export.h"
 
 /**
    \file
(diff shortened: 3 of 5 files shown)


lib/Optimizer/CodeGen/ConvertToQIRProfile.cpp
```diff
@@ -594,17 +594,10 @@ std::unique_ptr<Pass> cudaq::opt::createQIRProfilePreparationPass() {
 //===----------------------------------------------------------------------===//
 // The various passes defined here should be added as a pass pipeline.
 
-void cudaq::opt::addQIRProfileVerify(OpPassManager &pm,
```
sacpis, line 597: Seems like we will need to remove its declaration from `include/cudaq/Optimizer/CodeGen/Passes.h` as well.

include/cudaq/Optimizer/CodeGen/Passes.td
```diff
@@ -156,38 +156,6 @@ def RemoveMeasurements : Pass<"remove-measurements"> {
...
-    functions, i.e., starting with a __quantum prefix, or NVQIR runtime
-    functions.
-  }];
-  
-  let options = [
-    ListOption<"allowedFuncs", "allowed-funcs", "llvm::StringRef",
-        "Allowed list of functions in addition to NVQIR-defined function.">
-  ];
-
-  let constructor = "cudaq::opt::createVerifyNVQIRCallOpsPass({})";
```
sacpis, line 175: Let's remove  `createVerifyNVQIRCallOpsPass` declaration from `include/cudaq/Optimizer/CodeGen/Passes.h`.

lib/Verifier/QIRSpec.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  return isValidFloatBinaryInst(inst) || llvm::isa<llvm::FCmpInst>(inst) ||
+         llvm::isa<llvm::FPExtInst>(inst) ||
+         llvm::isa<llvm::FPTruncInst>(inst) ||
+         llvm::isa<llvm::SelectInst>(inst) || llvm::isa<llvm::PHINode>(inst);
+}
+
+static bool isValidOutputCallInstruction(llvm::Instruction &inst) {
+  if (auto *call = llvm::dyn_cast<llvm::CallBase>(&inst)) {
+    auto name = call->getCalledFunction()->getName().str();
```
sacpis: Let's guard it with 

```
if (auto *func = call->getCalledFunction()) {
    auto name = func->getName().str();
    ...
}
```

runtime/common/JIT.cpp
```diff
@@ -283,26 +285,30 @@ cudaq::JitEngine cudaq::createQIRJITEngine(mlir::ModuleOp &moduleOp,
...
-    if (mlir::failed(pm.run(module))) {
+    if (failed(pm.run(module))) {
       engine.eraseHandler(handlerId);
       throw std::runtime_error("[createQIRJITEngine] Lowering to QIR for "
                                "remote emulation failed.\n" +
                                error_msg);
     }
+    if (auto mod = dyn_cast<ModuleOp>(module))
+      if (failed(cudaq::verifier::checkQIRLLVMIRDialect(mod, profileName)))
+        return {};
```
sacpis: Are we silently returning on verification failure? Shouldn't we throw runtime_error?

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -107,17 +107,13 @@ void createTargetCodegenPipeline(PassManager &pm,
   pm.addPass(createConvertMathToFuncs());
   pm.addPass(createSymbolDCEPass());
   pm.addPass(cudaq::opt::createCCToLLVM());
-  if (options.appendDeprecatedVerifier)
```
sacpis, line 110: Let's remove `appendDeprecatedVerifier` declaration from `lib/Optimizer/CodeGen/Pipelines.cpp`.

lib/Verifier/QIRSpec.cpp
```diff
@@ -0,0 +1,140 @@
...
+    std::vector<const char *> outputFunctions{
+        cudaq::opt::QIRBoolRecordOutput, cudaq::opt::QIRIntegerRecordOutput,
+        cudaq::opt::QIRDoubleRecordOutput, cudaq::opt::QIRTupleRecordOutput,
+        cudaq::opt::QIRArrayRecordOutput};
+    return std::find(outputFunctions.begin(), outputFunctions.end(),
+                     name.c_str()) == outputFunctions.end();
```
sacpis: Question: Wouldn't find over std::vector<const char *> compare pointers? Would you please check it? Because name.c_str() returns a heap pointer, and we are comparing it here with the static constexpr pointers, no?

include/cudaq/Verifier/QIRSpec.h
```diff
@@ -0,0 +1,28 @@
...
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Target/LLVMIR/Export.h"
```
sacpis: nit: would forward declaration for `llvm::Module *`  be light here instead of importing `Export.h`?

lib/Verifier/QIRSpec.cpp
schweitzpgi: I'll double check. Interestingly enough, I believe this is the original code. :)

PR #4064, opened 2026-03-02
Fix bug in exp_pauli qubit concatenation

huaweil-nv:
### Description

Fix a Python operator precedence bug in PyKernel.exp_pauli() where individual qubits (RefType) are silently dropped when a quantum register (VeqType) is also provided.
The + operator has higher precedence than the ternary if/else expression in Python, so:
```
[quantumVal] if quantumVal is not None else [] + qubitsList
```
is parsed as:
```
[quantumVal] if quantumVal is not None else ([] + qubitsList)
```
This means when quantumVal is not None, the result is [quantumVal] alone — qubitsList is discarded entirely. The fix adds parentheses to ensure qubitsList is always concatenated:
```
([quantumVal] if quantumVal is not None else []) + qubitsList
```

**How to reproduce**:
```
import cudaq

kernel = cudaq.make_kernel()
qreg = kernel.qalloc(2)
q_extra = kernel.qalloc()
kernel.exp_pauli(1.0, qreg, q_extra, 'XXX')

print(kernel)
```
Before fix: The generated IR shows q_extra allocated but never used. exp_pauli operates on veq<2> (2 qubits) despite the 3-qubit Pauli word "XXX":
```
%1 = quake.alloca !quake.ref
quake.exp_pauli (%cst) %0 to "XXX" : (f64, !quake.veq<2>) -> ()
```
After fix: quake.concat correctly merges the register and the individual qubit into veq<3>:
```
%1 = quake.alloca !quake.ref
%2 = quake.concat %0, %1 : (!quake.veq<2>, !quake.ref) -> !quake.veq<3>
quake.exp_pauli (%cst) %2 to "XXX" : (f64, !quake.veq<3>) -> ()
```

### DIFF
--- a/python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
@@ -68,7 +68,7 @@ pySamplePTSBE(const std::string &shortName, MlirModule module,
 
   ptsbe::sample_result result;
   try {
-    result = ptsbe::runSamplingPTSBE(
+    result = ptsbe::detail::runSamplingPTSBE(
         [&]() mutable {
           [[maybe_unused]] auto res =
               clean_launch_module(shortName, mod, retTy, opaques);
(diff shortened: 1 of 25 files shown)


sacpis:
> /ok to test c06e3be

PR #4065, opened 2026-03-02
Fix segfault in `EvolveResult.final_expectation_values()` when no observables were provided

huaweil-nv:
### Description

EvolveResult.final_expectation_values() and EvolveResult.final_state() crash with a segfault when the underlying std::optional fields are nullopt. This happens when evolve() is called without observables — expectation_values is never populated, but the accessor directly dereferences it without checking.
The method's own docstring states "This value will be None if no observables were specified in the call", but the actual behavior is a segfault. Users can see this docstring in Python via:
```
>>> help(cudaq.EvolveResult.final_expectation_values)
```
**How to reproduce**:
```
import cudaq
from cudaq.operators import *
from cudaq.dynamics import *
import numpy as np

cudaq.set_target('density-matrix-cpu')

dimensions = {0: 2}
rho0 = cudaq.State.from_data(
    np.array([[1.0, 0.0], [0.0, 0.0]], dtype=np.complex128))

steps = np.linspace(0, 10, 11)
schedule = Schedule(steps, ["time"])

result = cudaq.evolve(hamiltonian, dimensions, schedule, rho0)
result.final_expectation_values()  # Segmentation fault
```
Fix:
Add has_value() and empty() checks in py_EvolveResult.cpp for both final_expectation_values() and final_state(). Return py::none() when the data is not available, consistent with the documented behavior.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
@@ -68,7 +68,7 @@ pySamplePTSBE(const std::string &shortName, MlirModule module,
 
   ptsbe::sample_result result;
   try {
-    result = ptsbe::runSamplingPTSBE(
+    result = ptsbe::detail::runSamplingPTSBE(
         [&]() mutable {
           [[maybe_unused]] auto res =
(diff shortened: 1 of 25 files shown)


sacpis:
> /ok to test 593f8aa

PR #4066, opened 2026-03-02
Fix a potential segfault in conda environment

### LINKED ISSUE #4062: Segfault in conda environment
### Required prerequisites

### Describe the bug

Segfault in test_observe_numpy_array
https://github.com/NVIDIA/cuda-quantum/actions/runs/22490877926/job/65162837231#step:7:22409

This is affecting publishing pipeline.

Not known yet. It appears in CI.

### Expected behavior

Should run successfully

Not a regression

### Environment

- **CUDA-Q version**: 0.13.0
- **Python version**: 3.12.3
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

_No response_

sacpis:
When a kernel raises inside a with ExecutionContext block (cudaq.observe called with a list argument that is too short), the __exit__ exception was still calling platform.finalizeExecutionContext(). On the custatevec GPU backend this triggers an API call against an uninitialized simulator state, causing a segfault.

Update: Have put back the `finalizeExecutionContext` in __exit__ and handling the 0 allocated qubits inside simulator.

Found second issue. When there are 0 qubit allocated, calling setToZeroState (in endExecution) on an unitialized GPU state causes segfault as was evident from the failing jobs in the first run of this PR.

Adding a fix to handle a case for the observe path via handleObservation.

Fixes #4062

### DIFF
--- a/python/runtime/common/py_EvolveResult.cpp
+++ b/python/runtime/common/py_EvolveResult.cpp
@@ -35,11 +35,15 @@ void bindEvolveResult(py::module &mod) {
       .def(py::init<std::vector<state>, std::vector<std::vector<double>>>())
       .def(
           "final_state",
-          [](evolve_result &self) { return self.states->back(); },
+          [](evolve_result &self) -> py::object {
+            if (!self.states.has_value() || self.states->empty())
+              return py::none();
+            return py::cast(self.states->back());
+          },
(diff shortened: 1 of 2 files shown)


taalexander:
Thanks @sacpis!

PR #4067, opened 2026-03-02
Add errors for misuse of `reuse_compiler_artifact`

atgeller:
This PR builds on #4054, adding checks and throwing runtime errors in the following situations where the compiler artifacts are incorrectly reused across launches:
1. The second launch is for a kernel with a different name.
2. The second launch has different explicit arguments.

This does not handle cases where any captured arguments (python kernel decorators) are changed in between launches when reusing artifacts.

### DIFF
--- a/python/tests/kernel/test_reuse_compiler.py
+++ b/python/tests/kernel/test_reuse_compiler.py
@@ -121,3 +121,21 @@ def test_reuse_complex_arguments():
         assert (res.count("11") == 1)
         with pytest.raises(RuntimeError):
             cudaq.sample(apply_complex_angles, different_angles, shots_count=1)
+
+
+def test_different_launch_mode():
+    """Reuse validation should reject different launch modes"""
+
+    @cudaq.kernel
+    def simple(numQubits: int):
+        qubits = cudaq.qvector(numQubits)
+        x(qubits.front())
+        for i, qubit in enumerate(qubits.front(numQubits - 1)):
+            x.ctrl(qubit, qubits[i + 1])
+
+    with cudaq.cudaq_runtime.reuse_compiler_artifacts():
+        res = cudaq.sample(simple, 4, shots_count=1)
+        assert (res.count("1111") == 1)
+
+        with pytest.raises(RuntimeError):
+            res = cudaq.get_state(simple, 4)
--- a/runtime/common/ExecutionContext.cpp
+++ b/runtime/common/ExecutionContext.cpp
@@ -23,11 +23,10 @@ thread_local cudaq::ExecutionContext *currentExecutionContext = nullptr;
 namespace cudaq {
 
 namespace compiler_artifact {
+thread_local bool reuseArtifact = false;
+
 class SavedCompilerArtifact {
 public:
-  const std::optional<cudaq::JitEngine> &getJitEngine() const { return jitEng; }
-  void setJitEngine(const cudaq::JitEngine engine) { jitEng = engine; }
-
   void checkArtifactReuse(const std::string &kernelName,
                           const std::vector<void *> &args,
                           const cudaq::JitEngine &engine,
@@ -69,12 +68,30 @@ public:
 
   SavedCompilerArtifact() : argBuff(nullptr, free) {}
 
+  void saveEngineForReuse(ExecutionContext *ctx) {
+    if (!reuseArtifact || !ctx)
+      return;
+    jitEng = ctx->jitEng;
+    launchMode = ctx->name;
+  }
+
+  void reuseEngineIfPresent(ExecutionContext *ctx) {
+    if (!reuseArtifact || !ctx || !jitEng.has_value())
+      return;
+
+    if (launchMode != ctx->name)
+      throw std::runtime_error(
+          "Detected reuse of compiler artifact with different launch mode");
+    ctx->jitEng = jitEng.value();
+  }
+
 private:
   std::optional<cudaq::JitEngine> jitEng = std::nullopt;
   // This is actually going to be a pointer into the jitEng,
   // but we have to store it explicitly due to linking issues.
   int64_t (*argsCreator)(const void *, void **);
   std::string kernelName;
+  std::string launchMode;
   std::unique_ptr<void, decltype(&free)> argBuff;
   size_t argSize = 0;
 
@@ -88,7 +105,6 @@ private:
   }
 };
 
-thread_local bool reuseArtifact = false;
 thread_local SavedCompilerArtifact savedArtifact;
 
 /// This will cause the JITEngine stored in the current execution context to be
@@ -109,19 +125,6 @@ void checkArtifactReuse(const std::string kernelName,
     return;
 
   savedArtifact.checkArtifactReuse(kernelName, args, jit, argsCreatorThunk);
-}
-
-void saveEngineForReuse(ExecutionContext *ctx) {
-  if (reuseArtifact && ctx && ctx->jitEng.has_value())
-    savedArtifact.setJitEngine(ctx->jitEng.value());
-}
-
-void reuseEngineIfPresent(ExecutionContext *ctx) {
-  if (!reuseArtifact || !ctx)
-    return;
-  auto engine = savedArtifact.getJitEngine();
-  if (engine.has_value())
-    ctx->jitEng = engine.value();
 }
 } // namespace compiler_artifact
 
@@ -148,14 +151,12 @@ std::size_t getCurrentQpuId() {
 }
 
 void detail::setExecutionContext(ExecutionContext *ctx) {
+  compiler_artifact::savedArtifact.reuseEngineIfPresent(ctx);
   currentExecutionContext = ctx;
-
-  compiler_artifact::reuseEngineIfPresent(ctx);
 }
 
 void detail::resetExecutionContext() {
-  compiler_artifact::saveEngineForReuse(currentExecutionContext);
-
+  compiler_artifact::savedArtifact.saveEngineForReuse(currentExecutionContext);
   currentExecutionContext = nullptr;
 }
 } // namespace cudaq


runtime/common/ExecutionContext.h
```diff
@@ -208,6 +208,11 @@ void resetExecutionContext();
 
 void enablePersistentJITEngine();
 void disablePersistentJITEngine();
+bool isPersistingJITEngine();
+void saveLaunchInfo(std::string_view kernelName, void *argMessageBuffer,
+                    size_t size);
+bool isKernelSame(std::string_view kernelName);
+bool isLaunchInfoSame(void *argMessageBuffer, size_t size);
```
Renaud-K: It might be time to provide some scope to these small functions. Maybe as static members of the savedCompilerArtifact class even if it means splitting the class with a header or in a CompilerArtifact namespace. The class is really a thread local singleton so we could make it explicit. I know this is temporary but I would be more comfortable with something a bit more structured.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -238,6 +244,27 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
+      if (!builtJIT && !cudaq::detail::isKernelSame(name))
+        throw std::runtime_error("Detected reuse of compiler artifact with "
+                                   "a different kernel.");
+      auto funcPtr = jit->lookupRawNameOrFail(name + ".argsCreator");
+      auto argsCreator =
+          reinterpret_cast<int64_t (*)(const void *, void **)>(funcPtr);
+      void *res_buffer;
+      const void *argBlock = rawArgs.data();
+      auto res_size = argsCreator(argBlock, &res_buffer);
+      if (builtJIT) {
```
Renaud-K: Something like 
```
  auto deleter = [](void* ptr) { std::free(ptr); };
  std::unique_ptr<void, decltype(deleter)> scopedRefBuffer(res_buffer, deleter);
```
makes it clearer whose responsibility it is to delete the buffer. 

You would have to move it when passing it `saveLaunchInfo` making it clear that it is its reponsibility. 

To isLaunchInfoSame, you pass the raw pointer with `saveLaunchInfo.get()`, to signify no ownership.

Finally, if it has not been moved, it will be automatically deleted when leaving the scope.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -238,6 +244,27 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
+        cudaq::detail::saveLaunchInfo(name, res_buffer, res_size);
+      } else {
+        auto can_reuse = cudaq::detail::isLaunchInfoSame(res_buffer, res_size);
```
Renaud-K: Could we be consistent with using camelCase throughout?

runtime/common/ExecutionContext.cpp
```diff
@@ -7,12 +7,63 @@
...
+  void setJitEngine(const cudaq::JitEngine &engine) { jitEng = engine; }
+
+  void reset() {
+    jitEng.reset();
+    clearArgs();
+  }
+
+  // We're responsible for freeing argMessageBuffer.
+  void saveLaunchInfo(std::string_view kernelName, void *argMessageBuffer,
+                      size_t size) {
```
sacpis: nit: Just to be on the safe side, should we call clearArgs() before entering this function? In case, if a second call to saveLaunchInfo was called before clearArgs, that would overwrite old argBuff and it would be leaked.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -238,6 +244,27 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
       jit = cudaq::createQIRJITEngine(module, "qir:");
     }
 
+    if (cudaq::detail::isPersistingJITEngine()) {
+      if (!builtJIT && !cudaq::detail::isKernelSame(name))
```
sacpis: Are we missing to call `isKernelSame` on the first call when builtJIT is true?

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Done.

runtime/common/ExecutionContext.h
atgeller: I think adding them to the SavedCompilerArtifact class makes sense, I'll make that change.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Done.

runtime/common/ExecutionContext.cpp
atgeller: ~~I've made `argBuff` into a unique pointer which should solve this issue.~~

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: If `builtJIT` is true, that means that there wasn't a saved JITEngine present, so there's nothing to compare to.

python/tests/kernel/test_reuse_compiler.py
```diff
@@ -31,9 +35,21 @@ def simple(numQubits: int):
     with cudaq.cudaq_runtime.reuse_compiler_artifacts():
         res = cudaq.sample(simple, 4, shots_count=1)
         assert (res.count("1111") == 1)
-        # Abuse the foot gun to make sure the cached kernel is rerun
-        # (and therefore the number of qubits is the same)
-        res = cudaq.sample(simple, 5, shots_count=1)
+
+        res = cudaq.sample(simple, 4, shots_count=1)
```
lmondada, line 39: This is no longer testing what it was testing before: i.e. making sure that the kernel is not JITed again.

Though thinking about it, I'm not sure how we'd check that from Python... This entire "hack" is annoying in this sense because it makes testing very difficult, but I think we'll have to live with it for now.

python/tests/kernel/test_reuse_compiler.py
```diff
@@ -31,9 +35,21 @@ def simple(numQubits: int):
...
         assert (res.count("1111") == 1)
+
+        @cudaq.kernel
+        def simple(numQubits: int):
+            qubits = cudaq.qvector(numQubits)
+        with pytest.raises(RuntimeError):
+            res = cudaq.sample(simple, 4, shots_count=1)
+
+        with pytest.raises(RuntimeError):
+            res = cudaq.sample(simple, 5, shots_count=1)
```
lmondada: Test this before re-defining `simple`, so that we actually test for passing a different argument, and not passing a different kernel as in the previous test.

runtime/common/ExecutionContext.cpp
```diff
@@ -7,12 +7,63 @@
  ******************************************************************************/
 
 #include "ExecutionContext.h"
+#include <cstdlib>
+#include <cstring>
+#include <string>
+
+namespace cudaq::detail {
+class savedCompilerArtifact {
```
lmondada:
```suggestion
class SavedCompilerArtifact {
```

runtime/cudaq/platform/default/python/QPU.cpp
lmondada: @atgeller I don't see this change. Did you push it? I agree with Renaud that using a unique_ptr here and in the `SavedCompilerArtifact` class would make ownership more explicit, and would also let you get rid of `clearArgs` and rely on the default destructor instead.

lmondada (commented):
Thanks @atgeller! Did you by any chance forget to push your last commit?

runtime/common/ExecutionContext.cpp
lmondada: You'd have to update the name elsewhere too, might be easier to use your IDE's renaming tool.

runtime/common/ExecutionContext.h
atgeller: I wound up making a `compiler_artifact` namespace.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Should be pushed now

runtime/common/ExecutionContext.cpp
atgeller: Turns out that a unique_ptr to void doesn't go over well, so I've given up on that and instead just called `clearArgs()` here as you suggested.

atgeller:
> Thanks @atgeller! Did you by any chance forget to push your last commit?

Sure did! Should be pushed now.

python/tests/kernel/test_reuse_compiler.py
atgeller: Sadly yes. I ran some local tests with debug statements, and I'll run one locally with some timing checks as well to double check.

runtime/common/ExecutionContext.cpp
atgeller: I think I fixed this typo as part of the changes that didn't get pushed earlier, so it should be good now.

runtime/common/ExecutionContext.h
Renaud-K: Thank you. It it still a great improvement.

runtime/common/ExecutionContext.cpp
Renaud-K: What was wrong with unique_ptr? `unique_ptr<void>` alone will not work, that's why you need to provide a custom deleter. But this could just be a type alias in the `compiler_artifact` namespace. Instead of a lambda instance, it could be a function object, so the instance is created internally.
```
struct FreeDeleter { void operator()(void* p) const { free(p); } };
using BufferUniquePtr = std::unique_ptr<void, FreeDeleter>
```

atgeller: I figured out how to make it work :)

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -244,22 +246,25 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
-      auto res_size = argsCreator(argBlock, &res_buffer);
+      void *resBuffer;
+      auto resSize = argsCreator(rawArgs.data(), &resBuffer);
+      auto deleter = [](void *ptr) { std::free(ptr); };
+      std::unique_ptr<void, decltype(deleter)> scopedRefBuffer(resBuffer,
+                                                               deleter);
       if (builtJIT) {
-        cudaq::detail::saveLaunchInfo(name, res_buffer, res_size);
+        cudaq::compiler_artifact::saveArtifactInfo(
+            name, scopedRefBuffer.release(), resSize);
```
lmondada: Here you shouldn't have to release the `unique_ptr`. Just move it and modify `saveArtifactInfo` to take a unique_ptr as argument, right?

runtime/common/ExecutionContext.cpp
```diff
@@ -97,48 +123,18 @@ std::size_t getCurrentQpuId() {
...
-void detail::enablePersistentJITEngine() { persistJITEngine = true; }
-
-void detail::disablePersistentJITEngine() {
-  persistJITEngine = false;
-  savedArtifact.reset();
-}
-
-bool detail::isPersistingJITEngine() { return persistJITEngine; }
-
-// We're responsible for freeing argMessageBuffer
```
lmondada: if you change the signature below to taking a `unique_ptr`, remove this comment as it is superfluous (and sounds like it isn't handled automatically).

runtime/common/ExecutionContext.cpp
```diff
@@ -11,66 +11,92 @@
...
+} // namespace
+
+namespace cudaq {
+
+namespace compiler_artifact {
+struct SavedCompilerArtifact {
 public:
   bool hasJitEngine() const { return jitEng.has_value(); }
-  const std::optional<cudaq::JitEngine> &getJitEngine() const { return jitEng; }
+  const cudaq::JitEngine &getJitEngine() const { return jitEng.value(); }
```
lmondada: You can return the JitEngine by value, as it is made to be cheap to copy.

runtime/common/ExecutionContext.cpp
```diff
@@ -11,66 +11,92 @@
...
+
+namespace {
+/// @brief Thread-local storage for the current execution context.
+thread_local cudaq::ExecutionContext *currentExecutionContext = nullptr;
+} // namespace
+
+namespace cudaq {
+
+namespace compiler_artifact {
+struct SavedCompilerArtifact {
```
lmondada: nit: I'm not sure whether our style guide dictates rules on this, but I'd say this is a `class` at this point.

runtime/common/ExecutionContext.h
```diff
@@ -205,14 +205,16 @@ void setExecutionContext(ExecutionContext *ctx);
 /// Use `quantum_platform::with_execution_context` instead of setting/resetting
 /// the execution context manually.
 void resetExecutionContext();
+} // namespace detail
 
+// TODO: document
```
lmondada: either document or remove TODO?

runtime/common/ExecutionContext.cpp
lmondada:
```suggestion
class SavedCompilerArtifact {
```

runtime/common/ExecutionContext.cpp
```diff
@@ -7,20 +7,97 @@
...
+    argSize = 0;
+  }
+
+  // We're responsible for freeing argMessageBuffer.
+  void saveInfo(std::string_view kernelName, void *argMessageBuffer,
+                size_t size) {
+    assert(argMessageBuffer);
+    argSize = size;
+    this->kernelName = kernelName;
+    argBuff = std::unique_ptr<void, decltype(&free)>(argMessageBuffer, free);
```
Renaud-K: The implementer of this interface has not been granted the right to take the ownership of the pointer. The client can assume that it can reuse the passed pointer after the call. The unique_ptr needs to be passed through the interface. This tells the client that it is no longer responsible for it and that the pointer is no longer viable. The callee needs to call `std::move(ptr)` instead of `ptr.release()`.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -238,6 +246,30 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
+      if (!builtJIT && !cudaq::compiler_artifact::isKernelSame(name))
+        throw std::runtime_error("Detected reuse of compiler artifact with "
+                                 "a different kernel.");
+      auto funcPtr = jit->lookupRawNameOrFail(name + ".argsCreator");
+      auto argsCreator =
+          reinterpret_cast<int64_t (*)(const void *, void **)>(funcPtr);
+      void *resBuffer;
+      auto resSize = argsCreator(rawArgs.data(), &resBuffer);
+      auto deleter = [](void *ptr) { std::free(ptr); };
+      std::unique_ptr<void, decltype(deleter)> scopedRefBuffer(resBuffer,
```
Renaud-K: To be able to move, the types should be the same.

runtime/common/ExecutionContext.cpp
```diff
@@ -7,20 +7,124 @@
...
-bool isUsingResourceCounterSimulator();
-} // namespace nvqir
-
 namespace cudaq {
 
+namespace compiler_artifact {
+class SavedCompilerArtifact {
+public:
+  const std::optional<cudaq::JitEngine> &getJitEngine() const { return jitEng; }
+  void setJitEngine(const cudaq::JitEngine &engine) { jitEng = engine; }
```
Renaud-K: nit: JitEngine use value semantic. Pass it by value.

Renaud-K (approved):
It looks good to me. Feel free to resolve the conversations as needed.
Thank you for applying the feedback.

lmondada (commented):
👍

runtime/common/ExecutionContext.cpp
```diff
@@ -7,20 +7,124 @@
...
+
+  void checkArtifactReuse(const std::string kernelName,
```
sacpis: nit: Should it be `&kernelName`?

PR #4072, opened 2026-03-03
[python] Use resolve_captured_arguments in kernel builder

lmondada:
A minor change that reuses the logic of `PyKernelDecorator.resolve_captured_arguments` to resolve the captured arguments of kernel decorators within the builder. Other than making sure the behaviour is consistent, the main benefit of this is that it no longer requires a `.defFrame` attribute to be defined.

### DIFF
--- a/python/cudaq/kernel/kernel_builder.py
+++ b/python/cudaq/kernel/kernel_builder.py
@@ -29,12 +29,12 @@ from cudaq.mlir._mlir_libs._quakeDialects import (
 from cudaq.kernel_types import qubit, qvector
 from .common.fermionic_swap import fermionic_swap_builder
 from .common.givens import givens_builder
-from .kernel_decorator import isa_kernel_decorator
+from .kernel_decorator import DecoratorCapture, isa_kernel_decorator
 from .quake_value import QuakeValue
 from .utils import (emitFatalError, emitWarning, nvqppPrefix, getMLIRContext,
                     recover_func_op, mlirTypeToPyType, cudaq__unique_attr_name,
                     mlirTypeFromPyType, emitErrorIfInvalidPauli,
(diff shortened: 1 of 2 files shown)


python/cudaq/kernel/kernel_builder.py
```diff
@@ -1356,42 +1356,44 @@ def apply_call(self, target, *target_arguments):
...
-                for var in target.captured_variables():
-                    v = recover_value_of(var.name, target.defFrame)
-                    if isa_kernel_decorator(v):
+                captured_args = target.resolved
+                for arg in captured_args:
+                    if isinstance(arg, DecoratorCapture):
                         # The recursive step
-                        v = self.resolve_callable_arg(inner, v)
+                        v = self.resolve_callable_arg(inner, arg)
                     else:
```
atgeller, line 1395: Is there support for C++ linked kernels in builder mode? I didn't add support when I was updating how C++ linked kernels are captured, so we may need to add logic to handle them here (similar to https://github.com/NVIDIA/cuda-quantum/blob/e89df77d4e9ffd1aaf3830f194871480dec856b6/python/cudaq/kernel/kernel_decorator.py#L507)

schweitzpgi (commented):
Looks like a good change.

Agree with Adam's comment that we might as well check against C++-defined kernels that may have gotten loaded when resolving callables here in the builder as well, if we're not already doing so.

PS: which means adding a test. :)

python/cudaq/kernel/kernel_builder.py
lmondada: Yes sure, I will add support for this. I am working on this but will make a separate PR. It's a bit of work as I need to somehow infer the callable type in Python, presumably by parsing the MLIR.

Issue #3727, opened 2026-01-08
Consider using python's async def coroutines instead of homegrown


schweitzpgi:
The CUDA-Q launchers have async variants. These have been implemented in C++ with python bindings, which makes choreographing the python objects, the python garbage collector, the C++ objects, the GIL, references between all these pieces, rather complicated.

It may be worth considering implementing the async aspects in Python and leaving only the quantum kernels themselves at the C++ layer.

schweitzpgi:
See #3724

sacpis:
Working on it.

Issue #3678, opened 2025-12-10
Seg fault in test_evolve_simulators


sacpis:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

We saw a segmentation fault in test_evolve_simulators in the following job. The changes in the PR are unrelated to this test.
https://github.com/NVIDIA/cuda-quantum/actions/runs/20087333345/job/57631843542?pr=3676#step:9:658

### Steps to reproduce the bug

N/A

### Expected behavior

This test should run successfully in CI.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 0.13.0
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: redhat

sacpis:
Reopening as seems like the root cause is in `reset_target` function and needs to be fixed.

PR #4077, opened 2026-03-03
Enabling test_evolve_async_no_intermediate_results

sacpis:
Enabling test_evolve_async_no_intermediate_results

Temp fix for #3678. Basically this is just a mitigation for the segafult. This needs a discussion with the team (issue #3727) as whether to go with the Python's async def coroutines or use internal ones.

@bettinaheim for viz.


sacpis:
This test is now causing the same seg fault in the test which runs after it. Disabling it.
https://github.com/NVIDIA/cuda-quantum/actions/runs/22785314123/job/66105277482?pr=4091#step:9:679

PR #4080, opened 2026-03-03
[scaleway] Follow-up work to address integration test failures

khalatepradnya:
* This PR follows the work done in https://github.com/NVIDIA/cuda-quantum/pull/3781
* Between the previous testing and now the integration tests started failing with -
  `RuntimeError: Scaleway Job Error: error: 'NoneType' object has no attribute 'get'`
* Fix was to include an empty options dictionary in quantum computation parameters.
* Add test file to be run against real test server and move limited tests to be checked against mock server

### DIFF
--- a/targettests/scaleway/test-int8_t.cpp
+++ b/targettests/scaleway/test-int8_t.cpp
@@ -7,4 +7,4 @@
  ******************************************************************************/
 
 // RUN: echo skipping
-#include "../execution/test-int8_t.cpp"
+#include "../execution/int8_t.cpp"
--- a/targettests/scaleway/test-int8_t_free_func.cpp
+++ b/targettests/scaleway/test-int8_t_free_func.cpp
@@ -7,4 +7,4 @@
  ******************************************************************************/
 
 // RUN: echo skipping
-#include "../execution/test-int8_t_free_func.cpp"
+#include "../execution/int8_t_free_func.cpp"


python/tests/backends/test_scaleway.py
```diff
@@ -10,61 +10,35 @@
...
-        "scaleway",
-        machine=TEST_PLATFORM,
-        max_duration=DEFAULT_DURATION,
-        max_idle_duration=DEFAULT_DURATION,
-        project_id=TEST_PROJECT_ID,
-        url=TEST_URL,
-        deduplication_id=DEFAULT_DEDUPLICATION_ID,
-    )
-
+def do_something():
```
sacpis, line 24: nit: let's change its name :) to something like setUp().

khalatepradnya: Yep, this ought to be fixed throughout... I had a PR for the update across the repo, I'll revive it as follow-up.

python/tests/backends/test_scaleway.py
```diff
@@ -489,13 +490,39 @@ def test():
...
-    assert not "01" in counts
-    assert not "10" in counts
+
+def test_toffoli():
+
+    @cudaq.kernel
+    def kernel():
+        q = cudaq.qvector(3)
+        x(q)
+        x.ctrl([q[0], q[1]], q[2])
```
sacpis, line 507: nit: let's use x.ctrl in `test_multi_qubit_kernel` as well?

khalatepradnya: For coverage, it is included in `test_qvector_kernel`.

sacpis: Sorry, coverage of?

khalatepradnya: Test coverage of `x.ctrl`, I thought that's what you are asking for?

sacpis: Ah no, I meant just to make it similar across all test, we can use `x.ctrl` in `test_multi_qubit_kernel` as well.

khalatepradnya: umm... `cx` is still supported so I want to check that it works, `x.ctrl` is checked in many places.

PR #4081, opened 2026-03-03
[macOS] Make installer paths user/sdk user-independent 

taalexander:
### Description
I was doing some local testing of the installer and found it was not relocatable to my system. While I had added a non-root test it was still using the same CI/root users and SDKs which did not fully exercise the variations we might see in practice.

I've added fixes and updated CI tests to make sure install is test as non-root *and* a different user.


.github/workflows/publishing.yml
```diff
@@ -1000,44 +1000,62 @@ jobs:
...
+          mkdir -p "$workdir"
+          cp -R docs/sphinx/examples "$workdir/examples"
+          cp -R docs/sphinx/applications "$workdir/applications"
+          cp -R docs/sphinx/targets "$workdir/targets"
+          rm -rf "$workdir/examples/python" "$workdir/applications/python" "$workdir/targets/python"
+          cp scripts/validate_installation.sh "$workdir/"
+          chown -R cudaqtest "$workdir"
+
+          touch "$GITHUB_STEP_SUMMARY"
+          chmod a+w "$GITHUB_STEP_SUMMARY"
```
mitchdz, line 1046: `GITHUB_STEP_SUMMARY` should already be created by the runner, did you have issues with it?

taalexander: I agree it should be present. I don't think I had issues with it not being there but rather the downstream permissions for the new user when debugging (hence the permissions). I don't think the touch should be harmful though?

mitchdz: yeah, if it was needed, then this is fine.

PR #4084, opened 2026-03-04
[PTSBE] Address flaky shot noise in tests

taalexander:
### Description
Attempted fix for tests that appear to be flaky. Attributing to shot noise right now.


taalexander:
> Thank you for the fix!
I'm curious if in any of these scenarios, setting random seed would make the test deterministic, while keeping shots count low?

I agree these tests are missing the seed being set. I think I agree that setting it in principal should fix this as the RNG does use the cudaq seed but also it will still be vulnerable to platform dependence, etc. of RNG. 

I'd prefer to set the seed and also keep the higher shot counts (they should be cheap and most tests use ~1000). Let me go back and update this.

PR #4086, opened 2026-03-04
[python] Compute kernel return type from module definition

lmondada:
When executing Python kernels, we were previously passing the return type as argument through the entire stack. This is not necessary as the provided return type is _always_ the same as the return type of the MLIR `FuncOp` (I have checked). Given that we are also passing along the `ModuleOp`, we can just compute the return type when required.

### DIFF
--- a/include/cudaq/Optimizer/Builder/Runtime.h
+++ b/include/cudaq/Optimizer/Builder/Runtime.h
@@ -9,7 +9,6 @@
 #pragma once
 
 #include "cudaq/Optimizer/Builder/Factory.h"
-#include <stdexcept>
 
 //===----------------------------------------------------------------------===//
 //


include/cudaq/Optimizer/Builder/Runtime.h
```diff
@@ -92,4 +93,22 @@ static constexpr const char extractDevPtr[] =
...
+
+  auto numResults = funcOp.getFunctionType().getNumResults();
+  switch (numResults) {
+  case 0:
+    return nullptr;
+  case 1:
+    return funcOp.getFunctionType().getResult(0);
+  default:
+    return nullptr;
+  }
```
sacpis: Wondering why do we need switch when we can just use ternary operator, as there are just 2 options to be returned from.

lmondada: You're right. This was a left-over from a previous version where I was treating `> 1` differently. I've fixed it as you suggest.

include/cudaq/Optimizer/Builder/Runtime.h
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "cudaq/Optimizer/Builder/Factory.h"
+#include <stdexcept>
```
sacpis: Do we need this import?

sacpis (approved):
LGTM. Thanks @lmondada. Left a comment.

include/cudaq/Optimizer/Builder/Runtime.h
lmondada: Good spot thanks, removing it!

PR #4089, opened 2026-03-04
[macOS] Fixes for transient CI cache miss

taalexander:
### Description
This is an attempt to cut down on locations where the macOS ci is missing cache and triggering a full LLVM rebuild even if not required. The root cause appears to be small Cmake hashing differences so we're trying to reduce these (this is a relatively fragile mechanism).


schweitzpgi:
Does this decrease the macOS builder time? It's a serious long pole in the CI at the moment.

taalexander:
Yes, indirectly. The intent is that it fixes the caching. 

When the caching is working macOS is about 60% of the Linux CI time. Macos CI is slow right now because some jobs are missing cache. It is not fully clear why but these are some potential mitigations I wish to try.

PR #4093, opened 2026-03-05
Fix integer overflow for 31+ qubits

huaweil-nv:
### Description

std::accumulate and std::reduce calls in CuDensityMatState.cpp and CuDensityMatOpConverter.cpp use the integer literal 1 as the init value. Since 1 has type int, the C++ template deduces the accumulator type as int (32-bit signed). When the product of Hilbert space dimensions exceeds INT_MAX (2^31 - 1 = 2,147,483,647), the accumulator overflows, producing a wrong result and causing initialize_cudm to reject valid state data with:
```
ValueError: Invalid hilbertSpaceDims for the state data
```
This PR replaces the int literal 1 with properly typed init values (std::size_t{1} or int64_t{1}) in all four affected call sites.
**Root Cause**
std::accumulate determines its accumulator type from the init parameter type, not from the container element type or the binary operation's return type:
```
// Before fix: T = int (32-bit), overflows when product > 2^31 - 1
std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>());

// After fix: T = size_t (64-bit), no overflow
std::accumulate(dims.begin(), dims.end(), std::size_t{1}, std::multiplies<>());
```
The effective (pseudo) expansion of the buggy code is:
```
int acc = 1;
for (auto& dim : hilbertSpaceDims)
    acc = acc * dim;   // int × int64_t → truncated back to int
return acc;            // wrong value after overflow
```

**Repro**
Environment
  GPU: NVIDIA A100-PCIE-40GB (requires >= 40 GB GPU memory)
  CUDA: 12.6
Trigger condition
The product of all Hilbert space dimensions must exceed INT_MAX (2,147,483,647). The minimum configuration is 31 qubits (each with dimension 2), giving a total dimension of 2^31 = 2,147,483,648.

Steps to reproduce：
```
import gc, sys
import numpy as np
import cudaq
from cudaq import spin

def run_evolve(num_qubits):
    dim = 2 ** num_qubits
    arr = np.zeros(dim, dtype=np.complex128)
    arr[0] = 1.0
    state = cudaq.State.from_data(arr)
    del arr
    gc.collect()
    cudaq.evolve(
        spin.z(0),
        {i: 2 for i in range(num_qubits)},
        cudaq.Schedule(np.linspace(0, 1, 2), ["t"]),
        initial_state=state,
    )

if __name__ == "__main__":
    cudaq.set_target("dynamics")

    # Control: 20 qubits (dim = 2^20, well below INT_MAX) — should pass
    print("[Control] 20 qubits...", end=" ")
    run_evolve(20)
    print("PASSED")

    # Test: 31 qubits (dim = 2^31 > INT_MAX, ~34 GB state vector)
    print("[Test]    31 qubits...", end=" ")
    run_evolve(31)
    print("PASSED")
```
Run the reproducer:
```
PYTHONPATH=build/python \
LD_LIBRARY_PATH=build/lib:$LD_LIBRARY_PATH \
CUDAQ_DEFAULT_SIMULATOR=qpp \
python3 repro_accumulate_overflow.py
```

Expected result (after fix)
```
[Control] 20 qubits... PASSED
[Test]    31 qubits... PASSED
```
(Note: on a 40 GB GPU, the 31-qubit test may fail with a CUDA out-of-memory error during evolve — this is expected since 34 GB state + working buffers exceeds 40 GB. The key verification is that the error is no longer "Invalid hilbertSpaceDims for the state data", confirming the overflow is fixed. A GPU with >= 80 GB is needed to run the full evolution.)

Actual result (before fix)
```
[Control] 20 qubits... PASSED
[Test]    31 qubits... FAILED: Invalid hilbertSpaceDims for the state data
```

### DIFF
--- a/unittests/ptsbe/PTSSamplingStrategyTester.cpp
+++ b/unittests/ptsbe/PTSSamplingStrategyTester.cpp
@@ -579,7 +579,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByErrorCount) {
     return traj.countErrors() == 1;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 2);
@@ -598,7 +598,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByNoErrors) {
     return traj.countErrors() == 0;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 1);
@@ -613,7 +613,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByProbabilityThreshold) {
     return traj.probability > 0.1;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 2);
@@ -630,7 +630,7 @@ TEST(ConditionalSamplingStrategyTest, FilterNonePass) {
     return traj.probability > 1.0;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 0);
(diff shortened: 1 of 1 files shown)


runtime/nvqir/cudensitymat/CuDensityMatOpConverter.cpp
```diff
@@ -249,7 +249,8 @@ cudaq::dynamics::CuDensityMatOpConverter::createElementaryOperator(
         if (diags.empty())
           return false;
         const auto dim = std::accumulate(
-            subspaceExtents.begin(), subspaceExtents.end(), 1,
+            subspaceExtents.begin(), subspaceExtents.end(),
+            static_cast<decltype(subspaceExtents)::value_type>(1),
```
sacpis: nit: Can we use `int64_t{1}` just as you have used it on line 276?

huaweil-nv: thanks, fixed in commit https://github.com/NVIDIA/cuda-quantum/pull/4093/changes/5297d4ec0113a5451c4c85167d87e38793c8bb78

huaweil-nv:
> /ok to test 3d7505aaa34644ed63233fb020920eadf243fbcf

huaweil-nv:
> /ok to test a2df5258b72ba44d8cfe0e148ff8d99f2cf6de26

sacpis:
> /ok to test c761092

sacpis:
> /ok to test 4ed2dfb

PR #4095, opened 2026-03-05
[Realtime] Remove dispatcher's DOCA dependency

boschmitt:
### Description

The aim here is mainly to remove the direct DOCA dependency from the dispatcher library, ensuring it remains transport-agnostic.

This also refines the unified dispatch mode by introducing a transport-agnostic API that encapsulates transport-specific details, allowing for a more flexible integration with various transport layers.

Key changes:
- Introduction of `cudaq_unified_dispatch_ctx_t` to bundle the launch function and transport context, keeping transport-specific types hidden from the caller.

- Moving Hololink-specific unified dispatch kernel that utilizes DOCA GPUNetIO verbs to the Hololink bridge library.

### Open question:

- It is not clear that changing `doca_gpu_dev_verbs_fence_release<DOCA_GPUNETIO_VERBS_SYNC_SCOPE_GPU>()` to `__threadfence_system()` is the best thing to do. We need a fence on those points, but we can't rely on DOCA for the dispatcher.


boschmitt:
Merged by other means.

PR #4098, opened 2026-03-05
[core] Rewrite the factor quantum allocations pass.

schweitzpgi:


lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -22,6 +22,54 @@ namespace cudaq::opt {
...
+    if (sub.hasConstantUpperBound())
+      return {sub.getConstantUpperBound()};
+    return cudaq::opt::factory::getIntIfConstant(sub.getUpper());
+  }();
+  if (!(lowInt && upInt))
+    return false;
+  for (auto *subUser : sub->getUsers())
+    if (!isUseConvertible(subUser))
+      return false;
+  return true;
```
sacpis, line 63: Wondering if we should create a function as it is being used on line 67-70?

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -102,22 +185,59 @@ class DeallocPat : public OpRewritePattern<quake::DeallocOp> {
...
+        Type memTy = iter.value();
+        auto mem = rewriter.create<quake::GetMemberOp>(loc, memTy, alloc,
+                                                       iter.index());
+        if (auto veqTy = dyn_cast<quake::VeqType>(memTy))
+          generateDeallocs(veqTy, rewriter, loc, mem);
+        else
+          rewriter.create<quake::DeallocOp>(loc, mem);
+      }
+    } else {
+      return dealloc.emitOpError("internal error: must be veq or struq.");
```
sacpis, line 223: Wondering if we will ever reach here as on line 190-205, we are only allowing `VeqType` or `StruqType` to move ahead.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
```diff
@@ -102,22 +185,59 @@ class DeallocPat : public OpRewritePattern<quake::DeallocOp> {
 
   LogicalResult matchAndRewrite(quake::DeallocOp dealloc,
                                 PatternRewriter &rewriter) const override {
```
sacpis, line 187: Do we need this check `if (isa<quake::StruqType>(allocOp.getType()) && !usesAreConvertible(allocOp))` for `StruqType`?

sacpis (approved):
Overall LGTM. Thanks @schweitzpgi. Left a few comments.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi: This will be done in a later PR. :)

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi: Left this for safety for now. I'll clean this one up in the next PR as well.

lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
schweitzpgi: No. We can always lower a struq type allocation to an allocation of its members followed by grouping those references as a struq.

PR #4100, opened 2026-03-05
Update notices

mitchdz:
This updates our `NOTICE` file to be more correct, and adds copyright years where available. Also adds attribution details for all packages installed via apt/pip in our docker container.

You can view the results at https://ghcr.io/nvidia/cuda-quantum-src

Running workflow: https://github.com/NVIDIA/cuda-quantum/actions/runs/22733929134

Also included is 2 scripts that are used in conjunction with the `-src` container build.
* `scripts/generate_pip_attribution.py`
* `scripts/generate_apt_attribution.py`

These scripts are used in the `build_package_sources` workflow to automatically generate attribution files.

The result is the source container has the following files:
```bash
root@b7b8a6f4231d:/sources/NOTICES# ls -l
total 260
-rw-r--r-- 1 root root    632 Mar  5 19:54 LICENSE
-rw-r--r-- 1 root root   8231 Mar  5 19:54 NOTICE
-rw-r--r-- 1 root root   4717 Mar  5 19:55 NOTICE_APT_cudaq_cu13
-rw-r--r-- 1 root root   3098 Mar  5 19:55 NOTICE_APT_cudaqx_cu13
-rw-r--r-- 1 root root 173668 Mar  5 19:55 NOTICE_PIP_cudaq_cu13
-rw-r--r-- 1 root root   1470 Mar  5 19:55 NOTICE_PIP_cudaqx_cu13
-rw-r--r-- 1 root root  56571 Mar  5 19:55 NOTICE_PIP_macos_cu13
```

Where,
* `LICENSE` is our license
* `NOTICE` is the notice for building the source code
* `NOTICE_APT_*` is auto-generated NOTICE files from APT. These are not the easiest to retrieve, so they mostly point to the actual notice file on the filesystem of the cuda-quantum container itself.
* `NOTICE_PIP_*` is auto-generated NOTICE files from PIP. These are very easy to extract and nicely formatted. On certain copyrights such as MIT, the entire copyright is included per the legal requirements.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -19,7 +19,6 @@ License at <https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT>
 Eigen - various licenses, primarily MPL2 and BSD
 <https://gitlab.com/libeigen/eigen>
 
-TODO: COPYRIGHT YEAR
 The incorporated source code and its licenses can be found as a submodule on the CUDA-Q repository.
 See also <https://gitlab.com/libeigen/eigen/-/blob/master/COPYING.README>.
 


NOTICE
```diff
@@ -1,21 +1,25 @@
...
+Originally developed by University of Illinois at Urbana-Champaign.
+The incorporated source code and its license can be found as a submodule on the CUDA-Q repository.
 License at <https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT>
 
 ----------------------------------------------------------------
 
 Eigen - various licenses, primarily MPL2 and BSD
 <https://gitlab.com/libeigen/eigen>
 
+TODO: COPYRIGHT YEAR
```
1tnguyen: Just confirming that we will still leave this as a `TODO`?

mitchdz: Good catch. This should be removed.

PR #4102, opened 2026-03-05
[bug-fix] Validate the simulation data used to create state from data

khalatepradnya:
This PR improves the error handling and validation for simulation state creation from raw data
 -  Enforce stricter checks for pointer validity and size, 
 - Throw a runtime error if the dimensions and data size in `simulationData` do not match, or if data is provided without dimensions, ensuring only well-formed simulation states are deserialized.


1tnguyen (approved):
LGTM 👍

PR #4103, opened 2026-03-06
Add option to get GKE to work for python kernel builder

atgeller:
Kernels built using the python kernel builder have dummy host entry points, ignore them in GKE so it doesn't refuse to run.


lib/Optimizer/Transforms/GenKernelExecution.cpp
```diff
@@ -1037,9 +1037,15 @@ class GenerateKernelExecution
         auto mangledAttr = mangledNameMap.getAs<StringAttr>(funcOp.getName());
         assert(mangledAttr && "funcOp must appear in mangled name map");
         StringRef mangledName = mangledAttr.getValue();
-        auto [hostEntryNeeded, hostFunc] =
-            cudaq::opt::marshal::lookupHostEntryPointFunc(mangledName, module,
-                                                          funcOp);
+        bool &&hostEntryNeeded = false;
+        func::FuncOp &&hostFunc = func::FuncOp{};
```
schweitzpgi:
```suggestion
        bool hostEntryNeeded = false;
        func::FuncOp hostFunc;
```
I don't think you need to declare these as r-value references. Simple variables should be fine.

PR #4104, opened 2026-03-06
[WIP] Restoring backward ABI compatibility with cudaq_solver 0.5.0

Renaud-K:


docs/sphinx/applications/python/logical_aim_sqale.ipynb
```diff
@@ -25,21 +25,21 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
Renaud-K, line 28: Did not mean to have this. Not sure how it got there.

sacpis: Seems like you missed to save the notebook after running it?

bettinaheim: doesn't matter either way - this is something that would be nice to add some scripts for to "sanitize" how notebooks are checked in at some point.

Renaud-K:
We will not be insuring ABI compatibility between for cudaq 0.14.0 and cudaqx 0.5.0.

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -153,17 +160,17 @@ std::optional<QubitConnectivity> quantum_platform::connectivity() {
   return platformQPUs.front()->getConnectivity();
 }
 
-bool quantum_platform::is_simulator(std::size_t qpu_id) const {
+bool quantum_platform::is_simulator(const std::size_t qpu_id) const {
```
schweitzpgi, line 163: Just curious: were these arguments declared `const` previously?

Renaud-K: Yes.

PR #4105, opened 2026-03-06
Bug fix for using `cudaq.translate` with kernel builder

huaweil-nv:
### Description
cudaq.translate() does not work with builder kernels created via cudaq.make_kernel(). Two layered bugs prevent it:

* translate.py references isa_dynamic_kernel and mk_decorator but never imports them, causing an immediate NameError.
* mk_decorator passes the raw builder.module (which lacks the cc.python_uniqued module attribute) instead of the compiled builder.qkeModule to PyKernelDecorator. The C++ translate_impl requires this attribute to locate the entry-point function.

### Minimal reproducer
```
import cudaq

kernel = cudaq.make_kernel()
q = kernel.qalloc(2)
kernel.h(q[0])
kernel.cx(q[0], q[1])
kernel.mz(q)

# Before fix:
#   NameError: name 'isa_dynamic_kernel' is not defined
# After fixing imports only:
#   RuntimeError: Module is malformed for python. Requires unique entry point attribute.
# After this PR:
#   Returns QIR string as expected.
print(cudaq.translate(kernel, format="qir"))
```

### Root cause
mk_decorator wraps a builder kernel into a PyKernelDecorator, but it forwards builder.module — the raw, uncompiled MLIR module. The compiled module (builder.qkeModule), which carries the cc.python_uniqued attribute, is only created by builder.compile(). Since mk_decorator skips compilation, any downstream C++ API that depends on this attribute (currently translate_impl) fails.

Other APIs are unaffected for different reasons: cudaq.sample() calls kernel(*args) directly (triggering PyKernel.__call__ → compile()), and cudaq.draw() does use mk_decorator but its C++ implementation does not check for cc.python_uniqued.

### Fix
* python/cudaq/runtime/translate.py: Add missing imports for mk_decorator and isa_dynamic_kernel.
* python/cudaq/kernel/kernel_decorator.py: In mk_decorator, call builder.compile() and use builder.qkeModule instead of builder.module. compile() is idempotent (guarded by if not hasattr(self, 'qkeModule')), so this is safe for all callers.
* python/tests/builder/test_translate.py (new): 9 tests — builder kernel translate across all formats (QIR, QIR-base, QIR-adaptive, OpenQASM2, parameterized QIR), error paths (invalid args, invalid type), and smoke tests for cudaq.draw and cudaq.get_unitary which share the mk_decorator code path.

### DIFF
--- a/unittests/ptsbe/PTSSamplingStrategyTester.cpp
+++ b/unittests/ptsbe/PTSSamplingStrategyTester.cpp
@@ -579,7 +579,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByErrorCount) {
     return traj.countErrors() == 1;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 2);
@@ -598,7 +598,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByNoErrors) {
     return traj.countErrors() == 0;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 1);
@@ -613,7 +613,7 @@ TEST(ConditionalSamplingStrategyTest, FilterByProbabilityThreshold) {
     return traj.probability > 0.1;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 2);
@@ -630,7 +630,7 @@ TEST(ConditionalSamplingStrategyTest, FilterNonePass) {
     return traj.probability > 1.0;
   };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 0);
@@ -641,7 +641,7 @@ TEST(ConditionalSamplingStrategyTest, FilterAllPass) {
 
   auto predicate = [](const cudaq::KrausTrajectory &) { return true; };
 
-  ConditionalSamplingStrategy strategy(predicate);
+  ConditionalSamplingStrategy strategy(predicate, /*seed=*/42);
   auto trajectories = strategy.generateTrajectories(noise_points, 10);
 
   EXPECT_EQ(trajectories.size(), 4);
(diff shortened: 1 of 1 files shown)


python/tests/builder/test_translate.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_translate_builder_invalid_extra_args():
+    kernel = cudaq.make_kernel()
+    q = kernel.qalloc(2)
+    kernel.h(q[0])
+    kernel.mz(q)
+    with pytest.raises(RuntimeError) as e:
+        cudaq.translate(kernel, 5, format="qir")
+    assert "Invalid number of argu" in repr(e)
```
schweitzpgi: Does this test appear somewhere else in another test file?

python/tests/builder/test_translate.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_translate_builder_invalid_type():
+    with pytest.raises(RuntimeError) as e:
+        cudaq.translate("not_a_kernel", format="qir")
+    assert "kernel is invalid type" in repr(e)
```
schweitzpgi: Same question.

It's a good idea to consolidate these tests in one place, but if there are other older tests for `.translate` they should be moved here or deleted. Let's not increase the volume of tests without moving the needle in actual coverage.

python/tests/builder/test_translate.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Smoke tests for other APIs sharing mk_decorator path
+def test_draw_builder():
+    kernel = cudaq.make_kernel()
+    q = kernel.qalloc(2)
+    kernel.h(q[0])
+    kernel.cx(q[0], q[1])
+    drawing = cudaq.draw(kernel)
+    assert "h" in drawing
+    assert "x" in drawing
```
schweitzpgi: These asserts are very, very low bar. What can be done to improve them so they don't pass with a burst of line noise as input?

python/tests/builder/test_translate.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_translate_builder_openqasm():
+    kernel = cudaq.make_kernel()
+    q = kernel.qalloc(2)
+    kernel.h(q[0])
+    kernel.cx(q[0], q[1])
+    kernel.mz(q)
+    asm = cudaq.translate(kernel, format="openqasm2")
+    assert "qreg" in asm
```
schweitzpgi: Generally, these assertion checks are very minimal. I think we can deploy pytest's API of checks and comparisons to our advantage a bit more here and check that the output is what we expect to a greater degree of certainty.

python/tests/builder/test_translate.py
huaweil-nv: Yes, test_kernel_translate.py already covers this error path with decorator kernels (e.g., test_translate_qir_ignored_args). Since the error-handling logic is shared after mk_decorator wraps the builder kernel, the builder variant doesn't add meaningful coverage. Removed in 15d6693.

python/tests/builder/test_translate.py
huaweil-nv: Good point. This particular error path ("kernel is invalid type") wasn't tested elsewhere, but the branch it covers is trivial and input-type-agnostic — it doesn't depend on whether the caller intended a builder or decorator kernel. Removed to keep the test file focused on the actual builder kernel translate functionality. See 15d6693.

python/tests/builder/test_translate.py
huaweil-nv: Replaced with an exact full-output comparison (assert drawing == expected_str) against the known circuit drawing. See 15d6693.

python/tests/builder/test_translate.py
huaweil-nv: Agreed. Strengthened assertions across the board:

* OpenQASM: now checks "OPENQASM 2.0;", "qreg var0[2];", "h var0[0];", and "cx var0[0], var0[1];".
* QIR: now also checks for __quantum__qis__h and __quantum__qis__x__ctl gate symbols.
* get_unitary: replaced single-element check with np.testing.assert_allclose against the full expected Hadamard matrix (tolerance tightened from 1e-6 to 1e-12).

See 15d6693.

sacpis:
> /ok to test 3e212d0

sacpis:
> /ok to test 53e3113

PR #4109, opened 2026-03-06
[DO NOT MERGE] Set the kernel reuse flag for broadcast observe.

schweitzpgi:
This is meant to improve performance for the broadcast observe launcher code path.

### DIFF
--- a/python/cudaq/runtime/observe.py
+++ b/python/cudaq/runtime/observe.py
@@ -35,12 +35,14 @@ def __broadcastObserve(kernel, spin_operator, *args, shots_count=0, qpu_id=0):
         ctx.totalIterations = N
         ctx.batchIteration = i
         ctx.setSpinOperator(spin_operator)
+        ctx.allowJitEngineCaching = True
         with ctx:
             kernel(*a)
         res = ctx.result
         results.append(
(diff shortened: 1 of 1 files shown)


python/cudaq/runtime/observe.py
```diff
@@ -35,12 +35,14 @@ def __broadcastObserve(kernel, spin_operator, *args, shots_count=0, qpu_id=0):
...
         ctx.batchIteration = i
         ctx.setSpinOperator(spin_operator)
+        ctx.allowJitEngineCaching = True
         with ctx:
             kernel(*a)
         res = ctx.result
         results.append(
             cudaq_runtime.ObserveResult(ctx.getExpectationValue(),
                                         spin_operator, res))
+        ctx.unset_jit_engine()
```
sacpis, line 45: I tried this change with the huge neural networks notebooks, but did not see any significant improvement. As this still trhows away the JIT engine and each time the kernel is compiled from scratch. Taking out the execution context, just creating it once and using it for subsequent iterations boost the perf a bit. But then we have to make few other changes to keep the args as live function parameters in MLIR.

schweitzpgi: Got it. Yep.

schweitzpgi:
This is a smoke test. Do not merge!

PR #4112, opened 2026-03-06
Latest hololink migration, transport-agnostic unified dispatch, and DGX Spark emulator support

cketcham2333:
- Latest hololink migration: Adapted the hololink wrapper to the latest internal hololink API (immediate connection, deferred QP, ring buffer getters).

- Transport-agnostic unified dispatch: Moved the DOCA-specific unified dispatch kernel into the hololink bridge layer (libcudaq-realtime-bridge-hololink.so), keeping libcudaq-realtime free of DOCA dependencies. Made the unified dispatch symbol weak so bridge tools that don't use it don't need to link the library.

- DGX Spark (iGPU) emulator support:
   - Fixed CUDA lazy module loading deadlock by setting CUDA_MODULE_LOADING=EAGER in orchestration scripts.
   - Fixed emulator buffer overrun when num_pages > NUM_BUFFERS.
   - Updated emulator ILA to 585-bit PTP format (19 banks).
   - Added serial_number and peer_ip metadata fields required by latest hololink library.
   - Added detect_cuda_arch helper for forward-compatible builds (e.g. CUDA 12.6 on GB10 targets sm_90).

### DIFF
--- a/realtime/include/cudaq/realtime/daemon/bridge/hololink/hololink_doca_transport_ctx.h
+++ b/realtime/include/cudaq/realtime/daemon/bridge/hololink/hololink_doca_transport_ctx.h
@@ -19,10 +19,10 @@ extern "C" {
 /// Packed by the Hololink bridge layer and passed as the opaque
 /// transport_ctx pointer through the transport-agnostic dispatcher API.
 typedef struct {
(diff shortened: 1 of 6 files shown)


1tnguyen (approved):
LGTM 👍

PR #4114, opened 2026-03-07
Fix potential seg fault in evolve_async

### LINKED ISSUE #3678: Seg fault in test_evolve_simulators

sacpis:
In evolve_internal.h, evolve_async called `p.set_value(evolve(...))` and then called `platform.set_noise(nullptr)`. Here the background thread held a reference to the platform and was executing `set_noise(nullptr)` even when `.get` returned. When the successive test's `set_target()` call triggered `platformQPUs.clear()`, it tried to execute even when the thread with `platform.set_noise(nullptr)` from the previous test was still running. That caused a segfault.

Fixes #3678. 

Also fixes the [deployment failure in main](https://github.com/NVIDIA/cuda-quantum/actions/runs/22787780036/job/66113787006#step:9:600).


taalexander (approved):
LGTM, I was running into similar classes of error with #4085. Thank you for fixing, I was running into this.

PR #4115, opened 2026-03-08
[macOS] Wheel publishing fixes

taalexander:
#4025 was preemptively merged while I was still working on verifying the macOS publishing pipelines. Unfortunately there were still a few latent bugs (wheels were missing their version) that we now try and fix.


bettinaheim:
@taalexander I'll pull this in as is as well - if this doesn't full resolve it, we'll see it in the next RC build and can follow up.

PR #4121, opened 2026-03-09
Fix divisive clustering notebook

sacpis:
Fix divisive clustering notebook by implementing iterative approach for getting linkage matrix.

Fixes the bug (5961748) filed by QA about the notebook crash with recursion error.


sacpis:
> /ok to test 65dc1a6

docs/sphinx/applications/python/divisive_clustering_coresets.ipynb
```diff
@@ -743,45 +743,31 @@
...
-      "q2 : ───────────────────────────┤ x ├┤ rz(0.0002127) ├┤ x ├──●──»\n",
-      "                                ╰───╯╰───────────────╯╰───╯╭─┴─╮»\n",
-      "q3 : ──────────────────────────────────────────────────────┤ x ├»\n",
-      "                                                           ╰───╯»\n",
-      "q4 : ───────────────────────────────────────────────────────────»\n",
-      "                                                                »\n",
+      "                                                                             »\n",
+      "q0 : ──●──────────────────●──────────────────────────────────────────────────»\n",
+      "     ╭─┴─╮╭────────────╮╭─┴─╮                                                »\n",
+      "q1 : ┤ x ├┤ rz(0.8218) ├┤ x ├──●──────────────────●──────────────────────────»\n",
```
bettinaheim: Why did the angles here change so much?

docs/sphinx/applications/python/divisive_clustering_src/divisive_clustering.py
```diff
@@ -1008,33 +1008,47 @@ def plot_clusters(
...
+                self.linkage_matrix.append([
+                    index1,
+                    index2,
+                    self.distance(index1) + self.distance(index2),
+                    self.cluster_len(index1) + self.cluster_len(index2),
+                ])
+                result_stack.append(
+                    len(self.linkage_matrix) - 1 + len(self.coreset_data))
+
+        return result_stack[0]
```
bettinaheim, line 1051: This logic is rather cryptic to read. Why is it always index zero of result_stack here? What about the other values?

sacpis: This function should return just one value (`len(self.linkage_matrix) - 1 + len(self.coreset_data)`) and this is what the result stack of index 0 contains. So returning the same.

docs/sphinx/applications/python/divisive_clustering_coresets.ipynb
sacpis: Here the rx gate is incorrectly applied. The second argument should be a qubit, instead a qvector is passed. CUDA-Q compiler has mishandled this case giving the wrong output.

sacpis: Fixed in [ca583bf](https://github.com/NVIDIA/cuda-quantum/pull/4121/commits/ca583bf0706047bfdd89c8b758fd50e7dfb415ae).

@mawolf2023 for visibility.

docs/sphinx/applications/python/divisive_clustering_src/main_divisive_clustering.py
```diff
@@ -179,7 +179,8 @@ def kernel(thetas: list[float], number_of_qubits: int, circuit_depth: int):
                 rz(2.0 * thetas[layer], qubits[(qubit + 1) % number_of_qubits])
                 cx(qubits[qubit], qubits[(qubit + 1) % number_of_qubits])
 
-            rx(2.0 * thetas[layer + layers], qubits)
+            for qubit in range(number_of_qubits):
+                rx(2.0 * thetas[layer + layers], qubits[qubit])
```
1tnguyen: Just wondering why we need this change?

sacpis: Addressed in 597aa57.

PR #4124, opened 2026-03-09
[ci] [quantinuum] Move the test that invokes `run` into the next-gen directory

khalatepradnya:
Fix for the error: 
``` 
 loc("targettests/quantinuum/../execution/conditional_run.cpp":18:3)
 what():  `run` is not yet supported on this target. 
```


1tnguyen (approved):
👍

PR #4127, opened 2026-03-09
[core] Move the delay measurements pass

schweitzpgi:
The pass is part of codegen and belongs in that library. Move it there. Also, update the code to use modern MLIR practices.

### DIFF
--- a/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/include/cudaq/Optimizer/Transforms/Passes.td
@@ -25,7 +25,8 @@ def AddMeasurements : Pass<"add-measurements", "mlir::func::FuncOp"> {
 
   }];
   
-  let dependentDialects = ["mlir::cf::ControlFlowDialect"];
+  let dependentDialects = ["quake::QuakeDialect",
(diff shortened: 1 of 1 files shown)


include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -277,38 +277,29 @@ def ResourceCountPreprocess : Pass<"resource-count-preprocess", "mlir::func::Fun
       }
     ```
 
-    The initial `h(q[0])` will always be run once, so we can pre count it and remove it.
-    Similarly, it is trivial to see that the loop will run 10 times, so we can pre count
-    `y(q[i])` ten times as well. However, because `x(q[i])` is gated by `mz(q[i])`, it is
-    not invariant, so it will remain and be counted during simulation.
+    The initial `h(q[0])` will always be run once, so we can pre count it and
```
sacpis: nit: pre-count? or precount?

schweitzpgi: Updated in 6c091ec64

PR #4130, opened 2026-03-10
Using 2D grid in cudaKronprod for state initialization

sacpis:
Using 2D grid in cudaKronprod to parallelize over both state dimensions.

Fixes QA bug 5961696.

The bug is in `kronprod` which calls `cudaKronprod ` as a 1D grid. The kernel uses `blockIdx.y / gridDim.y` to go over the user state (for n = 30 qubits, it will be 1 billion elments), which is tsize2 parameter. Now, when gridDim.y = 1, only the first block ran. It went over all 2^n elements sequentially. Other blocks just launched and exited.

The 2D grid shows that there are 65535 blocks each handling 4 chunks in parallel.

Used the code below

```
import cudaq
import numpy as np

cudaq.set_target("nvidia")

@cudaq.kernel
def kernel(vec : cudaq.State):
    p = cudaq.qubit()
    q = cudaq.qvector(vec)

    mz(p)
    mz(q)

n = 30
v = np.zeros(2**n, dtype=cudaq.complex())
v[-1] = 1.
state = cudaq.State.from_data(v)

print(cudaq.sample(kernel, state))
```

This change now executes the kernel within 1 minute.

### DIFF
--- a/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
+++ b/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
@@ -89,17 +89,16 @@ __global__ void cudaKronprod(size_t tsize1, const CudaDataType *arr1,
 #pragma pop
 
 template <typename CudaDataType>
-void kronprod(uint32_t /*n_blocks*/, int32_t threads_per_block,
+void kronprod(uint32_t maxGridDimY, int32_t threads_per_block,
               size_t tsize1, const void *arr1,
               size_t tsize2, const void *arr2,
               void *arr0) {
-  constexpr uint32_t kMaxGridDimY = 65535;
   uint32_t n_blocks_x =
       (static_cast<uint32_t>(tsize1) + threads_per_block - 1) / threads_per_block;
   uint32_t n_blocks_y = static_cast<uint32_t>(
       std::min((tsize2 + static_cast<size_t>(threads_per_block) - 1) /
                    static_cast<size_t>(threads_per_block),
-               static_cast<size_t>(kMaxGridDimY)));
+               static_cast<size_t>(maxGridDimY)));
   dim3 grid(n_blocks_x, n_blocks_y);
   cudaKronprod<<<grid, threads_per_block>>>(
     tsize1, reinterpret_cast<const CudaDataType *>(arr1),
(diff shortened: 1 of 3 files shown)


runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cu
```diff
@@ -89,12 +89,20 @@ __global__ void cudaKronprod(size_t tsize1, const CudaDataType *arr1,
...
 template <typename CudaDataType>
-void kronprod(uint32_t n_blocks, int32_t threads_per_block,
+void kronprod(uint32_t /*n_blocks*/, int32_t threads_per_block,
               size_t tsize1, const void *arr1,
-              size_t tsize2, const void *arr2, 
+              size_t tsize2, const void *arr2,
               void *arr0) {
-  cudaKronprod<<<n_blocks, threads_per_block>>>(
-    tsize1, reinterpret_cast<const CudaDataType *>(arr1), 
+  constexpr uint32_t kMaxGridDimY = 65535;
```
schweitzpgi: Should this always be this magic constant, `std::numeric_limits<std::int16_t>::max()`?

schweitzpgi: My corollary question is: what was `n_blocks` really for?

sacpis: The original code with `n_blocks` was trying to cover the total work (tsize1 * tsize2) with a 1D grid. Actually, the kernel's logic uses `blockIdx.x` for tsize1 and `blockIdx.y` for tsize2. With 1D grid, `blockIdx.y` was 0, so `blockIdx.x` went through 0 to tsize1 * tsize2 / 256. The kernel exited if blockIdx * threads_per_block >= tsize1, which was true for all blocks except the first one.

So n_blocks was flattened to a 1D grid, but the kernel expected a 2D grid.

sacpis: > Should this always be this magic constant, std::numeric_limits<std::int16_t>::max()?

65535 is not a std::numeric_limits<std::int16_t>::max(). This is a hardware property. On all new Nvidia GPUs this is 65535. Actually I can use cudaDeviceProp to fetch this number during initialization.

PR #4134, opened 2026-03-10
[Macos] Additional release documentation updates

taalexander:
Adds:
- Python pypi readme instructions with CI tie in
- Some minor updates to the quick start guide


docs/sphinx/using/quick_start.rst
```diff
@@ -36,8 +36,8 @@ Install CUDA-Q
 
 .. note::
 
-    If you are using CUDA 12.8 on Blackwell, you may need to install additional
-    dependencies to use the python wheels.
+    Linux only: If you are using CUDA 12.8 on Blackwell, you may need to install
```
efratshabtai: I think 'Linux only' here is redundant since it is mentioned a bit later that Linux only is supported on GPU

taalexander: Addressed in e321aed652

PR #4138, opened 2026-03-11
Throw exception when qvector state init exceeds 10 qubits

sacpis:
Throw exception when qvector state init exceeds 10 qubits

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -54,8 +54,11 @@ ALLOWED_TYPES_IN_A_DATACLASS = [int, float, bool, qview]
 
 def _get_qualified_name(node) -> str | None:
     """Return the dotted name of a decorator AST node (e.g. "cudaq.kernel"),
-    stripping through any `ast.Call` wrapper.  Returns `None` for unrecognized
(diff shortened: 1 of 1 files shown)


python/cudaq/kernel/ast_bridge.py
```diff
@@ -3403,14 +3403,24 @@ def check_vector_init():
...
                                     size = len(listScalar)
                                     numQubits = np.log2(size)
                                     if not numQubits.is_integer():
                                         self.emitFatalError(
                                             "Invalid input state size for "
                                             "qvector init (not a power of 2)",
                                             node)
+                                    if int(numQubits) > 10:
+                                        self.emitFatalError(
+                                            f"State vector initialization with more than 10 qubits is not supported. Requested {int(numQubits)} qubits.",
```
schweitzpgi, line 3357: 80 columns?

sacpis: I did run yapf. Let me check it for comment.

sacpis: Used docformatter. Fixed in f517200.

1tnguyen (approved):
👍

PR #4140, opened 2026-03-11
[Testing] - Fix test coverage

mitchdz:
# Fix codecov workflow

Codecov has been silently failing, thus creating no reports. This PR fixes that so now PRs will have a comment indicating the code coverage percentage.

Example message from codecov: https://github.com/NVIDIA/cuda-quantum/pull/4140#issuecomment-4043038427

### DIFF
--- a/requirements-tests-backend.txt
+++ b/requirements-tests-backend.txt
@@ -1,5 +1,5 @@
 # ============================================================================ #
-# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                   #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
 # All rights reserved.                                                         #
 #                                                                              #
 # This source code and the accompanying materials are made available under     #


scripts/generate_cc.sh
```diff
@@ -178,20 +180,25 @@ if $gen_cpp_coverage; then
 fi
 
 if $gen_py_coverage; then
+    venv_dir=${repo_root}/build/venv-coverage
+    python3 -mpip install iqm_client==16.1 -vvv --break-system-packages
+    python3 -m venv "$venv_dir"
```
schweitzpgi: Should these be `python` (without the 3) as well?

mitchdz: These should include the `3` since we are creating the venv. We can remove the `3` once we are inside a venv since it symlinks.

mitchdz: However, that line does have a glaring typo that I need to fix.

scripts/generate_cc.sh
```diff
@@ -93,7 +93,7 @@ if $gen_cpp_coverage; then
     use_llvm_cov=true
 
     # Run tests (C++ Unittests)
-    python3 -m pip install iqm-client==28.0.0
+    python3 -m pip install iqm-client==28.0.0 --break-system-packages
```
khalatepradnya: Q: What about `qio~=0.1.33`?

scripts/generate_cc.sh
```diff
@@ -178,20 +180,29 @@ if $gen_cpp_coverage; then
 fi
 
 if $gen_py_coverage; then
+    PY_VER=$(python3 -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")
+    apt install -y python${PY_VER}-venv
+ 
+    # Needs to be installed outside of venv
+    python3 -m pip install iqm-client==28.0.0 --break-system-packages
```
khalatepradnya: Same here... I wonder if such dependencies should be driven from a requirements file so we don't have to edit multiple places...

scripts/generate_cc.sh
mitchdz: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4140/commits/d891bf649456d8ca7f710c67f1bac9c81981d852

scripts/generate_cc.sh
mitchdz: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4140/commits/d891bf649456d8ca7f710c67f1bac9c81981d852

requirements-tests-backend.txt
```diff
@@ -0,0 +1,13 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
```
khalatepradnya: nit: New file can just have the year it was first created, no?

requirements-tests-backend.txt
```diff
@@ -0,0 +1,13 @@
...
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# Backend dependencies required for running tests against hardware provider
+# mock servers (e.g., IQM, Scaleway). Pinned here so that all CI workflows
+# and coverage scripts reference a single source of truth.
+iqm-client==28.0.0
+qio~=0.1.33
```
khalatepradnya, line 13: Thank you for this update.
Q: How do we distinguish between what goes here, v/s what goes in https://github.com/NVIDIA/cuda-quantum/blob/main/requirements-dev.txt?

requirements-tests-backend.txt
mitchdz: Completely true. I'll fix that in [f280148](https://github.com/NVIDIA/cuda-quantum/pull/4140/commits/f280148b7d44bf3435cb1e070d3744c2acb0b4ee)

requirements-tests-backend.txt
mitchdz: I think the description is a good guide for that. Generally this new file can be used for backend testing and tooling. It doesn't make sense to ship `iqm-client` in our image as an example, so it falls in this file which is necessary for testing, but not for actually running the code.

requirements-tests-backend.txt
```diff
@@ -0,0 +1,13 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                   #
```
khalatepradnya: nit: alignment of `#`

requirements-tests-backend.txt
khalatepradnya: not sure I follow... wouldn't you need these packages to run against the real hardware of that provider?

requirements-tests-backend.txt
mitchdz: Thanks. Fixed.

requirements-tests-backend.txt
mitchdz: Do we want to by default ship all of these dependencies?

khalatepradnya: I am not sure. Tagging @efratshabtai for guidance.

bettinaheim: Generally, we want to be careful about adding dependencies to our packages. Anything that is used for testing only (iqm-client) should certainly go to the test requirement and not be part of any of our images. For anything else, we usually want it to be an optional dependency - i.e. the user would be expected to install that dependency and our shipped packages don't formally depend on it. In the end, we need to discuss case by case.

PR #4144, opened 2026-03-11
Several python notebooks fixes

sacpis:
Fixes several notebook validation failures

- notebook_validation.py: Incerasing the timeout to 900 seconds as almost all notebooks cells had a timeout with 600 seconds (Have noticed that these notebooks cells finish execution within 630 - 700 seconds)
- hybrid_quantum_neural_networks.ipynb: Adding torch.cuda.is_available() guard
- skqd_src/pre_and_postprocessing.py: Explcitly freed the boolean mask and computed transformed_ints in 4096-row chunks (referred Google for the part B solution)
- qsci.ipynb: Using np.bytes_ as per suggested in the error message
- unitary_compilation_diffusion_models.ipynb: Adding huggingface_hub.login(token=os.environ.get('HF_TOKEN')) so that authenticated request are made with HF_TOKEN when available

### DIFF
--- a/.github/pre-commit/spelling_allowlist.txt
+++ b/.github/pre-commit/spelling_allowlist.txt
@@ -90,6 +90,7 @@ OpenSUSE
 Ou
 POSIX
 PSIRT
+PTSBE
 Pasqal
 Pauli
 Paulis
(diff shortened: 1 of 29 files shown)


1tnguyen (approved):
LGTM 👍

docs/sphinx/applications/python/unitary_compilation_diffusion_models.ipynb
```diff
@@ -108,6 +108,11 @@
     "\n",
     "os.environ['HF_HUB_DISABLE_PROGRESS_BARS'] = '1'\n",
     "\n",
+    "import huggingface_hub\n",
+    "hf_token = os.environ.get('HF_TOKEN')\n",
+    "if hf_token:\n",
+    "    huggingface_hub.login(token=hf_token, add_to_git_credential=False)\n",
```
1tnguyen, line 115: nit: maybe we could consider adding a log message here that users may hit the rate limit if not providing a token.

sacpis: Fixed in c9d748e.

PR #4151, opened 2026-03-12
Adding guard to handle numQubits > 10 as a kernel parameter

sacpis:
Adding guard to handle numQubits > 10 as a kernel parameter

### DIFF
--- a/.github/workflows/test_in_devenv.yml
+++ b/.github/workflows/test_in_devenv.yml
@@ -243,7 +243,8 @@ jobs:
             fi
             python -m pip install pytest
             python -m pytest -v --durations=0  python/tests/ \
-              --ignore python/tests/backends
+              --ignore python/tests/backends \
+              --ignore python/tests/contrib
             pytest_status=$?
             if [ ! $pytest_status -eq 0 ]; then
               echo "::error file=test_in_devenv.yml::Python tests failed with status $pytest_status."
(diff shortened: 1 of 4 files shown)


python/tests/kernel/test_kernel_qvector_state_init.py
```diff
@@ -534,3 +534,20 @@ def kernel():
...
+    v = [0.] * (2**n)
+    v[-1] = 1.
+
+    @cudaq.kernel
+    def kernel(state: list[complex]):
+        q = cudaq.qvector(state)
+
+    with pytest.raises(RuntimeError) as e:
+        cudaq.sample(kernel, v)
+    assert 'State vector initialization with more than 10 qubits is not supported. Requested 11 qubits.' in repr(
```
schweitzpgi: nit: use parenthesis and split the string literal for formatting.

sacpis:
> Thanks, Sachin. Are you going to back out the static compile-time check in the bridge?

Should we keep that check in the bridge as well? In case user tries to create a state using the complex array with more than 10 qubits.

sacpis:
> > Thanks, Sachin. Are you going to back out the static compile-time check in the bridge?
> 
> Should we keep that check in the bridge as well? In case user tries to create a state using the complex array with more than 10 qubits.

Yes, makes sense. Will remove it.

schweitzpgi:
> > Thanks, Sachin. Are you going to back out the static compile-time check in the bridge?
> 
> Should we keep that check in the bridge as well? In case user tries to create a state using the complex array with more than 10 qubits.

This check will catch both cases, right?

sacpis:
> > > Thanks, Sachin. Are you going to back out the static compile-time check in the bridge?
> > 
> > 
> > Should we keep that check in the bridge as well? In case user tries to create a state using the complex array with more than 10 qubits.
> 
> This check will catch both cases, right?

Yes.

python/tests/kernel/test_kernel_qvector_state_init.py
sacpis: Fixed in 666425b.

PR #4162, opened 2026-03-13
[core] Updates for value semantics

schweitzpgi:
- remove the quake.sink_cable op
- tidy up documentation so diagnostics are readable
- add code to convert a simple test in value semantics to full QIR
- add test for codegen in previous bullet
- update quake.concat operation
- improve the factor quantum allocations pass
- add cable rough-in pass and test for same


1tnguyen (approved):
LGTM 👍

schweitzpgi:
Opening #4179 for follow-up work.

PR #4166, opened 2026-03-16
[test] Clean up execution tests.

schweitzpgi:
Do some cleanup on these tests.

  - eliminate purely redundant command lines, not needed
  - add missing code, like "return 0;" in main()
  - remove FileCheck use from trivial tests that didn't need it
  - other misc.

### DIFF
--- a/targettests/execution/if_jit.cpp
+++ b/targettests/execution/if_jit.cpp
@@ -17,7 +17,6 @@
 // RUN: nvq++ --target quantinuum --emulate %s -o %t && %t | FileCheck %s
 // RUN: if %braket_avail; then nvq++ --target braket --emulate %s -o %t && %t | FileCheck %s; fi
 // RUN: if %qci_avail; then nvq++ --target qci --emulate %s -o %t && %t | FileCheck %s; fi
-// RUN: nvq++ --enable-mlir %s -o %t
 // clang-format on
 
 #include <cudaq.h>
--- a/targettests/execution/predefined_lib_mode.cpp
+++ b/targettests/execution/predefined_lib_mode.cpp
@@ -5,10 +5,8 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
-// clang-format off
+
 // RUN: nvq++ -DTEST_DEF -DMY_VAR=\"CUDAQ\" %s -o %t && %t | FileCheck %s
-// RUN: nvq++ --enable-mlir -DTEST_DEF -DMY_VAR=\"CUDAQ\" %s -o %t
-// clang-format on
 
 #include <iostream>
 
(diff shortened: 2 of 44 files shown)


targettests/execution/if_jit.cpp
```diff
@@ -17,7 +17,6 @@
 // RUN: nvq++ --target quantinuum --emulate %s -o %t && %t | FileCheck %s
 // RUN: if %braket_avail; then nvq++ --target braket --emulate %s -o %t && %t | FileCheck %s; fi
 // RUN: if %qci_avail; then nvq++ --target qci --emulate %s -o %t && %t | FileCheck %s; fi
-// RUN: nvq++ --enable-mlir %s -o %t
```
khalatepradnya, line 20: Perhaps some of these tests were from the time when library mode was the default.
Q: Do some of these now need explicit `--library-mode` check?

targettests/execution/predefined_lib_mode.cpp
```diff
@@ -5,10 +5,8 @@
  * This source code and the accompanying materials are made available under    *
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
-// clang-format off
+
 // RUN: nvq++ -DTEST_DEF -DMY_VAR=\"CUDAQ\" %s -o %t && %t | FileCheck %s
-// RUN: nvq++ --enable-mlir -DTEST_DEF -DMY_VAR=\"CUDAQ\" %s -o %t
```
khalatepradnya, line 10: Just going by the name of the test, this should have the `--library-mode` run command also?

targettests/execution/qir_op1_after_measure.cpp
```diff
@@ -7,7 +7,6 @@
  ******************************************************************************/
 
 // RUN: nvq++ -v %s -o %t --target oqc --emulate && %t 2>&1 | FileCheck %s
-// RUN: nvq++ --enable-mlir %s -o %t
```
khalatepradnya, line 10: Q: Doesn't the second command undergo a different compilation and execution path than emulation?
(Okay to remove if we don't get much out of the test itself).

targettests/execution/if_jit.cpp
schweitzpgi: I don't think so. library-mode is inching towards deprecation.

targettests/execution/qir_op1_after_measure.cpp
schweitzpgi: I don't think this is useful. The first compilation uses the C++ bridge. It looks like left over junk to me.

targettests/execution/predefined_lib_mode.cpp
schweitzpgi: Opening as an issue. I don't want to change the test behavior in this PR.

PR #4169, opened 2026-03-16
[Realtime] Fix unified dispatch for iGPU

cketcham2333:


realtime/include/cudaq/realtime/daemon/bridge/hololink/hololink_wrapper.h
```diff
@@ -111,6 +111,19 @@ size_t hololink_get_page_size(hololink_transceiver_t handle);
 /** Get the number of pages (slots) configured for this transceiver. */
 unsigned hololink_get_num_pages(hololink_transceiver_t handle);
 
+/**
+ * Pre-post receive WQEs and initialize the send WQE template.
+ *
+ * Calls GpuRoceTransceiverPrepareKernel with the transceiver's GPU QP,
+ * frame_size, and RX ring rkey.  Use this when start() didn't call the
```
bettinaheim:
```suggestion
 * frame_size, and RX ring `rkey`.  Use this when start() didn't call the
```

realtime/lib/daemon/bridge/hololink/unified_dispatch_kernel.cu
```diff
@@ -152,21 +166,34 @@ __global__ void hololink_unified_dispatch_kernel(
...
     }
 
-    // Send response via Hololink's BlueFlame helper
+    // Repost the receive WQE BEFORE sending the response.  On iGPU the
+    // send goes through the CPU proxy which may batch doorbell writes,
+    // delaying return to the kernel.  Reposting first ensures the NIC
+    // has a fresh WQE before any such delay.
+    sq_wqe_idx++;
+    repost_receive(qp, sq_wqe_idx);
+    cq_ticket = sq_wqe_idx;
```
1tnguyen: Just a question: so we need to move this `repost_receive` before sending for both iGPU and dGPU?

schweitzpgi:
nit: Can we get a more descriptive title for this one?

realtime/lib/daemon/bridge/hololink/unified_dispatch_kernel.cu
cketcham2333: The reorder is only strictly necessary for iGPU — on that path, send goes through the CPU proxy which may batch doorbell writes, so we want the receive WQE reposted before any such delay. On dGPU, send_bf writes directly to the NIC and returns immediately, so the order doesn't matter. I applied it to both paths to keep the code simpler rather than duplicating the send/repost logic in each branch. It's harmless on dGPU since the NIC processes send and receive WQEs independently. Happy to split it into separate paths if you'd prefer minimizing the dGPU behavioral change.

PR #4173, opened 2026-03-16
Support qvector arguments in custom operations

### LINKED ISSUE #2452: Error not thrown when broadcasting custom unitary to a qvector.
### Required prerequisites

### Describe the bug

When attempting to broadcast a custom unitary to a qvector, an error is not thrown. In the reproducer below, the kernel produces an incorrect result when sampled.


```
cudaq.register_operation("custom_i", np.array([1, 0, 0, 1]))                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                   
@cudaq.kernel                                                                                                                                                                                                                                                                      
def bell():                                                                                                                                                                                                                                                                        
    qubits = cudaq.qvector(2)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
    custom_i(qubits) 
```

### Expected behavior

An error is thrown when attempting to broadcast a custom unitary to a qvector.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 


### Suggestions

_No response_

### LINKED ISSUE #4120: Feature Request: Allow qvector unpacking (*) or direct qvector input for custom operations (cudaq.register_operation)
### Required prerequisites

### Describe the feature

When defining custom quantum operations from unitary matrices using `cudaq.register_operation`, the generated operation strictly requires discrete cudaq.qubit arguments.

When working with a cudaq.qvector, there is currently no way to dynamically unpack the register into the custom operation call. Python's standard argument unpacking operator (*) is not supported by  @cudaq.kernel definitions, and passing the `qvector` directly throws an error.

```import cudaq
import numpy as np

# 1. Define a unitary matrix for a custom operation (e.g., a 2-qubit operator)
custom_matrix = np.eye(4) # Using Identity for simplicity in this MWE
cudaq.register_operation("my_custom_op", custom_matrix)

@cudaq.kernel
def main_kernel():
    # Allocate a vector of 2 qubits
    qvec = cudaq.qvector(2)
    
    # ---------------------------------------------------------
    # DESIRED BEHAVIOR 1: Unpacking (Currently fails to parse/compile)
    # ---------------------------------------------------------
    # my_custom_op(*qvec) 
    
    # ---------------------------------------------------------
    # DESIRED BEHAVIOR 2: Passing the qvector directly 
    # ---------------------------------------------------------
    # my_custom_op(qvec)
    
    # ---------------------------------------------------------
    # CURRENT REQUIRED WORKAROUND: Hardcoding indices
    # ---------------------------------------------------------
    my_custom_op(qvec[0], qvec[1])

# Test the compilation
cudaq.sample(main_kernel)```

### PR DESCRIPTION
**Pull request by sacpis**
Support qvector arguments in custom operations

Custom operations registered via cudaq.register_operation required individual qubit refs to be passed explicitly.
This PR adds support for the following calls as well

- custom_op(qvec)
- custom_op(*qvec)

Adding unittests which covers all three calls above.

Fixes #4120 
Fixes #2452 
Fixes (partially, first case) #2484

### CODE REVIEW

File: python/tests/kernel/test_custom_operation.py [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+
+def test_starred_qvector():
+    """custom_swap(*qvec)"""
+
+    @cudaq.kernel
+    def kernel():
+        qvec = cudaq.qvector(2)
+        x(qvec[0])
+        custom_swap(*qvec)
+
```
1tnguyen: nit: I think we can add a couple of more test cases for more coverage:

1. Mixing `*` and individual ref should be supported

e.g., 
```
qvec = cudaq.qvector(1)
qbit = cudaq.qubit()
custom_swap(*qvec, qbit)
```

2. Mixing qvec (without `*`) and qref shouldn't be allowed (I think), as this may be the case for general gates. 

3. Error conditions, e.g., unknown veq size (size is an argument) or incorrect number of target qubits.

sacpis: >>incorrect number of target qubits.
 
 Will handle this in a follow up PR for ctrl gates.

sacpis: Fixed in [1d1c40f](https://github.com/NVIDIA/cuda-quantum/pull/4173/commits/1d1c40f3c0d4f953c4b8ac706d82b6a24290e10c).

python/tests/kernel/test_custom_operation.py
```diff
@@ -0,0 +1,146 @@
...
+
+def test_unknown_veq_size_incorrect_count():
+    """custom_swap(*qvec), qvec has more qubits than the operation requires."""
+
+    @cudaq.kernel
+    def kernel(n: int):
+        qvec = cudaq.qvector(n)
+        x(qvec[0])
+        custom_swap(*qvec)
```
1tnguyen: This test doesn't seem right: `custom_swap` expects 2 qubits, but 3 were provided (`qvec` has size 3) and we didn't check for an error?

sacpis: Yeah, I thought we should consume what is available in the `qvec`. But will throw a runtime error here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2926,14 +2961,8 @@ def bodyBuilder(iterVar):
...
-                targets = self.__groupValues(node.args,
-                                             [(numTargets, numTargets)])
-
-                for i, t in enumerate(targets):
-                    if not quake.RefType.isinstance(t.type):
-                        self.emitFatalError(
-                            f'invalid target operand {i}, broadcasting is not supported on custom operations.'
-                        )
+                targets = self.__expandCustomOpTargets(node.args, numTargets,
+                                                       node)
```
khalatepradnya, line 2999: Don't we need the same logic for control and adjoint variants of custom ops?

python/tests/custom/test_custom_operations.py
```diff
@@ -232,18 +232,6 @@ def bell():
 
 
 def test_bug_2452():
-    cudaq.register_operation("custom_i", np.array([1, 0, 0, 1]))
```
khalatepradnya, line 235: Q: Why delete this test altogether? If it no longer throws, perhaps update with expected result?

python/tests/custom/test_custom_operations.py
```diff
@@ -256,20 +244,6 @@ def kernel2():
     counts = cudaq.sample(kernel2)
     assert len(counts) == 1 and '111' in counts
 
-    cudaq.register_operation(
-        "custom_cz", np.array([1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0,
```
khalatepradnya, line 260: Same here.

python/tests/kernel/test_custom_operation.py
khalatepradnya, line 1: nit: Why not combine these under `python/tests/custom/test_custom_operations.py` ? So all are in the same place?

python/tests/kernel/test_custom_operation.py
```diff
@@ -0,0 +1,146 @@
...
+
+    counts = cudaq.sample(kernel, 3)
+    assert counts.most_probable() == "010"
```
khalatepradnya: Thank you for expanding tests! Can you add more to cover control and adjoint variants and builder mode too?

khalatepradnya (commented):
Thanks, @sacpis for addressing this long-awaited request! 
I suggest expanding the implementation to cover more scenarios.

python/cudaq/kernel/ast_bridge.py
sacpis: Yes, definitely. That's why I commented it [here](https://github.com/NVIDIA/cuda-quantum/pull/4173#discussion_r2944255549).

python/tests/custom/test_custom_operations.py
sacpis: Actually have added tests to cover these cases. So thought of getting rid of these old ones.

But if you prefer to update these tests, will definitely do.

python/tests/kernel/test_custom_operation.py
sacpis: Sure. Nice suggestion. Thank you.

python/tests/kernel/test_custom_operation.py
sacpis: Yes, when I implement for control and adjoint in a follow up PR, will do so.

khalatepradnya: Cover builder mode as well... (If I recall correctly, it follows a different logic for controlled variant )

sacpis: Yes, it does.

schweitzpgi:
I haven't had a chance to look at this @sacpis but wanted to drop a heads-up comment.

The builtins are overloaded in multiple dimensions. The custom ops should probably behave similarly (somehow).

```c++
   qubit q;
   qvector qs(n), ts(m);
   
   op(q);      // gate on single qubit
   op(qs);     // broadcast over several qubits;
   op(qs, q);  // controls on single qubit
   
   // Theoretically, we could even have both, but I don't think we do (yet).
   op(qs, ts);  // controls over a broadcast
```

khalatepradnya:
> I haven't had a chance to look at this @sacpis but wanted to drop a heads-up comment.
> 
> The builtins are overloaded in multiple dimensions. The custom ops should probably behave similarly (somehow).
> 
> ```c++
>    qubit q;
>    qvector qs(n), ts(m);
>    
>    op(q);      // gate on single qubit
>    op(qs);     // broadcast over several qubits;
>    op(qs, q);  // controls on single qubit
>    
>    // Theoretically, we could even have both, but I don't think we do (yet).
>    op(qs, ts);  // controls over a broadcast
> ```

I think for ` op(qs);     // broadcast over several qubits;` is okay on single qubit operations, but for 2+ qubit operations, doesn't the question arise of how to group them ?

python/cudaq/kernel/ast_bridge.py
sacpis: Sorry a little different. Because there will have to deal with source and target.

python/tests/kernel/test_custom_operation.py
sacpis: Fixed.

python/tests/kernel/test_custom_operation.py
sacpis: Fixed in [2d9988d](https://github.com/NVIDIA/cuda-quantum/pull/4173/commits/2d9988d51ca27c1d2ba0bb0a5e2633ddb0a382fd).

python/tests/custom/test_custom_operations.py
sacpis: Here we are fixing this broadcast issue listed in #2452. Throwing an error as now.

```
Traceback (most recent call last):
  File "/workspaces/cuda-quantum/test.py", line 11, in <module>
    result = cudaq.sample(bell)
             ^^^^^^^^^^^^^^^^^^
  File "/usr/local/cudaq/cudaq/runtime/sample.py", line 178, in sample
    kernel(*args)
  File "/usr/local/cudaq/cudaq/kernel/kernel_decorator.py", line 611, in __call__
    result = cudaq_runtime.marshal_and_launch_module(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: custom operation requires 1 qubit target(s), but 2 were provided
```

The `custom_i` unitary operation is incorrect as it is a 1-qubit gate op.
Actually we should remove this test as that issue will be marked as fixed with this change.

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -337,6 +337,10 @@ static constexpr IntrinsicCode intrinsicTable[] = {
   func.func private @__nvqpp_cudaq_state_numberOfQubits(%p : !cc.ptr<!quake.state>) -> i64
 )#"},
 
+    {"__nvqpp_customop_size_error", {}, R"#(
+  func.func private @__nvqpp_customop_size_error(i64, i64)
+)#"},
```
1tnguyen, line 342: Just a suggestion: the `__quantum__qis__trap` intrinsic that also serves this purpose (throwing runtime errors).
We can introduce a new trap code for this, but it wouldn't be able to log the expected vs. given number of qubits in the error message :)

sacpis: Yeah, actually want to inform users of the actual error which will be useful to them.

python/tests/custom/test_custom_operations.py
khalatepradnya: Okay to remove if covered elsewhere.

sacpis: This [test case](https://github.com/NVIDIA/cuda-quantum/pull/4173/changes#diff-5d561e474ab832afd404d67efb4470385048a130b7421b51c034cb8daa705cedR339) covers the above error message.

sacpis:
Thanks @schweitzpgi @khalatepradnya for the heads-up. I will consider these options in a follow up PR (for control, adjoint and builder). Will wait for your thumbs up on this one.

lib/Optimizer/Builder/Intrinsics.cpp
sacpis: Fixed in [d698591](https://github.com/NVIDIA/cuda-quantum/pull/4173/commits/d69859170fff088a2a5e4c6b64777888ffe84d56).

lib/Optimizer/Builder/Intrinsics.cpp
```diff
@@ -438,7 +438,7 @@ static constexpr IntrinsicCode intrinsicTable[] = {
 )#"},
 
     {cudaq::opt::QISTrap, {}, R"#(
-  func.func private @__quantum__qis__trap(i64)
+  func.func private @__quantum__qis__trap(i64, !cc.ptr<i8>)
```
schweitzpgi: The idea for the trap is to pass an error code only. The error code will be used by the runtime to generate the appropriate diagnostic. We don't want to encode the diagnostics in the IR.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2316,6 +2316,84 @@ def find_unique_decorator_name(self, name):
...
+                    sizesMatch = arith.CmpIOp(eqPred, actualSize,
+                                              expectedSizeVal).result
+                    ifOp = cc.IfOp([], sizesMatch, [])
+                    thenBlock = Block.create_at_start(ifOp.thenRegion, [])
+                    with InsertionPoint(thenBlock):
+                        cc.ContinueOp([])
+                    elseBlock = Block.create_at_start(ifOp.elseRegion, [])
+                    with InsertionPoint(elseBlock):
+                        errMsg = (f'custom operation requires {numTargets} '
+                                  f'qubit target(s), but got ')
```
schweitzpgi: Traps should not generate dynamic messages like this. They are intended to be static like `"division by 0"`.

schweitzpgi (changes requested):

lib/Optimizer/Builder/Intrinsics.cpp
sacpis: Thanks @schweitzpgi, Will revert this commit and have a separate intrinsic to throw runtime error with the appropriate message.

@1tnguyen for viz.

python/cudaq/kernel/ast_bridge.py
sacpis: Will revert this commit. Thank you.

lib/Optimizer/Builder/Intrinsics.cpp
sacpis: Change reverted with [2a6a5f0](https://github.com/NVIDIA/cuda-quantum/pull/4173/commits/2a6a5f0203df32b578023e92d284011d52289688).

python/cudaq/kernel/ast_bridge.py
sacpis: Change reverted with [2a6a5f0](https://github.com/NVIDIA/cuda-quantum/pull/4173/commits/2a6a5f0203df32b578023e92d284011d52289688).

schweitzpgi (approved):

PR #4174, opened 2026-03-17
Handles seg fault for state initialized with 20 qubits

sacpis:
Adding a guard that raises a RuntimeError before conversion when list[complex] argument exceeds 10 qubits.

Code to repro

```
import cudaq

n = 20
v = [0.] * (2**n)
v[-1] = 1.

@cudaq.kernel
def kernel(state: list[complex]):
    q = cudaq.qvector(state)

result = cudaq.sample(kernel, v)
print(result)
```

### DIFF
--- a/lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
+++ b/lib/Optimizer/Transforms/FactorQuantumAlloc.cpp
@@ -102,7 +102,7 @@ public:
           rewriter.create<quake::MakeStruqOp>(loc, stqTy, memAllocs);
       // 3. Walk all the uses. If they are quake.get_member operations, replace
       // them with direct uses.
-      for (auto *user : allocOp->getUsers())
+      for (auto *user : llvm::make_early_inc_range(allocOp->getUsers()))
         if (auto getMem = dyn_cast<quake::GetMemberOp>(user)) {
           auto index = getMem.getIndex();
           rewriter.replaceOp(getMem, memAllocs[index]);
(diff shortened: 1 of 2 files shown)


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -654,6 +654,17 @@ def process_argument(self, arg, arg_type):
                 f"Argument has callable type but the argument ({arg}) is not "
                 f"a kernel decorator.")
 
+        # Validate size limit for list[complex] arguments used for qvector
+        # state initialization.
+        if cc.StdvecType.isinstance(arg_type):
+            toEleTy = cc.StdvecType.getElementType(arg_type)
```
1tnguyen: nit: I think a name like `eleTy` is more readable (as we don't do any type conversion here).

1tnguyen (approved):
LGTM 👍

python/cudaq/kernel/kernel_decorator.py
sacpis: Fixed in 6fbda40.

khalatepradnya:
non-blocking Q: Why is the threshold chosen to be 10 qubits?

sacpis:
> non-blocking Q: Why is the threshold chosen to be 10 qubits?

Good question. It is after discussing with @bettinaheim and @schweitzpgi, we decided the reasonable number would be 10 qubits for constructing the state directly with complex numbers. We thought this would be the safe threshold number to fit in memory. Basically, we would want to encourage users to use cudaq.State api instead.

PR #4180, opened 2026-03-18
Add realtime host dispatcher

wsttiger:
## Summary

This PR refines the host-side dispatcher backend and adds end-to-end test coverage for the GRAPH_LAUNCH dispatch path.

**Restrict host dispatcher to GRAPH_LAUNCH only** — The host loop now only dispatches `GRAPH_LAUNCH` entries; `HOST_CALL` and `DEVICE_CALL` slots are dropped (cleared and advanced). Removes the unused `dispatch_host_call` path and updates comments/headers to reflect the GRAPH_LAUNCH-only design.

**Add host dispatcher tests + external mailbox support** — New test file `test_host_dispatcher.cu` with two tests:
- *Smoke test*: starts the host loop via the C API, sends an RPC with an unknown `function_id`, and verifies the slot is silently dropped.
- *GRAPH_LAUNCH round-trip*: full end-to-end test through the C API — allocates a pinned mailbox, captures an increment graph, wires the dispatcher, sends an RPC `{0,1,2,3}`, and asserts the graph produces `{1,2,3,4}` in-place.

**New C API: `cudaq_dispatcher_set_mailbox`** — Lets callers provide a caller-managed pinned (`cudaHostAllocMapped`) mailbox. This is required for GRAPH_LAUNCH because the graph must be captured with the device-side mailbox pointer before the dispatcher starts, and the internal allocation (plain `new`) is not device-visible. When no external mailbox is provided, the C API falls back to internal allocation (backward compatible).

### DIFF
--- a/realtime/CMakeLists.txt
+++ b/realtime/CMakeLists.txt
@@ -138,5 +138,6 @@ if (CUDAQ_REALTIME_BUILD_TESTS)
   add_subdirectory(unittests)
 endif()
 
+
 # Install the LICENSE and NOTICE files
 install(FILES LICENSE NOTICE DESTINATION ${CMAKE_INSTALL_PREFIX})


bmhowe23:
> /ok to test 6e48865b2b05762d7b0ba4be1038a6c14b08fa85

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
bmhowe23, line 1: Does this really need to be a .cu file, or should it just be .cpp?

realtime/lib/daemon/dispatcher/host_dispatcher.cu
bmhowe23, line 1: Does this really need to be a .cu file, or should it just be .cpp?

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
```diff
@@ -28,17 +28,40 @@ typedef enum {
...
+  CUDAQ_TX_IN_FLIGHT = 1,
+  CUDAQ_TX_ERROR = 2,
+  CUDAQ_TX_READY = 3
+} cudaq_tx_status_t;
+
+// RPC wire-format constants (must match dispatch_kernel_launch.h).
+#define CUDAQ_RPC_MAGIC_REQUEST 0x43555152u  /* 'CUQR' */
+#define CUDAQ_RPC_MAGIC_RESPONSE 0x43555153u /* 'CUQS' */
+#define CUDAQ_RPC_HEADER_SIZE                                                  \
+  24u /* `sizeof(RPCHeader)`: 4 x uint32_t + 1 x uint64_t */
```
1tnguyen: If these constants and types (e.g., `RPCHeader`) are common, I wonder whether we should pull them out into a separate header to be included both here and `dispatch_kernel_launch.h`.

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
```diff
@@ -0,0 +1,77 @@
...
+
+#pragma once
+
+#include "cudaq/realtime/daemon/dispatcher/cudaq_realtime.h"
+
+#include <cuda_runtime.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifndef QEC_CPU_RELAX
```
1tnguyen: Q: do we need to name this macro **QEC**_CPU_RELAX? i.e., could we name it `CUDAQ_REALTIME_CPU_RELAX`?

realtime/lib/daemon/dispatcher/cudaq_realtime_api.cpp
```diff
@@ -145,12 +166,25 @@ cudaq_status_t cudaq_dispatcher_set_control(cudaq_dispatcher_t *dispatcher,
...
+      launch_fn != nullptr)
+    return CUDAQ_ERR_INVALID_ARG;
+  if (dispatcher->config.backend != CUDAQ_BACKEND_HOST_LOOP && !launch_fn)
     return CUDAQ_ERR_INVALID_ARG;
   dispatcher->launch_fn = launch_fn;
   return CUDAQ_OK;
 }
 
+cudaq_status_t cudaq_dispatcher_set_mailbox(cudaq_dispatcher_t *dispatcher,
+                                            void **h_mailbox_bank) {
```
1tnguyen, line 180: Do we need a null check for `h_mailbox_bank`?

realtime/lib/daemon/dispatcher/host_dispatcher.cu
```diff
@@ -0,0 +1,200 @@
...
+    as_atomic_u64(config->tx_flags)[current_slot].store(
+        error_val, cuda::std::memory_order_release);
+    as_atomic_u64(config->idle_mask)
+        ->fetch_or(1ULL << worker_id, cuda::std::memory_order_release);
+  } else {
+    if (config->workers[w].post_launch_fn)
+      config->workers[w].post_launch_fn(config->workers[w].post_launch_data,
+                                        data_dev, config->workers[w].stream);
+    as_atomic_u64(config->tx_flags)[current_slot].store(
+        0xEEEEEEEEEEEEEEEEULL, cuda::std::memory_order_release);
```
1tnguyen: nit: perhaps we could define this magic number (also used in `cudaq_host_ringbuffer_poll_tx_flag`) somewhere in a header file.

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
```diff
@@ -0,0 +1,171 @@
...
+  if (handle->thread.joinable())
+    handle->thread.join();
+  for (size_t i = 0; i < handle->num_workers; ++i)
+    cudaStreamDestroy(handle->workers[i].stream);
+  delete[] handle->workers;
+  delete handle->idle_mask;
+  delete[] handle->inflight_slot_tags;
+  if (handle->owns_mailbox)
+    delete[] handle->h_mailbox_bank;
+  delete handle;
```
1tnguyen: nit: perhaps we could make this cleanup procedure a helper function.

realtime/lib/daemon/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ if(CUDA_FOUND)
   target_link_libraries(cudaq-realtime 
     PUBLIC 
       CUDA::cudart_static
+    PRIVATE
+      cudaq-realtime-host-dispatch
```
1tnguyen, line 34: Q: This is a bit different from the `cudaq-realtime-dispatch`, where it's separated (not directly linked).
We might need to set `rpath` if this additional linkage is required.

realtime/lib/daemon/CMakeLists.txt
```diff
@@ -85,6 +86,37 @@ if(CUDA_FOUND)
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
   )
 
+  # ============================================================================
+  # Host-side graph dispatcher (optional, for Grace Hopper / Grace Blackwell etc.)
+  # ============================================================================
+  # Compiled with nvcc so libcu++ (<cuda/std/atomic>) works without extra
+  # include paths. Host-only code; no device code in this TU.
+  add_library(cudaq-realtime-host-dispatch SHARED
```
1tnguyen: Q: I wonder whether we should use `STATIC` to be consistent with `cudaq-realtime-dispatch` lib.

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
wsttiger: These files are .cu only because they include <cuda/std/atomic> (libcu++ system-scope atomics), which requires nvcc to resolve the include paths automatically. There's no actual device code in this TU — the CMakeLists comment already notes this. Renaming to .cpp would require manually wiring the libcu++ include paths per toolkit version, which is fragile.

realtime/lib/daemon/dispatcher/host_dispatcher.cu
wsttiger: Same as the above comment

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
wsttiger: pushed into separate header

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
wsttiger: done

realtime/lib/daemon/dispatcher/cudaq_realtime_api.cpp
wsttiger: done

realtime/lib/daemon/dispatcher/host_dispatcher.cu
wsttiger: done

realtime/lib/daemon/CMakeLists.txt
wsttiger: converted to STATIC to match cudaq-realtime-dispatch

realtime/lib/daemon/CMakeLists.txt
wsttiger: converted to STATIC to match cudaq-realtime-dispatch

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
wsttiger: done

1tnguyen:
> /ok to test [9513105](https://github.com/NVIDIA/cuda-quantum/pull/4180/commits/95131056e79e09013bae7f8c340abb8874d809ae)

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
```diff
@@ -28,17 +28,37 @@ typedef enum {
...
+// TX flag status returned by cudaq_host_ringbuffer_poll_tx_flag.
+typedef enum {
+  CUDAQ_TX_EMPTY = 0,
+  CUDAQ_TX_IN_FLIGHT = 1,
+  CUDAQ_TX_ERROR = 2,
+  CUDAQ_TX_READY = 3
+} cudaq_tx_status_t;
+
+// RPC wire-format constants — single source of truth in rpc_wire_format.h.
+#include "cudaq/realtime/daemon/dispatcher/rpc_wire_format.h"
```
boschmitt: Seems like weird place the `#include` a file. Why does it need to be here ?

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
```diff
@@ -28,17 +28,37 @@ typedef enum {
   CUDAQ_ERR_CUDA = 3
 } cudaq_status_t;
 
+// Dispatcher backend: device persistent kernel vs host-side loop
+typedef enum {
+  CUDAQ_BACKEND_DEVICE_KERNEL = 0,
+  CUDAQ_BACKEND_HOST_LOOP = 1
+} cudaq_backend_t;
```
boschmitt: Can we find better names for these? "backend" is already used in CUDA-Q, and means something quite different.

Would it be fair to say that these are distinguishing between having the control path in the GPU vs CPU. That is, `CUDAQ_BACKEND_DEVICE_KERNEL` indicates that is the GPU that is controlling the NIC, while `CUDAQ_BACKEND_HOST_LOOP` indicates that the CPU is controlling the NIC?

realtime/include/cudaq/realtime/daemon/dispatcher/dispatch_kernel_launch.h
```diff
@@ -75,9 +77,9 @@ constexpr std::uint32_t fnv1a_hash(const char *str) {
   return hash;
 }
 
-// RPC framing magic values (ASCII: CUQ?).
-constexpr std::uint32_t RPC_MAGIC_REQUEST = 0x43555152;  // 'CUQR'
-constexpr std::uint32_t RPC_MAGIC_RESPONSE = 0x43555153; // 'CUQS'
+// RPC framing magic values — sourced from rpc_wire_format.h.
+constexpr std::uint32_t RPC_MAGIC_REQUEST = CUDAQ_RPC_MAGIC_REQUEST;
+constexpr std::uint32_t RPC_MAGIC_RESPONSE = CUDAQ_RPC_MAGIC_RESPONSE;
```
boschmitt, line 82: Do we need this indirection ?

wsttiger:
> /ok to test 4f3a446

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
```diff
@@ -0,0 +1,88 @@
...
+  /// NULL when tx_flags is already a device-accessible pointer.
+  volatile uint64_t *tx_flags_dev;
+
+  /// Per-worker GraphIOContext array for separate RX/TX buffer support.
+  /// When non-NULL, launch_graph_worker fills a GraphIOContext per dispatch
+  /// and writes its device address into h_mailbox_bank[worker_id].
+  /// When NULL, legacy mode: raw RX slot pointer written to mailbox.
+  void *io_ctxs_host; ///< host view of GraphIOContext[num_workers]
+  void *io_ctxs_dev;  ///< device view of same pinned mapped memory
+} cudaq_host_dispatcher_config_t;
```
boschmitt: This struct is odd. We already have `cudaq_dispatcher_t` and `cudaq_dispatcher_config_t`. This ones seems this one seems like a redundant flattening of both.

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
```diff
@@ -0,0 +1,195 @@
...
+  // Allocate per-worker GraphIOContext array only when the caller wired
+  // separate RX and TX data buffers.  When rx_data == tx_data (in-place),
+  // the legacy path writes a raw slot pointer into the mailbox instead.
+  void *io_ctxs_host_ptr = nullptr;
+  void *io_ctxs_dev_ptr = nullptr;
+  if (ringbuffer->rx_data != ringbuffer->tx_data) {
+    size_t io_ctxs_bytes =
+        num_workers * sizeof(cudaq::realtime::GraphIOContext);
+    if (cudaHostAlloc(&io_ctxs_host_ptr, io_ctxs_bytes,
+                      cudaHostAllocMapped) == cudaSuccess) {
```
boschmitt: If this call fails, we just silently keep going? I don't understand why this would be ok.

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
```diff
@@ -0,0 +1,88 @@
...
+} cudaq_host_dispatch_worker_t;
+
+typedef struct {
+  void *rx_flags; ///< opaque cuda::std::atomic<uint64_t>*
+  void *tx_flags; ///< opaque cuda::std::atomic<uint64_t>*
+  uint8_t *rx_data_host;
+  uint8_t *rx_data_dev;
+  uint8_t *tx_data_host;
+  uint8_t *tx_data_dev;
+  size_t tx_stride_sz;
```
boschmitt: I'm not sure if we actually need this struct (see another comment), but, if we do, are we assuming `tx_stride_sz == rx_stride_sz`? The `cudaq_ringbuffer_t` struct has both separate, shouldn't that be the case here ?

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
```diff
@@ -0,0 +1,195 @@
...
+  host_config.tx_data_dev = ringbuffer->tx_data;
+  host_config.tx_stride_sz = ringbuffer->tx_stride_sz;
+  host_config.h_mailbox_bank = handle->h_mailbox_bank;
+  host_config.num_slots = config->num_slots;
+  host_config.slot_size = config->slot_size;
+  host_config.workers = handle->workers;
+  host_config.num_workers = num_workers;
+  host_config.function_table = table->entries;
+  host_config.function_table_count = table->count;
+  host_config.shutdown_flag = (void *)(uintptr_t)shutdown_flag;
```
boschmitt: Shouldn't `host_config.shutdown_flag` be backed by a `cuda::std::atomic<int>` ? Here it is taking a `volatile int`, and thus the underlying storage won't be an atomic. Perhaps this is not a big issue for a shutdown flag, but different writers and readers will be seeing this differently.

wsttiger: The volatile int* is intentional here — this is a C API boundary (extern "C") so we need plain C types for ABI stability. Internally, the dispatch loop casts it to cuda::std::atomic<int>* to get acquire semantics on the reader side. This is safe because cuda::std::atomic<int> is lock-free and layout-compatible with int on all CUDA-supported platforms. The writer (caller) only ever does a single store of 1 to signal shutdown, so no atomic RMW or complex protocol is needed from the C side. Added a comment at the cast site to document this.

realtime/include/cudaq/realtime/daemon/dispatcher/dispatch_kernel_launch.h
wsttiger: The rpc_wire_format.h header defines the magic values as C preprocessor macros so they work in both C and CUDA device code. The constexpr wrappers here give C++ code typed constants within the cudaq::realtime namespace, which avoids macro pollution and plays nicely with templates. The macros remain the single source of truth — these just forward them into C++ land.

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
wsttiger: Renamed to cudaq_host_dispatch_loop_ctx_t to clarify this isn't a user-facing config — it's an internal runtime context captured by value into the dispatch loop thread. It bundles a subset of fields from the ringbuffer, dispatcher config, and function table, plus host-specific runtime state (workers, idle_mask, io_ctxs_*, etc.) that doesn't belong in the public API structs. The device kernel path passes these as individual kernel arguments; the host path needs a struct because std::thread captures a single lambda value. Composing the public types would add unused fields and force verbose access patterns with no real benefit.

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
wsttiger: slot_size serves as the RX stride — in a contiguous ring buffer, slot capacity equals stride. tx_stride_sz is a separate field because the GraphIOContext path supports TX slots with a different size than RX. So we're not assuming they're equal; they're intentionally tracked independently. Renamed the struct to _ctx_t to make it clearer this is an internal runtime context, not a mirror of cudaq_ringbuffer_t.

realtime/lib/daemon/dispatcher/host_dispatcher_capi.cu
wsttiger: Now both cudaHostAlloc and cudaHostGetDevicePointer failures are handled explicitly — they clean up streams and the handle, then return nullptr. No silent fallback to the wrong buffer path.

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
wsttiger: Renamed:

cudaq_backend_t → cudaq_dispatch_path_t
CUDAQ_BACKEND_DEVICE_KERNEL → CUDAQ_DISPATCH_PATH_DEVICE
CUDAQ_BACKEND_HOST_LOOP → CUDAQ_DISPATCH_PATH_HOST
.backend field → .dispatch_path

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
wsttiger: fixed

bmhowe23:
@wsttiger - it looks like this needs code formatting applied to pass the CI. It also needs a merge from main. Other than that, are we ready to merge this?

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
wsttiger: done

realtime/include/cudaq/realtime/daemon/dispatcher/host_dispatcher.h
wsttiger: Addressed and refactored in the following commit: https://github.com/NVIDIA/cuda-quantum/pull/4180/changes/842ae9fd445a0a45b37b2db01f718f446115c7f1

PR #4182, opened 2026-03-19
Pass pipeline logging instrumentation

atgeller:
Adds instrumentation to log pass pipelines when enabled. Use `CUDAQ_PIPELINE_LOG=path/to/log/file` to enable logging. Currently, this will only log JIT passes.


runtime/logger/pipeline/PassPipelineLogging.h
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include "llvm/ADT/StringRef.h"
```
Renaud-K, line 11: Forward definitions should be enough here. This will prevent these headers to bleed into other parts of the code.

runtime/logger/pipeline/PassPipelineLogging.h
```diff
@@ -0,0 +1,29 @@
...
+#include "mlir/Pass/PassManager.h"
+
+namespace cudaq {
```
Renaud-K: This should now be under the `cudaq_internal::compiler` namespace.
Ideally, it should even be under detail but we have passes still being run outside the compiler so it needs to be exposed through the interface.

runtime/internal/compiler/logging/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+add_mlir_library(cudaq-pass-pipeline-logging
```
Renaud-K: logger is not the right place for this because logger is a lower level library. And it will itself have to move to meet our [CppAPICodingStyle](https://github.com/Renaud-K/cuda-quantum/blob/7f37351841dc00bd8429f1f74c173eb742605a63/CppAPICodingStyle.md)

So we need to find a better home for this library and I am thinking that we will move it where the compiler (cudaq-mlir-runtime) is going to move. I think it should be under `/cuda-quantum/runtime/internal/compiler/logging`. And the compiler will be above it in `/cuda-quantum/runtime/internal/compiler`.

The header file is going to be more work. 
Can you move it under `/cuda-quantum/runtime/internal/compiler/include/cudaq_internal/compiler/logging`

This is part of a proposal to amend CppAPICodingStyle.md that has backing.
You will have to update the cudaq-quake tool accordingly.

targettests/pipeline/infleqtion.cpp
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  cudaq::sample(sampleKernel);
+  cudaq::observe(observeKernel, cudaq::spin_op::from_word("Z"));
+  // CHECK: "type":"configured"
+  // CHECK: "label":"
+  // CHECK: "pipeline":"
+  // CHECK: "type":"executed"
+  // CHECK: "label":"
+  // CHECK: "passes":[{"pass":"
```
Renaud-K, line 36: What are we testing for here? Are we not checking the passes?

Renaud-K (changes requested):

runtime/logger/pipeline/PassPipelineLogging.h
atgeller: I've moved to that namespace.

runtime/internal/compiler/logging/CMakeLists.txt
atgeller: I've moved it to `/cuda-quantum/runtime/internal/compiler` like we discussed.

runtime/logger/pipeline/PassPipelineLogging.h
atgeller: Done.

targettests/pipeline/infleqtion.cpp
atgeller: Just that the logging is happening, for now. This is an incremental PR to introduce an accounting tool that we can eventually use for raising errors/alerts when the pipeline changes, but I think more discussion is needed to determine what that error/alert should look like.

runtime/internal/compiler/logging/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  PassPipelineLogging.cpp
+
+  LINK_LIBS PRIVATE
+    MLIRIR
+    MLIRPass
+    MLIRSupport
+)
+
+target_include_directories(cudaq-pass-pipeline-logging
+  PRIVATE
```
Renaud-K: This should be PUBLIC. Here we define the public interface of the module. 
Then we use `target_link_libraries` we get the module's public interface in the search path.

runtime/internal/compiler/logging/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    ${CMAKE_SOURCE_DIR}/runtime/internal/compiler/include
+    ${CMAKE_SOURCE_DIR}/runtime
+)
+
+target_compile_options(cudaq-pass-pipeline-logging PRIVATE -fno-rtti)
```
Renaud-K: Is this still needed with `add_mlir_library`?

runtime/cudaq/platform/default/rest_server/CMakeLists.txt
```diff
@@ -70,12 +70,13 @@ add_library(${LIBRARY_NAME}
...
 target_include_directories(${LIBRARY_NAME} 
     PUBLIC 
+       $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/runtime/internal/compiler/include>
        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/runtime>
        $<BUILD_INTERFACE:${CUDA_INCLUDE_DIRS}>
        $<INSTALL_INTERFACE:include>
     PRIVATE . ../../)
 
-target_link_libraries(${LIBRARY_NAME} PRIVATE cudaq cudaq-logger cudaq-mlir-runtime fmt::fmt-header-only unzip_util nvqir)
+target_link_libraries(${LIBRARY_NAME} PRIVATE cudaq cudaq-logger cudaq-mlir-runtime cudaq-pass-pipeline-logging fmt::fmt-header-only unzip_util nvqir)
```
Renaud-K: Should already be included in cudaq-mlir-runtime.

runtime/cudaq/platform/default/rest_server/CMakeLists.txt
```diff
@@ -104,6 +106,7 @@ target_link_libraries(rest-remote-platform-server
     cudaq-operator
     cudaq-em-default 
     cudaq-mlir-runtime
+    cudaq-pass-pipeline-logging
```
Renaud-K: Should be coming from cudaq-mlir-runtime

runtime/cudaq/platform/default/rest_server/CMakeLists.txt
```diff
@@ -89,6 +90,7 @@ add_library(rest-remote-platform-server
 
 target_include_directories(rest-remote-platform-server 
   PUBLIC 
+    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/runtime/internal/compiler/include>
```
Renaud-K: Should be coming from `target_link_libraries`

python/extension/CMakeLists.txt
```diff
@@ -119,11 +119,13 @@ declare_mlir_python_extension(CUDAQuantumPythonSources.Extension
 target_include_directories(CUDAQuantumPythonSources.Extension INTERFACE 
     ${CMAKE_SOURCE_DIR}/python 
     ${CMAKE_SOURCE_DIR}/python/utils
+    ${CMAKE_SOURCE_DIR}/runtime/internal/compiler/include
```
Renaud-K: This should be coming from `target_link_libraries`

targettests/pipeline/infleqtion.cpp
Renaud-K: Sounds good.

unittests/Optimizer/CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ gtest_discover_tests(OptimizerUnitTests)
 add_executable(test_quake_synth QuakeSynthTester.cpp)
 target_link_libraries(test_quake_synth
   PRIVATE 
+  MLIRPass
```
Renaud-K, line 45: Do we still need this?

atgeller: Nope, pushing a fix.

runtime/common/PassPipelineLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+  std::string label;
+  std::vector<PassRecord> records;
+
+  explicit PipelineRecorder(llvm::StringRef label = {})
+      : label(label.empty() ? "<unnamed>" : label.str()) {}
+
+  ~PipelineRecorder() override { flushToLog(); }
+
+  void runBeforePass(mlir::Pass *pass, mlir::Operation *op) override {
+    records.push_back(
```
sacpis, line 89: A question. Is this records vector threadsafe? Seems like we will need to ensure that `disableMultithreading` is set before calling `pm.run()`?

runtime/cudaq/builder/kernel_builder.cpp
```diff
@@ -973,6 +975,7 @@ jitCode(ImplicitLocOpBuilder &builder, ExecutionEngine *jit,
     pm.addNestedPass<func::FuncOp>(createCSEPass());
     pm.addPass(cudaq::opt::createConvertToQIR());
     pm.addPass(createCanonicalizerPass());
+    cudaq_internal::maybeLogPassPipeline(pm, "cudaq::builder JIT");
```
sacpis, line 978: Seems like we should have different labels for these so that it will be helpful to distinguish?

python/tests/backends/test_pipeline_logging.py
```diff
@@ -0,0 +1,103 @@
...
+import cudaq
+import pytest
+from cudaq import spin
+
+
+@pytest.fixture(autouse=True)
+def reset_target():
+    cudaq.reset_target()
+    yield
+    cudaq.reset_target()
```
sacpis: Why are we calling `reset_target()` before and after the test? Should it be just after the test?

python/tests/backends/test_pipeline_logging.py
```diff
@@ -0,0 +1,103 @@
...
+@cudaq.kernel
+def observe_kernel():
+    qubit = cudaq.qubit()
+    h(qubit)
+
+
+@pytest.mark.parametrize("target,target_kwargs,supports_run", TARGET_CONFIGS)
+def test_pipeline_logging_decorator(tmp_path, monkeypatch, target,
+                                    target_kwargs, supports_run):
+    log_path = tmp_path / "pipeline.jsonl"
```
sacpis: Should we move out this hard coded value in a constant outside?

python/tests/backends/test_pipeline_logging.py
```diff
@@ -0,0 +1,103 @@
...
+        json.loads(line)
+        for line in log_path.read_text().splitlines()
+        if line.strip()
+    ]
+    assert any(entry.get("type") == "configured" for entry in entries)
+    assert any(entry.get("type") == "executed" for entry in entries)
+
+
+def test_pipeline_logging_builder_default(tmp_path, monkeypatch):
+    log_path = tmp_path / "pipeline.jsonl"
```
sacpis: same here. Wondering if the same log_file can be reused for these tests?

runtime/common/BaseRestRemoteClient.h
```diff
@@ -205,6 +206,8 @@ class BaseRemoteRestRuntimeClient : public RemoteRuntimeClient {
         moduleOp.getContext()->disableMultithreading();
         pm.enableIRPrinting();
       }
+      cudaq_internal::maybeLogPassPipeline(pm,
+                                           name + ":" + passName + "-synth");
```
sacpis: nit: just a thought. Should we have a function which constructs `name + ":" + passName + "-synth"` and returns it?

runtime/common/BaseRestRemoteClient.h
```diff
@@ -247,6 +250,7 @@ class BaseRemoteRestRuntimeClient : public RemoteRuntimeClient {
     tm.setEnabled(cudaq::isTimingTagEnabled(cudaq::TIMING_JIT_PASSES));
     auto timingScope = tm.getRootScope(); // starts the timer
     pm.enableTiming(timingScope);         // do this right before pm.run
+    cudaq_internal::maybeLogPassPipeline(pm, name + ":client");
```
sacpis: same here.

runtime/common/BaseRestRemoteClient.h
```diff
@@ -300,6 +304,7 @@ class BaseRemoteRestRuntimeClient : public RemoteRuntimeClient {
     mlir::PassManager pm(ctx);
     // For now, the server side expects full-QIR.
     opt::addAOTPipelineConvertToQIR(pm);
+    cudaq_internal::maybeLogPassPipeline(pm, name + ":aot-qir");
```
sacpis: same here.

runtime/common/PassPipelineLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+    case '\t':
+      out += "\\t";
+      break;
+    default:
+      out += c;
+      break;
+    }
+  }
+  return out;
+}
```
sacpis, line 51: I think we can optimize(instead of using +=) this function, we can collect all characters in some string and then flush it once at the end.

runtime/common/PassPipelineLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+      if (i > 0)
+        record += ",";
+      record += "{\"pass\":\"" + jsonEscape(records[i].passArg) +
+                "\",\"op\":\"" + jsonEscape(records[i].opName) + "\"";
+      if (records[i].failed)
+        record += ",\"failed\":true";
+      record += "}";
+    }
+    record += "]}";
+    appendToLogFile(logPath, record);
```
sacpis, line 116: Should we use `llvm::raw_string_ostream` here? This will avoid the creation of intermediate std::string (with +). We can also get rid of `\"`.

runtime/common/PassPipelineLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+    return;
+
+  std::string pipeline;
+  llvm::raw_string_ostream os(pipeline);
+  pm.printAsTextualPipeline(os);
+
+  llvm::StringRef effectiveLabel = label.empty() ? "<unnamed>" : label;
+  std::string record = "{\"type\":\"configured\",\"label\":\"" +
+                       jsonEscape(effectiveLabel) + "\",\"pipeline\":\"" +
+                       jsonEscape(pipeline) + "\"}";
```
sacpis, line 132: same here.

sacpis (commented):
Left few comments.

PR #4186, opened 2026-03-19
More robust catching/handling of possible aliasing in Phase Folding

atgeller:


lib/Optimizer/Transforms/PhaseFolding.cpp
```diff
@@ -54,6 +54,19 @@ static bool isSupportedValue(Value ref) {
   if (!isa<quake::AllocaOp>(ref.getDefiningOp()))
```
schweitzpgi:
```suggestion
  if (!ref.getDefiningOp<quake::AllocaOp>())
```

PR #4191, opened 2026-03-20
[Optimizer] Constant-fold cc::IfOp successor regions and invocation bounds

sanarang-nv:
**What does this PR implement? **

This PR applies two (similar) flavors of constant folding on the `IfOp`:
  - `cc::IfOp::getSuccessorRegions`: when the condition is a compile-time constant, only the live region is returned as a successor. 
  - `cc::IfOp::getRegionInvocationBounds`: when the condition is constant, the live region gets
  exact bounds {1,1} (always executes) and the dead region gets {0,0} (never executes).

**How do we test this?**

-  C++ unit tests in `unittests/Optimizer/IfOpConstantFoldTest.cpp` directly call both methods
  with constant and non-constant operands and assert the correct regions/bounds are returned.
- Lit test in `test/Transforms/cc_if_const_fold.qke` verifies end-to-end behavior


1tnguyen:
> /ok to test 672d4b4

1tnguyen:
> /ok to test [0bdb79a](https://github.com/NVIDIA/cuda-quantum/pull/4191/commits/0bdb79a879f3ba67ab1e0b9d4bd60bb6e52ce8b1)

sacpis:
Seems like there are some build failures, please try to build cudaq locally using `bash scripts/build_cudaq.sh -v`

sanarang-nv:
> Seems like there are some build failures, please try to build cudaq locally using `bash scripts/build_cudaq.sh -v`

@sacpis, thanks. I wasn't able to setup the devcontainer earlier, but it went through and I was able to run `build_cudaq` successfully inside the container.

sacpis:
> /ok to test 32bf42c

schweitzpgi:
I know folding IfOp was already done at one point. Let me see if I can find it...

schweitzpgi:
> I know folding IfOp was already done at one point. Let me see if I can find it...

OK. There is a bit of code in the `RewriteIf` pattern in LowerToCFGPatterns.inc that already attempted to do this folding. <s>We do not want both of them, as it will just be confusing. I think the solution in this PR is more in the spirit of MLIR, and would prefer it.

`RewriteIf` has a primary purpose of lowering `If` ops to a plain CFG (with branches), so it cannot be completely removed. Please remove the `rewriteOnlyIfConst` option from it however.</s>

Nevermind, this does not really fold the cc.if.

test/Transforms/cc_if_const_fold.qke
```diff
@@ -0,0 +1,60 @@
...
+func.func private @use_i32(i32)
+
+// Constant true: then-region is live {1,1}, else-region is dead {0,0}.
+// Verifies that a value used only in the then-region is sunk into it.
+func.func @sink_into_true_branch() {
+  %true = arith.constant true
+  %c42 = arith.constant 42 : i32
+  cc.if (%true) {
+    func.call @use_i32(%c42) : (i32) -> ()
+    cc.continue
```
schweitzpgi, line 26:
```suggestion
```
This `continue` isn't necessary.

PR #4193, opened 2026-03-21
Enable external backend integration via entry points and CUDAQ_BACKEND_PATH

sacpis:
A new hardware integration strategy that enables external backend packages to be discovered and used by CUDA-Q at runtime without requiring their code to be merged in CUDA-Q repository.

### What we need from hardware provider in a wheel
   - .so file
   - .yml file (for configuration)

- The hardware provider's `pyproject.toml` should have the following entry point
```toml
[project]
name = "<hw-provider>"
...

[project.entry-points."cudaq.backends"]
<hw-provider> = "<hw-provider>:register"
```

- Inside `<hw-provider>/__init__.py`
  - This sets the env var before the C extension is initialized

```python
import os

_pkg_dir = os.path.dirname(os.path.abspath(__file__))

def register():
    existing = os.environ.get('CUDAQ_BACKEND_PATH', '')
    os.environ['CUDAQ_BACKEND_PATH'] = _pkg_dir + (':' + existing if existing else '')
```

### How it will work is
   - User just needs to `pip install <hw-provider>`, which will install a wheel with <hw-provider>/targets/<hw-provider>.yml and <hw-provider>/lib/libcudaq-serverhelper-<hw-provider>.so and a register() entry point in group `cudaq.backends`
   - Issuing `import cudaq` will trigger a entry point discovery
      - It will run `register()` method
      - `register()` appends its package root to CUDAQ_BACKEND_PATH
      - LinkedLibraryHolder scans CUDAQ_BACKEND_PATH, then calls `findAvailableTargets()` through which `RuntimeTarget` gets serverHelperLibDir
      - When a user sets a target using `cudaq.set_target("<hw-provider>")`
         - set_target() will open the .so file from serverHelperLibDir using dlopen
         - Then it will put `__yml_path` in the backend config string
         - Now, `BaseRemoteRESTQPU::setTargetBackend()` reads the above YAML file
   
**Note:** `<hw-provider>` is hardware provider like TII, IQNQ, Quantinuum etc.
 
This PR will be followed by the docs PR.

### DIFF
--- a/python/cudaq/__init__.py
+++ b/python/cudaq/__init__.py
@@ -127,6 +127,17 @@ except Exception:
         print("Could not find a suitable cuQuantum Python package.")
     pass
 
+# Discover external backends
+try:
+    from importlib.metadata import entry_points as _entry_points
+    for _ep in _entry_points(group='cudaq.backends'):
+        try:
+            _ep.load()()
+        except Exception:
+            pass
+except Exception:
+    pass
+
 # ============================================================================ #
 # Module Imports
 # ============================================================================ #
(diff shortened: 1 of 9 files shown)


sacpis:
Closing this PR as it has been merged in @anpaz's plugin project.

PR #4195, opened 2026-03-22
[DOCS] [EXAMPLES] Add missing math import

benediktjohannes:
We use (e. g.) `math.pi`, but we didn't import is -> missing import was added.

Contributed by **Benedikt Johannes**


sacpis:
> /ok to test 8559a94

benediktjohannes:
@sacpis thanks for the review, status checks seem to fail

benediktjohannes:
Unforetunately, I guess that this is some kind of internal workflow and I can't do anything about it, so could you please have a look on this? Thanks!

PR #4198, opened 2026-03-22
Extend CompiledKernel to fat binary with JIT/MLIR support

lmondada:


lmondada:
> /ok to test ed52a81

lmondada:
> /ok to test c2f8b43

lmondada:
> /ok to test [3d7233e](https://github.com/NVIDIA/cuda-quantum/pull/4198/commits/3d7233e41a19b611ce52cc0790e3ca41f79714f1)

runtime/common/CompiledKernel.h
```diff
@@ -9,28 +9,91 @@
...
+  CompiledKernel(std::string kernelName, ResultInfo resultInfo);
+
+  // --- Queries ---
+
+  bool hasJit() const { return jitRepr.has_value(); }
+  bool hasMlir() const { return mlirRepr.has_value(); }
+
+  /// Whether the kernel is fully specialized (all arguments inlined). For JIT
+  /// kernels this means `argsCreator` is null.
+  /// Currently, MLIR-only kernels are always considered fully specialized.
```
atgeller, line 81: Is the reason for this just that if there's no JIT there's nothing to be specialized and therefore it's vacuously specified?

runtime/common/CompiledKernel.cpp
```diff
@@ -35,6 +47,25 @@ cudaq::CompiledKernel::execute(const std::vector<void *> &rawArgs) const {
   return {nullptr, 0};
 }
 
-void (*cudaq::CompiledKernel::getEntryPoint() const)() { return entryPoint; }
+cudaq::KernelThunkResultType cudaq::CompiledKernel::execute() const {
+  if (jitRepr && jitRepr->argsCreator)
```
atgeller: Can't we used `isFullySpecialized` here?

runtime/common/CompiledKernel.h
lmondada: The reason is simply that right now we don't handle mlir with arguments 😆 The "justification" for this is that currently in Python, there is no case where we need parametric calls for remote execution (which is when the MLIR representation is required). I'm sure this will be changed soon, though.

runtime/common/CompiledKernel.cpp
```diff
@@ -7,24 +7,36 @@
...
 cudaq::CompiledKernel::execute(const std::vector<void *> &rawArgs) const {
   auto funcPtr = getEntryPoint();
-  if (hasResult) {
+  if (resultInfo.hasResult()) {
     void *buff = const_cast<void *>(rawArgs.back());
     return reinterpret_cast<KernelThunkResultType (*)(void *, bool)>(funcPtr)(
         buff, /*client_server=*/false);
   }
-  if (argsCreator) {
+  if (jitRepr && jitRepr->argsCreator) {
```
atgeller: > Can't we used `isFullySpecialized` here?

Here too actually

runtime/common/CompiledKernel.cpp
lmondada: yep, will fix, thanks!

lmondada:
> /ok to test [7e78231](https://github.com/NVIDIA/cuda-quantum/pull/4198/commits/7e78231c0445e8cb507cde177ba23cf65e773ef9)

Renaud-K:
[Moving cudaq-mlir-runtime into its own directory](https://github.com/NVIDIA/cuda-quantum/pull/4233) has highlighted some dependencies that we should address. This [include](https://github.com/Renaud-K/cuda-quantum/blob/89c2a301092c685b139d6c84f4439e0b527bacb3/runtime/common/ExecutionContext.h#L17) is actually adding a dependency on `cudaq-mlir-runtime-headers` everywhere `ExecutionContext.h` is included which is almost everywhere and has forced me to make `cudaq-mlir-runtime-headers` [a public API of cudaq-common](https://github.com/Renaud-K/cuda-quantum/blob/e1e8cfab857a3b07a252884eb318eb3db3c0293a/runtime/common/CMakeLists.txt#L52). It has defeated the effort of [limiting header file visibility](https://github.com/NVIDIA/cuda-quantum/pull/4233).

I would not want the `CompiledKernel` to be designed to workaround the dependencies introduced by the `JitEngine` and I want to remove this `cudaq-mlir-runtime-headers` dependency.  The `JitEngine` could be in `cudaq-common`, if like the `CompiledKernel`, it only relied on `cudaq-runtime-mlir` for its creation. This can be done with [this PR](https://github.com/NVIDIA/cuda-quantum/pull/4245) whereby the methods of `JitEngine` all use closures defined during its creation. Then `JitEngine` could move into `CompiledKernel.h` or a sibling header. 

I think this would also provide more flexibility for the design on `CompiledKernel` as we would not have to worry about where the lookup is made and also more flexibility for where the `JitEngine` can be included.

PR #4199, opened 2026-03-23
Expose CompiledKernel to Python, delete JITExecutionCache

lmondada:
by passing around instances of `CompiledKernel` instead of function pointers, we no longer need to manage the lifetime of the `JITEngine` manually: it simply stays alive as long as a CompiledKernel instance is alive.

This changes the Python API in the following way:
- `PyKernelDecorator.beta_reduction` now returns a `CompiledKernel` instance instead of a raw function pointer. Neither type is particularly useful to the end-user at the moment, but exposing a dedicated type is a step in the right direction.

### DIFF
--- a/python/runtime/interop/PythonCppInterop.h
+++ b/python/runtime/interop/PythonCppInterop.h
@@ -7,10 +7,7 @@
  ******************************************************************************/
 #pragma once
 
-#include "common/CompiledKernel.h"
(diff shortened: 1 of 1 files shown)


lmondada:
> /ok to test 521a2e4

lmondada:
> /ok to test adf00ae

PR #4204, opened 2026-03-23
[Quake] Add a type for collection of measurements

khalatepradnya:
This PR introduces support for handling collections of quantum measurement results in the Quake dialect. 
Following have been added:
- a new `!quake.measurements<N>` type,
-  an operation to extract individual measurements from such collections, 
- associated parsing, verification, and testing


schweitzpgi (approved):
Great ideas! ;)

PR #4208, opened 2026-03-24
Use collection of measurements

khalatepradnya:
This PR introduces significant improvements to how multi-qubit measurements are represented and lowered throughout the CUDA-Q compiler stack. 
* Use the new type introduced in PR #4204 
* Replace the use of `std::vector` for multi-qubit measurement results with the dedicated `!quake.measurements<N>` type.
* Update the passes for adding, expanding and combining measurements
* Replace all the `!cc.stdvec<!quake.measure>` types with `!quake.measurements<N>`
* Introduced new runtime functions and corresponding lowering logic for handling result arrays: `__quantum__rt__result_array_create_1d` and `__quantum__rt__result_array_get_element_ptr_1d`.
* Implemented a new rewrite for lowering `quake.get_measure` operations to use the new QIR result array element pointer intrinsic, ensuring correct extraction of individual measurement results from a measurement array.


lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -135,6 +135,9 @@ struct QIRAPITypeConverter : public TypeConverter {
         [&](quake::CableType ty) { return getArrayType(ty.getContext()); });
     addConversion(
         [&](quake::MeasureType ty) { return getResultType(ty.getContext()); });
+    addConversion([&](quake::MeasurementsType ty) {
+      return LLVM::LLVMPointerType::get(ty.getContext());
```
schweitzpgi: Should this actually be an `Array*` type? As Measurements is really a short vector of `Result*` values.

lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -695,6 +695,51 @@ struct MergeRotationPattern : public OpRewritePattern<OP> {
   }
 };
 
+// %0 = quake.alloca !quake.veq<2>
```
schweitzpgi, line 718: Nice catch.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -117,6 +117,59 @@ class ExpandRewritePattern : public OpRewritePattern<A> {
...
+      TODO_loc(loc, "unsized measurements in `ExpandDiscriminatePattern`");
+    }
+
+    Value buff = rewriter.create<cudaq::cc::AllocaOp>(loc, i8Ty, totalToRead);
+
+    if (measTy.hasSpecifiedSize()) {
+      std::size_t n = measTy.getSize();
+      for (std::size_t i = 0; i < n; ++i) {
+        Value getMeas = rewriter.create<quake::GetMeasureOp>(loc, measVal, i);
+        Value bit = rewriter.create<quake::DiscriminateOp>(loc, i1Ty, getMeas);
```
schweitzpgi: I think we want to use `discOp.getResultType()` instead of `i1Ty` here for type correctness, no?

schweitzpgi (commented):
This looks great, Pradnya.

I'm really not sure if the QIR spec wants us to stick `Result*` in an `Array*` bucket or not. My hunch is that it does since using LLVM's `alloca` instruction will cause QIR verification failures.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
schweitzpgi: Specifically,
```mlir
%1 = quake.discriminate %0 : (!quake.measure) -> i4
```
is a valid operation. The results is not required to be `i1`.

khalatepradnya: I will update this... 
`ExpandRewritePattern` has the same issue (which I referred to), I will fix both places.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
khalatepradnya: I am not sure about this one... Looking at the latest QIR spec (v2.0), there is no `Array` type, just `Qubit` and `Result`.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
khalatepradnya: Fixed.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
khalatepradnya: Fixed now.

test/Transforms/result_array_alloca.qke
```diff
@@ -0,0 +1,21 @@
...
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --convert-to-qir-api %s | FileCheck %s
+
+func.func @test_measure_alloca() attributes {"cudaq-kernel"} {
+  %c3 = arith.constant 3 : i64
+  %buf = cc.alloca !quake.measure[%c3 : i64]
```
schweitzpgi: This is not legal IR. The point of introducing the `!quake.measurements<n>` type is so that we do not allocate pure abstractions in classical memory.

runtime/nvqir/QIRTypes.h
```diff
@@ -83,6 +83,7 @@ Array *__quantum__rt__array_slice(Array *array, int32_t dim,
 Array *__quantum__rt__array_slice_1d(Array *array, int64_t range_start,
                                      int64_t range_step, int64_t range_end);
 Array *quantum__rt__array_slice(Array *array, int32_t dim, Range range);
+Array *__quantum__rt__result_array_create_1d(int64_t count);
```
schweitzpgi:
```suggestion
Array *__quantum__rt__result_array_create_1d(std::int64_t count);
```

runtime/nvqir/QIRTypes.cpp
```diff
@@ -153,6 +153,10 @@ Array *__quantum__rt__array_create_1d(int32_t itemSizeInBytes,
   return array;
 }
 
+Array *__quantum__rt__result_array_create_1d(int64_t count) {
```
schweitzpgi:
```suggestion
Array *__quantum__rt__result_array_create_1d(std::int64_t count) {
```

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -203,11 +209,20 @@ class ExtendVeqMeasurePattern : public OpRewritePattern<quake::MzOp> {
...
+          resultType = quake::MeasurementsType::get(measure->getContext(),
+                                                    veqTy.getSize());
+        else
+          resultType =
+              quake::MeasurementsType::getUnsized(measure->getContext());
+        rewriter.replaceOpWithNewOp<quake::MzOp>(measure, TypeRange{resultType},
+                                                 ValueRange{veq},
+                                                 measure.getRegisterNameAttr());
+      } else if (measure.use_empty())
         rewriter.eraseOp(measure);
```
schweitzpgi:
```suggestion
      } else if (measure.use_empty()) {
rewriter.eraseOp(measure);
}
```

lib/Optimizer/Transforms/CombineMeasurements.cpp
```diff
@@ -132,7 +132,13 @@ class ExtendQubitMeasurePattern : public OpRewritePattern<quake::MzOp> {
       analysis.resultQubitVals[offset] =
           std::make_pair(idx, std::to_string(idx));
 
-      auto resultType = cudaq::cc::StdvecType::get(measure.getType(0));
+      Type resultType;
+      if (auto veqTy = dyn_cast<quake::VeqType>(veq.getType());
+          veqTy && veqTy.hasSpecifiedSize())
```
schweitzpgi:
```suggestion
      if (auto veqTy = quake::isConstantQuantumRefType(veq.getType()))
```

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1647,21 +1647,38 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
 
     if (funcName == "mx" || funcName == "my" || funcName == "mz") {
       // Measurements always return a bool or a std::vector<bool>.
-      bool useStdvec =
+      bool useCollection =
```
```suggestion
      bool useMeasurements =
```

test/Transforms/result_array_alloca.qke
khalatepradnya: This hand-generated input doesn't make sense.
Given https://github.com/NVIDIA/cuda-quantum/pull/4223 this will give error, removing it.

lib/Optimizer/CodeGen/ConvertToExecMgr.cpp
```diff
@@ -51,6 +51,9 @@ struct QuakeTypeConverter : public TypeConverter {
     addConversion([](quake::MeasureType ty) {
       return IntegerType::get(ty.getContext(), 64);
     });
+    addConversion([](quake::MeasurementsType ty) {
+      return cudaq::cc::PointerType::get(IntegerType::get(ty.getContext(), 8));
```
schweitzpgi: Why a pointer?

lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1647,21 +1647,38 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+          for (auto a : args) {
+            if (isa<quake::RefType>(a.getType())) {
+              totalSize++;
+            } else if (auto veqTy = dyn_cast<quake::VeqType>(a.getType())) {
+              if (veqTy.hasSpecifiedSize())
+                totalSize += veqTy.getSize();
+              else
+                allKnown = false;
+            }
+          }
```
schweitzpgi, line 1663: This should use `quake::isConstantQuantumRefType(type)` and `quake::getAllocationSize(type)` instead of rolling our own here.

test/Transforms/result_array_alloca.qke
khalatepradnya: yes, reverted this with 33ecca8.
I can add a meaningful test with deferred discriminate.

lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -715,6 +715,51 @@ struct MergeRotationPattern : public OpRewritePattern<OP> {
...
+    for (auto target : measOp.getTargets()) {
+      if (isa<quake::RefType>(target.getType())) {
+        totalSize++;
+      } else if (auto veqTy = dyn_cast<quake::VeqType>(target.getType())) {
+        if (!veqTy.hasSpecifiedSize())
+          return failure();
+        totalSize += veqTy.getSize();
+      } else {
+        return failure();
+      }
```
schweitzpgi: Again, let's use `quake::isConstantQuantumRef()` and `quake::getAllocationSize()`

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -51,13 +51,22 @@ class ExpandRewritePattern : public OpRewritePattern<A> {
...
+    // 2. Determine the element type from users (default to `i1`).
+    Type elemTy = rewriter.getI1Type();
+    for (auto *user : measureOp.getMeasOut().getUsers()) {
+      if (auto disc = dyn_cast<quake::DiscriminateOp>(user))
+        if (auto svTy =
+                dyn_cast<cudaq::cc::StdvecType>(disc.getResult().getType())) {
+          elemTy = svTy.getElementType();
+          break;
+        }
+    }
```
schweitzpgi: This appears to assume `i1` unless it can find an user that is a discriminate op with a `!cc.stdvec<t>` result type if which case it uses `t`.

However, `discriminate` can return any integer, `i<n>`, type value as well as a `stdvec`.

lib/Optimizer/CodeGen/ConvertToExecMgr.cpp
khalatepradnya: Since this is a collection of measurements, chose a pointer like we would have for say, `Array*`.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -163,6 +228,11 @@ class ExpandMeasurementsPass
         [](quake::MyOp x) { return usesIndividualQubit(x.getMeasOut()); });
     target.addDynamicallyLegalOp<quake::MzOp>(
         [](quake::MzOp x) { return usesIndividualQubit(x.getMeasOut()); });
+    // Convert only on the `MeasurementsType` operand, otherwise legal.
+    target.addDynamicallyLegalOp<quake::DiscriminateOp>(
+        [](quake::DiscriminateOp d) {
+          return !isa<quake::MeasurementsType>(d.getMeasurement().getType());
+        });
```
schweitzpgi: I'm thinking this algorithm may need to be altered a bit to avoid some of the pair-wise checking code.

1. Traverse and change all
```mlir
quake.discriminate %2 : (!quake.measurements<c>) -> !stdvec<i?>
```
to
```mlir
%2 = quake.get_measure %1[0] : (!quake.measurements<c>) -> !quake.measure
quake.discriminate %2 : (!quake.measure) -> i?
...
```

2. Traverse and change all
```mlir
%m = quake.m[xyz] %1 : (!!quake.veq<c>) -> quake.measurements<c>
```
to
```mlir
%2 = quake.extract_ref %1[0] : (!quake.veq<c>) -> !quake.ref
%3 = quake.m[xyz] %2 : (!quake.ref) -> !quake.measure
```
and substitute `%3` for any users of `%m` of the form
```mlir
quake.get_measure %m[0] : (!quake.measurements<c>) -> !quake.measure
```

schweitzpgi: At present, the only way to deal with measurements of unknown length is to have them feed into a discriminate user op.

```mlir
%1 = quake.mz %0 : (!quake.veq<?>) -> !quake.measurements<?>
%2 = quake.discriminate %1 : (!quake.measurements<?>) -> !cc.stdvec<i4>
```
becomes a loop over both operations, which is what this code did in all cases previously, because it was never the case that we might lower a measurement and a discriminate seperately.

khalatepradnya: > becomes a loop over both operations, which is what this code did in all cases previously, because it was never the case that we might lower a measurement and a discriminate seperately.

- It is still the case, we are doing eager discriminate as of this PR, the deferred discriminate will happen with PR #3800 .
- Q1: Does it still make sense to implement now? I was considering including it in that PR...
- Q2: Regarding the order, should the patterns (specifically `ExpandDiscriminatePattern`) be split into 2 phases, i.e. have two calls to `applyPartialConversion`, is that what's intended?

khalatepradnya: Discussed this offline and update the `ExpandMeasurements` pass accordingly.

lib/Optimizer/CodeGen/ConvertToExecMgr.cpp
khalatepradnya: Removed.

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -99,6 +99,8 @@ void createTargetCodegenPipeline(PassManager &pm,
                                  const TargetCodegenPipelineOptions &options) {
   createCommonTargetCodegenPipeline<isJIT>(pm, options);
   ::addQIRConversionPipeline(pm, options.target);
+  // QIR conversion may introduce cc.loop, lower to cf.
+  cudaq::opt::addLowerToCFG(pm);
```
schweitzpgi, line 103: Just noting that we may eventually want to generate the loop in CFG form directly in QIRAPI instead just to avoid running another pass for measurement loops.

lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -715,6 +715,51 @@ struct MergeRotationPattern : public OpRewritePattern<OP> {
...
+        dyn_cast<quake::MeasurementsType>(measOp.getMeasOut().getType());
+    if (!measTy || measTy.hasSpecifiedSize())
+      return failure();
+
+    std::size_t totalSize = 0;
+    for (auto target : measOp.getTargets()) {
+      if (quake::isConstantQuantumRefType(target.getType()))
+        totalSize += quake::getAllocationSize(target.getType());
+      else
+        return failure();
```
schweitzpgi:
```suggestion
      if (quake::isConstantQuantumRefType(target.getType())) {
        totalSize += quake::getAllocationSize(target.getType());
        continue;
      }
      return failure();
```
Alternative style and a bit more consistent with typical LLVM code possibly.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -2941,7 +2941,17 @@ def bodyBuilder(iterVar):
...
+                    all_known = True
+                    for q in qubits:
+                        if quake.isConstantQuantumRefType(q.type):
+                            total_size += quake.getAllocationSize(q.type)
+                        else:
+                            all_known = False
+                    if all_known and total_size > 0:
+                        measTy = quake.MeasurementsType.get(total_size)
+                    else:
+                        measTy = quake.MeasurementsType.get()
```
schweitzpgi, line 2954: I was wondering about the veq of length 0 case. If we know the number of qubits being measured is 0, we could simply erase the measurement entirely.

The case where the veq has size 1 is also interesting. We could get rid of the `measurements<1>` type and demote it as a canonicalization to `!quake.measure`.

test/AST-Quake/base_profile-1.cpp
schweitzpgi, line 1: The name of this file is misleading since it is testing adaptive profile too.

test/AST-Quake/qir_profiles.cpp
```diff
@@ -193,25 +193,20 @@ struct comprehensive {
...
 // ADAPT:         tail call void @__quantum__qis__u3__body(double 8.000000e-01, double 5.000000e-01, double -1.000000e+00, %Qubit* nonnull inttoptr (i64 3 to %Qubit*))
 // ADAPT:         tail call void @__quantum__qis__mz__body(%Qubit* null, %Result* null)
+// ADAPT:         tail call void @__quantum__rt__array_record_output(i64 7, i8* nonnull getelementptr inbounds ([14 x i8], [14 x i8]* @cstr.61727261793C6931207820373E00, i64 0, i64 0))
 // ADAPT:         tail call void @__quantum__rt__result_record_output(%Result* null, i8* nonnull getelementptr inbounds ([10 x i8], [10 x i8]* @cstr.73696E676C65746F6E00, i64 0, i64 0))
 // ADAPT:         tail call void @__quantum__qis__mz__body(%Qubit* nonnull inttoptr (i64 1 to %Qubit*), %Result* nonnull inttoptr (i64 1 to %Result*))
 // ADAPT:         tail call void @__quantum__rt__result_record_output(%Result* nonnull inttoptr (i64 1 to %Result*), i8* nonnull getelementptr inbounds ([5 x i8], [5 x i8]* @cstr.65696E7300, i64 0, i64 0))
-// ADAPT:         %[[VAL_2:.*]] = tail call i1 @__quantum__qis__read_result__body(%Result* nonnull inttoptr (i64 1 to %Result*))
```
schweitzpgi, line 199: These went away because the measurements aren't converted to integral types, right?

Maybe we should make one measurement save a bool so that the `read_result__body` still pops out.

test/Transforms/convert_to_qir_measurements.qke
```diff
@@ -0,0 +1,115 @@
...
+func.func @discriminate_i4(%m : !quake.measure) -> i4 attributes {"cudaq-kernel"} {
+  %bit = quake.discriminate %m : (!quake.measure) -> i4
+  return %bit : i4
+}
+
+// CHECK-LABEL:   func.func @discriminate_i4(
+// CHECK-SAME:                               %[[VAL_0:.*]]: !cc.ptr<!llvm.struct<"Result", opaque>>) -> i4 attributes {"cudaq-kernel", "qir-api"} {
+// CHECK:           %[[VAL_1:.*]] = cc.cast %[[VAL_0]] : (!cc.ptr<!llvm.struct<"Result", opaque>>) -> !cc.ptr<i1>
+// CHECK:           %[[VAL_2:.*]] = cc.load %[[VAL_1]] : !cc.ptr<i1>
+// CHECK:           %[[VAL_3:.*]] = arith.extui %[[VAL_2]] : i1 to i4
```
schweitzpgi:
```suggestion
// CHECK:           %[[VAL_1:.*]] = cc.cast %[[VAL_0]] : (!cc.ptr<!llvm.struct<"Result", opaque>>) -> !cc.ptr<i8>
// CHECK:           %[[VAL_2:.*]] = cc.load %[[VAL_1]] : !cc.ptr<i8>
// CHECK:           %[[VAL_3:.*]] = cc.cast %[[VAL_2]] : i8 to i4
```
This ought to look like.

- We want to use byte addressable memory (`i8*`)
- We want to use cc.cast to truncate a byte to a smaller bitstring

schweitzpgi (commented):
Looking good.

lib/Optimizer/CodeGen/Pipelines.cpp
khalatepradnya: Captured in new issue https://github.com/NVIDIA/cuda-quantum/issues/4243

test/AST-Quake/base_profile-1.cpp
khalatepradnya: Sure, renamed to `qir_profiles`

test/AST-Quake/qir_profiles.cpp
khalatepradnya: Added test.

test/Transforms/convert_to_qir_measurements.qke
khalatepradnya: Fixed the logic in `lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp` to get this o/p.

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Good points! But, I think both are orthogonal to the core plumbing in this PR, so I'd like to defer them to a follow-up. Captured in issue https://github.com/NVIDIA/cuda-quantum/issues/4244

schweitzpgi (approved):
Thanks, Pradnya.

PR #4210, opened 2026-03-24
Handle adjoint rotations in gate decompositions

lmondada:
When selecting gate decompositions, we were not distinguishing adjoint gates from their non-adjoint counterparts. This meant that we never selected decompositions such as `RxAdjToRx`, as the automatic selection algorithm was viewing it as a no-op and hence useless.

This PR fixes this.

Note that this would also allow us to distinguish between `S`/`T` and their adjoint, but this would require adding more decomposition rules and would not benefit us at the moment (because wherever `S` and `T` are supported we currently also support the adjoint forms).

PR #3644, opened 2025-12-02
Add Decomposition Pattern Selection logic

lmondada:
This PR introduces a "decomposition pattern graph", implemented in `DecompositionPatternSelection.cpp`. For encapsulation, it is not exposed in any header file, and hence can only be used through the new `selectDecompositionPatterns` function exposed in `DecompositionPatterns.h`.

The passes `Decomposition` and `BasisConversion` are modified to use this new pattern selection mechanism whenever the `basis` argument is specified and `enable-patterns` is empty. The `basis` argument was added to the `Decomposition` pass (and is optional).

`DecompositionGraph` is constructed using the metadata exposed by #3587 on the available decomposition patterns. The graph captures how gates can be decomposed into other gates (edges are decomposition patterns, vertices are gate types). For any desired target gate set, the set of all "useful" decompositions can be inferred by traversing the graph backwards, starting from the target gate set.

I have left the actual pass definitions in the target YAML files unchanged. It should be possible to substantially simplify them in a future PR, as using the `basis` argument should make most explicit lists of enabled and disabled patterns superfluous.

See also Jira issue CUDAQ-217

PR #4211, opened 2026-03-24
Use basis argument for Decomposition/BasisConversion passes

lmondada:
Done. ~TODO: rebase onto #4210~

This was an old outstanding task following https://github.com/NVIDIA/cuda-quantum/pull/3644. There is now a simpler 'declarative' way of defining which basis gates to decompose to, without having to explicitly list all patterns.

The pattern selection algorithm selects decomposition patterns via backward traversal of a decomposition pattern graph, choosing the shortest path to the target basis. This makes manually curating pattern lists unnecessary.


lmondada:
> /ok to test [ade6a82](https://github.com/NVIDIA/cuda-quantum/pull/4211/commits/ade6a8271ebcb7da1800809f2c132aeadbb17d55)

PR #4212, opened 2026-03-24
[core] Update quake.apply operation.

schweitzpgi:
Add a verifier and diagnostic message tests.

Fix bugs in apply-op-specialization and canonicalization. Add tests and handling for !quake.struq while we're here.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -397,27 +397,29 @@ def quake_ApplyOp : QuakeOp<"apply",
     Variadic<cc_CallableType>:$indirect_callee, // must be 0 or 1 element
     UnitAttr:$is_adj,
     Variadic<AnyQType>:$controls,
-    Variadic<AnyType>:$args
+    Variadic<AnyType>:$actuals
   );
   let results = (outs Variadic<AnyType>);
 
   let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+
   let builders = [
     OpBuilder<(ins "mlir::TypeRange":$retTy,
                    "mlir::SymbolRefAttr":$callee,
                    "mlir::UnitAttr":$is_adj,
                    "mlir::ValueRange":$controls,
                    "mlir::ValueRange":$args), [{
-      return build($_builder, $_state, retTy, callee, mlir::ValueRange{},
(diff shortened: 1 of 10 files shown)


include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
```diff
@@ -397,27 +397,29 @@ def quake_ApplyOp : QuakeOp<"apply",
     Variadic<cc_CallableType>:$indirect_callee, // must be 0 or 1 element
     UnitAttr:$is_adj,
     Variadic<AnyQType>:$controls,
-    Variadic<AnyType>:$args
+    Variadic<AnyType>:$actuals
```
khalatepradnya, line 400: nit: `args` is more meaningful than `actuals` IMO.

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -314,8 +314,12 @@ struct ApplyOpPattern : public OpRewritePattern<quake::ApplyOp> {
...
                                 PatternRewriter &rewriter) const override {
     std::string calleeOrigName;
-    if (apply.getCallee()) {
-      calleeOrigName = apply.getCallee()->getRootReference().str();
+    FunctionType calleeSignature;
+    if (auto callee = apply.getCallee()) {
+      calleeOrigName = callee->getRootReference().str();
+      auto fn =
+          SymbolTable::lookupNearestSymbolFrom<func::FuncOp>(apply, *callee);
+      calleeSignature = fn.getFunctionType();
```
khalatepradnya, line 322: Should we add a nullptr check on `fn` before this call?

include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
schweitzpgi: The trouble with `args` was it was overly general. One can pass "bonus" arguments to apply. These are still arguments, but they are not the actual arguments to the called function on the QPU device, which may necessitate some automatic code generation...

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
schweitzpgi: See the check at invalid.qke:180. The IR verifier will now reject any broken IR where functions are referenced and not declared. There is no need to re-verify that invariant.

PR #4220, opened 2026-03-29
Fix wrong BITWISE XOR

benediktjohannes:


sacpis:
> /ok to test 58a8470

test/AST-Quake/grover.cpp
```diff
@@ -30,7 +30,7 @@ struct run_grover {
   template <typename CallableKernel>
   __qpu__ auto operator()(const int n_qubits, CallableKernel &&oracle,
                           const long target_state) {
-    int n_iterations = round(0.25 * M_PI * sqrt(2 ^ n_qubits));
+    int n_iterations = round(0.25 * M_PI * sqrt(std::pow(2, n_qubits)));
```
sacpis, line 33: Can you please fix the MLIR below? You will need to replace `arith.xori`.

sacpis:
Would like to request to run the test or any relevant file after you make changes.

test/AST-Quake/grover.cpp
sacpis: nit: I don't think you need to specify the namespace `std::`, as already have included `cmath` lib.

PR #4233, opened 2026-04-01
Moving the cudaq-mlir-runtime library under runtime/internal/compiler.

Renaud-K:
These changes move the files that constitute the cudaq-mlir-runtime library under `runtime/internal/compiler` and from the `cudaq` namespace to the `cudaq_internal::compiler` namespace per [CppAPICodingStyle.md](https://github.com/Renaud-K/cuda-quantum/blob/main/CppAPICodingStyle.md).

### DIFF
--- a/runtime/internal/compiler/CMakeLists.txt
+++ b/runtime/internal/compiler/CMakeLists.txt
@@ -13,7 +13,6 @@ include(HandleLLVMOptions)
 add_library(cudaq-mlir-runtime-headers INTERFACE)
 target_include_directories(cudaq-mlir-runtime-headers INTERFACE
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>
 )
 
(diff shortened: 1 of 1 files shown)


runtime/internal/compiler/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+    MLIRLLVMToLLVMIRTranslation
+    cudaq-common
+    cudaq-qir-verifier
+    CUDAQTargetConfigUtil
+    cudaq-logger
+)
+
+install(TARGETS cudaq-mlir-runtime-headers
+        EXPORT cudaq-mlir-runtime-targets
+        DESTINATION include)
```
sacpis: Just wondering if the DESTINATION here is ignored for the INTERFACE libraries, as there are no files to install. `DESTINATION include` seems to incorrect? As the headers are already installed via `install(DIRECTORY ...)`. It seems a bit misleading.

Should we separate out the headers install line and the targets export?

Renaud-K: Thanks Sachin. That `DESTINATION` clause is indeed unneeded. It has no effect. There are also a few other clean-ups that could be made on this file.

schweitzpgi:
I realize this went in already, but these files are not "the compiler" and its sort of confusing to keep calling them "the compiler". I think they are compiler support code that is included in the runtime libraries. `compiler_support` would be more accurate IMHO.

Renaud-K:
I have sought input on this in several PRs: https://github.com/NVIDIA/cuda-quantum/pull/3941, https://github.com/NVIDIA/cuda-quantum/pull/4101 and https://github.com/NVIDIA/cuda-quantum/pull/4188. As well as several internal docs and slack discussions. I find `compiler_support` a bit long. `jit_compiler` maybe?

PR #4237, opened 2026-04-01
[Realtime] Add skip_stream_sweep option to host dispatch loop context

wsttiger:
Allow callers to disable automatic cudaStreamQuery-based worker recycling (sweep_completed_workers) so that worker lifetime can be managed externally via idle_mask.


1tnguyen (approved):
LGTM 👍

bmhowe23:
Direct merging this one now because there was nothing in the queue. (It was already approved.)

PR #4239, opened 2026-04-02
Add additional metrics to resource counting

taalexander:
Additional metrics that are useful (ie., num qubits, depth, 2q depth and two qubit gate counts).

### DIFF
--- a/runtime/nvqir/resourcecounter/ResourceCounter.h
+++ b/runtime/nvqir/resourcecounter/ResourceCounter.h
@@ -56,9 +56,7 @@ public:
   /// @param index 0-based index of qubit to reset
   void resetQubit(const std::size_t index) override {
     resourceCounts.appendInstruction("reset", {},
-                                     {
-                                         index,
-                                     });
+                                     std::vector<std::size_t>{index});
   }
 
   /// @brief Sample the multi-qubit state.


runtime/common/Resources.h
```diff
@@ -62,11 +62,19 @@ class Resources {
   /// @brief Return the total number of operations
   std::size_t count() const;
 
-  /// @brief Append the instruction with the given name and number of controls
-  /// to the resource estimate.
+  /// @brief Append instruction by name and control count. Updates gate
```
atgeller, line 65: At this point, it may make sense to somehow associate the qubits at compilation time with the circuit qubits at runtime to be able to track all this info better. As I recall, this was punted on initially because we didn't have a great use case for it, but it seems like there's more of an argument to doing this now. Thoughts?

runtime/common/Resources.h
```diff
@@ -80,6 +88,26 @@ class Resources {
...
+
+  /// @brief Return the 2-qubit circuit depth (longest chain of 2-qubit
+  /// gates on any qubit path, ignoring single-qubit gates).
+  std::size_t getCircuitDepth2Q() const;
+
+  /// @brief Return the total number of 2-qubit gates.
+  std::size_t getTwoQubitGateCount() const;
+
+  /// @brief Return the per-qubit depth map (all gates).
+  const std::unordered_map<std::size_t, std::size_t> &getPerQubitDepth() const;
```
atgeller, line 105: Since the qubits are all indices starting at 0, I think it would be better for these to be vectors using the qubit as the index, and growing the vector on qubit allocation.

runtime/common/Resources.h
taalexander: Absolutely agree and in fact am already working on this as we speak with the plan of adding into this PR :wink: for an unrelated reason. It turns out this is already done for the Rest backend, but not the simulator resource estimation. However, the simulator resource estimation needs to JIT compile via LLVM which turns out to be SLOW for long circuits and completely ruins benchmarking. On the other hand what the REST backends do is resource estimation at compile time. However, that pass is missing some of the depth metrics I need so I'm working on retrofitting the same metrics there in #4241 and then reproducing the same estimation strategy.

The one piece of nuance is that the way the resource estimation works for dynamic circuits is that the compile-time pass removes all static gates but leaves in the dynamic circuit and then runs those at simulation time. What I'd recommend is we use some heuristics for dynamic circuit compile-time analysis and then also support execution/simulation time reports as a separate path.

Happy to chat more about this.

runtime/common/Resources.h
taalexander: I think this makes it a tad more difficult as we don't encounter them in order necessarily so might need to handle logic for growing vector on the fly. That's simple of course but if it's not much of a performance bottleneck I'd prefer to keep as is for now.

PR #4240, opened 2026-04-02
[CMake] Add cuQuantum Find modules

boschmitt:
PR chain:
- #4238 
- \>\> #4240 \<\< (This)

Add standard CMake find modules for `cuStateVec`, `cuTensor`,
`cuTensorNet`, and `cuDensityMat` that create proper imported targets
(e.g.`cuStateVec::cuStateVec`), replacing ad-hoc find_library/find_path
calls scattered across the tree.


cmake/modules/FindcuStateVec.cmake
```diff
@@ -0,0 +1,71 @@
...
+
+# Sanity check CUDA version
+if(NOT DEFINED CUDAToolkit_VERSION)
+  find_package(CUDAToolkit REQUIRED)
+endif()
+
+set(CUSTATEVEC_MIN_CUDA "11")
+set(CUSTATEVEC_MAX_CUDA "13")
+if(CUDAToolkit_VERSION_MAJOR VERSION_LESS CUSTATEVEC_MIN_CUDA OR
+   CUDAToolkit_VERSION_MAJOR VERSION_GREATER CUSTATEVEC_MAX_CUDA)
```
sacpis: Seems like the `CUSTATEVEC_VERSION` will be empty at this point as the header has not been parsed yet. Should we move this check after version parsing?

runtime/nvqir/cudensitymat/CMakeLists.txt
```diff
@@ -103,34 +34,32 @@ if (${CUDENSITYMAT_VERSION} VERSION_GREATER_EQUAL "0.4")
...
                         cudaq-operator
-                        ${CUDENSITYMAT_LIB}
+                        cuDensityMat::cuDensityMat
                       )
   install(TARGETS ${LIBRARY_NAME} DESTINATION lib)
   add_target_config(dynamics)
 else()
-  message(FATAL_ERROR "Cudensitymat version ${CUDENSITYMAT_VERSION} is not supported. Please use version 0.4.0 or higher.")
+  message(FATAL_ERROR "Cudensitymat version ${cuDensityMat_VERSION} is not supported. Please use version 0.4.0 or higher.")
 endif()
```
sacpis: nit: Seems like we will need to keep this newline.

cmake/modules/FindcuStateVec.cmake
sacpis: Let's apply this in other files as well.

runtime/nvqir/custatevec/CMakeLists.txt
```diff
@@ -6,31 +6,11 @@
...
-
-if(NOT CUSTATEVEC_LIB)
-  message(FATAL_ERROR "\nUnable to find custatevec installation. Please ensure it is correctly installed and set and define CUSTATEVEC_ROOT if necessary (currently set to: ${CUSTATEVEC_ROOT}).")
-endif()
-message(STATUS "CUSTATEVEC_LIB: ${CUSTATEVEC_LIB}")
-
-get_filename_component(CUSTATEVEC_LIB_DIR ${CUSTATEVEC_LIB} DIRECTORY)
-SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${CUSTATEVEC_LIB_DIR}")
-
 macro (nvqir_create_cusv_plugin LIBRARY_NAME SOURCE_FILENAME)
```
sacpis, line 21: The `nvqir-custatevec-kernels` is built before `nvqir_create_cusv_plugin` macro here, and is not linked against `cuStateVec::cuStateVec`. I think the removed `${CUSTATEVEC_ROOT}/include` on line [43](https://github.com/NVIDIA/cuda-quantum/pull/4240/changes#diff-b03093528f1a9429c89b4931576c15e2c92ba28927f6b0482278f8d96e046899L43) below was its only path to `custatevec.h`. Can we please check whether `CuStateVecCircuitSimulator.cu` includes that header. Basically it should not, if it does it is a build regression?

runtime/nvqir/cudensitymat/CMakeLists.txt
```diff
@@ -6,81 +6,12 @@
...
+message (STATUS "cuDensityMat_FOUND - building dynamics NVQIR backend.")
+message(STATUS "Found cudensitymat version: ${cuDensityMat_VERSION}")
 
 # We need cudensitymat v0.4.0
-if (${CUDENSITYMAT_VERSION} VERSION_GREATER_EQUAL "0.4")
-
-  get_filename_component(CUDENSITYMAT_INCLUDE_DIR ${CUDENSITYMAT_INC} DIRECTORY)
-  get_filename_component(CUDENSITYMAT_LIB_DIR ${CUDENSITYMAT_LIB} DIRECTORY)
-  get_filename_component(CUTENSOR_LIB_DIR ${CUTENSOR_LIB} DIRECTORY)
-  SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:${CUDENSITYMAT_LIB_DIR}:${CUTENSOR_LIB_DIR}")
```
sacpis, line 79: Can we quickly do a small check to see whether `.so` files can find their cuQuantum dependencies at runtime on a non-default install path?

cmake/modules/FindcuStateVec.cmake
boschmitt: This is checking the version of the CUDA Toolkit, `CUDAToolkit_VERSION_MAJOR`, using the `CUSTATEVEC_MIN_CUDA`/`CUSTATEVEC_MAX_CUDA` variables which are in the lines just above. It is not checking the package version just yet.

sacpis: My comment was about `CUSTATEVEC_VERSION` which was stated in the error message. If the error is thrown, it will show empty version, no?

runtime/nvqir/cudensitymat/CMakeLists.txt
boschmitt: I think so, yes. Great catch.

runtime/nvqir/custatevec/CMakeLists.txt
boschmitt: Linking this library, line 39, against the imported target will propagate the `INTERFACE_INCLUDE_DIRECTORIES` of the imported target to the consumer's include path. So, linking `cuStateVec::cuStateVec` will give both the library link and the include directory where `custatevec.h` lives.

cmake/modules/FindcuStateVec.cmake
boschmitt: Yes, indeed. The comment is anchored in lines 28 and 29, so I missed the context. I will move the checks (:

runtime/nvqir/cudensitymat/CMakeLists.txt
boschmitt: I don't get this one. Do you want me to add a new line to the end of the file ?

sacpis: Yes, please :)

PR #4241, opened 2026-04-02
Add a circuit optimization target for resource estimation

taalexander:
This PR is based on #4239 and should be rebased after merging. This PR adds support for adding configured pipelines to resource estimation in Cuda-Q when using simulator (non-REST) targets. This enables us to a circuit optimization benchmarking target with a configured optimization pipeline. This is an intermediate step towards exposing better capabilities to python and for initial baselining. 

This PR would greatly benefit for the unified compilation @lmondada is working on in #4198 and maybe should only be merged after being updated with his changes to reduce duplication of pathways. Ideally we would only have one pipeline/pathway through compilation.

### DIFF
--- a/runtime/cudaq/platform/default/circuit-opt-bench.yml
+++ b/runtime/cudaq/platform/default/circuit-opt-bench.yml
@@ -22,9 +22,9 @@ config:
   # Decompose to CX basis, route via SABRE, then convert CX to CZ to match
   # the Benchpress target basis {id, sx, x, rz, cz}.
   # %DEVICE:bypass% defaults to bypass (no routing) when device is not set.
-  # Post-routing SwapToCX decomposes router-inserted SWAPs into 3 CX each,
-  # then CXToCZ converts all CX to CZ as the final 2Q gate.
-  jit-mid-level-pipeline: "func.func(expand-control-veqs,multicontrol-decomposition),decomposition{enable-patterns=SToR1,TToR1,R1ToU3,U3ToRotations,CHToCX,CCXToCCZ,CCZToCX,CRzToCX,CRyToCX,CRxToCX,CR1ToCX,RxAdjToRx,RyAdjToRy,RzAdjToRz,R1AdjToR1,SwapToCX,ExpPauliDecomposition},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{enable-patterns=SwapToCX,CXToCZ}"
+  # Post-routing decomposition converts router-inserted SWAPs and remaining
+  # CX gates to CZ as the final 2Q gate.
+  jit-mid-level-pipeline: "func.func(expand-control-veqs,multicontrol-decomposition),decomposition{basis=h,rx,ry,rz,x,x(1)},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{basis=h,rx,ry,rz,x,z(1)}"
   # Cleanup: symbol-dce removes wire_set ops left by add-wireset (prevents
   # the JIT from taking the wireset-to-profile QIR path).
   jit-low-level-pipeline: "func.func(apply-control-negations,canonicalize,cse),symbol-dce"


lmondada:
Thanks for this change! I've pressed merge for #4198 in case that helps, but presumably you'll need more than that to actually deduplicate your work...

I'll have a closer look when I'm back on Tuesday next week. By when do you want this in? Happy to work on this first thing as soon as I'm back if that unblocks you and lets us do this 'the right way'. If you need this immediately then go ahead and merge this, I'll deduplicate later.

taalexander:
I'm really not too worried about landing this as I'm unblocked locally. Just trying to push reasonable chunks of work up.

Happy to wait for you to get back to discuss how we'd rather land this to main. Ultimately I don't want to be using these hacky targets period.

lib/Optimizer/Transforms/ResourceCount.cpp
```diff
@@ -0,0 +1,45 @@
...
+  if (failed(pm.run(module)))
+    return failure();
+
+  // Count allocated qubits from the IR.
+  std::size_t allocated = 0;
+  module.walk([&](quake::AllocaOp alloc) {
+    if (isa<quake::RefType>(alloc.getType()))
+      allocated++;
+    else if (auto veqTy = dyn_cast<quake::VeqType>(alloc.getType()))
+      if (veqTy.hasSpecifiedSize())
```
atgeller: If a veq doesn't have a constant size, should we fail here since we would have failed to successfully count all qubits? If not, should we remove the veq (assuming it has no more uses), to prevent double counting later on?

taalexander: Great points. I added a shared `getVeqSize` helper that looks through RelaxSizeOp to resolve the inner sized type a8d39da. The walk now returns failure for unresolvable dynamic veqs instead of silently skipping them.

lib/Optimizer/Transforms/ExpandControlVeqs.cpp
```diff
@@ -38,16 +38,24 @@ class ExpandPat : public OpRewritePattern<OP> {
...
+      if (isa<quake::VeqType>(control.getType())) {
+        auto size = quake::getVeqSize(control);
+        if (!size)
           return failure();
 
+        // Use the inner sized veq for extract_ref when looking through
+        // RelaxSizeOp (extract_ref needs a sized veq operand).
+        Value veqVal = control;
+        if (auto relaxOp =
+                dyn_cast_or_null<quake::RelaxSizeOp>(control.getDefiningOp()))
```
atgeller: Nit: I believe you can use `control.getDefiningOp<quake::RelaxSizeOp>()` here instead of the getDefiningOp and dyn_cast

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
```diff
@@ -35,6 +36,42 @@ struct ResourceCountPreprocessPass
...
+  std::size_t nextQubitIndex = 0;
+
+  /// Assign a globally unique base index for a qvector Value.
+  std::size_t getVeqBase(Value veq) {
+    auto it = qubitIndexMap.find(veq);
+    if (it != qubitIndexMap.end())
+      return it->second;
+    auto base = nextQubitIndex;
+    if (auto size = quake::getVeqSize(veq))
+      nextQubitIndex += *size;
```
atgeller, line 52: The doc comment suggests this function should return a globally unique base index for a qvector, but if the qvector size is unknown it looks like it will reuse `nextQubitIndex` as the index for several qvectors. I assume this is not an issue because such qvectors can be reasoned about in general, so it doesn't matter that we don't have a unique index for them, but it would be good to say that explicitly.

taalexander:
After the merger of https://github.com/NVIDIA/cuda-quantum/pull/4292 this target should be updated to use.

lib/Optimizer/Transforms/ExpandControlVeqs.cpp
taalexander: Good catch. The original failed to compile because `control` is a dependent type in the template class `ExpandPat<OP>`. Fixed with the `template` keyword (TIL).

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
taalexander: Updated the doc comment to clarify the behavior for unknown-size veqs. For unsized veqs, the base index is shared (`nextQubitIndex` is not advanced) since individual qubits cannot be resolved.

lib/Optimizer/Transforms/ResourceCount.cpp
```diff
@@ -0,0 +1,62 @@
...
+  // createResourceCountPreprocess runs as addNestedPass<func::FuncOp>, which
+  // MLIR executes in parallel across functions. Disable threading for this
+  // PassManager so the callback is called sequentially.
+  auto *ctx = module.getContext();
+  bool wasThreadingEnabled = ctx->isMultithreadingEnabled();
+  ctx->disableMultithreading();
+  PassManager pm(ctx);
+  pm.addNestedPass<func::FuncOp>(createResourceCountPreprocess(opt));
+  pm.addPass(createCanonicalizerPass());
+  if (failed(pm.run(module))) {
```
lmondada: nit: I'd prefer having the pass definition first, then the enable-disable logic in close succession
```c++
ctx->disableMultithreading();
pm.run(module);
if (wasThreadingEnabled)
    ctx->enableMultithreading();
```
and handle pass failure after that. That way you'd only need a single `if (wasThreadingEnabled)` clause.

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
```diff
@@ -35,6 +36,45 @@ struct ResourceCountPreprocessPass
...
+    qubitIndexMap[veq] = base;
+    return base;
+  }
+
+  /// Resolve a quake value to a globally unique qubit index.
+  std::optional<std::size_t> resolveQubitIndex(Value v) {
```
lmondada, line 58: I'm surprised this is the first time a logic like this is needed in our codebase -- and wonder if there is a better place for this to live in. Maybe @schweitzpgi has a more concrete suggestion?

python/tests/builder/test_resource_metrics.py
```diff
@@ -61,27 +64,12 @@ def test_parallel_cx_disjoint_qubits():
     resources = cudaq.estimate_resources(kernel)
     assert resources.num_qubits == 4
     assert resources.depth == 1
-    assert resources.two_qubit_gate_count == 2
-    assert resources.depth_2q == 1
-
-
-def test_ccx_not_counted_as_two_qubit():
```
lmondada, line 68: Any reason this test was deleted? It's not particularly useful, but I see no harm in leaving it

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -101,6 +110,79 @@ specializeKernel(const std::string &name, ModuleOp module,
     throw std::runtime_error("Could not successfully apply argument synth.");
 }
 
+/// Replace %KEY% and %KEY:default% placeholders in a pipeline string with
+/// values from the runtime config map. If the key is in runtimeConfig, use
+/// that value. Otherwise use the inline default if provided (%KEY:val%).
+/// Keys in the pipeline are uppercase; runtimeConfig keys are lowercase.
+/// This is the Python JIT equivalent of ServerHelper::updatePassPipeline().
+static void substitutePipelinePlaceholders(
```
lmondada, line 118: As you've said, it would be very good indeed to be able to just rely on one version of this, which will be possible when we have one compiler pipeline (e.g. in this case not dependent on the implementation of the REST QPU backends).

But this will be easy to clean up later, so I am fine with merging this in.

python/tests/builder/test_resource_metrics.py
taalexander: I ended up changing how these were counted. There's some different counting now (for each of the arities). IIRC I had added this in my last PR.

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
taalexander: I had looked! If there's a better version I'd be keen to use it. I'll look again.

lib/Optimizer/Transforms/ResourceCount.cpp
taalexander: Fair enough. I'll try it out!

runtime/cudaq/platform/default/python/QPU.cpp
taalexander: Great!

python/tests/builder/test_resource_metrics.py
taalexander: Added it back

runtime/cudaq/platform/default/circuit-opt-bench.yml
```diff
@@ -0,0 +1,31 @@
...
+config:
+  nvqir-simulation-backend: qpp
+  # Synthesis: loop unrolling, matrix concretization, unitary synthesis.
+  jit-high-level-pipeline: "func.func(cc-loop-normalize,cc-loop-unroll,canonicalize,lift-array-alloc),globalize-array-values,func.func(canonicalize),get-concrete-matrix,unitary-synthesis"
+  # Decompose to CX basis, route via SABRE, then convert CX to CZ to match
+  # the Benchpress target basis {id, sx, x, rz, cz}.
+  # %DEVICE:bypass% defaults to bypass (no routing) when device is not set.
+  # Post-routing SwapToCX decomposes router-inserted SWAPs into 3 CX each,
+  # then CXToCZ converts all CX to CZ as the final 2Q gate.
+  jit-mid-level-pipeline: "func.func(expand-control-veqs,multicontrol-decomposition),decomposition{enable-patterns=SToR1,TToR1,R1ToU3,U3ToRotations,CHToCX,CCXToCCZ,CCZToCX,CRzToCX,CRyToCX,CRxToCX,CR1ToCX,RxAdjToRx,RyAdjToRy,RzAdjToRz,R1AdjToR1,SwapToCX,ExpPauliDecomposition},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{enable-patterns=SwapToCX,CXToCZ}"
```
atgeller: Do we want to add the phase folding opt somewhere :) between factor-quantum-alloc and memtoreg may actually just work.

lib/Optimizer/Transforms/ResourceCountPreprocess.cpp
atgeller: The `assign-wire-indices` pass was meant to add exactly this logic, but it's pretty restricted and rather old at this point.

runtime/cudaq/platform/default/circuit-opt-bench.yml
taalexander: Discussed we willa dd after baseline.

lmondada:
ah one more thing: can you add an issue about re-merging the pipeline parsing logic (and maybe a second one to centralise the resource counting logic). Will make it slightly easier to keep track of this.

PR #4245, opened 2026-04-02
Factory only dependency of JitEngine on cudaq-mlir-runtime

Renaud-K:
Apologies to the extra people that where added to the review. I think I had a dirty branch with many more changed files.

With this [move](https://github.com/NVIDIA/cuda-quantum/pull/4233), we needed to introduce `cudaq-mlir-runtime-headers`. The `cudaq-common` library has a dependency on these headers. This is because the `ExecutionContext` is including the `JITEngine` and almost every file includes the `ExecutionContext`.

This defeats the purpose of reducing the [visibility of module headers](https://github.com/NVIDIA/cuda-quantum/pull/4188) since it forces every part of the runtime to include them. 

This PR addresses this issue by implementing the methods of the `JitEngine` as closures that are defined on the factory side (cudaq-mlir-runtime). This allows us to move the `JitEngine` in a lower level header (`CompiledKernel.h) since any calls from it have no link-time dependency with `cudaq-mlir-runtime`.

### DIFF
--- a/runtime/common/CompiledKernel.h
+++ b/runtime/common/CompiledKernel.h
@@ -8,7 +8,10 @@
 #pragma once
 
 #include "common/ThunkInterface.h"
-#include "cudaq_internal/compiler/JIT.h"
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <memory>
 #include <optional>
 #include <string>
 #include <vector>
@@ -23,9 +26,48 @@
 namespace mlir {
 class Type;
 class ModuleOp;
+class ExecutionEngine;
 } // namespace mlir
 
 namespace cudaq {
+class ResultInfo;
+} // namespace cudaq
+
+namespace cudaq_internal::compiler {
+cudaq::ResultInfo createResultInfo(mlir::Type resultType, bool isEntryPoint,
+                                   mlir::ModuleOp module);
+} // namespace cudaq_internal::compiler
+
+namespace cudaq {
+
+/// JitEngine is a type-erased class that is wrapping an mlir::ExecutionEngine
+/// without introducing any link time dependency on MLIR for the client of the
+/// class. Memory management for of the mlir::ExecutionEngine is handled
+/// internally.
+class JitEngine {
+  using RawFnPtr = void (*)();
+  using LookupFn = std::function<RawFnPtr(const std::string &)>;
+
+  struct Base {
+    LookupFn lookupFn;
+    std::function<void(const std::string &)> runFn;
+  };
+
+public:
+  JitEngine(std::unique_ptr<mlir::ExecutionEngine>);
+
+  void run(const std::string &kernelName) const { impl->runFn(kernelName); }
+
+  void (*lookupRawNameOrFail(const std::string &kernelName) const)() {
+    return impl->lookupFn(kernelName);
+  }
+
+  std::size_t getKey() const;
+
+private:
+  class Impl;
+  std::shared_ptr<Base> impl;
+};
 
 /// Pre-computed result metadata, set at build time. Used at execution time
 /// for result buffer allocation and type conversion. Construct via
@@ -65,8 +107,6 @@ public:
 /// after construction.
 class CompiledKernel {
 public:
-  using JitEngine = cudaq_internal::compiler::JitEngine;
-
   // --- Construction ---
 
   CompiledKernel(std::string kernelName, ResultInfo resultInfo);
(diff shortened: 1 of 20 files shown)


runtime/internal/compiler/include/cudaq_internal/compiler/JIT.h
```diff
@@ -46,15 +46,28 @@ createWrappedKernel(std::string_view llvmIr, const std::string &kernelName,
...
+  void run(const std::string &kernelName) const { impl->runFn(kernelName); }
+
+  void (*lookupRawNameOrFail(const std::string &kernelName) const)() {
+    return impl->lookupFn(kernelName);
+  }
+
   std::size_t getKey() const;
 
 private:
   class Impl;
```
lmondada: I think your changes make this forward declaration redundant. It should be enough to have `class Impl` live within the compilation unit?

lmondada:
This is a step in the right direction, I'm in favour. Looking down the road, I think it might be possible in the future to replace most uses of `JitEngine` with `CompiledKernel`. At that point we would no longer need to expose `JitEngine` publicly.

runtime/internal/compiler/include/cudaq_internal/compiler/JIT.h
Renaud-K: It is counter-intuitive but Impl class is needed here although it is only used on the implementation side. It is a C++ requirement according to Claude. I have tried removing as well. Having a `Base` and `Impl` classes allows for any number of closures to be added without increasing the size of the object. It remains, the size of a `std::shared_ptr`.

Renaud-K:
> /ok to test 684951bb4bfc03fa81aa18e64224d362a48cbd66

runtime/common/CompiledKernel.h
```diff
@@ -8,10 +8,59 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <memory>
 #include <string>
 #include <vector>
 
+namespace mlir {
+class ExecutionEngine;
+} // namespace mlir
```
lmondada, line 30: When you merge in the changes from #4198, you can merge the MLIR forward declarations into one block

runtime/common/CompiledKernel.h
```diff
@@ -8,10 +8,59 @@
...
+  std::size_t getKey() const;
+
+private:
+  class Impl;
+  std::shared_ptr<Base> impl;
+};
+
+} // namespace cudaq
+
+namespace cudaq_internal::compiler {
```
lmondada: nit: This forward declaration is in a slightly odd place, wedged between `JitEngine` and `CompiledKernel` (which forces you to end the namespace block and then start a new one). Can you move it to the bottom or the top?

runtime/common/CompiledKernel.h
```diff
@@ -54,8 +101,9 @@ class CompiledKernel {
 
   // Use the following factory function (compiled into cudaq-mlir-runtime) to
   // construct CompiledKernels.
+  // TODO: Make this a friend of a factory that can be forward declared.
```
lmondada: Sorry, I don't understand this TODO. Could you expand on what you mean?

runtime/internal/compiler/include/cudaq_internal/compiler/JIT.h
```diff
@@ -41,32 +36,16 @@ std::tuple<std::unique_ptr<llvm::orc::LLJIT>, std::function<void()>>
...
-
-private:
-  class Impl;
-  std::shared_ptr<Impl> impl;
-};
-
 /// Lower ModuleOp to QIR/LLVM IR and create a JIT execution engine.
-JitEngine createQIRJITEngine(mlir::ModuleOp &moduleOp,
-                             llvm::StringRef convertTo);
+cudaq::JitEngine createQIRJITEngine(mlir::ModuleOp &moduleOp,
```
lmondada: Could we rename this to `createJITEngine`? I think the QIR bit is redundant and hard to read.

runtime/common/CompiledKernel.h
Renaud-K: The file had an 
```
#include "cudaq_internal/compiler/JIT.h"
```
which I have removed. It means that `createCompiledKernel` now has to be forward declared. But it takes a JitEngine and returns a CompiledKernel. Which means these 2 have to be declared before. I guess I can forward declare all 3 before.

runtime/common/CompiledKernel.h
Renaud-K: The forward declaration is a bit verbose:
```
friend cudaq::CompiledKernel cudaq_internal::compiler::createCompiledKernel(
      JitEngine engine, std::string kernelName, bool hasResult,
      bool isFullySpecialized);
```

Maybe instead, we could just have
```
friend class cudaq_internal::compiler::CompiledKernelFactory;
```

runtime/common/CompiledKernel.h
lmondada: ah I see. Up to you :)

Renaud-K:
> /ok to test 015584a781714c402bedc819403d6f257f3cf822

runtime/internal/compiler/include/cudaq_internal/compiler/JIT.h
Renaud-K: Done.

runtime/common/CompiledKernel.h
Renaud-K: Sure.

PR #4247, opened 2026-04-03
[CUDA-Q model] Use cudaq::state when dealing with state.

schweitzpgi:
We're evolving the CUDA-Q model such that amplitude vectors and density matrices are uniformly wrapped in `cudaq::state` objects. This change makes the CUDA-Q code easier to read and understand as a programming model and also greatly simplifies other components, which would otherwise need to be both generalized and specialized for logically disjoint use cases.

While amplitude vector data may still be passed-by-value as `std::vector<T>` data to a kernel, it is considered poor style. This PR is the first step towards enforcing quantum state to be wrapped in `cudaq::state` objects when passed to kernels, which as noted above will result in better readability as well as improved performance.

### DIFF
--- a/docs/notebook_validation.py
+++ b/docs/notebook_validation.py
@@ -62,6 +62,7 @@ LONG_RUNNING_NOTEBOOKS = [
     "qm_mm_pe.ipynb",
     "qsci.ipynb",
     "uccsd_wf_ansatz.ipynb",
+    "unitary_compilation_diffusion_models",
     "vqe_advanced.ipynb",
 ]
 


runtime/cudaq/qis/qudit.h
```diff
@@ -31,31 +31,6 @@ class qudit {
   /// Construct a qudit, will allocated a new unique index
   qudit() : idx(getExecutionManager()->allocateQudit(n_levels())) {}
 
-  qudit(const std::vector<complex> &state) : qudit() {
```
sacpis, line 34: nit: As it is a breaking change, I think we should state it in the PR description?

sacpis (approved):
LGTM. Thanks @schweitzpgi.

sacpis:
Seems like there is a test failure in IQM [here](https://github.com/NVIDIA/cuda-quantum/actions/runs/23949961367/job/69855582025?pr=4247#step:14:6909).

runtime/cudaq/qis/qudit.h
schweitzpgi: It's not breaking anything though from the user's perspective. clang++ will insert automatic conversions.

1tnguyen (approved):
LGTM 👍

PR #4257, opened 2026-04-04
Enabling the qpp_H2MoleculeTester tests to run unprivileged

Renaud-K:
I would like to be able to run these tests as an unprivileged user. If you do so, you will hit the following error:
```
C++ exception with description "PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.12/dist-packages/openfermion/testing/data/H2_sto-3g_singlet.hdf5'
```
Bu default `MolecularData` save `.hdf5` at the above location. Providing the `data_directory` argument allows for this file to be saved in the working directory.

### DIFF
--- a/python/cudaq/domains/chemistry/__init__.py
+++ b/python/cudaq/domains/chemistry/__init__.py
@@ -6,6 +6,8 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
+DATA_DIRECTORY = '.'
+
 
 def tryImport():
     """Import `openfermion` and `openfermionpyscf`."""
@@ -22,8 +24,7 @@ def create_molecular_hamiltonian(geometry: list,
                                  multiplicity=1,
                                  charge=0,
                                  n_active_electrons=None,
-                                 n_active_orbitals=None,
-                                 data_directory='.'):
+                                 n_active_orbitals=None):
     '''
     Create the molecular Hamiltonian corresponding to the provided 
     geometry, basis set, multiplicity, and charge.  One can also specify the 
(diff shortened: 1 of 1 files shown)


python/cudaq/domains/chemistry/__init__.py
```diff
@@ -46,8 +46,11 @@ def create_molecular_hamiltonian(geometry: list,
       Hamiltonian and the raw molecular data. 
     '''
     of, ofpyscf = tryImport()
-    molecule = ofpyscf.run_pyscf(of.MolecularData(geometry, basis, multiplicity,
-                                                  charge),
+    molecule = ofpyscf.run_pyscf(of.MolecularData(geometry,
+                                                  basis,
+                                                  multiplicity,
+                                                  charge,
+                                                  data_directory='.'),
```
sacpis: Should we make data_directory a parameter? Or Let's define it at the top as a global var?

Renaud-K: I made it a default parameter.

sacpis: Yes, I meant if we can pull it out as a constant (global) we won't need to change the default parameter in all places. Just will have to replace at one place.

Renaud-K: I made it a global this time.

sacpis: Have you pushed the change?

Renaud-K: Just did.

sacpis (approved):
LGTM. Thanks @Renaud-K.

Issue #2341, opened 2024-11-01
[Python] Aliasing cudaq name when importing it


boschmitt:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

In Python, it is quite common to alias the name of imported libs, e.g., `import ... as ...`. It is common practice even for libraries which don't have big names: `import numpy as np`. If one tries to alias CUDA-Q import, things break.

### Steps to reproduce the bug

```python
import cudaq as quda

@quda.kernel
def simple():
    q = quda.qubit()

counts = quda.sample(simple)
print(counts)
```

Result:
```
cudaq.kernel.ast_bridge.CompilerError: bug0.py:5: error: Invalid function call - 'quda' is unknown.
	 (offending source -> quda.qubit)
```

### Expected behavior

Correct execution. Users should be able to alias `cudaq`.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA Quantum version**: proto-0.8.0 (https://github.com/NVIDIA/cuda-quantum a9e2324f6504f02e47a87e1de82c8f31cbb0f46b)
- **Python version**: 3.12
- **Operating system**: ubuntu 24.04

mitchdz:
This should be possible with a small change. Let me take a stab at it.

mitchdz:
If we are doing this, we might as well allow full aliasing such as:
```python3
from cudaq import kernel as k, qubit as qb

@k
def f():
    q = qb()

f()
```

boschmitt:
> This should be possible with a small change. Let me take a stab at it.

I'm don't think it will be a small change. The strings `cudaq`, `numpy`, `np`, `qubit` and friends are hardcoded in the bridge. In the end this is a symptom in of a deeper issue. Given your example, which I agree it should be supported:

```python
from cudaq import kernel as k, qubit as qb

@k
def f():
    q = qb()

f()
```

[...]

PR #4258, opened 2026-04-06
[Python] Support aliased cudaq imports in kernel definitions

mitchdz:
Solves https://github.com/NVIDIA/cuda-quantum/issues/2341

  ## Problem
                                                                                                                                                                                                          
  Users couldn't alias the cudaq module with import cudaq as cq because the kernel compiler's AST bridge had the string 'cudaq' hardcoded in 7 locations across 4 files. Any code using an alias (e.g.,   
  cq.qubit(), cq.qvector(2)) would fail with CompilerError: 'cq' is unknown.                                                                                                                              
                                                                                                                                                                                                          
  ## Solution                                                           
  kernel_decorator.py                                                                                                                                                                                     
                                                                  
  - At kernel creation time, scans both f_locals and f_globals of the parent frame for any variable that is the cudaq module (via isinstance(var, types.ModuleType) and var.__name__ == 'cudaq'). Collects
   all alias names into self.cudaqAliases set.
  - Passes cudaqAliases through to KernelSignature.parse_from_ast() and compile_to_mlir().                                                                                                                
                                                                                                                                                                                                          
  ast_bridge.py
                                                                                                                                                                                                          
  - PyASTBridge.__init__ accepts and stores cudaqAliases.                                                                                                                                                 
  - New isCudaqName(name) method returns True if name is 'cudaq' or a known alias.
  - 6 hardcoded checks replaced: nested kernel detection, noise channel validation, decorator call detection, cudaq.complex/qvector/amplitudes handling, and Kraus channel detection.                     
                                                                                                                                                                                                          
  utils.py                                                                                                                                                                                                
                                                                                                                                                                                                          
  - mlirTypeFromAnnotation() accepts optional cudaqAliases parameter, uses it to recognize aliased type annotations like cq.qvector, cq.qubit, cq.pauli_word.                                             
   
  kernel_signature.py                                                                                                                                                                                     
                                                                  
  - parse_from_ast() and _mlir_type_from_annotation() thread cudaqAliases through to mlirTypeFromAnnotation.                                                                                              
   
  test_cudaq_alias.py (new)                                                                                                                                                                               
                                                                  
  - 13 tests covering: basic kernel, qvector, int/float args, type annotations, canonical+alias coexistence, gates, rotation gates, adjoint, control, return values, and multiple allocations.            
   
  Test Results                                                                                                                                                                                            
                                                                  
  - 13/13 new alias tests pass                                                                                                                                                                            
  - 597 existing kernel tests pass (2 skipped, 1 xfail) with no regressions


python/cudaq/kernel/kernel_signature.py
sacpis, line 213: Will need to update these with `cudaqAliases=None`.

python/tests/kernel/test_cudaq_alias.py
```diff
@@ -0,0 +1,235 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+import pytest
+import numpy as np
```
sacpis: nit: We are not using `numpy` here so can be removed.

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -179,9 +180,22 @@ def __init__(self,
...
+            # Collect all names that refer to the cudaq module so the AST
+            # bridge can recognize them alongside the canonical 'cudaq' name.
+            # Check both local and global scope since the alias may be at
+            # module level while the kernel is defined inside a function.
+            self.cudaqAliases = {'cudaq'}
+            for scope in (parentVars, self.parentFrame.f_globals):
+                for vname, var in scope.items():
+                    if (isinstance(var, types.ModuleType) and
+                            getattr(var, '__name__', None) == 'cudaq' and
+                            vname != 'cudaq'):
```
sacpis: nit: Redundant guard `vname != 'cudaq'` as `cudaq` is already in `cudaqAliases`.

sacpis (approved):
Overall LGTM. Thanks @mitchdz.

python/cudaq/kernel/kernel_signature.py
mitchdz: Nice catch. Fixed in [fe38f23](https://github.com/NVIDIA/cuda-quantum/pull/4258/commits/fe38f23e21e2b3b801969771a31370fabb516091)

PR #4261, opened 2026-04-06
Use pinned host memory for GPU state-to-NumPy conversion (#2797)

### LINKED ISSUE #2797: Potential performance issue when doing np.array on statevector representation
### Required prerequisites

### Describe the bug

When retrieving the statevector with `cudaq.get_state(kernel)`, converting the result to an np.array seems to take a long time.

I have the following python file named test_np_array.py:
```python3
import sys
import cudaq
import argparse
import numpy as np

parser = argparse.ArgumentParser(description="np array timing script with --np-array and --qubits flags.")

parser.add_argument('--np-array', action='store_true', default=False,
                    help='Use NumPy array if this flag is set')

parser.add_argument('--qubits', type=int, default=1,
                    help='Number of qubits to use (default: 1)')

args = parser.parse_args()


cudaq.set_target('nvidia', option='mgpu')

print(f"Running on target {cudaq.get_target().name}")
qubit_count = args.qubits

print(f"With {qubit_count} qubits")

@cudaq.kernel
def kernel():
    qubits = cudaq.qvector(qubit_count)
    h(qubits[0])
    for i in range(1, qubit_count):
        x.ctrl(qubits[0], qubits[i])
    mz(qubits)

#    print(result)
re = cudaq.get_state(kernel)
if args.np_array:
    print("Doing np.array:")
    # Do something new here
    sv = np.array(re)
```

Testing this on GH200 with 32 qubits, I see:
```
$ time python3 test_np_time.py --qubits 32
Running on target nvidia
With 32 qubits

real    0m15.617s
user    0m11.144s
sys     0m12.847s
$ time python3 test_np_time.py --qubits 32 --np-array
Running on target nvidia
With 32 qubits
Doing np.array:

real    0m20.968s
user    0m13.812s
sys     0m15.527s
```

### Expected behavior

I would expect np.array to be quick in this scenario, adding a 5s delay seems excessive.

Not Sure.

### Environment

- **CUDA-Q version**: nvq++ Version cu12-0.10.0 (https://github.com/NVIDIA/cuda-quantum 857dd2ce0a783c32416af8fba8664ff30f9ddc47)
- **CUDA-Q docker container**: cu12-0.10.0 container

### Suggestions

_No response_

mitchdz:
## Summary

  - Fix performance bottleneck in `np.array(cudaq.get_state(kernel))` for GPU-backed
    states by using pinned (`cudaMallocHost`) host memory instead of pageable (`new[]`)
    for the device-to-host transfer in the pybind11 buffer protocol handler
  - Add `SimulationState::toHostBuffer()` virtual method so GPU backends can provide
    optimized host allocation, with a pageable fallback for non-GPU backends and for
    systems that cannot pin the requested amount of memory
  - Fix `to_cupy()` passing a hardcoded 1024-byte size to `UnownedMemory` instead of
    the actual buffer size

  ## Motivation

  Closes #2797. When `np.array()` is called on a GPU-resident state, the buffer
  protocol handler allocates pageable host memory and copies via `cudaMemcpy`. With
  pageable destinations, CUDA must stage through an internal pinned buffer in chunks,
  severely limiting effective bandwidth. On an L4 GPU with 25 qubits (537 MB state):

  | Allocation | Bandwidth | Time   |
  |-----------|-----------|--------|
  | Pageable (before) | 1.8 GB/s | 0.29s |
  | Pinned (after)    | 13.2 GB/s | 0.04s |

  For the original reporter's 32-qubit case on GH200, the ~5.4s overhead should drop
  to under 1s.

### DIFF
--- a/python/extension/CMakeLists.txt
+++ b/python/extension/CMakeLists.txt
@@ -110,6 +110,7 @@ declare_mlir_python_extension(CUDAQuantumPythonSources.Extension
    cudaq-python-interop
    cudaq-platform-default
    cudaq-qir-verifier
+   cudaq-mlir-runtime-headers
 )
 
 target_include_directories(CUDAQuantumPythonSources.Extension INTERFACE 
--- a/python/runtime/cudaq/dynamics/CMakeLists.txt
+++ b/python/runtime/cudaq/dynamics/CMakeLists.txt
@@ -34,7 +34,6 @@ target_include_directories(nvqir_dynamics_bindings
 target_link_libraries(nvqir_dynamics_bindings PRIVATE
     cudaq-logger
     fmt::fmt-header-only
-    cudaq-mlir-runtime-headers
 )
 # Set output directory for ctest-based python test invocation, which uses cudaq python from the build directory. 
 set_target_properties(nvqir_dynamics_bindings PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/python/cudaq/dynamics)
--- a/python/runtime/interop/PythonCppInterop.cpp
+++ b/python/runtime/interop/PythonCppInterop.cpp
@@ -7,7 +7,7 @@
  ******************************************************************************/
 
 #include "PythonCppInterop.h"
-#include "cudaq.h" // unfortunately, cudaq::get_quake is here at top level
+#include "common/DeviceCodeRegistry.h"
 #include "cudaq/utils/cudaq_utils.h"
 
 cudaq::python::CppPyKernelDecorator::~CppPyKernelDecorator() {
--- a/runtime/common/BraketExecutor.h
+++ b/runtime/common/BraketExecutor.h
@@ -11,7 +11,6 @@
 #include "common/BraketServerHelper.h"
 #include "common/Executor.h"
 #include "common/SampleResult.h"
-#include "cudaq.h"
 #include "cudaq/runtime/logger/logger.h"
 #include <aws/braket/BraketClient.h>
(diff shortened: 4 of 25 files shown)


mitchdz:
> /ok to test 270b2a0

runtime/cudaq/qis/state.h
```diff
@@ -143,6 +143,12 @@ class state {
     internal->toHost(hostPtr, numElements);
   }
 
+  /// @brief Allocate a host buffer and copy state data from device.
+  /// GPU backends use pinned memory for optimal transfer speed.
+  SimulationState::HostBuffer toHostBuffer(std::size_t numElements) const {
```
1tnguyen: nit: perhaps we should use `snake_case` here to be consistent with the other APIs.

1tnguyen:
We'll need to update the downstream simulator to use pinned memory as well.

runtime/cudaq/qis/state.h
mitchdz: Done in [e4d8f99](https://github.com/NVIDIA/cuda-quantum/pull/4261/commits/e4d8f9980b92dcf36c11beffeb0a5911fa43a18d), thanks!

mitchdz:
Good call on the downstream simulator. Once this gets in, I'll go ahead and update that. This shouldn't break the simulator so I'm fine doing it asynchronously.

mitchdz:
Let's wait for https://github.com/NVIDIA/cuda-quantum/actions/runs/24163000317 to finish (and then go through publishing)

PR #4265, opened 2026-04-06
Directly invoke locally simulated python kernels with args

atgeller:
This PR removes argument synthesis by default for Python kernels run on the local simulator, instead directly invoking them with the arguments (currently, by constructing a message buffer through `.argsCreator` which is passed to the kernel's `thunk`). This only affects entry point kernels.

Benefits:
1. This makes it unnecessary to recompile kernels for different arguments in this setting, simplifying the `reuse_compiler_artifacts` logic.
2. It aligns the python local simulation path more closely with C++, where arguments are similarly not synthesized.
3. As a result of 1 and 2, it is a useful and important first step towards an inter-launch caching strategy for python.


runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -26,23 +26,23 @@
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Target/LLVMIR/Export.h"
 #include "mlir/Transforms/Passes.h"
+#include <cudaq/Optimizer/CodeGen/QIROpaqueStructTypes.h>
 #include <unordered_set>
```
sacpis: nit: Don't think we need this `unordered_set` anymore.

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
```diff
@@ -113,6 +113,11 @@ static constexpr const char QIRTupleRecordOutput[] =
 static constexpr const char QIRArrayRecordOutput[] =
     "__quantum__rt__array_record_output";
 
+// NVQPP span logging helpers for dynamic-size vector return types.
```
schweitzpgi, line 123: I don't get these changes.

The logging functions are for QIR {base,adaptive} profile code generation. They are very explicitly _not_ for return values. That is, `cudaq::run` is semantically broken. It's "return value" is _not_ a return value but a lie. Kernels with `return expression;` statements when called via `cudaq::run` _do not behave correctly_ and convert the return results into calls to output logging functions. It needs to be fixed.

But more to the point, CUDA-Q kernels can and should have _legitimate_ return values. Those have nothing to do with output logging. (Just like calling a function in C++ that returns a `std::vector<int>` has nothing to do with getting output printed to `std::cout`.)

Also, the output logging must follow a very specific format for the log parser to interpret the logs. It is simply not possible to dump variable length output due to the limitations of QIR. Specifically, we cannot use memory, so we cannot create a buffer in which to create the output logging message that is required.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -238,32 +220,31 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
-    const bool isFullySpecialized = varArgIndices.empty();
+    const bool hasResult = !!resultTy;
     auto resultInfo = createResultInfo(resultTy, isEntryPoint, module);
 
+    // Determine whether the kernel needs argument packing (argsCreator) by
+    // checking if any non-callable arguments are present. This must be done
+    // before the cache lookup so the cached path uses the correct value.
+    bool isFullySpecialized = true;
+    FunctionType fromFuncTy = funcOp.getFunctionType();
+    auto closureArgs = rawArgs;
```
sacpis: Seems like we can create a copy inside the conditional below (if loop). We can avoid redundant heap allocation and memcpy on every iteration for `run` and repeated `sample` calls.

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
atgeller: Yeah the comment and names are wrong here. It should be `float_span_record_output` or something like that. These are logging functions to record the values, nothing to do with returning values.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -191,21 +191,26 @@ void cudaq::handleStructMemberVariable(void *data, std::size_t offset,
...
+            };
 
         mlir::TypeSwitch<mlir::Type, void>(ty.getElementType())
             .Case([&](mlir::IntegerType type) {
               if (type.isInteger(1)) {
-                appendVectorValue(value, data, offset, char());
+                if (specialize)
+                  appendVectorValue(value, data, offset, char());
+                else
+                  appendVectorValue(value, data, offset, bool());
```
schweitzpgi: Why do we have both cases here? (specialize is confusing.)

- Sequences of bool _on the device_ are 1-D arrays of bytes.
- std::vector<bool> is _never_ on the device.

The `.argsCreator` function is always called on the host. The `.thunk` function is always called on the device. I suspect that we're trying to rig around that fundamental difference here?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -375,7 +404,10 @@ void cudaq::packArgs(OpaqueArguments &argData, py::list args,
         .Case([&](IntegerType ty) {
           if (ty.getIntOrFloatBitWidth() == 1) {
             checkArgumentType<py::bool_>(arg, i);
-            addArgument(argData, static_cast<char>(arg.cast<bool>()));
+            if (specialize)
```
schweitzpgi: I suspect that this really should be using parametric template rather than shuttling a special extra invariant argument all over.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -695,17 +735,27 @@ py::object cudaq::convertResult(ModuleOp module, Type ty, char *data) {
         return readPyObject<float>(ty, data);
       })
       .Case([&](cudaq::cc::StdvecType ty) -> py::object {
+        auto eleTy = ty.getElementType();
+        // Nested StdvecType elements have a different in-memory size than
+        // scalar types: span ({ptr,size_t} = 16 bytes) in direct-call context,
+        // std::vector ({ptr,ptr,ptr} = 24 bytes) in run context.
```
schweitzpgi, line 746: `cudaq::run` will never actually return, in the accepted semantic sense of a return value, any values.

schweitzpgi (changes requested):
Let's do a walkthrough on this next week. I think you blew through a few "street closed" barriers here.

sacpis (commented):
Reviewing still in progress...

include/cudaq/Optimizer/CodeGen/QIRFunctionNames.h
atgeller: Sounds like there is a need to gate the use of these functions behind a flag based on what QIR profile we're targetting. For full QIR (e.g., the simulators), this should be ok, right? I will split this part out into a separate incremental PR so we can discuss this in more detail.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: Yeah specialize is not the best name. Basically, the issue I was encountering is that `packArgs` is setting up the arguments for synthesis, which expects them in one format, whereas `.argsCreator` expects them in a different format. One option that may simplify all of this is to build the message buffer somewhere in the Python runtime directly, and then use hybridLaunchKernel when applicable. The current approach always uses streamlinedLaunchKernel and figures it out further down the stack, which I think adds to the complexity.

atgeller:
This will need to be rebased on top of #4303 once that is merged. That add support for dynamic sized vector result logging for the simulator.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Removed.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: I think I've changed this properly to avoid copying unless we're overwriting the args. I'm actually not sure it would be a problem to overwrite the args in place.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: `convertResult` is also called in `readRunResults`.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: Updated to use parametric template

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: I changed the name to `forSynthesis` to make it a little clearer. When `forSynthesis` is false we pack the args for the argsCreator function, and when it's true we pack them in the form expected by ArgumentSynthesis.

atgeller:
Will need to be rebased on top of #4332 to fix current failures.

python/tests/kernel/test_reuse_compiler.py
```diff
@@ -159,3 +158,62 @@ def test_reuse_of_builder():
...
+    @cudaq.kernel
+    def flip() -> bool:
+        q = cudaq.qubit()
+        x(q)
+        return mz(q)
+
+    with cudaq.cudaq_runtime.reuse_compiler_artifacts():
+        result = flip()
+        assert result == True
+        result = flip()  # cached kernel
```
lmondada, line 192: Am I correct in thinking that this would fail if the second call was a completely different kernel? Now that practically none of the tests in this file throw errors anymore, it might be a good test case to add (if only as documentation so people reading this in the future can see when `reuse_compiler_artifacts` doesn't work -- right now to me it looks like everything just works here 😅 ).

runtime/common/CompiledModule.cpp
```diff
@@ -64,21 +65,31 @@ cudaq::KernelThunkResultType
 cudaq::CompiledModule::execute(const std::vector<void *> &rawArgs) const {
```
lmondada, line 61: This method is quickly getting nasty :D don't worry about it though, I'll work on this in one of my upcoming PRs.

runtime/common/ExecutionContext.cpp
```diff
@@ -27,56 +27,39 @@ thread_local bool reuseArtifact = false;
 
 class SavedCompilerArtifact {
 public:
-  void saveArtifact(const std::string &kernelName,
-                    const std::vector<void *> &args, const JitEngine &engine,
-                    std::function<void *()> argsCreatorThunk) {
+  void saveArtifact(const std::string &kernelName, const JitEngine engine) {
```
lmondada, line 30: Nice simplification! Awesome!!

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -41,17 +41,16 @@ using namespace mlir;
...
+                             bool enablePythonCodegenDump = false,
+                             bool isEntryPoint = true,
+                             bool isFullySpecialized = true) {
   PassManager pm(module.getContext());
-  ArgumentConverter argCon(name, module);
-  if (varArgIndices.empty())
-    argCon.gen(name, module, rawArgs);
-  else
-    argCon.gen(rawArgs, varArgIndices);
+  cudaq_internal::compiler::ArgumentConverter argCon(name, module);
```
lmondada: You don't need to namespace `ArgumentConverter` here, but I guess this just comes down to taste.

runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -335,31 +316,38 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
+    FunctionType fromFuncTy = funcOp.getFunctionType();
+    // Specialization for direct calls will take care of partial specialization
+    // separately
+    bool isLocalSimulator =
+        !(cudaq::is_remote_platform() || cudaq::is_emulated_platform());
+
+    std::vector<void *> closureArgs;
+
+    // Special handling in case the arguments were already synthesized
+    // TODO: should ensure args have no uses if this is the case?
```
lmondada: Did you intend to leave these two TODOs in? I'm not sure what they mean, they sound a bit like mid-implementation thoughts. Up to you if you think you can leave this as-is.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -956,8 +1015,9 @@ static MlirModule synthesizeKernel(nanobind::object kernel,
   SmallVector<StringRef> substRefs{substs.begin(), substs.end()};
 
   PassManager pm(context);
-  pm.addPass(cudaq::opt::createArgumentSynthesisPass(
-      kernelRefs, substRefs, /*changeSemantics=*/false));
+  pm.addPass(
```
lmondada: nit: not sure why the formatting changed here, given that I can't see any changes. You can probably get this back to two lines by manually removing a newline.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -434,23 +471,28 @@ void cudaq::packArgs(
         })
         .Case([&](cc::StructType ty) {
           auto mod = kernelFuncOp->getParentOfType<mlir::ModuleOp>();
-          auto [size, offsets] = getTargetLayout(mod, ty);
+          cc::StructType layoutTy =
```
lmondada: nit again on formatting: my subjective opinion is that clangfmt is doing a terrible job here at formatting this. Any chance this can not be slouched over 5 lines like that 😆 ?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -178,6 +178,7 @@ nanobind::args cudaq::simplifiedValidateInputArguments(nanobind::args &args) {
   return processed;
 }
 
+template <bool forSynthesis>
```
lmondada: Nicely done on coming up with this parametrization, it's cool :D

I still find the name `forSynthesis` pretty unclear. Suggestion: introduce an enum backed by a bool:
```c++
enum class SynthesisStyle : bool {
    Style1  = false,
    Style2 = true
};
```
That would be much more readable imo.

You could also use `std::conditional_t` to map your template bool to either `char` or `bool`, that way you could get rid of 90% of your conditionals. But not sure the result would be less verbose tbh.

Just ideas, do as you like :)

lmondada (approved):
Super nice! Those simplifications on `SavedCompilerArtifact` are 💯

python/tests/kernel/test_reuse_compiler.py
atgeller: Ah, yeah that'd probably be a good test case to add. Lines 50-60ish already test for redefinition/shadowing, but not for outright calling a different kernel.

runtime/common/CompiledModule.cpp
atgeller: Yeah, I wonder if there's a nice way to overload it or something.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Oops, that may have been overeager autocorrect. If it's not needed I would rather not because it's less readable.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -206,13 +207,17 @@ void cudaq::handleStructMemberVariable(void *data, std::size_t offset,
           for (std::size_t i = 0; auto v : asList)
             (*values)[i++] = nanobind::cast<T>(v);
 
-          std::memcpy(((char *)data) + offset, values, 16);
+          std::size_t copySize = forSynthesis ? 16UL : sizeof(std::vector<T>);
```
schweitzpgi: nit: what is the 16? Is that `sizeof({void*, i64})`?

runtime/common/CompiledModule.h
```diff
@@ -106,12 +106,18 @@ class CompiledModule {
     JitEngine engine;
     void (*entryPoint)() = nullptr;
     int64_t (*argsCreator)(const void *, void **) = nullptr;
+    /// Offset (in bytes) of the result field within the argsCreator-packed
+    /// buffer. Only valid when argsCreator is non-null and the kernel has a
+    /// result. Use resultInfo.bufferSize to know how many bytes to copy.
+    int64_t (*returnOffset)() = nullptr;
```
schweitzpgi: nit: should use std::int64_t to foil "clever" hackers with `-D` options. :)

schweitzpgi (approved):
Looks great. Thanks, Adam.

runtime/cudaq/platform/default/python/QPU.cpp
atgeller: Ah, that TODO is indeed rather unhelpful. This is related to a broader issue that if a kernel is explicitly argument synthesized by cudaq.synthesize, the type is unchanged, but we shouldn't use the argsCreator. This case is pretty hard to detect because the type is unchanged. The only distinction is that there are no arguments passed, which should otherwise create an error upstream, but it still feels a bit fragile. This TODO is basically saying "if we're claiming the args were synthesized, let's make sure they really were". We may also just want to introduce a `synthesized` attribute for kernels.

atgeller: I'll just remove the todos for now

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: Now it's only over 4 lines! :)

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: I went with an enum and mapping function to char/bool, which (in some places) is much less verbose. I called the enum `PackingStyle` and the styles are `argsCreator` and `synthesis`. I'm still not entirely happy with the naming. Maybe it should be `CallingConvention` and `messageBuffer`/`argsCreator` and `synthesis`?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: Yup, updated to sizeof({void *, i64}) and added a comment.

lmondada:
Thanks for reworking the templating. It looks very clean now, great!

I'm taking the freedom to merge this in right away, as I am working on more PRs that will conflict. As far as I can tell work here is complete, but if something is missing feel free to open a new PR and ping me, I'll review it.

PR #4266, opened 2026-04-06
[quake] Add `quake.measurements_size` op

khalatepradnya:
- New op to return the size of a measurements collection, analogous to `quake.veq_size`


atgeller (approved):
LGTM

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
```diff
@@ -1178,6 +1178,31 @@ class GetVeqSizeOpRewrite : public OpConversionPattern<quake::VeqSizeOp> {
...
+  matchAndRewrite(quake::MeasurementsSizeOp msize, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto loc = msize->getLoc();
+    auto parentModule = msize->getParentOfType<ModuleOp>();
+    auto context = parentModule->getContext();
+    auto qFunctionName = cudaq::opt::QIRArrayGetSize;
+
+    auto symbolRef = cudaq::opt::factory::createLLVMFunctionSymbol(
+        qFunctionName, rewriter.getI64Type(),
+        {cudaq::opt::getArrayType(context)}, parentModule);
```
schweitzpgi, line 1196: We need to fix this pattern so we don't have multiple threads try to update a ModuleOp at the same time.

lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -100,6 +100,28 @@ struct ForwardConstantVeqSizePattern
...
+struct ForwardConstantMeasurementsSizePattern
+    : public OpRewritePattern<quake::MeasurementsSizeOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(quake::MeasurementsSizeOp msSize,
+                                PatternRewriter &rewriter) const override {
+    auto msTy =
+        dyn_cast<quake::MeasurementsType>(msSize.getMeasurements().getType());
+    if (!msTy)
+      return failure();
```
schweitzpgi, line 115: This check is redundant and should be removed.

khalatepradnya: Will do in a follow-up.

lib/Optimizer/CodeGen/QuakeToLLVM.cpp
khalatepradnya: Is it acceptable to pre-declare in the pass setup?

khalatepradnya: See https://github.com/NVIDIA/cuda-quantum/pull/4275

PR #4267, opened 2026-04-06
[quake] Extend `quake.relax_size` to support collection of measurements

khalatepradnya:
* Allow `quake.relax_size` to be run on collections of measurements as well (along with the existing veq), 
```
quake.relax_size %ms : (!quake.measurements<4>) -> !quake.measurements<?>
```


lib/Optimizer/Dialect/Quake/QuakeOps.cpp
```diff
@@ -702,8 +702,21 @@ LogicalResult quake::MakeStruqOp::verify() {
...
+      return emitOpError("result veq type must not specify a size");
+    if (!isa<quake::VeqType>(inTy))
+      return emitOpError("input and result must both be veq types");
+  } else if (auto measTy = dyn_cast<quake::MeasurementsType>(resTy)) {
+    if (measTy.hasSpecifiedSize())
+      return emitOpError("result measurements type must not specify a size");
+    if (!isa<quake::MeasurementsType>(inTy))
+      return emitOpError("input and result must both be measurements types");
+  } else {
+    return emitOpError("result must be veq or measurements type");
```
schweitzpgi: Not needed. The constraints in the .td file already guarantee the argument and result types. We don't need to check them again.

test/Transforms/invalid.qke
```diff
@@ -226,3 +226,20 @@ func.func private @wonk(!quake.veq<18>, i32) -> f64
...
+  return
+}
+
+// -----
+
+func.func @relax_size_type_mismatch(%q : !quake.veq<4>) {
+  // expected-error@+1 {{input and result must both be measurements types}}
+  %bad = quake.relax_size %q : (!quake.veq<4>) -> !quake.measurements<?>
+  return
+}
```
schweitzpgi, line 245: Can you make sure the other 2 diagnostics are emitted here as well? The fifth should not appear since it is covered by the autogenerated code before the custom verifier is even executed.

schweitzpgi (approved):
LGTM. Let's beef up the diagnostic tests so we don't get silent regressions. :)

lib/Optimizer/Dialect/Quake/QuakeOps.cpp
khalatepradnya: Removed.

test/Transforms/invalid.qke
khalatepradnya: Sure, updated.

PR #4275, opened 2026-04-07
Follow-up to PR#4266 - Addressing review comments

khalatepradnya:
* Follow-up to PR #4266 
* Remove redundant check
* Fix the pattern(s) by pre-declaring functions so we don't have multiple threads try to update a ModuleOp at the same time


lib/Optimizer/CodeGen/ConvertToQIR.cpp
```diff
@@ -183,6 +184,14 @@ class ConvertToQIR : public cudaq::opt::impl::ConvertToQIRBase<ConvertToQIR> {
     target.addLegalOp<ModuleOp>();
 
     auto op = getOperation();
+
+    // Pre-declare QIR runtime functions that patterns such as,
+    // `GetVeqSizeOpRewrite` and `GetMeasurementsSizeOpRewrite` reference, so
+    // that they don't need to update the `ModuleOp`.
+    cudaq::opt::factory::createLLVMFunctionSymbol(
+        cudaq::opt::QIRArrayGetSize, IntegerType::get(context, 64),
+        {cudaq::opt::getArrayType(context)}, op);
```
schweitzpgi, line 193: We should use the intrinsics loading machinery for this if we can. It will properly deal with redundant declarations, etc.

khalatepradnya: Yes, I tried using `IRBuilder::loadIntrinsic` here but it crashes with `Loading a dialect (cc) while in a multi-threaded execution context`.  Adding `cc` and `func` to `dependentDialects` fixes that, but then `applyFullConversion` hits "redefinition of symbol". I think because `createLLVMFunctionSymbol` misses the `func.func` from `loadIntrinsic` and creates a duplicate.

A potential fix would be to widen `createLLVMFunctionSymbol` such that it also finds `func::FuncOp` declarations, but that change is pervasive, and I wasn't sure whether it is right approach. 
Any pointers?

khalatepradnya:
Closing as this is no longer relevant.

PR #4276, opened 2026-04-08
Replace `remote-mqpu` target with direct MPI support for sub-communicators

### LINKED ISSUE #3405: Follow up from #3306 - remove env vars
#3306 introduces two environment variables for testing purposes, `CUDAQ_PHASE_FOLDING` and `CUDAQ_BYPASS_PHASE_FOLDING_MINS`. Preferably, these should be part of some sort of runtime configuration instead.

1tnguyen:
# Background

The `remote-mqpu`, with a remote REST client-server architecture, was designed to support 
1. multiple MPI simulator backends, such as `nvidia-mgpu`, `tensornet`, via process separation (each backend has its own hosting server app); and 
2. cloud deployment (NVQC). 

Since (2) has been removed, the remote REST client-server architecture is no longer applicable, e.g., performance overhead, portability (HPC clusters may not allow http communication between nodes), and maintenance (runtime and compiler support).

Hence, we provide a direct MPI-based solution for (1) instead.

# New APIs

The main API addition is `set_communicator` API (under `cudaq::mpi` namespace for both C++ and Python) to set the backend MPI communicator. This facilitates resource allocation/binding for the backend, i.e., the backend only uses/sees MPI ranks within that sub-communicator. 

The sub-communicator can be constructed with MPI-native tools, such as the MPI library in C++ or `mpi4py` in Python.
Alternatively, we also provide `split_communicator` API (under `cudaq::mpi` namespace) to support that.

# Changes

- Remove code supporting the `remote-mqpu` target, including dependencies, tests, validation scripts, docs, etc.

- Add examples, docs, validation for the replacement.

- Also, unifying the `GPUEmulatedQPU` with the `DefaultQPU`: use platform's `beginExecution` to set the device id rather than duplicating the whole QPU implementation.


Renaud-K (approved):
This looks good to me. A lot of simplifications. Thank you!

mitchdz (approved):
Very nice. Love all the simplification. Thanks!

schweitzpgi (approved):
Rubberstamping

sacpis (approved):

PR #4284, opened 2026-04-09
Release the GIL during CUDA-Q compilation to enable Python signal handling (Ctrl+C) 

taalexander:
This branch is based on #4241 and should only be merged after it is merged on upstream/main and rebased.

When Python calls into CUDA-Q for kernel compilation, the GIL is held for the entire duration. This means that `Ctrl+C` does nothing. Python signal handlers (including KeyboardInterrupt) can only run when the GIL is held by a Python thread. Since compilation holds the GIL in C++ the whole time, pressing Ctrl+C is silently queued and never delivered until compilation finishes. For large circuits or complex pass pipelines, this can mean minutes of uninterruptible execution. Another side effect is that there is no python thread concurrency. Other python threads (progress bars, async I/O, timeouts) are blocked from running during compilation.                                                                                 
  
This PR updates the execution to release the GIL at Python-to-C++ entry points and check for pending Python signals between MLIR passes through instrumentation of the pass pipelines. 

The interruption granularity is at a per-pass level, so a single long-running pass will still block until it completes (which can be an big issue with SABRE right now.
Ultimately I had to work around this in benchmarking by forking for my use-case). The reason being that MLIR/LLVM is built with `-fno-exceptions`, so signals cannot safely unwind through pass execution. Instead, pending signals are detected between passes via PyErr_CheckSignals and converted to an MLIR error diagnostic that stops the pipeline through normal control flow.


lmondada (approved):
Thanks, awesome! What do you reckon are the chances that we are introducing a deadlock somewhere? I'm thinking some part of our C++ bindings that is assumed to always be able to get the GIL but now suddenly can't have it.

taalexander:
> Thanks, awesome! What do you reckon are the chances that we are introducing a deadlock somewhere? I'm thinking some part of our C++ bindings that is assumed to always be able to get the GIL but now suddenly can't have it.

I haven't found anything yet. The only place I ran into an issue was with CPP calling back out to Python and not having the GIL in the noise model. That required a little bit of GIL handling. My guess is this might become a bigger issue if we start writing pass in Python more in the future?

taalexander:
> /ok to test b066e3f

Issue #4289, opened 2026-04-09
SABRE qubit-mapping pass: topology-aware initial placement to reduce SWAPs on irregular devices


taalexander:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

 The SABRE qubit-mapping pass (`Mapping.cpp`) can only route through physical qubits that were        
  assigned during initial placement. On a device with more physical qubits than the circuit requires,
  the router cannot use the extra qubits for SWAP routing, even when doing so would produce shorter    
  paths. This results in significantly worse routing quality compared to Qiskit and TKET on the same
  topologies.

### Steps to reproduce the bug

```python
  import cudaq    

  cudaq.set_target('circuit-opt-bench-routed', device='path(5)')

  kernel = cudaq.make_kernel()
  q = kernel.qalloc(3)
  kernel.cx(q[0], q[2])  # non-adjacent on path(5), but adjacent on path(3)
                                                                                                       
  resources = cudaq.estimate_resources(kernel)
  print(resources)                                                                                     
  # Observe: routing result is identical whether device is path(3) or path(5)
  # Expected: path(5) should allow routing q0->q1->q2 using intermediate qubit
```

### Expected behavior

  The router should be able to insert SWAPs that move circuit qubits through any physical qubit on the 
  device, not just the initially-placed subset. Qiskit's `SabreSwap` does this: it creates a full
  physical-to-virtual mapping for all device qubits (unmapped qubits get sentinel values) and can route
   through them freely.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: main
- **Python version**: *
- **C++ compiler**: *
- **Operating system**: * 

### Suggestions

  1. Allocate wires for all physical qubits in the device, not just initially-placed ones. Unmapped    
  physical qubits get ancilla wires with no corresponding virtual qubit.
[...]

PR #4292, opened 2026-04-10
Add BFS initial placement to SABRE qubit-mapping pass

taalexander:
*Note*: This is an experiment of having an agent prepare a fix for #4289 with no intervention (ie. we passed it the issue and then hit enter for all stages of speckit) following the spec-kit workflow I have been using. This was done as a part of a live tutorial. We agreed to use the output of this PR as an oppurtunity to review the PR comparing human led review with review agents.
                                                         
  ### Summary                                                                                     

❯ - Add topology-aware BFS placement as the default initial placement strategy for the SABRE    
  qubit-mapping pass, replacing identity placement                                                
  - Add a `placement` pass option (`identity`|`bfs`, default `bfs`) to allow users to select the  
  strategy                                                                                        
  - Add tests for BFS placement on star, path, and grid topologies, plus tests verifying auxiliary
   qubit routing                                                                                  
                                                                                                  
  ### Context                                                                                     
                                                                                                  
  Addresses #4289. Investigation found that the routing engine already supports full-device
  routing through unmapped (auxiliary) qubits. The root cause of poor routing quality on irregular
   topologies is identity placement (`v_i -> p_i`), which ignores device topology structure.      
                                       
  On heavy-hex and star topologies, identity placement puts circuit qubits on low-degree          
  peripheral nodes. BFS placement starts from the highest-degree node and assigns circuit qubits
  to the most centrally-connected positions, giving the SABRE router a better starting point.
                                                                                                  
  ### Changes                                            
                                                                                                  
  - `lib/Optimizer/Transforms/Mapping.cpp`: Added `bfsPlacement()` function (BFS from
  highest-degree node, O(V+E)), placement option wiring with validation for unknown values        
  - `include/cudaq/Optimizer/Transforms/Passes.td`: Added `placement` option to `MappingFunc` pass
  - `test/Transforms/mapping_bfs_placement.qke`: New test for BFS on star(5,0), path(5), grid(3,3)
  - `test/Transforms/mapping_aux_routing.qke`: New test confirming auxiliary qubit routing works
  (star, path, single-qubit edge case)                                                            
  - 4 existing FileCheck test files updated with `placement=identity` to preserve expected output
                                                                                                  
  ### Test plan                                                                    
                                                         
  - [x] All 8 mapping lit tests pass (6 existing + 2 new)
  - [x] `mapping_unitaries.qke` passes with BFS default on path(10) via CircuitCheck (functional
  correctness)                         
  - [x] `mapping_bfs_placement.qke` verifies BFS produces valid mapped output on star, path, and
  grid topologies                                                                                 
  - [ ] `mapping_aux_routing.qke` verifies auxiliary qubit allocation and N=M regression
  - [x] No regressions in existing tests (FileCheck tests pinned to identity placement)

### DIFF
--- a/lib/Optimizer/Transforms/Mapping.cpp
+++ b/lib/Optimizer/Transforms/Mapping.cpp
@@ -49,6 +49,53 @@ void identityPlacement(Placement &placement) {
     placement.map(Placement::VirtualQ(i), Placement::DeviceQ(i));
 }
 
+/// Place circuit qubits on the most centrally-connected device positions using
+/// BFS from the highest-degree node. Remaining device qubits receive auxiliary
+/// virtual qubits in BFS order.
+void bfsPlacement(Placement &placement, const Device &device) {
+  unsigned numDeviceQubits = device.getNumQubits();
+  if (numDeviceQubits == 0)
+    return;
+
+  // Find the highest-degree node (tie-break: lowest index).
+  unsigned bestNode = 0;
+  unsigned bestDegree = 0;
+  for (unsigned i = 0; i < numDeviceQubits; ++i) {
+    unsigned degree = device.getNeighbours(Device::Qubit(i)).size();
+    if (degree > bestDegree) {
+      bestDegree = degree;
+      bestNode = i;
+    }
+  }
+
+  // BFS from bestNode to produce a device-qubit ordering.
+  SmallVector<unsigned> bfsOrder;
+  bfsOrder.reserve(numDeviceQubits);
+  SmallVector<bool> visited(numDeviceQubits, false);
+  SmallVector<unsigned> queue;
+  queue.push_back(bestNode);
+  visited[bestNode] = true;
+  while (!queue.empty()) {
+    SmallVector<unsigned> nextQueue;
+    for (unsigned node : queue) {
+      bfsOrder.push_back(node);
+      for (auto neighbor : device.getNeighbours(Device::Qubit(node))) {
+        if (!visited[neighbor.index]) {
+          visited[neighbor.index] = true;
+          nextQueue.push_back(neighbor.index);
+        }
+      }
+    }
+    queue = std::move(nextQueue);
+  }
+
+  // Assign virtual qubits to physical qubits in BFS order.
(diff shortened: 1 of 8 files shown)


lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -49,6 +49,53 @@ void identityPlacement(Placement &placement) {
...
+  // BFS from bestNode to produce a device-qubit ordering.
+  SmallVector<unsigned> bfsOrder;
+  bfsOrder.reserve(numDeviceQubits);
+  SmallVector<bool> visited(numDeviceQubits, false);
+  SmallVector<unsigned> queue;
+  queue.push_back(bestNode);
+  visited[bestNode] = true;
+  while (!queue.empty()) {
+    SmallVector<unsigned> nextQueue;
+    for (unsigned node : queue) {
```
atgeller, line 80: This BFS implementation looks a little wonky to me. I would expect it to go one node at a time on the queue, adding neighbors. That way we don't need a separate queue to avoid writes to the first queue through loop iterations. Maybe it's more optimal, but it looks a little funny and overly complex to me.

test/Transforms/mapping_bfs_placement.qke
```diff
@@ -0,0 +1,51 @@
...
+  %3:2 = quake.x [%0] %1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
+  %4:2 = quake.x [%3#0] %2 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
+  %5:2 = quake.x [%4#1] %3#1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
+  quake.return_wire %4#0 : !quake.wire
+  quake.return_wire %5#0 : !quake.wire
+  quake.return_wire %5#1 : !quake.wire
+  return
+}
+
+// Verify star(5,0) BFS produces valid mapped output.
```
atgeller, line 38: Doesn't look like we have any tests to make sure that the BFS strategy actually performs better than identity?

taalexander:
Closing as we want to open this problem to Unitary Hack challenge and this is only half a solution.

PR #4297, opened 2026-04-10
 [CMake] Add FindMinizip module

boschmitt:
PR chain:
- #4238 
- #4240
- \>\> #4297 \<\< (This)

Introduced a new CMake module to find and configure the Minizip library,
which is an addon for zlib.

### DIFF
--- a/cmake/modules/FindMinizip.cmake
+++ b/cmake/modules/FindMinizip.cmake
@@ -1,0 +1,63 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# Minizip is an addon library bundled with zlib (contrib/minizip) but not
+# installed by default.  This project's install_prerequisites.sh builds it.
+# The module tries pkg-config first, then falls back to manual search.
+
+set(_minizip_hints
+  "${ZLIB_ROOT}"
+  "$ENV{ZLIB_INSTALL_PREFIX}"
+  "${ZLIB_INCLUDE_DIR}/.."
+)
+
+find_package(PkgConfig QUIET)
+if(PkgConfig_FOUND)
+  find_path(_minizip_pc_dir NAMES minizip.pc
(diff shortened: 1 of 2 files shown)


cmake/modules/FindMinizip.cmake
```diff
@@ -0,0 +1,63 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
```
sacpis: nit: Should the year be just 2026?

runtime/cudaq/platform/default/rest_server/CMakeLists.txt
```diff
@@ -6,61 +6,11 @@
...
-    ${ZLIB_ROOT}/lib/pkgconfig
-    $ENV{ZLIB_INSTALL_PREFIX}/lib/pkgconfig
-    ${ZLIB_INCLUDE_DIR}/../lib/pkgconfig
-)
-if (PkgConfig_FOUND AND MINIZIP_PKG_CONFIG_DIR) 
-  set(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${MINIZIP_PKG_CONFIG_DIR}")
-  pkg_check_modules(MINI_ZIP IMPORTED_TARGET minizip)
-  # Make sure that we link to minizip static library
-  if (MINI_ZIP_FOUND)
-    find_library(MINI_ZIP_LIB NAMES libminizip.a
```
sacpis, line 30: This used to search explicitly for `libminzip.a`. The new cmake uses `find_library` which will find whichever variant cmake finds first. Was the intent here to link `minzip` statically?

sacpis (approved):
Overall LGTM. Thanks @boschmitt. Just one question.

runtime/cudaq/platform/default/rest_server/CMakeLists.txt
boschmitt: This PR changes the behavior. There is not reason for us to enforce the use of a static library on all builds. This is related to #4301. When packaging for distribution, using static libraries simplify things. However, not all dev environments need a static version of `minizip` (or any of the other dependencies).

sacpis: Got it. Thank you @boschmitt.

PR #4301, opened 2026-04-10
[CMake] Split static linking into two options

boschmitt:
PR chain:
- #4238 
- #4240
- #4297
- \>\> #4301 \<\< (This)

Replace the single `CUDAQ_ENABLE_STATIC_LINKING` flag with two
orthogonal options:

- `CUDAQ_STATIC_CXX_RUNTIME`: controls -static-libstdc++/-static-libgcc
  for ABI isolation when distributing shared libraries.

- `CUDAQ_STATIC_DEPS`: controls whether external dependencies (zlib,
  OpenSSL, curl, BLAS) are found as static archives.

Remove unconditional defaults that forced static dependencies even for
development builds.

The old `CUDAQ_ENABLE_STATIC_LINKING` env variable is still supported
with a deprecation warning that maps it to both new flags.


runtime/cudaq/algorithms/optimizers/ensmallen/CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ set(LIBRARY_NAME cudaq-ensmallen)
 
 set (CUDAQ_TPLS_DIR ${CMAKE_SOURCE_DIR}/tpls)
 
-set(BLA_STATIC ON)
+if(DEFINED ENV{BLAS_INSTALL_PREFIX})
+  find_library(BLAS_LIBRARIES blas
+    HINTS "$ENV{BLAS_INSTALL_PREFIX}" "$ENV{BLAS_INSTALL_PREFIX}/lib"
+    REQUIRED)
+endif()
```
sacpis, line 17: Don't think find_library honors `BLA_STATIC`. I think if the prefix directory contains both `libblas.a` and `libblas.so`, this will prefer `libblas.so` on Linux. The [removed code](https://github.com/NVIDIA/cuda-quantum/pull/4301/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL211) pinned `BLAS_LIBRARIES` to `libblas.a`.

CMakeLists.txt
```diff
@@ -172,6 +172,21 @@ option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Fron
 option(CUDAQ_ENABLE_CC "Enable CUDA-Q code coverage generation." OFF)
 option(CUDAQ_REQUIRE_OPENMP "Fail the build if OpenMP is not found." OFF)
 option(CUDAQ_ENABLE_SANITIZERS "Enable Address Sanitizer (ASan) and Undefined Behavior Sanitizer (UBSan)." OFF)
+# Backward compatibility: map the old CUDAQ_ENABLE_STATIC_LINKING env variable
+# to the two new flags. Setting cache variables here means the option() calls
+# below will respect them (option() does not overwrite existing cache entries).
+if(DEFINED ENV{CUDAQ_ENABLE_STATIC_LINKING} AND "$ENV{CUDAQ_ENABLE_STATIC_LINKING}")
```
sacpis: Are we missing to check the cache env here? Seems like `DEFINED ENV{CUDAQ_ENABLE_STATIC_LINKING}` only checks the process env. It does not look into cmake cache. So `-DCUDAQ_ENABLE_STATIC_LINKING=ON` no longer triggers the shim. So both new flags stays off and no deprecation warning is fired.

Should we add a check like this?

```
if ((DEFINED ENV{CUDAQ_ENABLE_STATIC_LINKING} AND "$ENV{CUDAQ_ENABLE_STATIC_LINKING}") OR (DEFINED CUDAQ_ENABLE_STATIC_LINKING AND CUDAQ_ENABLE_STATIC_LINKING))
```

sacpis (approved):
Overall LGTM. Thanks @boschmitt. Left 2 comments.

runtime/cudaq/algorithms/optimizers/ensmallen/CMakeLists.txt
boschmitt: You are correct that `find_library` doesn't honor `BLA_STATIC`, but that is not how this variable works. This variable is used by CMake's `FindBLAS` module, which doesn't use `find_library` for its primary search.

Note: I actually tried something like you suggestion. However, it is a global sledgehammer that breaks unrelated libraries. Setting it will make every subsequent `find_library` and `find_package` across the entire project only find static archives. This would break finding CUDA libraries that are only available as shared libraries, e.g., `libcudart`.

CMakeLists.txt
boschmitt: Indeed. I will add an extra check. (I'm really not sure if anyone uses this setting, AFAICT it is not used anywhere in our build scripts.)

runtime/cudaq/algorithms/optimizers/ensmallen/CMakeLists.txt
sacpis: I think there is some confusion here. The fix I am proposing doesn't touch `CMAKE_FIND_LIBRARY_SUFFIXES` at all. It just uses find_library's NAME arg, which is local to that single call.

I am proposing something like below to kind of match the old comparison.

```
if(DEFINED ENV{BLAS_INSTALL_PREFIX})
    if(CUDAQ_STATIC_DEPS)
        find_library(BLAS_LIBRARIES NAMES libblas.a
          HINTS "$ENV{BLAS_INSTALL_PREFIX}" "$ENV{BLAS_INSTALL_PREFIX}/lib"
        REQUIRED)
    else()
        find_library(BLAS_LIBRARIES NAMES blas
          HINTS "$ENV{BLAS_INSTALL_PREFIX}" "$ENV{BLAS_INSTALL_PREFIX}/lib"
        REQUIRED)
    endif()
endif()
find_package(BLAS REQUIRED)
```

On macOS, I think the whole block is gated by `DEFINED ENV{BLAS_INSTALL_PREFIX` . I think macOS uses the native BLAS.

sacpis: Removed the suggested code in my first comment. That should not have been there.

PR #4303, opened 2026-04-10
Add support for dynamic sized vector results in run for full QIR

atgeller:


targettests/execution/cudaq_run_dynamic_result.cpp
```diff
@@ -0,0 +1,26 @@
...
+// A kernel with a dynamic result size
+__qpu__ std::vector<bool> branch_vec_test(bool flip) {
+  cudaq::qubit ctrl;
+  if (flip)
+    x(ctrl);
+  bool b = mz(ctrl);
+  int sz = b ? 2 : 4;
+  cudaq::qvector data(sz);
+  return mz(data);
+}
```
khalatepradnya, line 53: Can you add a simple test with vector of `int` and `float` as well?

targettests/execution/cudaq_run_dynamic_result.cpp
```diff
@@ -0,0 +1,26 @@
+// RUN: nvq++ %s -o %t && %t | FileCheck %s
```
khalatepradnya, line 9: nit: Missing copyright banner - CI should flag this...

runtime/nvqir/NVQIR.cpp
```diff
@@ -485,6 +485,67 @@ void __quantum__rt__array_record_output(std::uint64_t len, const char *label) {
...
+    case 2:
+      val = static_cast<int64_t>(*reinterpret_cast<int16_t *>(data + i * 2));
+      break;
+    case 4:
+      val = static_cast<int64_t>(*reinterpret_cast<int32_t *>(data + i * 4));
+      break;
+    case 8:
+      val = *reinterpret_cast<int64_t *>(data + i * 8);
+      break;
+    }
```
khalatepradnya, line 525: May be add a `default` case, it should throw for unsupported types?

runtime/nvqir/NVQIR.cpp
```diff
@@ -485,6 +485,67 @@ void __quantum__rt__array_record_output(std::uint64_t len, const char *label) {
...
+  for (int64_t i = 0; i < count; ++i) {
+    std::string elemLabel = std::string("[") + std::to_string(i) + "]";
+    double val = 0.0;
+    switch (elemSizeBytes) {
+    case 4:
+      val = static_cast<double>(*reinterpret_cast<float *>(data + i * 4));
+      break;
+    case 8:
+      val = *reinterpret_cast<double *>(data + i * 8);
+      break;
```
khalatepradnya, line 545: Same here.

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -156,9 +156,46 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
...
+          if (!allowDynamic)
+            return;
+          auto eleTy = vecTy.getElementType();
+          auto i8PtrTy = cudaq::cc::PointerType::get(rewriter.getI8Type());
+          Value size = rewriter.create<cudaq::cc::StdvecSizeOp>(
+              loc, rewriter.getI64Type(), val);
+          Value rawData =
+              rewriter.create<cudaq::cc::StdvecDataOp>(loc, i8PtrTy, val);
+          if (auto intTy = dyn_cast<IntegerType>(eleTy)) {
+            if (intTy.getWidth() == 1) {
```
schweitzpgi: Another approach is an identity test.
```c++
if (eleTy == rewriter.getI1Type()) ...
```

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -156,9 +156,46 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
...
+              rewriter.create<func::CallOp>(loc, TypeRange{},
+                                            cudaq::opt::QIRBoolSpanRecordOutput,
+                                            ArrayRef<Value>{rawData, size});
```
schweitzpgi, line 176: On the device, a bool span is exactly the same representation as a byte span. I suspect you don't need to special case this other than the QIR logging tag string values?

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
```diff
@@ -156,9 +156,46 @@ class ReturnRewrite : public OpRewritePattern<cudaq::cc::LogOutputOp> {
...
+            auto floatTy = cast<FloatType>(eleTy);
+            std::int32_t byteSize = floatTy.getWidth() / 8;
+            Value elemSize =
+                rewriter.create<arith::ConstantIntOp>(loc, byteSize, 32);
+            rewriter.create<func::CallOp>(
+                loc, TypeRange{}, cudaq::opt::QIRFloatSpanRecordOutput,
+                ArrayRef<Value>{rawData, size, elemSize});
+          } else {
+            // Unsupported element type — trap.
+            Value one = rewriter.create<arith::ConstantIntOp>(loc, 1, 64);
```
schweitzpgi, line 198: Can we add an LLVM_DEBUG() here? It's possible to have lists of structs and lists of lists of ...

schweitzpgi (approved):
LGTM. A couple of nits.

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
atgeller: I combined the cases and just branch on the logging tag value instead.

targettests/execution/cudaq_run_dynamic_result.cpp
atgeller: I added simple tests, but I need to double check they exercise this case.

atgeller: Yup, they do. For the emulated hardware test case, the QIR verification error is caused by an alloca because the vector has a dynamic size, so that's the same for both cases.

khalatepradnya (approved):
Thank you, @atgeller !

lib/Optimizer/CodeGen/ReturnToOutputLog.cpp
atgeller: Ah, no, that fails because they have different signatures (QIRIntSpanRecordOutput needs to know the element size which adds an extra instruction that's unnecessary for QIRBoolSpanRecordOutput).

PR #4305, opened 2026-04-11
Fix macOS xdist crash in GHZ noise validation test

KevinSailema:
Fixes #4281.

- This PR addresses an intermittent macOS CI crash where an xdist worker dies while running test_simple_run_ghz_with_noise.
- The change is intentionally minimal and limited to test/validation flow, without touching product runtime code.

Root cause addressed
- The failing path uses global target state in a noisy kernel test, and this can be fragile under parallel test execution in macOS CI.
- The failure mode is a worker crash, not a deterministic assertion failure.

What changed
- Hardened target lifecycle in test_run_kernel.py:
- test_simple_run_ghz_with_noise now guarantees target reset via try/finally.
- Results are materialized before resetting target to avoid teardown/lifecycle races.
- Updated macOS validation strategy in validate_pycudaq.sh:
- Keep core tests parallel with xdist.
- Exclude test_simple_run_ghz_with_noise from the parallel core batch on macOS.
- Run that single test serially immediately after core tests.

Why this approach
- Keeps maximum parallel coverage for the rest of the suite.
- Isolates only the unstable test in macOS CI.
- Minimizes blast radius and avoids broad skips or module-wide serialization.

Validation
- Target test passed in serial and xdist locally.
- Stress runs of the target test in xdist were stable across repeated iterations.
- Adjacent GHZ tests remained green in serial and xdist after the change.
- Script syntax validated and change scope confirmed to the two intended files only.

Impact
- Reduces flaky macOS worker crashes for the validation pipeline.
- Preserves test coverage while improving CI reliability.

### DIFF
--- a/scripts/validate_pycudaq.sh
+++ b/scripts/validate_pycudaq.sh
@@ -333,15 +333,47 @@ fi
 
 # Run core tests
 echo "Running core tests."
-python3 -m pytest -v -n auto "$root_folder/tests" \
-    --ignore "$root_folder/tests/backends" \
-    --ignore "$root_folder/tests/dynamics/integrators" \
-    --ignore "$root_folder/tests/parallel" \
-    --ignore "$root_folder/tests/domains"
+core_test_args=(
+    -v
+    -n
+    auto
+    "$root_folder/tests"
+    --ignore
+    "$root_folder/tests/backends"
+    --ignore
+    "$root_folder/tests/dynamics/integrators"
+    --ignore
+    "$root_folder/tests/parallel"
+    --ignore
+    "$root_folder/tests/domains"
+)
+
+macos_serial_core_tests=()
+if $is_macos; then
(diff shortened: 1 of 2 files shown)


scripts/validate_pycudaq.sh
```diff
@@ -333,16 +333,48 @@ fi
...
+    "$root_folder/tests/dynamics/integrators"
+    --ignore
+    "$root_folder/tests/parallel"
+    --ignore
+    "$root_folder/tests/domains"
+)
+
+macos_serial_core_tests=()
+if $is_macos; then
+    ghz_noise_test="test_simple_run_ghz_with_noise"
```
mitchdz, line 353: I'm not a fan of hardcoding the test here.

I don't believe it always fails this test specifically - I started creating GH issues once I see them to try and paint a better picture. I think the failing test is random, and has something to do with the runner environment likely.

PR #4306, opened 2026-04-12
Fix/issue 4268 forward declared adjoint

### LINKED ISSUE #4268: `apply-op-specialization` crashes instead of erroring when adjointing a forward-declared kernel

**Reproduction**

prep.cpp:
```cpp
#include <cudaq.h>
```

main.cpp:
```cpp
#include <cudaq.h>
__qpu__ bool kernel() {
  cudaq::qubit q;
  cudaq::adjoint(prep, q);
  return mz(q);
}
int main() {}
```

```bash
nvq++ -c prep.cpp -o prep.o
nvq++ -c main.cpp   # crashes
```

**Actual:**
```
...
ApplyOpSpecialization.cpp:588
```

KevinSailema:
Fixes #4268

This PR fixes a crash in apply-op-specialization when adjoint is requested for a forward-declared kernel whose body is not available in the current translation unit.

Instead of crashing with an internal assertion, the compiler now emits a clear diagnostic and fails gracefully.

## Root Cause
During specialization, the pass attempted to generate an adjoint variant for declaration-only functions. Since no function body exists, the transformation pipeline eventually hit an invalid internal state and aborted with an assertion.

## What Changed
- Added a guard in apply-op-specialization before generating adjoint/adjoint-control variants.
- If the target function is declaration-only, emit a diagnostic:
  cannot create adjoint variant for forward-declared kernel ... (kernel body is unavailable in this translation unit)
- Added a regression test that reproduces this case and checks for the expected error.

## Validation
- Regression test passes: the expected diagnostic is emitted.
- Original 2-translation-unit repro now behaves correctly:
  - prep.cpp compiles
  - main.cpp reports the diagnostic
  - no internal assertion crash (isKnownSentinel) appears

## Impact
- No behavior change for valid kernels with available definitions.
- Unsupported forward-declared adjoint now fails predictably with a user-facing error instead of a compiler crash.


atgeller:
> /ok to test 7d0aca7

lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -403,6 +422,17 @@ class ApplySpecializationPass
       assert(func && "global must be a FuncOp");
       auto &variant = variantIter->second;
 
+      // We cannot synthesize an adjoint if only a declaration is available.
+      if ((variant.needsAdjointVariant || variant.needsAdjointControlVariant) &&
```
mitchdz, line 426: Can you also add a `variant.needsControlVariant` to the check?

A user who writes quake.apply [%ctl] @prep %q against a forward-declared @prep will hit the same class of crash this PR is fixing, just through a different code path.

e.g.
```
// ctl_decl.qke
module {
  func.func private @prep(!quake.ref)
  func.func @kernel() {
    %ctl = quake.alloca !quake.ref
    %q = quake.alloca !quake.ref
    quake.apply @prep [%ctl] %q : (!quake.ref, !quake.ref) -> ()
    return
  }
}
```
```
$ cudaq-opt --apply-op-specialization ctl_decl.qke
cudaq-opt: /opt/llvm/include/llvm/ADT/ilist_iterator.h:138: ...
Aborted (core dumped)
```

schweitzpgi:
Closing as a duplicate #4715

PR #4225, opened 2026-03-31
Reduce cudaq import time of quakeDialects by lazily loading the targets

taalexander:
This PR is based on #4224 and should only be merged after. It further reduces the loading of quake dialects by ensuring that we only load the shared libraries for those targets we use when we load them (including deferring the loading of the default target to only when/if it is needed).

It reduces the load time to basically only the cost of pybind bindings. Changes:

- Lazy-load simulator and platform .so files: scan filenames at init, dlopen on demand when set_target() or get_target() is first called     
- Pass explicit library paths to `getUniquePluginInstance` instead of searching all globally loaded symbols via dlopen(NULL)
- Pre-populate `runtimeTarget` on the platform before `setTargetBackend()` to skip redundant YAML file re-parsing                                
- Defer target initialization to first use, with graceful fallback to `qpp-cpu` if the `nvidia` simulator fails to load
- Add lazy init callbacks to NVQIR and quantum_platform so the simulator
   and platform are loaded on first access rather than at import time

  ### Import timing comparison (local machine)                                                                                                 
                                                                                                                                               
  | Metric | Baseline | #4224| #4225 | Reduction |
  |---|---|---|---|---|                                                                                                                        
  | `_quakeDialects` self | 822ms | 478ms | 147ms | **82%** |                                                                                  
  | Total cudaq cumulative | 1,569ms | 704ms | 331ms | **79%** |                                                                               
  | Modules loaded | 1,174 | 339 | 339 | **71%** |                                                                                             
                                                                                                                                               
  ### `_quakeDialects` constructor breakdown                                                                                                   
                                                                                                                                               
  | Phase | Baseline (ms) | #4225 (ms) |                                                   
  |---|---|---|                                             
  | nvidia-smi / GPU detection | 171 | 0 (deferred) |                                                                                          
  | dlopen 13 simulator `.so` files | 63 | 0 (on-demand) |                                                                                     
  | `resetTarget` (simulator init) | 241 | 0 (deferred) |                                                                                      
  | Target YAML parsing | 1.6 | 1.6 |                                                                                                          
  | Core lib dlopen | 0.05 | 0.05 |                                                                                                            
  | Filename scan | - | 0.1 |                                                                                                                  
  | pybind11 binding setup | ~145 | ~145 |                                                                                       
  | **Total** | **~476** | **~147** |                                                                                                            
                                                                                                                                               
[...]

PR #4307, opened 2026-04-12
Add back eager loading of `libcudaq-em-default` on import

taalexander:
Downstream libraries like `CUDA-QX` reference
symbols in `libcudaq-em-default.so` at dlopen time via inline functions in `execution_manager.h`. Since this library is neither a simulator nor a platform, the lazy loading from #4225 never loaded it, causing an undefined symbol error on import as noticed by @bmhowe23 in this comment https://github.com/NVIDIA/cuda-quantum/pull/4225#issuecomment-4230312258.

I tried to find a way to achieve this with pure Cmake but due to circular dependencies it was not easy so adding this back. Performance hit will be minimal for import time (most a few ms).


mitchdz (approved):
This looks good to me. Let's get it in to unblock cudaqx.

My question is are we missing any other dependencies like this that should always get loaded?

bmhowe23 (approved):

PR #4308, opened 2026-04-13
Fix zero() and identity() scalar fallback

huaweil-nv:
## Summary
- fix the empty-sequence check in `cudaq.operators.zero()` and `cudaq.operators.identity()`
- replace `hasattr(degrees, "len")` with `hasattr(degrees, "__len__")`
- add a regression test for no-argument calls

## Root cause
`zero()` and `identity()` used `hasattr(degrees, "len")` to detect an empty sequence.

That check is incorrect in Python: standard containers expose `__len__`, not `len`. Because of that, the empty-input branch never triggered, and no-argument calls fell through to the `MatrixOperatorTerm` path instead of returning the documented `ScalarOperator`.

## User-visible behavior
Before this change:
- `operators.zero()` returned `MatrixOperatorTerm`
- `operators.identity()` returned `MatrixOperatorTerm`

After this change:
- `operators.zero()` returns `ScalarOperator.const(0)`
- `operators.identity()` returns `ScalarOperator.const(1)`

### DIFF
--- a/python/cudaq/operators/custom/backwards_compatibility.py
+++ b/python/cudaq/operators/custom/backwards_compatibility.py
@@ -35,7 +35,7 @@ def const(constant_value: NumericType) -> ScalarOperator:
 def zero(
         degrees: Sequence[int] | int = []
 ) -> ScalarOperator | MatrixOperatorTerm:
-    if hasattr(degrees, "len") and len(degrees) == 0:
+    if hasattr(degrees, "__len__") and len(degrees) == 0:
         return ScalarOperator.const(0)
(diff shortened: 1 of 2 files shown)


1tnguyen:
> /ok to test [783ca51](https://github.com/NVIDIA/cuda-quantum/pull/4308/commits/783ca51dc300196ba1f6d6ec62bc9502690c60a7)

1tnguyen (approved):
LGTM 👍

sacpis (approved):

sacpis:
> /ok to test 7c47c32

PR #4313, opened 2026-04-13
[Testing] Use python to test phase folding

1tnguyen:
Currently, phase folding uses `remote-mqpu` for testing. One reason is that we need to compute the state overlap between the resulting states with and without the optimization. 

This PR refactors those tests to use Python instead:

- The pipeline (with and without phase folding) is embedded in the target config file. We add dummy targets for this.

- Python is perfect for this as it allows us to switch targets. Provided that the underlying simulator is not changed (qpp in this case), we can compute the overlap.

- Teach `random_gen.py` to generate python code rather than C++.

**Tested by**:

- Use some uncommitted test code to inject bugs into the phase folding pass implementation and see that some of the generated tests actually failed the overlap validation.


python/tests/mlir/phase_folding/phase-folding-bench-mins.yml
```diff
@@ -0,0 +1,16 @@
+# ============================================================================ #
+# Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                   #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+name: phase-folding-bench
```
atgeller: I think this should be phase-folding-bench-mins, hence the CI errors.

atgeller:
Thanks @1tnguyen! This looks like a nicer way of structuring these tests.

python/tests/mlir/phase_folding/phase-folding-bench-mins.yml
1tnguyen: Indeed. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4313/changes/797ffa1546a6664b90e744359badabbb662fc840

atgeller (approved):
LGTM. Thanks @1tnguyen!

PR #4314, opened 2026-04-13
Add apply-op-specialization and inlining to circuit-opt-bench pipeline

taalexander:
Without these passes, unitary-synthesis creates `quake.apply` ops that call KAK-decomposed helper functions, but the helpers are never inlined into the main kernel. symbol-dce then removes them as dead code, producing 0 2Q gates for any custom unitary operation which breaks resource estimation.


atgeller (approved):
LGTM

PR #4322, opened 2026-04-14
Rename CompiledKernel to CompiledModule

lmondada:
This is a first follow-up from yesterday's conversation. It renames `CompiledKernel` to `CompiledModule`. This is to clarify the exact scope of this type:

> A `CompiledModule` is the collection of artifacts produced by CUDAQ when compiling a Quake MLIR module.

This is slightly different from a "compiled kernel", as it may contain zero, one or multiple entrypoints, depending on what was in the module and how it is compiled.

I'm working separately on another PR that will make a clearer distinction between the CompiledModule and its artifacts (e.g. `JitArtifact` come with an entrypoint, whereas `MlirArtifact`s do not). It will also move the logic on how to "execute" `CompiledModule`s out of the type itself.


Renaud-K (approved):
Looks good. Thank you.

PR #4327, opened 2026-04-15
Fix CuPy state initialization for strided device arrays and adjacent dynamics issues

huaweil-nv:
## Summary

This PR fixes `cudaq.State.from_data(...)` for strided CuPy device arrays.

Before this change, CuPy arrays with explicit stride metadata could be misinterpreted during state construction. In particular, transposed views, Fortran-ordered arrays, and other non-contiguous CuPy arrays could be read as if they were flat contiguous buffers, which silently changed their logical layout.

This affected both state-vector-like inputs and 2D density matrix inputs, with the 2D case being especially dangerous because the matrix contents could be silently reordered.

## How to reproduce

A minimal reproducer is to pass a transposed or Fortran-ordered CuPy array into `cudaq.State.from_data(...)`.

Example:

```python
import cupy as cp
import cudaq
import numpy as np
from cudaq.operators import spin

cudaq.set_target("dynamics")

base = cp.array([[1.0 + 0.0j, 2.0 + 0.0j],
                 [3.0 + 0.0j, 4.0 + 0.0j]], dtype=cp.complex128)

rho = base.T  # or cp.asfortranarray(base)
state = cudaq.State.from_data(rho)

result = cudaq.evolve(
    0.0 * spin.z(0),
    {0: 2},
    Schedule([0.0], ["t"]),
    state,
    observables=[],
    collapse_operators=[],
    store_intermediate_results=cudaq.IntermediateResultSave.NONE,
)

print(np.array(result.final_state()))
print(cp.asnumpy(rho))
```
Before this change, these two values could differ for strided CuPy inputs even though they should represent the same logical matrix.

## Root Cause
The CuPy import path did not consistently preserve logical layout information.

CuPy arrays expose device memory through __cuda_array_interface__, including shape and stride metadata. However, the previous implementation could reduce CuPy inputs to a raw device pointer plus element count, which is only safe for contiguous layouts. For strided arrays such as a.T or cp.asfortranarray(a), this loses the logical indexing semantics and can cause the array to be interpreted using its underlying flat memory layout instead of its intended logical values.

## What this PR changes
* Read CuPy stride metadata from __cuda_array_interface__
* Preserve safe fast paths for contiguous inputs
* Canonicalize CuPy arrays when needed before constructing the state
* Add regression coverage for:
   - strided 1D CuPy views
   - C-order 2D CuPy arrays
   - Fortran-order 2D CuPy arrays
   - transposed 2D CuPy views

## Why this matters
Users expect cudaq.State.from_data(cupy_array) to preserve the logical values of the CuPy array, regardless of whether the array is contiguous, transposed, or stored with non-default strides. This PR fixes cases where that expectation was not met and prevents silent layout corruption for GPU-backed inputs.

------ update in 4/24
 ## Adjacent fixes discovered during review

  Verifying the fix under the post-nanobind merge surfaced two independent
  correctness gaps on the dynamics target. They are committed separately so
  each can be reviewed or reverted on its own.

  ### 1. Reject non-square 2D CuPy arrays at `from_data` time (commit `0f762a09`)

  A non-square 2D CuPy array on the dynamics target previously slipped through
  `createStateFromPyBuffer` and was flattened into a 1D buffer inside
  `TensorStateData`. The failure was deferred until `initialize_cudm()` raised
  a cryptic `Invalid hilbertSpaceDims for the state data` with no pointer to
  the real cause. Now mirrors the host 2D validation path and rejects at
  `from_data` time with `state.from_data 2D array (density matrix) input must
  be square matrix data.`.

  ### 2. Propagate `isDensityMatrix` in `createFromSizeAndPtr` (commit `21e4bdde`)

  PR #2853 ("Migrate Python dynamics solver implementation to pybind11",
  May 2025) trimmed the `CuDensityMatState` constructor's `isDm` parameter
  but did not add an equivalent assignment in `createFromSizeAndPtr`, leaving
  the locally computed `isDm` dead for ~11 months. As a result,
  `cudaq.State.from_data(cupy_2d)` on the dynamics target produced a state
  whose `getTensor().extents` returned `[N*N]` instead of `[N, N]` and whose
  `np.array(state).shape` was `(N*N,)` instead of `(N, N)`. The inconsistency
  was masked because `evolve()` re-infers the shape during `initialize_cudm()`,
  so no existing test inspected the state before `evolve`.

  Restored the flag after construction. `dimension` already stores the flat
  element count and `getTensor()`/`operator()` already `sqrt` it when
  `isDensityMatrix` is true, so no other field needs to change.


python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -15,16 +15,157 @@
...
+      /*dataPtr=*/reinterpret_cast<void *>(dataInfo[0].cast<int64_t>()),
+      /*readOnly=*/dataInfo[1].cast<bool>(),
+      /*typeStr=*/typeStr,
+      /*dataTypeSize=*/dataTypeSize,
+      /*formatDescriptor=*/formatDescriptor,
+      /*shape=*/std::move(shape),
+      /*strides=*/std::move(strides)};
+}
+
+bool isCContiguous(const std::vector<ssize_t> &shape,
```
sacpis, line 142: Should we `assert(shape.size() == strides.size())` or return early to make it safer?

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -15,16 +15,157 @@
...
+  bool needsCanon = (info.shape.size() == 1) ||
+                    (info.shape.size() == 2 && targetName == "dynamics");
+  return needsCanon && !isCContiguous(info.shape, info.strides, info.itemsize);
+}
+
+py::object canonicalizeCupyArrayToNumpy(py::handle cupyArray) {
+  auto cupy = py::module_::import("cupy");
+  auto contiguous = cupy.attr("ascontiguousarray")(
+      py::reinterpret_borrow<py::object>(cupyArray));
+  return cupy.attr("asnumpy")(contiguous);
```
sacpis: Can we just do `return cupy.attr("asnumpy")(py::reinterpret_borrow<py::object>(cupyArray));`?

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -15,16 +15,157 @@
...
+  shape.reserve(shapeTuple.size());
+  for (auto dim : shapeTuple)
+    shape.push_back(dim.cast<ssize_t>());
+
+  const std::string typeStr = cupyArrayInfo["typestr"].cast<std::string>();
+  auto [dataTypeSize, formatDescriptor] = getCupyComplexTypeInfo(typeStr);
+  auto strides = getCupyArrayStrides(cupyArrayInfo, shape, dataTypeSize);
+
+  return CuPyArrayMetadata{
+      /*dataPtr=*/reinterpret_cast<void *>(dataInfo[0].cast<int64_t>()),
```
sacpis: Should we do `reinterpret_cast<void *>(dataInfo[0].cast<uintptr_t>())`? As on some platforms, casting through `int64_t` is undefined behavior (intptr_t != int64_t)?

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -15,16 +15,157 @@
...
+    return py::buffer_info(dataPtr, dataTypeSize, /* itemsize */
+                           formatDescriptor,      /* format */
+                           static_cast<py::ssize_t>(shape.size()), /* ndim */
+                           shape,                                  /* shape */
+                           strides,                                /* strides */
+                           readOnly /* readonly */);
+  }
+
+  std::vector<std::size_t> getExtents() const {
+    return std::vector<std::size_t>(shape.begin(), shape.end());
```
sacpis: Can we please confirm that return `size_t` vector would not cause any issue, as shape is of `ssize_t` (signed size_t)?

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -15,16 +15,157 @@
...
+  for (int i = static_cast<int>(shape.size()) - 1; i >= 0; --i) {
+    if (shape[i] > 1 && strides[i] != expectedStride)
+      return false;
+    expectedStride *= shape[i];
+  }
+
+  return true;
+}
+
+bool shouldCanonicalizeCupyArray(const py::buffer_info &info,
```
sacpis, line 162: Do we need this intermediate `py::buffer_info` as CuPyArrayMetadata already holds shape, strides, and dataTypeSize? Can we directly pass `CuPyArrayMetadata` instead?

sacpis (approved):
Overall LGTM. Thanks @huaweil-nv. Left few comments.

python/runtime/cudaq/algorithms/py_state.cpp
huaweil-nv: Added a guard in `isCContiguous()` for inconsistent shape/stride metadata. In that case we now conservatively return `false` so the caller falls back to the safer canonicalization path instead of taking the fast path.

python/runtime/cudaq/algorithms/py_state.cpp
huaweil-nv: Yes, updated. Since we only need a normalized host-side representation here, `cupy.asnumpy(...)` is sufficient and the extra `ascontiguousarray(...)` step was unnecessary.

python/runtime/cudaq/algorithms/py_state.cpp
huaweil-nv: Good catch. Updated the device pointer cast to use `std::uintptr_t` instead of `int64_t`, and added `<cstdint>` explicitly.

python/runtime/cudaq/algorithms/py_state.cpp
huaweil-nv: Good point. I updated `getExtents()` to validate the shape values explicitly before converting them to `size_t`. Negative extents now fail fast with an exception instead of being silently converted.

python/runtime/cudaq/algorithms/py_state.cpp
huaweil-nv: Agreed. `CuPyArrayMetadata` already carries the relevant fields, so we could pass it through directly. I left the intermediate `py::buffer_info` in place here to keep this PR smaller and focused on the correctness fix.

huaweil-nv:
Hi @sacpis @1tnguyen 
Could you take another look at this change when you have time?
Let me summarize the three main issues addressed by this PR.
1.
```
  Fix 1 — Strided CuPy arrays (original scope, pre-existing commits a4f0c1b7 + bcbeaae9)

  Problem

  Before this PR, cudaq.State.from_data(cupy_array) read CuPy inputs as a flat contiguous buffer starting from array.data.ptr. If the caller passed a non-contiguous CuPy array — a transposed view, a Fortran-ordered matrix, a ::step slice, etc. — the logical values were
  silently reordered. No error was raised; the state just contained the wrong numbers. For 2D density matrices this is especially dangerous because row/column ordering matters for the physics.

  Fix

  - Read the full __cuda_array_interface__ (shape + strides + typestr + data pointer) rather than trusting .data.ptr + element count alone.
  - Detect non-C-contiguous layouts via a stride consistency check.
  - For non-contiguous 1D arrays, and for 2D arrays on the dynamics target, fall back to cupy.asnumpy(...) which returns a C-contiguous host NumPy array; recurse into the existing NumPy path from there.
  - Contiguous arrays are unchanged: they still take the zero-copy GPU fast path (direct device pointer → state::from_data).

  Known trade-off 

  The canonicalization strategy is correct but not optimal for performance:

  - For non-contiguous inputs we go device → host → state → device, i.e. two cross-boundary copies. An arguably better strategy would be cupy.ascontiguousarray(x) (a device-side copy that produces a contiguous CuPy array), then keep the state on device. That would be one copy instead of two, and would preserve the "state lives on GPU" invariant for these inputs.  It's worth explicitly reconsidering.
  - For contiguous inputs (the overwhelmingly common case — anything created by cp.array(...), cp.zeros(...), cp.complex128(...) without subsequent slicing/transposing), there is zero performance impact. The stride-check short-circuits immediately and the code takes the same zero-copy GPU path as before. Reviewers should not worry about regressing the hot path.

  So the argument for landing the current form is: previously we silently produced wrong states for non-contiguous input; now we produce correct states, at the cost of one extra host round-trip on the rare path. I think that's clearly the right end of the trade-off, and ascontiguousarray can be a follow-up optimization.
```

2.
```
 Fix 2 — Reject non-square 2D CuPy on the dynamics target (commit 12ee343e)

  Problem

  A non-square 2D CuPy array passed to state.from_data on the dynamics target used to pass createStateFromPyBuffer silently, get flattened into a 1D buffer inside TensorStateData, and only fail much later inside initialize_cudm() with a cryptic Invalid hilbertSpaceDims for the state data. Users would have no pointer back to the real issue (their input wasn't a density matrix shape). Crucially, the host NumPy 2D path has always correctly rejected non-square input at from_data time with a clear message — it's only the CuPy path that was asymmetric.

  Fix

  Four lines in createStateFromPyBuffer, mirroring the host validation:

  if (info.shape.size() == 2 && info.shape[0] != info.shape[1])
    throw std::runtime_error(
        "state.from_data 2D array (density matrix) input must be "
        "square matrix data.");

  Added test_from_data_cupy_2d_non_square_rejected to lock in the behaviour.
```

3. 
```
  Fix 3 — isDensityMatrix propagation in CuDensityMatState::createFromSizeAndPtr (commit be839f3b)

  This is an independent, pre-existing bug that surfaced while verifying Fix 1. I'd especially appreciate a careful look here.

  What's wrong

  CuDensityMatState::createFromSizeAndPtr has historically computed a local bool isDm (true when the incoming extents are 2D) and used it to decide how to interpret size. Since PR #2853 (May 2025) — "Migrate Python dynamics solver implementation to pybind11" — the
  constructor's isDm parameter was trimmed, and the function was updated to call the new two-arg constructor, but the equivalent assignment (result->isDensityMatrix = isDm;) was never added. The isDm local has been dead for ~11 months.

  Observable consequence

  For any cudaq.State.from_data(cupy_2d_array) on the dynamics target, the resulting state had:

  - state.getTensor().extents == [N*N] instead of [N, N]
  - np.array(state).shape == (N*N,) instead of (N, N)
  - state.is_density_matrix-ish code path thinks it's a state vector

  The inconsistency was masked because cudaq.evolve() internally calls initialize_cudm(), which re-infers the shape from hilbertSpaceDims, so end-to-end evolve tests still produced the right final state. No existing test inspected the state between from_data and evolve, so the bug went undetected.

  I verified this is a real observable bug (not a theoretical one):

  cupy 2x2 square C-contiguous on dynamics:
    Before fix: getTensor().extents = [4]      np.array(state).shape = (4,)
    After fix:  getTensor().extents = [2, 2]   np.array(state).shape = (2, 2)

  host numpy 2x2 square on dynamics (unaffected baseline):
    Both:       getTensor().extents = [2, 2]   np.array(state).shape = (2, 2)

  Fix

  One line after construction:

  auto result = std::make_unique<CuDensityMatState>(size, devicePtr);
  result->isDensityMatrix = isDm;
  return result;

  dimension already stores the flat element count (N*N for density matrices), and getTensor() / operator() already apply sqrt() to it when isDensityMatrix is true, so no other field needs to change. I followed the exact same post-construction assignment style that createFromData already uses for the complex_matrix host path a few lines above, so this should feel familiar.

  Added test_from_data_cupy_2d_square_metadata_preserved_pre_evolve (parametrized over c_order, fortran_order, transpose_view) that inspects the state before evolve(). I deliberately verified that this test fails on c_order without the fix (asserts [4] == [2, 2]) — so it's a real regression guard, not a tautology.
```

1tnguyen (approved):
LGTM 👍 Thanks @huaweil-nv

sacpis:
Will review the latest comments from @huaweil-nv in some time.

sacpis:
LGTM. Thanks @huaweil-nv.

sacpis:
> /ok to test 0d30a45

huaweil-nv:
> /ok to test 6d30f1a280ac5d8a7ff1d57f98d453694db9f7e8

huaweil-nv:
> /ok to test dea00498067d61cc156d37180b16621eafd085f7

PR #4329, opened 2026-04-15
Add CompiledModuleBuilder

lmondada:
I was seeing how the building of `CompiledModule`s was getting more complicated (with too many complicated `friend` declarations), so I've implemented your earlier suggestion, i.e. using a builder class to construct `CompiledModule` instances.

I've called the new file in the compiler library `CompiledModuleHelper` rather than `CompiledModuleBuilder` because there will be more types that could fit in there, such as ways to access the type erased data within `CompiledModule` with proper MLIR casting around them.

### DIFF
--- a/runtime/cudaq/platform/default/python/QPU.cpp
+++ b/runtime/cudaq/platform/default/python/QPU.cpp
@@ -23,6 +23,7 @@
 #include "cudaq/Verifier/QIRLLVMIRDialect.h"
 #include "cudaq/platform.h"
 #include "cudaq_internal/compiler/ArgumentConversion.h"
+#include "cudaq_internal/compiler/CompiledModuleHelper.h"
 #include "cudaq_internal/compiler/JIT.h"
 #include "cudaq_internal/compiler/RuntimeMLIR.h"
 #include "mlir/ExecutionEngine/ExecutionEngine.h"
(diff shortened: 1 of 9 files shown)


runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -359,12 +360,12 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
...
-    auto resultInfo = createResultInfo(resultTy, isEntryPoint, module);
 
     if (auto jit = alreadyBuiltJITCode(name, rawArgs)) {
-      cudaq::CompiledModule ck(name, resultInfo);
-      ck.attachJit(*jit, isFullySpecialized);
-      return ck;
+      CompiledModuleBuilder factory(name);
+      factory.setResultInfo(resultTy, isEntryPoint, module);
+      factory.attachJit(*jit, isFullySpecialized);
+      return factory.finish();
```
Renaud-K, line 368: Thank you for considering a factory. I think of them as stateless though. 
Something like this maybe:
```
CompiledModule cm = CompiledModuleBuilder::create(name, resultInfo);
cm.attachJit(*jit, isFullySpecialized)
```

lmondada: Yes, I would like that too.

The static factory method wasn't working well for me because whilst `attachJit` is MLIR-free (and thus can be a method on `CompiledModule`), other parts of the factory will require MLIR linking (such as `attachMlir`). Whether we want to have methods on `CompiledModule` that are implemented in a different library is a design decision we can discuss, but it seemed wrong to me.

So far we've resolved this using free functions (such as `createResultInfo`), but we'll have to add more free functions in the future, such as `attachMlir`.

lmondada: writing down from our conversation: the alternative is to have multiple static functions within the `CompiledModuleFactory` type, corresponding to all the ways in which the `CompiledModule` can be constructed.

This is how LLVM/MLIR typically does it.

It's a good (and more idiomatic) approach, as long as we can convince ourselves that there won't be an unmanageable number of overloads.

Renaud-K (approved):
Thank you for considering a factory,

lmondada:
Superseded by #4338

PR #4336, opened 2026-04-16
Fix sampling with mixed-basis measurements (mx/my)

### LINKED ISSUE #4333: Kernels with mx measurements not reporting correct results during sampling
### Required prerequisites

### Describe the bug

When a kernel contains an mx measurement after an earlier measurement such as mz, cudaq.sample(...) only reports the measurement outcomes collected after the mx instruction instead of returning the full bitstring for all measurements in the kernel.

For example, in a 2-qubit kernel that measures qubit 0 with mz and qubit 1 with mx, the sampled result should contain 2-bit outcomes, but instead only a 1-bit outcome is returned.

Python example:
```
import cudaq

@cudaq.kernel
def kernel_mx():
    qubits = cudaq.qvector(2)
    x(qubits[0])
    h(qubits[1])
    mz(qubits[0])
    mx(qubits[1])

result = cudaq.sample(kernel_mx, shots_count=10)
print(result)
```

Observed output:
```
{ 0:10 }
```

Expected output shape:
```
{ 00:10 }
```

or, more generally, a 2-bit result that includes both the mz outcome for qubit 0 and the mx outcome for qubit 1.

C++ example:
```
#include <cudaq.h>

__qpu__ void kernel() {
  cudaq::qvector qubits(2);
  x(qubits[0]);
  h(qubits[1]);
  mz(qubits[0]);
  mx(qubits[1]);
}

int main() {
  auto result = cudaq::sample(kernel);
  result.dump();
}
```

This appears to reproduce in C++ as well.

### Expected behavior

cudaq.sample(...) should return bitstrings that include all measurement outcomes performed in the kernel, regardless of whether those measurements are in the Z basis, X basis, or mixed bases.

For the reproducer above, the result should include both measured bits, not just the measurements that occur after the mx instruction.

Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 0.14.0
- **Python version**: 3.12.3
- **C++ compiler**: g++ 13.3.0
- **Operating system**: ubuntu (wsl)

### Suggestions

This looks like sampled measurement aggregation may be dropping previously recorded measurement results once an mx instruction is encountered, or only collecting results from a later measurement register/path. It would be worth checking how mixed-basis measurement results are appended into the final sampled bitstring.

anpaz:
Fixes #4333

### Root Cause

`mx(q)` is implemented as `h(q); mz(q)` — a basis-change gate followed by a Z-basis measurement. The problem is that in sampling mode every gate application calls `flushAnySamplingTasks()`, which samples all previously recorded qubits and **clears the measurement buffer** (`sampleQubits`). So the intrinsic `h` gate inside `mx` prematurely flushes measurements recorded before it, and only the qubits measured after the flush end up in the final result.

For example, in a kernel that does `mz(q[0]); mx(q[1])`:
1. `mz(q[0])` records qubit 0 in `sampleQubits`
2. `mx(q[1])` calls `h(q[1])`, which triggers `flushAnySamplingTasks()` — this samples qubit 0 and clears `sampleQubits`
3. `mz(q[1])` (inside `mx`) records qubit 1 in `sampleQubits`
4. At finalization, only qubit 1 is in `sampleQubits`, producing 1-bit results instead of 2-bit

### Fix

Gate operations now only trigger a sampling flush when they operate on a qubit that has already been measured (i.e., already present in the sampling buffer). Previously, every gate unconditionally called `flushAnySamplingTasks()`, which caused basis-change gates from `mx`/`my` decomposition (`h(q)`, `s†(q)`) to prematurely sample only the qubits measured so far, truncating bitstrings.

With this change, `h(q)` from the `mx(q)` → `h(q); mz(q)` decomposition no longer triggers a flush because `q` hasn't been measured yet. A flush still occurs when a gate operates on an already-measured qubit, preserving mid-circuit measurement semantics (measure, apply gate, re-measure).

### Testing

Added regression tests (C++ and Python) that verify mixed-basis measurement kernels return correctly-sized bitstrings with proper qubit ordering.


.gitignore
```diff
@@ -87,6 +87,7 @@ _version.py
 
 # Generated files
 python/tests/mlir/generated
+Testing
```
mitchdz, line 90: Is this file being generated from somewhere new? If it's just local for you, add it to your `.git/info/exclude`

anpaz: Not sure. This folder appeared with test results. I ran the tests from the cli using `ctest` and `pytest`. I also triggered the tests from VSCode, so I'm not sure what exactly is creating the folder.

bettinaheim: I think it may be a VS Code extension that is generating that if I remember correctly.

runtime/common/ExecutionContext.h
```diff
@@ -63,6 +63,11 @@ class ExecutionContext {
   /// statements on measure results.
   bool hasConditionalsOnMeasureResults = false;
 
+  /// @brief When true, suppress mid-circuit sampling flushes. Used by
+  /// measurement basis-change operations (`mx`/`my`) to prevent their internal
+  /// gate decomposition from prematurely flushing the sample qubit buffer.
+  bool deferSamplingFlush = false;
```
khalatepradnya: I think we don't want to add new fields to the `ExecutionContext`...
Tagging @Renaud-K who can share more.

Renaud-K: `qubit_qis.h` has managed to avoid using the ExecutionContext so far. It uses the ExecutionManager instead to propagate instructions to the simulators. We should be able to use a similar mechanism to propagate this flag.

anpaz: Yeah, I did ping Renaud about this. I found a different solution that doesn't require flags on ExecutionContext.

runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -430,7 +436,26 @@ class StimCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
           fmt::format("Gate not supported by Stim simulator: {}. Note that "
                       "Stim can only simulate Clifford gates.",
                       task.operationName));
-    else if (gateName == "SDG")
```
Renaud-K, line 433: Are these changes related to this fix?

anpaz: Yes, in the sense that the test I added were failing with the stim simulator. I believe that the tests automatically run against all simulators and `my` uses the `R1` gate instead of an `S_adj` gate (they are equivalent with the same angle).

runtime/nvqir/CircuitSimulator.h
```diff
@@ -640,6 +640,18 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
+  /// should trigger a sampling flush (only needed when the gate operates
+  /// on a qubit that was already measured, i.e. mid-circuit measurement).
+  bool operatesOnMeasuredQubit(const std::vector<std::size_t> &qubits) const {
+    for (auto q : qubits)
+      if (std::find(sampleQubits.begin(), sampleQubits.end(), q) !=
+          sampleQubits.end())
+        return true;
+    return false;
+  }
+
```
Renaud-K, line 654: I have looked around the file for how we look up qubits. I see a lot of `std::find` uses. I guess the input qubits vector should be small because these are the gate's qubits, right? Otherwise we would have quadratic complexity. I am ok with it but it might be something we need to keep an eye on.

Renaud-K:
Things look good to me. It might be worth testing this in cudaqx which will test the simulation much more deeply.

Renaud-K (approved):
Things look good to me. Thank you for looking at an alternative. It might be worth testing these changes against cudaq-x.

runtime/nvqir/CircuitSimulator.h
anpaz: Agreed. Input qubits is very small (typically <=2), also, `sampleQubits` is also pretty small as it only contains qubits that have been queued for measurement, so this shouldn't be problematic.
CI wise, I don't see any degradation on runtime for the tests in the CI build, so I think we should be good.

PR #4338, opened 2026-04-16
Add stateless CompiledModuleHelper

lmondada:
This is a rewrite of #4329, using a stateless class with static functions rather than a builder pattern.


Renaud-K (approved):
Looks good. Thank you!

PR #4339, opened 2026-04-16
Introducing execution policies. Starting with providing results as return values.

Renaud-K:
This is the initial PR in a series aimed at reducing the use of the ExecutionContext and ultimately providing kernel launch overloads based on execution policy. The end goal is to arrive at a launch API on the QPU side that would look like as follows:
```cpp
sample_result QPU::launchKernel(cudaq::sample_policy,  const std::string &kernelName, CompiledModule cm); 
observe_result QPU::launchKernel(cudaq::observe_policy, const std::string &kernelName, CompiledModule cm);
...
```
Whereby, each overload returns its own result, instead of populating an overextended ExecutionContext.

This one PR focuses on getting the `CircuitSimulator` and the `ExecutionManager` family of classes to return an actual value instead of populating the ExecutionContext. Because of the incremental nature of the work the class hierarchies are untouched and the ExecutionContext is still populated but this step can easily be eliminated when the whole flow comes together.

## 1. API Changes
The  API of `CircuitSimulator` and `ExecutionManager` changes from
```cpp
virtual void measureSpinOp(const cudaq::spin_op &op);
virtual void finalizeExecutionContext(cudaq::ExecutionContext &ctx);
```
to
```cpp
virtual cudaq::SpinMeasureResult measureSpinOp(const cudaq::spin_op &op);
void finalizeExecutionContext(cudaq::ExecutionContext &ctx);
virtual cudaq::sample_result finalizeExecutionContext(const cudaq::sample_policy &policy, cudaq::ExecutionContext &ctx);
virtual void                            finalizeExecutionContext(const cudaq::other_policies &policy,       cudaq::ExecutionContext &ctx);
```

- `measureSpinOp` now returns a value.

- The virtual `finalizeExecutionContext` becomes non-virtual and dispatches to policy-based virtual overloads. 

- The `cudaq::sample_policy` overload returns a `cudaq::sample_result`.

- There is a fallback overload `cudaq::other_policies`, for all other policies. This is useful for policies that do not require any particular specialization.  It is easy to add overloads as needed. It would not require any dispatching changes. 

## 2. Internal changes
 `CircuitSimulator` and `ExecutionManager` used to rely on the ExecutionContext to store intermediate simulation results. This is no longer the case. Intermediate results are now stored internally in `internalResults`. This is also reflected in the API of `measureSpinOp` which gets its value from internals. 

## 3. Dispatching
1. Runtime dispatching.
A runtime dispatching is implemented using a utility provided by the policies. We can see below that the dispatcher catches the results from the overload and updates the `ExecutionContext`. When the flow is complete, this step will be eliminated and the overload will be called directly.

```cpp
void ExecutionManager::finalizeExecutionContext(ExecutionContext &ctx) {
  policies::withPolicy(ctx.name, [&](auto policy) {
    policies::visitResult(
        [&]() { return finalizeExecutionContext(policy, ctx); },        
        [&](sample_result &&r) { ctx.result = std::move(r); },  // <---- Updates ctx!
        [&](policies::void_result &&r) {});
  });
```
2. Customization Point Objects
To provide a fallback mechanism in the many cases when an overload does not exist for a given policy,
CPOs are introduced in `policy_cpos.h`. They provide a default implementation for policy overloads that are absent.
These can be discovered through ADL with hidden friends in the policies as illustrated in this PR.

### File location
Policies were added under `runtime/cudaq/algorithm` in a directory corresponding to the algorithm.
Common files (`any_policy.h`, `policies.h` and `policy_dispatch.h`) are under `runtime/cudaq/algorithm`
```
runtime/cudaq/algorithm
                    ├── any_policy.h
                     ...
                    ├── observe
                    │   ├── options.h
                    │   └── policy.h
                    ├── observe.h
                    ...
                    ├── policies.h
                    ├── policy_dispatch.h
                    ├── resource_estimation.h
                    ├── run
                    ├── run.cpp
                    ├── run.h
                    ├── sample
                    │   ├── options.h
                    │   └── policy.h
                    ├── sample.h

```

### Design

Policies are small objects, largely used as tags for dispatching. 

- They provide hidden friends for ADL lookup of policy overloads.
- Although not currently used, they are also meant to carry launch options.

```cpp
struct sample_policy : public any_policy {
  using result_type = sample_result;
  sample_options options;

  friend sample_result
  finalize_execution_manager_impl(ExecutionManager &mgr,
                                  const sample_policy &policy,
                                  ExecutionContext &ctx);
  friend sample_result
  finalize_simulation_circuit_impl(nvqir::CircuitSimulator &sim,
                                   const sample_policy &policy,
                                   ExecutionContext &ctx);
};
```

## 5. Future work

1. Consolidate all launch apis into one.
There are currently 3 entry points in the QPU, that should be consolidated into one in order to avoid a sprawl of overloads
```cpp
  launchKernel(const std::string &name, KernelThunkType kernelFunc, void *args,
               std::uint64_t, std::uint64_t,
               const std::vector<void *> &rawArgs) = 0;

  launchModule(const std::string &name, mlir::ModuleOp module,
               const std::vector<void *> &rawArgs);

  [[nodiscard]] virtual CompiledModule
  specializeModule(const std::string &name, mlir::ModuleOp module,
                   const std::vector<void *> &rawArgs, bool isEntryPoint);
```

2. Manage ExecutionManager locally
The unified `launchKernel` is still being designed but it might look like something below. The result would be extracted from the ExecutionManager upon leaving the launch. 
```cpp
sample_result QPU::launchKernel(cudaq::sample_policy, const std::string &kernelName, CompiledModule cm) {
  ExecutionManager em(getCurrentContext());
  cm.launch(kernelName);
  return em.finalize();
}
```

4. Split the JIT compiler with overloads as well.

```cpp
  std::vector<cudaq::KernelExecution>
  lowerQuakeCode(cudaq::ExecutionContext *executionContext,
                 const std::string &kernelName, void *kernelArgs,
                 const std::vector<void *> &rawArgs);
```
```cpp
  std::vector<cudaq::KernelExecution>
  lowerQuakeCode(cudaq::sample_policy, CompiledModule);
```
5. Implement all policies
Currently, only the sample_policy and observe_policy are implemented. This has been enough for this initial work.

6. Bring it all together

To tie it all together, we will be using Architecture's executor_api  proposal.
This will allow us to register a closure within the ExecutionContext to carry return value from QPU launch to user launch.
```cpp
ctx->execute_kernel =
        [qpu_ptr](const char *kernelName, KernelThunkType kernel, void *args,
                  std::uint64_t argsSize, std::uint64_t resultOffset,
                  const std::vector<void *> &rawArgs) {
          return qpu_ptr->launch_kernel(kernelName, kernel, args, argsSize,
                                        resultOffset, rawArgs);
        };

```


Renaud-K:
> /ok to test 27977ff0f634a8053e38e75a6eec136f950db950

runtime/cudaq/algorithms/any_policy.h
```diff
@@ -0,0 +1,13 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+namespace cudaq {
+struct any_policy {};
```
amccaskey: Instead of `any_policy` base class, why not start this incremental move toward the v2 runtime with Concepts and hidden friend / CPO? Wondering if something like the following would retain the incremental nature of this PR but also help move us closer to the intended design without requiring subsequent removals later on: 

```cpp 
namespace cudaq {
/// Any type with a result_type alias is a policy.                                                                         
template <typename T>
concept ExecutionPolicy = requires { typename T::result_type; };                                                                                                                                                        

namespace detail {                                                                                                                             
  void tag_invoke_finalize() = delete;  // poison pill: blocks unqualified lookup                                            
  
  struct finalize_fn {                                                                                                       
    template <ExecutionPolicy P>
      requires requires(const P &p, ExecutionContext &c, ExecutionManager &m) {
        tag_invoke_finalize(p, c, m);                                                                                        
      }                                                                                                                      
    auto operator()(const P &p, ExecutionContext &c, ExecutionManager &m) const {                                            
      return tag_invoke_finalize(p, c, m);                                                                                   
    }                                                                                                                        
  };
                                                                                                                             
} // namespace detail
 
/// CPO: external policies provide a hidden friend `tag_invoke_finalize`.                                                  
inline constexpr detail::finalize_fn finalize_execution{};
                                                                                                                             
} // namespace cudaq
``` 
and then 
```cpp
struct sample_policy {                      // no base class                                                               
  using result_type = sample_result;
  sample_options options;                                                                                                  
                                                                                                                             
  // Hidden friend — found by ADL through cudaq::finalize_execution
  friend sample_result                                                                                                     
  tag_invoke_finalize(const sample_policy &p, ExecutionContext &ctx,                                                       
                      ExecutionManager &mgr) { 
     return mgr.finalize_sample(p, ctx);       // typed virtual
  }                                                 
};
```
then you keep your `withPolicy` / `withResult` pattern, but call to the CPO to get the right execution policy friend function and process results 
```cpp
void ExecutionManager::finalizeExecutionContext(ExecutionContext &ctx) {                                                   
    policies::withPolicy(ctx.name, [&](auto policy) {
      policies::visitResult(                                                                                                 
          [&]{ return cudaq::finalize_execution(policy, ctx, *this); },                                                      
          [&](sample_result&&  r) { ctx.result = std::move(r); },                                                            
          [&](observe_result&& r) { /* expectation + raw_data write-back */ },                                               
          [&](policies::void_result) {});                                                                                    
    });                                                                                                                      
  }   
```
and same for CircuitSimulator.

IMO, this remains incremental but gets us closer to where we want to be:                                                                                                                              
Gets:
  - No need for the any_policy base that would have to go away later anyway.                     
  - The set of policies is open at the type axis (the CPO) even though the manager is still virtual at the receiver axis.    
  - The hidden friends on sample_policy / observe_policy are (or can be similar to) the same hidden friends the v2 `cudaq::launch` will dispatch to — just with the QPU as the third argument instead of the manager. Backends written today against this CPO carry forward without rewrite.                                                                                                           

Doesn't get:                                                                                                               
  - Virtual dispatch isn't eliminated — `mgr.finalize_sample(p, ctx)` stays virtual until the simulator type becomes statically known at the call site. That's a separate (later) refactor.                                                               
  - `withPolicy` string registry is unchanged. The right fix for that is still to plumb the policy type through from cudaq::sample() / observe() so this layer doesn't need a string lookup at all — but that's an orthogonal change and a separate comment.

runtime/cudaq/algorithms/policy_dispatch.h
```diff
@@ -0,0 +1,199 @@
...
+template <typename Func>
+void withPolicy(std::string_view name, Func &&func) {
+  using FuncRef = std::remove_reference_t<Func> &;
+  using Entry = std::pair<std::string_view, void (*)(FuncRef)>;
+
+  // One static array per Func instantiation — initialized once, no heap
+  // allocation. To add a new policy, append an entry here and define the policy
+  // struct above.
+  static const Entry registry[] = {
+      {"sample", [](FuncRef f) { f(sample_policy{}); }},
```
amccaskey, line 83: The callers in `runtime/cudaq/algorithms/sample.h` and `observe.h` already know the policy type at the point where they construct `ExecutionContext("sample", …)`. Can they call `finalizeExecutionContext(sample_policy{…}, ctx)` directly and bypass the string lookup? Then `withPolicy(ctx.name, …)` is only needed at the Python / C-API seam where the type really is erased.

runtime/cudaq/algorithms/sample/policy.h
```diff
@@ -0,0 +1,38 @@
...
+/// @brief Tag and options for sampling quantum circuit measurements.
+struct sample_policy : public any_policy {
+  /// Associated result type for synchronous APIs keyed off this policy.
+  using result_type = sample_result;
+
+  /// Sampling  options.
+  sample_options options;
+};
+
+namespace async {
```
amccaskey: You could instead have a single async_policy generic wrapper: 
```cpp
template<typename InnerPolicy>
struct async_policy { 
  using result_type = std::future<typename InnerPolicy::result_type>; 
  ... execution hidden friend is just std::async(...) on InnerPolicy::execute 
} 
```

runtime/nvqir/cudensitymat/CuDensityMatSim.cpp
```diff
@@ -137,6 +138,19 @@ class CuDensityMatSim : public nvqir::CircuitSimulatorBase<double> {
           context.name));
   }
 
+  cudaq::sample_result
+  finalizeExecutionContext(const cudaq::sample_policy &policy,
```
amccaskey, line 142: Does this mean that the method should not be pure-virtual at CircuitSimulator level? Should we add a default on the base that throws an exception for now?

runtime/cudaq/algorithms/policy_dispatch.h
amccaskey: if kept, should we use `std::array`?

runtime/cudaq/qis/execution_manager.h
```diff
@@ -58,6 +59,11 @@ class ExecutionManager {
   /// Internal - At qudit deallocation, return the qudit index
   void returnIndex(std::size_t idx) { tracker.returnIndex(idx); }
 
+  virtual void finalizeExecutionContext(const any_policy &policy,
```
amccaskey: fully understand that this is incremental, but just want to note that in a world where we remove library mode we will just delete all this ExecutionManager code.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1292,22 +1321,19 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
-  void measureSpinOp(const cudaq::spin_op &op) override {
+  cudaq::SpinMeasureResult measureSpinOp(const cudaq::spin_op &op) override {
     auto executionContext = cudaq::getExecutionContext();
 
     if (nQubitsAllocated == 0) {
       if (executionContext)
-        executionContext->expectationValue = 0.0;
-      return;
+        return cudaq::SpinMeasureResult(0.0, {});
     }
```
amccaskey, line 1330: if nQubits != 0 and executionContext is null, what happens?

amccaskey:
Thanks @Renaud-K, this is a nice incremental change. I posted a few comments above. My main questions are surrounding the use of an `any_policy` base type. Wondering if we can retain the usage of the virtual simulator methods, execution context, etc. but move toward the policies carrying their intended execution dispatch ADL/CPO code.

runtime/cudaq/algorithms/policy_dispatch.h
Renaud-K: Absolutely. They are meant to be called directly. The dispatch is only intended to be scafolding.

runtime/cudaq/algorithms/sample/policy.h
Renaud-K: I was thinking along similar lines. We have our own future type that we will need to keep but I can see about generalizing the definition. I might remove them from this PR because they are unused and any work on them would be done in a vacuum. Maybe we can revisit when they are re-introduced?

runtime/cudaq/algorithms/any_policy.h
Renaud-K: I do need to keep `any_policy` for the fallback mechanism right? The other policies just do not have to derive from it.

Renaud-K: Also, I really do not like the fact that there is now a circular dependency between the ExecutionManager and the policies. The execution ought to be outside of the policies. I am currently looking at a possible implementation.

runtime/cudaq/algorithms/policy_dispatch.h
Renaud-K: Plus, we are only passing a default constructed policy. It will have very limited use. Ultimately we want to be able to pass meaningful information.

I do not have an opinion about `std::array`. C++ was able to use a range-based for loop on the C-style array as well. `std::array` would make sense if we needed the size as an integral template argument.

runtime/cudaq/qis/execution_manager.h
Renaud-K: I did not know this was introduced for library mode but that's fine.

runtime/nvqir/cudensitymat/CuDensityMatSim.cpp
Renaud-K: `CircuitSimulatorBase` has an implementation, so all simulators have one. So a pure virtual should be fine. Did I miss something?

runtime/nvqir/CircuitSimulator.h
Renaud-K: This is why I hate these one-liner rules on if statements. There should always be curly braces. Good catch. It should not matter if there is an executionContext or not. There is nothing useful to report.

runtime/cudaq/algorithms/any_policy.h
Renaud-K: I have pushed changes that remove `any_policy` has base class and use CPOs instead. It would have been simplest to just add the CPOs in the respective files (execution_manager.h and CircuitSimulator.h) which would have required no ADL but to illustrate the ADL mechanism and create a precedent for it, I have added hidden friends in the policies. I did it in a manner that does not introduce any circular dependencies.

Renaud-K:
> /ok to test 9858e04275f2f9d01818ed6066695d1ae1f75663

runtime/cudaq/algorithms/any_policy.h
Renaud-K: I also want to offer with 9246400dbcd2b0410b2ff47c399caa45f23ed774, a version of the CPO that does not require ADL. It just uses a qualified name. Both the ADL and non-ADL versions need a forward declared anchor. Once it is found in the namespace all the other overloads are found. The ADL is useful to find functions with unqualified names. I am not sure we need it. You pick.

Renaud-K:
> /ok to test 9246400dbcd2b0410b2ff47c399caa45f23ed774

runtime/cudaq/algorithms/policy_dispatch.h
amccaskey: I had called out std::array because I saw that <array> was included. Either is fine with me, but you might remove the array include then.

runtime/cudaq/algorithms/any_policy.h
amccaskey: Either approach works. The CPO + ADL sets us up for where we want to be. CPO with qualified works now since this still ultimately uses virtual dispatch. My gut would say get us thinking now about ADL.

Renaud-K: I restored the ADL version.

runtime/cudaq/algorithms/policy_dispatch.h
Renaud-K: I have cleaned up unnecessary headers.

lmondada (commented):
Thank you Renaud, this is a great step towards moving things out of the `ExecutionContext`. Using overloads based on policy is a very neat solution to handling the heterogeneous types. Are you waiting to get additional eyes on the PR? Happy to approve whenever you feel ready to merge.

Question about the direction we're going in (you don't have to have an answer): I have in mind your "n + m" picture from discussions months ago. The idea was that if we have `n` launch policies and `m` QPUs, we want to try to get as close to `n + m` implementations as possible, as opposed to `n * m`. In this PR, we are introducing a parametrisation on the `n` launch policies and are letting each simulator implement these. There's hence a risk that we will end up with `n * m`.

Do you have an opinion on where you would like to see a boundary between functions parametrised on policy versus parametrised/dispatched on QPU/simulator?

lmondada:
I confirm that I am working in the same direction as what you are outlining in the PR description :)

On the ideal signature of `QPU::launchKernel`:
> ```c++
> sample_result QPU::launchKernel(cudaq::sample_policy, const std::string &kernelName, CompiledModule cm) {
>    ...
> }
> ```

If we follow the current plans, then `kernelName` should be redundant (stored within `CompiledModule`), but you would need to pass kernel arguments. I've been using a placeholder type named `KernelArgs` for this (does not exist at the moment).

Same comment for `lowerQuakeCode`.

Renaud-K:
>In this PR, we are introducing a parametrization on the n launch policies

Not really. If anything, we are exposing existing parametrization. We are just hiding `if (ctx->name == "sample")` in a dispatching utility. Now the goal is to eliminate this dispatch with direct calls to overloads but this parametrization has been there all along. 

Note that most of it is being done in the base class which makes m closer to one. But this change does not improve or worsen this situation.

Renaud-K:
> /ok to test 32ca65a2b939a09061f2da271abc8cf63ad2b329

lmondada (approved):

runtime/cudaq/algorithms/policy_cpos.h
```diff
@@ -0,0 +1,107 @@
...
+
+namespace detail {
+
+/// @brief Detects whether a policy-specific @c finalize_execution_manager_impl
+///        overload exists for type @p T (found via ADL through hidden friends).
+template <class T>
+concept has_em_custom_finalize =
+    requires(cudaq::ExecutionManager &mgr, const T &policy,
+             cudaq::ExecutionContext &ctx) {
+      cudaq::finalize_execution_manager_impl(mgr, policy, ctx);
```
sacpis: Question, isn't using `finalize_execution_manager_impl` disables ADL? This is contrary to the comment on line 14.

runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
```diff
@@ -165,6 +165,17 @@ class DefaultExecutionManager : public cudaq::BasicExecutionManager {
...
+  sample_result finalizeExecutionContext(const sample_policy &policy,
+                                         ExecutionContext &ctx) override {
+    finalizeExecutionContextImpl(ctx);
+    return simulator()->finalizeExecutionContext(policy, ctx);
+  }
+
+  void finalizeExecutionContext(const other_policies &policy,
+                                ExecutionContext &ctx) override {
+    finalizeExecutionContextImpl(ctx);
     simulator()->finalizeExecutionContext(ctx);
```
sacpis, line 179: Same as [Alex's comment](https://github.com/NVIDIA/cuda-quantum/pull/4339/changes#r3100947258) for policy_dispatch.

runtime/cudaq/algorithms/policies.h
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+
+#include "cudaq/algorithms/observe/policy.h"
+#include "cudaq/algorithms/sample/policy.h"
+
+namespace cudaq {
+
+/// @brief Fallback policy tag used when no specific policy matches.
+struct other_policies {};
```
sacpis, line 17: nit: Should we rename it to something else? like `default_policies` or `fallback_policies` ?

runtime/nvqir/CircuitSimulator.h
```diff
@@ -15,6 +15,9 @@
 #include "common/QuditIdTracker.h"
 #include "common/SampleResult.h"
 #include "common/Timing.h"
+#include "cudaq/algorithms/policies.h"
+#include "cudaq/algorithms/policy_cpos.h"
+#include "cudaq/algorithms/policy_dispatch.h"
```
sacpis, line 20: Should we forward declare these policy tags?

sacpis (approved):

runtime/cudaq/qis/managers/default/DefaultExecutionManager.cpp
Renaud-K: It is the same answer. Ultimately, yes. The goal is to remove the policy_dispatch. It is just scaffolding. But it can only be removed when all the parts come together. (See future work above).

runtime/cudaq/algorithms/policies.h
Renaud-K: I like `other_policies` better. There are no default or fallback policies, but default or fallback implementations.

runtime/nvqir/CircuitSimulator.h
Renaud-K: The implementation is in CircuitSimulator.h so they need to be fully known here. The policies are the ones doing the forward declarations with hidden friends and CPOs. In the future, the policies will be participating in TMP so they will always need to be fully included.

runtime/cudaq/algorithms/sample/policy.h
```diff
@@ -0,0 +1,32 @@
...
+#include "cudaq/algorithms/sample/options.h"
+
+namespace nvqir {
+class CircuitSimulator;
+}
+
+namespace cudaq {
+
+class ExecutionManager;
+class ExecutionContext;
```
1tnguyen, line 21: Just a question: are these forward declarations intentional as they seem not needed for this file?

Renaud-K: Something went wrong when I tried to restore the ADL version. Let me try again.

runtime/cudaq/algorithms/policy_cpos.h
Renaud-K: Something went wrong when I tried to restore the ADL version. Let me try again.

Renaud-K:
> /ok to test 98c368bf3ebf1af80c10da9a24499359357a8ca6

Renaud-K:
> /ok to test c0cebdc3e03ca8eb9b9e03d22f2269ef17a39795

PR #4346, opened 2026-04-17
[codegen] Restore the OpenQASM2.0 pipeline

### LINKED ISSUE #4341: Braket target submission fails in v0.14.0 due to multiple `creg` declarations in OQ2 output
### Required prerequisites

### Describe the bug

Starting in CUDA-Q v0.14.0, running `cudaq.sample()` with the Braket target (`cudaq.set_target("braket", ...)`) fails with:

```
RuntimeError: [line 10] cannot declare bit register. Only 1 bit register(s) is/are supported
```

This appears to be a regression from v0.13.0.

```python
import cudaq

@cudaq.kernel
def bell_state():
    qubits = cudaq.qvector(2)
    h(qubits[0])
    cx(qubits[0], qubits[1])
    mz(qubits)

cudaq.set_target("braket", machine="arn:aws:braket:::device/quantum-simulator/amazon/sv1")
result = cudaq.sample(bell_state, shots_count=100)
```

gives the following error returned from Braket:
```
RuntimeError: [line 10] cannot declare bit register. Only 1 bit register(s) is/are supported
```

### Expected behavior

This code should run correctly, successfully submit a task to Braket, and return the result, as it did prior to version 0.14.0.

0.13.0

### Environment

- **CUDA-Q version**: 0.14.0
- **Python version**: 3.13.5
- **C++ compiler**: n/a (using cudaq 0.14.0 installed from pypi)
- **Operating system**: tested on both macOS (ARM) and Linux (x86_64)

#### More details

In v0.14.0, `addPipelineTranslateToOpenQASM` in `lib/Optimizer/CodeGen/Pipelines.cpp` was refactored to call `createCommonTargetCodegenPipeline</*isJIT=*/true>()`, which includes `createExpandMeasurementsPass()`.

This appears to have been introduced in https://github.com/NVIDIA/cuda-quantum/pull/3423.

The pass pipeline for the Braket target executes in this order:

1. `braket.yml` `jit-mid-level-pipeline` runs `combine-measurements`, which is a no-op because the Python frontend already emits a single `mz` on the whole `qvector`.
2. `jit-finalize-pipeline` runs (no `ExpandMeasurements` here).
3. `add-measurements` runs (no-op, measurements already exist).
4. OpenQASM translation calls `addPipelineTranslateToOpenQASM`, which now includes `ExpandMeasurements`. This expands the single `mz %veq<2>` into a loop of individual per-qubit `mz` ops.
5. `LoopUnroll` unrolls the loop into two separate `quake.mz` ops.
6. The OpenQASM emitter generates a separate `creg` declaration for each `mz` op.
7. Amazon Braket's OpenQASM parser rejects the program because it only allows one classical bit register.

The generated OpenQASM looks like:
```
qreg var0[2];
h var0[0];
cx var0[0], var0[1];
creg var3[1];          // first creg
measure var0[0] -> var3[0];
creg var4[1];          // second creg - Braket rejects this
measure var0[1] -> var4[0];
```

When it should be:
```
qreg var0[2];
creg var1[2];          // single creg
h var0[0];
cx var0[0], var0[1];
measure var0 -> var1;
```

#### Possible fixes

One of these approaches may work:
1. Remove `ExpandMeasurements` from the `addPipelineTranslateToOpenQASM` pipeline (restoring the v0.13.0 behavior)
2. Run `combine-measurements` again after `ExpandMeasurements` + `LoopUnroll` in the OpenQASM translation pipeline.

khalatepradnya:
## Background

`cudaq.sample` with `set_target("braket")` fails on v0.14.0+ with:

    RuntimeError: [line 10] cannot declare bit register. Only 1 bit register(s) is/are supported

Amazon Braket's OpenQASM 2.0 parser enforces exactly one classical register per circuit. The payload CUDA-Q emits for the Bell-state reproducer in #4341 contains two.

## Root cause

`addPipelineTranslateToOpenQASM` (`lib/Optimizer/CodeGen/Pipelines.cpp`) was refactored in #3693 to run  `ExpandMeasurements` unconditionally. For `qasm2` backends that run `combine-measurements` in the mid pipeline (Braket, Scaleway, Quantum Machines), the sequence becomes:

1. Mid pipeline: `combine-measurements` merges per-qubit measurements into a single `quake.mz` on the whole `!quake.veq` -   the intent being "emit one `creg` spanning all qubits".
2. Translate pipeline: `ExpandMeasurements` re-expands the combined `mz` into one `mz` per qubit, then loop-unrolls.
3. OpenQASM2.0 emitter: writes one `creg` declaration per `mz`.

Target-specific YAML intent is silently overridden in the translate pipeline.

## Fix

1. `lib/Optimizer/CodeGen/Pipelines.cpp`: revert `addPipelineTranslateToOpenQASM` to the thin cleanup it was before
   #3693 . Each backend's YAML now drives measurement expansion.
2. `infleqtion.yml` and `tii.yml`: add `jit-high-level-pipeline: "expand-measurements"`.  These targets previously depended on the unconditional expansion to get one `creg` per measured qubit; the explicit entry preserves that behavior.
3. `test/Translate/OpenQASM/basic.qke` and `test/Translate/openqasm2_*.cpp`: update CHECK lines to match the single-`creg` output for a vector `mz` (which is what the emitter produces after the fix).

## Impact

| Backend | creg count for `mz(qvector(n))` |
|---|---|
| Braket, Scaleway, Quantum Machines | 1 (single `creg` of size n) |
| Infleqtion, TII | n (preserved via new YAML entry) |
| Quantinuum, IQM, OQC, Anyon, QCI | n (unchanged; already had `expand-measurements` in YAML) |

The change is scoped to `addPipelineTranslateToOpenQASM`, which only runs for `codegen-emission: qasm2`. Simulators and non-OpenQASM2.0 backends are unaffected.

## Testing

- `ninja check-cudaq-mlir` passes with the updated CHECK lines.
- `cudaq.translate(kernel, format="openqasm2")` under `set_target(...)`
  for Braket, Scaleway, Infleqtion, TII — creg counts match the matrix
  above.
- Reproducer from #4341 now emits exactly the "expected" OpenQASM2.0 shown in the issue: `creg var3[2]; measure var0 -> var3;`.
- Manually tested against real servers: `test_braket.py`, `test_Infleqtion.py`, `test_tii.py`, `test_scaleway.py`.

## Follow-up

An automated local test set up for OpenQASM payload validator will be added in a separate PR.

Fixes #4341.


sacpis (approved):
Overall LGTM. Thanks @khalatepradnya. Just one suggestion. Should we add a small regression test for a target who mid-level pipeline contains combine-measurements now translates cleanly? Just as a confirmation that this bug is fixed.

runtime/cudaq/platform/default/rest/helpers/infleqtion/infleqtion.yml
```diff
@@ -18,7 +18,8 @@ config:
   gen-target-backend: true
   # Add preprocessor defines to compilation
   preprocessor-defines: ["-D CUDAQ_QUANTUM_DEVICE"]
-  # Define the JIT lowering pipeline
+    # Define the JIT lowering pipeline
+  jit-high-level-pipeline: "expand-measurements"
```
mitchdz: Remove whitespace
```suggestion
  # Define the JIT lowering pipeline
  jit-high-level-pipeline: "expand-measurements"
```

khalatepradnya:
> Overall LGTM. Thanks @khalatepradnya. Just one suggestion. Should we add a small regression test for a target who mid-level pipeline contains combine-measurements now translates cleanly? Just as a confirmation that this bug is fixed.

Thanks! Good suggestion - I added a test to cover that.

PR #4348, opened 2026-04-18
Fixing Krylov notebook failure

sacpis:
Skipping identity terms when building the Pauli word and coefficient lists passed to the Krylov kernel. Controlled exp_pauli does not handle the identity terms. We add their contribution back when assembling the Hamiltonian matrix.

Fixes https://github.com/NVIDIA/cuda-quantum/actions/runs/24584888146/job/71904057326#step:5:1955


1tnguyen (approved):
LGTM 👍

PR #4351, opened 2026-04-20
[Python] Fix overload resolution for `state.from_data` with MPS tensors

1tnguyen:
Fixed the `test_state_mps.py - AttributeError: 'list' object has no attribute 'dtype'` errors in https://github.com/NVIDIA/cuda-quantum/actions/runs/24624569814/job/72005503960#step:7:43857

The fix for the rest of the failure (`RuntimeError: invalid value`) will come in a separate PR.


sacpis (approved):
LGTM. Thanks @1tnguyen. Should we add a test for the overload path?

1tnguyen:
> LGTM. Thanks @1tnguyen. Should we add a test for the overload path?

Indeed, those tests in `test_state_mps.py` test the 2 overload paths (for a list of tensors either as ndarray or tensor objects). The other paths are tested by other `state` tests for state-vector simulators.

PR #4352, opened 2026-04-20
Fixes for state creation within kernel execution

### LINKED ISSUE #4350: [Publishing failure] Invalid value in kernel tests
### Required prerequisites

### Describe the bug

Calling sample API is erroring out on invalid values for the following tests

FAILED ../tests/builder/test_qalloc_init.py::test_kernel_complex_force_kron - RuntimeError: invalid value
FAILED ../tests/builder/test_qalloc_init.py::test_kernel_complex_params_f64 - RuntimeError: invalid value
FAILED ../tests/builder/test_qalloc_init.py::test_kernel_complex_capture_f64 - RuntimeError: invalid value
FAILED ../tests/builder/test_qalloc_init.py::test_kernel_complex_np_array_from_capture_f64 - RuntimeError: invalid value

https://github.com/NVIDIA/cuda-quantum/actions/runs/24273974731/job/70888783446#step:7:42750

Run these tests in CI.

### Expected behavior

Should pass successfully.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

Seems like the PR #4247 is the cause of these failures. Thien is looking into it.

1tnguyen:
With https://github.com/NVIDIA/cuda-quantum/pull/4247, we now create the state in the middle of circuit execution. 

**Note**: previously, the state is usually constructed outside, before kernel execution. 

Hence, the default implementation of `createStateFromData`, via `getSimulationState`, is no longer appropriate because `getSimulationState` has the semantics of the simulator giving up the ownership its current state.

This extends the fix in https://github.com/NVIDIA/cuda-quantum/pull/4304 to all simulators by removing the default implementation that is no longer valid; rather, requires subclasses to implement state creation without interfering with the current simulator state. 

Fixed https://github.com/NVIDIA/cuda-quantum/issues/4350

Publishing run: https://github.com/NVIDIA/cuda-quantum/actions/runs/24686493909/ (Python metapackage validation is passing now)


sacpis (approved):
LGTM. Thanks @1tnguyen.

schweitzpgi (approved):
Awesome. Thanks, Thien!

schweitzpgi:
The next step is to switch tests, examples, etc. to create the state outside of the kernels and pass it correctly.

There are also bugs in places that disallow entry-point kernel signatures like
```c++
__qpu__ void qernel(cudaq::state& mississippi) {
  ...
}
```

schweitzpgi:
We might want to add a regression or target test for these fixes, no?

1tnguyen:
> We might want to add a regression or target test for these fixes, no?

Good point. I added a target test for this in C++ and also extended the Python test (which triggered this bug) to cover all simulator backends.

1tnguyen:
Removed the newly-added Python test as it is not compatible with our parallel execution of pytest-xdist during Publishing (GPU resource conflicts for tensornet/tensornet-mps targets) (https://github.com/NVIDIA/cuda-quantum/pull/4352/changes/91b86ba5e4d91996a32ecd58f737f587b670807f) This will be addressed in a follow up PR.  

Confirmed Publishing Python metapackage validation passed: https://github.com/NVIDIA/cuda-quantum/actions/runs/24705966799/job/72261865535

PR #4354, opened 2026-04-21
[Dynamics] Support custom time stepper injection 

1tnguyen:
The dynamics integrator API exposes a custom stepper injection, but it did not work.

This PR fixes that:

- Make sure optional custom time stepper is used if provided. Also, this custom stepper injection won't interfere with the existing default stepper construction whenever the system dynamics changes.

- In the torch integrator: make the code more generic when handling steppers that are not the built-in one (which optimizes the data conversion). Also, move some code out of the `compute_rhs` for better performance. 

- Add test cases for custom stepper injection.


python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
```diff
@@ -121,21 +123,25 @@ def compute_rhs(self, t, vec):
...
+            result_state = bindings.initializeState(result_vec.data_ptr(), size,
+                                                    self._dimensions_list,
+                                                    self.batchSize)
+            self.stepper.compute_inplace(temp_state, t_scalar, result_state)
+        else:
+            # Stepper only provides compute(); call it and convert the returned
+            # state back to a torch tensor via `dlpack` (no extra copy).
+            result_state_obj = self.stepper.compute(temp_state, t_scalar)
+            result_cupy = to_cupy_array(result_state_obj)
+            result_vec = torch.from_dlpack(result_cupy).clone()
```
sacpis, line 144: nit: seems like comment says 'no extra copy', but here we explicitly copy.

python/cudaq/dynamics/integrators/scipy_integrators.py
```diff
@@ -31,23 +31,21 @@ class ScipyZvodeIntegrator(BaseIntegrator[cudaq_runtime.State]):
     rtol = 1e-6
     order = 12
 
-    def __init__(self, stepper: BaseTimeStepper[cudaq_runtime.State], **kwargs):
+    def __init__(self,
+                 stepper: BaseTimeStepper[cudaq_runtime.State] = None,
```
sacpis, line 35: I think we can write this as `Optional(BaseTimeStepper[cudaq_runtime.State])`.

python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
sacpis, line 74: Should we make this `None` as default or put Optional as seems like we have kept it required but the subclasses mark it as None?

python/cudaq/dynamics/integrator.py
```diff
@@ -35,7 +35,12 @@ def __init__(self, **kwargs):
         self.dimensions = None
         self.schedule = None
         self.hamiltonian = None
+        # The actual stepper used for integration. This may be set in the constructor with a user-provided stepper, or it may be auto-created by the integrator when `integrate()` is called based on the system dynamics.
```
sacpis, line 38: Strange, why yapf did not complain here?

sacpis (approved):
Overall LGTM. Thanks @1tnguyen. Left a few comments.

huaweil-nv (approved):
LGTM👍, thanks for the quick fix.

python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
1tnguyen: Good catch! `clone` is not needed here. I audited the code before we introduced the in-place optimization and confirmed that it was not needed.

Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4354/changes/d13453586d3fcda610759399955f73ac0e92ee75

python/cudaq/dynamics/integrators/scipy_integrators.py
1tnguyen: Agreed. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4354/changes/d13453586d3fcda610759399955f73ac0e92ee75.

python/cudaq/dynamics/integrators/cuda_torchdiffeq_integrator.py
1tnguyen: Good point! Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4354/changes/d13453586d3fcda610759399955f73ac0e92ee75

python/cudaq/dynamics/integrator.py
1tnguyen: I have no idea. I manually broke this long comment in https://github.com/NVIDIA/cuda-quantum/pull/4354/changes/d13453586d3fcda610759399955f73ac0e92ee75

PR #4355, opened 2026-04-21
Increasing the number of colors

sacpis:
Increasing the number of colors in the palette so VQE runs producing upto 10 flat clusters at threshold height of 1.

Fixes: https://github.com/NVIDIA/cuda-quantum/actions/runs/24686493909/job/72220359142#step:5:1914


1tnguyen (approved):
Thanks @sacpis

PR #4356, opened 2026-04-21
Slim down CompiledModule, move out execution logic

lmondada:
This is the second follow up after #4322 from our discussion last week.

This PR 'dumbs down' the `CompiledModule` type in the following sense
- it does not make any assumptions anymore on what the compilation artifacts contain (e.g. before `JitArtifact` kept track of the various named entrypoints according to some naming convention)
- it does not know what 'executing' a `CompiledModule` means

The result is a pretty clean 'container' type that just knows how to store artifacts of different kinds. The only methods that still rely on some naming conventions are `getArgsCreator` and `getReturnOffset`, which exist purely for convenience (can be removed if you wish).

This logic (which is dependent on conventions), now lives outside of the type. The naming conventions and entrypoint resolution was fully moved to construction (in `runtime/internal/compiler/include/cudaq_internal/compiler/CompiledModuleHelper.h`). The execution logic on the other hand was moved to the only placed it is currently used in (`runtime/cudaq/platform/qpu.cpp`). We might need to move this further as we unify kernel execution for C++ and Python, but wasn't sure where it would eventually land (and in what form).

### DIFF
--- a/runtime/cudaq/platform/qpu.cpp
+++ b/runtime/cudaq/platform/qpu.cpp
@@ -8,10 +8,49 @@
 
 #include "qpu.h"
 #include "mlir/IR/BuiltinOps.h"
+#include <cstring>
 
 using namespace cudaq_internal::compiler;
 
 LLVM_INSTANTIATE_REGISTRY(cudaq::ModuleLauncher::RegistryType)
+
+/// Execute a JIT-compiled kernel with provided arguments.
+///
+/// Handles argument marshaling via `argsCreator` (if not fully specialized) and
+/// result buffer allocation.
+cudaq::KernelThunkResultType
+launchCompiledModule(const cudaq::CompiledModule &compiled,
+                     const std::vector<void *> &rawArgs) {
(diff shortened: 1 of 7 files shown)


runtime/cudaq/platform/qpu.cpp
```diff
@@ -8,11 +8,50 @@
...
 using namespace cudaq_internal::compiler;
 
 LLVM_INSTANTIATE_REGISTRY(cudaq::ModuleLauncher::RegistryType)
 
+/// Execute a JIT-compiled kernel with provided arguments.
+///
+/// Handles argument marshaling via `argsCreator` (if not fully specialized) and
+/// result buffer allocation.
+cudaq::KernelThunkResultType
+launchCompiledModule(const cudaq::CompiledModule &compiled,
```
lmondada, line 22: This was just moved over from `runtime/common/CompiledModule.cpp` (previously called `CompiledModule::execute`

Renaud-K (approved):

Issue #2593, opened 2025-02-06
cannot use @cudaq.kernel in the interpreter: OSError: could not get source code


jeffhammond:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

I tried to run the demo on https://nvidia.github.io/cuda-quantum/latest/applications/python/vqe_advanced.html directly in the interpreter.  It fails with the following:
```python
>>> import cudaq
>>> @cudaq.kernel
... def kernel(qubit_num: int, electron_num: int, thetas: list[float]):
...     qubits = cudaq.qvector(qubit_num)
...
OSError: could not get source code
```

### Steps to reproduce the bug

Type this into the interpreter and press return:
```python
>>> import cudaq
>>> @cudaq.kernel
... def kernel(qubit_num: int, electron_num: int, thetas: list[float]):
...     qubits = cudaq.qvector(qubit_num)
...
OSError: could not get source code
```

### Expected behavior

What the documentation describes on the aforementioned page.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: latest.  I ran `pip install --upgrade cudaq` today.
- **Python version**: 3.10.12
- **C++ compiler**: GCC 11.4
- **Operating system**: Ubuntu 22.04.5 LTS

bebora:
# Possible solution
The problem highlighted from @jeffhammond can be solved by [using Python 3.13](https://github.com/python/cpython/issues/57129). Unfortunately, CUDA-Q does not support Python 3.13 at the moment (#2228).
# `inspect` module related problems
There are two other situations that results in the same error. In these two cases, simply updating Python is not enough. The `inspect` module might have some limitations in these scenarios.
## Interpreting from stdin
Assuming to have a valid CUDA-Q Python program, run:
`cat program.py | python3`
instead of the usual:
`python3 program.py`
The OSError will be raised.
[...]

PR #4365, opened 2026-04-22
Fix opaque OSError when @cudaq.kernel is used in the REPL (#2593)

mitchdz:
This adds better error messages for #2593.

### DIFF
--- a/python/cudaq/kernel/utils.py
+++ b/python/cudaq/kernel/utils.py
@@ -269,8 +269,8 @@ def _format_missing_source_error(function, filename):
     """
     Build a user-facing diagnostic explaining why source for `function` could
     not be retrieved. Distinguishes between three buckets:
-      - REPL-defined (`<stdin>` or `<python-input-...>`).
-      - Other synthetic filenames (eval, exec with custom name).
+      - Interactive interpreter-defined (`<stdin>` or `<python-input-...>`).
+      - Other synthetic filenames (code compiled with a non-file name).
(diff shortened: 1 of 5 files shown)


schweitzpgi (commented):
LGTM. Thanks, Mitch!

python/cudaq/kernel/utils.py
```diff
@@ -265,6 +265,63 @@ def emitWarning(msg):
...
+    `inspect.getsource`. If any fail (most commonly because `function` was
+    defined in the Python REPL), raise `RuntimeError` with a diagnostic
+    tailored to the failure mode, chained from the underlying exception.
+    """
+    filename = None
+    try:
+        filename = inspect.getfile(function)
+        first_line = inspect.getsourcelines(function)[1]
+        src = inspect.getsource(function)
+    except (OSError, TypeError) as e:
```
anpaz: why `TypeError `here? is it the same root cause (that we can't read the source file) or would the error message we know return be missleading?

anpaz (approved):
There is a minor comment about the error might be miss leading as we're reporting the same message for Type exceptions.

python/cudaq/kernel/utils.py
mitchdz: Great catch. In practice, TypeError shouldn't happen in this scenario, so I'll remove it since it is a bit misleading.

1tnguyen (approved):
LGTM 👍 Thanks, @mitchdz!

PR #4367, opened 2026-04-22
[NFC] Breaking up Compiler.cpp to isolate common code from policy specific code.

Renaud-K:
The goal of this PR is to prepare `lowerQuakeCode` and `runPassPipeline` to be overloaded based on the policies introduced in #4339. Practically speaking, the code located between `ctx->name == "..."` has been moved into utility methods which do not depend on the `executionContext`.

5 private helper methods have been extracted.
1. The lambda `runPassPipeline` has been renamed to `applyPipeline` as to not be confused with the new `runPassPipeline` method and is now a method as it is being called from several other methods
2. `prepareModule` builds the module via lowerQuakeCodeBuildModule, merges closures, performs argument/quake synthesis.

3. `executeMainPipeline` handles the combine-measurements delay for emulation, then runs the main pass pipeline. Returns whether combine-measurements was delayed.

4. `extractMappingReorderIdx` extracts qubit-mapping reorder indices from entry-point attributes. Moved into the anonymous namespace.

5. `assembleCompiledModule` creates JIT artifacts (when needed), runs deferred combine-measurements, creates MLIR artifacts, and assembles the final CompiledModule.

### DIFF
--- a/runtime/internal/compiler/Compiler.cpp
+++ b/runtime/internal/compiler/Compiler.cpp
@@ -90,6 +90,22 @@ nlohmann::json formOutputNames(const std::string &codegenTranslation,
   }
   return output_names;
 }
+/// Extract qubit-mapping reorder indices from the entry-point attributes.
+std::vector<std::size_t> extractMappingReorderIdx(mlir::ModuleOp moduleOp,
+                                                  mlir::func::FuncOp epFunc) {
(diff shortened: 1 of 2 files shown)


Renaud-K:
> /ok to test a61e715b63ad67ba60b22e5bf28c648179c2ed13

lmondada (approved):
Thanks, this is actually complementary to what I had in mind, works very well.

PR #4369, opened 2026-04-22
[Publishing fix] Using xdist_group

sacpis:
The remote platform test module sets pytest.mark.xdist_group to keep its tests on one worker, but pytest-xdist only honors that marker under --dist=loadgroup.

Fixes: https://github.com/NVIDIA/cuda-quantum/actions/runs/24750269888/job/72426003995#step:7:2816

### DIFF
--- a/scripts/validate_pycudaq.sh
+++ b/scripts/validate_pycudaq.sh
@@ -333,9 +333,14 @@ else
     done
 fi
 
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
LGTM 👍

mitchdz (approved):

PR #4372, opened 2026-04-23
Fixes and tests for nested control and adjoint modifiers

### LINKED ISSUE #854: Allow applying a controlled, adjoint kernel
### Required prerequisites

- [X] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to check if your feature has already been mentioned or rejected in other issues.

### Describe the feature

Currently, it is possible to apply an adjoint variant of a kernel using `cudaq::adjoint` and a controlled variant of a kernel with `cudaq::control`, but there isn't a way to apply them both.
It seems like some infrastructure work was already performed to allow this (e.g. the `apply-op-specialization` pass can handle `quake.apply` operators which have both adjoint set and control registers) but I couldn't find a way to create such usages of `quake.apply` with C++.

1tnguyen:
We have mostly supported nested control/adjoint as requested in #854.

- `ctrl(adj(k))` works

- `adj(ctrl(k))` does not

This is because `createAdjointVariantOf` doesn't handle `ApplyOp`, while `createControlVariantOf` does.

Hence, update it so that `createAdjointVariantOf` can handle `ApplyOp` as well (toggle the adjoint flag).

Added quake and execution tests.

Resolved #854


atgeller (approved):
LGTM, thanks @1tnguyen!

PR #4375, opened 2026-04-23
Adding coverage for if condition in a list

### LINKED ISSUE #4207: Python @cuda.kernel list creation issues
### Required prerequisites

### Describe the bug

Multiple issues noticed related to list comprehension while attempting to convert a bitmask into a list of qubits:

1. Within cuda kernel, `if` in list comprehensions is ignored.
2. When the length of the list is calculated using a loop, using it in list comprehension causes kernel output to fail entirely.

Context:

> We have a bitmask like `0b10011` and we desire to turn that into a list of qubits, to be used as a quantum condition like `[q[0], q[1], q[4]]` (notice bits 0, 1, and 4 are set in the example mask `0b10011`)
> 
> While trying to accomplish this using loops and list comprehension, encountered multiple issues. The `list.append()` method for lists is not supported, and the obvious list comprehension approach...
```python
    control_list = [q[x] for x in range(control_mask_bit_length) if ((1 << x) & control_mask)]
```
> ...fails for multiple reasons, as described

These tests illustrate list comprehension problems encountered:

```python
import cudaq
cudaq.set_target('nvidia', option = 'fp64')

@cudaq.kernel
def test1_kernel() -> list[int]:
    # Test "if" in list comprehension
    return [x for x in range(16) if False]

@cudaq.kernel
def test2a_kernel() -> list[int]:
    # Test scanning bit length and using result in list comprehension
    mask = 0b10011
    bit_length = 0
    while (1 << bit_length) <= mask:
        bit_length += 1
    return [x for x in range(bit_length)]

@cudaq.kernel
def test2b_kernel() -> list[int]:
    # Demo that scanning list length works
    mask = 0b10011
    bit_length = 0
    while (1 << bit_length) <= mask:
        bit_length += 1
    return [bit_length]

@cudaq.kernel
def test2c_kernel() -> list[int]:
    # Demo that list length works if we don't scan
    bit_length = 5
    return [x for x in range(bit_length)]

def test_kernels():
    print('test1:', cudaq.run(test1_kernel, shots_count=1))
    print('    expected: [[]] \n')
    print('test2a:', cudaq.run(test2a_kernel, shots_count=1))
    print('    expected: [[0, 1, 2, 3, 4]] \n')
    print('test2b:', cudaq.run(test2b_kernel, shots_count=1))
    print('    expected: [[5]] <--- this works \n')
    print('test2c:', cudaq.run(test2c_kernel, shots_count=1))
    print('    expected: [[0, 1, 2, 3, 4]] <--- this works \n')

test_kernels()
```

### Expected behavior

The above tests each indicate the expected output.

Also, any solution to turn a mask into a list would solve the problem.

Not a regression

### Environment

- **CUDA-Q version**: cuda-quantum-cu13-nightly==0.13.10, cupy-cuda13x==13.6.0, nvidia-cuda-runtime==13.2.51
- **Python version**: 
- **C++ compiler**: N/A (but it's g++ 13.3.0)

### Suggestions

_No response_

sacpis:
- Lower if`-filtered list comprehensions (handles chained if) in Python kernel AST bridge
- Works for int/bool element types and qubit-ref results
- Adds `test_list_comprehension_filter` covering constant, variable, captured-list, chained-if, and control-qubit cases

Fixes #4207

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -3797,12 +3797,16 @@ class PyASTBridge(ast.NodeVisitor):
         the MLIR.
 
         By simple, we mean expressions like `[expr(iter) for iter in iterable]`
-        or `myList = [exprThatReturns(iter) for iter in iterable]`.
+        or `myList = [exprThatReturns(iter) for iter in iterable]`, optionally
+        with `if` filter clause.
         """
         if len(node.generators) > 1:
             self.emitFatalError(
                 "CUDA-Q only supports single generators for list comprehension.",
                 node)
+
+        if_clauses = node.generators[0].ifs
+        hasFilter = len(if_clauses) > 0
 
         self.visit(node.generators[0].iter)
         iterable = self.popValue()
@@ -3841,6 +3845,15 @@ class PyASTBridge(ast.NodeVisitor):
             # This loop could be marked as invariant if we didn't use
             # `visit_For`, but that would be premature optimization.
             self.visit_For(forNode)
+
+        def evalFilter():
+            cond = None
+            for if_node in if_clauses:
+                self.visit(if_node)
+                this_cond = self.__arithmetic_to_bool(self.popValue())
+                cond = this_cond if cond is None else arith.AndIOp(
+                    cond, this_cond).result
+            return cond
 
         target_types = {}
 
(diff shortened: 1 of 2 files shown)


1tnguyen (approved):
LGTM 👍

PR #4376, opened 2026-04-23
Fix operator examples in docs

### LINKED ISSUE #3341: Documentation Errors
https://nvidia.github.io/cuda-quantum/latest/using/examples/operators.html#constructing-spin-operators
Example is missing.

https://nvidia.github.io/cuda-quantum/latest/using/examples/executing_kernels.html#get-state
This ordering is referred to as MBS or big endian. This seems incorrect, or at the very least ill-defined.
As far as I can tell, the state vectors are indexed in the way that is given, [00, 10, 01, 11], but this is traditionally little-endian.
The confusion may come because qiskit is also little endian, but uses a different qubit ordering in their register? 
May also be related to how the tensor expansion of operators has to be somewhat reversed in your qubit order.
[ i.e. spin.z(0) is IIZ, spin.z(1) is IZI, spin.z(2) is ZII... ]
https://nvidia.github.io/cuda-quantum/latest/api/languages/python_api.html#cudaq.operators.spin.SpinOperator.degrees
This part of the API seems to contradict the above definition of state vector indexing as well.

Perhaps a section of documentation covering how CUDA-Qs chosen qubit ordering conventions relates to the tensor products which build the state vectors and operators would be helpful?

https://nvidia.github.io/cuda-quantum/latest/using/examples/operators.html#pauli-words-and-exponentiating-pauli-words

I believe this should be the product of exponentials, not the sum.
Try the code snippet with different pauli words, and see that the order the pauli words are in can change the answer.
`words = ['ZYX', 'ZXX']
coefficients = [0.432, 0.324]

@cudaq.kernel
def kernel(coefficients: list[float], words: list[cudaq.pauli_word]):
    q = cudaq.qvector(3)

    for i in range(len(coefficients)):
        exp_pauli(coefficients[i], q, words[i])

cudaq.get_state(kernel, coefficients, words).dump()
cudaq.get_state(kernel, coefficients[::-1], words[::-1]).dump()` 

returns two different states as an example.

mitchdz:
Closes #3341.

  - Fill in the missing spin-operator construction example in `operators.{py,cpp}` (the `[Begin Spin]` block was empty).
  - Fix the `get_state` endianness description: text said "Big-endian" while the table showed LSB-on-left. Rewrite with an explicit qubit-to-bit mapping and a note covering `SampleResult`, `get_pauli_word()`, and the
    binary form of the statevector index.
  - `exp_pauli` docs called the result a sum of exponentials; it's a product. Also swap the demo words from `XYZ`/`IXX` (commute, hides the point) to `ZYX`/`ZXX`.
  - Default `shots_count` for `observe` is `-1`, not `1`.
  - Rewrite `SpinOperator.degrees` docstring for consistency; fix `max_degree` docstring (said "smallest").


docs/sphinx/examples/cpp/operators.cpp
```diff
@@ -8,13 +8,21 @@
 
 // [Begin Spin]
 #include <cudaq.h>
+#include <iostream>
+
+void build_operator() {
```
sacpis, line 13: Seems like we are missing to call `build_operator`?

docs/sphinx/using/examples/executing_kernels.rst
```diff
@@ -218,19 +218,35 @@ The `get_state` function gives us access to the quantum statevector of the compu
         :start-after: [Begin `GetStateOutput`]
         :end-before: [End `GetStateOutput`]
 
-The statevector generated by the `get_state` command follows Big-endian convention for associating numbers with their binary representations, which places the least significant bit on the left. That is, for the example of a 2-bit system, we have the following translation between integers and bits:
+The statevector returned by `get_state` indexes basis states by interpreting each
+qubit's amplitude bit with qubit 0 as the least significant bit, qubit 1 as the
```
sacpis: nit: Can we please show an example as where the LSB and MSB is?

sacpis (approved):
LGTM. Thanks @mitchdz. Left few comments.

docs/sphinx/using/examples/executing_kernels.rst
mitchdz: Added in [ee49d1f](https://github.com/NVIDIA/cuda-quantum/pull/4376/commits/ee49d1fe867b04e85013c102532ef63ecf4bd23c)

docs/sphinx/examples/cpp/operators.cpp
mitchdz: Added in [ee49d1f](https://github.com/NVIDIA/cuda-quantum/pull/4376/commits/ee49d1fe867b04e85013c102532ef63ecf4bd23c)

1tnguyen (approved):
LGTM 👍

PR #4378, opened 2026-04-23
Merge lowerQuakeCode overloads

lmondada:
This unifies the two overloads of `Compiler::lowerQuakeCode` into one by moving the responsibility of loading the kernel into an MLIR ModuleOp to the caller.

This also means that `extractQuakeCodeAndContext` is now a public method of `Compiler`. I've taken this opportunity to rename it to `loadQuakeCodeByName` which I found clearer. Happy to revert that.


Renaud-K (approved):
Looks good. Thank you.

PR #4384, opened 2026-04-23
Skipping full Mottonen decomposition

### LINKED ISSUE #765: state preparation in from_state() function is inefficient
When the from_state() function is used to initialize a state, it adds a bunch of rotations gates and x gates to it.  This preparation is really inefficient and slows down circuit execution quite significantly. 

Here is the code that uses the from_state() function to prepare an all zero state for a 16-qubit circuit.  

```
#include "cudaq/algorithms/state.h"
#include "cudaq/builder/kernels.h"
#include <cudaq/algorithm.h>
#include <cudaq.h>

int main()
{
    const std::size_t numQubits = 16;
    
    std::vector<std::complex<double>> state(1ULL << numQubits, 0.0);
    // state[0] = 1.0; # set one of the component to 1 
    auto kernel = cudaq::make_kernel();
    auto qubits = kernel.qalloc(numQubits);

    cudaq::from_state(kernel, qubits, state);
    cudaq::sample(kernel);
}
```

sacpis:
Before this change, a 16-qubit basis state input emitted `2^n controlled rotations` and took `29 seconds` to build and sample. Here we are adding a path that detects input with one non-zero amplitude and emit `X` gates on the set bits of that index.

Result: This has sped up the execution time by **289x** (For 16 qubits, 29 s -> 100 ms).

Fixes #765

### DIFF
--- a/runtime/cudaq/builder/kernels.h
+++ b/runtime/cudaq/builder/kernels.h
@@ -80,13 +80,17 @@ void from_state(Kernel &&kernel, QuakeValue &qubits,
         break;
     }
   }
-  if (nonZeroCount <= 1) {
+  if (nonZeroCount == 0)
+    throw std::invalid_argument(
+        "[from_state] input state vector is all zeros; a quantum state "
+        "must have unit norm.");
+  if (nonZeroCount == 1) {
     // Möttönen ordering: state-vector index MSB maps to qubits[0], LSB to
-    // qubits[numQubits-1]
-    if (nonZeroCount == 1)
-      for (std::make_signed_t<std::size_t> q = 0; q < numQubits; ++q)
-        if ((nonZeroIdx >> (numQubits - 1 - q)) & 1)
-          kernel.x(qubits[q]);
+    // qubits[numQubits-1].
+    auto nq = static_cast<std::size_t>(numQubits);
+    for (std::size_t q = 0; q < nq; ++q)
+      if ((nonZeroIdx >> (nq - 1 - q)) & 1)
+        kernel.x(qubits[q]);
     return;
   }
 
--- a/runtime/cudaq/builder/kernels.h
+++ b/runtime/cudaq/builder/kernels.h
@@ -69,6 +69,27 @@ void from_state(Kernel &&kernel, QuakeValue &qubits,
         "[from_state] cannot infer size of input quantum register, please "
         "specify the number of qubits via the from_state() final argument.");
 
+  constexpr double basisTol = 1e-12;
+  std::size_t nonZeroCount = 0;
+  std::size_t nonZeroIdx = 0;
+  for (std::size_t i = 0; i < data.size(); ++i) {
+    if (std::abs(data[i]) > basisTol) {
+      ++nonZeroCount;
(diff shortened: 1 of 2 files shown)


runtime/cudaq/builder/kernels.h
```diff
@@ -69,6 +69,27 @@ void from_state(Kernel &&kernel, QuakeValue &qubits,
...
+  std::size_t nonZeroIdx = 0;
+  for (std::size_t i = 0; i < data.size(); ++i) {
+    if (std::abs(data[i]) > basisTol) {
+      ++nonZeroCount;
+      nonZeroIdx = i;
+      if (nonZeroCount > 1)
+        break;
+    }
+  }
+  if (nonZeroCount <= 1) {
```
1tnguyen: nit: should we throw if `nonZeroCount` is zero?

runtime/cudaq/builder/kernels.h
```diff
@@ -69,6 +69,27 @@ void from_state(Kernel &&kernel, QuakeValue &qubits,
...
+    // Möttönen ordering: state-vector index MSB maps to qubits[0], LSB to
+    // qubits[numQubits-1]
+    if (nonZeroCount == 1)
+      for (std::make_signed_t<std::size_t> q = 0; q < numQubits; ++q)
```
1tnguyen: nit: could we write `std::size_t q = 0; q < ...`?

unittests/integration/kernels_tester.cpp
```diff
@@ -181,6 +181,49 @@ CUDAQ_TEST(KernelsTester, checkFromState) {
...
+
+  verifyBasis(4, 0);
+  verifyBasis(4, 5);
+  verifyBasis(4, 15);
+
+  {
+    std::vector<std::complex<double>> zero(8, 0.0);
+    auto kernel = cudaq::make_kernel();
+    auto qubits = kernel.qalloc(3);
+    cudaq::from_state(kernel, qubits, zero);
```
1tnguyen: I think this test is actually invalid and we should throw as an all zero vector is not a valid state vector.

runtime/cudaq/builder/kernels.h
sacpis: Fixed in [2f1e16e](https://github.com/NVIDIA/cuda-quantum/pull/4384/commits/2f1e16e11399f7280e1160c5e8ac598cbe083661).

runtime/cudaq/builder/kernels.h
sacpis: Fixed in [2f1e16e](https://github.com/NVIDIA/cuda-quantum/pull/4384/commits/2f1e16e11399f7280e1160c5e8ac598cbe083661).

unittests/integration/kernels_tester.cpp
sacpis: Fixed in [2f1e16e](https://github.com/NVIDIA/cuda-quantum/pull/4384/commits/2f1e16e11399f7280e1160c5e8ac598cbe083661).

1tnguyen (approved):
LGTM 👍

PR #4385, opened 2026-04-23
Replacing llvm type registry with our own

Renaud-K:
While exploring and prototyping options for compile-time checks of QPUs, I keep running into the issue that llvm headers bleed into user code. I would like to put an end to this issue once and for all by replacing it with our own registry.

This registry is widely borrowed from LLVM by Claude and reviewed by codex. I have asked it to keep the current model, whereby it is instantiated in one place so that a registration from a shared lib is visible to all shared libs.

I believe, from discussions with Bruno, that this should also help the LLVM update work.

### DIFF
--- a/lib/Optimizer/Transforms/DecompositionPatterns.cpp
+++ b/lib/Optimizer/Transforms/DecompositionPatterns.cpp
@@ -39,7 +39,7 @@
 
 using namespace mlir;
 
-CUDAQ_INSTANTIATE_REGISTRY(cudaq::DecompositionPatternType::RegistryType)
+LLVM_INSTANTIATE_REGISTRY(cudaq::DecompositionPatternTypeRegistry)
 
 namespace {
 
@@ -309,6 +309,8 @@ LogicalResult checkAndExtractControls(quake::OperatorInterface op,
 }
 
 // From here on, we define the decomposition patterns ==========================
+#define CONCAT(a, b) CONCAT_INNER(a, b)
+#define CONCAT_INNER(a, b) a##b
 
 /// Macro to register a decomposition pattern with its metadata
 /// Usage: REGISTER_DECOMPOSITION_PATTERN(PatternName, "source_op", "target1",
@@ -332,7 +334,8 @@ LogicalResult checkAndExtractControls(quake::OperatorInterface op,
       return pattern;                                                          \
     }                                                                          \
   };                                                                           \
-  CUDAQ_REGISTER_TYPE(cudaq::DecompositionPatternType, PATTERN##Type, PATTERN)
+  static cudaq::DecompositionPatternTypeRegistry::Add<PATTERN##Type> CONCAT(   \
+      TEMPNAME_, PATTERN)(#PATTERN, "");
 
 // NOTE: The patterns SToR1, TToR1, R1ToU3, and U3ToRotations handle arbitrary
 // control counts and are registered with (n) metadata. R1ToRz explicitly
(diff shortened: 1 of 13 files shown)


Renaud-K:
> /ok to test 4916cae2429044625de8396ec6db0606cf9e8beb

lib/Optimizer/Transforms/DecompositionPatterns.cpp
```diff
@@ -39,7 +39,7 @@
 
 using namespace mlir;
 
-LLVM_INSTANTIATE_REGISTRY(cudaq::DecompositionPatternType::RegistryType)
+CUDAQ_INSTANTIATE_REGISTRY(cudaq::DecompositionPatternType::RegistryType)
```
schweitzpgi: This introduces a dependence from the core compiler to the runtime which is completely backwards.

schweitzpgi (changes requested):

runtime/common/Registry.h
```diff
@@ -7,41 +7,154 @@
...
 
 #pragma once
+
+// Suppress the LLVM ABI breaking check for translation units that include this
+// header. Several server helper DSOs transitively include LLVM headers (e.g.
+// llvm/Support/Base64.h) but do not link against LLVM libraries, so the ABI
+// check symbol would be unresolved. This define must be set before any LLVM
+// header is included. TODO: remove once those DSOs stop including LLVM headers.
 #define LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING 1
-#include "llvm/Support/Registry.h"
```
schweitzpgi, line 11: Why do we think the runtime cannot use LLVM? The dependence chain looks like
```
runtime depends on
  core depends on
    MLIR depends on
      LLVM depends on
        ...
```

lib/Optimizer/Transforms/DecompositionPatterns.cpp
boschmitt: Agree. The compiler should rely on LLVM's registry. I also think the runtime should have its own registry implementation. No need for them to be the same. (More about this in the next comment.)

runtime/common/Registry.h
boschmitt: It is not that the runtime cannot use LLVM. However, we do need to minimize the surface of this dependency. In this case, I think the change is good, the runtime and the compiler should have different registries. (So I agree with your other comment about not using the runtime registry in the compiler.)

For context, the issue I discussed with @Renaud-K is that LLVM 22 changed `llvm::Registry<T>::Head/Tail` from `static` to `static inline`, causing each DSO to get its own copy of the registry head pointer. Our runtime is split into various DSOs, so when Python loads extensions, plugins registered in one DSO become invisible to code in another, breaking `QPU`, `ServerHelper`, and `RemoteRuntimeClient` lookups at runtime. (We basically end up if various registries instead of one.)

Admittedly, this is an issue when compiling the project with `clang/llvm`, GCC shouldn't have the same issue because it uses `STB_GNU_UNIQUE` instead of weak symbols.

Renaud-K: Thanks @boschmitt. We also have the architectural requirement that `cuda.h`, or any user visible header, does not transitively include an llvm/mlir include or any other third party library such as fmtlib for instance.

lib/Optimizer/Transforms/DecompositionPatterns.cpp
Renaud-K: I will change this. I did not realize that this was in the compiler.

Renaud-K: Wait a minute. The runtime dependency has been there [all along](https://github.com/Renaud-K/cuda-quantum/blob/188d1c1d98d10cfb4d6d83fc6c6bf4a046aaa1b1/lib/Optimizer/Transforms/DecompositionPatterns.h#L11).

I will see what it will take to remove it.

Renaud-K: It could be done with using the LLVM API.

Renaud-K:
> /ok to test 3b858e0b189d3a43a996d9f4062a8204b54a5321

runtime/common/Registry.h
schweitzpgi: I think it is exactly right to not slurp in "bonus" include headers.

Just wanted to make sure we didn't think there was no dependence whatsoever, which would be wrong. We just want to hide those details behind functional interfaces.

schweitzpgi (approved):

PR #4388, opened 2026-04-24
Introduce QPU::compileModule, use it for launch + specialize

lmondada:
This removes `specializeModule` within `QPU` and `quantum_platform`. Instead, both when specializing and launching a kernel from Python, `QPU::compileModule` is called.

### DIFF
--- a/runtime/cudaq/platform/nvqpp_interface.h
+++ b/runtime/cudaq/platform/nvqpp_interface.h
@@ -43,6 +43,7 @@ streamlinedLaunchKernel(const char *kernelName,
 hybridLaunchKernel(const char *kernelName, KernelThunkType kernel, void *args,
                    std::uint64_t argsSize, std::uint64_t resultOffset,
                    const std::vector<void *> &rawArgs);
+} // extern "C"
 
 //===----------------------------------------------------------------------===//
 // Launch module entry points.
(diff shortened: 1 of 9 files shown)


runtime/cudaq/platform/nvqpp_interface.h
```diff
@@ -53,23 +54,13 @@ hybridLaunchKernel(const char *kernelName, KernelThunkType kernel, void *args,
 // directly.
 //===----------------------------------------------------------------------===//
 
-// Streamlined interface for launching kernels. Argument synthesis and JIT
-// compilation *must* happen on the local machine. The caller must provide an
-// mlir::ModuleOp and the short name of the entry point kernel function to be
-// called,
-[[nodiscard]] KernelThunkResultType
-streamlinedLaunchModule(const char *kernelName, mlir::ModuleOp moduleOp,
```
lmondada, line 61: I've removed this C-API function -- it's not used for Python and I don't see how anyone other than Python would pass us live `moduleOp` objects. Is this wrong?

atgeller: I would double check that it isn't being called from generate code somewhere. I don't believe it is used on the Python side.

atgeller (approved):
LGTM, thanks @lmondada!

PR #4389, opened 2026-04-24
Add runtime tracer with spdlog and Chrome backends  

taalexander:
## Summary                                                   

- Introduces a process-wide `Tracer` that dispatches span begin/end events to
pluggable backends.
- `SpdlogTraceBackend` preserves identical output with the existing
`ScopedTrace` implementation.                               
- `ChromeTraceBackend` captures events in memory and serializes as Chrome
Trace Event Format JSON (viewable in Perfetto / speedscope).             
- `TracePassInstrumentation` attached at every in-repo MLIR `PassManager`     
construction sites.
- Env-var flow: `CUDAQ_TRACE_FORMAT=chrome|spdlog` / `CUDAQ_TRACE_PATH=<path>`   
enables tracing at `initializeLogger` time with no code changes.              

## Design notes                                                               
                                                           
- `TraceBackend` inherits `std::enable_shared_from_this` so a follow-up       
Python-bindings PR can hold backends as first-class objects with independent
C++ / Python shared ownership.                                                  
- Fork safety: `ChromeTraceBackend` records `ownerPid` at construction and
skips the destructor file write in forked children, avoiding parent-output    
clobber.                                                     
                                                                                                                                      
## Dependencies                                                               
                                                           
Based on `feature/gil-release-compilation`. Precedes PR #4284              
(`feature/tracer-python`) which stacks on this.

### DIFF
--- a/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
+++ b/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
@@ -22,6 +22,7 @@
 #include "cudaq/platform/qpu.h"
 #include "cudaq_internal/compiler/ArgumentConversion.h"
 #include "cudaq_internal/compiler/LayoutInfo.h"
+#include "cudaq_internal/compiler/TracePassInstrumentation.h"
 #include "runtime/cudaq/algorithms/py_utils.h"
 #include "runtime/cudaq/platform/PythonSignalCheck.h"
 #include "utils/LinkedLibraryHolder.h"
@@ -1016,6 +1017,7 @@ static MlirModule synthesizeKernel(nanobind::object kernel,
 
   PassManager pm(context);
   cudaq::addPythonSignalInstrumentation(pm);
+  pm.addInstrumentation(std::make_unique<cudaq::TracePassInstrumentation>());
   pm.addPass(cudaq::opt::createArgumentSynthesisPass(
       kernelRefs, substRefs, /*changeSemantics=*/false));
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
@@ -1107,6 +1109,7 @@ static ModuleOp cleanLowerToCodegenKernel(ModuleOp mod,
     auto *ctx = mod.getContext();
     PassManager pm(ctx);
     cudaq::addPythonSignalInstrumentation(pm);
+    pm.addInstrumentation(std::make_unique<cudaq::TracePassInstrumentation>());
     std::string transport = getTransportLayer();
     cudaq::opt::addAOTPipelineConvertToQIR(pm, transport);
     executeMLIRPassManager(mod, pm);
@@ -1256,6 +1259,8 @@ void cudaq::bindAltLaunchKernel(nanobind::module_ &mod,
         auto context = m.getContext();
         PassManager pm(context);
         cudaq::addPythonSignalInstrumentation(pm);
+        pm.addInstrumentation(
+            std::make_unique<cudaq::TracePassInstrumentation>());
         pm.addNestedPass<func::FuncOp>(
             cudaq::opt::createPySynthCallableBlockArgs(
                 SmallVector<StringRef>(funcNames.begin(), funcNames.end()),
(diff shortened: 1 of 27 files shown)


python/runtime/common/py_NoiseModel.cpp
```diff
@@ -132,6 +132,7 @@ void bindNoiseModel(nanobind::module_ &mod) {
             auto key = nanobind::hash(krausT);
             std::function<kraus_channel(const std::vector<double> &)> lambda =
                 [krausT](const std::vector<double> &p) -> kraus_channel {
+              nanobind::gil_scoped_acquire acquire;
```
Renaud-K, line 137: How do we know when these locks are required?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -920,7 +922,14 @@ nanobind::object cudaq::marshal_and_launch_module(const std::string &name,
   auto mod = unwrap(module);
   Type retTy = cudaq::runtime::getReturnType(kernelFunc);
   auto args = marshal_arguments_for_module_launch(mod, runtimeArgs, kernelFunc);
-  [[maybe_unused]] auto resultPtr = clean_launch_module(name, mod, args);
+
+  {
+    // Release the GIL for MLIR compilation + JIT. This allows Python signal
+    // handling (Ctrl-C) via addPythonSignalInstrumentation which uses
+    // PyGILState_Ensure from the compilation thread.
+    nanobind::gil_scoped_release release;
```
Renaud-K: Could this not be done internally?

runtime/internal/compiler/RuntimeMLIR.cpp
```diff
@@ -644,6 +648,8 @@ static void registerToIQMJsonTranslation() {
          bool printIntermediateMLIR, bool printStats) {
         ScopedTraceWithContext(cudaq::TIMING_JIT, "iqm translation");
         PassManager pm(op->getContext());
+        pm.addInstrumentation(
```
Renaud-K: This could be run from python as well. Do we not need `runPassManagerReleasingGIL` here? Note this also runs with no python dependencies. We would have a rather clean way to deal with it if we needed to.

python/runtime/common/py_NoiseModel.cpp
taalexander: See this PR for where the GIL comes from #4284. This PR is based on that chain of work. In general the lock is required when CPP calls out to a Python object and requires the gil (in this case the lambda is later captured and invoked from the simulation which requires recapture).

taalexander: I added comments in 0159e4d342

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
taalexander: Good point. I moved the `gil_scoped_release` into `clean_launch_module` as its a NOOP if not held in 0159e4d342

runtime/internal/compiler/RuntimeMLIR.cpp
taalexander: I did a search and updated all accessors that could trigger lazy execution in 0159e4d342 to make sure they now all release the gil.

Renaud-K: What I meant, is that we have `RuntimeCppMLIR.cpp` and `RuntimePyMLIR.cpp` both with the same interface. Except a Python flow only links in the latter while a C++ flow only links in the former. This can be used as a way to perform different actions in Python and C++ while calling the same code.

taalexander: Thanks for clarifying, I did not understand initially. Looking into it.

taalexander: This was a great suggestion and much cleaner. I've implemented in 3f6fb44df6

Renaud-K (approved):

PR #4391, opened 2026-04-24
[PTSBE] fully qualify names and drop `using namespace` directives

taalexander:
Address comments from @schweitzpgi that pointed out the style did not match the existing codebase.

### DIFF
--- a/unittests/ptsbe/ExecutePTSBETester.cpp
+++ b/unittests/ptsbe/ExecutePTSBETester.cpp
@@ -15,23 +15,21 @@
 #include <numeric>
 
 using namespace cudaq;
-using namespace cudaq::ptsbe;
-using namespace cudaq::ptsbe::detail;
 
 // Use QPP simulator for testing samplePTSBE
 using QppSimulator =
     QppCircuitSimulatorTester<nvqir::QppCircuitSimulator<qpp::ket>>;
 
-const PTSBETrace kHadamardTrace = {
+const ptsbe::PTSBETrace kHadamardTrace = {
     {ptsbe::TraceInstructionType::Gate, "h", {0}, {}, {}}};
-const PTSBETrace kXTrace = {
+const ptsbe::PTSBETrace kXTrace = {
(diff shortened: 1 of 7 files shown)


python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
```diff
@@ -30,52 +30,52 @@
 #include <nanobind/stl/unordered_map.h>
 #include <nanobind/stl/vector.h>
 
-using namespace cudaq;
```
schweitzpgi, line 33: FWIW, having one wide open namespace is fine. We just do not want more than one.

unittests/ptsbe/ExecutePTSBETester.cpp
```diff
@@ -14,54 +14,50 @@
 #include <cmath>
 #include <numeric>
 
-using namespace cudaq;
-using namespace cudaq::ptsbe;
-using namespace cudaq::ptsbe::detail;
```
schweitzpgi, line 19: Like here. You could leave
```
using namespace cudaq;
```
But we don't want the other 2.

schweitzpgi (approved):
I think this can be a much smaller PR and just have focused on the files with more than 1
```c++
using namespace foo;
using namespace bar;
....
````
open namespace in the .cpp files.

For .h files, the number of open namespaces should be zero.

python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
taalexander: Thanks addressed in [79485df](https://github.com/NVIDIA/cuda-quantum/commit/79485df8e045064fcb2e1012ff0751a1b4f1dac5)

unittests/ptsbe/ExecutePTSBETester.cpp
taalexander: Thanks addressed in [79485df](https://github.com/NVIDIA/cuda-quantum/commit/79485df8e045064fcb2e1012ff0751a1b4f1dac5)

PR #4392, opened 2026-04-24
[NFC] Split out CompiledModule container into class

lmondada:
On top of #4388.

This PR splits out the container that is used in `CompiledModule` into its own type. This is so that it can be re-used by other upcoming types that look very similar, e.g. `KernelArgs`.

I took the opportunity to change to using a vector of pairs instead of a `std::map` to store the artifacts. This should be faster (most of the time, there will be <5 artifacts) and means that several artifacts of different types can share the same name. This removes the need to adopt some naming convention to differentiate multiple artifact types for the same kernel, as they can share the same name.

### DIFF
--- a/runtime/common/BaseRemoteSimulatorQPU.h
+++ b/runtime/common/BaseRemoteSimulatorQPU.h
@@ -204,7 +204,7 @@ public:
               name, jit, {}, true);
       auto mlirArtifact =
           cudaq_internal::compiler::CompiledModuleHelper::createMlirArtifact(
-              name + ".mlir", moduleOp);
+              name, moduleOp);
       artifacts.push_back(mlirArtifact);
       return cudaq_internal::compiler::CompiledModuleHelper::
           createCompiledModule(name, resultInfo, std::move(artifacts));
@@ -212,7 +212,7 @@ public:
 
     auto mlirArtifact =
         cudaq_internal::compiler::CompiledModuleHelper::createMlirArtifact(
-            name + ".mlir", prefabMod);
+            name, prefabMod);
 
     return cudaq_internal::compiler::CompiledModuleHelper::createCompiledModule(
         name, resultInfo, {mlirArtifact});
--- a/runtime/common/NamedVariantStore.h
+++ b/runtime/common/NamedVariantStore.h
@@ -1,0 +1,79 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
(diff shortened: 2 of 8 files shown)


runtime/common/BaseRemoteSimulatorQPU.h
```diff
@@ -204,15 +204,15 @@ class BaseRemoteSimulatorQPU : public QPU {
               name, jit, {}, true);
       auto mlirArtifact =
           cudaq_internal::compiler::CompiledModuleHelper::createMlirArtifact(
-              name + ".mlir", moduleOp);
+              name, moduleOp);
```
schweitzpgi, line 207: Where did this ".mlir" suffix come from? What is `name` here?

runtime/common/NamedVariantStore.h
```diff
@@ -0,0 +1,79 @@
...
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <variant>
+#include <vector>
+
+namespace cudaq::detail {
+
+/// A container of named heterogeneous name-value pairs, retrievable by name and
```
schweitzpgi, line 21: Is there any documentation on what you are doing?

runtime/common/BaseRemoteSimulatorQPU.h
Renaud-K: FWIW, This file will be gone in #4276

runtime/common/NamedVariantStore.h
lmondada: Yes -> https://gitlab-master.nvidia.com/cuda-quantum/cudaq-spec/-/merge_requests/62

I had rewritten the entire document after your comments. They should have been addressed, but we can discuss tomorrow.

runtime/common/BaseRemoteSimulatorQPU.h
lmondada: `name` is the kernel name that is passed in as argument to `launchModule`. The string identifiers are there to distinguish multiple kernel entrypoints (currently, there can be the main entrypoint, `argsCreator` and `resultOffset`). This string identifier is useless when there is a unique entrypoint.

I had temporarily introduced the suffix `mlir` to distinguish MLIR kernels from JITed kernels, but this was clunky. This PR fixes this.

The choice of strings as identifiers was discussed [here](https://github.com/NVIDIA/cuda-quantum/pull/4280#issuecomment-4215582204). This is to model that 

> a kernel isn't one function, it's a collection of functions.

I understand you don't like them. The only alternative I have been able to think of is storing entrypoints in an ordered list. This would mean we'd rely on an entrypoint ordering convention, which seems worse than a naming convention.

Happy to change to a better alternative if you have suggestions.

Renaud-K (approved):

PR #4393, opened 2026-04-24
[PTSBE] Expose noise channel and parameters on TraceInstruction

taalexander:
When calling `cudaq.ptsbe.sample(..., return_execution_data=True)`, the TraceInstruction objects returned for Noise-type instructions now carry the channel's numeric parameters and the full kraus_channel object.

Previously a Noise TraceInstruction had an empty `params` list, and its underlying `cudaq::kraus_channel` (populated in C++) was not bound to Python. Users could see which channel fired via `inst.name` and which Kraus operator was selected via `kraus_selections[i].kraus_operator_index`, but could not recover the channel probability or the Kraus matrices from the trace without re-inspecting the NoiseModel.


sacpis (approved):
LGTM. Thanks @taalexander.

PR #4395, opened 2026-04-26
Fix observe broadcast fallback on REST QPU targets

### LINKED ISSUE #4363: `cudaq.observe` broadcast crashes with `TypeError` on all REST QPU targets (OQC, Quantinuum, …)
When `cudaq.observe` is called with a list or 2D-array argument -> triggers the broadcast path -> it crashes with a `TypeError` on any REST-based QPU target (including emulated targets like `oqc` and `quantinuum`):

```
TypeError: __init__(): incompatible function arguments. The following argument types are supported:
    1. __init__(self, arg0: float, arg1: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SpinOperator, arg2: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SampleResult, /) -> None
    2. __init__(self, arg0: float, arg1: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SpinOperator, arg2: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SampleResult, /) -> None
    3. __init__(self, arg0: float, arg1: object, arg2: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SampleResult, /) -> None

Invoked with types: cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.ObserveResult, NoneType, cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SpinOperator, cudaq.mlir._mlir_libs._quakeDialects.cudaq_runtime.SampleResult
```

The crash occurs because `ctx.getExpectationValue()` returns `None` for all REST QPU backends, and `__broadcastObserve` in `observe.py`) passes it directly to `ObserveResult` without a `None` fallback.

# Steps to reproduce

```
import cudaq
from cudaq import spin
import numpy as np
from typing import List

cudaq.set_target("oqc", emulate=True)   # or "quantinuum", emulate=True

qubit_count = 5
sample_count = 10
parameters = np.random.default_rng(13).uniform(low=0, high=1,
                                               size=(sample_count, qubit_count))

@cudaq.kernel
def kernel(qubit_count: int, parameters: List[float]):
    qvector = cudaq.qvector(qubit_count)
    for i in range(qubit_count - 1):
        rx(parameters[i], qvector[i])

results = cudaq.observe(kernel, spin.z(0), [qubit_count] * sample_count, parameters)
print([r.expectation() for r in results])
```

# Root cause

REST QPU backends (OQC, Quantinuum, IonQ, etc.)  (`BaseRemoteRESTQPU`) never write to `executionContext->expectationValue`, i.e., `None`.

The **non-broadcast** path in `observe.py` already handles this correctly (computes the expectation value from the sample counts). `__broadcastObserve` is missing the same fallback mechanism.

zeel2104:
## Summary

This change fixes `cudaq.observe()` broadcasting on REST-based QPU targets when `ExecutionContext.getExpectationValue()` returns `None`.

REST backends such as OQC and Quantinuum do not always populate `executionContext->expectationValue`. The non-broadcast `observe()` path already handled that by reconstructing the expectation value from sampled term results, but `__broadcastObserve()` passed the `None` value directly into `ObserveResult`, which caused a crash.

This patch makes the broadcast path use the same fallback behavior as the non-broadcast path.

## What changed

- Added a shared helper in `python/cudaq/runtime/observe.py` to resolve the expectation value:
  - return `ctx.getExpectationValue()` when available
  - otherwise reconstruct it from the sampled term expectations
- Updated `__broadcastObserve()` to use that helper
- Updated the existing non-broadcast path to reuse the same helper instead of duplicating the fallback logic
- Added backend regression tests for:
  - `python/tests/backends/test_OQC.py`
  - `python/tests/backends/test_Quantinuum_kernel.py`

## Why this fixes the issue

Previously, the broadcast path assumed the expectation value was always present in the execution context. On REST targets that assumption is false, so `ObserveResult(...)` received `None` and raised a `TypeError`.

With this change, broadcasted `observe()` calls now fall back to computing the expectation value from the returned sample counts, matching the behavior already used in the non-broadcast path.

## Testing

I added regression tests covering broadcasted `observe()` calls for OQC and Quantinuum.

What I was able to verify locally:
- the runtime fix is present in `python/cudaq/runtime/observe.py`
- the new backend regression tests are present and selected by pytest

What I could not fully verify locally:
- end-to-end execution of the new tests in my WSL environment

Reason:
- local runs abort during kernel compilation / MLIR lowering before `observe()` execution begins
- the crash occurs in `cudaq/kernel/ast_bridge.py` / `compile_to_mlir`
- because of that, the local environment does not reach the broadcast observe path, so it does not validate the new fallback behavior end-to-end

This appears to be unrelated to the `observe` broadcast fix itself, since the abort happens before the `observe()` runtime path is exercised.

## Local environment notes

During local setup I had to:
- build CUDA-Q from source in WSL
- build a custom LLVM/MLIR toolchain
- disable the Braket backend locally to avoid unrelated AWS SDK dependency issues

Even after that, the backend tests still abort earlier during kernel compilation in this environment.


1tnguyen:
> /ok to test c6cbfda8f7f7fb6183a33b8dd35175c9c79af050

1tnguyen:
Thanks @zeel2104 for your contribution.
Could you please take a look at the CI [checks](https://github.com/NVIDIA/cuda-quantum/pull/4395/checks)? There are DCO and code formatting issues. For DCO, just follow the instructions there to resolve the pushed commit without sign-off.

zeel2104:
@1tnguyen 
Done

1tnguyen:
> /ok to test 6721a6a

1tnguyen:
@zeel2104 There is a test failure in the new test case: https://github.com/NVIDIA/cuda-quantum/actions/runs/25017183407/job/73268586629?pr=4395#step:15:8084

It looks like this may be related to shot-based expectation evaluation in these hardware QPU mock tests. You might consider increase the number of shots and adjust the tolerance to be reasonable for stability.

zeel2104:
@1tnguyen 
Updated the OQC and Quantinuum broadcast observe regression tests to compare shot-based results against the analytic expectation with explicit shots and tolerance.

Local WSL validation:
python -m pytest python/tests/backends/test_OQC.py::test_OQC_observe_broadcast python/tests/backends/test_Quantinuum_kernel.py::test_quantinuum_observe_broadcast -q -s
2 passed, 1 warning in 40.54s

1tnguyen:
> /ok to test a4f8b8af318ccba8a62bfec8bc5161df98fc4f68

splch:
Thanks for tackling this! Verified the fix works against the real IonQ cloud REST endpoint (`target='ionq', qpu='simulator'`), not just the local emulator - the broadcasted `observe` now returns correct expectation values, matching analytical `cos(θ)` to within shot noise.

One small gap worth filling: the issue explicitly lists IonQ as affected, but the new tests only cover OQC and Quantinuum. Since IonQ goes through the same `BaseRemoteRESTQPU` path, an analogous test in `python/tests/backends/test_IonQ.py` would close the loop. Drop-in addition mirroring the style of the new OQC test:

```python
def test_ionq_observe_broadcast():
    qubit_count = 5
    sample_count = 4
    shots_count = 10000
    parameters = np.random.default_rng(13).uniform(low=0,
                                                   high=1,
                                                   size=(sample_count,
                                                         qubit_count))

    @cudaq.kernel
    def kernel(qubit_count: int, parameters: List[float]):
        qvector = cudaq.qvector(qubit_count)
        for i in range(qubit_count - 1):
            rx(parameters[i], qvector[i])

    results = cudaq.observe(kernel,
                            spin.z(0), [qubit_count] * sample_count,
                            parameters,
                            shots_count=shots_count)
    expected = np.cos(parameters[:, 0])

    assert len(results) == sample_count
    assert np.allclose([r.expectation() for r in results], expected, atol=0.1)
```

Happy to open a follow-up PR if that's easier than tacking it on.

1tnguyen:
> Happy to open a follow-up PR if that's easier than tacking it on.

@splch Thank you very much for verifying this fix.
Yeah, I think it'd be great if you could open a follow-up PR after this PR is merged in.

1tnguyen:
> /ok to test a9484d100e2775e88f44d0ee83e073b78cf87c36

1tnguyen (approved):
LGTM 👍

PR #4397, opened 2026-04-27
[NFC] Use std::span in remote client / argument-conversion APIs

lmondada:
Change signatures in `RemoteRuntimeClient`, `ArgumentConverter::gen*` and `mergeAllCallableClosures` to take `std::span`s instead of `const std::vector<void *> *` or `const std::vector<void *> &`.

This is more general and casting from a reference to span is implicit, so most call sites remain unchanged. Note that technically, a pointer to a vector can distinguish the empty vector from no vector (`nullptr`), but this distinction is never made in the code, so using `span`s removes this unnecessary distinction as well.

### DIFF
--- a/runtime/common/BaseRemoteSimulatorQPU.h
+++ b/runtime/common/BaseRemoteSimulatorQPU.h
@@ -194,7 +194,9 @@ public:
           return m_client->lowerKernelInPlace(prefabMod, name, *rawArgs);
         }
         return m_client->lowerKernel(*m_mlirContext, name, args, voidStarSize,
-                                     0, rawArgs);
+                                     0,
+                                     rawArgs ? std::span<void *const>{*rawArgs}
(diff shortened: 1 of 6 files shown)


runtime/common/BaseRemoteSimulatorQPU.h
```diff
@@ -194,7 +194,9 @@ class BaseRemoteSimulatorQPU : public QPU {
           return m_client->lowerKernelInPlace(prefabMod, name, *rawArgs);
         }
         return m_client->lowerKernel(*m_mlirContext, name, args, voidStarSize,
-                                     0, rawArgs);
+                                     0,
```
lmondada, line 197: this new ternary logic is unfortunate. This will be removed in my next PR, where spans will be passed in as arguments.

taalexander (approved):

PR #4398, opened 2026-04-27
Introduce KernelArgs to unify kernel argument types

lmondada:
Stacks on top of #4392

This PR introduces a `KernelArgs` type that stores kernel arguments EITHER as 'packed' arguments in a contiguous memory buffer OR as a vector of void*. `KernelArgs` also supports storing both representations, which is used by `hybridLaunchKernel`.

Using this type allows us to make the signatures of the launch endpoints more homogeneous, effectively hiding the different conventions as implementation details that only need to be handled within the function implementations.

Resulting signatures:

```c++
  launchKernel(const std::string &name, KernelThunkType kernelFunc, KernelArgs args);

  launchModule(const CompiledModule &compiled, KernelArgs args);
```


runtime/cudaq/platform/default/python/QPU.cpp
```diff
@@ -331,25 +329,23 @@ struct PythonLauncher : public cudaq::ModuleLauncher {
     bool isLocalSimulator =
         !(cudaq::is_remote_platform() || cudaq::is_emulated_platform());
 
-    std::vector<void *> closureArgs;
-
     // Special handling in case the arguments were already synthesized
+    auto rawArgs =
+        args.hasTypeErased() ? *args.getTypeErased() : std::span<void *>();
     size_t numArgs = rawArgs.size() - (hasResult ? 1 : 0);
+    std::vector<void *> closureArgs =
```
atgeller: Is always copying necessary now?

lmondada: No, if anything less copying is required :) e.g. in `python/runtime/utils/PyRemoteSimulatorQPU.cpp`, where `subspan` can be used instead of a new vector.

However in this case I had to leave the copying as is, because `closureArgs` gets modified on line 343. Previously, the copy was done by `operator=` on line 340.

python/runtime/utils/PyRemoteSimulatorQPU.cpp
```diff
@@ -48,43 +48,35 @@ static void
...
+                 void (*kernelFunc)(void *), cudaq::KernelArgs args) {
+  const auto packed = args.getPacked();
+  const auto rawArgs = args.getTypeErased();
+
+  mlir::ModuleOp module;
+  void *sendArgs = nullptr;
+  std::uint64_t sendSize = 0;
+  std::span<void *const> rawArgsForSend;
+
+  if (kernelFunc) {
```
lmondada, line 61: This if-else block merges `launchKernelImpl` (if branch) and `launchKernelStreamlineImpl` (else branch) that existed before.

It looks like a big diff but if you go line by line you will see that it should do the same!

runtime/cudaq/platform/default/python/QPU.cpp
lmondada: ah sorry, just saw that in the old version there was a `std::move` in the else clause! That move is kind of wild, as it is moving out of a const ref... 😅 

However, you're right that with my brand new spans I should be able to implement no-copying!

python/runtime/utils/PyRemoteSimulatorQPU.cpp
schweitzpgi: Are you saying this is merging two functions by putting them under a single name and embedding an `if then else` in the new function?

runtime/cudaq/platform/default/python/QPU.cpp
schweitzpgi: Are we sure about the lifetimes of the arguments? That is, does the actual argument values ever need to outlive the call itself, such as in asynchronous calls?

lmondada: This change has not affected any lifetime semantics.

Either way, it seems the previous code was correct (up to the move that I have removed): the arguments are no longer required once they have been specialized, so having the references valid for the duration of the method is enough.

python/runtime/utils/PyRemoteSimulatorQPU.cpp
lmondada: yes, but I only need to branch on a small part of the function. The two functions were broadly repeating themselves, so this change reduces duplication, as can be seen by the decrease in this file's line count.

That being said, this file is getting deleted in #4276 anyways

Renaud-K (approved):

mitchdz:
I removed from the queue, as I am blocking merge queues for the llvm22 update work now. I'll re-enqueue once open if no conflicts.

PR #4403, opened 2026-04-28
[cc] Measure handle type

khalatepradnya:
## Summary
Introduce `!cc.measure_handle` - the IR alias for the source-language `cudaq::measure_handle` - and widen `quake.mz`/`mx`/`my` and `quake.discriminate` ODS / verifiers to admit it alongside the existing `!quake.measure` form. Pure IR vocabulary: no path yet produces or consumes the new type. This is the prologue of a small stack; lowering through `convert-to-qir-api` lands in a follow-up PR, frontend bindings after that.

## Motivation
`cudaq::measure_handle` is a distinct source-language type from both raw integers and the existing measurement token: integer-shaped at the bit level, but identity-preserving for analyses that need to distinguish a measurement event from arbitrary i64 traffic. Landing the IR vocabulary first gives the QIR conversion PR and the frontend PRs a stable target without forcing the ODS contracts to churn step by step.

## What Changed
- **New type** `!cc.measure_handle` in the CC dialect; i64 payload, opaque to the IR. Registered with the CC dialect, lowered to `i64` in the CC->LLVM type converter, and `cc.cast` admits no-op `i64 <-> !cc.measure_handle`.
- **ODS widening** on `quake.mz`/`mx`/`my` results and `quake.discriminate` operand: now `!cc.measure_handle` or `!cc.stdvec<!cc.measure_handle>` are admitted in addition to the prior `!quake.measure` forms.
- **Verifier widening**: `verifyMeasurements` and `DiscriminateOp::verify` accept the new shape; arity diagnostics mention both spellings so users see why a scalar-typed result is rejected when measuring a register.
- **Tests**: `test/Transforms/roundtrip-ops.qke` (passthrough + `i64 <-> !cc.measure_handle` `cc.cast` round-trip), `test/Transforms/invalid.qke` (verifier negatives).

## Risks
No behavioral change in this PR: no path produces or consumes `!cc.measure_handle` until the follow-up PR lands. Risk is bounded to ODS coverage gaps that would surface in the consumer; the follow-up wires up `--convert-to-qir-api` and tests it.

## Downstream Impact
- CUDA-QX: none.
- Public API: none.
- Stack: lowering through `--convert-to-qir-api` lands in a follow-up PR built on this branch; C++/Python frontend bindings land after that.

PR #4404, opened 2026-04-28
[opt][qir] Lower `!cc.measure_handle` through `ConvertToQIRAPI`

khalatepradnya:
## Summary
* Lower `!cc.measure_handle` to its `i64` payload through `--convert-to-qir-api`'s existing `TypeConverter`, completing the IR-side of the `cudaq::measure_handle` feature. 
* Builds on #4403.

## Motivation
#4403 introduced `!cc.measure_handle` as IR vocabulary; nothing yet routes it to QIR. This PR adds the converter rule plus boundary bridging on `quake.mz` (which still calls a QIR function returning `Result*`) and `quake.discriminate` (which still consumes `Result*`), so handle-form kernels reach the QIR pipeline as `i64` payloads through the same `TypeConverter` machinery the rest of QIR conversion already uses.

## What Changed
- **`QIRAPITypeConverter`** gains three `addConversion` rules: `!cc.measure_handle -> i64`, plus recursive descent through `!cc.array<...>` and `!cc.stdvec<...>` so container-shaped function signatures, allocations, and pointers see consistent post-conversion element types. The `!cc.ptr<...>` recursion was already in place.
- **`MeasurementOpPattern`** detects when the original `quake.mz` produced a handle (its `measOut` is `!cc.measure_handle`) and emits a `cc.cast Result* -> i64` so downstream uses see the converted payload. The cast is materialized in the mz call's block, ahead of the optional terminator-relative insertion point used for record-output, so it dominates downstream `quake.discriminate` uses.
- **`DiscriminateOpToCallRewrite`** mirrors this on the read side: when the post-conversion operand is integer-typed it emits `cc.cast i64 -> Result*` before delegating to the existing read-result lowering. In the full-QIR (`!discriminateToClassical`) branch the bridge cast and the inner double-cast fold against each other, leaving a single `cc.cast i64 -> !cc.ptr<i1>` + `cc.load`.
- **`ExpandMeasurements`** accepts `!cc.measure_handle` alongside `!quake.measure` in `usesIndividualQubit`, so single-qubit handle measurements aren't rewritten as registers.
- **Predicate rename**: the misnamed `hasQuakeType` is now `needsTypeConversion`, leaf check extended to include `!cc.measure_handle`, recursion extended to descend through `!cc.array`/`!cc.stdvec`. The old name was incorrect — it has always reported "this op carries a type the converter rewrites," not "this op carries a Quake type."
- **Test**: `test/Transforms/qir_api_measure_handle.qke` covering scalar handle measurement + discriminate, function signature with handle parameter and return, `cc.alloca` of a scalar handle, static- and dynamic-size arrays of handles, `cc.stdvec<!cc.measure_handle>` in a function signature, `cc.indirect_callable<() -> !cc.measure_handle>`, and a no-handle negative.

## Risks
- `cc.loop` iter-args carrying `!cc.measure_handle` are not exercised by the conversion's region-aware patterns. Low immediate risk because no current frontend or test produces such IR; flagged as a follow-up.
- Container types beyond `cc.array`/`cc.stdvec` (e.g., a `cc.struct` with a handle field) are not in the converter's recursion. None of the current frontends produce these; not a regression vs. the prototype.

## Downstream Impact
- CUDA-QX: none.
- Public API: none.
- Stack: the next PR adds C++/Python frontend bindings that produce handle-form IR, which this PR now correctly routes.


schweitzpgi:
Heads-up: we may want to do the LLVM-22 rebase first. Not sure.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -1341,6 +1388,15 @@ struct MeasurementOpPattern : public OpConversionPattern<quake::MzOp> {
...
       call->setAttr(cudaq::opt::QIRRegisterNameAttr, regNameAttr);
+      // For handle-form callers, materialize the back-cast `Result* -> i64`
+      // here so it dominates downstream uses. The `!discriminateToClassical`
+      // branch below moves the insertion point to the block terminator for
+      // the record-output call, after which a cast would not dominate.
+      Value handleRes;
+      if (measOutIsHandle) {
+        auto i64Ty = rewriter.getI64Type();
+        handleRes = rewriter.create<cudaq::cc::CastOp>(loc, i64Ty, res);
+      }
```
1tnguyen: nit: would it be more readable if we write it as:
```
auto i64Ty = rewriter.getI64Type();
Value handleRes = measOutIsHandle ? rewriter.create<cudaq::cc::CastOp>(loc, i64Ty, res) : res;
```

then later 

```
SmallVector<Value> results = {handleRes};
```
?

I was wondering about uninitialized `handleRes` until I saw that we use the same `measOutIsHandle` condition in both places :)

1tnguyen (approved):
LGTM 👍

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
khalatepradnya: Fixed in [20a3361](https://github.com/NVIDIA/cuda-quantum/pull/4404/commits/20a3361f0762302c7db79f904f4c53e4549a8881)

sacpis:
Reviewing it now...

sacpis (approved):

PR #4405, opened 2026-04-28
[opt] Expand measurements: handle `!cc.stdvec<!cc.measure_handle>`

khalatepradnya:
## Summary
* Extend `--expand-measurements` to scalarize `quake.mz`/`mx`/`my` `%veq -> !cc.stdvec<!cc.measure_handle>`. 
* Builds on #4404
* No source-language or runtime API change.

## Motivation
The pass previously hardcoded `!quake.measure` for per-element output and only handled `quake.discriminate` consumers of the vector result. Handle-typed vector measurements require per-element `!cc.measure_handle` output and can flow to non-discriminate consumers (returns, stores, calls), neither of which the legacy `vector<bool>`-only rewrite supported.

## What Changed
- `ExpandRewritePattern` tracks the input stdvec's element type and emits per-element measurements of the matching type (`!quake.measure` or `!cc.measure_handle`).
- Consumers are classified as discriminate vs non-discriminate. Handle inputs allocate each buffer only when its consumer class is present; legacy `!cc.stdvec<!quake.measure>` inputs always allocate the i1 buffer so existing AST-Quake CHECK lines stay stable.
- Original op is replaced via `replaceOp` (atomic) instead of `eraseOp`, so partial conversion does not try to re-legalize downstream `func.return` consumers.
- New lit test `test/Transforms/expand_measurements_handle.qke` covers handle stdvec with each consumer class (return-only, discriminate-only, mixed, `cc.store`), mixed `ref + veq` operands, and `mx`/`my` parity.


test/Transforms/expand_measurements_handle.qke
```diff
@@ -0,0 +1,121 @@
...
+func.func @handle_disc(%v: !quake.veq<?>) -> !cc.stdvec<i1> {
+  %mz = quake.mz %v : (!quake.veq<?>) -> !cc.stdvec<!cc.measure_handle>
+  %d = quake.discriminate %mz : (!cc.stdvec<!cc.measure_handle>) -> !cc.stdvec<i1>
+  return %d : !cc.stdvec<i1>
+}
+
+// CHECK-LABEL:   func.func @handle_disc(
+// CHECK:           %[[I8BUF:.*]] = cc.alloca i8{{\[}}%{{.*}}
+// CHECK-NOT:       cc.alloca !cc.measure_handle
+// CHECK:             %[[H:.*]] = quake.mz %{{.*}} : (!quake.ref) -> !cc.measure_handle
```
atgeller, line 55: Is there meant to be a `cc.loop while` here that was just omitted from the CHECKs? If so, is there a reason for the omission? The CHECKs are a little confusing without it explicitly there. (Same in below cases)

khalatepradnya: Omitted to keep the test smaller, but I will add them back for clarity...

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -84,37 +154,63 @@ class ExpandRewritePattern : public OpRewritePattern<A> {
...
+    assert((replacementVal != nullptr) == hasNonDiscUser &&
+           "handle-vector replacement must exist iff a non-discriminate "
+           "consumer was present");
+
+    SmallVector<Value> replacements;
+    replacements.push_back(replacementVal);
+    for (auto wire : measureOp.getWires()) {
+      (void)wire;
+      replacements.push_back(nullptr);
+    }
```
sacpis: Is this loop ever run? Because on line 197 we have an assertion that makes sure this loop does not run. So either we remove the assertion or this loop.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -32,6 +32,18 @@ bool usesIndividualQubit(A x) {
...
 // Generalized pattern for expanding a multiple qubit measurement (whether it is
 // mx, my, or mz) to a series of individual measurements.
+//
+// Handles both result-type families that the vector form of `quake.mz`/`mx`/
+// `my` can carry:
+//   - `!cc.stdvec<!quake.measure>` -- the legacy form. The only legitimate
+//     consumer is `quake.discriminate`, so the rewrite folds the per-element
+//     measurements straight into a `cc.stdvec_init -> !cc.stdvec<i1>`.
+//   - `!cc.stdvec<!cc.measure_handle>` -- the handle-vector value can have
+//   non-discriminate consumers  Those consumers
```
sacpis: nit: `non-discriminate consumers. Those ...`

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -40,6 +52,42 @@ class ExpandRewritePattern : public OpRewritePattern<A> {
...
+    // consumers by construction; the handle path may have either, both,
+    // or none.
+    bool hasDiscUser = false;
+    bool hasNonDiscUser = false;
+    for (auto *u : measureOp.getMeasOut().getUsers()) {
+      if (isa<quake::DiscriminateOp>(u))
+        hasDiscUser = true;
+      else
+        hasNonDiscUser = true;
+    }
```
sacpis, line 82: I think here we can exit early once we have both flags set.

```
if (hasDiscUser && hasNonDiscUser)
    break;
```

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -84,37 +154,63 @@ class ExpandRewritePattern : public OpRewritePattern<A> {
...
-        rewriter.template replaceOpWithNewOp<cudaq::cc::StdvecInitOp>(
-            disc, stdvecTy, buffCast, totalToRead);
-      }
+    // 4. Replace each `quake.discriminate` consumer with a
+    // `cc.stdvec_init -> !cc.stdvec<i1>` over the i1 buffer.
+    if (needI1Buf) {
+      auto stdvecI1Ty = cudaq::cc::StdvecType::get(ctx, i1Ty);
+      auto ptrArrI1Ty =
+          cudaq::cc::PointerType::get(cudaq::cc::ArrayType::get(i1Ty));
+      for (auto *out : llvm::to_vector(measureOp.getMeasOut().getUsers()))
```
sacpis: Can we collect these disc users in some vector during the [classification](https://github.com/NVIDIA/cuda-quantum/pull/4405/changes#diff-702ee1f7eb124f627680555a27e727a4ed25842fa89ce339ab5222ac9ebe0cf4R78) and reuse it here? That way we can avoid `to_vector` copy of user-iterator range and `dyn_cast` per user below.

lib/Optimizer/Transforms/ExpandMeasurements.cpp
sacpis: I think we can collect disc users here itself so those can be used below on line 173, something like this

```
SmallVector<quake::DiscriminateOp> discUsers;
bool hasNonDiscUser = false;
for (auto *u : measureOp.getMeasOut().getUsers()) {
    if (auto d = dyn_cast<quake::DiscriminateOp>(u))                                                                                                                      
        discUsers.push_back(d);                                                                                                                                             
    else                                                                                                                                                                  
        hasNonDiscUser = true;                                                                                                                                              
  }               

bool needI1Buf = !isHandleResult || !discUsers.empty();
```

lib/Optimizer/Transforms/ExpandMeasurements.cpp
```diff
@@ -124,6 +220,78 @@ using MxRewrite = ExpandRewritePattern<quake::MxOp>;
...
+        cudaq::cc::StdvecSizeOp::create(rewriter, loc, i64Ty, handleVec);
+    auto handleArrPtrTy =
+        cudaq::cc::PointerType::get(cudaq::cc::ArrayType::get(handleTy));
+    Value handleData = cudaq::cc::StdvecDataOp::create(
+        rewriter, loc, handleArrPtrTy, handleVec);
+    // Output is held in an i8 buffer, then bitcast to `!cc.ptr<!cc.array
+    // <i1 x ?>>` for the wrap. This matches the convention used by the
+    // measurement-op pattern above (steps 2 + 4) so downstream passes see
+    // the same shape regardless of which path produced the i1 vector.
+    Value i1Buff = cudaq::cc::AllocaOp::create(rewriter, loc, i8Ty, vecSize);
```
sacpis, line 257: Does only the legacy pattern requires `i8` buffer? I don't see there is any restriction on this new pattern.
So can we try to allocate `i1Ty` directly?

```
Value i1Buff = cudaq::cc::AllocaOp::create(rewriter, loc, i1Ty, vecSize);
<code>
cudaq::cc::StoreOp::create(builder, loc, bit, byteAddr);
<code>
rewriter.replaceOpWithNewOp<cudaq::cc::StdvecInitOp>(disc, stdvecI1Ty,
                                                         i1Buff, vecSize);
```

I think that can remove one cast per loop iteration and the bitcast after the loop as well.

sacpis (approved):

lib/Optimizer/Transforms/ExpandMeasurements.cpp
khalatepradnya: Good points, it improves readability too. Updated to traverse the users only once, a9eedc3073fb060c35e9a5589a1f4c414362301e

lib/Optimizer/Transforms/ExpandMeasurements.cpp
khalatepradnya: I agree on the redundancy, but, I think this should be addressed separately, if we want to.
If I change only `ExpandStdvecHandleDiscriminate` here, then the legacy `ExpandRewritePattern` would produce structurally different IR for the same` cc.stdvec<i1>`, which I think is harder to read than one redundant cast.
And practically, the byte-size accounting is identical either way ((1+7)/8 = 1), so the optimization changes only the IR shape.

PR #4412, opened 2026-04-29
Fixes a miscompile of cc.loop while-loop

### LINKED ISSUE #4401: Incorrect behavior for ``cc.loop while``
### Required prerequisites

### Describe the bug

1) The CUDA-Q while loop incorrectly yields the loop-carried variable's state.
2) The CUDA-Q compiler backend incorrectly lowers the sge (signed greater-than-or-equal) condition inside cc.loop blocks, effectively treating it as a strict greater-than (>) operation and causing the loop to drop its final boundary iteration.

1) Purely classical ``while``loop:
```
import cudaq

@cudaq.kernel
def test_kernel() -> int:

    val = 3
    while val >= 0:
        val -= 1
    return val   

#print(test_kernel)
cudaq.run(test_kernel, shots_count=1)
```
> [3] # Expected -1

```
@cudaq.kernel
def test_kernel() -> int:

    val = 5
    while val <= 10:
        val += 1
    return val   

#print(test_kernel) 
cudaq.run(test_kernel, shots_count=10)
```
> [6] # Expected 11

2) **Quantum gates** inside ``while`` loop:
```
@cudaq.kernel
def multi_qubit_kernel() -> int:
    q = cudaq.qvector(3)

    val = 2
    while val >= 0:
        x(q[val])
        val -= 1

    return mz(q[0])

#print(multi_qubit_kernel)
cudaq.run(multi_qubit_kernel, shots_count=1)
```
> [0] # Expected 1

### Expected behavior

In the first example, we would expect that the correctly accumulated ``val``is returned.

In the second example, we would expect that an x-gate is applied to each qubit, and the measurement of qubit 0 will yield 1 (not 0). The output is 0, thus the expected x-gate is not applied to qubit 0. 

Note that, the printed MLIR looks correct (we would expect an x-gate applied to qubits 2, 1 and 0):

```
    %c1_i64 = arith.constant 1 : i64
    %c2_i64 = arith.constant 2 : i64
    %0 = quake.alloca !quake.veq<3>
    %1 = cc.loop while ((%arg0 = %c2_i64) -> (i64)) {
      %5 = arith.cmpi sge, %arg0, %c0_i64 : i64
      cc.condition %5(%arg0 : i64)
    } do {
    ^bb0(%arg0: i64):
      %5 = quake.extract_ref %0[%arg0] : (!quake.veq<3>, i64) -> !quake.ref
      quake.x %5 : (!quake.ref) -> ()
      %6 = arith.subi %arg0, %c1_i64 : i64
      cc.continue %6 : i64
    } step {
    ^bb0(%arg0: i64):
      cc.continue %arg0 : i64
    }
```

An equivalent loop with "> -1" works es expected:

```
@cudaq.kernel
def multi_qubit_kernel() -> int:
    q = cudaq.qvector(3)

    val = 2
    while val > -1:
        x(q[val])
        val -= 1

    return mz(q[0])

#print(multi_qubit_kernel)
cudaq.run(multi_qubit_kernel, shots_count=1)
```
> [1]

Not a regression

### Environment

- **CUDA-Q version**: 0.14.0
- **Python version**: 3.11.14
- **Operating system**: macOS 15.7.4

### Suggestions

_No response_

sacpis:
Fixes a miscompile of cc.loop while-loop with closed-interval comparisons (>=, <=).

- Adding uge/sge to isClosedIntervalForm.
- After the loop, replacing external uses of the induction-position result

Fixes #4401

### DIFF
--- a/lib/Optimizer/Transforms/LoopAnalysis.cpp
+++ b/lib/Optimizer/Transforms/LoopAnalysis.cpp
@@ -75,8 +75,11 @@ static bool isaConstantOf(Value v, std::int64_t hasVal) {
 }
 
 static bool isClosedIntervalForm(arith::CmpIPredicate p) {
-  return p == arith::CmpIPredicate::ule || p == arith::CmpIPredicate::sle ||
-         p == arith::CmpIPredicate::uge || p == arith::CmpIPredicate::sge;
+  return p == arith::CmpIPredicate::ule || p == arith::CmpIPredicate::sle;
+}
+
+static bool isClosedIntervalDownForm(arith::CmpIPredicate p) {
+  return p == arith::CmpIPredicate::uge || p == arith::CmpIPredicate::sge;
 }
 
 static bool isSemiOpenIntervalForm(arith::CmpIPredicate p) {
@@ -333,8 +336,8 @@ bool opt::LoopComponents::shouldCommuteStepOp() const {
 }
 
 bool opt::LoopComponents::isClosedIntervalForm() const {
-  auto cmp = cast<arith::CmpIOp>(compareOp);
-  return ::isClosedIntervalForm(cmp.getPredicate());
+  auto p = cast<arith::CmpIOp>(compareOp).getPredicate();
+  return ::isClosedIntervalForm(p) || ::isClosedIntervalDownForm(p);
 }
 
 bool opt::LoopComponents::isLinearExpr() const {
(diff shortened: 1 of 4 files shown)


schweitzpgi (commented):
.

schweitzpgi (approved):
LGTM.

PR #4416, opened 2026-04-29
Reduced QPU headers dependency on JSON and LLVM/MLIR

Renaud-K:
The end goal is to be able to have QPU headers visible from user code, so compile-time checks can be performed based on QPU traits without leaking headers from 3rd party libraries.  

1. Importing the latest `nlohmann::json` library which introduces a header for forward declaration of the `nlohmann::json` type.
   - Using FectchContent
   - Removed tpls/json
   - Exporting json_fwd.hpp in the install tree
   - Introducing the `cudaq_json` type to encapsulate the forward declared `nlohmann::json`. Only needed by `KernelExecution` 
   - Only including the full `nlohmann::json` type in `.cpp` files.

2. Adding to the build tree, the files newly needed at runtime to enable `nvq++` to run from the build tree
  (TargetConfig.h Version.h  json_fwd.hpp)

3. Moved Fermioniq implementation that was using `nlohmann::json` to the `.cpp` file. Merged FermioniqQPU with FermioniqBaseQPU.

4. Introduced qpu_utils.{h,cpp} to prevent header includes to bleed through the qpu headers. This, for now, stops the inclusion of:
```
#include "cudaq/Optimizer/Builder/RuntimeNames.h"
#include "cudaq/Support/TargetConfig.h"
#include "cudaq/Support/TargetConfigYaml.h"
#include "llvm/Support/Base64.h"
```

5. Using `ModuleOp::getFromOpaquePointer`, `ModuleOp::getAsOpaquePointer` and an MLIRContextDesctuctor to pass the MLIR modules and contexts through the QPUs. 
This is a stop gap until we have a unified launch interface @lmondada is working on.


CMakeLists.txt
```diff
@@ -544,7 +544,19 @@ if(POLICY CMP0146)
 endif()
 add_subdirectory(tpls/fmt)
 cmake_policy(POP)
-include_directories(SYSTEM tpls/json/include)
+include(FetchContent)
+FetchContent_Declare(json
```
1tnguyen: In the past, we seem to prefer submodule over `FetchContent` (e.g., https://github.com/NVIDIA/cuda-quantum/pull/1380)

1tnguyen (approved):
LGTM 💯 

Just one question about `FetchContent` vs. submodule convention.

CMakeLists.txt
Renaud-K: @amccaskey had asked about using `FetchContent` over submodules for `nanobind` when I had introduced it. cudaqx is using `FetchContent` [for it](https://github.com/NVIDIA/cudaqx/pull/512). So, I am not really sure what to do now. I will talk to @bettinaheim tomorrow.

runtime/common/BaseRemoteSimulatorQPU.h
```diff
@@ -142,11 +142,12 @@ class BaseRemoteSimulatorQPU : public QPU {
   }
 
   CompiledModule compileModule(const std::string &kernelName,
-                               mlir::ModuleOp module,
+                               const void *modulePtr,
```
lmondada, line 145: as you correctly pointed out, this will be resolved with https://github.com/NVIDIA/cuda-quantum/pull/4413

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
```diff
@@ -100,7 +104,8 @@ class Compiler {
       std::shared_ptr<mlir::MLIRContext> context);
 
 public:
-  static std::pair<mlir::ModuleOp, std::unique_ptr<mlir::MLIRContext>>
+  static std::pair<const void *,
```
lmondada: Instead of using the custom deleter `MLIRContextDeleter`, it would be easier to use `std::shared_ptr<mlir::MLIRContext>`. The `context` will have to be promoted to a `shared_ptr` anyway once stored in `CompiledModule` so you might as well construct it here. That way you get the type-erased destructor for free :)

lmondada (approved):
Thank you, nice clean up!

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
Renaud-K: I am using a shared_ptr now. But if the semantic of this pointer is going to be to have a single owner then I would rather we make it ultimately a unique_ptr, albeit with a nicer wrapper.

Renaud-K:
> /ok to test f4af53f97ee339e45f961d9c7dd44ce96b1e23fd

CMakeLists.txt
Renaud-K: After talking to @bettinaheim, I reverted to using tpls.

Renaud-K:
> /ok to test e54f5578f78b8ae071b596617d9d6a59139e21e6

Renaud-K:
> /ok to test 2e103275b4fa0153e63a9b60817c266cc71c1bc7

Renaud-K:
> /ok to test 89704757e7a67a1de741ac9e58f12d1b3c5cab4b

Renaud-K:
> /ok to test 44c04ad0eb08563799b4756c9f2fde3bea64867f

PR #4417, opened 2026-04-30
Restoring static linkage

sacpis:
Move the anonymous namespace to wrap only BufferInfo and mark hostDataFromDevice plus the CuPy helper functions static, matching the convention used elsewhere in the file.


1tnguyen (approved):
LGTM 👍 Thanks @sacpis

PR #4418, opened 2026-04-30
Updated `sample` semantics

### LINKED ISSUE #4153: Explicit Measurements QPU Compatability
### Required prerequisites

### Description

In my CUDA-Q program, I need the output bitstring ordering to follow the order of the measurement statements.

We are running a noisy simulation (included `noise_model`) in `sample()` parameters of a large circuit for many shots, using the `nvidia` backend target. Below is a simplified example of the issue:

```python
import cudaq

@cudaq.kernel
def circuit():
    q = cudaq.qvector(4)

    x(q[2])   # Apply X gate to the 3rd qubit (index 2)

    mz(q[2])
    mz(q[0])
    mz(q[1])
    mz(q[3])

if __name__ == "__main__":
    cudaq.set_target('nvidia')
    result = cudaq.sample(circuit)
    result_explicit = cudaq.sample(circuit, explicit_measurements=True)
```

For this circuit:

* `cudaq.sample(circuit)` returns `0010`
* `cudaq.sample(circuit, explicit_measurements=True)` returns `1000`

The `1000` result is the behavior I want, since the measured `1` comes from `q[2]`, which is measured first and should therefore appear in the first bit position of the returned bitstring.

### Issue

Without `explicit_measurements=True`, CUDA-Q appears to ignore the explicit order of the `mz(...)` statements when constructing output bitstrings.

Using `explicit_measurements=True` fixes the ordering, but in my case it appears to no longer use the GPU effectively. Performance drops dramatically and GPU utilization falls significantly compared to the default sampling path.

### Request

It would be very helpful if CUDA-Q could support explicit measurement ordering while still using the fast GPU sampling path.

If this is not currently supported, it would also be helpful to clarify whether `explicit_measurements=True` disables GPU-accelerated sampling, and whether there is another recommended way to preserve measurement order while retaining GPU performance.

anpaz:
**Summary**

Fixes #4153 by updating `cudaq::sample` / `cudaq.sample` measurement result semantics so sampled bitstrings follow user measurement order when measurements are present; if the kernel includes no measurements, the current behavior of returning a bitstring based on the allocation order remains.

**What Changed**

- Updated default sample semantics:
  - Kernels with no measurements use implicit final sampling over allocated qubits.
  - Kernels with measurements normally return `__global__` bitstrings in measurement/program order.
  - Terminal `mz` / `mx` / `my` measurements that already follow allocation order remain allocation-order compatible.
  - `explicit_measurements=False` is rejected when it would change returned bitstrings.
- Added conservative MLIR measurement analysis and metadata to record when a kernel requires explicit measurement-order semantics.
- Updated Python `sample` / `sample_async` default `explicit_measurements` to auto mode via `None`.
- Kept C++ `sample_options::explicit_measurements` as a `bool`, but documented it as a deprecated compatibility option.
- Preserved named measurement registers in `sample_result` where available for compatibility.
- Updated docs to describe the result contract in terms of outcome semantics rather than implementation fast paths.
- Moved the QIR Base Profile measurement-order verifier into shared verifier code so both runtime and `cudaq-translate` paths can use it.
- Updated affected tests and added coverage for measurement-order defaults, allocation-order-compatible measurements, rejected legacy requests, no-measurement kernels, Python sync/async, and C++ target tests.

> Note: `explicit_measurements` is now deprecated as a user-facing semantic switch. Users should rely on the default sample behavior: kernels with measurements return results in measurement order when that affects the bitstring, while kernels without measurements use implicit final allocation-order sampling. The flag remains mainly as a compatibility/backend capability mechanism: explicit_measurements=False requests legacy allocation-order behavior and is only accepted when it does not change the returned outcomes, while targets that cannot support measurement-order sampling can reject kernels that require it.


include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -62,6 +62,12 @@ void createClassicalOptimizationPipeline(
     std::optional<bool> allowBreak = std::nullopt,
     std::optional<bool> allowClosedInterval = std::nullopt);
 
+/// Recompute Quake function metadata and propagate it to the module. This is
+/// useful after transformations such as measurement expansion, loop unrolling,
+/// and allocation combining, which can expose a more precise measurement shape
+/// than was available earlier in the pipeline.
+void addQuakeMetadataRefresh(mlir::OpPassManager &pm);
```
schweitzpgi, line 74: We already have several "metadata" passes. Do we need yet another one?

lib/Optimizer/CodeGen/Pipelines.cpp
```diff
@@ -71,8 +71,7 @@ void createCommonTargetCodegenPipeline(
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createCSEPass());
     pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddDeallocs());
-    pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddMetadata());
-    pm.addPass(cudaq::opt::createQuakePropagateMetadata());
+    cudaq::opt::addQuakeMetadataRefresh(pm);
```
schweitzpgi, line 74: It appears "no" is the answer to my question.

include/cudaq/Optimizer/Transforms/SampleMeasurementAnalysis.h
```diff
@@ -0,0 +1,284 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
```
schweitzpgi, line 9: This appears to be in the wrong place.

Is it a pure analysis? It appears to be. Why is it entirely implemented in a header file? Why not provide an API? It looks to be used from both the runtime and the new "refresh metadata" pass. Why?

schweitzpgi (commented):
In general, the compiler makes no guarantee to maintain qubits or maintain their relative order. The correct approach is to tag measurement operations with identifiers (StringAttr) and post-process them as needed.

khalatepradnya (commented):
Thanks for working on this. I see that the default `sample` semantics have been updated so measurement order is preserved automatically when needed, and that seems like the right user-facing direction.

One thing I am still not fully understanding: does this PR address the original performance concern in #4153? My reading is that kernels requiring explicit measurement-order semantics will still use `explicitMeasurements` internally, and for non-Stim local simulators that do not support buffered explicit sampling, that path still appears to execute one shot at a time. So the default result would now be semantically correct, but the explicit-measurements performance issue may still remain for targets like `nvidia` / other non-Stim local simulators.

Am I missing something?

anpaz:
We decided to not move forward with this breaking change for now. See comments in #4153 for more details.

PR #4419, opened 2026-04-30
[NFC] Split out CompiledModule container into class

lmondada:
NOTE: This is a re-post of #4392, which I merged into the wrong branch! It's already been reviewed, discussed and approved.

---

This PR splits out the container that is used in CompiledModule into its own type. This is so that it can be re-used by other upcoming types that look very similar, e.g. KernelArgs.

I took the opportunity to change to using a vector of pairs instead of a std::map to store the artifacts. This should be faster (most of the time, there will be <5 artifacts) and means that several artifacts of different types can share the same name. This removes the need to adopt some naming convention to differentiate multiple artifact types for the same kernel, as they can share the same name.


taalexander (approved):
Reapproving

PR #4422, opened 2026-04-30
[NFC] Let's follow the coding conventions

schweitzpgi:


atgeller (approved):
Thanks @schweitzpgi! Is there anyway we can add a check for this to pre-commit?

PR #4427, opened 2026-05-01
[unittests] Build unit tests in library mode

khalatepradnya:
## Summary

Set `CUDAQ_LIBRARY_MODE` at directory scope in `unittests/CMakeLists.txt` so every unit-test binary in the tree compiles against the library-mode inline implementations of the runtime headers.

## Motivation

The unit-test binaries instantiate `__qpu__` functor structs and invoke `cudaq::sample(myKernel)` directly from C++ test code, with no `nvq++` / `cudaq-quake` step in between. The AST bridge never runs, so measurement and gate calls reach the inline bodies in `runtime/cudaq/qis/qubit_qis.h` on the host. That is exactly the library-mode execution model.

Today this happens to work because the historic inline `mz` / `mx` / `my` bodies execute on host. Future runtime header changes can make this implicit dependency explicit by trapping host invocation of the MLIR-mode entry points (e.g., `std::abort()` stubs); pre-emptively setting the macro keeps these test binaries on the supported execution path.

### DIFF
--- a/runtime/cudaq/qis/execution_manager.h
+++ b/runtime/cudaq/qis/execution_manager.h
@@ -51,7 +51,7 @@ private:
   int result = 0;
 
   /// Unique integer for measure result identification
-  std::size_t uniqueId = 0;
+  [[maybe_unused]] std::size_t uniqueId = 0;
 
 public:
   measure_result(int res, std::size_t id) : result(res), uniqueId(id) {}
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -13,6 +13,20 @@ set (CMAKE_CXX_FLAGS
 SET(CMAKE_EXE_LINKER_FLAGS "")
 SET(CMAKE_SHARED_LINKER_FLAGS "")
 
+# These unit test executables instantiate `__qpu__` kernel functor structs and
+# call `cudaq::sample(myKernel)` directly from C++ test code, with no `nvq++`
+# / `cudaq-quake` step. The AST bridge never runs, so measurement and gate
+# calls reach the inline bodies in `runtime/cudaq/qis/qubit_qis.h` on the
+# host. That is the same execution model as library mode, but without the
+# `CUDAQ_LIBRARY_MODE` macro the headers' MLIR-mode branches are selected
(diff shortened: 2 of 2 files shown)


1tnguyen (approved):
👍

schweitzpgi (commented):
Q: is there a way to focus this on specific tests? or do none of the unittests really expect to behave properly?

khalatepradnya:
> Q: is there a way to focus this on specific tests? or do none of the unittests really expect to behave properly?

Good question. I think the useful boundary here is the CMake test target/subtree, rather than trying to identify individual gtest cases.

We could try to narrow this to only the tests that currently instantiate `__qpu__` kernels and execute them directly from C++, but I’m afraid that would be a bit fragile. Most of these unit tests are not going through the normal `nvq++` / `cudaq-quake` path, so new tests added here could easily miss the define and regress in the same way.

Since library mode is on its way out anyway, this probably needs a proper revisit when that cleanup happens. I’d rather keep this workaround broad and explicit for now, then fix the test harness correctly as part of removing/replacing library mode.

khalatepradnya:
> /ok to test 1cc79eb

khalatepradnya:
> /ok to test 193d1bd

khalatepradnya:
> /ok to test 53f2a87

lmondada (approved):

lmondada (approved):
We should discourage writing tests using library mode, so ideally library mode would be opt-in and only enabled on the tests that actually require it. For instance, tests within `common` or `domains` definitely do not need it. 

However, I understand that figuring out where exactly library mode is required and only enabling it at this finer granularity might be more work than it's worth, so feel free to merge this in as-is if you think this is the simplest solution.

lmondada (approved):
EDIT: ah never mind, just saw Eric made the same comment! Please ignore :)

~We should discourage writing tests using library mode, so ideally library mode would be opt-in and only enabled on the tests that actually require it. For instance, tests within `common` or `domains` definitely do not need it.~

~However, I understand that figuring out where exactly library mode is required and only enabling it at this finer granularity might be more work than it's worth, so feel free to merge this in as-is if you think this is the simplest solution.~

PR #4430, opened 2026-05-01
fixing GIL deadlock in mqpu async kernel compilation

sacpis:
The test `test_mpi_mqpu.py` was hanging where PyGILState_Ensure() on the worker waits for the main thread to release the GIL, but the main thread is in `asyncResult.get()` holding the GIL.

`addPythonSignalInstrumentation` now skips installing the per-pass PyGILState_Ensure based instrumentation when called from a thread that doesn't currently hold the GIL.


1tnguyen (approved):
Thanks @sacpis!

PR #4431, opened 2026-05-01
Fixing memory leak in async API

### LINKED ISSUE #3355: MQPU sample_async memory leak
### Required prerequisites

### Describe the bug

When using the mqpu option with sample_async, after running the circuit on different qpu and waiting for the result with get(), the memory of the CPU (host memory) keeps increasing as a function of the number of iterations of the kernel (nb in the code).
In fact, the leak is present even if qpu_id is set to 0, but with multiple qpu it's even faster.

```
import cudaq


cudaq.set_target("nvidia", option="mqpu")

@cudaq.kernel 
def kernel(N:int): # some circuit
    q = cudaq.qvector(N) 
    for i in range(100):
        h(q)
    for j in range(N):
        x(q[j])
        h(q)

N=30
nb=100


for t in range(nb): 
    L=[]
    for i in range(8): 
        L.append(cudaq.sample_async(kernel, N,qpu_id=i%8, shots_count=1))
    for j in range(8): 
        L[j].get()

```
### Expected behavior

I thought that the CPU memory would reset after get() is done but it is not the case.

Not a regression

### Environment

- **CUDA-Q version**: latest
- **Python version**: 3.12
- **C++ compiler**: 
- **Operating system**: Linux

### Suggestions

_No response_

sacpis:
The clone captured into the async task lambda was never erased. That left the operation in the MLIR context. Now we are wrapping it in a shared pointer with an erasing deleter that ties cleanup to lambda destruction. Similar solution as how we already handled it in [pyLaunchModule](https://github.com/NVIDIA/cuda-quantum/blob/main/python/runtime/cudaq/platform/py_alt_launch_kernel.cpp#L682).

The memory leak of around 15 KB per call is no more. 

Fixes #3355

### DIFF
--- a/python/runtime/cudaq/algorithms/py_observe_async.cpp
+++ b/python/runtime/cudaq/algorithms/py_observe_async.cpp
@@ -76,14 +76,19 @@ static async_observe_result pyObserveAsync(const std::string &shortName,
 
   // Launch the asynchronous execution.
   nanobind::gil_scoped_release release;
+  auto clonedMod = std::shared_ptr<mlir::ModuleOp>(
+      new mlir::ModuleOp(mod.clone()), [](mlir::ModuleOp *p) {
+        p->erase();
+        delete p;
+      });
   return details::runObservationAsync(
-      detail::make_copyable_function([opaques = std::move(opaques), shortName,
-                                      mod = mod.clone()]() mutable {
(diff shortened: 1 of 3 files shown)


1tnguyen (approved):
LGTM 👍

PR #4432, opened 2026-05-01
Additional tracing and resource estimation target improvements

taalexander:
Follow ups as noted by @schweitzpgi and @boschmitt.

### DIFF
--- a/runtime/cudaq/platform/default/circuit-opt-bench.yml
+++ b/runtime/cudaq/platform/default/circuit-opt-bench.yml
@@ -24,7 +24,6 @@ config:
   # routing basis, prepare wire metadata, optionally route, then lower routed
   # two-qubit work to CZ. With no device argument, routing is bypassed.
   jit-mid-level-pipeline: "apply-op-specialization,aggressive-inlining,decomposition{basis=h,rx,ry,rz,x,x(1)},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{basis=h,rx,ry,rz,x,z(1)}"
-  # Target low-level cleanup stage: normalize controls, simplify IR, and remove
-  # wire-set symbols left by routing prep.
-  jit-low-level-pipeline: "func.func(apply-control-negations,canonicalize,cse),symbol-dce"
+  # Target low-level cleanup stage: remove wire-set symbols left by routing prep.
+  jit-low-level-pipeline: "symbol-dce"
   library-mode: false
--- a/runtime/cudaq/platform/default/circuit-opt-bench.yml
+++ b/runtime/cudaq/platform/default/circuit-opt-bench.yml
@@ -17,27 +17,14 @@ target-arguments:
 
 config:
   nvqir-simulation-backend: qpp
-  # Retained high-level pass group:
-  # - get-concrete-matrix materializes matrices for registered custom ops.
-  # - unitary-synthesis lowers those custom unitaries, notably Benchpress QV
-  #   SU(4) blocks, to explicit gates before decomposition and routing.
+  # Target high-level stage: materialize registered custom-op matrices and
+  # synthesize them to gates before target lowering.
   jit-high-level-pipeline: "get-concrete-matrix,unitary-synthesis"
-  # Retained mid-level pass groups:
-  # - apply-op-specialization and aggressive-inlining expose synthesized helper
-  #   functions so later passes see the KAK-decomposed gates, not opaque
-  #   quake.apply ops.
-  # - The first decomposition targets a CX-capable basis for the router.
-  # - Allocation, wire-set, and wire-index passes produce the IR shape expected
-  #   by qubit-mapping and later lowering.
-  # - qubit-mapping runs SABRE routing when device is set. %DEVICE:bypass%
-  #   defaults to bypass, so unconstrained runs do not route.
-  # - delay-measurements/regtomem restore the expected memory form after
-  #   routing.
-  # - The final decomposition converts router-inserted SWAPs and remaining CX
-  #   gates to CZ as the final 2Q gate. Single-qubit gates remain in CUDA-Q's
-  #   decomposition vocabulary: h, rx, ry, rz, x.
+  # Target mid-level stage: inline synthesized helper calls, decompose to a CX
+  # routing basis, prepare wire metadata, optionally route, then lower routed
+  # two-qubit work to CZ. With no device argument, routing is bypassed.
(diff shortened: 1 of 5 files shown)


runtime/cudaq/platform/default/circuit-opt-bench.yml
```diff
@@ -17,17 +17,13 @@ target-arguments:
...
-  # CX gates to CZ as the final 2Q gate.
-  # Inline KAK-decomposed helper functions before decomposition and routing
-  # so the router sees the individual CX gates, not opaque quake.apply ops.
-  jit-mid-level-pipeline: "apply-op-specialization,aggressive-inlining,func.func(expand-control-veqs,multicontrol-decomposition),decomposition{basis=h,rx,ry,rz,x,x(1)},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{basis=h,rx,ry,rz,x,z(1)}"
-  # Cleanup: symbol-dce removes wire_set ops left by add-wireset (prevents
-  # the JIT from taking the wireset-to-profile QIR path).
-  jit-low-level-pipeline: "func.func(apply-control-negations,canonicalize,cse),symbol-dce"
+  # Target high-level stage: materialize registered custom-op matrices and
+  # synthesize them to gates before target lowering.
+  jit-high-level-pipeline: "get-concrete-matrix,unitary-synthesis"
```
schweitzpgi, line 22: Before this pipeline is run, we should always run
```c++
  pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddDeallocs());
  pm.addNestedPass<func::FuncOp>(cudaq::opt::createQuakeAddMetadata());
  pm.addPass(cudaq::opt::createQuakePropagateMetadata());
  pm.addNestedPass<func::FuncOp>(cudaq::opt::createUnwindLowering());
  cudaq::opt::createClassicalOptimizationPipeline(pm, std::nullopt,
                                                  {options.allowEarlyExit});
  pm.addPass(cudaq::opt::createGlobalizeArrayValues());
  pm.addPass(cudaq::opt::createUnitarySynthesis());
  pm.addPass(cudaq::opt::createApplySpecialization(
      {.constantPropagation = options.applyConstProp}));
  cudaq::opt::addAggressiveInlining(pm);
```
as well as some other passes.

runtime/cudaq/platform/default/circuit-opt-bench.yml
```diff
@@ -17,17 +17,13 @@ target-arguments:
...
+  # Target mid-level stage: inline synthesized helper calls, decompose to a CX
+  # routing basis, prepare wire metadata, optionally route, then lower routed
+  # two-qubit work to CZ. With no device argument, routing is bypassed.
+  jit-mid-level-pipeline: "apply-op-specialization,aggressive-inlining,decomposition{basis=h,rx,ry,rz,x,x(1)},func.func(add-dealloc,combine-quantum-alloc,canonicalize,factor-quantum-alloc,memtoreg),add-wireset,func.func(assign-wire-indices),qubit-mapping{device=%DEVICE:bypass%},func.func(delay-measurements,regtomem),decomposition{basis=h,rx,ry,rz,x,z(1)}"
```
schweitzpgi, line 26: Before this pipeline, we'll run
```c++
  cudaq::opt::createClassicalOptimizationPipeline(pm);
  cudaq::opt::addDecomposition(pm, {std::string("U3ToRotations")});
  pm.addNestedPass<func::FuncOp>(cudaq::opt::createMultiControlDecomposition());
```
The mid-level pipeline is where we normally would put custom (per target) gate set mappings, routings, etc. So this is good. Do you really want `delay-measurements`? That's sort of a weird hack pass for the IQM target.

runtime/cudaq/platform/default/circuit-opt-bench.yml
```diff
@@ -17,17 +17,13 @@ target-arguments:
...
+  # Target low-level cleanup stage: remove wire-set symbols left by routing prep.
+  jit-low-level-pipeline: "symbol-dce"
```
schweitzpgi, line 28: And before this pipeline string
```c++
```
So, here we're re-running symbol-dce.

Unfortunately, there is the caveat that instead of simply applying the pipeline as constructed, the pipelines that are strings can be mutated by the launch code in ad hoc ways, which is a bad idea.

schweitzpgi: There is another user-defined pipeline that can be specified to run after this low-level one called post-codegen-passes. I think we can ignore that one and just use low-level for our purposes here.

runtime/cudaq/platform/default/circuit-opt-bench.yml
schweitzpgi: So unitary-synthesis is already part of that pipeline and GetConcreteMatrix realy only matters if you're using user-defined custom ops.

runtime/cudaq/platform/default/circuit-opt-bench.yml
schweitzpgi: Here, apply op specialization and inlining should have already been done in the high-level pipeline. If we're doing them here again and they are actually transforming code in some way, it would be good to have an example and understand what's wrong.

runtime/cudaq/platform/default/circuit-opt-bench.yml
taalexander: I agree it's a part of the default. However, I think I was running into issues with exactly 

> GetConcreteMatrix realy only matters if you're using user-defined custom ops

These are a part of the benchmarks that I was running and this is the magic sauce it took to get them running. I will try again and see if I can remove.

runtime/cudaq/platform/default/circuit-opt-bench.yml
taalexander: > Do you really want delay-measurements? That's sort of a weird hack pass for the IQM target.

I think delay measurements could be removed. I was using IQM to copy from and it snuck in.

> Here, apply op specialization and inlining should have already been done in the high-level pipeline. If we're doing them here again and they are actually transforming code in some way, it would be good to have an example and understand what's wrong.

This ties into the unitary synthesis/concrete matrix above when defining custom ops for the benchmarks. I can spit out what's happening in this benchmark to give a better example.

runtime/cudaq/platform/default/circuit-opt-bench.yml
taalexander: I think the `symbol-dce` was explicitly an issue with running this in estimate resources.  I'm not sure I fully understand your comment here.

schweitzpgi (approved):
Pipelines to be revistied in another follow-up PR.

PR #4433, opened 2026-05-01
[Tensornet] Fix adjoint handling of non-unitary ops

1tnguyen:
Previously, we had a special code handling the adjoint of projector tensors (used to construct tensor network state from state vector data). This turned out to be a workaround for an unknown bug, which has now been fixed.

Hence, remove the workaround code block. Also, guard against the cutensornet fixed version, which should have been updated in https://github.com/NVIDIA/cuda-quantum/pull/4342/.

Also, extend the test to cover more cases beyond the simple bell state.


runtime/nvqir/cutensornet/tn_simulation_state.inc
```diff
@@ -49,30 +49,9 @@ std::complex<double> TensorNetSimulationState<ScalarType>::overlap(
   // Reverse them
   std::reverse(tensorOps.begin(), tensorOps.end());
   std::vector<void *> tempDeviceBuffers;
```
sacpis: Seems like an unused vector?

runtime/nvqir/cutensornet/tn_simulation_state.inc
sacpis: Once we remove unused vector, we won't need this code as well.

sacpis (approved):
LGTM. Thanks @1tnguyen. Left a few comments.

cabreraam (approved):
LGTM

runtime/nvqir/cutensornet/tn_simulation_state.inc
1tnguyen: Indeed, good catch! Removed in https://github.com/NVIDIA/cuda-quantum/pull/4433/changes/55e97f74e74e2f13f95166bab97ad58707ca36c5

runtime/nvqir/cutensornet/tn_simulation_state.inc
1tnguyen: Definitely. Removed in https://github.com/NVIDIA/cuda-quantum/pull/4433/changes/55e97f74e74e2f13f95166bab97ad58707ca36c5

PR #4434, opened 2026-05-01
[Testing] Reduce tensornet test time in explicit measurement tests

1tnguyen:
These dynamics circuits require incremental contractions (in the `n_rounds` loop), hence are very slow on `tensornet` backends.

Use the test macro to reduce the number of rounds/shots for `tensornet` backends. 

With 4 tensor network backends (full tensor network/MPS for single/double precision) x 2 tests, the test time reduction (for developers) is quite significant (~1000 secs)


khalatepradnya (approved):
LGTM. Thanks, @1tnguyen

PR #4409, opened 2026-04-29
[C++] Implement `cudaq::measure_handle`

khalatepradnya:
## Summary

* Implements the C++ half of the `measure_handle` proposal

_NOTE: These all apply in MLIR mode only.
Library mode (`#ifdef CUDAQ_LIBRARY_MODE`) keeps the legacy `class measure_result` block in `runtime/cudaq/qis/execution_manager.h` verbatim, and is source-compatible._

### Type identity

`cudaq::measure_result` is now a `using` alias for `cudaq::measure_handle`. Existing source code that spells `measure_result` continues to compile, but the underlying class changed:
  * `measure_handle` is not implicitly convertible to `bool` at host scope - `operator bool()` throws `std::runtime_error` in MLIR mode. The bridge intercepts every legitimate bool coercion inside `__qpu__` and emits `quake.discriminate`; reaching the host body means a coercion site escaped the bridge.
  * `measure_handle` is not integer-constructible. The legacy `measure_result(uint8_t)` form no longer compiles; use `bool` for host-side bit storage.
  * `decltype(mz(q))` resolves to `measure_handle`, not `bool`. Visible to `auto`, template specialization, and `std::is_same` checks.

### Return type of `mz` / `mx` / `my`

* Scalar form: `mz(q)` / `mx(q)` / `my(q)` return `measure_handle` instead of `bool`.
* Vector form: `mz(qvec)` / `mx(qvec)` / `my(qvec)` return `std::vector<measure_handle>` instead of `std::vector<bool>` (which is what the prior `using measure_result = bool` alias resolved to).

### Required source migrations

* `__qpu__ auto kernel() { return mz(...); }` - `auto` now deduces `measure_handle` (or `std::vector<measure_handle>`), which the host-device boundary rejects (see below). Spell the return type explicitly as `bool` / `std::vector<bool>`; for the vector case, also wrap with `cudaq::to_bools(...)`.
* `cudaq::to_integer(mz(qvec))` - the bridge rejects this shim with a  spec-named frontend diagnostic.  The new
  `cudaq::to_bools(std::vector<measure_handle>)` API is the spec-mandated bulk discrimination surface.
* Host-scope `mz` / `mx` / `my` -calling these outside a `__qpu__` kernel now throws `std::runtime_error` instead of silently returning a meaningless `bool`. Move the call into a kernel.

### New compile-time errors

[...]

PR #4439, opened 2026-05-02
[Python] Measurement handle type

khalatepradnya:
### Summary

* Python-side counterpart to PR #4409. 
* Binds `cudaq.measure_handle`, rewires the Python AST bridge so `mz`/`mx`/`my` emit handles directly, lowers every Python bool-coercion site to `quake.discriminate`, surfaces `cudaq.to_bools` for bulk discrimination, and rejects `measure_handle` in entry-point kernel signatures with the spec-canonical diagnostic. 

### What Changed

- `cudaq.measure_handle` is a kernel-only stub that raises on host instantiation; `!cc.measure_handle` registered as a Python-visible MLIR type.
- `mz`/`mx`/`my` produce `!cc.measure_handle` (scalar) or `!cc.stdvec<!cc.measure_handle>` (vector); `__discriminateIfMeasureHandle` is inserted at every bool-coercion site (`if`/`while`/`not`/`bool()`/`==`/`!=`/`return`, `and`/`or` short-circuit RHS, list-of-bool element stores via `i8` storage); `cudaq.to_bools` lowers to vectorized `quake.discriminate`
- entry-point Python kernels with `measure_handle` (direct or transitive) in the signature get the spec-canonical 'measure_handle cannot cross the host-device boundary; entry-point kernels must discriminate first' diagnostic, mirroring the C++ wording for cross-frontend consistency.

### Breaking Changes

- Python kernels that relied on `mz` returning `bool` directly (e.g., `if mz(q):` used to type-check as `bool`; now type-checks as `measure_handle` and is implicitly discriminated by the new bridge path). User-visible behavior in the common case is unchanged, but deeper integrations that introspect the return type will see the new class.
- `cudaq.to_integer(list[cudaq.measure_handle])` is now a hard rejection. Calls of the form `cudaq.to_integer(mz(qv))` previously read as an integer projection of the bit vector via implicit discrimination; the diagnostic is now `cudaq.to_integer` does not accept a `list[cudaq.measure_handle]`; compose with `cudaq.to_integer(cudaq.to_bools(handles))`.


python/cudaq/kernel/ast_bridge.py
```diff
@@ -4828,6 +4974,13 @@ def compare_equality(item1, item2):
 
             # TODO: the In/NotIn case should be recursive such that we can
             # search for a list in a list of lists.
+            # `mz(q1) == mz(q2)` (or with one side being a handle and the
+            # other a `bool`): discriminate each handle to `i1` first and let
+            # `convert_arithmetic_types` finish the integer comparison.
+            if cc.MeasureHandleType.isinstance(item1.type):
+                item1 = self.__discriminateIfMeasureHandle(item1, node)
+            if cc.MeasureHandleType.isinstance(item2.type):
+                item2 = self.__discriminateIfMeasureHandle(item2, node)
```
sacpis: Actually here, we can directly call `__discriminateIfMeasureHandle` as it already handles `cc.MeasureHandleType.isinstance` in it.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -5171,6 +5348,15 @@ def __process_binary_op(self, left, right, nodeType):
...
         """
 
+        # `measure_handle` operands in arithmetic context discriminate
+        # through `bool` first. Pre-discriminating here lets
+        # `__get_superior_type` and the `isArithmeticType` check below treat
+        # handles uniformly with any other `i1`-valued operand.
+        if cc.MeasureHandleType.isinstance(left.type):
+            left = self.__discriminateIfMeasureHandle(left, self.currentNode)
+        if cc.MeasureHandleType.isinstance(right.type):
+            right = self.__discriminateIfMeasureHandle(right, self.currentNode)
```
sacpis: same here.

python/cudaq/kernel/ast_bridge.py
```diff
@@ -618,6 +623,24 @@ def changeOperandToType(self, ty, operand, allowDemotion=False):
...
+        # discriminate. `i8` also covers element stores into a `list[bool]`,
+        # which the bridge backs with `i8` storage but logically owns as bool.
+        if cc.MeasureHandleType.isinstance(
+                operand.type) and (IntegerType.isinstance(ty) or
+                                   F32Type.isinstance(ty) or
+                                   F64Type.isinstance(ty)):
+            disc = self.__discriminateIfMeasureHandle(operand, self.currentNode)
+            return self.changeOperandToType(ty, disc, allowDemotion)
+        if (cc.StdvecType.isinstance(ty) and
+                cc.StdvecType.getElementType(ty) == self.getIntegerType(1) and
```
sacpis: nit: `self.getIntegerType(1)` can be saved in a local variable. We can save a couple of attribute lookups and method calls per call site,

python/cudaq/kernel/kernel_builder.py
```diff
@@ -1197,22 +1202,16 @@ def mx(self, target, regName=None):
...
                 res = quake.MxOp(measTy, [], [target.mlirValue],
                                  registerName=StringAttr.get(regName,
                                                              context=self.ctx))
             else:
                 res = quake.MxOp(measTy, [], [target.mlirValue])
-            disc = quake.DiscriminateOp(retTy, res)
-            return self.__createQuakeValue(disc.result)
+            return self.__createQuakeValue(res.measOut)
 
     def my(self, target, regName=None):
```
sacpis, line 1212: nit: Seems like `mx`, `my`, and `mz` are almost identical. We can pull out some common functionality in a helper method.

sacpis (approved):

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Fixed in 63c4e95

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Fixed in 63c4e95

python/cudaq/kernel/kernel_builder.py
khalatepradnya: Fixed in 63c4e95

python/cudaq/kernel/ast_bridge.py
khalatepradnya: Fixed in 63c4e95

Issue #4441, opened 2026-05-02
Publishing failure test is failing on CuPy


sacpis:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

Following test

test_cupy_to_state_ownership_semantics

is failing with

```
E               AssertionError: Contiguous CuPy path is expected to alias source memory.
E               assert np.complex128(1+1j) == (999 + 0j)
```

https://github.com/NVIDIA/cuda-quantum/actions/runs/25252920938/job/74051805635#step:7:22525
https://github.com/NVIDIA/cuda-quantum/actions/runs/25252920938/job/74051805645#step:7:13399

### Steps to reproduce the bug

Occurring in CI only

### Expected behavior

Should pass successfully.

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: 
- **Python version**: 
- **C++ compiler**: 
- **Operating system**: 

### Suggestions

@huaweil-nv Mind taking a look at it?

huaweil-nv:
Sorry about this. After further analysis, I confirmed that the failure is due to an issue in the test itself. Quick root cause and fix below.
**Root cause:**
The assertion at fault is the "contiguous CuPy aliases source memory" half of test_cupy_to_state_ownership_semantics:
```
  contig = cp.array([1+1j, 2+2j, 3+3j, 4+4j], dtype=cp.complex128)
  s_contig = cudaq.State.from_data(contig)
  contig[0] = 999 + 0j
  aliased = np.array(s_contig)
  assert aliased[0] == 999 + 0j  # <-- this
```
[...]

PR #4442, opened 2026-05-03
[tests] Drop brittle contiguous-aliasing assertion in CuPy ownership …

huaweil-nv:
Issue #4441: test_cupy_to_state_ownership_semantics fails on the CI's NVIDIA/CUDA stack because it asserts that the contiguous CuPy fast path aliases the source device buffer (mutating contig[0] should be visible through np.array(state)). That assertion was written from a local observation, not a contract — whether CusvState adopts the user-supplied device pointer or eagerly cudaMemcpy's it into its own pool is a backend implementation detail that varies with cupy version, CUDA runtime version, and memory pool configuration.

Drop the contiguous half of the test and keep only the strided half, renaming it to test_cupy_to_state_strided_canonicalization_is_independent to reflect what is actually being asserted. Independence is part of the strided-import fix's contract because we explicitly canonicalize non-contiguous CuPy input through cupy.asnumpy(), which always returns a host copy.

### DIFF
--- a/python/tests/builder/test_cupy_integration.py
+++ b/python/tests/builder/test_cupy_integration.py
@@ -145,35 +145,28 @@ def test_cupy_to_state_strided_1d():
     cudaq.reset_target()
 
 
-def test_cupy_to_state_ownership_semantics():
-    """Document (and lock in) the ownership contract exposed by `from_data`:
+def test_cupy_to_state_strided_canonicalization_is_independent():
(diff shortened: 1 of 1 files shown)


huaweil-nv:
> /ok to test 1d4e5b0511a07d81d27be2fd164912f057e17d41

sacpis (approved):
LGTM. Thanks @huaweil-nv for the quick fix.

Issue #4359, opened 2026-04-21
`make_kernel` segfaults when called concurrently from async dispatch loop; needs guard or thread-safety


khalatepradnya:
### Context

Spun off from #2821. When `cudaq.make_kernel()` is called inside a loop that also dispatches `observe_async` / `sample_async`, the MLIR context is accessed from multiple threads without synchronization, producing a bare segfault with no actionable error message.

The workaround (build the kernel before the loop) was confirmed by the reporter, but the failure mode — a silent segfault inside `kernel_builder.py` → `_func_ops_ext.py` — is hostile to users.

### Reproducer

```python
import cudaq

cudaq.set_target("nvidia", option="mqpu")

def build_kernel():
    k, params = cudaq.make_kernel(list[float])
    qubits = k.qalloc(5)
    for i in range(5):
        k.rx(params[i], qubits[i])
    return k

# This segfaults because make_kernel is not thread-safe
for i in range(100):
    ker = build_kernel()  # <-- called inside async dispatch
    cudaq.sample_async(ker, [0.1]*5, qpu_id=i % 4)
```

### Expected behavior
Either:

1. Preferred: make_kernel() is thread-safe and works correctly when called concurrently, or
2. Minimum: A clear RuntimeError is raised (e.g., "Kernel construction is not thread-safe; build kernels before dispatching async work") instead of a segfault.

### Additionally
The thread-safety constraint should be documented in the [multi-GPU workflows guide](https://nvidia.github.io/cuda-quantum/latest/using/examples/multi_gpu_workflows.html) and in the make_kernel API docstring.

### Environment
Originally reported on CUDA-Q 0.10.0
Workaround confirmed on 0.10.0+

### Related
#2821 (original report, being closed with workaround)

PR #4443, opened 2026-05-03
Prevent dynamic kernel construction during pending async work

shreyaskommuri:
## Summary

Fixes the minimum failure mode from #4359 by guarding \`cudaq.make_kernel()\` while async \`sample_async\` or \`observe_async\` work is pending. Instead of allowing unsafe concurrent access to the process-wide MLIR context and risking a segfault, CUDA-Q now raises an actionable \`RuntimeError\`.

This does not make dynamic kernel construction fully thread-safe; it documents and enforces the current constraint.

## Changes

**\`python/cudaq/kernel/utils.py\`**
- Adds \`_active_async_work_count\` (protected by \`threading.Lock\`) and three helpers: \`register_async_work\`, \`unregister_async_work\`, \`check_no_active_async_work\`.

**\`python/cudaq/kernel/kernel_builder.py\`**
- \`make_kernel()\` calls \`check_no_active_async_work()\` before touching the global MLIR context.
- Adds a constraint note to the docstring.

**\`python/cudaq/runtime/sample.py\`**
- \`AsyncSampleResult\` now accepts \`async_work_registered\` to avoid double-counting when constructed from \`sample_async\`.
- \`get()\` unregisters async work in a \`finally\` block (correct even when \`impl.get()\` throws).
- \`__del__\` unregisters async work when the result is discarded without calling \`get()\`, preventing the counter from leaking and permanently blocking \`make_kernel()\`.
- \`sample_async\` registers before the async launch and unregisters on exception.

**\`python/cudaq/runtime/observe.py\`**
- Adds \`_AsyncObserveResult\` wrapping the C++ future with the same register/unregister lifecycle as \`AsyncSampleResult\`, including \`__del__\`.
- \`observe_async\` now returns \`_AsyncObserveResult\` instead of the raw C++ future.

**\`docs/sphinx/using/examples/multi_gpu_workflows.rst\`**
- Adds a \`.. note::\` documenting the build-before-dispatch constraint for \`make_kernel\`.

## Testing

- \`python3 -m py_compile\` across all changed Python files — passes.
- \`git diff --check\` — no whitespace errors.
- Two new unit tests in \`python/tests/builder/test_kernel_builder.py\`:
  - \`test_make_kernel_rejects_pending_async_work\` — verifies \`make_kernel\` raises while an \`AsyncSampleResult\` is live, then succeeds after \`get()\`.
  - \`test_make_kernel_allowed_after_gc_of_async_result\` — verifies \`make_kernel\` succeeds after the result is dropped (\`del\`) without calling \`get()\`, exercising the \`__del__\` path.

## Not tested

- Targeted pytest, because local Python 3.14 environment does not have \`pytest\` installed.
- GPU/mqpu reproducer, because this host is macOS CPU-only.


shreyaskommuri:
## Follow-up fixes on top of the original guard

After reviewing the initial implementation, I found two correctness bugs and one missing test case. These are now fixed in the same commit.

---

### Bug 1 — \`AsyncSampleResult.__del__\` leaked the async-work count

**Problem:** If a caller discards an \`AsyncSampleResult\` without ever calling \`.get()\` (e.g. fire-and-forget, or an exception skips the call), \`_active_async_work_count\` stays > 0 permanently. Every subsequent \`cudaq.make_kernel()\` call in that process raises \`RuntimeError\`, which is worse than the segfault the guard was meant to prevent.

**Fix:** \`__del__\` now calls \`unregister_async_work()\` when \`_async_work_registered\` is still \`True\`. \`unregister_async_work\` already guards against going below zero, so a double-decrement is impossible.

---

### Bug 2 — \`_AsyncObserveResult\` had no \`__del__\`

**Problem:** Same permanent-blockage issue on the observe path. The new \`_AsyncObserveResult\` class had \`get()\` correctly unregistering via \`finally\`, but nothing handled the GC path.

**Fix:** Added \`__del__\` with the same pattern as the sample fix.

---

### New test — GC path

Added \`test_make_kernel_allowed_after_gc_of_async_result\` to \`test_kernel_builder.py\`. The existing \`test_make_kernel_rejects_pending_async_work\` only exercised the \`.get()\` path; this new test uses \`del future\` to exercise \`__del__\` directly.

PR #4445, opened 2026-05-04
Improve header isolation and parse time of logger.h

Renaud-K:
As we prepare to surface qpu headers in user code,  `cudaq.h` is become significantly slower to parse in a prototyping sandbox. This PR addresses a slowdown introduced by logger.h as can be seen in the profile below.

The compiled code is:
```cpp
#include <cudaq.h>

int main() {
return 0;
}
```
The compile time is 3.6s.
```
time clang++ cudaq_inc.cpp -std=c++20  -I ~/cudaq/cq2/install/cudaq/include
real    0m3.603s
```

The profile shows logger.h takes 441ms of parse time.

logger-profile-1

With this patch, the compilation time becomes 2.9s
```
time clang++ cudaq_inc.cpp -std=c++20  -I ~/cudaq/cq2/install/cudaq/include
real    0m2.936s
```

The profile show logger.h only taking 7ms.

logger-profile-2

This is achieved by replacing the `std::variant`  with a `FormatArgument` which only stores a pointer and an out-of-line appending callback. The callback is instantiated in logger.cpp.


Renaud-K:
> /ok to test b2a36a747c561cb2c92f8bcb4586bfcfd6bc13d7

Renaud-K:
> /ok to test ded40b51d1809b8d4b891309519f08752d636c78

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
```diff
@@ -8,120 +8,82 @@
...
 //
 // Functions substitutes for fmt::format and fmt::print
 //
 template <typename... Args>
 std::string format(const std::string_view message, Args &&...args) {
-  auto array = std::array<fmt_arg, sizeof...(Args)>{
-      fmt_arg(std::forward<Args>(args))...};
-  return details::format_packed(message, array);
+  auto array = std::array<FormatArgument, sizeof...(Args)>{
+      FormatArgument(std::forward<Args>(args))...};
```
sacpis, line 69: Here and on line 84, instead of `std::forward` can we use `static_cast<Args&&>(args)...` and drop the `utility` include?

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
```diff
@@ -8,120 +8,82 @@
...
-
-//
-// Packed versions of format and print, implemented in logger.cpp
-//
-std::string format_packed(const std::string_view message,
-                          const std::span<fmt_arg> &arr);
+  template <typename T>
+  FormatArgument(const T &value)
+      : value(std::addressof(value)), append(&appendArgument<std::decay_t<T>>) {
+  }
```
sacpis, line 42: Should we add a `static_assert` so unsupported types fail at compile time?

runtime/logger/logger.cpp
```diff
@@ -125,36 +131,138 @@ std::string pathToFileName(const std::string_view fullFilePath) {
...
+  case LogLevel::info:
+    info(msg);
+    return;
+  case LogLevel::warn:
+    warn(msg);
+    return;
+  case LogLevel::error:
+    error(msg);
+    return;
+  }
```
sacpis, line 170: Shall we add a default case here?

runtime/logger/logger.cpp
```diff
@@ -125,36 +131,138 @@ std::string pathToFileName(const std::string_view fullFilePath) {
   const std::filesystem::path file(fullFilePath);
   return file.filename().string();
 }
+
+void logMessagePacked(LogLevel logLevel, const std::string_view message,
+                      const std::span<const cudaq_fmt::FormatArgument> &args,
+                      const char *funcName, const char *fileName, int lineNo) {
+  (void)funcName;
```
sacpis: nit: Should we delete it as we are not using it in this function?

runtime/include/cudaq/runtime/logger/logger.h
```diff
@@ -29,34 +37,46 @@ void debug(const std::string_view msg);
 void warn(const std::string_view msg);
 void error(const std::string_view msg);
 std::string pathToFileName(const std::string_view fullFilePath);
+
+void logMessagePacked(LogLevel logLevel, const std::string_view message,
+                      const std::span<const cudaq_fmt::FormatArgument> &args,
+                      const char *funcName, const char *fileName, int lineNo);
```
sacpis: nit: are we using `funcName` in the definition? If not, should we remove it?

runtime/include/cudaq/runtime/logger/logger.h
```diff
@@ -29,34 +37,46 @@ void debug(const std::string_view msg);
...
+    const std::span<const cudaq_fmt::FormatArgument> &args);
+
+template <typename... Args>
+std::string formatMessage(const std::string_view message, Args &&...args) {
+  return cudaq_fmt::format(message, std::forward<Args>(args)...);
+}
+
+template <typename... Args>
+void logMessage(LogLevel logLevel, const std::string_view message,
+                const char *funcName, const char *fileName, int lineNo,
```
sacpis: nit: `funcName` is not being used.

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
```diff
@@ -8,120 +8,82 @@
...
-      std::reference_wrapper<const std::vector<float>>,
-      std::reference_wrapper<const std::vector<double>>,
-      std::reference_wrapper<const std::vector<std::string>>,
-      std::reference_wrapper<const std::map<std::string, std::string>>,
-      std::reference_wrapper<const std::vector<std::complex<float>>>,
-      std::reference_wrapper<const std::vector<std::complex<double>>>>;
-  storage_t value;
+  template <std::size_t N>
+  FormatArgument(const char (&value)[N])
+      : value(value), append(&appendCString) {}
```
sacpis: Wondering if we are using this overloaded function?

runtime/logger/logger.cpp
```diff
@@ -125,36 +131,138 @@ std::string pathToFileName(const std::string_view fullFilePath) {
...
+    return;
+  }
+}
+
+void logWithTimestampPacked(
+    const std::string_view message,
+    const std::span<const cudaq_fmt::FormatArgument> &args) {
+  const auto timestamp = std::chrono::system_clock::now();
+  const auto now_c = std::chrono::system_clock::to_time_t(timestamp);
+  std::tm now_tm = *std::localtime(&now_c);
```
sacpis: `cudaq::log()` is callable from any thread. `std::localtime` writes to a shared static tm and races. Seems like a data bug?
Should we do

```
std::tm now_tm{};
localtime_r(&now_c, &now_tm);
```

note: I think it is `localtime_s` for Windows, please double check.

runtime/logger/logger.cpp
sacpis, line 126: nit: Are we missing static_assert on `LogLevel::error`?

sacpis (approved):

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
Renaud-K: You cannot prevent `utility` from being included. `cudaq_fmt.h` comes from `cudaq.h`. If not included in this header it will be included from other places. It is unavoidable.

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
Renaud-K: We would have to bring back the headers we were trying to avoid in the first place. Yes, we are trading off compile time errors for speed and replacing them with link time errors.

runtime/logger/logger.cpp
Renaud-K: We are and want to handle all the cases. clang will error if we miss to handle a case. With a default case, it no longer does and will not warn us about new unhandled cases. I think it is better not to have a default.

runtime/logger/logger.cpp
Renaud-K: That's right. And it seems we have had this bug for awhile. `_r` is the POSIX version.

runtime/include/cudaq/runtime/logger/logger.h
Renaud-K: I have been reluctant to remove it because it seems like we went to quite an extend to print it but just fall short of doing it. I am not sure if we were planning on adding it eventually.

runtime/logger/logger.cpp
Renaud-K: This is the same comment about funcName. It is passed through but ultimately unused. Like I said, I have been reluctant to remove it altogether considering that we seem to have gone to some extend to add it.

runtime/logger/logger.cpp
Renaud-K: We are. :)

runtime/logger/logger.cpp
sacpis: Thanks @Renaud-K. I think if it not being used we can remove it. @khalatepradnya Do you recall why we added this funcName here?

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
Renaud-K: We are in the case of the string literals. Their types are char[N]. 
```
cudaq_fmt::format("hello")
```
Now, there is a bit of a battle of the AIs here. Because one reviewer agent told me:
```
When a string literal (e.g. `"hello"`, type `const char[N]`) is passed as a format argument, the unconstrained template constructor `FormatArgument(const T&)` wins over the non-template `FormatArgument(const char*)` because binding a reference is a better implicit conversion sequence than array-to-pointer decay. This causes `T` to deduce as `const char[N]` and `std::decay_t<T>` as `const char*`,..., it would interpret the raw string bytes as a pointer (UB).
```

But then another agent says the `char *` path would be taken, digging deeper in the C++ standard. It even built a testcase to prove it. 

I can remove the array overloads because the string literals would fallback into the proper non template overloads. But we will have to able to argue our case with AI later.

Maybe, what I am finding out, is that what is really missing here are tests. I will add those.

sacpis: Yes, I wouldn't rely on AI agents. The basic test would be to check if this path is hit. I didn't see any code hitting this. Or may have overlooked it. So a test would be sufficient if this overloaded function is used by any code.

Renaud-K: There is plenty of code hitting this. All string literal would hit this.

runtime/include/cudaq/runtime/logger/logger.h
Renaud-K: I have removed it.

Renaud-K:
> /ok to test ad446480297047fa6073ebc1a0e8aa47b484ca64

taalexander (commented):
I didn't do a detailed review of the code, but the overall high-level approach looks like a nice improvement for performance.

Renaud-K:
> /ok to test ffdba0009e72cc42013b6066ba71dbb2addf282b

runtime/include/cudaq/runtime/logger/cudaq_fmt.h
```diff
@@ -8,120 +8,75 @@
...
 #include <type_traits>
-#include <variant>
-#include <vector>
+#include <utility>
 
 //
-// This header provides a wrapper around some of the fmtlib functions, so that
-// fmtlib headers, which are not distributed, do not bleed into user code.
+// This header provides a wrapper around some of the `fmtlib` functions, so that
+// `fmtlib` headers, which are not distributed, do not leak into user code.
```
schweitzpgi: Maybe propagate instead of leak (which has negative slang connotations)?

runtime/include/cudaq/runtime/logger/logger.h
```diff
@@ -11,13 +11,21 @@
 #include "cudaq/runtime/logger/cudaq_fmt.h"
 #include "cudaq/runtime/logger/tracer.h"
```
schweitzpgi: nit: no blank lines between `#include` lines.

runtime/include/cudaq/runtime/logger/logger.h
```diff
@@ -29,34 +37,53 @@ void debug(const std::string_view msg);
 void warn(const std::string_view msg);
 void error(const std::string_view msg);
 std::string pathToFileName(const std::string_view fullFilePath);
+
+// Test/debug helpers. Production callers configure the level via the
+// CUDAQ_LOG_LEVEL environment variable.
+void setLogLevel(LogLevel level);
+LogLevel getLogLevel();
+// Flushes any buffered log output. Useful in tests that need to inspect
```
schweitzpgi, line 45:
```suggestion

// Flushes any buffered log output. Useful in tests that need to inspect
```

runtime/logger/logger.cpp
```diff
@@ -11,14 +11,20 @@
...
+#include <functional>
+#include <map>
 #include <set>
 #include <spdlog/cfg/env.h>
 #include <spdlog/cfg/helpers.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/rotating_file_sink.h>
 #include <spdlog/spdlog.h>
 #include <sstream>
+#include <vector>
```
schweitzpgi, line 26: Are all these headers actually needed?

Our rule is to include the minimum number of headers without exceptions and carve outs like the "except we put standard headers in every file". Due to their transitive nature, it's not always necessary to have to include all the C++ standard headers.

I'm also asking because the idea of this PR is, from the description, to reduce the time to compile these headers, and redundantly including header files when not required actually increases the time.

runtime/logger/logger.cpp
```diff
@@ -121,40 +127,150 @@ static_assert(static_cast<int>(LogLevel::warn) ==
 bool should_log(const LogLevel logLevel) {
   return spdlog::should_log(static_cast<spdlog::level::level_enum>(logLevel));
 }
+void setLogLevel(LogLevel level) {
+  spdlog::set_level(static_cast<spdlog::level::level_enum>(level));
+}
+LogLevel getLogLevel() { return static_cast<LogLevel>(spdlog::get_level()); }
+void flushLogs() {
```
schweitzpgi, line 137:
```suggestion

LogLevel getLogLevel() { return static_cast<LogLevel>(spdlog::get_level()); }

void flushLogs() {
```

runtime/logger/logger.cpp
```diff
@@ -121,40 +127,150 @@ static_assert(static_cast<int>(LogLevel::warn) ==
...
+CUDAQ_INSTANTIATE_FORMAT_ARGUMENT(const void *);
+CUDAQ_INSTANTIATE_FORMAT_ARGUMENT(std::chrono::milliseconds);
+CUDAQ_INSTANTIATE_FORMAT_ARGUMENT(std::chrono::system_clock::time_point);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::string);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<int>);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<unsigned int>);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<long>);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<unsigned long>);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<long long>);
+CUDAQ_INSTANTIATE_FORMAT_REF_ARGUMENT(std::vector<unsigned long long>);
```
schweitzpgi, line 253: Would it make sense to use `std::int32_t`, `std::uint32_t`, `std::int64_t`, and `std::uint64_t` here instead of the overlapping types of `int`, `long`, and `long long`?

runtime/logger/logger.cpp
Renaud-K: These headers are in a `.cpp` file. They do not impact `cudaq.h`. The idea here and #4469 is to remove unnecessary headers from **headers** and move them into `.cpp` files. I would not remove a standard header on account that it might be already included from other headers. It is not portable as can be seen here: 7933dccc8cbd5b750a71f4033a55259fc662164d

runtime/logger/logger.cpp
Renaud-K: Sure. I can make this change.

Renaud-K:
> /ok to test 008d90bd85937de18cd61d7bd1a5f2f3428fb972

Renaud-K:
> /ok to test abf349bf5a9bb37b1a426df9840e99befebef611

Renaud-K:
> /ok to test a3fc133031631980e2db2c5fe7176f30479f9da1

PR #4446, opened 2026-05-04
Introduce SourceModule to unify kernel definition types

lmondada:
NOTE: This is a re-post of #4413, which I merged into the wrong branch! It's already been reviewed, discussed and approved.

---

Stacks on top of #4398

PR #4398 made the first of two steps towards homogenizing all launch endpoints into a same function signature by introducing `KernelArgs`.

This PR introduces a `SourceModule` type that stores the kernels themselves in different formats depending on the host language:
- a C++ kernel is stored as a name + function pointer. The name is used in the runtime to retrieve the Quake representation of the kernel if required. The function pointer points to the function that was compiled by nvq++ and is used for local simulation.
- a Python kernel is stored as a name + MLIR ModuleOp. Both local and remote executions use the MLIR ModuleOp as the source of truth for the kernel definition. If the kernel launch must be executed locally, the MLIR will be compiled and JITed, otherwise it will be compiled and submitted to the remote endpoint.

Using this type, we can effectively hide ALL differences between the host languages and the various launching conventions behind a homogeneous API.

Resulting signatures:

```c++
  launchKernel(const SourceModule &src, KernelArgs args);

  launchModule(const CompiledModule &compiled, KernelArgs args);

  [[nodiscard]] virtual CompiledModule
  compileModule(const SourceModule &src, KernelArgs args, bool isEntryPoint);
```

Note that for Python, the current execution is broken up into `compileModule` -> `launchModule`, whereas for C++, both compile and launch steps are still in one -- hence the signature difference. Changing the C++ launch path to mirror the Python one is planned upcoming work.


Renaud-K (approved):

Renaud-K (changes requested):
If all these changes are leading to #4447, then I don't see how we are unifying the launchKernel API.  Introducing `SourceModule` only emphasizes their difference.

I thought we were going to move these concerns into CompiledModule so we could have:
```
  launchKernel(const CompiledModule &src, KernelArgs args);

  launchModule(const CompiledModule &compiled, KernelArgs args);
```
and ultimately:

```
  launchKernel(const CompiledModule &src, KernelArgs args);
```

Renaud-K (approved):
Looking good. Thank you!

PR #4447, opened 2026-05-04
Introduce unifiedLaunchModule dispatcher

lmondada:
Stacks on top of #4398 and #4446.

This PR replaces all uses of launch endpoints (launchKernel and launchModule) with a single `unifiedLaunchModule`.

This unified launch is temporary: it dispatches to either `launchKernel` or `launchModule`. The rest of the unification can proceed behind the scenes, decoupled from API changes.

## How `launchKernel` and `launchModule` differ

`launchModule` is used for Python kernels. This launch path has already been refactored to split out compilation from execution. `launchModule` thus gets passed a `CompiledModule` as argument, generated and passed to it by `compileModule`.

`launchKernel` on the other hand is used for C++ kernels. This launch path has not yet been refactored, and hence both compilation and kernel execution happen within the same function. As a result, `launchKernel` expects an uncompiled `SourceModule` as argument.

To bridge this gap, a variant `AnyModule` is introduced in this PR that can represent either a `CompiledModule` or a `SourceModule`. Once compilation has been refactored out of `launchKernel`, its arguments will be modified to match `launchModule` and thus this dispatcher and variant type will become redundant.


runtime/cudaq/platform/qpu.cpp
```diff
@@ -54,28 +56,38 @@ launchCompiledModule(const cudaq::CompiledModule &compiled,
...
+  return launchCompiledModule(module, args);
 }
 
-cudaq::CompiledModule
-cudaq::QPU::compileModule(const std::string &name, const void *modulePtr,
-                          const std::vector<void *> &rawArgs,
-                          bool isEntryPoint) {
+cudaq::KernelThunkResultType
+cudaq::QPU::unifiedLaunchModule(const AnyModule &module,
+                                const KernelArgs &args) {
```
Renaud-K, line 72: I am not sure what this buys us. We could have used a trick like this 2 weeks ago by just passing a default constructed ModuleOp in the C++ case.

Renaud-K:
>Once compilation has been refactored out of launchKernel, its arguments will be modified to match launchModule and thus this dispatcher and variant type will become redundant.

I am concerned by this statement. The idea of the work was not to have a dispatcher but a single API to enable policy work and compiler refactoring to happen in parallel.

PR #4451, opened 2026-05-05
Re-enable gcc12 toolchain

atgeller:
cuda 12.6 doesn't work with clang++ 22.1. Re-enable gcc12 toolchain support to work around this.


mitchdz (approved):
This looks fine to me. Let's get it in to unblock Deployments.

PR #4454, opened 2026-05-05
Suppress false positive std::swap warnings in matrix files

atgeller:


schweitzpgi (approved):

sacpis:
Wondering where is `std::swap` as mentioned in the PR title?

Issue #4407, opened 2026-04-28
Quantum memory management within CUDA-Q kernels does not work


renezander90:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [x] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

When **sequentially executing a sub-kernel** that allocates a local cudaq.qvector, the quantum memory management fails to actually free and reuse the allocated qubits after the sub-kernel returns. This causes total qubit usage to scale linearly with the number of calls, eventually causing a crash due to qubit exhaustion/OOM.

Crucially, inspecting the MLIR confirms that the frontend correctly inserts the quake.dealloc instruction. The issue appears to be that the execution runtime or simulator backend is ignoring this instruction or failing to return the qubits to the available pool.

Please note that many quantum algorithms will also require memory management **within a single kernel** (e.g., for allocating ancillary qubits for an mcx gate, and deallocating them afterwards https://www.iccs-meeting.org/archive/iccs2022/papers/133530169.pdf). But, this might rather be a feature request.

### Steps to reproduce the bug

```
import cudaq

@cudaq.kernel
def test_kernel() -> int:
    q = cudaq.qvector(10)
    h(q[0])
    return mz(q[0])

@cudaq.kernel
def main_kernel() -> int:

    a = test_kernel()
    b = test_kernel()
    c = test_kernel()
    return a + b + c

# print(test_kernel)
cudaq.run(main_kernel, shots_count=10)
```

This will cause a crash due to excessive memory usage. (It will work for 1 or 2 calls of ``test_kernel``).

### Expected behavior

The quantum resources go out of scope at the end of each sub-kernel. The program contains a ``quake.dealloc`` statement:

```
    %0 = quake.alloca !quake.veq<10>
    %1 = quake.extract_ref %0[0] : (!quake.veq<10>) -> !quake.ref
    quake.h %1 : (!quake.ref) -> ()
    %2 = quake.extract_ref %0[0] : (!quake.veq<10>) -> !quake.ref
[...]

PR #4455, opened 2026-05-05
Fix for issue #4407.

schweitzpgi:
These changes make sure that inlined functions have their scopes preserved through the inlining process. Doing this facilitates keeping track of live ranges from variables declared in the called function's body, which allows for more precise allocation and deallocation of qubits in simulation, etc.

Add python regression test.


1tnguyen (approved):
LGTM 👍

schweitzpgi:
Depends on #4474

lib/Optimizer/Dialect/CC/CCOps.cpp
```diff
@@ -1327,13 +1327,23 @@ struct CollapseCastToStdvecInit
...
+            if (!initLen)
+              return arrSize;
+            if (auto optInt = cudaq::opt::factory::getIntIfConstant(initLen))
+              return *optInt;
+            return arrSize + 1; // do not replace this one
+          }();
+          if (!initLen || arrSize == initSize)
             rewriter.replaceOpWithNewOp<cudaq::cc::StdvecInitOp>(
                 init, init.getType(), castVal);
-          return success();
```
sacpis: Seems like the tests in macOS are failing because after replacing the op the code falls through to return failure.
We need
```
if (!initLen || arrSize == initSize) {
    rewriter.replaceOpWithNewOp<cudaq::cc::StdvecInitOp>(init, init.getType(), castVal);
    return success();
}
```

python/tests/regression/test_4407.py
```diff
@@ -0,0 +1,37 @@
...
+    cudaq.__clearKernelRegistries()
+
+
+# Qubit scoping test from
+# https://github.com/NVIDIA/cuda-quantum/issues/4407
+def test_issue_4407():
+
+    @cudaq.kernel
+    def test_kernel() -> int:
+        q = cudaq.qvector(5)
```
renezander90, line 25: `cudaq.qvector(5)` is too small for the test. Calling this three times will result in 15 qubits, which is still easy to simulate if memory management does not work. `cudaq.qvector(10)` (as in the example from #4407) will result in 30 qubits which would be challenging.

schweitzpgi: Understood. We don't add tests to the regressions that may potentially exhaust the server's resources or crash when run with dozens of other tests in parallel. Feel free to add your own tests downstream however.

PR #4458, opened 2026-05-05
fixing auto-launch to bail out fast when cudaq-qpud dies

sacpis:
- If the launched server exits before becoming reachable, waitpid(WNOHANG) breaks out of the 50s ping loop so we move to the next port immediately, instead of throwing `RuntimeError: No usable ports available` only after few minutes.
- Dropping static from the mt19937 so seed_offset is honoured on every construction, not just the first one.

Fixes: 

```
 @pytest.fixture(scope="session", autouse=True)
    def startUpMockServer():
>       cudaq.set_target("remote-mqpu", auto_launch=str(num_qpus))
E       RuntimeError: No usable ports available

tmp/tests/remote/test_remote_platform.py:71: RuntimeError
```

https://github.com/NVIDIA/cuda-quantum/actions/runs/25393104770/job/74490650292#step:7:1282

### DIFF
--- a/runtime/cudaq/platform/mqpu/helpers/MQPUUtils.cpp
+++ b/runtime/cudaq/platform/mqpu/helpers/MQPUUtils.cpp
@@ -23,6 +23,7 @@
 #include <set>
 #include <signal.h>
 #include <sys/socket.h>
+#include <sys/wait.h>
 #include <thread>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


atgeller (approved):
LGTM, thanks @sacpis!

1tnguyen (approved):
👍

PR #4462, opened 2026-05-06
[DNM] Debugging remote platform test for amd64

sacpis:
Debugging remote platform test for amd64


sacpis:
Closing this as we will be disabling this test as we have a PR for remote QPU [here](https://github.com/NVIDIA/cuda-quantum/pull/4276).

PR #4464, opened 2026-05-06
[python] Validate callable arg signature at call site

lmondada:
Previously, the following

```python
import cudaq
from cudaq.qis import qis
from typing import Callable
@cudaq.kernel
def k1(fct: Callable[[int, cudaq.qubit], None]):
    qubit = cudaq.qubit()
    fct(0,qubit)
@cudaq.kernel
def k2(cond: bool, q: cudaq.qubit):
    if cond:
        x(q)
print(k1(k2))
```
 
was raising a very cryptic error:

```
loc("-":5:7): error: 'func.call' op operand type mismatch: expected operand type 'i1', but provided 'i64' for operand number 0
error: substitution module must have been created
RuntimeError: Could not successfully apply argument synth.
```

This PR cleans up the error message by checking the type of the callable.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -650,6 +650,7 @@ class PyKernelDecorator(object):
 
     def process_argument(self, arg, arg_type):
         if isa_kernel_decorator(arg):
+            _validate_kernel_callable_arg(arg, arg_type)
             return DecoratorCapture(arg)
 
         arg = self.convertStringsToPauli(arg)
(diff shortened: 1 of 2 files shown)


python/cudaq/kernel/kernel_decorator.py
```diff
@@ -775,3 +776,48 @@ def _parse_ast(funcSrc: str, verbose: bool = False):
...
+    if mismatches or expectedResults != actualResults:
+        details = "; ".join(f"argument {i}: expected `{exp}`, got `{act}`"
+                            for i, exp, act in mismatches)
+        if expectedResults != actualResults:
+            ret_detail = (f"return: expected `{_format([], expectedResults)}`, "
+                          f"got `{_format([], actualResults)}`")
+            details = f"{details}; {ret_detail}" if details else ret_detail
+        emitFatalError(
+            f"Expected callable with signature {_format(expectedInputs, expectedResults)}, "
+            f"got callable kernel `{kernel.name}` with signature {_format(actualInputs, actualResults)}"
```
sacpis: nit: Should we include `details` in this error message?

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -775,3 +776,48 @@ def _parse_ast(funcSrc: str, verbose: bool = False):
...
+
+def _validate_kernel_callable_arg(kernel, arg_type):
+    """
+    Validate that `kernel` matches the `arg_type` callable signature.
+    """
+    if not cc.CallableType.isinstance(arg_type):
+        emitFatalError(
+            f"Expected argument of type `{arg_type}`, got callable kernel `{kernel.name}`"
+        )
+
```
sacpis, line 789: We can include happy path here.

```
actual_type = kernel.signature.get_callable_type()
if actual_type == arg_type:
    return ...

# Mismatch can go here
```

This way we will only have one `==` on uniqued MLIR types. For a kernel like `k1(fct, a, b, c, d, ...)` invoked in a loop, this is a difference between O(n) per call and O(1) per call. Here n is the number of parameters.
In the mismatch path, we will still have O(n) to walk args and report mismatch. I think that's fine for the failure case.

sacpis (approved):
Overall LGTM. Thanks @lmondada. Left 2 comments.

atgeller (approved):
Thanks @lmondada!

python/cudaq/kernel/kernel_decorator.py
lmondada: Ah right, I had simplified the error message but then left in the unused variables. Fixed!

python/cudaq/kernel/kernel_decorator.py
lmondada: Oh yes, great idea, thanks!

PR #4469, opened 2026-05-07
Surface QPU headers into user code to enable compile-time checks.

Renaud-K:
With these changes, header files are created for the QPU classes and are included in `platform.h`. This is a necessary condition to enable trait based design and compile-time checks. The dependencies between algorithms and QPUs is turned around. It used to be that QPU would include directly or indirectly header files used for user code (`cudaq.h`) thereby introducing circular dependencies. This is no longer the case. QPU have no dependency with user code and including them into user code does not leak any unwanted header. 

This PR is the fourth in a series aimed at surfacing header files:

- [x] #4282 removes unnecessary inclusion of `cudaq.h` in the runtime itsef
- [x] #4416 prevents nlohmanm/json.h from leaking into user code
- [x] #4445 addresses the `cudaq.h` compilation performance from the added headers.
- [ ] This PR surfaces headers
- [ ] The next PR will address another compilation performance issue preventing ExecutionManager and SimulationState to leak into cudaq.h

This has been achieved by

1. Breaking up header files
    -   `KernelExecution` has been moved out of `ServerHelper`
    - `apply_noise` has been moved out of `qubit_qis.h`

1. Moving non-templatized code into `.cpp` files.
    Types are forward declared in a new `qpu_utils.h` files. The implementation in `qpu_utils.cpp` includes the full headers.

1. Introducing a  `cudaq::owning_ptr` utility which has been created as an `std::unique_ptr` taking an out-of-line deleter argument.
    This allows for compilation of forward declared `std::unique_ptr` types. A capability typically provided by `std::shared_ptr` but that is extended to `std::unique_ptr` to maintain the single ownership semantic. 

1. Restricting the included headers to the strict minimum

1. Introducing free function APIs in `platform_iface.h` and `platform_iface.cpp` which lets the QPU use the platform as a C API. 
    Often APIs are accessed through global objects such as  `getPlatform()`, `getCurrentSimulator()`, `getExecutionManager()`, `getSimulationState()`, for which the API requires very few headers but the object itself leaks a lot of unwanted ones. With this technique, `getObject().do_something()` is called in the `object_iface.cpp` file while `object_iface.h` declares a free function `object::do_something` to be called from the client code. All unwanted headers are trapped in  `object_iface.cpp` while `object_iface.h` only needs to include types used by the free function. 
    Example:
    ```
    #include <functional>

    namespace cudaq {
    class ExecutionContext;
    
    namespace platform {
    void with_execution_context(ExecutionContext &ctx, std::function<void()> f)
    } // namespace platform
    } // namespace cudaq
    ```


Renaud-K:
> /ok to test 64e1b3526cd348c2810a9f385f16f21dcfa2cfe2

runtime/cudaq/apply_noise.h
```diff
@@ -0,0 +1,133 @@
+/****************************************************************-*- C++ -*-****
```
lmondada, line 1: Moving `apply_noise`  from `qubit_qis_h` to here seems like something that is likely to 'break' user-code. (I'm putting 'break' in quotes, because our C++ AST parser might still work fine -- maybe it only breaks library mode?). I'm not sure how we flag breaking changes, but this seems like a thing that we should mention in the release notes.

runtime/cudaq/platform.h
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "cudaq/builder/kernel_builder.h"
+#include "cudaq/platform/qpu_types.h"
```
lmondada, line 12: Is it necessary to include all QPUs by default? To me it seems fine to ask the users to include the appropriate header when they want to use a specific QPU, but I might be missing the reason these are here.

lmondada (approved):
Thanks a lot Renaud, your changes make a lot of sense and are a welcome clean up.

One question (not for this PR): given that each QPU is compiled into its own shared library, do you think they should each be in their own directory? It might be surprising that all header are side by side when they are in fact each in a separate library.

runtime/cudaq/apply_noise.h
Renaud-K: This is not breaking. `apply_noise.h` is explicitly added to `cudaq.h`. So, all of what used to be `qubit_qis.h` is still visible from `cudaq.h`. But qpus can now include `qubit_qis.h` with no circular dependency.

runtime/cudaq/platform.h
Renaud-K: This is there to stress test the changes. Without it, we really do not know if user code (all of our tests that is) can be compiled. I am on the fence on keeping it as well but the downside, of removing it, is that we will find out late if bad dependencies are reintroduced. 

I like the fact that it also stresses the compilation performance: https://github.com/NVIDIA/cuda-quantum/pull/4445.

I would rather we remove the kernel builder from `cudaq.h` and have it as an opt-in because it would improve performance the most,

Renaud-K:
> /ok to test 361dd0dec754f03969a222cd87910ed77bc73737

PR #4470, opened 2026-05-07
Fix run_async crash on shots_count=0 and leak in async_run_result

huaweil-nv:
## Summary

  Two small fixes in `python/runtime/cudaq/algorithms/py_run.cpp`,
  split into two commits.

  ### 1. `run_async(..., shots_count=0).get()` no longer crashes

  `run_async_impl` had an early `if (!shots_count) return {};` that produced a
  value-initialized `async_run_result` whose `std::future` had no shared state.
  The follow-up `.get()` then threw `std::future_error: No associated state`.

  The function already had a `shots_count == 0` branch further down that
  allocates the buffers and returns a ready future — the early return was
  short-circuiting it. Drop the early return.

  This restores parity with the synchronous `cudaq.run(..., shots_count=0)`
  (which returns `[]`) and matches the documented contract that `shots_count`
  is non-negative and the result list length equals `shots_count`.

  ### 2. `async_run_result` no longer leaks memory

  `async_run_result` held `std::vector<nb::object>*` and `std::string*` as raw
  pointers with no destructor. Cleanup lived only in `AsyncRunResultImpl.get()`,
  and was asymmetric — the success path freed only `results`, the error path
  freed only `error`. The mismatched buffer leaked on every call. If `.get()`
  was never invoked, both leaked.

  Switch ownership to `std::unique_ptr` so the struct destructor cleans up on
  every path. The deferred future captures non-owning raw pointers obtained via
  `.get()` on the unique_ptrs; ordering of struct members guarantees the future
  is destroyed before the unique_ptrs, so the captures are valid for the
  future's lifetime.

  `get()` is also simplified: throw straight from `*self.error`, and `std::move`
  the results vector out instead of `copy + delete`. This removes per-element
  Python refcount churn for large `shots_count`, and the success path becomes
  `noexcept`.

  ## Reproducer (before this PR)

  ```python
  import cudaq

  @cudaq.kernel
  def bell() -> int:
      q = cudaq.qubit()
      h(q)
      return mz(q)

  # Bug 1: crashes with "std::future_error: No associated state"
  cudaq.run_async(bell, shots_count=0).get()

  # Bug 2: heap-allocated string/vector leaked on every call
  for _ in range(10000):
      cudaq.run_async(bell, shots_count=10).get()

### DIFF
--- a/python/runtime/cudaq/algorithms/py_run.cpp
+++ b/python/runtime/cudaq/algorithms/py_run.cpp
@@ -14,6 +14,7 @@
 #include "utils/OpaqueArguments.h"
 #include "mlir/Bindings/Python/NanobindAdaptors.h"
 #include <future>
+#include <memory>
 #include <nanobind/nanobind.h>
 #include <nanobind/ndarray.h>
 #include <nanobind/stl/complex.h>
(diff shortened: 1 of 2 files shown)


sacpis (approved):
LGTM. Thanks @huaweil-nv.

1tnguyen (approved):
LGTM 👍

huaweil-nv:
> /ok to test 7d9fbcccfacf2930b69f029f2a62bf3775e48968

PR #4471, opened 2026-05-07
Merge launchKernel + launchModule into unifiedLaunchModule

lmondada:
~Stacks on top of #4446~ Done.

This PR unifies the `launchKernel` and `launchModule` endpoints in `quantum_platform`, `QPU` and its derived classes into a single `unifiedLaunchModule`.

This is achieved by moving the differences between the two methods into `if` and `else` branches within the `unifiedLaunchModule` implementation. This in itself reduces code duplication somewhat by removing the identical parts across the two execution paths. More importantly, it unblocks us to proceed from here on in parallel on:

- consolidating and reducing further the differences between the compilation and launch pipelines for different QPUs and host languages. In particular, we will be able to remove `AnyModule` once the C++ path is split into two phases (mlir-opt + launch) that mirror the current Python path
- introducing overloads of the launch endpoint based on execution policies

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -220,41 +220,39 @@ public:
   /// representation required by the targeted backend. Handle all pertinent
   /// modifications for the execution context as well as asynchronous or
   /// synchronous invocation.
-  KernelThunkResultType launchKernel(const SourceModule &src,
-                                     KernelArgs args) override {
-    const auto &kernelName = src.getName();
-    CUDAQ_INFO("launching remote rest kernel ({})", kernelName);
-
-    auto executionContext = cudaq::getExecutionContext();
-
-    // TODO future iterations of this should support non-void return types.
-    if (!executionContext)
-      throw std::runtime_error(
-          "Remote rest execution can only be performed via cudaq::sample(), "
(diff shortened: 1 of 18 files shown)


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -262,41 +262,39 @@ class BaseRemoteRESTQPU : public QPU {
...
-    CUDAQ_INFO("launching remote rest kernel ({})", kernelName);
+  KernelThunkResultType unifiedLaunchModule(const AnyModule &module,
+                                            const KernelArgs &args) override {
+    Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
+                      noiseModel, emulate);
+    std::string kernelName;
+    std::vector<cudaq::KernelExecution> codes;
 
-    auto executionContext = cudaq::getExecutionContext();
+    if (std::holds_alternative<SourceModule>(module)) {
```
Renaud-K, line 230: This is better than #4447. Thank you. It is still code exposed in the QPU, that I thought would be hidden in the CompiledModule. I am not sold on the idea that we should have AnyModule, SourceModule and CompiledModule.
As a hack, however, it is ok. I could work with it. But, I would have to understand the long term evolution of it.

Renaud-K (approved):
Thank you for this great step towards unification.

PR #4475, opened 2026-05-07
remove per-PR repo check links as it is done nightly in integration tests

mitchdz:
Let's not check links on every PR. It is fine if bad links go through in a PR and we can catch it nightly to fix it.

Check Links is done nightly per https://github.com/NVIDIA/cuda-quantum/actions/runs/25713066304


schweitzpgi (commented):
Good idea. Reduce getting throttled by other sites as a DNS attacker. :)

sacpis:
Wondering if we should have this for all jobs which ran in PR's CI.

sacpis:
Basically, we can run CI on the combination of PRs (let's say a batch of 3) in the merge queue. These PRs had already gone through the CI, so we know the result of each. In which case, we can use merge queue to combine and run CI (don't think it as really useful in terms of running the jobs, but can act as a queue) and then merge each one.

schweitzpgi:
> Basically, we can run CI on the combination of PRs (let's say a batch of 3) in the merge queue. These PRs had already gone through the CI, so we know the result of each. In which case, we can use merge queue to combine and run CI (don't think it as really useful in terms of running the jobs, but can act as a queue) and then merge each one.

Riffing on this. Do we really need to DNS attack everyone all the time? If we could scan the diffs for links *before* running a link check, we'd discover that 99% of the time no links were added or changed. If you coupled that with a nightly run to verify that the existing links were still valid, you'd cover that case as well without all the overhead we have today.

khalatepradnya:
I thought the "Check links" was supposed to happen only nightly?
Ref: https://github.com/NVIDIA/cuda-quantum/actions/runs/25713066304

mitchdz:
I think this was meant to be something that is ran only nightly, which I would +1 for that.

mitchdz:
I updated this PR to simply remove Check links per PR, and leave that to the nightly test.

mitchdz:
@sacpis Yeah, my goal is to make it so as little as possible is re-ran that doesn't make sense per-PR and per-merge queue.

1tnguyen (approved):

PR #4476, opened 2026-05-07
[core] Codegen for cable ops.

schweitzpgi:


lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
```diff
@@ -716,6 +716,40 @@ struct ConcatOpRewrite
...
+        Base::wrapQubitAsArray(loc, rewriter, firstOperand, firstTy);
+    for (auto [i, next] : llvm::enumerate(adaptor.getOperands().drop_front())) {
+      Type nextTy = getInitialType(bundle, i + 1);
+      Value wrapNext = Base::wrapQubitAsArray(loc, rewriter, next, nextTy);
+      auto appended = func::CallOp::create(
+          rewriter, loc, arrayTy, cudaq::opt::QIRArrayConcatArray,
+          ArrayRef<Value>{resultArray, wrapNext});
+      resultArray = appended.getResult(0);
+    }
+    rewriter.replaceOp(bundle, resultArray);
```
sacpis, line 748: Seems like to bundle N wires the code creates N one-slot arrays and then sticks them together 1 pair at a time. I think each glueing step copies everything done so far, so the total work grows like N^2 time. 

Since we already know the final size up front, we could just make one array of size N and drop each wire into its slot. That will be much faster (O(N)) I think.

We can apply similar logic then to  ConcatOp as it is similar (as per the comment and later verified as well).

sacpis (commented):
Thanks Eric. Left a comment.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
schweitzpgi: Not exactly. quake.concat is totally dynamic. There may be no chance that the compiler can determine the length; it simply has to be done at runtime. Also, quake.concat can have arguments of `!ref` and `!veq` types.

Here, I'm not concerned with optimizations. This code is meant to be correct.

sacpis (approved):
LGTM. Thanks @schweitzpgi.

lib/Optimizer/CodeGen/ConvertToQIRAPI.cpp
sacpis: Yeah, you are right about quake.concat. Thanks @schweitzpgi.

PR #4477, opened 2026-05-07
cmake: scope --param=evrp-mode=legacy to CXX compiles only

cketcham2333:
The gcc-12 -Wrestrict workaround in runtime/common/CMakeLists.txt and realtime/unittests/CMakeLists.txt was added as a PUBLIC compile option, so cmake propagates it through interface inheritance into nvcc command lines on consumer CUDA targets. nvcc forwards it to its host gcc, which may be a different gcc than CMAKE_CXX_COMPILER (e.g. gcc-12 vs gcc-13) and may not recognize the flag. Wrap it in a $<$<COMPILE_LANGUAGE:CXX>:..> generator expression so it only appears on CXX command lines.


bmhowe23 (approved):
Thanks, Chuck!

mitchdz (approved):
Looks good to me.

PR #4478, opened 2026-05-07
Jit target compilation pipeline unification

taalexander:
### Description
This is based on #4432 and should only be looked at after merging.

Python target lowering was not running the same target JIT pipeline stages as the normal compiler flow. This broke targets that rely on custom operation synthesis, especially registered SU(4) operations used with `estimate_resources`. The synthesis created helper functions, but the Python path was missing the specialization and inlining stages needed to expose the synthesized two-qubit gates, so resource counting saw only one-qubit gates. We fixed this by sharing the JIT target pipeline assembly between the compiler path and the Python local QPU path. Python local targets now use the shared target pipeline only when the target explicitly configures target pass-pipeline fields, which preserves the existing behavior for local simulator targets that do not override these pipelines (which is what happened already, just incorrectly). This centralizes the target YAML pipeline behavior while keeping Python lowering aligned with the standard compiler flow. 

Note the test `test_custom_unitary_produces_2q_gates` would fail if we just removed the passes from the target without this unification.

This allows me to remove the passes that in https://github.com/NVIDIA/cuda-quantum/pull/4432 @schweitzpgi noted shouldn't be required.

One possible follow-up question is whether the local Python QPU path should eventually delegate more of its module specialization and target-pipeline execution to `Compiler.cpp`, similar to the REST path, while preserving Python-specific JIT caching, resource counting, and result-handling behavior.

### DIFF
--- a/runtime/internal/compiler/include/cudaq_internal/compiler/JITTargetPipeline.h
+++ b/runtime/internal/compiler/include/cudaq_internal/compiler/JITTargetPipeline.h
@@ -16,55 +16,35 @@ class TargetConfig;
 
 namespace cudaq_internal::compiler {
 
-/// Materialized target JIT pipeline settings derived from a target YAML config
-/// and runtime target arguments.
+/// Target JIT pipeline settings resolved from target config and runtime args.
 struct JITTargetPipelineConfig {
-  /// MLIR pass-pipeline string to run. Defaults to the compiler's baseline
-  /// canonicalization-only pipeline when the target has no backend config.
+  /// MLIR pass pipeline to run.
   std::string passPipelineConfig = "canonicalize";
-  /// Codegen emission selected by the target config and runtime args, e.g.
-  /// "qir-adaptive", "qir-base", "qasm2", or "nop".
+  /// Codegen emission selected by the target config.
   std::string codegenTranslation;
-  /// Optional pass-pipeline string to run after codegen translation.
+  /// Optional pass pipeline to run after codegen.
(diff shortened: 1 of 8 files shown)


runtime/internal/compiler/include/cudaq_internal/compiler/JITTargetPipeline.h
```diff
@@ -0,0 +1,50 @@
...
+  /// Backend data provides target pass pipeline fields.
+  bool hasConfiguredPassPipeline = false;
+  /// `target-pass-pipeline` is an exact pipeline override.
+  bool usesTargetPassPipelineOverride = false;
+  /// Standard `jit-finalize-pipeline` is part of `passPipelineConfig`.
+  bool runsStandardFinalize = false;
+};
+
+/// Build the staged JIT target pipeline or exact target override.
+JITTargetPipelineConfig buildJITTargetPipelineConfig(
```
lmondada: This wouldn't look out of place as a static factory method in `JITTargetPipelineConfig`. You could call it `createFromTargetConfig` or something, if you wanted.

lmondada (approved):
Thanks for the consolidation!

runtime/internal/compiler/include/cudaq_internal/compiler/JITTargetPipeline.h
taalexander: Moved this into `JITTargetPipelineConfig::createFromTargetConfig`

PR #4481, opened 2026-05-08
[Testing][Build] Fix a test file build config for gcc-12

1tnguyen:
Similar to https://github.com/NVIDIA/cuda-quantum/pull/4477, we need this fix for the `gtest` target, which will be used to compile `get_state_tester.cu`.


atgeller (approved):

atgeller:
Thanks @1tnguyen!

PR #4485, opened 2026-05-08
[pasqal/quera] Fix segfault in C++ analog target binaries 

khalatepradnya:
### Summary

* Pasqal nightly has been segfaulting since April 11 because PR #4282 inadvertently broke the static-initializer chain that wires `nvq++`-compiled C++ binaries to their target backend. 
* Move the `TargetSetter` block to `cudaq/host_config.h` so every public CUDA-Q header transitively pulls it in.
* Add a CMake test that catches this class of regression at PR time.

### What Changed

- **Permanent fix**: Move the `TargetSetter` declaration and `NVQPP_TARGET_BACKEND_CONFIG`-gated inline global from `runtime/cudaq.h` to `runtime/cudaq/host_config.h`. `host_config.h` is universally included by every public CUDA-Q algorithm header, so any C++ TU targeting a `gen-target-backend: true` backend now reliably wires `setTargetBackend()` at static init.
- **Defense in depth**: Add `#include <cudaq.h>` to the three analog docs examples (`pasqal.cpp`, `quera_basic.cpp`, `quera_intro.cpp`) to match the include pattern used by every other target's example.

### Testing

- New ctest `backend_target_setter_check` exercises the link-time symbol presence for all three analog examples.
- End-to-end: `pasqal.cpp` compiled with `nvq++ --target pasqal` and run against EMU_FREE returns valid sample results
  (`{ 000:56 001:12 010:16 100:16 }`) instead of segfaulting.


mitchdz (approved):

sacpis:
@khalatepradnya Have you tried passing `PREPROCESSOR_DEFINES="${PREPROCESSOR_DEFINES} -include cudaq/host_config.h" ` in `nvq++`? With this `nvq++` guarantees the runtime backend wiring. We could have avoided including all the headers, no? `PREPROCESSOR_DEFINES` is forwarded to both the C++ compiler and cudaq-quake, so `-include cudaq/host_config.h` will reach both.

runtime/cudaq/host_config.h
```diff
@@ -32,4 +32,23 @@ using real = double;
 
 using complex = std::complex<real>;
 
+namespace __internal__ {
```
schweitzpgi, line 35:
```suggestion
namespace detail {
```

khalatepradnya:
> @khalatepradnya Have you tried passing `PREPROCESSOR_DEFINES="${PREPROCESSOR_DEFINES} -include cudaq/host_config.h" ` in `nvq++`? With this `nvq++` guarantees the runtime backend wiring. We could have avoided including all the headers, no? `PREPROCESSOR_DEFINES` is forwarded to both the C++ compiler and cudaq-quake, so `-include cudaq/host_config.h` will reach both.

Thanks @sacpis!  I tried this before settling on the header relocation, and it doesn't quite work end-to-end. `nvq++` invokes `cudaq-quake` first on every source file outside library mode, and `cudaq-quake` parses its arguments with LLVM `cl::opt`. It only declares `-D`, `-I`, `-J`, and `-Xcudaq` ; `-include` is an unknown flag and aborts the compile:

    $ cudaq-quake -D CUDAQ_ANALOG_TARGET ... \
        -include cudaq/host_config.h \
        -I .../runtime/include ... pasqal.cpp -o pasqal.qke
    cudaq-quake: Unknown command line argument '-include'.
    cudaq-quake: Too many positional arguments specified!
    exit: 1

Same outcome with an absolute path or the `-include=...` / `--include` forms.
To fix it this way would be a larger change, so I'm inclined to keep this PR scoped to the header relocation.

The example edits I'd lean toward keeping regardless: they make the docs match the canonical include pattern other targets use, and switch to angle-bracket form.

sacpis:
> > @khalatepradnya Have you tried passing `PREPROCESSOR_DEFINES="${PREPROCESSOR_DEFINES} -include cudaq/host_config.h" ` in `nvq++`? With this `nvq++` guarantees the runtime backend wiring. We could have avoided including all the headers, no? `PREPROCESSOR_DEFINES` is forwarded to both the C++ compiler and cudaq-quake, so `-include cudaq/host_config.h` will reach both.
> 
> Thanks @sacpis! I tried this before settling on the header relocation, and it doesn't quite work end-to-end. `nvq++` invokes `cudaq-quake` first on every source file outside library mode, and `cudaq-quake` parses its arguments with LLVM `cl::opt`. It only declares `-D`, `-I`, `-J`, and `-Xcudaq` ; `-include` is an unknown flag and aborts the compile:
> 
> ```
> $ cudaq-quake -D CUDAQ_ANALOG_TARGET ... \
>     -include cudaq/host_config.h \
>     -I .../runtime/include ... pasqal.cpp -o pasqal.qke
> cudaq-quake: Unknown command line argument '-include'.
> cudaq-quake: Too many positional arguments specified!
> exit: 1
> ```
> 
> Same outcome with an absolute path or the `-include=...` / `--include` forms. To fix it this way would be a larger change, so I'm inclined to keep this PR scoped to the header relocation.
> 
> The example edits I'd lean toward keeping regardless: they make the docs match the canonical include pattern other targets use, and switch to angle-bracket form.

Actually was suggessting something like this in `nvq++`

```
TARGET_CONFIG="${TARGET_CONFIG}${PLATFORM_EXTRA_ARGS}"
PREPROCESSOR_DEFINES="${PREPROCESSOR_DEFINES} -D NVQPP_TARGET_BACKEND_CONFIG="\"${TARGET_CONFIG}\"""
PREPROCESSOR_DEFINES="${PREPROCESSOR_DEFINES} -include cudaq/host_config.h"
```

runtime/cudaq/host_config.h
khalatepradnya: Thanks @schweitzpgi. A couple of reasons I'd rather keep this PR as-is and not accept this change:

1. This PR is a relocation, not a re-namespacing. `class TargetSetter` has lived in `cudaq::__internal__` since #3805. I'm just moving the block from `cudaq.h` to `host_config.h` to restore the static-init chain that #4282 broke.
2. Applying the suggestion in-place breaks the build: the constructor body in `cudaq.cpp:203` and the friend grant in `quantum_platform.h:42,239` are both written against `cudaq::__internal__::TargetSetter`.
3. `cudaq::__internal__` is used in ~25 other files across `runtime/`, and `cudaq::detail` already exists as a separate namespace for unrelated implementation-detail code (operators, QPU helpers, etc.). A partial rename here would create more inconsistency than it removes.

If we want to revisit the namespace choice, I'd value @Renaud-K 's input given he authored the original placement and would prefer to do it uniformly in a follow-up rather than here.

sacpis:
Got it. Thanks for trying it.

sacpis:
Tried this patch which requires change only to `nvq++.in`

```
if ${GEN_TARGET_BACKEND} && ${SET_TARGET_BACKEND}; then
		# Add a function that will run before main and set the target
		# backend on the quantum_platform
		TARGET_CONFIG="${TARGET_CONFIG};emulate;${CUDAQ_EMULATE_REMOTE}"
		TARGET_CONFIG="${TARGET_CONFIG};disable_qubit_mapping;${DISABLE_QUBIT_MAPPING}"
		if [ -n "${MAPPING_FILE}" ]; then
			TARGET_CONFIG="${TARGET_CONFIG};mapping_file;${MAPPING_FILE}"
		fi
		TARGET_CONFIG="${TARGET_CONFIG}${PLATFORM_EXTRA_ARGS}"
		STUB_CPP=$(mktemp -t nvqpp_target_setter_XXXXXX.cpp)
		STUB_OBJ="${STUB_CPP%.cpp}.o"
		TMPFILES="${TMPFILES} ${STUB_CPP} ${STUB_OBJ}"
		cat > "${STUB_CPP}" <<EOF
namespace cudaq { namespace __internal__ {
class TargetSetter { public: TargetSetter(const char *); };
} }
namespace {
cudaq::__internal__::TargetSetter target_setter_stub("${TARGET_CONFIG}");
}
EOF
		run ${CXX} ${CLANG_RESOURCE_DIR} ${COMPILER_FLAGS} -c "${STUB_CPP}" -o "${STUB_OBJ}"
		OBJS="${OBJS} ${STUB_OBJ}"
	fi
```

runtime/cudaq/host_config.h
Renaud-K: I did not know about our desire to move all private namespaces to `detail` so I used the one I found in `cudaq.h` at the time I introduced the `TargetSetter`. Renaming it to `detail` is fine for me.

Renaud-K (approved):
`host_config.h` seems like it is a better fit since not all our tests include `cudaq.h`. 
Thank you for the fix!

runtime/cudaq/host_config.h
khalatepradnya: Ack, given that this would mean changes across multiple files, I will capture this in a follow-up.

khalatepradnya: https://github.com/NVIDIA/cuda-quantum/issues/4489

PR #4488, opened 2026-05-08
Reuse LLVM build tree for MLIR python bindings

mitchdz:
The wheel container previously re-ran build_llvm.sh per Python version with
  python-bindings added, which reconfigured from scratch and recompiled the
  full LLVM/MLIR/clang/lld tree (~5,800 objects, ~65 min).

  The manylinux base already builds and retains /llvm-project/build without
  bindings. Add a focused script that reuses that tree, flips
  MLIR_ENABLE_BINDINGS_PYTHON on, and runs only the install-MLIRPythonModules
  and install-mlir-python-sources targets. nanobind is still rebuilt per
  Python since it ABI-binds to the interpreter.


atgeller (approved):
Nice. Thanks @mitchdz!

schweitzpgi (approved):

PR #4493, opened 2026-05-09
[runtime] Introduce `nvqir::AnalysisScope` and migrate resource counter

khalatepradnya:
### Summary

* Introduces `nvqir::AnalysisScope`, a single RAII-managed type that owns the thread-local simulator override used by non-sampling analyses, and migrates the resource counter onto it via `nvqir::resource_counter::make_scope`. 
* Generalizes the NVQIR resolver's analysis check, retires the parallel `nvqir::*` and `python::detail::*` machinery, and lays the foundation for the incoming Detector Error Model engine without further plumbing changes.

### Motivation

The resource counter and the upcoming DEM engine each need to temporarily swap the active circuit simulator for an analysis-specific one for the duration of a kernel run. The pre-existing resource-counter machinery was fragmented across:

* A thread-local `using_resource_counter` boolean in NVQIR.
* Six `nvqir::` free functions: `switchToResourceCounterSimulator`, `stopUsingResourceCounterSimulator`, `isUsingResourceCounterSimulator`, `setChoiceFunction`, `getResourceCounts`, `setResourceCounts`.
* Four `python::detail::*` shim functions mirroring the nvqir surface.
* Manual `try`/`catch` cleanup in the Python driver, no cleanup in the C++ driver.

The DEM prototype had a third shape (manual push/pop). Each new analysis would have repeated the pattern with its own thread-locals and free functions. This PR consolidates the pattern onto one shared, RAII-driven type so DEM and any future analysis plug in with a single factory function.

### Impact

User-facing APIs (`cudaq::estimate_resources`, `cudaq.estimate_resources`) are unchanged.


runtime/nvqir/resourcecounter/ResourceCounter.cpp
```diff
@@ -19,18 +22,34 @@ ResourceCounter *getResourceCounterSimulator() {
...
+namespace cudaq::analysis::resource_counter {
+
+scope make_scope(std::function<bool()> choice) {
+  auto *rc = nvqir::getResourceCounterSimulator();
+  rc->setChoiceFunction(std::move(choice));
+  return scope{
+      "resource_counter",
+      *rc,
+      {.on_enter = nullptr, .on_exit = [](nvqir::CircuitSimulator &sim) {
+         static_cast<nvqir::ResourceCounter &>(sim).setToZeroState();
```
atgeller: Is this cast necessary here? I thought `setToZeroState` is a virtual function in `CircuitSimulator`

khalatepradnya: No, removed.

runtime/cudaq/analysis/scope.h
```diff
@@ -0,0 +1,89 @@
...
+
+#include <functional>
+#include <string>
+#include <string_view>
+
+namespace nvqir {
+class CircuitSimulator;
+}
+
+namespace cudaq::analysis {
```
Renaud-K: Can this be in the nvqir directory, together with the .cpp file with name AnalysisScope.h
The .cpp file is compiled in the nvqir library, so this utility should be part of the nvqir namespace, shouldn't it?
Then, if the filename is CamelCase, the code probably should too.

runtime/nvqir/resourcecounter/ResourceCounter.cpp
```diff
@@ -19,18 +22,45 @@ ResourceCounter *getResourceCounterSimulator() {
   return resource_counter_simulator;
 }
 
-void setChoiceFunction(std::function<bool()> choice) {
-  getResourceCounterSimulator()->setChoiceFunction(choice);
+} // namespace nvqir
+
+namespace cudaq::analysis::resource_counter {
```
Renaud-K: I think this should be in the nvqir namespace as well. And `resource_counter.h` can probably merge with `ResourceCounter.h`.

runtime/cudaq/analysis/scope.h
khalatepradnya: Good catch on the filename inconsistency!
I've renamed `AnalysisScope.cpp` -> `analysis_scope.cpp` to match the "snake_case" used by the public namespace (`cudaq::analysis::scope`) and the rest of the `runtime/cudaq/` subtree (`qubit_qis.h`, `kernel_builder.h`, `resource_counter.h`).

I'd like to keep the header at `runtime/cudaq/analysis/scope.h` and the namespace as `cudaq::analysis`, for two reasons:

1. CUDA-Q's existing convention is `cudaq::*` for the user-facing API surface and `nvqir::*` for the engine implementation, even when the .cpp lives in libnvqir. `cudaq::sample`, `cudaq::observe`, `cudaq::run`, `cudaq::draw`, and `cudaq::estimate_resources` all follow this pattern. Moving `cudaq::analysis::scope` to `nvqir::AnalysisScope` would be inconsistent, no?

2. The `cudaq::analysis::*` umbrella was chosen so the upcoming Detector Error Model engine fits as a sibling (`cudaq::analysis::dem_engine`) without further restructuring. A flat `nvqir::AnalysisScope` removes that ability.

runtime/nvqir/resourcecounter/ResourceCounter.cpp
khalatepradnya: Could we keep these split? 
`resource_counter.h` is a small facade (only `common/Resources.h` + a forward-declared simulator) - what `cudaq::estimate_resources` ends up pulling in transitively.
`ResourceCounter.h` brings in `CircuitSimulator.h`, `Gates.h`, and the full engine class body. Merging would put all of that on every user TU that calls `estimate_resources`, no?

On the namespace: same reasoning as the scope.h thread, I'd like to keep `cudaq::*` for the API tier so it stays consistent with `cudaq::sample` / `observe` / `estimate_resources`. What do you think?

runtime/cudaq/analysis/scope.h
Renaud-K: This is just a utility very much tailored for an nvqir object. It is not user facing. It is the nvqir in the library. It belongs in the nvqir namespace. The inconsistency would be for it not to be in the nvqir namespace.

Having a header file in one library with one namespace and an implementation in another library with an inconsistent namespace raises flags.

> A flat nvqir::AnalysisScope removes that ability.

You can still have `nvqir::resource_count`, `nvqir::dem`.

runtime/nvqir/resourcecounter/ResourceCounter.cpp
Renaud-K: In that case, they should remain separate. I see that you are including it `resource_estimation.h` which should not leak `CircuitSimulation.h`. But then we need to think about names because we have 2 resource_counters now, case aside, and the new header belongs in `nvqir`.

runtime/cudaq/analysis/scope.h
khalatepradnya: Fair points! I've moved both headers under `runtime/nvqir/` and flipped the namespaces

runtime/nvqir/resourcecounter/ResourceCounter.cpp
khalatepradnya: Done. The new header is `runtime/nvqir/resourcecounter/ResourceCounterScope.h` in `nvqir::resource_counter`. Distinct filename from `ResourceCounter.h` so there's no case-insensitive collision.

python/utils/LinkedLibraryHolder.h
```diff
@@ -8,23 +8,17 @@
...
 #include <filesystem>
-#include <functional>
 #include <map>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 namespace nvqir {
 class CircuitSimulator;
-void switchToResourceCounterSimulator();
```
Renaud-K, line 24: Great simplification!

Renaud-K (approved):
This looks very good and well structured. Thank you for the changes.

atgeller (approved):
Thanks @khalatepradnya! These are very nice changes!

PR #4496, opened 2026-05-11
[CI] - Implement exponential backoff retry to various network operations

mitchdz:
Adds exponential backoff retry to various components in our CI pipeline.

Most of these are surgical, and are directly targeting actual failures we have seen in the last month that have been solved by retries.

The retry implementation is relatively surgical because if the failure is due to rate limiting, we do not want to just keep trying as that will compound the issue, yet most of the transient failures are due to network issues.

One thing to note is there are reusable workflows for retry mechanisms - namely `nick-fields/retry@v4` is useful. I chose not to use this reusable worfklow because it runs just a single command, and most of our failures are embedded in scripts and other logic that it doesn't make sense to retry _everything_. If a seg fault during the build fails the job at 2 hours, we do not want to try it 3 times in a row just to seg fault again with the same failure.

### DIFF
--- a/.github/workflows/build_package_sources.yml
+++ b/.github/workflows/build_package_sources.yml
@@ -78,8 +78,9 @@ jobs:
 
       - name: Pull images
         run: |
-          docker pull "${{ steps.images.outputs.base_image }}"
-          docker pull "${{ steps.images.outputs.target_image }}"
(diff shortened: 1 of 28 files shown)


1tnguyen (approved):
Nice! Thanks @mitchdz.

schweitzpgi (approved):

bettinaheim (approved):
looks reasonable

PR #4498, opened 2026-05-11
Preserve the `cc.loop` for `nop` codegen (raw CUDA-Q MLIR code)

1tnguyen:
If the target is expected to consume high-level CUDA-Q MLIR (`nop` codegen), preserve `cc.loop` ops in the IR.


cudaq/include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -61,11 +61,13 @@ void registerJITPipelines();
...
 /// Default values are threshold = 1024, allow break = true, and allow closed
-/// interval = true.
+/// interval = true. If preserve loops is true, the pipeline keeps cc.loop
+/// operations instead of unrolling them.
 void createClassicalOptimizationPipeline(
     mlir::OpPassManager &pm, std::optional<unsigned> threshold = std::nullopt,
     std::optional<bool> allowBreak = std::nullopt,
-    std::optional<bool> allowClosedInterval = std::nullopt);
+    std::optional<bool> allowClosedInterval = std::nullopt,
+    std::optional<bool> preserveLoops = std::nullopt);
```
schweitzpgi: The correct thing to do here is to _not_ run cc-loop-unroll.

lib/Optimizer/Transforms/Pipelines.cpp
```diff
@@ -27,6 +27,20 @@ struct TargetPrepPipelineOptions
       llvm::cl::desc(
           "Enable loop unrolling on loops with early exit conditions."),
       llvm::cl::init(false)};
+  PassOptions::Option<bool> preserveLoops{
+      *this, "preserve-loops",
```
schweitzpgi: Can we rename this to `no-loop-unroll`?

unittests/backends/quake_backend/test_loop_app.cpp
```diff
@@ -0,0 +1,51 @@
...
+  if (mz(q[3]))
+    result |= 8;
+
+  return result;
+}
+
+int main() {
+  const auto results = cudaq::run(3, loop_payload_stress);
+  for (auto result : results) {
+    if (result != 15) {
```
schweitzpgi, line 43: How does this test that the loop wasn't unrolled? I think a more targeted test in the regression tests is required for this one.

schweitzpgi (commented):
I think the option name should directly refer to what it does: `no-loop-unroll`.

Also, can we make a regression test in `test/Transforms` that explicitly exercises this pass and the new option? I can't tell how the unittest is actually verifying that loop unrolling occurred and did not occur.

cudaq/include/cudaq/Optimizer/Transforms/Passes.h
1tnguyen: It turned out that we only run the `cc-loop-unroll` as a pass later (QIR codegen). In these earlier pipelines that we want to skip for `nop` codegen, we use the `UnrollCountedLoop` pattern directly.

lib/Optimizer/Transforms/Pipelines.cpp
1tnguyen: Absolutely, renamed in https://github.com/NVIDIA/cuda-quantum/pull/4498/changes/6db04d1e0b7b68edf6096813e09cbe28cec7d5e7

unittests/backends/quake_backend/test_loop_app.cpp
1tnguyen: We have a check embedded in the test script so that we can check the submitted IR as well as the execution on the mock server.

Agree, it's a good idea to have a separate `lit` regression check for this, hence added in https://github.com/NVIDIA/cuda-quantum/pull/4498/changes/6db04d1e0b7b68edf6096813e09cbe28cec7d5e7 (`test/AST-Quake/jit_pipeline_no_loop_unroll.cpp`)

Update: made it a transform only test in https://github.com/NVIDIA/cuda-quantum/pull/4498/changes/7e3edd4b9149d0d392370e9022cd3349d3fe7cae as I read the other comment :)

1tnguyen:
> I think the option name should directly refer to what it does: `no-loop-unroll`.
> 
> Also, can we make a regression test in `test/Transforms` that explicitly exercises this pass and the new option? I can't tell how the unittest is actually verifying that loop unrolling occurred and did not occur.

Fixed: rename the option to `no-loop-unroll` as suggested and add a targeted lit test (`test/Transforms/jit_pipeline_no_loop_unroll.qke`.

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -210,7 +210,9 @@ def ClassicalOptimization : Pass<"classical-optimization"> {
     Option<"allowClosedInterval", "allow-closed-interval", "bool",
       /*default=*/"true", "Allow loop iterations on a closed interval.">,
     Option<"allowBreak", "allow-early-exit", "bool", /*default=*/"false",
-      "Allow unrolling of loop with early exit (i.e. break statement).">
+      "Allow unrolling of loop with early exit (i.e. break statement).">,
+    Option<"preserveLoops", "no-loop-unroll", "bool", /*default=*/"false",
```
schweitzpgi: I still see "preserve". Can we get rid of that? `enableLoopUnrolling` or `disableLoopUnrolling`?

1tnguyen: Definitely. Renamed the variable to `disableLoopUnrolling` in 1d5a3dc1906eb9400bcd16907dee0a12c1263118

schweitzpgi (approved):

PR #4500, opened 2026-05-11
Move quake namespace.

schweitzpgi:
This moves the quake namespace to cudaq::quake in preparation for adding formal subprojects to the CUDA-Q monorepo.

### DIFF
--- a/runtime/common/DeviceCodeRegistry.cpp
+++ b/runtime/common/DeviceCodeRegistry.cpp
@@ -9,10 +9,10 @@
 #include "DeviceCodeRegistry.h"
 #include "cudaq/qis/qkernel.h"
 #include "cudaq/runtime/logger/logger.h"
+#include <algorithm>
 #include <map>
 #include <mutex>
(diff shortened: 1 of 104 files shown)


runtime/common/DeviceCodeRegistry.cpp
```diff
@@ -9,10 +9,10 @@
 #include "DeviceCodeRegistry.h"
 #include "cudaq/qis/qkernel.h"
 #include "cudaq/runtime/logger/logger.h"
+#include <algorithm>
```
atgeller, line 12: Are these include changes relevant to this PR, or linter changes that snuck in?

atgeller (approved):
Just one minor question, otherwise LGTM.

runtime/common/DeviceCodeRegistry.cpp
schweitzpgi: There was this 1 to address a build failure in my environment.

PR #4505, opened 2026-05-12
Fix cudaq-common missing transitive include

mitchdz:
PR https://github.com/NVIDIA/cuda-quantum/pull/4469 promoted cudaq-mlir-runtime-headers to PUBLIC on the cudaq
  library so consumers transitively get the runtime/internal/compiler
  include path. cudaq-common was not updated in lockstep, so downstream
  targets that link only cudaq-common (e.g. gpu-emulated-qpu) and
  transitively reach BaseRemoteRESTQPU.h fail to find
  cudaq_internal/compiler/CompiledModuleHelper.h.

  Add cudaq-mlir-runtime-headers to cudaq-common's PUBLIC deps.

### DIFF
--- a/runtime/common/CMakeLists.txt
+++ b/runtime/common/CMakeLists.txt
@@ -51,6 +51,7 @@ target_include_directories(${LIBRARY_NAME}
 target_link_libraries(${LIBRARY_NAME}
   PUBLIC
     cudaq-operator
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
LGTM 👍

PR #4508, opened 2026-05-12
Support for static Rotation Gate Synthesis (Clifford + T)

sacpis:
Quantum error-corrected hardware can only run a small set of gates. Arbitrary rotations like `rz(0.6)` are not among them. This PR adds the mechanism to replace those rotations with an equivalent sequence of gates the hardware can run (H, S, T, and X (the "Clifford+T" set)) to whatever accuracy you ask for.

# This PR includes

1. A new math library (cudaq::synth)
    - An embedded implementation of the gridsynth algorithm (Ross & Selinger, [arXiv:1403.2975](https://arxiv.org/pdf/1403.2975)). Given an angle and a tolerance epsilon, it finds a short Clifford + T sequence approximating `Rz(theta)`.
    - It needs exact arbitrary precission arithmetic, so `GMP` and `MPFR` are now dependencies (built as shared libs in the prerequisites scripts, with `LGPL/GPL` licenses shipped alongside and a compliance check script).

2. New compiler passes
    - clifford-t-synthesis
        - rewrites constant angle `quake.rz` ops into `H/S/T/X` sequences.
        - works with both memory and value semantics
        - options for epsilon, solver timeouts, retries, and what to do with non-constant angles
    - RxToRz / RytoRz decomposition patterns
        - Rx, Ry, and R1 get reduced to Rz before synthesis runs
    - cudaq-fault-tolerant-target
        - pipeline that chains all of the above in the right order

3. A Python API

```
from cudaq import synth

seq = synth.gridsynth(theta=0.6, epsilon=1e-10)
# number of T gates
seq.t_count 
# how far off the approximation actually is
synth.rz_error(0.6, seq) 
```

`CliffordTSequence` can also build a CUDA-Q kernel directly.

# Testing
- Unit tests for the number theory and grid search layers
- FileCheck tests for each pass and for the full pipeline
- Python tests for the bindings

# Notes / Limitation
- The global phase is currently being developed in another PR. For now this is fine for a kernel used on its own, but it becomes an observable relative phase if the kernel is wrapped in `cudaq::control`. We already have a TODO to preserve it once we add support for global phase.
- Angles must already be constants when the pass runs, argument synthesis and constant propagation need to have happened first.

@vadym-kl for viz


python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+NB_MODULE(_cudaq_synth, m) {
+  m.doc() = "Internal bindings for the Clifford+T rotation synthesis "
+            "library (cudaq-synth).";
+
+  m.def(
+      "gridsynth", &gridsynthBinding, nanobind::arg("theta"),
+      nanobind::arg("epsilon"), nanobind::arg("diophantine_timeout_ms") = 200,
+      nanobind::arg("factoring_timeout_ms") = 50,
+      R"doc(Synthesize a Clifford+T circuit approximating R_z(theta) to precision epsilon.
```
vadym-kl, line 108: Let's be precise about how we measure precision here., that is which matrix norm we are using,

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,81 @@
...
+    theta: Target rotation angle (float, or str for arbitrary precision).
+    epsilon: Approximation precision, must be > 0 (float, or str).
+    diophantine_timeout_ms: Per-candidate timeout for the Diophantine
+        solver. Higher values improve optimality at the cost of
+        worst-case latency. Default 200.
+    factoring_timeout_ms: Per-candidate timeout for integer factoring
+        inside the Diophantine solver. Default 50.
+
+Returns:
+    A string of gate characters from the alphabet {H, S, T, X, W}. The
```
vadym-kl: We need to clarify what W is and if characters are returned in circuit order or matrix multiplication order.

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+} // namespace
+
+NB_MODULE(_cudaq_synth, m) {
+  m.doc() = "Internal bindings for the Clifford+T rotation synthesis "
+            "library (cudaq-synth).";
+
+  m.def(
+      "gridsynth", &gridsynthBinding, nanobind::arg("theta"),
```
vadym-kl, line 105: Should this be reflected in `.pyi` file ?

unittests/Optimizer/CliffordTSynthesisTester.cpp
```diff
@@ -0,0 +1,198 @@
...
+    EXPECT_FALSE((isa<cudaq::quake::RxOp, cudaq::quake::RyOp,
+                      cudaq::quake::RzOp, cudaq::quake::R1Op>(op)))
+        << "rotation gate survived synthesis: "
+        << op->getName().getStringRef().str();
+  });
+
+  const M2 U = reconstructUnitary(*module);
+  const M2 expected = idealUnitary(param.gate, param.theta);
+
+  const std::complex<double> inner =
```
vadym-kl: Feel free to disregard, but it would be nice if there was a distance function ( consistent with rz_gate_sequence_error ) for distance between two matrices if the test comparison was that distance is less than `opts.epsilon` .

cudaq/unittests/Optimizer/CliffordTSynthesisTester.cpp
```diff
@@ -0,0 +1,198 @@
...
+    cudaq::quake::R1Op::create(builder, loc, /*isAdj=*/false, ValueRange{angle},
+                               ValueRange{}, q);
+    break;
+  }
+
+  func::ReturnOp::create(builder, loc);
+  return OwningOpRef<ModuleOp>(module);
+}
+
+M2 reconstructUnitary(ModuleOp module) {
```
vadym-kl, line 175: At which point do we run into rounding errors being an issue for this test ?

sacpis: The test has a comfortable safety margin of ~6 orders of magnitude. Trouble starts around epsilon 10⁻¹³ to 10⁻¹⁴.

python/runtime/synth/py_synth.cpp
sacpis: We don't need `.pyi` update.

python/runtime/synth/py_synth.cpp
sacpis: Fixed in ed33366.

python/runtime/synth/py_synth.cpp
sacpis: Fixed in ed33366.

unittests/Optimizer/CliffordTSynthesisTester.cpp
sacpis: Fixed in ed33366.

sacpis:
> /ok to test b82dfb9

sacpis:
> /ok to test fab5937

sacpis:
> /ok to test 9f40281

sacpis:
> /ok to test 80ecfc7

sacpis:
> /ok to test 363a3a6

sacpis:
> /ok to test 757421c

CMakeLists.txt
```diff
@@ -260,6 +260,7 @@ option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Fron
 option(CUDAQ_ENABLE_CC "Enable CUDA-Q code coverage generation." OFF)
 option(CUDAQ_REQUIRE_OPENMP "Fail the build if OpenMP is not found." OFF)
 option(CUDAQ_ENABLE_SANITIZERS "Enable Address Sanitizer (ASan) and Undefined Behavior Sanitizer (UBSan)." OFF)
+option(CUDAQ_ENABLE_CLIFFORD_T_SYNTHESIS "Enable the Clifford+T rotation synthesis library (requires GMP and MPFR; the build fails if those are missing)." OFF)
```
taalexander: Is the intent to distribute CUDAQ without these capabilities? I do not believe this is acceptable as rotation synthesis is a core requirement for a fault-tolerant compiler. This must be generally be available with the Python package.

cudaq/lib/Synthesis/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+find_path(GMP_INCLUDE_DIR NAMES gmp.h)
+find_path(MPFR_INCLUDE_DIR NAMES mpfr.h)
+
+if(APPLE)
+  set(_synth_saved_suffixes ${CMAKE_FIND_LIBRARY_SUFFIXES})
+  set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
```
taalexander: This explictly will only link find and link gmp/mpfr statically into the library which we *explicitly* do not ant to do.

cudaq/lib/Synthesis/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  find_library(GMP_LIBRARIES NAMES gmp libgmp)
+  find_library(MPFR_LIBRARIES NAMES mpfr libmpfr)
+  set(CMAKE_FIND_LIBRARY_SUFFIXES ${_synth_saved_suffixes})
+else()
+  find_library(GMP_LIBRARIES NAMES gmp libgmp)
+  find_library(MPFR_LIBRARIES NAMES mpfr libmpfr)
+endif()
```
taalexander: I would be much more comfortable if the build explicitly required dynamic GMP/MPFR linkage for distributed CUDA-Q artifacts. 

Right now the macOS branch forces `CMAKE_FIND_LIBRARY_SUFFIXES` to `.a`, and the Linux path leaves the static-vs-shared result up to `find_library`. 

Given the LGPL obligations around static linking, can we make the policy explicit here, for example by finding only shared libraries for GMP/MPFR in release/package builds and failing configure if only static archives are available? It doesn't need to always explicitly disallow static linking. Something like would also work to be used for all distribution. This is tied to my top-level comment about licensing and distribution.

```
  option(CUDAQ_SYNTH_REQUIRE_SHARED_GMP_MPFR
    "Require shared GMP/MPFR libraries for Clifford+T synthesis."
    ON)
```

cudaq/include/cudaq/Synthesis/Math/Integer.h
```diff
@@ -0,0 +1,680 @@
...
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#pragma once
+
+#include <cassert>
+#include <gmp.h>
```
taalexander, line 12: This and other headers make GMP/MPFR part of the public Cuda-Q C++ API/ABI surface. For example, `Integer.h` includes `<gmp.h>` and stores `mpz_t`, while `Real.h` includes `<mpfr.h>` and stores `mpfr_t`. Both headers are installed by `cudaq/include/cudaq/Synthesis/CMakeLists.txt`. This means downstream C++ users that include or re-export these CUDA-Q headers will need GMP/MPFR development headers and may need to account for runtime dependency and LGPL compliance considerations, especially if they statically link or redistribute binaries.

Can we make the intended public API boundary and downstream dependency story explicit as part of the packaging plan? Is it possible avoid publicly exposing GMP/MPFR in the headers via shims?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,430 @@
+/*******************************************************************************
```
taalexander, line 1: I am concerned that the current integration model makes the synthesis pass more complicated than necessary and obscures the distribution boundary. `CliffordTSynthesis.cpp` is compiled into `OptTransforms` unconditionally, but the real implementation is hidden behind `CUDAQ_HAS_CLIFFORD_T_SYNTHESIS`; when `cudaq-synth` is not built, the pass still exists and fails at runtime with a "built without the synthesis library" diagnostic.

Can we clarify the intended linking and registration model here? If rotation synthesis is meant to be a shipped CUDA-Q capability, should `cudaq-synth` be built and linked as part of the normal compiler/runtime distribution instead of compiling a disabled stub into `OptTransforms`? If it is intentionally optional, should the pass, pipeline entry points, tests, Python exposure, and user-facing documentation all be gated consistently so users do not see a registered pass that cannot run in default builds?

python/runtime/synth/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+else()
+  set(_origin_prefix "$ORIGIN")
+endif()
+
+set_target_properties(_cudaq_synth PROPERTIES
+  INSTALL_RPATH "${_origin_prefix}/../../lib"
+  BUILD_RPATH   "${CMAKE_BINARY_DIR}/lib")
+
+set_target_properties(_cudaq_synth PROPERTIES
+  LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/python/cudaq/synth)
```
taalexander, line 30: Can we clarify how `_cudaq_synth -> cudaq-synth -> GMP/MPFR` is expected to resolve in distributed artifacts? For wheels, containers, and installers, are `libgmp` and `libmpfr` bundled, provided by the base system, or declared as runtime dependencies?

python/cudaq/synth/__init__.py
```diff
@@ -0,0 +1,11 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+from ._cudaq_synth import gridsynth
```
taalexander: `cudaq.synth` appears to be packaged unconditionally, but `_cudaq_synth` is only built when `CUDAQ_ENABLE_CLIFFORD_T_SYNTHESIS` is enabled. Should this package be gated with the extension, or should it provide a clear "synthesis not available" warning/error instead of failing on the missing native module?

cudaq/lib/Synthesis/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  Math/Geometry/Ellipse.cpp
+  Math/Geometry/Rectangle.cpp
+  Math/Geometry/ToUpright.cpp
+  Math/Grid/Odgp.cpp
+  Math/Grid/Tdgp.cpp
+  Math/Real.cpp
+  Synthesis/Gridsynth.cpp
+  Synthesis/KmmSynthesize.cpp
+
+  LINK_LIBS PRIVATE
```
taalexander, line 39: AI-generated review comment:
`cudaq-synth` exports GMP/MPFR include directories publicly but links the GMP/MPFR libraries privately. Because the installed synthesis headers contain inline GMP/MPFR calls, downstream CMake consumers of `cudaq::cudaq-synth` may compile but fail to link unless they manually discover and link GMP/MPFR. Can the exported target model these dependencies explicitly, or should the GMP/MPFR-dependent headers stay out of the public target?

taalexander (changes requested):
Review Phase 1 - Packaging and distribution: Here I am looking and packaging/distribution/build concerns.

My chief concern right now is that totation synthesis is a critical CUDA-Q FTQC capability, but the new `CUDAQ_ENABLE_CLIFFORD_T_SYNTHESIS` option defaults to `OFF`, and I do not see any release, wheel, container, or packaging path in this PR that enables it. If that remains true, the feature will be present only for source builds that opt in manually, not in normal CUDA-Q distributed artifacts. This can't be the case as we will need to ship this functionality to users. 

As we discussed offline we need a clear packaging/licensing plan for the new GMP/MPFR dependency chain and I can't see this? The synthesis CMake links GMP and MPFR into cudaq-synth, and the macOS path searches static archives. 

Before merging, can we clarify the intended distribution model. Whether CUDA-Q packages and Python wheels will ship this feature by default, whether GMP/MPFR will be dynamically linked or bundled. If bundled where the required third-party notices/license texts will be carried, and how LGPL relinkability/source obligations are handled for wheels, containers, and installers? If not bundled what will the installation/dependency story be for users?

sacpis:
> Review Phase 1 - Packaging and distribution: Here I am looking and packaging/distribution/build concerns.
> 
> My chief concern right now is that totation synthesis is a critical CUDA-Q FTQC capability, but the new `CUDAQ_ENABLE_CLIFFORD_T_SYNTHESIS` option defaults to `OFF`, and I do not see any release, wheel, container, or packaging path in this PR that enables it. If that remains true, the feature will be present only for source builds that opt in manually, not in normal CUDA-Q distributed artifacts. This can't be the case as we will need to ship this functionality to users.
> 
> As we discussed offline we need a clear packaging/licensing plan for the new GMP/MPFR dependency chain and I can't see this? The synthesis CMake links GMP and MPFR into cudaq-synth, and the macOS path searches static archives.
> 
> Before merging, can we clarify the intended distribution model. Whether CUDA-Q packages and Python wheels will ship this feature by default, whether GMP/MPFR will be dynamically linked or bundled. If bundled where the required third-party notices/license texts will be carried, and how LGPL relinkability/source obligations are handled for wheels, containers, and installers? If not bundled what will the installation/dependency story be for users?

The feature is no longer optional. CUDAQ_ENABLE_CLIFFORD_T_SYNTHESIS has been removed entirely. The clifford-t-synthesis pass and the cudaq-synth Python module are unconditional part of every CUDA-Q build. So all the distributed artifacts (containers, installer, wheels) ship rotation synthesis by default.

Here is the distribution model
1. `install_prerequisites.sh` builds GMP and MPFR from source as shared libraries only.
2. These are dynamically linked. cmake installs the shared libraries and soname symlinks into the CUDA-Q `lib/` directory.
3. Container and the installer inherits them from the install tree.
4. Wheels pick them up through the exisiting auditwheel step.

Note: Since the libraries are dynamically linked and replaceable by substituting the shared library file and we distribute them with source pointers to `gmplib.org` and `mpfr.org`, I think the LGPL v3 relinking and source obligations are met without carrying source in the artifacts. Having said that we are building a container which include source of these third-party libs to comply with the security requirements.

CMakeLists.txt
sacpis: This option is gone. Rotation synthesis is now always on. That means a build without GMP/MPFR fails at the configuration time. We do provide instructions for the user.

cudaq/lib/Synthesis/CMakeLists.txt
sacpis: Fixed. The `.a` suffix override is removed. Discovering these libs on all platforms is done through `CMAKE_SHARED_LIBRARY_SUFFIX`.

cudaq/lib/Synthesis/CMakeLists.txt
sacpis: Fixed. Same as above comment.

cudaq/include/cudaq/Synthesis/Math/Integer.h
sacpis: Thanks. Fixed by drawing the API boundary explicitly. The `cudaq/Synthesis/*` headers are now internal. Have deleted their install rules. GMP and MPFR include dirs are `$<BUILD_INTERFACE:>` only. Also have linked the libs as PRIVATE. Made sure that no installed headers include `gmp.h` or `mpfr.h`. That ensures downstream users would never need GMP/MPFR headers. Right now, we have the public surfaces as the `clifford-t-synthesis` pass and `cudaq.synth.gridsynth` in Python. I am thinking if we later want a public C++ Synthesis API, that would be a shim layer, no?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Sure. Here is what is happening now. `cudaq-synth` is built and linked into `OptTransforms` (and `cudaq-mlir-runtime`) unconditionally. We have got rid of `CUDAQ_HAS_CLIFFORD_T_SYNTHESIS` and the `build without the synthesis library` runtime stubs. This registered pass will always be on by default.

python/runtime/synth/CMakeLists.txt
sacpis: Yes. `cudaq-synth` is a static library that is linked into the consuming binaries, which reference `lbgmp` and/or `libmpfr` dynamically. Those shared libs are bundled. cmake installs them into the CUDA-Q `lib/` dir, container and installer inherit that tree, and wheel repair (auditwheel, delocate) vendors, and re-rpaths them for wheels. They are not system dependencies and not pip runtime dependencies.

python/cudaq/synth/__init__.py
sacpis: Now `_cudaq_synth` is built unconditionally, so the package and its native module always ship together.

cudaq/lib/Synthesis/CMakeLists.txt
sacpis: This scenario is gone now. The GMP/MPFR dependent headers are no longer installed. So downstream consumers can't compile against them. The include dirs are build-interface-only. GMP/MPFR remain PRIVATE on `cudaq-synth`.

scripts/install_prerequisites.sh
```diff
@@ -14,7 +14,7 @@
 #
```
taalexander, line 14: why is this duplicated with bootstrap_prequistes?

taalexander:
Thank you @sacpis for your changes. I've been going through them to understand compatiability with the LGPL license.

I am ok with the approach you had added flagging the library build and think it is ok to include if it is enabled by default with the resolved distribution story. This is ultimately up to you though if we wish to flag it.

I've been using https://licensecheck.io/blog/lgpl-dynamic-linking as a good reference for the requirements to dynamically link our code to GMP/MPFR and what the consequences are for downstream users.

The cores requirements being:

```
LGPL Requirements for Dynamic Linking
When dynamically linking to LGPL libraries, you must:

- Provide LGPL license text with your distribution
- Include copyright notices for the LGPL library
- Enable library replacement by users
- Provide interface information if using modified LGPL code

You do NOT need to:
- Release your source code
- License your code under LGPL
- Provide build scripts for your application
```

Checking against this what I don't see right now are:
- The distribution artifacts (wheels, cpp installer and container) must have the full license texts /notices. It's not sufficient to have the current NOTICE entry alone AFAIK
- We must document how GMP/MPFR can be replaced in the distribution to demonstrate that it is possible. Found an example in Shapely's installation - https://shapely.readthedocs.io/en/latest/installation.html.
- We likely need some integration tests to verify these properties (eg., verify licenses are shipped with, that we can replace libraries and that all links to libraries are dynamic) and that we are not accidentally shipping distributions in violation of the license.  

As I am far from an expert at this I think we should also run this by our open source team (cc @mitchdz)

scripts/install_prerequisites.sh
sacpis: Yeah, one is being used by gcc and another by llvm builds.

taalexander: Hmm, that's too bad. My preference would be to avoid duplication by extracting as script to call but not required.

sacpis: I know. I had the same feeling. Will keep this item to fix as P1.

sacpis:
> Thank you @sacpis for your changes. I've been going through them to understand compatiability with the LGPL license.
> 
> I am ok with the approach you had added flagging the library build and think it is ok to include if it is enabled by default with the resolved distribution story. This is ultimately up to you though if we wish to flag it.
> 
> I've been using https://licensecheck.io/blog/lgpl-dynamic-linking as a good reference for the requirements to dynamically link our code to GMP/MPFR and what the consequences are for downstream users.
> 
> The cores requirements being:
> 
> ```
> LGPL Requirements for Dynamic Linking
> When dynamically linking to LGPL libraries, you must:
> 
> - Provide LGPL license text with your distribution
> - Include copyright notices for the LGPL library
> - Enable library replacement by users
> - Provide interface information if using modified LGPL code
> 
> You do NOT need to:
> - Release your source code
> - License your code under LGPL
> - Provide build scripts for your application
> ```
> 
> Checking against this what I don't see right now are:
> 
> * The distribution artifacts (wheels, cpp installer and container) must have the full license texts /notices. It's not sufficient to have the current NOTICE entry alone AFAIK
> * We must document how GMP/MPFR can be replaced in the distribution to demonstrate that it is possible. Found an example in Shapely's installation - https://shapely.readthedocs.io/en/latest/installation.html.
> * We likely need some integration tests to verify these properties (eg., verify licenses are shipped with, that we can replace libraries and that all links to libraries are dynamic) and that we are not accidentally shipping distributions in violation of the license.
> 
> As I am far from an expert at this I think we should also run this by our open source team (cc @mitchdz)

Hey @taalexander, Thank you for reviewing this PR and providing feedback. Have made the requested changes.

sacpis:
>> I am ok with the approach you had added flagging the library build and think it is ok to include if it is enabled by >> default with the resolved distribution story. This is ultimately up to you though if we wish to flag it.

I don't have a valid reason to keep the flag when we already have the libs available.

scripts/build_cudaq.sh
```diff
@@ -340,6 +340,8 @@ fi
 
 cp "$repo_root/LICENSE" "$CUDAQ_INSTALL_PREFIX/LICENSE"
 cp "$repo_root/NOTICE" "$CUDAQ_INSTALL_PREFIX/NOTICE"
+mkdir -p "$CUDAQ_INSTALL_PREFIX/LICENSES"
+cp "$repo_root/LICENSES/"* "$CUDAQ_INSTALL_PREFIX/LICENSES/"
```
taalexander: I think we should do this via cmake instead so this covers the path where build_cudaq.sh is used if possible.

scripts/validate_license_compliance.sh
```diff
@@ -0,0 +1,245 @@
...
+        "$install_root/bin/cudaq-opt" --clifford-t-synthesis='epsilon=1e-3' \
+            "$smoke_dir/rz.qke" 2>/dev/null | grep -q "__cliffordt_rz_"
+    }
+else
+    synthesis_smoke_test() {
+        python3 -c "import cudaq; gates = cudaq.synth.gridsynth(0.7853981633974483, 1e-3); assert 'T' in gates, gates" 2>/dev/null
+    }
+fi
+
+# Back up the shipped libgmp files and guarantee they are restored.
```
taalexander: Should we also test mpfr?

docs/sphinx/using/install/local_installation.rst
```diff
@@ -919,6 +919,35 @@ Detailed information about supported drivers for different CUDA versions and be
...
+
+CUDA-Q links to GMP and MPFR exclusively dynamically, and the two libraries
+can be replaced with compatible versions without rebuilding CUDA-Q by
+substituting the shared library files (`libgmp.so*` and `libmpfr.so*` on
+Linux, `libgmp*.dylib` and `libmpfr*.dylib` on macOS):
+
+- Docker image and installer - the libraries are located in the `lib`
+  folder of the CUDA-Q installation directory, that is
+  `${CUDA_QUANTUM_PATH}/lib`.
+- Python wheels - the libraries are located in the `lib` folder that is
```
taalexander, line 942: Can you confirm this path matches the final repaired wheel layout after `auditwheel`/`delocate`?

taalexander (commented):
Thanks Sachin, this addresses most of my packaging review comments. I will move onto the API surface now.

scripts/build_cudaq.sh
sacpis: Oh, I just followed the process how the root license is copied on line 341. But will put it in cmake.

scripts/validate_license_compliance.sh
sacpis: Good catch. Yes, definitely. I missed this.

scripts/build_cudaq.sh
taalexander: Ahh, missed that. It might make sense to put that in cmake as well but it's up to you?

sacpis: For now, will leave the old stuff as it is so that we don't break anything with respect to the security review for the same. Fixed the new one in 71cc832.

scripts/validate_license_compliance.sh
sacpis: Fixed in 71cc832.

docs/sphinx/using/install/local_installation.rst
sacpis: Good catch. Fixed in 71cc832.

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,97 @@
...
+      nanobind::arg("epsilon"), nanobind::arg("diophantine_timeout_ms") = 200,
+      nanobind::arg("factoring_timeout_ms") = 50,
+      R"doc(Synthesize a Clifford+T circuit approximating R_z(theta) to precision epsilon.
+
+Implements the grid-synthesis algorithm of Ross & Selinger (arXiv:1403.2975,
+Algorithm 7.6). The returned gate string is in Matsumoto-Amano normal form
+with minimum T-count.
```
taalexander:
```suggestion
Implements the grid-synthesis algorithm of Ross & Selinger (arXiv:1403.2975,
Algorithm 7.6). The returned gate string is in Matsumoto-Amano normal form
with minimum T-count up to search timeouts. 
```

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+Precision is measured in the operator norm (a.k.a. spectral norm, the
+induced 2-norm ||A|| = sigma_max(A)). The synthesized unitary U satisfies
+||R_z(theta) - U|| <= epsilon. This is the norm used in Ross & Selinger
+section 7.1, equation (13).
+
+Args:
+    theta: Target rotation angle (float, or str for arbitrary precision).
```
taalexander:
```suggestion
    theta: Target rotation angle (float, or decimal str for arbitrary precision).
```

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <nanobind/stl/variant.h>
+
+#include <string>
+#include <variant>
+
+namespace {
+
+using RealArg = std::variant<double, std::string>;
+
+cudaq::synth::Real toReal(const RealArg &arg) {
```
taalexander: What happens if the string is invalid?

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,97 @@
+/*******************************************************************************
```
taalexander, line 1: Could `gridsynth` return a small sequence object instead of a bare gate string? `str(result)` could preserve the current representation. It could also support a method to build a Cuda-Q kernel for the  sequence? I imagine a `CliffordTSequence` object.

cudaq/include/cudaq/Synthesis/Circuit/Circuit.h
```diff
@@ -0,0 +1,191 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
```
taalexander: Copyright should be updated to 2026 only (elsewhere in the PR too).

cudaq/include/cudaq/Synthesis/Math/Integer.h
```diff
@@ -0,0 +1,680 @@
...
+//     library ABI demands; the casts are lossless on LP64 (Linux).
+
+namespace cudaq::synth {
+
+//===----------------------------------------------------------------------===//
+// Integer
+//===----------------------------------------------------------------------===//
+
+/// Arbitrary-precision integer backed by GMP's mpz_t.
+class Integer {
```
taalexander, line 47: I'm happy there's a composition wrapper. Makes the potential replacement/lgpl licensing story more clear.

taalexander (changes requested):
Phase 2 of review - Focused on public interfaces.

I really like the ring implementations/numerics implementations. It makes it much more simple to understand the core components.

The main API question is whether we intend to distribute the full synthesis header tree as supported CUDA-Q API, or keep a smaller user-facing surface.

For the user API, should we expose algorithm-specific functions directly, or provide an operation-level entry point such as `cudaq.synth.approximate_rz(...)` that uses GridSynth today? Algorithm-specific calls could then live under `cudaq.synth.algorithms`.

That would give us room to add or change synthesis methods without changing existing user code, and would let us limit the installed C++ headers to the types users actually need.

python/runtime/synth/py_synth.cpp
sacpis: Handled in 42356b7.

cudaq/include/cudaq/Synthesis/Circuit/Circuit.h
sacpis: Fixed in 3eb0779.

python/runtime/synth/py_synth.cpp
sacpis: Fixed in 96f1b23.

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -165,6 +165,50 @@ def BasisConversion : Pass<"basis-conversion", "mlir::ModuleOp"> {
...
+    (Ross & Selinger, arXiv:1403.2975) provided by the embedded
+    `cudaq::synth` library.
+
+    Preconditions (must be enforced by upstream passes):
+    - Rotation angles are folded to constants - run `ArgumentSynthesis` and
+      constant-propagation passes first.
+    - Controls and adjoints are materialized - run `ApplyOpSpecialization`
+      first. The omega global-phase contributions are dropped on the
+      assumption that no consumer wraps the kernel with `cudaq::control`
+      after this pass runs.
```
taalexander, line 191: FYI @cabreraam this will need to be updated when the global phase work is landed. @sacpis could we add a TODO (or follow on issue) to fix this when global phase lands

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -165,6 +165,50 @@ def BasisConversion : Pass<"basis-conversion", "mlir::ModuleOp"> {
...
+
+    The pass is gated at build time on `CUDAQ_HAS_CLIFFORD_T_SYNTHESIS`.
+    When the synthesis library is not available, `runOnOperation` emits a
+    fatal diagnostic.
```
taalexander: I don't think this is true anymore right?

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -165,6 +165,50 @@ def BasisConversion : Pass<"basis-conversion", "mlir::ModuleOp"> {
...
+      /*default=*/"200", "Per-candidate Diophantine solver timeout (ms).">,
+    Option<"factoringTimeoutMs", "factoring-timeout-ms", "int32_t",
+      /*default=*/"50", "Per-candidate integer-factoring timeout (ms).">,
+    Option<"retryCount", "retry-count", "int32_t", /*default=*/"3",
+      "Retry attempts with doubled timeouts on gridsynth failure.">,
+    Option<"onDynamicAngle", "on-dynamic-angle", "std::string",
+      /*default=*/"\"error\"",
+      "Behavior on non-constant rotation angles: 'error' or 'skip'.">,
+    Option<"skipBelow", "skip-below", "double", /*default=*/"1e-12",
+      "Skip rotations whose |theta| is below this threshold.">,
```
taalexander: Do we skip or do we remove? If we don't remove is there a pruning pass that will?

cudaq/include/cudaq/Optimizer/Transforms/Passes.h
```diff
@@ -54,6 +55,16 @@ void addDecomposition(mlir::OpPassManager &pm,
...
 
+/// Clifford+T fault-tolerant synthesis sub-pipeline
+/// UnitarySynthesis
+/// ApplyOpSpecialization
+/// constant propagation
+/// CliffordTSynthesis
+/// Decomposition to the {H, S, T, X, Z, CNOT} basis
+/// Opt-in only.
+/// This helper is not added to default target pipelines.
+void addCliffordTSynthesis(mlir::OpPassManager &pm, double epsilon = 1e-10);
```
taalexander, line 76: Is the intent of this pipeline to be run from a specific stage in the pass pipeline or for testing only? I could see some of the passes being duplicated if not just for testing but for testing I think this is ok.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+#define DEBUG_TYPE "clifford-t-synthesis"
+
+using namespace mlir;
+
+namespace {
+
+struct RotationOptions {
+  double epsilon;
+  int32_t diophantineTimeoutMs;
+  int32_t factoringTimeoutMs;
```
taalexander, line 42: values are int32-t might be surprising? It might be more natural to have these as floats?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+      cudaq::quake::SOp::create(b, loc, ValueRange{qubit});
+      break;
+    case cudaq::synth::Gate::T:
+      cudaq::quake::TOp::create(b, loc, ValueRange{qubit});
+      break;
+    case cudaq::synth::Gate::X:
+      cudaq::quake::XOp::create(b, loc, ValueRange{qubit});
+      break;
+    case cudaq::synth::Gate::W:
+      // omega = e^{i*pi/4} global phase - dropped.
```
taalexander, line 119: TODO here for phase inclusion.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+  FloatAttr attr;
+  if (!matchPattern(angleVal, m_Constant<FloatAttr>(&attr))) {
+    if (opts.onDynamicAngle == "skip") {
+      op->emitRemark("clifford-t-synthesis: skipping non-constant rotation "
+                     "angle (--on-dynamic-angle=skip)");
+    } else {
+      op->emitError("clifford-t-synthesis: rotation angle is not a "
+                    "compile-time constant. Fold the angle upstream or set "
+                    "--on-dynamic-angle=skip to leave the op alone.");
```
taalexander, line 250: Should this signal pass failure?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  return {PreCheck::Action::Lower, theta};
+}
+
+struct RzPattern : OpRewritePattern<cudaq::quake::RzOp> {
+  RzPattern(MLIRContext *ctx, RotationOptions opts, SynthState *state)
+      : OpRewritePattern(ctx), opts(std::move(opts)), state(state) {}
+
+  LogicalResult matchAndRewrite(cudaq::quake::RzOp op,
```
taalexander, line 278: Why do we not handle adjoint operations? Are these resolved into angles at this point? If this is the case should we emit an error?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+struct SynthState {
+  mlir::ModuleOp module;
+  llvm::DenseMap<uint64_t, mlir::FlatSymbolRefAttr> cache;
+  uint64_t hits = 0;
+};
+
+// Materialize a body of Clifford+T gates onto `qubit` from a synthesized
+// `Circuit`.
+void emitCircuitBody(OpBuilder &b, Location loc, Value qubit,
```
taalexander: I believe the free functions here should be static as per the LLVM coding standards?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+    cudaq::quake::RzOp::create(rewriter, loc, ValueRange{angle}, ValueRange{},
+                               ValueRange{target});
+    rewriter.eraseOp(op);
+    return success();
+  }
+
+  RotationOptions opts;
+};
+
+} // namespace
```
taalexander: Why do we include these transformations if we already require prelude passes to be run? It's minimal but is duplication logic as we already support gateset decomposition in the compiler. (Eg., could we force the gateset to be Rz+Clifford only by this point and then only support Rz?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+        return WalkResult::advance();
+      });
+      if (walk.wasInterrupted()) {
+        signalPassFailure();
+        return;
+      }
+    }
+
+    auto prec = static_cast<mpfr_prec_t>(
+        std::max<double>(64.0, std::ceil(-std::log2(epsilon) * 4.0 + 64.0)));
```
taalexander: Where does this equation come from?

cudaq/test/Transforms/CliffordTSynthesis/skip_tiny.qke
```diff
@@ -0,0 +1,27 @@
...
+
+// RUN: cudaq-opt --clifford-t-synthesis='epsilon=1e-3' %s | \
+// RUN:   FileCheck %s --implicit-check-not='quake.r{{[xyz1]}}' \
+// RUN:                --implicit-check-not='quake.{{[hstx]}}'
+
+// |theta| = 1e-15 is below the default --skip-below=1e-12, so the rotation
+// is erased. The function body must end up with only the alloca and return.
+// No rotation ops and no Clifford+T gates.
+
+func.func @tiny() {
```
taalexander, line 18: I wonder if small rotations should be removed through a separate pass available to users?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+    cudaq::synth::Real::set_default_precision(prec);
```
taalexander: AI caught this - Is this changing a global? I worry this won't be later reset?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  cudaq::synth::Real thetaReal(theta);
+  cudaq::synth::Real epsilonReal(opts.epsilon);
+  llvm::FailureOr<cudaq::synth::Circuit> circuit = llvm::failure();
+  for (int32_t attempt = 0; attempt <= opts.retryCount; ++attempt) {
+    circuit = cudaq::synth::gridsynth(
+        thetaReal, epsilonReal,
+        static_cast<int32_t>(opts.diophantineTimeoutMs << attempt),
+        static_cast<int32_t>(opts.factoringTimeoutMs << attempt));
```
taalexander: AI caught this - These are missing validation?

cudaq/unittests/Optimizer/CliffordTSynthesisTester.cpp
```diff
@@ -0,0 +1,395 @@
+/*******************************************************************************
```
taalexander, line 1: I would limit the unittests to:
- Numerical correctness
- Approximation tolerance
- - Rotation and adjoint coverage

and otherwise relay on lit tests.

taalexander (changes requested):
Phase 3 review - focused on the MLIR infrastructure.

This should be a value-semantics pass, or support both value and reference semantics as by this stage in compilation/optimizations predominantly all sequences will be in value-semantics and will incur a large-cost to go between. 

Could the synthesized helper consume and return a `!quake.wire`, threading each gate result through the helper? Reference targets could remain supported separately.

I think we might be missing a test which shows the full expected lowering pipeline `cudaq-fault-tolerant-target` for an example?

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
sacpis: Yes, right. Sorry, forgot to remove the text.

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
sacpis: Fixed in fb32fa4.

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
sacpis: Fixed in c758bdb.

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
sacpis: We remove it. A sub-threshold rotation is erased and treated as identity in this pass. Have updated the help text to state that explicitly and to document that `skip-below` disables the behavior.

I kept option name `skip-below` as is. I am leaning towards making it more clear with names as erase-below or identity-threshold. Let me know if you have any preference.

Addressed the original comment with 7a629ec.

cudaq/include/cudaq/Optimizer/Transforms/Passes.h
sacpis: Fixed in da6c193.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Deliberately kept it as `int32_t` to match the underlying gridsynth API.

```
llvm::FailureOr<Circuit> gridsynth(
    const Real &theta, const Real &epsilon,
    int32_t diophantine_timeout_ms = details::DEFAULT_DIOPHANTINE_TIMEOUT_MS,
    int32_t factoring_timeout_ms = details::DEFAULT_FACTORING_TIMEOUT_MS);
```

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Added with 9aa9fc8.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Yes, it should. Seems like a bug as the pass emits an error yet reports success.

sacpis: Fixed in 0406383.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Adjoint rotation carry an `UnitAttr:$is_adj`, and the pass was silently ignoring it. As per the precondition, `ApplyOpSpecialization` is supposed to fold adjoints into negated angles upstream, but nothing enforced it. So have added a hard-error check in bbe953d that rejects any adjoint rotation reaching the pass.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Fixed in f9d567e.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Have added a TODO comment in the code explaining that `Rx/Ry/R1` patterns exist because the shared decomposition library has no `Rx to Rz` and `Ry to Rz` path. It only has Rx/Ry to PhasedRx. So there is no way to force Rz only gateset upstream today.

Will create a follow up PR for this TODO.

sacpis: Added a TODO in 62763a7.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Added a comment explaining this empirical precision heuristic in 2c59ae7. Would be happy to replace this if we have a better one.

cudaq/test/Transforms/CliffordTSynthesis/skip_tiny.qke
sacpis: Added a TODO in 8ae7996.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Good catch. Fixed by AI in a2dcb30.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Fixed by AI in c09060e.

cudaq/unittests/Optimizer/CliffordTSynthesisTester.cpp
sacpis: Fixed in 07bb5cd.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
taalexander: Thanks Sachin, a follow-up PR is perfect.

sacpis:
> This should be a value-semantics pass, or support both value and reference semantics as by this stage in compilation/optimizations predominantly all sequences will be in value-semantics and will incur a large-cost to go between.
> 
> Could the synthesized helper consume and return a `!quake.wire`, threading each gate result through the helper? Reference targets could remain supported separately.

Addressed in f86a9ed.

sacpis:
> I think we might be missing a test which shows the full expected lowering pipeline `cudaq-fault-tolerant-target` for an example?

Added in 4587f4a.

sacpis:
> would let us limit the installed C++ headers to the types users actually need.

Seems like it implies some installed C++ synthesis headers. But for now the only user facing surface is Python (with `cudaq.synth`), and every C++ consumer of the `Synthesis/` headers builds against the source tree. That's why I was thinking of Python only, which will exclude the entire `Synthesis/` tree from install expose no supported C++ API.

> provide an operation-level entry point such as cudaq.synth.approximate_rz(...)

Is `cudaq.synth.approximate_rz(...)` the name you want, or prefer something like `synthesize_rz` / `decompose_rz` or somehting else?

> Algorithm-specific calls could then live under cudaq.synth.algorithms

Do we intend the algorithm-specific calls to be a supported public surface or an advanced or escape hatch below the stable `approximate_rz`?

sacpis:
> /ok to test cfab788

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,374 @@
...
+  double theta = attr.getValueAsDouble();
+  if (!std::isfinite(theta)) {
+    op->emitError("clifford-t-synthesis: rotation angle is NaN or Inf");
+    *hadHardError = true;
+    return {PreCheck::Action::LeaveInPlace, 0.0};
+  }
+
+  // gridsynth is never invoked for an angle that would be erased anyway.
+  if (std::abs(theta) < opts.skipBelow) {
+    rewriter.eraseOp(op);
```
taalexander: I think there's a pretty critical bug here if this value flows through another wire (likely will cause a crash). It should be forwarded to the user.

python/runtime/synth/py_synth.cpp
```diff
@@ -0,0 +1,111 @@
...
+  if (!(epsilonReal > 0))
+    throw nanobind::value_error("epsilon must be strictly positive");
+
+  llvm::FailureOr<cudaq::synth::Circuit> result = cudaq::synth::gridsynth(
+      thetaReal, epsilonReal, diophantine_timeout_ms, factoring_timeout_ms);
+  if (llvm::failed(result))
+    throw nanobind::value_error(
+        "gridsynth: failed to synthesize a Clifford+T approximation "
+        "(degenerate epsilon region or search exhausted)");
+  return result->to_string();
```
taalexander, line 74: Should this release the gil while calling since it can block?

cudaq/include/cudaq/Synthesis/Math/Real.h
```diff
@@ -0,0 +1,839 @@
...
+    return value;
+  }
+  static const Real &pi() {
+    static const Real value = []() {
+      Real v;
+      mpfr_const_pi(v.get_mpfr(), MPFR_RNDN);
+      return v;
+    }();
+    return value;
+  }
```
taalexander, line 433: AI noted these won't pick up precision changes in the global which could be an issue (same with other constants likely).

cudaq/lib/Synthesis/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+   NOT MPFR_INCLUDE_DIR OR NOT MPFR_LIBRARIES)
+  message(FATAL_ERROR
+    "The GMP/MPFR shared libraries were not found. They are required to "
+    "build CUDA-Q (Clifford+T rotation synthesis). Run "
+    "scripts/install_prerequisites.sh to build them from source, or install "
+    "them via your package manager (e.g. libgmp-dev and libmpfr-dev on apt) "
+    "and/or set GMP_INSTALL_PREFIX / MPFR_INSTALL_PREFIX.")
+endif()
+
+add_cudaq_library(cudaq-synth
```
taalexander, line 28: AI-generated review comment:

`add_cudaq_library()` installs this target into the `CUDAQTargets` export set, but the installed `CUDAQTargets.cmake` is generated from the separate `cudaq-targets` export set. As a result, the library artifact may be installed without defining `cudaq::cudaq-synth` for consumers of `find_package(CUDAQ)`.

Is `cudaq-synth` intended to be a supported external C++ target? If so, could it be added to the installed export set and covered by an install-tree consumer test? If it is internal only, could that boundary be made explicit so the installed headers do not imply a supported standalone C++ library?

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
```diff
@@ -0,0 +1,374 @@
...
+}
+
+// Materialize a body of Clifford+T gates onto `qubit` from a synthesized
+// `Circuit`, threading value-semantics wires when `qubit` is a `!quake.wire`.
+// Returns the final qubit value (the last wire for value semantics, or `qubit`
+// unchanged for memory semantics).
+static Value emitCircuitBody(OpBuilder &b, Location loc, Value qubit,
+                             const cudaq::synth::Circuit &circuit) {
+  Value cur = qubit;
+  for (cudaq::synth::Gate g : circuit) {
```
taalexander: `Circuit` stores gates in matrix-product order, but his loop emits them in forward execution order,producing the reversed product. Current Rz synthesis
  masks this because the gates and Rz are symmetric.

cudaq/unittests/Synthesis/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+include(GoogleTest)
+
+function(add_synthesis_test target_name source_file)
+  add_executable(${target_name} ${source_file})
```
taalexander, line 12: These test executables bypass  `add_cudaq_unittest()`, so none is added as a dependency of the standard `CudaQUnitTests` target?

taalexander (changes requested):
A few more issues discovered. Still looking at the core algorithm. Thank for splitting out the gate removal.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Fixed in 2017121.

cudaq/lib/Optimizer/Transforms/CliffordTSynthesis.cpp
sacpis: Fixed in a7ee052.

python/runtime/synth/py_synth.cpp
sacpis: Fixed in 06fda60.

cudaq/include/cudaq/Synthesis/Math/Real.h
sacpis: Fixed in 1dbb01e.

cudaq/lib/Synthesis/CMakeLists.txt
sacpis: Fixed in 07ec200.

cudaq/unittests/Synthesis/CMakeLists.txt
sacpis: Fixed in c82b363.

sacpis:
> A few more issues discovered. Still looking at the core algorithm. Thank for splitting out the gate removal.

Thanks @taalexander. Have addressed the feedback.

cudaq/include/cudaq/Synthesis/Math/Unitary.h
```diff
@@ -0,0 +1,321 @@
...
+
+/// Operator-norm proxy for |R_z(theta) - U|.
+///
+/// Reference: Ross & Selinger, arXiv:1403.2975, equation (13).
+///
+/// Builds E = U - R_z(theta) entry by entry from the complex matrix of u
+/// (via `to_complex_matrix`) and returns `sqrt`(|`det`(E)|). For small epsilon
+/// the two singular values of E are approximately equal, so this proxy is
+/// accurate in the regime the synthesizer targets.
+inline Real rz_approximation_error(const DOmegaUnitary &u, const Real &theta) {
```
taalexander, line 247: This computes

$$
\sqrt{|\det(E)|},
$$

which is not the spectral norm for an arbitrary circuit. For example,
`rz_gate_sequence_error("0", "T")` returns `0`, while the true error is
approximately `0.7653669`. Since this is a 2-by-2 matrix, we can compute the
exact spectral norm cheaply as

$$
\sqrt{\lambda_{\max}(E^\dagger E)}.
$$

Could we make that change and add this example
as a regression?

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
```diff
@@ -0,0 +1,398 @@
...
+  ///   A x^2 + 2B xy + C y^2 + Dx x + Dy y <= 1
+  ///
+  /// Returns failure() if Ellipse::create rejects the parameters; this
+  /// should not happen for epsilon > 0.
+  static llvm::FailureOr<Ellipse>
+  make_bounding_ellipse(const Real &z_x, const Real &z_y,
+                        const Real &dot_threshold, const Real &epsilon) {
+    Real inv_eps2 = 1 / (epsilon * epsilon);
+    Real inv_eps4 = inv_eps2 * inv_eps2;
+    Real lambda_x = 64 * inv_eps4;       // D2(0, 0)
```
taalexander: These ellipse constants are also based on the smaller region above. Changing only the threshold still returned the same four-T circuit. Updating the ellipse with it returned identity. Could we rederive these constants for the corrected region?

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
```diff
@@ -0,0 +1,398 @@
...
+  };
+
+  /// Precompute the three scalars that depend on (theta, epsilon):
+  ///   dot_threshold = sqrt(1 - epsilon^2/4) = cos(arcsin(epsilon/2))
+  ///   z             = (cos(-theta/2), sin(-theta/2))
+  /// Uses mpfr_sin_cos to get both trig values from a single argument.
+  static Precomputed compute_precomputed(const Real &theta,
+                                         const Real &epsilon) {
+    Precomputed pre;
+    pre.dot_threshold = sqrt(1 - ((epsilon * epsilon) / 4));
```
taalexander: [Ross-Selinger](https://arxiv.org/pdf/1403.2975) equation (13) defines the epsilon-region using

$$
\widetilde z\cdot\widetilde u
\geq 1-\frac{\epsilon^2}{2},
$$

but this uses

$$
\sqrt{1-\frac{\epsilon^2}{4}},
$$

which searches a smaller region and can add unnecessary T gates. For example,
asking gridsynth to approximate `Rz(0.5)` within error `0.3`:

```python
seq = cudaq.synth.gridsynth(0.5, 0.3)
print(str(seq), seq.t_count)
```

returned:

```text
HTSHTSHTHTSHXSWWWWWW 4
```

However, applying no gates already meets the requested tolerance:

$$
\lVert R_z(0.5)-I\rVert_2
=2\sin(0.5/4)
\approx 0.24935 < 0.3.
$$

The minimum-T result should therefore be:

```text
I 0
```

I tried the threshold change in an isolated rebuild, and by itself it still
returned the same four-T circuit because the enclosing ellipse remained sized
for the old cap. Updating both the threshold and the ellipse (see comment below) returned `I 0` for
this input, and all six committed synthesis test executables still passed.
Could we update both pieces and add this case (are there others?) as a regression test?

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+} // namespace
+
+namespace cudaq::synth {
+
+//===----------------------------------------------------------------------===//
+// gridsynth_unitary
+//===----------------------------------------------------------------------===//
+
+llvm::FailureOr<DOmegaUnitary> gridsynth_unitary(const Real &theta,
```
taalexander, line 319: The synthesis entry point does not handle loose or invalid `epsilon` values
consistently. Ross-Selinger Lemma 7.2 shows that a zero-T Clifford solution
always exists when:

$$
\epsilon\geq\left|1-e^{i\pi/8}\right|
\approx0.39018.
$$

Could we validate that `epsilon` is finite and positive at the C++ boundary,
then use the paper's construction to return the nearest zero-T Clifford above
this threshold instead of entering the grid search? For example:

```python
seq = cudaq.synth.gridsynth(0.5, 0.4)
print(str(seq), seq.t_count)
```

should return:

```text
I 0
```

At minimum, values with `epsilon >= 2` should be handled immediately (eg., I 0) because

$$
\lVert U-V\rVert_2\leq2
$$

for any two unitaries. Tests should cover zero, negative, NaN, infinity, the
Lemma 7.2 threshold, `2`, and values greater than `2` through both the C++ and
Python APIs.

cudaq/lib/Synthesis/Circuit/Clifford.h
```diff
@@ -0,0 +1,451 @@
...
+  {0, 0, 1, 0},
+  {0, 1, 0, 0},
+  {0, 1, 1, 6},
+  {0, 1, 2, 4},
+  {0, 1, 3, 2},
+  {2, 0, 0, 0},
+  {1, 0, 1, 2},
+  {2, 1, 0, 0},
+  {1, 1, 3, 4},
+  {2, 1, 1, 2},
```
taalexander: AI caught. For the Clifford represented by

$$
(a,b,c,d)=(1,1,0,0),
$$

calling `inv()` and multiplying the result back does not produce identity. The
table returns `{2,1,1,2}`, while literal matrices require `{2,1,2,2}`. Could we
correct this row and add one finite exhaustive test over all 192 inputs that
expects

$$
CC^{-1}=C^{-1}C=I
$$

for every element? This appears minor because the current synthesis path only
uses `inv()` for fixed `H` and `SH` values and does not reach the defective
row.

cudaq/include/cudaq/Synthesis/Math/Integer.h
```diff
@@ -0,0 +1,680 @@
...
+  Integer result(rhs);
+  result += lhs;
+  return result;
+}
+inline Integer operator-(const Integer &lhs, int64_t rhs) {
+  Integer result(lhs);
+  result -= rhs;
+  return result;
+}
+inline Integer operator-(int64_t lhs, const Integer &rhs) {
```
taalexander, line 490: AI caught - `1 - (Integer(1) << 100)` returns `1` instead of

$$
-1267650600228229401496703205375,
$$

because the arbitrary-precision right operand is converted to `int64_t` before
subtraction. Could this overload use exact GMP arithmetic and add the expression
above as a regression? I did not find a current production caller of this
mixed overload, so this also appears minor.

taalexander (changes requested):
The public API accepts a synthesis error tolerance but Python only returns the
gate sequence, so users cannot inspect the achieved error or evaluate a saved
sequence. Could we correct the existing (see my other comment) C++ helper to compute the spectral norm for every accepted sequence and expose the same operation in Python, for example `cudaq.synth.rz_error(theta, sequence)`?

This way the user can ensure `rz_error(theta, gridsynth(theta, epsilon)) <= epsilon` (or check how much error there was).

cudaq/include/cudaq/Synthesis/Math/Integer.h
sacpis: Fixed in 12f8e58.

cudaq/lib/Synthesis/Circuit/Clifford.h
sacpis: Fixed in 128abea.

cudaq/include/cudaq/Synthesis/Math/Unitary.h
sacpis: Fixed in 624c397.

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
sacpis: Fixed in 474c1f3.

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
sacpis: Fixed in 1a5ad50.

cudaq/lib/Synthesis/Synthesis/Gridsynth.cpp
sacpis: Fixed in 1a5ad50.

sacpis:
> The public API accepts a synthesis error tolerance but Python only returns the gate sequence, so users cannot inspect the achieved error or evaluate a saved sequence. Could we correct the existing (see my other comment) C++ helper to compute the spectral norm for every accepted sequence and expose the same operation in Python, for example `cudaq.synth.rz_error(theta, sequence)`?
> 
> This way the user can ensure `rz_error(theta, gridsynth(theta, epsilon)) <= epsilon` (or check how much error there was).

Thanks @taalexander for a thorough review here. Have pushed the fixes.

taalexander (approved):
LGTM. Thank you for your hard work @sacpis and @boschmitt.

1tnguyen (approved):
Reviewed the workflow updates. LGTM 👍

PR #4515, opened 2026-05-13
[photonics] Guard `finalizeExecutionContext` against empty `sampleQudits`

khalatepradnya:
### Summary

Fix a SIGSEGV in `PhotonicsExecutionManager::finalizeExecutionContext` when an `orca-photonics` kernel raises before reaching any `mz()`.

### Motivation

`pytest -n auto python/tests/handlers/test_photonics_kernel.py` crashes intermittently in CI - latest sighting on arm64-llvm. Standalone reproducer (deterministic):
``` 
$ pytest python/tests/handlers/test_photonics_kernel.py::test_qudit_level
    Fatal Python error: Segmentation fault    
```

The faulting line is `qpp::sample(shots, state, ids, sampleQudits.begin()->levels)`: `sampleQudits` is empty, `_M_start` is nullptr, the load of `->levels` segfaults.

### Why it suddenly popped up

The bare dereference has been in the photonics handler since the original landing - latent. 
Two things had to coincide for it to become reachable:

1. PR #4051 made `ExecutionContext.__exit__` call `finalizeExecutionContext` + `endExecution` on the exception path. Before that, finalize was skipped when the kernel threw, so the latent deref never ran   on throw-before-mz kernels.

2. `pytest-xdist` scheduling: Single-process pytest runs are deterministic: `test_qudit` is the first test in the file, calls
   `mz()`, populates `sampleQudits`, and the subsequent `clear()` leaves the vector's backing buffer mapped. From then on
   `begin()->levels` reads stale-but-mapped bytes and `qpp::sample` either returns empty or throws (caught by `invoke_no_throw`). CI runs `pytest -n auto` with the default per-test load distribution, so any worker that picks up `test_qudit_level`,   `test_qudit_invalid`, `test_unsupported_gates`, or `test_target_handler` before a measuring test gets a nullptr deref. Which CI target reproduces is a function of xdist scheduling and heap layout, not the target itself.


khalatepradnya:
Logs [link ](https://github.com/NVIDIA/cuda-quantum/actions/runs/25820996903/job/75862653315?pr=4439)for failing CI.

atgeller (approved):
LGTM, thanks @khalatepradnya!

PR #4516, opened 2026-05-13
Updated sample semantics (simple fix)

### LINKED ISSUE #4153: Explicit Measurements QPU Compatability

anpaz:
## Summary

Fixes #4153 by making default `sample()` global-register bitstrings follow the order of explicit terminal `mz` measurements, without enabling the broader `explicit_measurements=True` behavior.

For example, a kernel that ends with:

```cpp
mz(q[2]);
mz(q[0]);
mz(q[1]);
mz(q[3]);
```

now reports the `__global__` bitstring in that same order.

## Details

`sample()` already tracks which qubits are measured through `sampleQubits`, but the simulator path sorted those qubits before building the `__global__` result. That made the bitstring follow qubit allocation order instead of the order in which the user wrote the measurements.

This PR preserves the original measurement encounter order before sorting for backend execution. For local simulator execution, `__global__` is rebuilt from the backend result using that preserved order.

Some execution paths run the qubit-mapping compiler pass before sampling. That pass can change which physical backend qubit represents each source-level qubit. The compiler already uses `mapping_reorder_idx` to translate result bit positions after that transformation. This PR also updates `mapping_reorder_idx` so that translation reconstructs the user-facing `__global__` register in measurement encounter order instead of source qubit allocation order.

Implicit measurements inserted by the mapping pass still use allocation order, because there is no explicit user-written measurement order to preserve in that case.

### DIFF
--- a/lib/Optimizer/Transforms/Mapping.cpp
+++ b/lib/Optimizer/Transforms/Mapping.cpp
@@ -768,10 +768,14 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
           return;
         }
 
-        // Save which qubits are measured
+        // Save which qubits are measured, preserving measurement encounter
+        // order and ignoring duplicate measurements for the global register.
         if (isa<quake::MeasurementInterface>(op))
-          for (const auto &wire : wireOperands)
-            userQubitsMeasured.push_back(wireToVirtualQ[wire].index);
+          for (const auto &wire : wireOperands) {
+            auto virtualQubit = wireToVirtualQ[wire].index;
+            if (!llvm::is_contained(userQubitsMeasured, virtualQubit))
+              userQubitsMeasured.push_back(virtualQubit);
+          }
 
         // Map the result wires to the appropriate virtual qubits.
         for (auto &&[wire, newWire] :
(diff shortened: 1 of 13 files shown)


anpaz:
We decided to not move forward with this breaking change for now. See comments in #4153 for more details

PR #4519, opened 2026-05-14
Get the noise model from the simulator instead of the executionContext.

Renaud-K:
In preparation for the policy work, I am removing occurrences where simulation gets the noise model from the a global execution context. Instead the noise is saved in the simulator itself and accessed by a new `getNoiseModel()` method on the simulators. 

The context is only used to pass the information. In the future the noise model will come from the policy which is not global and will require this local storage.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -138,10 +138,6 @@ public:
   virtual std::unique_ptr<cudaq::SimulationState>
   createStateFromData(const cudaq::state_data &) = 0;
 
-  /// @brief Set the current noise model to consider when
-  /// simulating the state.
-  void setNoiseModel(const cudaq::noise_model *noise) { noiseModel = noise; }
-
   /// @brief Get the current noise model.
   const cudaq::noise_model *getNoiseModel() const { return noiseModel; }
 
@@ -826,7 +822,6 @@ protected:
   /// @brief Flush the gate queue, run all queued gate
   /// application tasks.
   void flushGateQueueImpl() override {
-    auto executionContext = cudaq::getExecutionContext();
 
     while (!gateQueue.empty()) {
       auto &next = gateQueue.front();
@@ -1294,7 +1289,6 @@ public:
   /// context, just measure, collapse, and return the bit.
   bool mz(const std::size_t qubitIdx,
           const std::string &registerName) override {
-    auto executionContext = cudaq::getExecutionContext();
 
     // Flush the Gate Queue
     flushGateQueue();
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -109,6 +109,9 @@ protected:
   /// @brief Internal result
   cudaq::sample_result internalResult = {};
 
+  /// @brief Noise model to apply to the current execution.
+  const cudaq::noise_model *noiseModel = nullptr;
+
 public:
   /// @brief The constructor
   CircuitSimulator() = default;
@@ -136,9 +139,11 @@ public:
   createStateFromData(const cudaq::state_data &) = 0;
 
   /// @brief Set the current noise model to consider when
-  /// simulating the state. This should be overridden by
-  /// simulation strategies that support noise modeling.
-  virtual void setNoiseModel(cudaq::noise_model &noise) = 0;
+  /// simulating the state.
+  void setNoiseModel(const cudaq::noise_model *noise) { noiseModel = noise; }
+
+  /// @brief Get the current noise model.
+  const cudaq::noise_model *getNoiseModel() const { return noiseModel; }
 
   virtual void setRandomSeed(std::size_t seed) {
     // do nothing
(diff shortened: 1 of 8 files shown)


1tnguyen (approved):
LGTM 👍 

I think we need a matching change in the `mgpu` downstream simulator as well. We also query the noise model from the `cudaq::getExecutionContext()` there.

Renaud-K:
Can you show me where this is? I had temporarily renamed the `noiseModel` field of the execution context to catch all the cases where it is accessed.

runtime/nvqir/cutensornet/simulator_mps.h
```diff
@@ -317,7 +317,7 @@ class SimulatorMPS : public SimulatorTensorNetBase<ScalarType> {
     auto executionContext = cudaq::getExecutionContext();
 
     LOG_API_TIME();
-    const bool hasNoise = executionContext && executionContext->noiseModel;
+    const bool hasNoise = this->getNoiseModel() != nullptr;
```
sacpis, line 320: nit: I see an existence of `auto executionContext = cudaq::getExecutionContext();` in this file. Can we please get rid of it as we are not using `executionContext` anymore?

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1315,8 +1312,7 @@ class CircuitSimulatorBase : public CircuitSimulator {
 
     // Apply measurement noise (if any)
     // Note: gate noises are applied during flushGateQueue
-    if (executionContext && executionContext->noiseModel &&
-        !executionContext->noiseModel->empty())
+    if (getNoiseModel() && !getNoiseModel()->empty())
```
sacpis, line 1315: I think we can cache this `getNoiseModel() && !getNoiseModel()->empty()` inside `configureExecutionContext` in order to avoid making multiple calls, loads.

In `configureExecutionContext(...)`
```
hasNoise = getNoiseModel() && !getNoiseModel()->empty();
```

and then just use it here in `if(hasNoise)`.

runtime/nvqir/NVQIR.cpp
```diff
@@ -808,7 +808,7 @@ void __quantum__qis__apply_kraus_channel_double(std::int64_t krausChannelKey,
   if (!ctx)
     return;
 
-  auto *noise = ctx->noiseModel;
+  auto *noise = nvqir::getCircuitSimulatorInternal()->getNoiseModel();
```
sacpis, line 811: Can we cache the active noise pointer in an `nvqir::` during configureExecutionContext? In `CircuitSimulator.h`, inside `configureExecutionContext`, we can set something like

```
nvqir::activeNoiseModel = noiseModel;
```

and then here, just call it directly

```
auto *noise = nvqir::activeNoiseModel;
```

That way we can avoid calling `getCircuitSimulatorInternal` for every function call.

runtime/nvqir/CircuitSimulator.h
sacpis, line 848: Not related to this PR. Just saw this and few other places in the code base. 

Curious, do we know why we are copying `next.parameters` to `std::vector<double> params`? Isn't `next.parameters` already a `std::vector<double>`? In which case we can pass it directly.

sacpis: There may be a reason to copying it. Just asking as I am unaware of it.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1113,6 +1103,8 @@ class CircuitSimulatorBase : public CircuitSimulator {
   /// @brief Set the execution context
   void configureExecutionContext(cudaq::ExecutionContext &context) override {
     context.canHandleObserve = canHandleObserve();
+    noiseModel = context.noiseModel;
+    context.noiseModel = nullptr;
```
sacpis, line 1107: Sorry, I do not get this change.

sacpis (approved):

runtime/nvqir/cutensornet/simulator_mps.h
Renaud-K: This turns out the be a compilation error with the llvm toolchain so I took care of it.

runtime/nvqir/NVQIR.cpp
Renaud-K: We already call `nvqir::getCircuitSimulatorInternal()` in every function call all over this file. If caching this call is needed, it should be a separate body of work.

runtime/nvqir/CircuitSimulator.h
Renaud-K: The context is only used pass the information. Once it is passed, it is nullified to stress test that no one else use it.

runtime/nvqir/CircuitSimulator.h
Renaud-K: These are inlined calls.

runtime/nvqir/CircuitSimulator.h
Renaud-K: I am not familiar with this code.

runtime/nvqir/CircuitSimulator.h
sacpis: Thanks @Renaud-K for the clarification.

PR #4521, opened 2026-05-15
Fix custatevec GPU memory leak in `state::from_data` from host pointer

huaweil-nv:
## Summary
`CusvState::createFromSizeAndPtr` decided ownership of the device buffer from the `state_data` variant index **before** calling `maybeCopyToDevice`:

  ```cpp
  bool weOwnTheData = type < 2 ? true : false;   // pair variants -> false
  ptr = maybeCopyToDevice(size, ptr);             // may have just cudaMalloc'd
```
For the std::pair<std::complex<T>*, std::size_t> variants (indices 2 and 3), weOwnTheData is false. But maybeCopyToDevice invokes cudaMalloc + cudaMemcpy whenever the input pointer is host memory, and the freshly allocated device buffer is then handed to a state whose destroyState() early-returns on !ownsDevicePtr and never calls cudaFree.

This is the path taken by every cudaq.State.from_data(numpy_array) call on the nvidia target (py_state.cpp always wraps numpy data as a pair). Each call therefore permanently leaks size * sizeof(complex) bytes of device memory.

**Reproducer**
```
  import ctypes, gc, numpy as np, cudaq
  cudaq.set_target('nvidia')
  cudart = ctypes.CDLL('libcudart.so')

  def free_mb():
      f = ctypes.c_size_t(0); t = ctypes.c_size_t(0)
      cudart.cudaMemGetInfo(ctypes.byref(f), ctypes.byref(t))
      return f.value / (1024.0*1024.0)

  arr = np.zeros(1 << 22, dtype=np.complex64); arr[0] = 1.0
  for _ in range(20):  # warmup
      s = cudaq.State.from_data(arr); del s; gc.collect()
  base = free_mb()
  for _ in range(50):
      s = cudaq.State.from_data(arr); del s; gc.collect()
  print(f'leaked: {base - free_mb():.1f} MB (expected 0)')
```
50 calls (post-warmup), 22-qubit FP32
before: 1600 MB leaked
fixed: 0.0 MB


1tnguyen (approved):
LGTM 👍 Thanks @huaweil-nv

huaweil-nv:
> /ok to test 0d7c75004fce01f385b4a5a1531bfc45524bae37

sacpis (approved):
LGTM. Thanks @huaweil-nv. Wondering if we can add a regression test that asserts no device memory leak across N from_data calls.

huaweil-nv:
> LGTM. Thanks @huaweil-nv. Wondering if we can add a regression test that asserts no device memory leak across N from_data calls.

That makes sense. I checked the existing coverage, and GetStateTester.checkOverlapFromHostPointer already exercises the from_data(host pointer, size) path that leaked the device allocation.
For the no-leak assertion, I think compute-sanitizer is a better fit than adding a cudaMemGetInfo-based assertion inside the gtest. cudaMemGetInfo observes global GPU free memory, so it can be flaky on shared GPU systems or due to driver/runtime allocator bookkeeping.
I can enable a compute-sanitizer memcheck job in NVIDIA's internal DVS-SC pipeline to run this coverage periodically. That should catch this class of device memory leaks reliably without making the regular unit test flaky. how do you think?

huaweil-nv:
> /ok to test 7f150b54e02d905684b2b2ffbcae9900ea87b792

schweitzpgi (approved):
Thank you for the bug fix.

PR #4523, opened 2026-05-15
Moving code up and down class hierarchies in preparation for policies

Renaud-K:
This is another PR in preparation for using policies. 
1. `configureExecutionContext` is made non-virtual in `ExecutionManager` and `CircuitSimulator`. 
    Under the existing code, it does not really needed to be virtual. This will greatly simplify using policies, since each policy could introduce and overloaded version. Code is moved up the hierarchy so the base class can do all the work. When policies are intorduced, there will only be one place to check and specialize. 
1. The `BaseRemoteRESTQPU` kernel launch is moved into the concrete `RemoteRESTQPU` class because every other subclass of `BaseRemoteRESTQPU` overwrites the launch.
1.  Clean-ups in the `OrcaRemoteRESTQPU.h`

### DIFF
--- a/runtime/cudaq/qis/execution_manager.h
+++ b/runtime/cudaq/qis/execution_manager.h
@@ -114,7 +114,9 @@ public:
   bool memoryLeaked() { return !tracker.allDeallocated(); }
 
   /// Configure the execution context before an execution.
-  virtual void configureExecutionContext(ExecutionContext &ctx) {}
+  void configureExecutionContext(const sample_policy &policy,
+                                 ExecutionContext &ctx);
+  void configureExecutionContext(ExecutionContext &ctx);
 
   /// Finalize the execution context after an execution.
   void finalizeExecutionContext(ExecutionContext &ctx);
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -109,6 +109,12 @@ protected:
   /// @brief Internal result
   cudaq::sample_result internalResult = {};
 
+  /// @brief Reference to the current circuit name.
+  std::string currentCircuitName = "";
+
+  /// @brief Get the name of the current circuit being executed.
+  std::string getCircuitName() const { return currentCircuitName; }
+
   /// @brief Noise model to apply to the current execution.
   const cudaq::noise_model *noiseModel = nullptr;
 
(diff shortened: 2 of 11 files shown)


runtime/cudaq/qis/execution_manager.h
```diff
@@ -114,7 +114,9 @@ class ExecutionManager {
   bool memoryLeaked() { return !tracker.allDeallocated(); }
 
   /// Configure the execution context before an execution.
-  virtual void configureExecutionContext(ExecutionContext &ctx) {}
+  void configureExecutionContext(const sample_policy &policy,
+                                 ExecutionContext &ctx);
```
sacpis, line 118: Did we miss to define it in `.cpp` file?

runtime/nvqir/CircuitSimulator.h
```diff
@@ -273,8 +279,21 @@ class CircuitSimulator {
...
+  /// handle <psi | H | psi> instead of NVQIR applying measure
+  /// basis quantum gates to change to the Z basis and sample.
+  virtual bool canHandleObserve() { return false; }
+
+  /// @brief Set the execution context
+  void configureExecutionContext(cudaq::ExecutionContext &context) {
+    context.canHandleObserve = canHandleObserve();
+    noiseModel = context.noiseModel;
+    // Stress testing the fact that the context is just used topass the noise
+    // and no one is supposed to use it after that.
```
sacpis, line 292: nit: we can remove this comment?

runtime/nvqir/resourcecounter/ResourceCounter.h
```diff
@@ -87,14 +87,6 @@ class ResourceCounter : public nvqir::CircuitSimulatorBase<double> {
     prepopulated = false;
   }
 
-  void configureExecutionContext(cudaq::ExecutionContext &context) override {
-    if (context.name != "resource-count")
```
sacpis, line 91: Just checking, if we don't need this check?

sacpis (commented):
LGTM. Thanks @Renaud-K. Left a few comments.

sacpis (approved):

runtime/cudaq/qis/execution_manager.h
Renaud-K: I submit the smaller PRs from code from the bigger change. So sometimes, the future code is not cleaned up. It will be implemented in a future PR. It is harmless for now, so if you don't mind, I would like to leave it.

runtime/nvqir/CircuitSimulator.h
Renaud-K: You asked in the previous PR. So, I thought a comment was warranted. I do think that nullify the pointer deserves an explanation.

runtime/cudaq/qis/execution_manager.h
sacpis: Oh, got it. Thanks @Renaud-K. Please go ahead.

runtime/nvqir/resourcecounter/ResourceCounter.h
Renaud-K: I don't think so. Since #4493, the use of the ResourceCounter is very well controlled in a scoped object and we have [other checks](https://github.com/Renaud-K/cuda-quantum/blob/main/runtime/common/BaseRemoteRESTQPU.h#L129).
We would need a bug to trigger this, not a user error. 
Instead of a context, we will have an estimate policy and in this overload we can check for the proper simulator. But the end goal would be to have a compile time error for simulators not supporting algorithm.

Renaud-K:
> /ok to test a42246783b1c42492a3bc149ca5e4c8e38791e16

PR #4524, opened 2026-05-15
[core] Reorganizing the directory structure.

schweitzpgi:
Continuation of the effort to make the code base more modular.

CI is stepping on itself with its checks. The code is being moved, not changed. Errors on moved code will not be fixed as that just clutters up the diffs for issues that are irrelevant to the work being done here.


mitchdz (approved):
Thanks for this, this looks fine to me. It is failing CI checks due to moved files now not being happy with formatting. Since there is no way to get into the merge queue, I am fine pushing directly to main and then if any fallout we can handle it there.

PR #4525, opened 2026-05-15
[CI] macOS: exclude Xcode release candidates from latest-stable selector

khalatepradnya:
### Summary

Tighten the macOS Xcode selector to exclude release candidates (in addition to betas), and invalidate the macOS LLVM devdeps cache so the existing tarball — built against a now-deleted RC SDK — is replaced on the next CI run.

### Motivation

macOS CI has been failing in Build & Test and the py3.11 / py3.13 wheel jobs with:
```
CMake Error in lib/LineEditor/CMakeLists.txt:
  Imported target "LibEdit::LibEdit" includes non-existent path
    "/Applications/Xcode_26.5_Release_Candidate.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.5.sdk/usr/include"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does
    not provide.
CMake Generate step failed.  Build files cannot be regenerated correctly.
ninja: error: rebuilding 'build.ninja': subcommand failed
Failed to build compiler components.
```
py3.12 wheel passed on the same run — the classic runner-roulette signature when a cached absolute path resolves on some `macos-26` VMs but not others.

Root cause in two parts, both in our CI plumbing:

1. `.github/actions/setup-xcode-stable/action.yml` only filtered by `LicenseInfo.plist`'s `licenseType == "beta"`. Apple ships release candidates with `licenseType` set to `"GM"` or `"Release"`, so `Xcode_26.5_Release_Candidate.app` was selected as "newest stable" when the LLVM devdeps cache was built. Its SDK's absolute path got baked into LLVM's `LibEdit::LibEdit` imported target inside the cached `~/.llvm-project/build` tree. Once GA Xcode 26.5 rolled out, the RC was removed from some `macos-26` images, and the baked-in path stopped resolving.

2. The Xcode selector file was not part of `dev_environment_macos.yml`'s `scripts_hash`. Fixing the selector alone would have no effect — consumer jobs would keep pulling the same RC-poisoned ORAS tarball.

### DIFF
--- a/.github/actions/setup-xcode-stable/action.yml
+++ b/.github/actions/setup-xcode-stable/action.yml
@@ -1,5 +1,8 @@
 name: Set up latest stable Xcode
-description: Select the newest non-beta Xcode installed on the macOS runner.
+description: >-
+  Select the newest stable Xcode installed on the macOS runner. Excludes
+  betas and release candidates (matches the "latest-stable" semantics of
+  maxim-lobanov/setup-xcode).
 
 outputs:
   version:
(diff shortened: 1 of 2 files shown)


mitchdz (approved):

taalexander (approved):
LGTM thank you. I wonder if we should test against the next release candidate in some way to identify future breaks @mitchdz

mitchdz:
I say we make our pipelines much more stable and not hitting egress/402 issues before we start testing devel frameworks, but it is a good idea.

PR #4532, opened 2026-05-18
Add `unitary` as an op trait and add it trait to all unitary ops in the Quake Dialect

cabreraam:
This PR aims to add a  `unitary` trait to all unitary ops in the Quake dialect. This will allow, in part, for easy filtering on unitary ops in mixed quantum/classical MLIR circuit representations.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Common/Traits.h
+++ b/include/cudaq/Optimizer/Dialect/Common/Traits.h
@@ -71,6 +71,15 @@ public:
 };
 
 template <typename ConcreteType>
+class Unitary : public mlir::OpTrait::TraitBase<ConcreteType, Unitary> {
+public:
+  static mlir::LogicalResult verifyTrait(mlir::Operation *op) {
+    // todo: we should verify that the operation is unitary, potentially
+    return mlir::success();
+  }
+};
+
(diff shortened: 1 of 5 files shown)


cabreraam:
A good point brought up by @schweitzpgi: is a  `Unitary` trait even necessary? Could we just filter on [`QuantumGate`](https://github.com/NVIDIA/cuda-quantum/blob/main/include/cudaq/Optimizer/Dialect/Common/Traits.td#L18-L20)?

cabreraam:
Further inspection shows that [`Reset` op](https://github.com/NVIDIA/cuda-quantum/blob/main/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td) is marked with the `QuantumGate` trait, but if this is the only op that we need to worry about when determining whether a given op is unitary, something like

```
isa<cudaq::quake::QuantumGate>(op) && !isa<cudaq::quake::ResetOp>(op);
```
could suffice.

schweitzpgi:
After conversion with @cabreraam , I think this one can be closed.

schweitzpgi (changes requested):
Not needed.

cabreraam:
Agree with @schweitzpgi. Closing this!

PR #4533, opened 2026-05-18
[Python] Clear `_resolving_arguments` on exception in `ensure_not_recursive`

khalatepradnya:
### Summary

`PyKernelDecorator.ensure_not_recursive` left `_resolving_arguments=True` when the wrapped `resolve_captured_arguments` method raised. Every subsequent invocation on the same instance then misreported the real error as `"recursive kernel call detected in <name>"`. Wrap the body in `try/finally` so the flag is cleared on every exit path.

### DIFF
--- a/python/cudaq/kernel/kernel_decorator.py
+++ b/python/cudaq/kernel/kernel_decorator.py
@@ -67,9 +67,14 @@ def ensure_not_recursive(method):
             )
 
         self._resolving_arguments = True
-        ret = method(self, *args, **kwargs)
-        self._resolving_arguments = False
-        return ret
+        try:
+            return method(self, *args, **kwargs)
+        finally:
+            # Clear the flag on every exit path. Without `finally`, an
+            # exception raised by `method` leaks `_resolving_arguments=True`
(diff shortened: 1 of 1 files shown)


sacpis (approved):
LGTM. Thanks @khalatepradnya. Can we please add a regression test?

khalatepradnya:
> LGTM. Thanks @khalatepradnya. Can we please add a regression test?

The test set up seemed tedious to me, but let me add that in a follow-up...

khalatepradnya:
> > LGTM. Thanks @khalatepradnya. Can we please add a regression test?
> 
> The test set up seemed tedious to me, but let me add that in a follow-up...

Addressed in https://github.com/NVIDIA/cuda-quantum/pull/4683

sacpis:
> > > LGTM. Thanks @khalatepradnya. Can we please add a regression test?
> > 
> > 
> > The test set up seemed tedious to me, but let me add that in a follow-up...
> 
> Addressed in #4683

Thank you @khalatepradnya. Approved the PR #4683.

PR #4537, opened 2026-05-18
[Python] measure_handle frontend cleanups

khalatepradnya:
- Thread `raiseError` and `cudaqAliases` through `mlirTypeFromAnnotation`'s recursive calls (`Callable` / `list` / `tuple`) so type aliases like `import cudaq as cq` resolve `list[cq.measure_handle]` to the boundary diagnostic instead of falling through to the generic "not yet a supported type" error.

- Run the host-device `measure_handle` boundary check before creating the `FuncOp` so a rejection does not leave an orphaned op in the module.

- Discriminate per-element when returning a tuple expression whose type is a struct, so an intermediate struct with `!cc.measure_handle` in its element types is never constructed. Without this, the intermediate type survives to QIR lowering and fails to legalize on `cudaq.run` even though the outer struct is later coerced to the expected type.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -1781,17 +1781,11 @@ class PyASTBridge(ast.NodeVisitor):
             # function name
             fullName = nvqppPrefix + self.name
 
-            # Create the FuncOp
-            f = func.FuncOp(fullName,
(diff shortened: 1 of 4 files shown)


sacpis (approved):
LGTM. Thanks @khalatepradnya.

PR #4538, opened 2026-05-18
Group classical, unitary, measurement/reset, and remaining Quake ops

cabreraam:
This PR aims to create a compiler pass that groups operations while respecting any boundaries that are created due to ops depending on other ops and potential side effects.

The groupings, which can be most plainly seen in `pickNextReady`, group by the following categories:

- Classical (non-Quake ops)
- Unitary ops
- Measurements or resets
- Remaining Quake Ops
- Everything else

If it turns out that #4532 isn't necessary, I can refactor this to not use the  `Unitary` trait.

### DIFF
--- a/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
+++ b/include/cudaq/Optimizer/Dialect/Quake/QuakeOps.td
@@ -1773,7 +1773,7 @@ def CustomUnitarySymbolOp :
       return build($_builder, $_state, mlir::TypeRange{}, generator, is_adj,
                    parameters, controls, targets, negates);
     }]>,
-    OpBuilder<(ins "mlir::SymbolRefAttr":$generator,
+    OpBuilder<(ins "mlir::SymbolRAttr":$generator,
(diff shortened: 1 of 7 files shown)


schweitzpgi:
What happened to writing an analysis?

schweitzpgi (changes requested):
I'm expecting an analysis, not a mechanical shuffling of operations.

- if we were going to move the operations, we'd do so by cloning them, not by backdooring the updates to the links
- how does this work with control-flow? were those cases tested?
- the CHECK lines ought to be generated using tpls/llvm/mlir/utils/generate-test-checks.py

cabreraam:
> I'm expecting an analysis, not a mechanical shuffling of operations.
> 
> * if we were going to move the operations, we'd do so by cloning them, not by backdooring the updates to the links
> * how does this work with control-flow? were those cases tested?
> * the CHECK lines ought to be generated using tpls/llvm/mlir/utils/generate-test-checks.py

All good points. I'm going to give this another shot!

PR #4539, opened 2026-05-18
[opt][qec] Add `EraseQEC` pass for hardware-prep pipeline

khalatepradnya:
### Summary

Adds the `--erase-qec` MLIR pass that strips `qec.detector`, `qec.observable`, and `qec.detectors` ops from the IR,
and slots it into the hardware-prep pipeline. The emulation pipeline is deliberately untouched so the analysis-engine path can observe those ops as needed

Made-with: Cursor

### DIFF
--- a/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
@@ -449,6 +449,17 @@ def EraseNopCalls : Pass<"erase-nop-calls"> {
   }];
 }
 
+def EraseQEC : Pass<"erase-qec"> {
+  let summary = "Erase QEC detector and observable operations.";
+  let description = [{
+    Removes `qec.detector`, `qec.observable`, and `qec.pair_detectors` operations
+    from the IR. These are side-effecting declaration ops: they record a
+    parity-check or logical-observable relationship over prior measurement
+    results so a detector error model can be built from the kernel. This pass
+    strips them.
+  }];
+}
+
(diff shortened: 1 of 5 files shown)


cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -449,6 +449,17 @@ def EraseNopCalls : Pass<"erase-nop-calls"> {
   }];
 }
 
+def EraseQEC : Pass<"erase-qec"> {
+  let summary = "Erase QEC detector and observable operations.";
+  let description = [{
+    Removes `qec.detector`, `qec.observable`, and `qec.detectors` operations
+    from the IR. These ops are inert annotations: they declare relationships
```
schweitzpgi: This sounds like it shouldn't have gone in this way by calling them "inert annotations". Why aren't they simply generated as calls to designated function names?

schweitzpgi (dismissed):

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
khalatepradnya: I see this has incorrect use of the term "annotations", stale from my early prototype and borrowing "stim" terminology. Let me fix that...

sacpis (approved):
LGTM. Thanks @khalatepradnya.

anpaz (approved):
lgtm

atgeller (approved):

PR #4542, opened 2026-05-19
[core] Make the QEC dialect look more like the other dialects.

schweitzpgi:
We want our code and IR to look consistent as much as possible.

Update tests.


Renaud-K (approved):
Looks reasonable to me.

PR #4555, opened 2026-05-20
Fix qpp density matrix overlap calculation

huaweil-nv:
## Summary

This removes the determinant-based overlap formula from the qpp density-matrix backend and aligns it with the existing density-matrix behavior used by the dynamics backend.

The previous implementation used:

```cpp
Tr(rho * sigma) + 2 * sqrt(det(rho) * det(sigma))
```
That expression is only valid as a 2x2 density-matrix closed form, but the backend applies it to arbitrary 2^n x 2^n density matrices. For full-rank multi-qubit mixed states this returns a value that is neither the Hilbert-Schmidt inner product nor the density-matrix fidelity.

This patch does not attempt to redefine the public overlap contract for density matrices. It only removes the invalid 2x2-only formula and makes density-matrix-cpu consistent with the existing dynamics density-matrix implementation, which computes Tr(rho† sigma).

**Reproducer**
For a normalized 2-qubit density matrix:
```
import cudaq, numpy as np

rho = np.diag([0.4, 0.3, 0.2, 0.1]).astype(np.complex128)

for target in ["density-matrix-cpu", "dynamics"]:
    cudaq.reset_target()
    cudaq.set_target(target)
    state = cudaq.State.from_data(rho)
    print(target, state.overlap(state))

print("expected HS:", np.trace(rho.conj().T @ rho))
```
Before:
```
density-matrix-cpu: 0.3048
dynamics:           0.3
expected HS:        0.3
```
After:
```
density-matrix-cpu: 0.3
dynamics:           0.3
expected HS:        0.3
```

**Note**: If maintainers feel density-matrix `overlap` should remain unspecified, or that the intended fix should instead define a fidelity-based API, feel free to modify/close this PR.

### DIFF
--- a/python/tests/builder/test_state.py
+++ b/python/tests/builder/test_state.py
@@ -160,6 +160,18 @@ def test_state_density_matrix_simple():
     cudaq.reset_target()
 
 
+def test_state_density_matrix_mixed_overlap():
+    cudaq.set_target('density-matrix-cpu')
+
+    rho = np.diag([0.4, 0.3, 0.2, 0.1]).astype(np.complex128)
+    state = cudaq.State.from_data(rho)
+
+    # Density matrix overlap is the Hilbert-Schmidt inner product.
+    assert np.isclose(state.overlap(state), np.trace(rho.conj().T @ rho))
+
+    cudaq.reset_target()
+
+
 def test_state_density_matrix_integration():
     """
     An integration test on the state density matrix class. Uses a CUDA-Q
--- a/python/tests/builder/test_state.py
+++ b/python/tests/builder/test_state.py
@@ -160,7 +160,7 @@ def test_state_density_matrix_simple():
     cudaq.reset_target()
 
 
-def test_state_density_matrix_mixed_overlap():
+def test_state_density_matrix_self_overlap():
     cudaq.set_target('density-matrix-cpu')
 
     rho = np.diag([0.4, 0.3, 0.2, 0.1]).astype(np.complex128)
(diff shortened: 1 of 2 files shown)


python/tests/builder/test_state.py
```diff
@@ -160,6 +160,18 @@ def test_state_density_matrix_simple():
     cudaq.reset_target()
 
 
+def test_state_density_matrix_mixed_overlap():
```
sacpis: nit: can be named as `test_state_density_matrix_self_overlap`.

python/tests/builder/test_state.py
```diff
@@ -160,6 +160,18 @@ def test_state_density_matrix_simple():
...
+
+    rho = np.diag([0.4, 0.3, 0.2, 0.1]).astype(np.complex128)
+    state = cudaq.State.from_data(rho)
+
+    # Density matrix overlap is the Hilbert-Schmidt inner product.
+    assert np.isclose(state.overlap(state), np.trace(rho.conj().T @ rho))
+
+    cudaq.reset_target()
+
+
```
sacpis, line 174: We can also add a test for cross-state assertion, something like `a.overlap(b) where a != b`.

sacpis (approved):
Overall LGTM. Thanks @huaweil-nv. Left few comments.

python/tests/builder/test_state.py
huaweil-nv: Thanks for the review. I renamed the test to `test_state_density_matrix_self_overlap` in new commit.

python/tests/builder/test_state.py
huaweil-nv: Thanks, I added a cross-state overlap test for `a.overlap(b)` where `a !=b` in new commit.

sacpis:
> /ok to test b06a9b4

1tnguyen (approved):
Thanks @huaweil-nv!

PR #4556, opened 2026-05-20
Add FTQC compiler benchmark target

taalexander:
Add a ftqc fault-tolerant target compiling to the ` 'h', 's', 'sdg', 't', 'tdg', 'rx', 'ry', 'rz', 'x', 'y', 'z', 'cx', 'mz'` basis which is intended as a representation of Clifford+Pauli-product-measurement+T/Tdg+Rotation synthesis as is possible today in Cuda-Q.

This also renames the existing target and fixes some bugs/translations we need to make this happen.


lmondada (changes requested):
I like the gist, thanks @taalexander !

Regarding the pattern selection logic: introducing multiple versions of the same decomposition pattern is a bit of a hack. I've spent some time thinking about it. A better way to solve this problem might be:

1. Introduce the notion of "decomposition variant" in the `DecompositionPatternType` metadata, which acts as an enumeration of the concrete decompositions that the pattern implements. For instance, the pattern `SToR1` might have two variants: the "s -> r1" variant and the "s(n) -> r1(n)" variant. In the future, we could further register "s(c) -> n(c)" for a constant `c` if that is useful -- but that is always optional.
2. The variants are defined when registering the decomposition pattern, i.e. the updated REGISTER_DECOMPOSITION_PATTERN macro would be used as follows
    ```c++
    REGISTER_DECOMPOSITION_PATTERN(SToR1,
                                   {"s", "r1"},
                                   {"s(n)", "r1(n)"})
    ```
    The source gates of the variants should all be distinct.
3. When traversing the decomposition pattern graph backwards, we propagate the target gate to source gates depending on the first variant that matches. This way, our pattern selection algorithm can reason about the fact that `s` produces a `r1` and not `r1(n)`. We never check for "cover" nor do we implicitly handle "unbounded" ops. Instead, whenever we add gate `g(c)` for some constant `c` to the set of gates to visit, we also add `g(n)`. In other words, we add an edge `g(c) -> g(n)` of cost 0 in the graph that Dijkstra explores. Thus in the above example, no variants would be found for "r1(1)", but the algorithm would still select the second variant once "r1(n)" gets visited.
4. We add the option to blacklist variants of patterns at pattern construction time. In DecompositionGraph::selectPatterns, we can then disable all variants whose source ops are in the target basis already. The pattern matchAndRewrite logic must be updated to respect this blacklist.

Finally, regarding the benefit attached to a rewrite rule: breaking ties according to whether the source is bounded or unbounded will be brittle. Ideally we'd be able to replace this tie breaking mechanism by using the ordering of variants within a pattern.

This is a bit of work, but might be worth trying to solve this the right way once and for all. AI should make this rather easy.

Happy to discuss/defer to another PR if this ends up being bigger/harder than anticipated. However, I'd rather not put this on the pile of "let's do this maybe one day", because as it stands the Dijkstra logic is getting quite convoluted here, to the point that I'm no longer convinced it behaves as we'd expect it to. I'd feel more comfortable with the refactor I propose because the graph that Dijkstra is run over can then be cleanly understood as:
- edges are variants of patterns + implicit edges that promote `g(c)` to `g(n)`
- vertices are `g(c)` for any gate `g` and constant `c`, plus `g(n)`.

Hope this makes sense!

taalexander:
Thanks Luca, I agree it is somewhat of a hack. I also agree that with AI it's the opportunity to do it right. I'll give it a go.

taalexander:
Thanks @lmondada, I've implemented your suggestions.

lmondada:
@taalexander this is roughly what I had in mind. Does that work for you?

lmondada (approved):

taalexander:
Thank you for the changes Luca. These are helpful and make more sense long term.

PR #4557, opened 2026-05-20
End-to-end sample_policy

Renaud-K:
## Summary

This change continues the runtime refactor from a single global `ExecutionContext` (control + inputs + results via `ctx.name` and `ctx.result`) toward **typed policies** that carry launch inputs and return results by value.

For **`sample`**, the policy path is wired end-to-end in C++ and Python:

## The key changes in this PR are
1. We are no longer calling `with_execution_context`. 
    1. We want to get away from the virtual family of calls: configure, begin, end, finalize
    2. `with_execution_context` is calling `ExecutionManager::finalizeContext` too late. We need to result sooner.
    3. The RAII pattern of `with_execution_context` is still helpful so it has been broken up into 
        - `with_policy_and_ctx`: only installs the ExecutionContext. Does not do any of the configure, begin, end, finalize calls
        - `with_default_em`: takes care of returning the result based on the policy from the ExecutionManager
    1. `with_default_em` is helpful for emulation and library mode where no QPU is called but a result is still needed.

2. The ExecutionContext is no longer passed through the policy overloads. For instance, it has been removed from the hidden friends of the policy:
```
  friend sample_result
  finalize_execution_manager_impl(ExecutionManager &mgr,
                                  const sample_policy &policy);
  friend sample_result
  finalize_simulation_circuit_impl(nvqir::CircuitSimulator &sim,
                                   const sample_policy &policy);
```

3. We are a lot more explicit about the use of the async versus the sync policy is used. We only use the async policy for remote backends and we use the synchronous policy for all other backends (running in a separate thread to emulate asynchronous behavior). This is not a change of behavior since this is exactly what was done before but it is now a lot more apparent. Example:
```
async_sample_result
  completeLaunchKernel(async_sample_policy &policy,
                       const std::string &kernelName,
                       std::vector<cudaq::KernelExecution> &&codes) {
...
    // Execute the codes produced in quake lowering
    // Allow developer to disable remote sending (useful for debugging IR)
    assert(!emulate);
    if (getEnvBool("DISABLE_REMOTE_SEND", false))
      return {};
    // Cannot be observe and run at the same time
    const cudaq::details::ExecutionContextType execType =
        cudaq::details::ExecutionContextType::sample;

    auto future = executor->execute(codes, execType,
                                    &executionContext->invocationResultBuffer);
    return async_sample_result(std::move(future));
  }
```

4.  Because the same code might be called from the different policy overloads, the biggest part of these diffs is about moving code into common function. They are called:
    - samplePreamble
    - compileModulePreamble
    - compileKernelExecutions
    - compileModuleImpl
Some of which are templatized to work with any policy type.

5. Of course, a new API is added to the QPU:
```
  virtual sample_result launchKernel(sample_policy &policy,
                                     const AnyModule &module, KernelArgs args);

  virtual async_sample_result launchKernel(async_sample_policy &policy,
                                           const AnyModule &module,
                                           KernelArgs args);
```

6. Which are called by installing a dispatcher (`executeKernelApi) on the ExcutionContext (the technique used by the architecture team)
```
template <typename Policy, typename Callable, typename... Args>
auto launch(Policy &policy, std::size_t qpu_id, ExecutionContext &ctx,
            quantum_platform &platform, Callable &&f, Args &&...args)
    -> Policy::result_type {
...
  typename Policy::result_type result;
  auto &qpu = platform.getQPU(qpu_id);
  ctx.executeKernelApi = [&qpu, &result, &policy, &ctx](
                             const AnyModule &module, const KernelArgs &args) {
    result = qpu.launchKernel(policy, module, args);
  };

  detail::try_finally(
      [&] {
        detail::with_policy_and_ctx(policy, ctx, std::forward<Callable>(f),
                                    std::forward<Args>(args)...);
      },
      [&] { ctx.executeKernelApi = nullptr; });
  return result;
}
```

7. In the callbacks, the `executeKernelApi` is called if it is installed, which is going to call the policy overload instead of the `unifiedLaunchModule` API.
```
cudaq::KernelThunkResultType
cudaq::altLaunchKernel(const char *kernelName,
                       cudaq::KernelThunkType kernelFunc, void *kernelArgs,
                       std::uint64_t argsSize, std::uint64_t resultOffset) {
  ScopedTraceWithContext("altLaunchKernel", kernelName, argsSize);
  auto &platform = *getQuantumPlatformInternal();
  std::string kernName = kernelName;
  KernelArgs args{KernelArgs::PackedArgs{kernelArgs, argsSize, resultOffset}};
  SourceModule src{kernName, kernelFunc};
  auto ctx = cudaq::getExecutionContext();
  if (ctx && ctx->executeKernelApi) {
    ctx->executeKernelApi(src, args);
    return {};
  }
  std::size_t qpu_id = cudaq::getCurrentQpuId();
  return platform.unifiedLaunchModule(src, args, qpu_id);
}
```

8. In the python case, we are able to call the C++ `detail::launch` API, by simply passing a sample_policy object. `py_sample.h` and `py_sample.cpp` were introduced for this mechanism. The asynchronous flow was already calling the C++ runAsyncSampling API, so nothing had to be done. 

9. The compiler has overloads as well. Example:
```
  cudaq::CompiledModule
  runPassPipeline(cudaq::sample_policy &policy, const std::string &kernelName,
                  const void *modulePtr, cudaq::KernelArgs args,
                  std::shared_ptr<mlir::MLIRContext> context = nullptr);
```
When an overload has been added, the code relying on the execution context has been removed. Forcing all paths to use the overloads. In particular the `compileModule` member of the QPUs now calls a common template based implementation that passes a policy:
```
  CompiledModule compileModule(sample_policy &policy, const SourceModule &src,
                               KernelArgs args, bool isEntryPoint) override {
    return compileModuleImpl(policy, src, args, isEntryPoint);
  }
```

Since this is only called from python, the policy dispatch happens based on the policy name.
```
CompiledModule quantum_platform::compileModule(const SourceModule &src,
                                               const KernelArgs &args,
                                               std::size_t qpu_id,
                                               bool isEntryPoint) {
  validateQpuId(qpu_id);
  auto &qpu = platformQPUs[qpu_id];
  auto ctx = cudaq::getExecutionContext();
  if (ctx && ctx->name == "sample") {
    sample_policy policy;
    return qpu->compileModule(policy, src, args, isEntryPoint);
  }
  return qpu->compileModule(src, args, isEntryPoint);
}
```
This is then only part of the flow that is not end-to-end yet and for which more refactoring might be needed.

### DIFF
--- a/python/cudaq/runtime/sample.py
+++ b/python/cudaq/runtime/sample.py
@@ -74,8 +74,8 @@ def __broadcastSample(kernel,
         ctx.totalIterations = N
         ctx.batchIteration = i
         ctx.explicitMeasurements = explicit_measurements
-        policy = cudaq_runtime.sample_policy(ctx, kernel_name,
-                                             explicit_measurements)
+        policy = cudaq_runtime.SamplePolicy(ctx, kernel_name,
+                                            explicit_measurements)
         res = cudaq_runtime.launch_sample(policy, ctx, lambda: kernel(*a))
         results.append(res)
 
@@ -178,8 +178,7 @@ def sample(kernel,
     ctx.kernelName = kernel_name
     ctx.explicitMeasurements = explicit_measurements
     ctx.allowJitEngineCaching = True
-    policy = cudaq_runtime.sample_policy(ctx, kernel_name,
-                                         explicit_measurements)
+    policy = cudaq_runtime.SamplePolicy(ctx, kernel_name, explicit_measurements)
 
     counts = cudaq_runtime.SampleResult()
     while counts.get_total_shots() < shots_count:
--- a/python/cudaq/runtime/sample.py
+++ b/python/cudaq/runtime/sample.py
@@ -68,13 +68,15 @@ def __broadcastSample(kernel,
     N = len(argSet)
     results = []
     for i, a in enumerate(argSet):
+        kernel_name = kernel.name if hasattr(kernel, 'name') else ''
         ctx = cudaq_runtime.ExecutionContext('sample', shots_count)
+        ctx.kernelName = kernel_name
         ctx.totalIterations = N
         ctx.batchIteration = i
         ctx.explicitMeasurements = explicit_measurements
-        with ctx:
-            kernel(*a)
-        res = ctx.result
+        policy = cudaq_runtime.sample_policy(ctx, kernel_name,
+                                             explicit_measurements)
+        res = cudaq_runtime.launch_sample(policy, ctx, lambda: kernel(*a))
         results.append(res)
 
     return results
@@ -171,25 +173,27 @@ def sample(kernel,
         cudaq_runtime.unset_noise()
         return res
 
+    kernel_name = kernel.name if hasattr(kernel, 'name') else ''
     ctx = cudaq_runtime.ExecutionContext("sample", shots_count)
-    ctx.kernelName = kernel.name if hasattr(kernel, 'name') else ''
+    ctx.kernelName = kernel_name
     ctx.explicitMeasurements = explicit_measurements
     ctx.allowJitEngineCaching = True
+    policy = cudaq_runtime.sample_policy(ctx, kernel_name,
+                                         explicit_measurements)
 
     counts = cudaq_runtime.SampleResult()
     while counts.get_total_shots() < shots_count:
-        with ctx:
-            kernel(*args)
+        result = cudaq_runtime.launch_sample(policy, ctx, lambda: kernel(*args))
         # If the platform is a hardware QPU, launch only once
         countsTotalIsZero = counts.get_total_shots() == 0
-        resultTotalWasReached = ctx.result.get_total_shots() == shots_count
+        resultTotalWasReached = result.get_total_shots() == shots_count
         if (countsTotalIsZero and
                 resultTotalWasReached) or cudaq_runtime.isQuantumDevice():
             # Early return for case where all shots were gathered the first time
             # through this loop.This avoids an additional copy.
             cudaq_runtime.unset_noise()
-            return ctx.result
-        counts += ctx.result
+            return result
+        counts += result
         if counts.get_total_shots() == 0:
             if explicit_measurements:
                 raise RuntimeError(
@@ -199,7 +203,6 @@ def sample(kernel,
                   "results when executed. Exiting shot loop to avoid infinite "
                   "loop.")
             break
-        ctx.result.clear()
     cudaq_runtime.unset_noise()
     ctx.unset_jit_engine()
     return counts
--- a/python/runtime/cudaq/algorithms/py_sample.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample.cpp
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "py_sample.h"
+#include "cudaq/algorithms/launch.h"
+#include "cudaq/algorithms/sample.h"
+#include <nanobind/stl/string.h>
+
+using namespace cudaq;
+
+static void construct_sample_policy(sample_policy *self, ExecutionContext &ctx,
+                                    const std::string &kernelName,
+                                    bool explicitMeasurements) {
+  new (self) sample_policy();
+  self->kernelName = kernelName;
+  self->options.shots = ctx.shots;
+  self->options.explicit_measurements = explicitMeasurements;
+  self->noiseModel = get_platform().get_noise(ctx.qpuId);
+}
+
+static sample_result launch_sample(sample_policy &policy, ExecutionContext &ctx,
+                                   nanobind::callable callable) {
+  return detail::launch(policy, ctx.qpuId, ctx, get_platform(),
+                        [&]() { callable(); });
+}
+
+void cudaq::bindPySample(nanobind::module_ &mod) {
+  nanobind::class_<sample_policy>(mod, "sample_policy")
+      .def("__init__", construct_sample_policy, nanobind::arg("ctx"),
+           nanobind::arg("kernel_name"),
+           nanobind::arg("explicit_measurements"));
+
+  mod.def("launch_sample", launch_sample, "FIXME: document",
+          nanobind::arg("policy"), nanobind::arg("ctx"),
+          nanobind::arg("callable"));
+}
--- a/python/runtime/cudaq/algorithms/py_sample.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample.cpp
@@ -21,10 +21,22 @@ static void construct_sample_policy(sample_policy *self, ExecutionContext &ctx,
   self->options.shots = ctx.shots;
   self->options.explicit_measurements = explicitMeasurements;
   self->noiseModel = get_platform().get_noise(ctx.qpuId);
+  // For now the noise model has to be duplicated on the policy unfortunately.
+  // get_noise() still expects it on the execution context.
+  // TODO: Store the noise on the platform or QPU instead.
+  ctx.noiseModel = self->noiseModel;
 }
 
 static sample_result launch_sample(sample_policy &policy, ExecutionContext &ctx,
                                    nanobind::callable callable) {
+  auto &platform = get_platform();
+  if (platform.is_remote()) {
+    async_sample_policy async_policy;
+    async_policy.inner = policy;
+    auto res = detail::launch(async_policy, ctx.qpuId, ctx, platform,
+                              [&]() { callable(); });
+    return res.get();
+  }
   return detail::launch(policy, ctx.qpuId, ctx, get_platform(),
                         [&]() { callable(); });
 }
--- a/python/runtime/cudaq/algorithms/py_sample.cpp
+++ b/python/runtime/cudaq/algorithms/py_sample.cpp
@@ -9,6 +9,7 @@
 #include "py_sample.h"
 #include "cudaq/algorithms/launch.h"
 #include "cudaq/algorithms/sample.h"
+#include "cudaq/algorithms/sample/policy.h"
 #include <nanobind/stl/string.h>
 
 using namespace cudaq;
@@ -42,12 +43,29 @@ static sample_result launch_sample(sample_policy &policy, ExecutionContext &ctx,
 }
 
 void cudaq::bindPySample(nanobind::module_ &mod) {
-  nanobind::class_<sample_policy>(mod, "sample_policy")
+  nanobind::class_<sample_policy>(mod, "SamplePolicy")
       .def("__init__", construct_sample_policy, nanobind::arg("ctx"),
-           nanobind::arg("kernel_name"),
-           nanobind::arg("explicit_measurements"));
+           nanobind::arg("kernel_name"), nanobind::arg("explicit_measurements"))
+      .def_prop_ro(
+          "kernel_name",
+          [](const sample_policy &policy) { return policy.kernelName; },
+          "The kernel name.")
+      .def_prop_ro(
+          "shots",
+          [](const sample_policy &policy) { return policy.options.shots; },
+          "The number of shots.")
+      .def_prop_ro("explicit_measurements",
+                   [](const sample_policy &policy) {
+                     return policy.options.explicit_measurements;
+                   })
+      .def("__repr__", [](const sample_policy &p) {
+        return cudaq_fmt::format("SamplePolicy(kernel_name='{}', shots={}, "
+                                 "explicit_measurements={})",
+                                 p.kernelName, p.options.shots,
+                                 p.options.explicit_measurements);
+      });
 
-  mod.def("launch_sample", launch_sample, "FIXME: document",
+  mod.def("launch_sample", launch_sample, "Policy based sample launch.",
           nanobind::arg("policy"), nanobind::arg("ctx"),
           nanobind::arg("callable"));
 }
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -93,9 +93,12 @@ protected:
     auto modulePtr = compileModulePreamble(src);
     CUDAQ_INFO("specializing remote rest kernel via module ({}) with {} policy",
                kernelName, policy.name);
+    // Temporary hack until we have a proper way of configuring the compiler
+    // based on the policy.
+    auto ctx = cudaq::getExecutionContext();
     Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
                       noiseModel, emulate);
-    return compiler.runPassPipeline(policy, kernelName, modulePtr, args);
+    return compiler.runPassPipeline(ctx, kernelName, modulePtr, args);
   }
 
 public:
@@ -274,6 +277,9 @@ public:
                       noiseModel, emulate);
     std::vector<cudaq::KernelExecution> codes;
 
+    // Temporary hack until we have a proper way of configuring the compiler
+    // based on the policy.
+    auto ctx = cudaq::getExecutionContext();
     if (std::holds_alternative<SourceModule>(module)) {
       const auto &src = std::get<SourceModule>(module);
       kernelName = src.getName();
@@ -281,7 +287,7 @@ public:
 
       auto [moduleOp, context] = Compiler::loadQuakeCodeByName(kernelName);
 
-      codes = compiler.lowerQuakeCode(policy, kernelName, moduleOp, args);
+      codes = compiler.lowerQuakeCode(ctx, kernelName, moduleOp, args);
     } else {
       const auto &compiled = std::get<CompiledModule>(module);
       kernelName = compiled.getName();
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -270,13 +270,13 @@ public:
   /// configured pass pipeline; pre-compiled modules are emitted directly.
   /// The resolved kernel name is returned via @p kernelName.
   template <typename Policy>
-  std::vector<cudaq::KernelExecution>
+  std::pair<std::string, std::vector<cudaq::KernelExecution>>
   compileKernelExecutions(Policy &policy, const AnyModule &module,
-                          KernelArgs args, std::string &kernelName) {
+                          KernelArgs args) {
     Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
                       noiseModel, emulate);
     std::vector<cudaq::KernelExecution> codes;
-
+    std::string kernelName;
     // Temporary hack until we have a proper way of configuring the compiler
     // based on the policy.
     auto ctx = cudaq::getExecutionContext();
@@ -295,7 +295,7 @@ public:
       codes = compiler.emitKernelExecutions(compiled);
     }
 
-    return codes;
+    return {kernelName, codes};
   }
 
   void completeLaunchKernel(const std::string &kernelName,
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -19,6 +19,7 @@
 #include "nvqir/AnalysisScope.h"
 #include "nvqir/resourcecounter/ResourceCounterScope.h"
 #include "cudaq/Support/TargetConfig.h"
+#include "cudaq/algorithms/sample/policy.h"
 #include "cudaq/platform/platform_iface.h"
 #include "cudaq/platform/qpu.h"
 #include "cudaq/platform/qpu_utils.h"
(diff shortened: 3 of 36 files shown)


python/cudaq/runtime/sample.py
```diff
@@ -68,13 +68,15 @@ def __broadcastSample(kernel,
...
+        kernel_name = kernel.name if hasattr(kernel, 'name') else ''
         ctx = cudaq_runtime.ExecutionContext('sample', shots_count)
+        ctx.kernelName = kernel_name
         ctx.totalIterations = N
         ctx.batchIteration = i
         ctx.explicitMeasurements = explicit_measurements
-        with ctx:
-            kernel(*a)
-        res = ctx.result
+        policy = cudaq_runtime.sample_policy(ctx, kernel_name,
```
lmondada: This is great! Happy that we are relying less and less on an implicit context

python/runtime/cudaq/algorithms/py_sample.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+static sample_result launch_sample(sample_policy &policy, ExecutionContext &ctx,
+                                   nanobind::callable callable) {
+  return detail::launch(policy, ctx.qpuId, ctx, get_platform(),
+                        [&]() { callable(); });
+}
+
+void cudaq::bindPySample(nanobind::module_ &mod) {
+  nanobind::class_<sample_policy>(mod, "sample_policy")
```
lmondada: I think we should follow Python's `CamelCase` naming convention:


```suggestion
  nanobind::class_<sample_policy>(mod, "SamplePolicy")
```

python/runtime/cudaq/algorithms/py_sample.cpp
```diff
@@ -0,0 +1,41 @@
...
+      .def("__init__", construct_sample_policy, nanobind::arg("ctx"),
+           nanobind::arg("kernel_name"),
+           nanobind::arg("explicit_measurements"));
```
lmondada: It's not strictly necessary but it would be easy (and nice to have) to expose some of the properties of the `SamplePolicy` as well as a string representation. I'm thinking something like
```c++
      .def_prop_ro("kernel_name", [](const SamplePolicy &policy) { return policy.kernelName; }, "..")
      .def_prop_ro("shots", [](const SamplePolicy &policy) { return policy.options.shots; }, "..")
      .def_prop_ro("explicit_measurements", [](const SamplePolicy &policy) { return policy.options.explicit_measurements; }, "..")
      .def("__repr__", [](const SamplePlolicy &p) {
            return std::format("SamplePolicy(kernel_name='{}', shots={}, explicit_measurements={})", p.kernelName, p.options.shots, p.options.measurements);
        });
```

Things to consider but that I don't have strong opinions on:
- which properties to expose (I suppose that depends on what types have python bindings)
- whether you want to mirror the C++ structure of having an "options" attribute or if we're fine with just exposing a couple of ad-hoc properties
- whether all properties should be read-only or some can be read-write
- what `__repr__` should look like. The [Python guidance](https://docs.python.org/3/reference/datamodel.html#object.__repr__) is quite rigid
  >  If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value
  
  but in this case this would involve also introducing a good `__repr__` for ExecutionContext, which I think is taking it too far.

python/runtime/cudaq/algorithms/py_sample.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+  mod.def("launch_sample", launch_sample, "FIXME: document",
```
lmondada: Just a note that you probably intended to add some doc here. We can keep it minimal.

runtime/common/Future.h
```diff
@@ -204,4 +204,11 @@ using async_observe_result = async_result<observe_result>;
 /// @brief Return type for asynchronous sampling.
 using async_sample_result = async_result<sample_result>;
 
+/// @brief Wrapper for a policy to return an async_result.
+template <typename InnerPolicy>
+struct policy_wrapper {
```
lmondada: maybe call this `async_policy_wrapper`?

runtime/cudaq/algorithms/launch.h
```diff
@@ -0,0 +1,71 @@
...
+namespace cudaq {
+namespace detail {
+
+/// @brief Execute the given function within the given execution context.
+template <typename Policy, typename Callable, typename... Args>
+auto launch(Policy &policy, std::size_t qpu_id, ExecutionContext &ctx,
+            quantum_platform &platform, Callable &&f, Args &&...args)
+    -> Policy::result_type {
+
+  // Python builds without CUDAQ_LIBRARY_MODE defined, so we need to check for
```
lmondada, line 29: This comment scares me: does library mode within Python actually exist?!

runtime/cudaq/algorithms/launch.h
```diff
@@ -0,0 +1,71 @@
...
+      return detail::with_policy_and_ctx(policy, ctx, [&]() {
+        return cudaq::ExecutionManager::with_default_em(
+            policy, [&]() { f(std::forward<Args>(args)...); });
+      });
+    }
+  }
+
+  typename Policy::result_type result;
+  auto &qpu = platform.getQPU(qpu_id);
+  ctx.executeKernelApi = [&qpu, &result, &policy, &ctx](
```
lmondada: I think you don't need to capture `ctx` in the lambda here

runtime/cudaq/algorithms/sample/policy.h
```diff
@@ -22,20 +23,36 @@ class ExecutionContext;
...
   /// Associated result type for synchronous APIs keyed off this policy.
   using result_type = sample_result;
 
   /// Sampling  options.
   sample_options options;
 
+  /// @brief The name of the kernel being executed.
+  std::string kernelName;
+
+  /// @brief Whether or not to simply concatenate measurements in execution
```
lmondada: This docstring isn't quite correct: my understanding is that `warnedNamedMeasurements` serves to track whether a warning has been emitted, to avoid emitting a warning multiple times.

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
```diff
@@ -11,6 +11,24 @@
 using namespace cudaq;
 cudaq::RemoteRESTQPU::~RemoteRESTQPU() = default;
 
+sample_result RemoteRESTQPU::launchKernel(sample_policy &policy,
+                                          const AnyModule &module,
+                                          KernelArgs args) {
+  CUDAQ_INFO("RemoteRESTQPU::launchKernel {}", policy.name);
+  std::string kernelName;
+  auto codes = compileKernelExecutions(policy, module, args, kernelName);
```
lmondada: I understand why you're passing the `policy` through to `compileKernelExecutions` here: it allows you to define multiple policy overloads of `Compiler::runPassPipeline`. However it would be better if we were able to pull out the differences between policies out of the compiler, so that no policy-based overloads are required on compiler methods.

The reason for that is that we want to avoid that each QPU and each policy define their own compiler pipeline. Instead, we should aim to have one unified pipeline, which can be tailored using custom "hooks" (or options or callbacks...) by the QPU & policies. In a way, we would keep the policy-based overloads more "localized", tailoring parts of the compilation pipeline rather than redefining it from scratch.

I've implemented a concrete proposal for this in PR #4599. The idea is to introduce a `getCompileTarget` method on QPUs, which can be overloaded and dispatched by QPU and policy. The `Compiler` is then passed the returned `CompileTarget` for configuration. Compilation then proceeds in the same way, independently of QPU and policy.

runtime/cudaq/platform/default/DefaultQPU.cpp
```diff
@@ -36,6 +36,25 @@ cudaq::DefaultQPU::unifiedLaunchModule(const cudaq::AnyModule &module,
   return rawFn->getFn()(argData, /*isRemote=*/false);
 }
 
+cudaq::sample_result
+cudaq::DefaultQPU::launchKernel(cudaq::sample_policy &policy,
+                                const cudaq::AnyModule &module,
+                                cudaq::KernelArgs args) {
+  CUDAQ_INFO("DefaultQPU::launchKernel {}", policy.name);
+  return cudaq::ExecutionManager::with_default_em(
```
lmondada, line 44: intriguing that this is the only case where the ExecutionManager is still needed. Can I ask why, out of curiosity? Do you plan to leave this or is this a stepping stone towards something else?

runtime/cudaq/platform/qpu.cpp
```diff
@@ -45,6 +46,20 @@ cudaq::QPU::unifiedLaunchModule(const AnyModule &module, KernelArgs args) {
   return runJITCompiledModule(compiled, args);
 }
 
+sample_result cudaq::QPU::launchKernel(sample_policy &policy,
```
lmondada, line 49: All the overloads that you had to explicitly mark as "not supported" made me realize how hard it is to mix templates and virtual methods: it's pretty much impossible to provide a default "not supported" implementation.

Presumably, if the QPU base class was a CRTP these problems would go away. Is this something you are still aiming for in future work? I don't have an opinion here, just a thought.

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -274,6 +281,14 @@ cudaq::CodeGenConfig quantum_platform::get_codegen_config() {
 const RuntimeTarget *quantum_platform::get_runtime_target() const {
   return runtimeTarget.get();
 }
+
+bool quantum_platform::is_library_mode() const {
```
lmondada, line 285: RE above comment: I don't know what library mode in Python is meant for, but it would be good if we could at least check with the team that this is needed.

runtime/internal/compiler/Compiler.cpp
```diff
@@ -416,6 +416,56 @@ cudaq::CompiledModule cudaq_internal::compiler::Compiler::runPassPipeline(
...
+  std::vector<std::pair<std::string, mlir::ModuleOp>> modules;
+  modules.emplace_back(kernelName, moduleOp);
+
+  bool needJit = emulate;
+  return assembleCompiledModule(kernelName, modules, needJit,
+                                emulate && combineMeasurements, std::nullopt,
+                                mapping_reorder_idx, context);
+}
+
+cudaq::CompiledModule cudaq_internal::compiler::Compiler::runPassPipeline(
```
lmondada: It's unfortunate because whilst I agree that these policy overloads remove the ugly conditional branchings within the method body, my problem is that my task for next release is to reduce the number of places where compilation pipelines are being defined. This does the opposite.

I think this is easily solvable though, let's talk

lmondada (commented):
Heroic set of changes, thank you @Renaud-K!

I agree with the vast majority of the changes and really think that it's going to make overloading launch policies easier down the line. Thank you again!

My one concern regards the policy-based overloads of the compiler (specifically, `QPU::compileModule`, `Compiler::lowerQuakeCode` and `Compiler:runPassPipeline`). I don't think we should introduce multiple overloads of these methods. The reason for that is that this will allow any combination of QPU and policy to define a fully custom compilation pipeline. This moves us in the opposite direction of where we need to go, namely one unified compilation pipeline shared by all QPUs with specific insertion points and configuration options (which QPUs and policies can customize to tweak compilation, inject custom passes or override behaviour).

Concretely, my proposal is that we move the policy-based overloads for compilation to a finer granularity: QPUs and policies can specify how to configure the `Compiler` instance at construction time, but then compilation proceeds identically, based on those settings, for all QPUs and policies. I have implemented what I propose in #4599.

I'm flexible on how to proceed, let's discuss to see what's easiest for you and let's us merge in your changes fastest.

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
Renaud-K: In this [commit](https://github.com/NVIDIA/cuda-quantum/pull/4557/changes/cc7ee62bba9e856d672ed338c08bd77627cd8489), I have undone the changes to the compiler interface. I have kept however the policy based API on the QPU but stop just shy of passing them to the compiler. It was the easiest way to proceed and I am thinking that it might be helpful to you as well since you would have the policy up to the point you would want the `CompileTarget`.

runtime/cudaq/platform/qpu.cpp
Renaud-K: This is still the plan and why we have #4416, #4445, #4469. These PRs have also helped with the inclusion of `qpu.h` into `launch.h` which is user visible.
```
#include "cudaq/platform/qpu.h"
...

namespace cudaq {
namespace detail {

/// @brief Execute the given function within the given execution context.
template <typename Policy, typename Callable, typename... Args>
auto launch(Policy &policy, std::size_t qpu_id, ExecutionContext &ctx,
            quantum_platform &platform, Callable &&f, Args &&...args)
    -> Policy::result_type {
...

  typename Policy::result_type result;
  auto &qpu = platform.getQPU(qpu_id);
  ctx.executeKernelApi = [&qpu, &result, &policy](const AnyModule &module,
                                                  const KernelArgs &args) {
    result = qpu.launchKernel(policy, module, args);
  };

  detail::try_finally(
      [&] {
        detail::with_policy_and_ctx(policy, ctx, std::forward<Callable>(f),
                                    std::forward<Args>(args)...);
      },
      [&] { ctx.executeKernelApi = nullptr; });
  return result;
}
```
 So that we can call:
 `result = qpu.launchKernel(policy, module, args);`

runtime/internal/compiler/Compiler.cpp
Renaud-K: These changes are now undone.

runtime/cudaq/platform/quantum_platform.cpp
Renaud-K: Needed by `test_photonics_kernel.py`

runtime/cudaq/algorithms/launch.h
Renaud-K: Yup. The [photonics tests](https://github.com/Renaud-K/cuda-quantum/blob/main/python/tests/handlers/test_photonics_kernel.py)

runtime/cudaq/algorithms/sample/policy.h
Renaud-K: Good catch. Thank you.

lmondada (approved):
Thank you for addressing the feedback! This latest version works very well for me.

lmondada:
hmm, I don't know why the main CI checks are not getting triggered... I've tried restarting the checks, but still nothing. https://www.githubstatus.com/ is not reporting any issues.

Renaud-K:
> /ok to test 7c8b4592dc819ef7ffd9d09d280d078579126fea

runtime/cudaq/algorithms/sample.h
```diff
@@ -103,36 +116,63 @@ runSampling(KernelFunctor &&wrappedKernel, quantum_platform &platform,
...
+            std::size_t qpu_id = 0, std::size_t batchIteration = 0,
+            std::size_t totalBatchIters = 0) {
 
-  // Indicate that this is an asynchronous execution.
-  ctx.asyncExec = futureResult != nullptr;
+  if (platform.is_remote(qpu_id)) {
+    auto noise = platform.get_noise(qpu_id);
+    std::optional<noise_model> noise_opt =
+        noise ? std::make_optional(*noise) : std::nullopt;
+    auto restult = runSamplingAsync(wrappedKernel, platform, kernelName, shots,
```
sacpis: nit: result

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -387,6 +447,83 @@ class BaseRemoteRESTQPU : public QPU {
...
+    if (policy.inner.options.shots != std::numeric_limits<std::size_t>::max() &&
+        policy.inner.options.shots != 0)
+      localShots = policy.inner.options.shots;
+
+    executor->setShots(localShots);
+
+    auto executionContext = cudaq::getExecutionContext();
+    // Execute the codes produced in quake lowering
+    // Allow developer to disable remote sending (useful for debugging IR)
+    assert(!emulate);
```
sacpis, line 466: I think we should use `std::runtime_error` or `std::logic_error` as these asserts will vanish in the release builds.

runtime/cudaq/platform/default/DefaultQPU.cpp
```diff
@@ -36,6 +36,25 @@ cudaq::DefaultQPU::unifiedLaunchModule(const cudaq::AnyModule &module,
...
+      [this, &module, &args]() { this->unifiedLaunchModule(module, args); });
+}
+
+cudaq::async_sample_result
+cudaq::DefaultQPU::launchKernel(async_sample_policy &policy,
+                                const cudaq::AnyModule &module,
+                                cudaq::KernelArgs args) {
+  throw std::runtime_error(
+      "DefaultQPU does not support launching the async_sample_policy.");
+  return async_sample_result{};
```
sacpis: Seems like a dead code after throw.

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -235,6 +237,11 @@ CompiledModule quantum_platform::compileModule(const SourceModule &src,
                                                bool isEntryPoint) {
   validateQpuId(qpu_id);
   auto &qpu = platformQPUs[qpu_id];
+  auto ctx = cudaq::getExecutionContext();
+  if (ctx && ctx->name == "sample") {
```
sacpis, line 241: Instead of checking this, should we lookup the policy on the ctx or push the policy into this signature? Seems like it defeats the point of moving to policy objects.

runtime/cudaq/qis/execution_manager.h
```diff
@@ -16,7 +16,9 @@
 #include "cudaq/host_config.h"
 #include "cudaq/operators.h"
 #include "cudaq/qis/measure_handle.h"
+#include "cudaq/utils/cudaq_utils.h"
 #include <deque>
+#include <iostream>
```
sacpis: nit: can be removed, as we are not using it.

runtime/common/BaseRemoteRESTQPU.h
Renaud-K: I do believe that an assertion is appropriate here. This would not be a user error. We use assertions a lot to emphasize or communicate our assumptions.

runtime/cudaq/platform/quantum_platform.cpp
Renaud-K: In the python compilation flow, the policies are not generated from the top yet. There is still a lot of refactoring happening in this space. This is the place, selected to introduce them as it does not interfere with it. When the support for observe is added, this will be nicer with the policy dispatcher utility but it will still map a string to a policy overload. As the compilation refactor proceeds they should bubble up closer to the top. It does not defeat anything. We are in a transition phase where policies gain territory and execution context retreats.

runtime/common/BaseRemoteRESTQPU.h
sacpis: It might be. But it will not appear in the release build, just FYI. Unless we don't want it in there.

runtime/cudaq/qis/execution_manager.h
Renaud-K: fixed.

runtime/cudaq/algorithms/sample.h
Renaud-K: fixed.

Renaud-K:
> /ok to test f052c03092e6601e1f0fb27fc3e8565ab0dabc9f

runtime/cudaq/algorithms/sample.h
```diff
@@ -103,36 +116,63 @@ runSampling(KernelFunctor &&wrappedKernel, quantum_platform &platform,
...
+      counts += result;
 
-    ctx.result.clear();
     if (counts.get_total_shots() == 0) {
       if (explicitMeasurements)
         throw std::runtime_error(
-            "The sampling option `explicit_measurements` is not supported on a "
+            "The sampling option `explicit_measurements` is not supported on "
+            "a "
             "kernel without any measurement operation.");
```
1tnguyen: nit: perhaps we can combine the text lines

runtime/cudaq/platform/fermioniq/FermioniqQPU.cpp
```diff
@@ -75,4 +75,30 @@ void cudaq::FermioniqQPU::launchImpl(const cudaq::CompiledModule &compiled) {
   completeLaunchKernel(compiled.getName(), std::move(codes));
 }
 
+cudaq::sample_result
+cudaq::FermioniqQPU::launchKernel(cudaq::sample_policy &policy,
+                                  const AnyModule &module, KernelArgs args) {
+  std::string kernelName;
+  auto codes = compileKernelExecutions(policy, module, args, kernelName);
```
1tnguyen: Just a comment: what do you think about this helper to return both the `codes` and `kernelName`, e.g.,

`auto [kernelName, codes] = compileKernelExecutions(policy, module, args);`

?

1tnguyen (approved):
LGTM 👍

runtime/cudaq/algorithms/sample.h
Renaud-K: Looks like something clang-format did.

runtime/cudaq/platform/fermioniq/FermioniqQPU.cpp
Renaud-K: It would be a bit better, I guess.

Renaud-K:
> /ok to test 0458d09f1e2bc77a8d41b488235ad35a111dee7f

runtime/cudaq/algorithms/sample.h
Renaud-K: fixed.

runtime/cudaq/platform/fermioniq/FermioniqQPU.cpp
Renaud-K: fixed.

sacpis (approved):

runtime/cudaq/platform/quantum_platform.cpp
sacpis: Got it. I was unaware of the following PRs for this effort. Thanks @Renaud-K.

PR #4559, opened 2026-05-20
remove codecov from deployments

mitchdz:
code coverage is failing in deployments. Let's just remove it. It doesn't make a ton of sense there, as code coverage will be done in CI anyways.

It is currently disabled in CI while I work on getting the performance better, but once it is enabled in CI, then deployments is redundant and wasting runner time.


mitchdz:
No need to run this one through merge queue, can just push directly to main.

bettinaheim (approved):

PR #4561, opened 2026-05-20
Lower `qec` ops to QIR runtime

khalatepradnya:
### Summary

Lower `qec.detector` / `qec.observable` / `qec.pair_detectors` to QIR runtime calls inside the existing `--convert-to-qir-api` patterns. 

### What Changed

- Three `OpConversionPattern`s in `ConvertToQIRAPI.cpp` plus a `packMeasurementHandles` helper that handles the all-scalar / single-stdvec / mixed-or-multi-stdvec shapes in one place.
- Single-stdvec form: variadic ops dispatch one `(Result**, count)` call regardless of operand mix. `pair_detectors` keeps `(prev, prev_size,  curr, curr_size)` so the runtime can enforce length agreement.
- Runtime symbols: `__quantum__qis__detector`, `__quantum__qis__logical_observable`, `__quantum__qis__pair_detectors`.


sacpis (approved):
LGTM. Thanks @khalatepradnya.

PR #4564, opened 2026-05-20
[runtime][nvqir] Wire `qec` runtime calls to `stim`

khalatepradnya:
### Summary

- Wire the three QIR runtime symbols PR #4561 declared (`__quantum__qis__detector`, `__quantum__qis__logical_observable`, `__quantum__qis__pair_detectors`) into the Stim NVQIR backend so kernels containing `qec.*` ops do something at runtime instead of resolving to unresolved externals.

### What changed

- NVQIR adapters (`runtime/nvqir/NVQIR.cpp`): three `extern "C"` functions. Each `Result*` entry in the ABI buffer is the bit pattern of an `i64` chronological measurement index; the adapter reinterprets it through `intptr_t` and dispatches the resulting `std::vector<int64_t>` to the active simulator's QEC virtual. 

- Simulator base virtual functions (`runtime/nvqir/CircuitSimulator.h`): Non-QEC-capable backends silently drop QEC declarations; no per-target erase pass needed.

- Stim override (`runtime/nvqir/stim/StimCircuitSimulator.cpp`): new `stim::Circuit recordedCircuit` member accumulating gates, measurements, and noise (`applyOpToSims` and the noise-channel branch each append). The three overrides:

- Lifecycle: `recordedCircuit` is cleared alongside `num_measurements` in `deallocateStateImpl` and in `setToZeroState`'s preserve-simulators branch. Successive `cudaq::sample` invocations on the reused thread-local simulator would otherwise stack new gates on top of the previous kernel's record while `rec[-N]` restarts from zero, producing a malformed DEM. Consumers that need the recorded circuit must read it before `endExecution` runs.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -23,6 +23,7 @@
 #include <concepts>
 #include <cstdarg>
 #include <cstddef>
+#include <cstdint>
 #include <iostream>
 #include <queue>
 #include <sstream>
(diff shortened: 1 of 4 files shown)


bmhowe23 (approved):
Looking good ... thanks, @khalatepradnya!

PR #4565, opened 2026-05-21
[CUDAQ-790] `cc.device_call` lowering to realtime dispatch for local simulators

1tnguyen:
Support lowering of `cc.device_call` to realtime-based ring-buffer dispatch.

 `cc.device_call` is lowered to three ABI calls that operate on a transport-neutral frame lease - a leased `RX`/`TX` slot pair owned by the runtime for the duration of one RPC:

```
  __cudaq_device_call_acquire_realtime_frame  // lease an in-flight frame
  __cudaq_device_call_dispatch_realtime_frame // publish + wait
  __cudaq_device_call_safely_release_realtime_frame
```
  
The compiler writes function arguments directly into the leased `RX` payload and reads results directly from the leased `TX` payload. 

- Scalars are stored at aligned offsets in the request slot
- `std::vector<T>` args use a length prefix followed by element bytes packed in place. 
- Output `std::vector<T>&` arguments (for CUDA-QX interop) are signalled via an attribute on `cc.device_call` and read back through the same zero-copy response slot. 

  The runtime side (`runtime/internal/device_call/`) wraps `cudaq_ringbuffer_t` from CUDA-Q Realtime: a singleton driver routes per-device sessions to a `DeviceCallChannel`, with two built-in shared-memory channels: `device_dispatch` (persistent GPU dispatch kernel) and `host_dispatch` (graph-launch with a pinned mailbox).

**Notes**:
 - This PR handles local shared-memory only; no process separation or RDMA transport in this branch. This is deferred to a follow-up.
  
 - CUDA-Q builds without `realtime` by default; the path is opt-in via `CUDAQ_REALTIME_DIR` at configure time (points to a `realtime` installation) and
 
  - CI here is build-only: `realtime_integration_ci.yml` compiles a realtime-enabled CUDA-Q against the realtime installation.

**Tested by:**

  - `DeviceCallDispatchTester.cu`:  test runtime ring-buffer / channel behavior
  - End-to-end `NVQPP/device_call_realtime_{scalar,array}.cpp` tests of shared-memory and host-dispatch channels for different arguments and return type with simulators in the loop.


realtime/CMakeLists.txt
```diff
@@ -120,6 +123,28 @@ SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libstdc++ -static-
...
 
+if (CUDA_FOUND)
+  configure_package_config_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/cmake/cudaq-realtime-config.cmake.in
+    ${CMAKE_CURRENT_BINARY_DIR}/cmake/cudaq-realtime-config.cmake
+    INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/cudaq-realtime)
+
+  write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/cmake/cudaq-realtime-config-version.cmake
+    VERSION 0.1.0
```
bmhowe23: I thought we were past version 0.1.0 now?

1tnguyen: Yeah, this is a boilerplate which should have been removed as we don't need this.
Removed in https://github.com/NVIDIA/cuda-quantum/pull/4565/changes/b20e088ee8a011965286f46642cc7827ccb7e0d2

cmake/modules/CUDAQConfig.cmake
```diff
@@ -133,6 +143,20 @@ if(NOT TARGET cudaq::cudaq)
     include("${CUDAQ_CMAKE_DIR}/CUDAQTargets.cmake")
 endif()
 
+foreach(_cudaq_config DEBUG RELEASE RELWITHDEBINFO MINSIZEREL NOCONFIG)
+  get_target_property(_cudaq_link_deps cudaq::cudaq
+    IMPORTED_LINK_DEPENDENT_LIBRARIES_${_cudaq_config})
+  if (_cudaq_link_deps MATCHES "cudaq::cudaq-realtime" AND
```
sacpis: Seems like this loop iterates `IMPORTED_LINK_DEPENDENT_LIBRARIES_${_cudaq_config}` but only `cudaq-device-call-runtime` actually links against `cudaq::cudaq-realtime` [here](https://github.com/NVIDIA/cuda-quantum/pull/4565/changes#diff-e3ec6c5f5eea0bbe955158d2557c1481956cf571b9d86b6521e533e69164eebdR37). Seems like this check no-ops on every install? Should we either inspect `cudaq::cudaq-device-call-runtime` or drop the loop?

sacpis (approved):
LGTM. Thanks @1tnguyen.

cmake/modules/CUDAQConfig.cmake
1tnguyen: Good catch! This block was stale: after I converted the `cudaq::realtime::initialize/finalize` to plugin-based to remove the dependency, it is no longer needed. 

Removed in https://github.com/NVIDIA/cuda-quantum/pull/4565/changes/50ad8a746991cbf350788f18119da72245df0a07

runtime/internal/device_call/DeviceCallChannel.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+#include "cudaq/runtime/logger/logger.h"
+#include "cudaq/utils/cudaq_utils.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Path.h"
+
+#include <cstdlib>
+#include <dlfcn.h>
```
Destroyerrrocket, line 20: We're only supporting posix right now? genuine question, I don't know

1tnguyen: We're officially supporting Linux and MacOS. For this particular component, which depends on CUDA-Q `realtime` (with CUDA dependency), only Linux is expected to support.

.github/workflows/realtime_integration_ci.yml
```diff
@@ -0,0 +1,168 @@
+on:
+  workflow_dispatch:
+    inputs:
+      cache_base:
+        required: false
+        type: string
+        description: 'The name of the branch to use as cache base.'
+        default: main
+  push:
```
mitchdz, line 9: Do you want to run this on `merge_group`? Then it will have to be added in the push section.

mitchdz (approved):
I'm mainly just reviewing the github workflows. It looks good to me, other than maybe needing to add to merge group.

.github/workflows/realtime_integration_ci.yml
1tnguyen: Good point! Added in https://github.com/NVIDIA/cuda-quantum/pull/4565/changes/bac6c1c3300a29faa7ce3d65abb1ec74c45c2542

PR #4569, opened 2026-05-21
Introduce CompileOptions and plumb through Compiler API

lmondada:
This PR sets up the plumbing to define compilation options in a new `CompileOptions` struct. Such an options struct can get constructed from an execution context using `CompileOptions::fromExecutionContext`. The first goal is to remove any dependency on the execution context within `Compiler::runPassPipeline` (and hence also from `Compiler::lowerQuakeCode`).

I wanted to get this first PR out today. I'll open a follow up PR tomorrow with more config options.

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -234,8 +234,11 @@ public:
 
     Compiler compiler(serverHelper.get(), backendConfig, targetConfig,
                       noiseModel, emulate);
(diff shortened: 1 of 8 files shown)


Renaud-K:
The dependencies on the execution context will be removed by the policies. I would like to get this #4557 in first. Then under the hood, compiler options can be created.

lmondada:
I'm reworking this in favour of a `CompileTarget` type

PR #4573, opened 2026-05-21
[C++] Allow `std::vector<measure_handle>::operator=` in kernels

khalatepradnya:
Unblocks the QEC cross-round detector pattern `(auto m = mz(qv); for (...) { auto m_new = mz(qv); ...; m = m_new; })` by enabling `std::vector<measure_handle>::operator=` in `__qpu__` kernels. Handle-vector locals now get stack storage when they're declared, and assignment lowers to `cc.store` through the LHS address. Mirrors the scalar `measure_handle` precedent; non-handle stdvec types stay on the legacy SSA-binding path.

### DIFF
--- a/cudaq/lib/Frontend/nvqpp/ConvertDecl.cpp
+++ b/cudaq/lib/Frontend/nvqpp/ConvertDecl.cpp
@@ -743,11 +743,40 @@ bool QuakeBridgeVisitor::VisitVarDecl(clang::VarDecl *x) {
     if (x->getInit()) {
       // At the very least, its a vector var = vec_init;
       auto initVec = popValue();
-      symbolTable.insert(x->getName(), initVec);
+      auto elementType = vecType.getElementType();
+
+      // For `std::vector<measure_handle>` locals only, allocate stack
+      // storage for the descriptor so reassignment (`v = w;`) and
+      // cross-scope mutation (`if (cond) { v = w; }`) lower to `cc.store`
+      // into a stable per-name address.
+      //
+      // Mirrors the scalar `measure_handle` allocation in the basic-type
+      // path and the `measure_handle::operator=` arm. Within a kernel scope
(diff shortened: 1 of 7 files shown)


cudaq/lib/Frontend/nvqpp/ConvertDecl.cpp
```diff
@@ -743,11 +743,40 @@ bool QuakeBridgeVisitor::VisitVarDecl(clang::VarDecl *x) {
...
+      // For `std::vector<measure_handle>` locals only, allocate stack
+      // storage for the descriptor so reassignment (`v = w;`) and
+      // cross-scope mutation (`if (cond) { v = w; }`) lower to `cc.store`
+      // into a stable per-name address.
+      //
+      // Mirrors the scalar `measure_handle` allocation in the basic-type
+      // path and the `measure_handle::operator=` arm. Within a kernel scope
+      // `std::vector<measure_handle>` has `std::span<measure_handle>`-
+      // like value semantics: the `(data_ptr, size)` descriptor is what
+      // gets copied; element mutations land in the underlying storage
```
1tnguyen, line 757: If I read this correctly, we're doing a shallow copy (pointer only) here.

This might be problematic for the case like in the new test:

```
    cudaq::qvector qv(2);
    auto m = mz(qv);
    if (cond) {
      cudaq::qvector qv2(2);
      auto m_new = mz(qv2);
      m = m_new;
    }
    return m[0];
```

If `m` holds a pointer to `m_new`, maybe we could have a scoping problem (`m_new` went out-of-scope). 
Perhaps we can add an execution test as well to make sure.

khalatepradnya: This is something I missed by using "span"-like behavior.
Let me implement proper fix, converting to "draft".

khalatepradnya:
Captured in new issue https://github.com/NVIDIA/cuda-quantum/issues/4601, closing this PR.

PR #4599, opened 2026-05-22
Introduce CompileTarget and plumb it through the Compiler API

lmondada:
I've reworked #4569 to work better with @Renaud-K's proposed changes in #4557. ~Needs to be merged after https://github.com/NVIDIA/cuda-quantum/pull/4614. For review, it is sufficient to look at commit [8cc13fc47b](https://github.com/NVIDIA/cuda-quantum/pull/4599/changes/8cc13fc47bfd74cdd37a39ad15919b9d16780f1f).~ done.

This PR sets up the plumbing to define compilation options in a new `CompileTarget` struct. This struct is currently constructed and populated by the QPU class using the `QPU::getCompileTarget` method. The method has overloads for each launch policy, and, as a temporary stopgap, falls back to taking an executionContext as argument.

The main conceptual change is _where_ the compilation pipeline is constructed.

Before:
1. TargetConfig + RuntimeConfig are parsed in `BaseRemoteRESTQPU`
2. TargetConfig + RuntimeConfig + ServerHelper are passed to `Compiler` constructor, which builds the compilation pipeline. _This had to happen within the `QPU`, as it owns the inputs to the constructor._
3. `Compiler::runPassPipeline` executes the computed pipeline

After:
1. TargetConfig + RuntimeConfig are parsed in `BaseRemoteRESTQPU`
2. TargetConfig + RuntimeConfig + ServerHelper are combined into a `CompileTarget`, which the QPU constructs and returns.
3. A `Compiler` instance is constructed from the `CompileTarget`.
4. `Compiler::runPassPipeline` computes the pass pipeline from the compile target using `cudaq_internal::compiler::getPassPipeline` and executes the pipeline

The crucial difference is that the new flow allows us to construct the `Compiler` instance and execute the pipeline outside of the QPU. In other words, `CompileTarget` represents the minimum amount of information from the QPU config that is required to determine the pass pipeline.

---

Note that the arguments to `getCompileTarget` are currently unused because the inner code still relies on the global execution context. This will change in an upcoming pr, which will remove any dependency on the execution context within `Compiler::runPassPipeline` (and hence also from `Compiler::lowerQuakeCode`): the compiler pipeline would instead be fully determined from the `CompileTarget`.

In the future, we may choose to separate `CompileTarget` out of `QPU` and have the source of truth be outside of the QPU class.


runtime/cudaq/platform/qpu.cpp
```diff
@@ -104,6 +104,22 @@ cudaq::CompiledModule cudaq::QPU::compileModule(sample_policy &,
...
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::QPU::getCompileTarget(ExecutionContext *context) {
+  throw std::runtime_error("no CompileTarget defined for this QPU");
+}
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::QPU::getCompileTarget(sample_policy &) {
+  return getCompileTarget();
+}
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::QPU::getCompileTarget(async_sample_policy &) {
```
Renaud-K: I have been avoiding overloads of async policies when the sync vs async does not apply. You should be able to pass the inner policy in the calling scope.

runtime/cudaq/platform/qpu.cpp
```diff
@@ -104,6 +104,22 @@ cudaq::CompiledModule cudaq::QPU::compileModule(sample_policy &,
   return compileModule(src, args, isEntryPoint);
 }
 
+std::unique_ptr<cudaq::CompileTarget> cudaq::QPU::getCompileTarget() {
+  return getCompileTarget(getExecutionContext());
+}
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::QPU::getCompileTarget(ExecutionContext *context) {
+  throw std::runtime_error("no CompileTarget defined for this QPU");
```
Renaud-K, line 111: I did not see where this is overridden.

cudaq/include/cudaq/Target/CompileTarget.h
```diff
@@ -0,0 +1,64 @@
...
+  /// Hook to update the pass pipeline before compilation.
+  virtual void updatePassPipeline(std::string &passPipeline) const {}
+
+  /// Resolved MLIR pass-pipeline and `codegen` settings.
+  struct PipelineConfig {
+    /// If set, override compilation pipeline with this string.
+    std::string overridePassPipeline;
+
+    /// Compilation pipeline to insert at the appropriate stages.
+    std::string JITHighLevelPipeline;
```
Renaud-K: Lower case JIT?

runtime/cudaq/platform/qpu.cpp
lmondada: You're right, it was superfluous. Removed.

runtime/cudaq/platform/qpu.cpp
lmondada: My mistake. When introducing overloads I implemented the wrong one in `BaseRemoteRESTQPU`. Fixed.

cudaq/include/cudaq/Target/CompileTarget.h
lmondada: This was mirroring the names in `TargetConfig`; but I agree that it should be lowercase. I've solved the conundrum by removing the "JIT" prefix altogether -- it's redundant.

runtime/cudaq/platform/qpu.h
```diff
@@ -153,6 +154,15 @@ class QPU : public registry::RegisteredType<QPU> {
   [[nodiscard]] virtual KernelThunkResultType
   unifiedLaunchModule(const AnyModule &module, KernelArgs args);
 
+  /// Get the compile target of the QPU
+  // Overload for sample policy
+  [[nodiscard]] virtual std::unique_ptr<CompileTarget>
+  getCompileTarget(sample_policy &policy);
+  // Overloads for currently unsupported policies (to be removed).
+  [[nodiscard]] virtual std::unique_ptr<CompileTarget> getCompileTarget();
```
Renaud-K, line 162: Does this have to be virtual? It seems it always takes the global execution context and calls another virtual method. Maybe we can just call a cat, a cat and only call the version with the explicit execution context?

lmondada: I used to only have the version with the execution context. I thought you'd want to introduce policy overloads and phase out the execution context dependency asap, hence the overloads. In the next PRs I will introduce flags in CompileTarget that get set depending on the launch policy, so the overloads will be useful for that. See e.g. [here](https://github.com/NVIDIA/cuda-quantum/pull/4608/changes#diff-a1cdb82adcbf0128e8d4caf1a988ab134711dbe83d47f7195eafb6e7c0a589a5R283). Happy to remove them, though, if you prefer.

As to whether this has to be virtual: QPUs will define their own overloads (and possibly return pointers to custom derived classes of CompileTarget), so I think there is no way around this polymorphism.

Renaud-K: I was asking specifically about ` virtual std::unique_ptr<CompileTarget> getCompileTarget();`.

Renaud-K: It's not a blocking issue. I can push the PR so it is in by tomorrow.

Renaud-K (approved):
It looks good. Thank you.

runtime/cudaq/platform/qpu.h
lmondada: ah I see. Yes that overload does not need to be virtual. I'll change that in the next PR,

PR #4566, opened 2026-05-21
[qec] Add `detector`, `logical_observable`, `detectors` source-language support

khalatepradnya:
### Summary

Adds the QEC source-language surfaces for detector, logical-observable, and paired-detector declarations in both C++ and Python.

Kernel source and builder calls now emit the `qec.detector`, `qec.observable`, and `qec.pair_detectors` ops added by earlier PRs. 

_Note: `cudaq::dem_from_kernel` remains follow-up scope._

### Motivation

Prior PRs added the QEC dialect, QIR lowering, and Stim runtime wiring. This PR makes that path reachable from user kernels and programmatic builders, while keeping the DEM analysis API isolated to the next PR.

_Note: https://github.com/NVIDIA/cuda-quantum/pull/4539 required for hardware targets._

### What Changed

- Adds C++ `cudaq::detector`, `cudaq::logical_observable`, and `cudaq::detectors` kernel-side declarations.
- Lowers those declarations through the C++ AST bridge and C++ `make_kernel` builder.
- Adds Python AST-bridge and `cudaq.make_kernel` support for the same operations.
- Adds Python QEC dialect bindings.
- Supports scalar/list/mixed measurement-handle operands for detector and logical-observable declarations where the dialect accepts them.

### Risks and Downstream Impact

The new source surfaces emit QEC declaration ops, but this PR does not add the user-facing DEM-generation API. End-to-end `.dem` generation lands in the follow-up PR. No CUDA-QX API change.

PR #4520, opened 2026-05-15
Add `qec` dialect skeleton

khalatepradnya:
### Summary

* First PR of the QEC detectors and observables task. 
* Adds the `qec` MLIR dialect with three ops:
```
  qec.detector(measurements...) : (...)
  qec.detectors(prev, curr) : (!cc.stdvec<!cc.measure_handle>, ...)
  qec.observable(measurements...) [index N] : (...)
```
* Operands are `!cc.measure_handle` or `!cc.stdvec<!cc.measure_handle>`. The dialect ships with parse/print support, registry hooks, explicit links from `cudaq-opt`, `cudaq-quake`, `cudaq-translate`, and the runtime MLIR layer.

* No transforms, lowerings, runtime, or AST bridge yet; those land in follow-up PRs.

Made-with: Cursor

PR #4602, opened 2026-05-22
Add support to compute detector error model (DEM) from kernel

khalatepradnya:
### Summary

* Adds the user-facing `cudaq::dem_from_kernel` analysis primitive and Python binding `cudaq.dem_from_kernel`. 
* The call runs a CUDA-Q kernel under an internal execution context, drives the Stim NVQIR plugin, and returns
the resulting detector error model as a UTF-8 string in Stim's standard [`.dem` file format](https://github.com/quantumlib/Stim/blob/main/doc/file_format_dem_detector_error_model.md).
* The active CUDA-Q target is unaffected.

This PR closes the source → AST → QEC ops → QIR → NVQIR → Stim → `.dem` path opened by #4520,
#4539, #4561, #4564, and #4566.

### Design

* `cudaq::dem_from_kernel` mirrors the `cudaq::estimate_resources` shape: it creates an internal `"dem"` execution context, claims the thread-local NVQIR analysis simulator slot via `nvqir::dem::make_scope`, runs the kernel through the platform execution path, and returns the DEM text populated during `CircuitSimulator::finalizeExecutionContext`.

* Stim remains behind the NVQIR plugin interface.

### What Changed

- **Analysis core** — `runtime/cudaq/analysis/Dem.cpp` sets up the `"dem"` context, rejects conditional-feedback kernels, runs the kernel under `nvqir::dem::make_scope`, and returns `ctx.dem_text` after normal platform finalization.
- **Compiler / QPU integration** — `Compiler::runPassPipeline`, `emitKernelExecutions`, and REST QPU completion now treat `"dem"` as an analysis context: preserve the analysis JIT artifact, avoid target transport code emission, and run locally for remote emulation like resource counting.
- **JIT cleanup** — `runtime/internal/compiler/JIT.cpp` separates WireSet qubit setup/teardown from result-map cleanup and injects `__quantum__rt__clear_result_maps` for all JIT entry points.
- **Python binding** — `cudaq.dem_from_kernel(kernel, *args, noise_model=None)` mirrors the C++ API and returns Stim `.dem` text.


runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // Set the thread-local context without running the active platform's
+    // configure/finalize hooks. DEM drives the Stim analysis simulator directly
+    // through `demScope`; the launch hook only redirects kernels to local JIT /
+    // raw-function execution so remote targets do not send analysis work to
+    // their transport.
+    cudaq::detail::try_finally(
+        [&]() {
+          withDemExecutionContext(ctx, [&]() {
+            kernel();
+            recorded = *recorder.circuit();
```
sacpis: Can we move the recorded circuit and not copy it? As it involves full copy of every gate, noise channel, detector. As the recorder is about to reset on scope exit, we can define `release()` on RecordedCircuit that returns `stim::Circuit&&` so that the caller moves the circuit out and then reset becomes a no op.

runtime/nvqir/dem/DemScope.cpp
```diff
@@ -0,0 +1,64 @@
...
+      recorder->reset();
+    else
+      throw std::runtime_error(
+          "`nvqir::dem::make_scope`: plugin simulator does not implement "
+          "`nvqir::RecordedCircuit` and therefore cannot drive DEM "
+          "analysis.");
+  };
+  return AnalysisScope::from_plugin(
+      std::move(name), std::move(plugin_name),
+      {.on_enter = resetRecordedCircuit, .on_exit = resetRecordedCircuit});
```
sacpis: Seems like we are passing the same `resetRecordedCircuit` for both `on_enter` and `on_exit`. Once we have extracted the circuit, the on_exit reset is kind of a no op?

runtime/nvqir/dem/DemScope.cpp
```diff
@@ -0,0 +1,64 @@
...
+#endif
+  const std::string lib = "libnvqir-" + plugin_name + libExt;
+  void *handle = dlopen(lib.c_str(), RTLD_NOW | RTLD_GLOBAL);
+  if (!handle) {
+    const char *err = dlerror();
+    throw std::runtime_error(
+        "`nvqir::dem::make_scope`: failed to load NVQIR plugin '" + lib +
+        "': " + (err ? err : "unknown dlerror"));
+  }
+}
```
sacpis, line 48: Instead of calling dlopen on every `make_scope`, can we cache it using similar pattern we used in dem header [here](https://github.com/NVIDIA/cuda-quantum/pull/4602/changes#diff-3d8a0f8d057d72e83a9908e12547f58489486e5f82130c9f384d7b45f051efb3R53).

runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,214 @@
...
+          // `__quantum__qis__mz_handle__to__register`.
+          __quantum__rt__clear_result_maps();
+        });
+  }
+
+  stim::DetectorErrorModel stimDem =
+      stim::ErrorAnalyzer::circuit_to_detector_error_model(
+          recorded,
+          /*decompose_errors=*/false,
+          /*fold_loops=*/false,
```
sacpis: For multi-circuits emitted by `cudaq::detector` loops, enabling fold_loops lets Stim compress repeated structure.
Please see this
```
When a circuit contains a `REPEAT` block, the structure of the
detectors often settles into a form that is identical from iteration
to iteration. Specifying the `--fold_loops` option tells Stim to
watch for periodicity in the structure of detectors by using a
"tortoise and hare" algorithm (see
https://en.wikipedia.org/wiki/Cycle_detection ).
This improves the asymptotic complexity of analyzing the loop from
O(total_repetitions) to O(cycle_period).
```
https://github.com/quantumlib/Stim/blob/main/doc/usage_command_line.md

Can we expose it as an arg or auto enable it when the recorded circuit contain REPEAT blocks?

runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,214 @@
...
+/// Serialize concurrent `dem_from_kernel` calls across threads.
+///
+/// `nvqir::AnalysisScope` claims a thread-local slot, but the simulator
+/// pointer it resolves through `dlsym(getCircuitSimulator_<plugin>)` is a
+/// process-wide singleton owned by the NVQIR plugin. The mutex therefore guards
+/// the kernel-execution + circuit-readback pair plus the on-entry simulator
+/// reset done by `nvqir::dem::make_scope`'s `on_enter` hook; once we have a
+/// local `stim::Circuit` copy, `ErrorAnalyzer` operates on it lock-free so
+/// concurrent analyses do not serialize on the error analysis step.
+std::mutex &demMutex() {
```
sacpis: Can we make the Analysis simulator per thread?

sacpis (commented):
Overall LGTM. Thanks @khalatepradnya.

runtime/cudaq/algorithms/dem.h
```diff
@@ -0,0 +1,117 @@
...
+                                    std::string plugin_name = "stim") {
+  using FnT = decltype(&cudaq_runDemFromKernel);
+  static FnT fn = []() -> FnT {
+#if defined(__APPLE__)
+    constexpr const char *kLib = "libcudaq-analysis.dylib";
+#else
+    constexpr const char *kLib = "libcudaq-analysis.so";
+#endif
+    constexpr const char *kSym = "cudaq_runDemFromKernel";
+    void *handle = dlopen(kLib, RTLD_NOW | RTLD_GLOBAL);
```
Renaud-K: I see where the library is used now. If we go by this logic all algorithms and analyses dynamically load their implementation now. The runtime is not going down that path. This seems unnecessary. You are already achieving zero cost by not including dem.h in cudaq.h.

runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,214 @@
...
+  cudaq::detail::setExecutionContext(&ctx);
+  auto cleanup = [&]() {
+    cudaq::detail::resetExecutionContext();
+    if (outerContext)
+      cudaq::detail::setExecutionContext(outerContext);
+  };
+  cudaq::detail::try_finally(std::forward<Callable>(callable), cleanup);
+}
+
+std::string runDemFromKernelImpl(const std::string &kernelName,
```
Renaud-K: We need to take a step back here and understand why you need all this specialization. As much as possible we need to leverage our architecture. It seems we are creating completely new pathways here.

runtime/cudaq/algorithms/dem.h
```diff
@@ -0,0 +1,117 @@
...
+                                    const std::function<void()> &wrappedKernel,
+                                    std::string plugin_name = "stim") {
+  using FnT = decltype(&cudaq_runDemFromKernel);
+  static FnT fn = []() -> FnT {
+#if defined(__APPLE__)
+    constexpr const char *kLib = "libcudaq-analysis.dylib";
+#else
+    constexpr const char *kLib = "libcudaq-analysis.so";
+#endif
+    constexpr const char *kSym = "cudaq_runDemFromKernel";
```
1tnguyen: I think we can make the code a bit shorter using the existing `cudaq::getUniquePluginInstance` util. 
For example, returning the function as the unique plugin instance:

```
using DemFromKernelFn = std::string(const std::string &kernelName, ....);

static DemFromKernelFn *func = cudaq::getUniquePluginInstance<DemFromKernelFn>(
      "cudaq_getDemFromKernelFunc",
#if defined(__APPLE__)
      "libcudaq-analysis.dylib"
#else
      "libcudaq-analysis.so"
#endif
  );

  func(...);
```
with

```
extern "C" DemFromKernelFn *cudaq_getDemFromKernelFunc() {
  return &<actual_function_name>;
}
```

This might also eliminate the `return-type-c-linkage` warning.

runtime/cudaq/analysis/Dem.cpp
khalatepradnya: This is only when the input circuit also contains "REPEAT" blocks, right? We feed into the analyzer fully unrolled circuit for now. This can be a future optimization.

runtime/cudaq/analysis/Dem.cpp
khalatepradnya: I think this would need a change to the current NVQIR plugin contract (`getCircuitSimulator_<name>`) which returns a process-wide singleton today. So, unless I'm mistaken, making it per-thread requires widening that contract, which is broader work, outside the scope of this PR.

runtime/cudaq/analysis/Dem.cpp
khalatepradnya: Fixed in 0de0302.

runtime/nvqir/dem/DemScope.cpp
khalatepradnya: Fixed in 0de0302.

runtime/nvqir/dem/DemScope.cpp
khalatepradnya: Kept both intentionally:
- `on_enter` defensively clears state if a prior call ended abnormally.
- `on_exit` runs on every exit path including exceptions; it frees the per-execution tableau / sample-simulator state via `deallocateState()` so subsequent direct uses of the plugin simulator do not inherit DEM residue.

After the `release()` change that you suggested above, the circuit-clear portion of `on_exit`'s `reset()` is a cheap no-op on an empty buffer, so the cost is in the noise. The state-deallocation portion is load-bearing for exception
safety, hence the retention.

runtime/cudaq/algorithms/dem.h
khalatepradnya: Fixed in 0de0302.

runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,49 @@
...
+  ctx.kernelName = kernelName;
+  ctx.qpuId = cudaq::getCurrentQpuId();
+  ctx.asyncExec = false;
+  if (noise)
+    ctx.noiseModel = noise;
+
+  // RAII: claim the thread-local analysis-simulator slot backed by the `stim`
+  // plugin. The scope starts from a clean simulator and releases the override
+  // on every exit path.
+  auto demScope = nvqir::dem::make_scope(plugin_name);
```
Renaud-K, line 42: This is a lot simpler and the way to go. Thank you.

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
```diff
@@ -112,8 +112,12 @@ class Compiler {
 
   /// @brief Emit target-specific code for each `MlirArtifact` in the
   /// `CompiledModule` and produce `KernelExecution` objects.
+  ///
+  /// The execution context determines whether target transport code is needed
+  /// or whether a local analysis JIT artifact is sufficient.
   std::vector<cudaq::KernelExecution>
-  emitKernelExecutions(const cudaq::CompiledModule &compiled);
+  emitKernelExecutions(const cudaq::CompiledModule &compiled,
```
Renaud-K: You are being caught in the midst of Luca's changes #4608 and probably will need to wait for this PR to land. Luca is introducing an intermediate `CompileTarget` data structure to direct the compilation. So instead  `executionContext` -> `pipeline` . We will have `executionContext` -> `CompileTarget` -> `pipeline`. And eventually `policy` -> `CompileTarget` -> `pipeline`. I have talked to Luca about your needs. He is happy to talk to you and help you through the changes.

runtime/nvqir/NVQIR.cpp
```diff
@@ -713,6 +716,11 @@ bool __quantum__qis__read_result__body(Result *result) {
   auto iter = measRes2Val.find(result);
   if (iter != measRes2Val.end())
     return iter->second;
+  auto handle =
+      static_cast<std::int64_t>(reinterpret_cast<std::intptr_t>(result));
+  auto handleIter = measHandle2Val.find(handle);
```
sacpis: Would you please check if the `measHandle2Val` is getting triggered? As we already find the result using `measRes2Val`, the subsequent lookup can never fire. In `__quantum__qis__mz_handle__to__register`, we are storing the same handle and r (reinterpreted as Result*). `measRes2Val.find(result)` call will have already returned the key. `measHandle2Val` map, lookup and clear() appears to be dead code.

runtime/internal/compiler/JIT.cpp
```diff
@@ -124,8 +125,36 @@ static void insertSetupAndCleanupOperations(Operation *module) {
     mlir::LLVM::CallOp::create(builder, loc, mlir::TypeRange{voidTy},
                                setDynamicSymbol,
                                mlir::ValueRange{origMode.getResult()});
-    mlir::LLVM::CallOp::create(builder, loc, mlir::TypeRange{voidTy},
-                               clearResultMapsSymbol, mlir::ValueRange{});
+  }
+}
+
+static void
+insertResultMapCleanupOperations(Operation *module,
```
sacpis, line 132: A quick check. Can we confirm if this works for nested kernels? Also, can we check if a function has multiple return in different blocks, are those covered?

runtime/nvqir/NVQIR.cpp
```diff
@@ -48,6 +48,9 @@ inline static constexpr std::string_view GetCircuitSimulatorSymbol =
 static thread_local std::map<Qubit *, Result *> measQB2Res;
 static thread_local std::map<Result *, Qubit *> measRes2QB;
 static thread_local std::map<Result *, Result> measRes2Val;
+static thread_local std::map<std::int64_t, Result> measHandle2Val;
+static thread_local std::map<std::int64_t, std::int64_t> measHandle2Index;
```
sacpis: `measHandle2Index` is keyed by nextMeasureHandle (a monotonic counter from 0). Map has a lookup of `log(n)`. We can switch to `std::vector<std::int64_t>` indexed by handle. This will give us a constant lookup.

sacpis (approved):

runtime/internal/compiler/Compiler.cpp
```diff
@@ -453,7 +461,8 @@ cudaq::CompiledModule cudaq_internal::compiler::Compiler::runPassPipeline(
     modules.emplace_back(kernelName, moduleOp);
   }
 
-  bool needJit = emulate || target->generateResourceCounts;
+  bool needJit =
+      emulate || target->generateResourceCounts || target->generateJitArtifacts;
```
lmondada: By OR-ing flags here we might override a `target->generateJitArtifacts = false`, which would be a bit confusing.

I'd suggest removing the `emulate || target->generateResourceCounts` part (i.e. just leave `generateJitArtifacts` and then modify the CompileTarget constructor to set `generateJitArtifacts` to true whenever `emulate` or `generateResouceCounts` is set.

cudaq/include/cudaq/Target/CompileTarget.h
```diff
@@ -71,6 +71,27 @@ class CompileTarget {
...
+  /// Analysis contexts that execute locally, but are entered through a remote
+  /// platform, use this to run the kernel under the analysis simulator instead
+  /// of submitting it to the remote executor.
+  bool generateJitArtifacts = false;
+
+  /// Whether to translate MLIR artifacts into target transport code.
+  ///
+  /// Local analysis contexts can set this to false when they only need the JIT
+  /// artifact and do not need a QIR/QASM payload for the remote backend.
+  bool emitTargetCode = true;
```
lmondada: We could unify the naming conventions for these three flags: `emitResourceCounts`, `emitJit` and `emitTargetCode`

runtime/internal/compiler/JIT.cpp
```diff
@@ -211,8 +240,10 @@ cudaq_internal::compiler::createJITEngine(ModuleOp &moduleOp,
 
     // Insert necessary calls to qubit allocations and qubit releases if the
     // original module contained WireSetOp's.
+    const auto entryPointFuncs = collectEntryPointFunctions(module);
     if (containsWireSet)
-      insertSetupAndCleanupOperations(module);
+      insertWireSetSetupAndCleanupOperations(module, entryPointFuncs);
+    insertResultMapCleanupOperations(module, entryPointFuncs);
```
1tnguyen, line 253: Q: This `insertResultMapCleanupOperations` is only used in this `JIT.cpp` file. Do we need to insert it for the AOT case as well (`nvq++` with local simulation)?

1tnguyen (approved):
LGTM 👍

runtime/nvqir/CircuitSimulator.h
```diff
@@ -170,6 +171,13 @@ class CircuitSimulator {
   /// https://arxiv.org/pdf/2407.13826.
   virtual void generateMSM() {}
 
+  /// @brief For simulators that support detector error model (DEM) analysis,
+  /// compute the DEM from the recorded circuit and return it as `.dem` text.
```
lmondada, line 175: Is the `.dem` text you mention a format that is defined by Stim? Or is it some standard? It might be good to add a note this is currently only implemented by the Stim backend

lmondada (approved):
Thank you @khalatepradnya for making these changes. This now aligns perfectly with the recent (and upcoming) changes to the Compiler.

Renaud-K (approved):
Looks good.

runtime/nvqir/NVQIR.cpp
khalatepradnya: Good catch! Removed.

runtime/nvqir/NVQIR.cpp
khalatepradnya: Yes, updated in [07ead31](https://github.com/NVIDIA/cuda-quantum/pull/4602/commits/07ead31315a93b8f9d9e120e564a707176e4f2b5)

runtime/internal/compiler/JIT.cpp
khalatepradnya: Confirmed both, and hardened the multi-return case in `JIT.cpp`

runtime/internal/compiler/JIT.cpp
khalatepradnya: The result-map clear has only ever been on the JIT path, I only split up the existing function.

cudaq/include/cudaq/Target/CompileTarget.h
khalatepradnya: Sure, done in 07ead31

runtime/nvqir/CircuitSimulator.h
khalatepradnya: Added clarifying comment here as well.

runtime/internal/compiler/Compiler.cpp
khalatepradnya: Can we fold this in https://github.com/NVIDIA/cuda-quantum/pull/4636? Deferring.

runtime/cudaq/analysis/Dem.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  platform.with_execution_context(ctx, kernel);
+
+  return std::move(ctx.dem_text);
```
schweitzpgi, line 46: Generally, you shouldn't use `std::move` in return statements as it can force the compiler to generate strictly worse performing code by disabling RVO.

khalatepradnya: This would be applicable when move the result (`dem_text`) outside of the `ExecutionContext`.

PR #4603, opened 2026-05-22
Give `std::vector<measure_handle>` locals a stable address (fix #4601, part 1)

### LINKED ISSUE #4601: Support `std::vector<measure_handle>::operator=` in kernels
### Summary

Enable the QEC vectorized cross-round detector patterns shown below. They are the vectorized counterparts to the scalar `prev_s0 = s0;` idiom from the internal spec, and they do not compile today.

#### C++ — vectorized cross-round, local-only (loop)
```cpp
#include <cudaq.h>
#include <vector>
__qpu__ void rep_code_vectorized(int nRounds) {
  cudaq::qvector ancillas(2);
  auto prev = mz(ancillas);
  for (int r = 1; r < nRounds; r++) {
    auto curr = mz(ancillas);
    cudaq::detectors(prev, curr);
    prev = curr;                    // carry current round forward
  }
}
```

#### C++ — vectorized cross-round with reference parameter 
```cpp
#include <cudaq.h>
#include <vector>
__qpu__ void surface_code_round(cudaq::qvector<>& data,
                                cudaq::qvector<>& ancillas,
                                std::vector<cudaq::measure_result>& prev) {
  auto syndromes = mz(ancillas);
  cudaq::detectors(prev, syndromes);
  prev = syndromes;                 // update caller's `prev` for next round
}
```

### Current Behavior

Compilation fails
```
$ cudaq-quake rep_code_vectorized.cpp
error: ../cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp:1595: not yet implemented: unhandled std::vector member function, operator=
LLVM ERROR: fatal error, aborting.
```

### Expected Behavior

Program compiles and runs successfully.

### Prior work

https://github.com/NVIDIA/cuda-quantum/pull/4573 attempted to fix by using shallow copy semantics, but, was generating wrong IR.

anpaz:
## Problem

`std::vector<measure_handle>` locals dangle. `ConvertDecl::VisitVarDecl`'s `cc::StdvecType` arm registered the **descriptor value** returned by the initializer directly in the symbol table — there was no stack backing. Any subsequent lvalue use (reassignment, range-for, passing to `qec.detectors`/`qec.observable`/`qec.pair_detectors`, etc.) then operated on a stale SSA value whose underlying buffer pointer was no longer valid, producing wrong IR or crashes downstream. Scalar `measure_handle` locals were already memory-domain (alloca + load/store), so the two shapes were inconsistent.

The previous attempt at a fix (#4573) was closed because it relied on shallow-copy semantics that broke aliasing.

## Change

In `ConvertDecl::VisitVarDecl`, when the local's element type is `measure_handle`:

1. `cc.alloca !cc.stdvec<!cc.measure_handle>` a descriptor slot.
2. Normalize the initializer to value form (`cc.load` if it arrived as a pointer-form lvalue from another handle-vec read).
3. `cc.store` the descriptor into the slot.
4. Insert the **slot pointer** into the symbol table.

Other `std::vector<T>` element types (`bool`, numeric) keep the existing value-domain symbol-table entry — no behavior change for them.

Consumers must materialize the value form before consuming the descriptor. The existing `loadHandleVectorIfPointer` helper in `ConvertExpr.cpp` already does this at the 5 sites that read a handle-vec (`to_integer`, bulk-discriminate, `qec.detector(s)`, `qec.observable`, `qec.pair_detectors`). This PR:

- Promotes `loadHandleIfPointer` and `loadHandleVectorIfPointer` from the `ConvertExpr.cpp` anonymous namespace to `inline` helpers in `cudaq::details` in `ASTBridge.h` so they can be shared.
- Adds one new call site in `ConvertStmt::TraverseCXXForRangeStmt` (range-for over a handle-vec) using the same helper.

## Why this doesn't break shallow-copy semantics

The slot owns the descriptor of the local, not the underlying handle buffer. Two locals initialized from the same RHS still share the same handle buffer pointer (same C++ semantics as before). Deep-copy semantics for `m = m_new;` are out of scope here and will land as a follow-up that intercepts `operator=` on `std::vector<measure_handle>`.

## Test updates

Updated CHECK lines for the new alloca/store/load shape in:
- vector_bool.cpp (`t1`)
- measure_bell.cpp (3 functions)
- qec_ops.cpp (`DetectorVector`, `DetectorMixed`, `ObservableVectorDefault`, `ObservableVectorIndexed`, `ObservableMixed`, `PairDetectors`, `RepCodeD3`)

`ninja check-nvqpp`: 361 passed, 2 expectedly failed (pre-existing XFAILs).

## Follow-ups

- **PR B**: intercept `operator=` on `std::vector<measure_handle>` to deep-copy the descriptor's buffer, addressing the aliasing concern raised on #4573.


1tnguyen (approved):
LGTM 👍

khalatepradnya (approved):
Thanks, @anpaz !

PR #4604, opened 2026-05-24
Support `std::vector<measure_result>&` ref-parameter assignment in kernels (#4601, part 2)

### LINKED ISSUE #4601: Support `std::vector<measure_handle>::operator=` in kernels

anpaz:
**Summary**

Fixes #4601. Builds on the previous commit (`prev = curr;` lvalue deep-copy) to also support the surface-code-style pattern where the measurement-vector being assigned into is a *reference parameter*, e.g.

```cpp
__qpu__ void surface_code_round(cudaq::qview<> qv,
                                std::vector<cudaq::measure_result>& prev) {
  auto curr = mz(qv);
  // ...
  prev = curr;   // must write through to the caller's vector
}
```

Before this change, the assignment compiled but wrote into a local descriptor copy and was invisible to the caller, because `std::vector<T>&` parameters were lowered to a bare `!cc.stdvec<...>` descriptor passed by value (the standing convention for all `SpanLikeType` references). Element-level mutation (`v[i] = ...`) happened to work because the descriptor's `data*` still aliased the caller's heap buffer, but whole-descriptor replacement could not.

**Change**

In `VisitLValueReferenceType`, carve out exactly one case from the existing `SpanLikeType` fall-through: a *non-const* `std::vector<cudaq::measure_result>&` parameter is now lowered to `!cc.ptr<!cc.stdvec<!cc.measure_handle>>`. The existing `operator=` intercept then emits `cc.store newDescr, thisPtr`, which lands in the caller's slot.

Every other reference type — const `std::vector<measure_result>&`, all `std::vector<T>&` with non-measure_handle elements, `pauli_word&`, etc. — keeps the prior bare-descriptor ABI. The carve-out is intentionally as narrow as the set of assignments the bridge currently knows how to lower; widening it would require either an extended `operator=` intercept or downstream updates to const-ref consumers (and is left as future work).

**Files**

- ConvertType.cpp — the targeted ABI change (one nested `if`).
- measure_handle.cpp — adds a `__qpu__` callee + caller exercising the new pointer-form parameter; AST CHECKs pin the new IR.
- separate_compilation.cpp — updates an existing CHECK to expect the new pointer-form call site for `otherKernel(std::vector<measure_result>&)`.
- measure_handle_vec_assign.cpp — adds `ref_param_last_round` (writes through) and `assign_then_mutate_source` (future-proofing coverage for the assign-then-modify-source pattern; not a strict deep-vs-shallow witness — see in-file comment).

**Tests**

- `ninja check-nvqpp` — baseline preserved (361 pass + 2 XFAIL).
- measure_handle_vec_assign.cpp lit run prints `outer:01`, `carry:10`, `refp:10`, `snap:10` as expected.


cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1471,6 +1447,66 @@ bool QuakeBridgeVisitor::VisitCallExpr(clang::CallExpr *x) {
...
+        auto i8PtrTy = cc::PointerType::get(builder.getI8Type());
+        Value srcData =
+            cc::StdvecDataOp::create(builder, loc, i8PtrTy, rhsDescr);
+        Value size = cc::StdvecSizeOp::create(builder, loc,
+                                              builder.getI64Type(), rhsDescr);
+        IRBuilder irb(builder);
+        Value eltSize =
+            irb.getByteSizeOfType(loc, lhsStdvecTy.getElementType());
+        Value heap = func::CallOp::create(builder, loc, i8PtrTy,
+                                          "__nvqpp_vectorCopyCtor",
```
1tnguyen, line 1499: I think it might be worth checking the QIR lowering of this in the use cases of interest (e.g., QEC). This intrinsic has `malloc`/`memcpy` which might or might not be optimized away by the final LLVM optimization.

1tnguyen (approved):
LGTM 👍

cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp
anpaz: I checked the QIR lowering for the QEC carry pattern (prev = curr inside a per-round loop). The intrinsic is inlined to a malloc(N*8) + memcpy(N*8) per iteration, and LLVM -O3 does not eliminate it.

schweitzpgi (approved):

PR #4606, opened 2026-05-25
Updated QASM gate support for benchmarking

taalexander:
This is based on #4556 and should only be merged after.

This PR improves OpenQASM interoperability through the Qiskit conversion path by handling gate forms that commonly appear in imported QASM circuits. It adds direct coverage for controlled phase-style gates and expands Qiskit open-control definitions so CUDA-Q preserves the intended control-state behavior. This has quite a few LOC change but really it's a refactor of the existing inline logic to something a bit more scalable (although not perfect and probably long run will need additional refactor).


atgeller (approved):
LGTM!

PR #4607, opened 2026-05-26
[Python] Cache JIT compiled module for Python kernels

### LINKED ISSUE #4507: Performance Regression for Kernel Launching 13.0->14.0
### Required prerequisites

### Describe the bug

There seems to be a slowdown in kernel launches between CUDA-Q v13.0 and 14.0(2).   

The code below, produced the following timings.

Image

```
import time
import cudaq
from cudaq import spin

cudaq.set_target("nvidia", option="fp64")

@cudaq.kernel
def trivial(n: int):
    q = cudaq.qvector(n)
    for i in range(n):
        h(q[i])

H = spin.z(0)  # the simplest possible observable

# ---- Warm-up: trigger JIT for both code paths ----
for _ in range(5):
    cudaq.get_state(trivial, 4)
    cudaq.observe(trivial, H, 4).expectation()

N = 1000

# ---- Measure get_state ----
t0 = time.time()
for _ in range(N):
    cudaq.get_state(trivial, 4)
t_state_total = time.time() - t0
t_state_per   = t_state_total / N

# ---- Measure observe ----
t0 = time.time()
for _ in range(N):
    cudaq.observe(trivial, H, 4).expectation()
t_obs_total = time.time() - t0
t_obs_per   = t_obs_total / N

print(f"cudaq:    {cudaq.__version__.split()[2]}")
print(f"target:   nvidia fp64")
print(f"kernel:   trivial 4-qubit Hadamard")
print(f"calls:    {N} per op")
print()
print(f"  get_state  total: {t_state_total:7.2f} s   per call: {t_state_per*1000:7.3f} ms")
print(f"  observe    total: {t_obs_total:7.2f} s   per call: {t_obs_per*1000:7.3f} ms")

```

### Expected behavior

Faster times

13.0

### Environment

- **CUDA-Q version**:  13.0 compared to 14.2 (and 14.0)
- **Python version**:  3.11
- **Operating system**:  WSL

### Suggestions

_No response_

atgeller:
Attaches a slot to Python kernel decorators/builders to store a saved CompiledModule (when run on a local simulator), which can be used to skip re-compilation in the future. Fixes #4507.

Does not include logic to make sure a `CompiledModule` was compiled for a specific target/launch, as this distinction is currently irrelevant for local simulation, where the same passes will be run no matter what.

I've deleted the `reuse_compiler_artifacts` environment and related code in `ExecutionContext.cpp`/`.h` as it is no longer needed.


python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -1179,6 +1223,12 @@ void cudaq::bindAltLaunchKernel(nanobind::module_ &mod,
                    &cudaq::CompiledModule::isFullySpecialized,
                    "Whether all arguments have been specialized.");
 
+  nanobind::class_<cudaq::CompiledModulePtr>(mod, "CompiledModulePtr")
+      .def(nanobind::init<>())
+      .def("empty",
+           [](const cudaq::CompiledModulePtr &p) { return p == nullptr; })
+      .def("clear", [](cudaq::CompiledModulePtr &p) { p.reset(); });
```
sacpis: Do we need these? Seems like we are not using it.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -674,10 +674,51 @@ static void appendTheResultValue(ModuleOp module, const std::string &name,
...
+    return cudaq::streamlinedLaunchModule(**compiled, rawArgs);
+
   auto clone = mod.clone();
-  auto compiled = cudaq::streamlinedCompileModule(name, clone, rawArgs, true);
-  auto res = cudaq::streamlinedLaunchModule(compiled, rawArgs);
+  auto jitted = cudaq::streamlinedCompileModule(name, clone, rawArgs, true);
+  auto res = cudaq::streamlinedLaunchModule(jitted, rawArgs);
+  *compiled = std::make_shared<cudaq::CompiledModule>(std::move(jitted));
   clone.erase();
   return res;
```
sacpis, line 713: This code looks similar to the code from 703-707. Should we put it in a lambda?

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -674,10 +674,51 @@ static void appendTheResultValue(ModuleOp module, const std::string &name,
...
+    mlir::Type resultTy = cudaq::runtime::getReturnType(func);
+    auto hasResult = !!resultTy;
+    size_t num_args = rawArgs.size() - (hasResult ? 1 : 0);
+    // Check for args synthesized.
+    if (num_args != func.getArgumentTypes().size())
+      return false;
+    return true;
+  }();
+
+  if (!is_cachable) {
```
sacpis: nit: Should we be using camelCase here for `is_cachable` and `num_args`?

python/runtime/cudaq/platform/py_alt_launch_kernel.h
```diff
@@ -18,11 +19,17 @@
 #include <nanobind/stl/function.h>
 #include <nanobind/stl/string.h>
 #include <nanobind/stl/vector.h>
+#include <optional>
```
sacpis, line 22: nit: Don't think we are using it.

runtime/common/CompiledModule.h
```diff
@@ -186,6 +186,7 @@ class FatQuakeModule {
   struct CompilationMetadata {
     /// Qubit reorder indices emitted by the qubit-mapping pass.
     std::vector<std::size_t> reorderIdx;
+    // TODO: Add hash of initial module to serve as cache key.
```
sacpis: Will this PR be followed by another to cover this TODO?

python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
```diff
@@ -41,7 +41,8 @@ using namespace cudaq;
 // std::optional types for all nullable parameters instead.
 static ptsbe::sample_result
 pySamplePTSBE(const std::string &shortName, MlirModule module,
-              std::size_t shots_count, noise_model noiseModel,
+              nanobind::handle compiledHandle, std::size_t shots_count,
```
sacpis: Should we just accept `CompiledModulePtr` like other entry points and let nanobind convert?

sacpis (approved):
Overall LGTM. Thanks @atgeller. Left a few comments.

python/cudaq/kernel/kernel_builder.py
```diff
@@ -195,6 +196,7 @@ def __generalCustomOperation(self, opName, *args):
     qubits = []
     if hasattr(self, 'qkeModule'):
         del self.qkeModule
+    self._compiled_module = cudaq_runtime.CompiledModulePtr()
```
lmondada: nit: thinking about it, I find it slightly strange that there is an opaque `_compiled_module` variable set even before anything was compiled (and before `qkeModule`) was set.

It might be worth copying the same pattern used for qkeModule, i.e. if it's set it's not empty, otherwise the variable does not exist. You would just need to add a line

```python
self._compiled_module = cudaq_runtime.CompiledModulePtr()
```
before calling `marshal_and_launch_module`. Then `_clear_cache` deletes the variable, in nice symmetry to how it handles `qkeModule`.

python/cudaq/kernel/kernel_builder.py
```diff
@@ -58,6 +58,7 @@ def __generalOperation(self,
     opCtor = getattr(quake, '{}Op'.format(opName.title()))
     if hasattr(self, 'qkeModule'):
         del self.qkeModule
+    self._compiled_module = cudaq_runtime.CompiledModulePtr()
```
lmondada: You could put these three lines in a `_clear_cache` method and reuse that across the file

python/cudaq/kernel/kernel_decorator.py
```diff
@@ -588,7 +591,7 @@ def prepare_call(self, *args, allow_no_args=False):
         with trace.span("kernel.clone_module"):
             module = cudaq_runtime.cloneModule(self.qkeModule)
 
-        return processed_args, module
+        return processed_args, module, self._compiled_module
```
lmondada: You don't gain much from returning this here and have to make a bunch of call sites more noisy. The call sites could just get the `_compiled_module` variable directly. Or, if you follow my suggestion above, you could expose a method `get_or_create_compiled_module()` and use that where you need it.

python/runtime/cudaq/platform/py_alt_launch_kernel.h
```diff
@@ -18,11 +19,17 @@
...
 #include <string>
 #include <vector>
 
 namespace cudaq {
 
+/// A Python-owned mutable handle to a CompiledModule. Python passes one of
+/// these through the launch entry points so that C++ can install (cache) a
+/// JIT-compiled module and reuse it on subsequent calls.
+using CompiledModulePtr = std::shared_ptr<CompiledModule>;
```
lmondada: I've thought about it and I don't see why you couldn't just define

```
using PyCompiledModule = std::optional<CompiledModule>
```

Then in the binding code you can avoid the double pointer situation. Am I missing something?

In fact, this makes me think that we should make it possible to default construct a `CompiledModule`, then you wouldn't even need the `optional` wrapper.

python/tests/kernel/test_caching.py
```diff
@@ -0,0 +1,260 @@
+# ============================================================================ #
```
lmondada, line 1: Nice set of tests! Do we have any way of verifying that caching really happens? I guess we could check that `_compiled_module` got filled with something... Really a shame that we don't have a good way to 'deactivate' the JIT compiler. But this should get easier with the changes we're making, so let's just be patient :P

runtime/common/CompiledModule.h
```diff
@@ -245,6 +246,7 @@ class FatQuakeModule {
   const std::string &getName() const { return name; }
   const ResultInfo &getResultInfo() const { return resultInfo; }
   const CompilationMetadata &getMetadata() const { return metadata; }
+  CompilationMetadata &getMutableMetadata() { return metadata; }
```
lmondada: I think C++ would allow an overload of the same name `getMetadata` (at least I think that's what the STL often does?). I don't mind being explicit though.

lmondada: btw, are you using this?

python/runtime/cudaq/platform/py_alt_launch_kernel.h
```diff
@@ -49,12 +56,14 @@ void bindAltLaunchKernel(nanobind::module_ &mod,
 /// type is obtained from the kernel's FuncOp. \p module must be modifiable.
 nanobind::object marshal_and_launch_module(const std::string &kernelName,
                                            MlirModule module,
+                                           CompiledModulePtr *compiled,
```
lmondada: This is silly, but if you moved `*compiled` to last position in the signature and added a `= nullptr` default, then you wouldn't have to udpate as many call sites.

lmondada (approved):
This is such a nice simplification! Makes me happy, thanks!!

python/runtime/cudaq/platform/py_alt_launch_kernel.h
lmondada: Actually, having thought about it, I'd just add a public `CompiledModule` default construct and an `empty()` method, then just expose `CompiledModule` as-is to Python. IMO keeps things simpler.

Your bound functions in C++ can then just take a `CompiledModule&` as argument, or if you need to handle the case where no caching should happen, then you'd need an `std::optional<CompiledModule>&` or a pointer.

runtime/common/CompiledModule.h
atgeller: Oops, I added that along with some extra metadata fields to be able to set them, but it turned out I didn't need them. Looks like I forgot to clean up this function as well.

runtime/common/CompiledModule.h
atgeller: Yes, although the TODO is slightly wrong: the hashed cache key should be of the CompileTarget Luca is working on adding, not the module.

python/tests/kernel/test_caching.py
atgeller: Good point. I was relying on local time observations, but it would be good to make sure we don't silently break something. I'll add a check that the _compiled_module got filled, and maybe add logging.

python/runtime/cudaq/platform/py_alt_launch_kernel.h
atgeller: Yeah a default `CompiledModule` would be nice and simplify some of this. Unfortunately, you can't pass by reference across the nanobind interface (at least, I think so, it wasn't working when I was trying it), so it would have to be a pointer.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: Looks like we may be able to get rid of `CompiledModulePtr` as whole anyway

python/runtime/cudaq/algorithms/py_sample_ptsbe.cpp
atgeller: This is because PTSBE include nanobind std::shared_ptr which unwraps the `CompiledModulePtr` definiton and caused all sorts of problems. I think with the changes removing `CompiledModulePtr` this problem should go away so we can just use a `CompiledModule*`

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: I refactored the logic a bit to be a little cleaner.

PR #4608, opened 2026-05-26
Parametrise compiler based on CompileTarget options 

lmondada:
~This needs to be merged after #4599. Until I rebase, you should ignore the first commit when reviewing.~ Done.

In this PR I introduce compiler options within the `CompileTarget` type. This allows us to finally remove any dependency of the compilation on the execution context!

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -269,12 +269,19 @@ public:
   };
 
   using QPU::getCompileTarget;
-  std::unique_ptr<CompileTarget> getCompileTarget(ExecutionContext *) override {
+  std::unique_ptr<CompileTarget>
+  getCompileTarget(ExecutionContext *ctx) override {
     std::filesystem::path cudaqLibPath{cudaq::getCUDAQLibraryPath()};
     auto platformPath = cudaqLibPath.parent_path().parent_path() / "targets";
 
-    return std::make_unique<BaseRemoteRESTQPUCompileTarget>(
+    auto target = std::make_unique<BaseRemoteRESTQPUCompileTarget>(
         serverHelper.get(), platformPath, targetConfig, backendConfig, emulate);
+    if (ctx && ctx->name == "observe") {
+      if (!ctx->spin.has_value())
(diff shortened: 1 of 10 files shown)


lmondada:
/ok to test [98a8e05](https://github.com/NVIDIA/cuda-quantum/pull/4608/commits/98a8e058ca6078bf23ea439ac9b0e80aac7bb2a7)

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -252,12 +263,35 @@ class BaseRemoteRESTQPU : public QPU {
...
+  std::unique_ptr<CompileTarget>
+  getCompileTarget(ExecutionContext *ctx) override {
+    std::filesystem::path cudaqLibPath{cudaq::getCUDAQLibraryPath()};
+    auto platformPath = cudaqLibPath.parent_path().parent_path() / "targets";
+
+    auto target = std::make_unique<BaseRemoteRESTQPUCompileTarget>(
+        serverHelper.get(), platformPath, targetConfig, backendConfig, emulate);
+    if (ctx && ctx->name == "observe") {
+      if (!ctx->spin.has_value())
+        throw std::runtime_error("observe execution requires a spin_op");
```
Destroyerrrocket, line 271: Maybe there's no better place, but isn't it a bit weird that we're checking for the semantics of ctx when we create the compile target, no?

1tnguyen (approved):
LGTM 👍

runtime/common/BaseRemoteRESTQPU.h
lmondada: Yes, 100% agreed! We are trying to remove dependencies on the ExecutionContext as much as possible. This PR is a step in that direction, as it removes any look ups on `ctx` within the compilation _with the exception of getCompileTarget_. There is no way around that at the moment, but that's precisely what @Renaud-K is working on.

He is introducing launch policies that will allow us to define overloads of `getCompileTarget` based on policy rather than having to look into the execution context. There is already one example of such an overload for the `sample_policy`. Others will follow soon.

Renaud-K (approved):
LGTM.

PR #4610, opened 2026-05-26
[Python] Fix handling of empty veqs.

### LINKED ISSUE #4609: `quake.subveq` slicing: Empty `subveq` slices (`[0:0]`) cause either crashes (static) or underflow evaluation (dynamic)
### Required prerequisites

### Describe the bug

Applying gates to an empty slice should result in no-op. The actual behaviour is:

- if the slice boundary is **static**: kernel crashes (related to: https://github.com/NVIDIA/cuda-quantum/issues/2625 and  https://github.com/annagrin/cuda-quantum/commit/32745a65c3bb5d907709ef4ab951d67b1afbaaf5) 
- if the slice boundary is **dynamic**: Gates are applied to all qubits in the list.

**Static boundary**
```
import cudaq

@cudaq.kernel
def static_empty_slice() -> int:
    q = cudaq.qvector(5)
    index = 0
    
    # Evaluates to q[0:0]. Expected: no-op. Actual: Crash
    x(q[0:index])

    return int(mz(q[0])) + int(mz(q[1])) + int(mz(q[2])) + int(mz(q[3])) + int(mz(q[4])) 

#cudaq.run(static_empty_slice, shots_count=10)

print(static_empty_slice)
```

> %1 = cc.poison !quake.veq<?>

is inserted and running the kernel results in a crash

**Dynamic boundary**
```
import cudaq

@cudaq.kernel
def dynamic_empty_slice() -> int:
    q = cudaq.qvector(5)
    
    # mz(q[4]) is deterministically 0
    index = int(mz(q[4]))
    
    # Evaluates to q[0:0] logically, but lowers dynamically.
    x(q[0:index])

    return int(mz(q[0])) + int(mz(q[1])) + int(mz(q[2])) + int(mz(q[3])) + int(mz(q[4])) 

result = cudaq.run(dynamic_empty_slice, shots_count=10)
print(result)
```
> [5, 5, 5, 5, 5, 5, 5, 5, 5, 5]

I.e., the `x` gate is applied to all qubits.

> print(dynamic_empty_slice)
>    %2 = quake.discriminate %measOut : (!quake.measure) -> i1
    %3 = cc.cast unsigned %2 : (i1) -> i64
    %4 = arith.subi %3, %c1_i64 : i64
    %5 = quake.subveq %0, 0, %4 : (!quake.veq<5>, i64) -> !quake.veq<?>

This results in upper boundary `-1` which appears to be interpreted as `npos` (rest of array).

### Expected behavior

Both static and dynamic empty slices (e.g., q[0:0]) should safely resolve as no-ops, matching standard Python array/list semantics.

Not a regression

### Environment

- **CUDA-Q version**: 0.14.2
- **Python version**: 3.11.15
- **Operating system**: MacOs 26.4.1

### Suggestions

_No response_

schweitzpgi:
In Python, we want to allow an escape hatch such that the quantum operations called with 0 qubits will become NOPs instead of get flagged as errors. This change was approved/recommended by architecture.

Fix #4609.

### DIFF
--- a/cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
+++ b/cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
@@ -117,8 +117,11 @@ struct ForwardEmptyVeqSizePattern
     if (veqTy.hasSpecifiedSize())
       return failure();
     auto undef = veqSize.getVeq().getDefiningOp<cudaq::cc::UndefOp>();
-    if (!undef)
-      return failure();
+    if (!undef) {
+      auto poison = veqSize.getVeq().getDefiningOp<cudaq::cc::PoisonOp>();
+      if (!poison)
+        return failure();
+    }
     rewriter.replaceOpWithNewOp<arith::ConstantIntOp>(veqSize,
                                                       veqSize.getType(), 0);
     return success();
--- a/cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
+++ b/cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
@@ -102,7 +102,7 @@ struct ForwardConstantVeqSizePattern
 };
 
 // %3 = cc.undef !quake.veq<?>
-// %4 = quake.veq_size %3 : (!quake.veq<?>) -> 164
+// %4 = quake.veq_size %3 : (!quake.veq<?>) -> i64
 // ────────────────────────────────────────────────
 // %4 = constant 0 : i64
 struct ForwardEmptyVeqSizePattern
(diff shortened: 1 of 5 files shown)


cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -101,6 +101,30 @@ struct ForwardConstantVeqSizePattern
   }
 };
 
+// %3 = cc.undef !quake.veq<?>
+// %4 = quake.veq_size %3 : (!quake.veq<?>) -> 164
```
1tnguyen: nit: 
```suggestion
// %4 = quake.veq_size %3 : (!quake.veq<?>) -> i64
```

1tnguyen (approved):
LGTM 👍 

Perhaps we can also add a Python regression test similar to the empty range slicing in 4609.

PR #4611, opened 2026-05-27
[Analysis] Create simple analysis pass to group unitary operations

cabreraam:
# Overview

This PR aims to create a simple analysis pass to group unitary operations. This is an intermediary step for more complex groupings (e.g., coalescing groups) and downstream passes (consuming the grouping to inform lowering to hardware). The analysis pass creates a vector of `UnitaryOpGroup` structs that, where each struct contains, in part, a vector of all the unitary ops grouped in a given pass. 

## Implementation
For now, this pass groups unbroken sequences of unitary operations, i.e., ops that contain the  `QuantumGate` type but are not  `quake.reset` ops.

The entry point for the analysis are `func` ops. With the `func` op region as the starting point, we recursively search regions to find groups of unitary operations. To account for nested regions within ops (e.g., `cc.if`), we examine all ops in a given block, and then we check each op to see if there are nested regions that require examining.

## Adding infrastructure for Analysis Passes
Currently, there is no formal infrastructure for pure analysis passes in the code base; while there are analysis passes that exist in CUDA-Q, there are one off analyses that are tied to a transformation pass. We introduce a structure that tests analysis passes as unit tests, as shown for testing this simple unitary op grouping. Namely:
- Create corresponding unit test for the pass under `cudaq/unittests/Optimizer`
- Modifying the `cudaq/unittests/Optimizer/CMakeLists.txt` file to 
  - create an executable target `AnalysisPassUnitTests` for GTest driven unit tests, and include the above test as a source for the executable.
  - create a custom CMake target `check-analysis-passes` to drive unit tests from `ninja` or `make`


cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "mlir/Support/TypeID.h"
+
+#define DEBUG_TYPE "unitary-op-grouping-analysis"
+
+using namespace mlir;
+
+namespace {
+static bool isUnitaryOp(Operation *op) {
+  return op->hasTrait<cudaq::QuantumGate>() && !isa<cudaq::quake::ResetOp>(op);
+}
```
taalexander, line 22: A reset is not unitary.

taalexander (commented):
Hi @cabreraam, thanks for putting this together. I was wondering if you could frame this analysis in how it would eventually be used? I can imagine either:
1. Circuit outlining to a canonical IR form
2. Circuit lowering to a hardware form

In both cases I imagine the analysis may only be required once or twice. That being said if I were to use it for the above I would expect that I would need methods to interact with the analysis as I was writing the transformation pass. Something like:

- `const UnitaryOpGroup *getGroupContaining(Operation *op) const;`
- `ArrayRef<UnitaryOpGroup> getGroupsIn(Block *block) const;`
- `bool inSameGroup(Operation *a, Operation *b) const;`

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: Yes, this is my understanding as well. I believe the current logic verifies that an `op` is a unitary iff:
- it possesses the  `QuantumGate` trait
and
- it is _not_ a `ResetOp`

taalexander: Thanks @cabreraam for pointing out my obvious error (I missed the not :man_facepalming: ).

Follow up question. Why do we use the unitary check to delineate boundaries?

cabreraam:
Thanks for the feedback @taalexander! 

> I was wondering if you could frame this analysis in how it would eventually be used? I can imagine either:
> 
> 1. Circuit outlining to a canonical IR form
Yes, definitely! A transformation pass that aims to canonicalize a circuit by grouping unitary ops together could consume the output of this analysis pass to shuffle the IR around to eventually group sequences of unitaries within a mixed classical-quantum program.   

> 2. Circuit lowering to a hardware form
Also would be useful for this purpose. I believe that it would be useful in lowering a mixed classical-quantum program to mirror the groupings in [this QIR document](https://www.qir-alliance.org/qat/DeveloperGuide/LogicSeparation/#block-separation-strategy) where a program is neatly delineated by 
- classical setup
- quantum ops
- readout
- classical post-processing

> 
> In both cases I imagine the analysis may only be required once or twice. That being said if I were to use it for the above I would expect that I would need methods to interact with the analysis as I was writing the transformation pass. Something like:
> 
> * `const UnitaryOpGroup *getGroupContaining(Operation *op) const;`
> * `ArrayRef<UnitaryOpGroup> getGroupsIn(Block *block) const;`
> * `bool inSameGroup(Operation *a, Operation *b) const;`

Yes, good idea. I'll implement some of these ~~ASAP~~ in a follow-on PR. The next feature on the work queue is to coalesce groups if possible (e.g., no ops with side effects between two given groups in the same block), and these features would be great to implement toward that end.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: No problem!

It's a good follow-up question. The goal of the pass was to implement a simple way to find groups of unitaries, so this seemed to fit the bill. However, future iterations of this or maybe different analysis passes may do something more complex/clever.

cabreraam:
> Yes, good idea. I'll implement some of these ~ASAP~ in a follow-on PR. 

Actually, I just included them in this PR. It felt more modular to implement interface functions in this PR rather than string it across a PR which will include at least one more pass.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -51,8 +53,49 @@ static void flushGroupIfNonEmpty(cudaq::quake::detail::UnitaryOpGroups &groups,
...
+
+const cudaq::quake::detail::UnitaryOpGroup *
+cudaq::quake::detail::UnitaryOpGroupingAnalysis::getGroupContainingOp(
+    mlir::Operation *op) const {
+  if (!op || !isUnitaryOp(op))
+    return nullptr;
+
+  for (const UnitaryOpGroup &group : groups)
+    if (llvm::is_contained(group.ops, op))
+      return &group;
```
taalexander: I worry that this might be the most common operation requiring an O(n) lookup. It ultimately depends if we're lowering on a per-block or per-operation basis.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -51,8 +53,49 @@ static void flushGroupIfNonEmpty(cudaq::quake::detail::UnitaryOpGroups &groups,
...
+  return nullptr;
+}
+
+llvm::SmallVector<const cudaq::quake::detail::UnitaryOpGroup *>
+cudaq::quake::detail::UnitaryOpGroupingAnalysis::getGroupsIn(
+    const mlir::Block *block) const {
+  llvm::SmallVector<const UnitaryOpGroup *> groupsInBlock;
+  for (const UnitaryOpGroup &group : groups) {
+    if (group.block == block)
+      groupsInBlock.push_back(&group);
```
taalexander: This is also an O(n) lookup

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -51,8 +53,49 @@ static void flushGroupIfNonEmpty(cudaq::quake::detail::UnitaryOpGroups &groups,
...
+  return groupsInBlock;
+}
+
+bool cudaq::quake::detail::UnitaryOpGroupingAnalysis::inSameGroup(
+    mlir::Operation *op1, mlir::Operation *op2) const {
+  if (!op1 || !op2 || !isUnitaryOp(op1) || !isUnitaryOp(op2))
+    return false;
+
+  const UnitaryOpGroup *group1 = getGroupContainingOp(op1);
+  const UnitaryOpGroup *group2 = getGroupContainingOp(op2);
```
taalexander: Each of these requires an O(n) lookup.

cudaq/test/lib/Analysis/TestUnitaryOpGrouping.cpp
```diff
@@ -7,18 +7,144 @@
...
 namespace {
+using UnitaryOpGroup = cudaq::quake::detail::UnitaryOpGroup;
+using UnitaryOpGroupingAnalysis =
+    cudaq::quake::detail::UnitaryOpGroupingAnalysis;
+
+static void printBool(bool value) {
+  llvm::errs() << (value ? "true" : "false");
+}
+
+static void printOperationMembership(
```
taalexander: Side note even just for printing for test validation it might be simple to rewrite all of our groups as new functions (eg., the outlining). How much work would it be to test in this way? If large we can defer

taalexander (changes requested):

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: I now maintain a [map](https://github.com/cabreraam/cuda-quantum/blob/513b895b8f4eb373d293e4c405a7817db6341fd7/cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h#L56) for 
`op --> group`

[And rewritten this to be an O(1) lookup](https://github.com/cabreraam/cuda-quantum/blob/513b895b8f4eb373d293e4c405a7817db6341fd7/cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp#L79-L90)

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: Similarly here, I now maintain a [mapping](https://github.com/cabreraam/cuda-quantum/blob/513b895b8f4eb373d293e4c405a7817db6341fd7/cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h#L57-L58) of 
`block -> [group indices]`

[So that lookups happen in `O(1)` now](https://github.com/cabreraam/cuda-quantum/blob/513b895b8f4eb373d293e4c405a7817db6341fd7/cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp#L92-L107)

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: Resolved now that the `getGroupContainingOp` lookup happens in `O(1)` time

cudaq/test/lib/Analysis/TestUnitaryOpGrouping.cpp
cabreraam: Yes, I'd like to defer this to a follow-on PR so that I can make sure that the core logic of the pass works. 

But once this is upstreamed, I'll start working on it immediately.

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
```diff
@@ -0,0 +1,60 @@
...
+class Region;
+} // namespace mlir
+
+namespace cudaq::quake::detail {
+struct UnitaryOpGroup {
+  UnitaryOpGroup() = default;
+
+  mlir::Block *block = nullptr;
+  mlir::Operation *firstOp = nullptr;
+  mlir::Operation *lastOp = nullptr;
```
taalexander: Aren't these trivially available from the `block` are are these the first op from ops (also trivial from the `ops` vector)?

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
```diff
@@ -0,0 +1,60 @@
...
+    performAnalysis(op);
+  }
+
+  const UnitaryOpGroups &getGroups() const { return groups; }
+
+  const mlir::Block *getBlockForGroup(const UnitaryOpGroup &group) const;
+  const UnitaryOpGroup *getGroupContainingOp(mlir::Operation *op) const;
+  llvm::SmallVector<const UnitaryOpGroup *>
+  getGroupsIn(const mlir::Block *block) const;
+  bool inSameGroup(mlir::Operation *op1, mlir::Operation *op2) const;
```
taalexander, line 74: These are relatively self-describing but it might be good to put a Doxygen comment for the public interface.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+
+const mlir::Block *
+cudaq::quake::detail::UnitaryOpGroupingAnalysis::getBlockForGroup(
+    const UnitaryOpGroup &group) const {
+  return group.block;
+}
+
+const cudaq::quake::detail::UnitaryOpGroup *
```
taalexander, line 63: I'm not sure what the pattern in the rest of the codebase is but i prefer to use `std::optional` to communicate present or not. Maybe see what pattern is in the codebase?

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,171 @@
...
+      continue;
+    }
+
+    flushGroupIfNonEmpty(groups, block, currUnitaryOps);
+  }
+
+  flushGroupIfNonEmpty(groups, block, currUnitaryOps);
+}
+
+void cudaq::quake::detail::UnitaryOpGroupingAnalysis::scanRegion(
```
taalexander: Could this be folded into a single block traversal? `scanBlock(block)` already walks every op in the block, then `scanRegion` walks the same ops again only to find nested regions. It seems simpler to recurse while scanning the block then when a non-unitary op flushes the current group, immediately scan any regions on that op. That preserves the “groups do not cross parent control-flow ops” rule while removing the second pass over each block.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  groups.push_back(std::move(group));
+  currUnitaryOps.clear();
+}
+} // namespace
+
+/// Maps 1) each unitary op to its corresponding group
+///      2) each block to a vector of group indices contained within that block
+///         - group indices are indices into the UnitaryOpGroups group vector
+void cudaq::quake::detail::UnitaryOpGroupingAnalysis::buildLookupTables() {
```
taalexander: Do we need a separate post-pass to build these maps? Since group indices are known when a group is flushed, `opToGroupIndex` and `blockToGroupIndices` could be populated at the same time `groups.push_back()` happens. That would remove `buildLookupTables()` and keep the analysis state construction in one place. It may also make `firstOp/lastOp` unnecessary?

taalexander (changes requested):
Thanks Anthony, a few additional potential opportunities for improvement.

cudaq/lib/Optimizer/Analysis/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+add_cudaq_library(OptAnalysis
+  UnitaryOpGrouping.cpp 
+  
+  DEPENDS
+    QuakeDialect
+)
```
schweitzpgi: heh. The eof monster appears!

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "cudaq/Optimizer/Analysis/UnitaryOpGrouping.h"
+#include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/Debug.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Region.h"
+#include "mlir/Support/TypeID.h"
```
schweitzpgi: This one is already in the .h file.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+#define DEBUG_TYPE "unitary-op-grouping-analysis"
+
+using namespace mlir;
+
+namespace {
```
schweitzpgi:
```suggestion
```
Should omit the anonymous namespace, since you're already (correctly) using `static`.

cabreraam: Done! Thanks for the catch

cudaq/test/lib/Analysis/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# ============================================================================ #
```
schweitzpgi, line 1: This doesn't look right. The tests in `cudaq/test` are llvm-lit tests and are not compiled and executed.

We probably need a new unittests for this actually. The unittests at the top-level are mostly (all?) related to testing the runtime rather than internals to the core. Take a look at the unit tests in tpls/llvm/llvm. We really should have something like a
```console
% make check-core-unit
...
```
that runs unittests on code in the `cudaq/` subdirectory.

cudaq/tools/cudaq-opt/cudaq-opt.cpp
```diff
@@ -60,6 +66,10 @@ int main(int argc, char **argv) {
   cudaq::registerAllCLOptions();
   cudaq::registerAllPasses();
 
+#ifdef CUDAQ_INCLUDE_TESTS
+  cudaq::test::registerTestUnitaryOpGroupingPass();
```
schweitzpgi: Yeah, we really want this to be run by gtest as a unit test...

schweitzpgi (commented):
I'd also suggest looking at the union-find data structure from LLVM to build up sets of Operations. The regtomem pass uses it, IIRC.

Great start, Anthony. Thanks for working on this.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: Yep, good catch. Resolved!

cudaq/lib/Optimizer/Analysis/CMakeLists.txt
cabreraam: Added eof newline!

cudaq/test/lib/Analysis/CMakeLists.txt
cabreraam: This might also relate to @taalexander's comment about 
> rewrite[ing] all of our groups as new functions (eg., the outlining). How much work would it be to test in this way? 
If I generate a `.qke` file to compare against, would this fit more into the mold of a standard  `lit` test?

Originally, I was trying to mimic how Flang tests its alias analysis pass:
- [`CMakeLists.txt` in `flang/test/lib/Analysis/AliasAnalysis`](https://github.com/llvm/llvm-project/blob/main/flang/test/lib/Analysis/AliasAnalysis/CMakeLists.txt)
- [test class to drive the analysis](https://github.com/llvm/llvm-project/blob/main/flang/test/lib/Analysis/AliasAnalysis/TestAliasAnalysis.cpp)
- [registering the test in  `fir-opt`](https://github.com/llvm/llvm-project/blob/main/flang/tools/fir-opt/fir-opt.cpp#L42-L46)

I am not against looking into rewriting the test to be located under `unittest` and driven by `gtest`, but just wanted to get your thoughts on my original justification for authoring the test the way that I did and why I located it in  `cudaq/test/lib/Analysis`. What do you think?

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
cabreraam: Yep, I agree. Done!

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
cabreraam: Yes, agreed. Removed now.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: This is a good question. I actually did think about doing this. I peeked at the codebase when I was implementing this (and just now as well for good measure) and found that if things were returned by value, `optional` was used. Otherwise, for references, a `nullptr` is returned. 

Here's an example of the [former](http://github.com/NVIDIA/cuda-quantum/blob/main/cudaq/lib/Optimizer/Transforms/RegToMem.cpp#L68-L72) and the [latter](https://github.com/NVIDIA/cuda-quantum/blob/main/cudaq/lib/Optimizer/Transforms/DependencyAnalysis.cpp#L1318-L1329)

So, I figured I'd keep this as is and save  `std::optional` when I need to return something by value

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: My original thinking was that the intent felt clearer when factored into a different function. But you're absolutely right; the info you need is all available at `groups.push_back()`. I've changed this now to implement the functionality of `buildLookupTables` in `flushGroupIfNonEmpty` in [this commit](http://github.com/NVIDIA/cuda-quantum/pull/4611/changes/ece0ef00270f2e0da8eaa9f3c56858c2e0fcd1ec).

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: A great suggestion; I've taken that approach in [this commit](https://github.com/NVIDIA/cuda-quantum/pull/4611/changes/ac003c334416344d3d1fe285b75026c87ba5ed5b).

cabreraam:
> I'd also suggest looking at the union-find data structure from LLVM to build up sets of Operations. The regtomem pass uses it, IIRC.

Had a chance to look through it; thanks for the pointer! That seems really useful for follow-on work to this PR. E.g., we could create equivalence classes for groups and then union the ones that are safe to merge after checking that there are only classical set-up ops/side-effect free ops between two sets.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  flushGroupIfNonEmpty(block, currUnitaryOps);
+}
+
+/// Scan every block in a region. Nested regions are reached from scanBlock when
+/// their owning operation is encountered.
+void cudaq::quake::detail::UnitaryOpGroupingAnalysis::scanRegion(
+    Region &region) {
+  for (Block &block : region)
+    scanBlock(block);
```
atgeller: Now that this function just defers to `scanBlock`, I wonder if it makes sense to have them split up and not just inline the loop from `scanBlock`. While I like the compositional approach of a function per container, I think it may be more readable/clear to have one recursive function sifting through all the containers. Just a preference though so feel free to ignore if you disagree :)

cudaq/test/lib/Analysis/TestUnitaryOpGrouping.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+using namespace mlir;
+
+namespace {
+using UnitaryOpGroup = cudaq::quake::detail::UnitaryOpGroup;
+using UnitaryOpGroupingAnalysis =
+    cudaq::quake::detail::UnitaryOpGroupingAnalysis;
+
+static void printBool(bool value) {
+  llvm::errs() << (value ? "true" : "false");
```
atgeller: Is the use of `errs` rather than `outs` important for testing? I would expect "print" to go to `outs`, whereas I think "dump" usually goes to `errs`.

atgeller (commented):
Thanks @cabreraam! Overall looks nice, just a few nits.

taalexander (approved):
Thank you for addressing my comments!

cudaq/test/lib/Analysis/CMakeLists.txt
schweitzpgi: Thanks for explaining.

Yeah, I still lean towards this approach being a hack. I'd rather not see `test/` be polluted with (more) tests done in hacky ways. (I reorganized these a few times already, once creating the end-to-end `targettests/`.) [Stroustrup said (paraphrasing): if you have to resort to using the C preprocessor to write your code, you're doing it wrong. This flang trick feels like that.]

We do really need a core unit tests area, and it shouldn't be too hard to refactor your test case into the gtest framework.

schweitzpgi (changes requested):
Let's see about reworking the executable test.

cudaq/lib/Optimizer/Analysis/UnitaryOpGrouping.cpp
cabreraam: Great catch! 110% agree. [I've inlined `scanRegion` so now it's just the one recursive call to `scanBlock`](https://github.com/NVIDIA/cuda-quantum/pull/4611/changes/1cbdcf40b77db9664f897d7a13d383a8ab237cf1#diff-857849852cff031811ac2f6bdcb05623322e4e4fd8b98ca2c51cf25d0b09f17f). Thanks @atgeller!

cudaq/test/lib/Analysis/TestUnitaryOpGrouping.cpp
cabreraam: Agreed. Though not necessary anymore in this case because of a refactor, definitely will remember going forward.

cabreraam:
> Let's see about reworking the executable test.

Thanks for the suggestion! For posterity, the reason this suggestion is good is because it seems hack-y to wrap a pass around the analysis and then register that pass conditionally to `cudaq-opt` and then place a `lit` test somewhere in `cudaq/test`. Placing it in `unittests` instead makes more sense since it better fits the mold of taking pieces of library code and testing it. And it doesn't require the aforementioned gymnastics just to test. This method is cleaner and is what we should use going forward for testing analysis-only passes.

[I made the changes](https://github.com/NVIDIA/cuda-quantum/pull/4611/changes/5d8438a065c371f81300c288437fb71596ba6707)! Notably:
- we construct the same `.qke` test case we had previously [but just use the MLIR builder API to construct it instead](https://github.com/cabreraam/cuda-quantum/blob/5d8438a065c371f81300c288437fb71596ba6707/unittests/Optimizer/UnitaryOpGroupingTest.cpp)
- I did my best to create a `make check-` style target [here](https://github.com/cabreraam/cuda-quantum/blob/5d8438a065c371f81300c288437fb71596ba6707/unittests/Optimizer/CMakeLists.txt#L63-L67). From the top level directory, it can be invoked using
  ```
  {make, ninja} -C build check-analysis-passes
  ```
  In LLVM you get the nice `add_lit_testsuite` CMake function, but it's really just a nice wrapper around `add_custom_target`. So that's what I used. Let me know if you think I should take a different approach @schweitzpgi

cudaq/tools/cudaq-opt/cudaq-opt.cpp
cabreraam: Testing now refactored to  `unittests` instead

atgeller (approved):
Thanks @cabreraam for setting up the analysis+testing infrastructure! This looks like it's in good shape to be merged.

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
```diff
@@ -0,0 +1,88 @@
...
+class Block;
+class Region;
+} // namespace mlir
+
+namespace cudaq::quake::detail {
+struct UnitaryOpGroup {
+  UnitaryOpGroup() = default;
+
+  mlir::Block *block = nullptr;
+  llvm::SmallVector<mlir::Operation *> ops;
```
schweitzpgi: nit: I think we use the re-exported name `mlir::SmallVector` in all (most?) places?

schweitzpgi (commented):
Thanks, Anthony. This is looking really great.

Can you add the hooks (files and directories) so that `make check-cudaq-unit` runs the unit tests in `cudaq/unittests/`?

Can be a separate issue and PR.

schweitzpgi (approved):
Clearing "request changes".

cudaq/include/cudaq/Optimizer/Analysis/UnitaryOpGrouping.h
cabreraam: Ah, thank you! I've switched over to `mlir::SmallVector`

PR #4614, opened 2026-05-27
Break up CUDAQTargetConfigUtils

lmondada:
This PR
- breaks up the `CUDAQTargetConfigUtils` library into an LLVM-free part (`CUDAQTargetConfig`) and the YAML file parsing utility (`CUDAQTargetConfigYaml`)
- Moves `cudaq/Support/TargetConfig.h` to `cudaq/Target/StaticConfig.h` and renames the type of the same name accordingly. This makes the distinction to the "runtime config" (as used e.g. in [BaseRemoteRESTQPU.h](https://github.com/NVIDIA/cuda-quantum/blob/f718fda3ca899f959ac3a8dbc6bce60d0f9f2d77/runtime/common/BaseRemoteRESTQPU.h#L69) explicit.

The upcoming `CompileTarget` type will also live within the `CUDAQTargetConfig` library (and needs to be LLVM-free)


taalexander (approved):
This refactor looks good to me. Just one question about the follow up work

> The upcoming CompileTarget type will also live within the CUDAQTargetConfig library (and needs to be LLVM-free)

I would have expected the `CompileTarget` would need MLIR symbols at some stage, eg., to configure a pass manager or to provide a module for further lowering?

lmondada:
> I would have expected the CompileTarget would need MLIR symbols at some stage, eg., to configure a pass manager or to provide a module for further lowering?

Agreed. For now, no MLIR is required, as all the pass manipulation is done on strings. I agree that using MLIR for this purpose (and for further purposes in the future) makes sense. My thinking is that we'd forward declare the MLIR symbols required for the method declarations and that it would be up to the implementations to link in MLIR if they want to provide implementations for those methods. This way we don't "force" MLIR on projects for which setting a couple of boolean flags would suffice.

When I say:

> needs to be LLVM-free

I mean that the runtime, the QPUs etc all need to include `CompileTarget`. I want this to be possible without having to link in MLIR everywhere.

taalexander:
Makes sense thank you!

Renaud-K:
I agree with removing LLVM dependency and breaking up libraries if it helps achieve the goal. I am having a harder time with the `StaticConfig` naming. Everybody knows what the yml target config file is. So `TargetConfig` seemed natural. And the information in `StaticConfig`  is probably used at runtime. So, I don't find that this name is communicating what it is.

lmondada:
@Renaud-K Ok, removed from the queue. I'll reconsider the naming and we can discuss before merging this in.

lmondada:
I've undone the name change

Renaud-K (approved):
Thank you Luca. I appreciate the name change.

PR #4616, opened 2026-05-27
[core] Harden regtomem and fix its use of eraseOp.

schweitzpgi:
This squashes a long-standing bug in regtomem. Under certain conditions, an Op might be erased before all of its uses were rewritten appropriately, resulting in stale pointers potentionally being dereferenced.

These bugs were observed on MacOS less than 0.2% of the time with a focused test to tickle this exact set of bugs.

These patches correct the issue and also add self-checking so as to monitor any regressions in the future.


schweitzpgi:
needs more work...

PR #4617, opened 2026-05-27
trim jupyter notebooks

mitchdz:


sacpis:
Curious, why are we removing these notebooks?

mitchdz:
This is mainly to simplify the notebooks that we carry in this repository as they create a maintenance burden. They might be moved to a different repo later, but that remains to be seen what happens to these.

docs/sphinx/applications/python/ptsbe.ipynb
bettinaheim, line 1: We should keep this - I had listed it under ML/AI focused, but there are other uses for ptsbe. If you remove it from this PR (meaning keep it, along with the corresponding links in the rst files), I'll approve this PR while I check if we intended to migrate ptsbe somewhere else.

mitchdz: Fixed in 3bccd1b22d1f94a80f9edff41a4a03d50e74b502

bettinaheim (approved):

PR #4618, opened 2026-05-27
[core] Fix bug in regtomem.

schweitzpgi:
The pass was erasing operations that may have still had uses instead of replacing those uses. This led to a small chance that a dead pointer would be dereferenced in the IR. This happened less than 0.2% of the time on MacOS with a focused test.


taalexander (approved):
Can confirm this fixes the transient reproducer on OSX. Thank you!

PR #4622, opened 2026-05-28
[cmake] Add CUDAQ_ENABLE_PROJECTS cmake variable

schweitzpgi:
Continue to restructure the build system to be more modular. This patch adds a CUDAQ_ENABLE_PROJECTS variable that should be used to enable different pieces of the CUDA-Q project as a whole. There is more refactoring to do to decompose and separate some of the build concerns. This is an initial step only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,64 @@ option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Fron
 option(CUDAQ_ENABLE_CC "Enable CUDA-Q code coverage generation." OFF)
 option(CUDAQ_REQUIRE_OPENMP "Fail the build if OpenMP is not found." OFF)
 option(CUDAQ_ENABLE_SANITIZERS "Enable Address Sanitizer (ASan) and Undefined Behavior Sanitizer (UBSan)." OFF)
+
+# Sub-projects that can be enabled independently, in the style of
+# LLVM_ENABLE_PROJECTS. Accepts a semicolon-separated list.
+set(CUDAQ_ALL_PROJECTS "cudaq;runtime;python")
+set(CUDAQ_ENABLE_PROJECTS "" CACHE STRING
+    "Semicolon-separated list of CUDA-Q sub-projects to build. Valid values: ${CUDAQ_ALL_PROJECTS}. Defaults to 'cudaq;runtime' (plus 'python' if CUDAQ_ENABLE_PYTHON is set).")
+
+# Back-compat: if CUDAQ_ENABLE_PROJECTS wasn't given, derive it from the
+# legacy per-project flags so existing build scripts keep working, and warn
+# the user that those flags are deprecated.
+if(NOT CUDAQ_ENABLE_PROJECTS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -172,6 +172,64 @@ option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Fron
...
+  if(CUDAQ_ENABLE_PYTHON)
+    list(APPEND CUDAQ_ENABLE_PROJECTS "python")
+  endif()
+
+  if(CUDAQ_DISABLE_RUNTIME)
+    message(DEPRECATION
+      "CUDAQ_DISABLE_RUNTIME is deprecated. Use CUDAQ_ENABLE_PROJECTS to "
+      "control which sub-projects are built (omit 'runtime' to disable it).")
+  endif()
+  if(DEFINED CUDAQ_ENABLE_PYTHON)
```
sacpis, line 199: Should we use one type of check `if(CUDAQ_DISABLE_RUNTIME)` or `if(DEFINED CUDAQ_DISABLE_RUNTIME)`?

CMakeLists.txt
```diff
@@ -172,6 +172,64 @@ option(CUDAQ_DISABLE_CPP_FRONTEND "Build without the CUDA-Q C++ Clang-based Fron
...
+  endforeach()
+endif()
+
+# Map CUDAQ_ENABLE_PROJECTS onto the legacy per-project flags so the rest of
+# this file and sub-CMakes keep working unchanged.
+if(NOT "runtime" IN_LIST CUDAQ_ENABLE_PROJECTS)
+  set(CUDAQ_DISABLE_RUNTIME ON CACHE BOOL "" FORCE)
+endif()
+if("python" IN_LIST CUDAQ_ENABLE_PROJECTS)
+  set(CUDAQ_ENABLE_PYTHON ON CACHE BOOL "" FORCE)
```
sacpis, line 230: Seems like here we are forcefully setting `CUDAQ_ENABLE_PYTHON` to "" in the cache. On the next configure, `CUDAQ_ENABLE_PROJECTS` is still empty in the cache (default is set to "" on line [179](https://github.com/NVIDIA/cuda-quantum/pull/4622/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR179)). In this case, we renter the backward compatibility branch and trigger the deprecation at line [199](https://github.com/NVIDIA/cuda-quantum/pull/4622/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR199) even though the user has not set `CUDAQ_ENABLE_PYTHON` flag.

sacpis (approved):
Overall LGTM. Thanks @schweitzpgi. Left few comments.

CMakeLists.txt
schweitzpgi: Those are different tests, eh?

CMakeLists.txt
schweitzpgi: I'm not too worried about optimizing the deprecation warning. If it comes out every time, it will get lost in the sea of other warnings that get dumped out. heh

PR #4623, opened 2026-05-28
[cmake, end-to-end tests] Make the backends selectable

schweitzpgi:
We want the user to be able to select which backends they want to include in their build rather than presuming that they want every backend to be built. This is an incremental step in that direction.


CMakeLists.txt
```diff
@@ -89,6 +89,26 @@ if (NOT DEFINED CUDAQ_ENABLE_BRAKET_BACKEND)
...
+# Enable IQM target by default.
+if (NOT DEFINED CUDAQ_ENABLE_IQM_BACKEND)
+  set(CUDAQ_ENABLE_IQM_BACKEND ON CACHE BOOL "Enable building the IQM target.")
+endif()
+
+# Enable OQC target by default.
+if (NOT DEFINED CUDAQ_ENABLE_OQC_BACKEND)
+  set(CUDAQ_ENABLE_OQC_BACKEND ON CACHE BOOL "Enable building the OQC target.")
+endif()
+
```
cabreraam, line 111: Ah, okay. So if the user just wanted the OQC backend and were using the dev container, they might issue something like

```
bash scripts/build_cudaq.sh -- -DCUDAQ_ENABLE_ANYON_BACKEND=OFF -DCUDAQ_ENABLE_IONQ_BACKEND=OFF -DCUDAQ_ENABLE_IQM_BACKEND=OFF
```
since by default all of them are turned on?

cabreraam:
LGTM. Just added a clarification point but definitely makes sense. 

My only thought is that, if we shouldn't presume the user wants to build all backends, should they be default _off_ instead of _on_?

targettests/execution/bug_qubit.cpp
```diff
@@ -9,15 +9,15 @@
...
-// RUN: nvq++ --target anyon      --emulate %s -o %t && %t
+// RUN: if %anyon_avail; then nvq++ --target anyon      --emulate %s -o %t && %t; fi
 // RUN: nvq++ --target infleqtion --emulate %s -o %t && %t
-// RUN: nvq++ --target ionq       --emulate %s -o %t && %t
-// RUN: nvq++ --target iqm        --emulate %s -o %t
+// RUN: if %ionq_avail; then nvq++ --target ionq       --emulate %s -o %t && %t; fi
+// RUN: if %iqm_avail; then nvq++ --target iqm        --emulate %s -o %t; fi
 // RUN: IQM_QPU_QA=%iqm_tests_dir/Crystal_5.txt  %t
 // RUN: IQM_QPU_QA=%iqm_tests_dir/Crystal_20.txt %t
 // RUN: IQM_QPU_QA=%iqm_tests_dir/Crystal_54.txt %t
```
sacpis: Just a question. what happens when `iqm` is disabled? Does `%t` exists or is it a stale binary from another RUN? In that case should we make this gated as well?

runtime/cudaq/platform/default/rest/helpers/CMakeLists.txt
```diff
@@ -6,14 +6,22 @@
...
 # ============================================================================ #
 
-add_subdirectory(anyon)
+if (CUDAQ_ENABLE_ANYON_BACKEND)
+  add_subdirectory(anyon)
+endif()
 if (AWSSDK_ROOT)
   add_subdirectory(braket)
 endif()
 add_subdirectory(infleqtion)
```
sacpis, line 15: Should we also add a condition for `Infleqtion` similar to others?

unittests/backends/CMakeLists.txt
```diff
@@ -84,14 +84,22 @@ endfunction()
...
 find_package(Python COMPONENTS Interpreter)
 if (OPENSSL_FOUND AND CUDAQ_ENABLE_PYTHON AND CUDAQ_TEST_MOCK_SERVERS)
-  add_subdirectory(anyon)
+  if (CUDAQ_ENABLE_ANYON_BACKEND)
+    add_subdirectory(anyon)
+  endif()
   if (CUDAQ_ENABLE_BRAKET_BACKEND)
     add_subdirectory(braket)
   endif()
   add_subdirectory(infleqtion)
```
sacpis, line 93: same here?

sacpis (approved):
Overall LGTM. Thanks @schweitzpgi. Left a few comments.

runtime/cudaq/platform/default/rest/helpers/CMakeLists.txt
schweitzpgi: I left that one alone because it is set up in a different part of the source tree, etc. It seemed better to avoid it. :)

targettests/execution/bug_qubit.cpp
schweitzpgi: Thanks. Will fix this one.

schweitzpgi: [17ea0cf](https://github.com/NVIDIA/cuda-quantum/pull/4623/commits/17ea0cf0ecf05241599e6636b0a99f1f7631a18b)

schweitzpgi:
> LGTM. Just added a clarification point but definitely makes sense.
> 
> My only thought is that, if we shouldn't presume the user wants to build all backends, should they be default _off_ instead of _on_?

I think you are right that they should be opt-in instead of opt-out. But everything about the code base presumes you've opted-in as it stands, so to try and keep this smallish and not break everything ... I did this as step 0.5. :)

cabreraam (approved):
LGTM!

PR #4624, opened 2026-05-28
Build observe_result during finalization.

Renaud-K:
This PR is moving the production of the observe_result in the simulator finalization phase, using the observe_policy. This is in order to bring symmetry with how the sample_result is produced. It builds on the changes of #4339 and adds a new observe_policy overload. 

This does not bring end-to-end support for the observe_policy, but like #4339, is a first step in this direction and a change worth testing on its own.

### DIFF
--- a/runtime/cudaq/qis/execution_manager.cpp
+++ b/runtime/cudaq/qis/execution_manager.cpp
@@ -12,6 +12,7 @@
 #include "nvqir/CircuitSimulator.h"
 #include "cudaq/algorithms/policy_cpos.h"
 #include "cudaq/algorithms/policy_dispatch.h"
+#include <stdexcept>
 
 using namespace cudaq;
 
@@ -55,6 +56,10 @@ void ExecutionManager::finalizeExecutionContext(ExecutionContext &ctx) {
     policies::visitResult(
         [&]() { return cudaq::finalize_execution_manager(*this, policy, ctx); },
         [&](sample_result &&r) { ctx.result = std::move(r); },
+        [&](observe_result &&r) {
+          ctx.result = r.raw_data();
+          ctx.expectationValue = r.expectation();
+        },
         [&](policies::void_result &&r) {});
   });
 }
--- a/runtime/cudaq/qis/execution_manager.cpp
+++ b/runtime/cudaq/qis/execution_manager.cpp
@@ -12,7 +12,6 @@
 #include "nvqir/CircuitSimulator.h"
 #include "cudaq/algorithms/policy_cpos.h"
 #include "cudaq/algorithms/policy_dispatch.h"
-#include <stdexcept>
 
 using namespace cudaq;
 
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -267,12 +267,19 @@ public:
       cudaq::policies::visitResult(
           [&]() { return finalize_simulation_circuit(*this, policy, ctx); },
           [&](cudaq::sample_result &&r) { ctx.result = std::move(r); },
+          [&](cudaq::observe_result &&r) {
+            ctx.result = r.raw_data();
+            ctx.expectationValue = r.expectation();
+          },
           [&](cudaq::policies::void_result &&r) {});
     });
   }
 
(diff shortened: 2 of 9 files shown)


Renaud-K:
> /ok to test d6b70de6f0346b8b7ba371b623f406f55a8d3e5d

runtime/cudaq/qis/execution_manager.cpp
```diff
@@ -12,6 +12,7 @@
 #include "nvqir/CircuitSimulator.h"
 #include "cudaq/algorithms/policy_cpos.h"
 #include "cudaq/algorithms/policy_dispatch.h"
+#include <stdexcept>
```
sacpis: nit: let's remove it.

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1017,6 +1024,38 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
+      double sum = 0.0;
+      for (const auto &term : H) {
+        if (term.is_identity())
+          sum += term.evaluate_coefficient().real();
+        else {
+          auto [exp, data] = measureSpinOp(term);
+          results.emplace_back(data.to_map(), term.get_term_id(), exp);
+          sum += term.evaluate_coefficient().real() * exp;
+        }
+      };
```
sacpis: nit: remove ;

runtime/nvqir/cudensitymat/CuDensityMatSim.cpp
```diff
@@ -155,6 +156,13 @@ class CuDensityMatSim : public nvqir::CircuitSimulatorBase<double>,
     return cudaq::sample_result();
   }
 
+  cudaq::observe_result
+  finalizeExecutionContext(const cudaq::observe_policy &,
+                           cudaq::ExecutionContext &ctx) override {
+    finalizeExecutionContextImpl(ctx);
+    return cudaq::observe_result();
```
sacpis: I think this is unreachable code. Since with dynamics target, we call observe through evolve api. The `finalizeExecutionContextImpl` is only reached if the a user calls `cudaq::observe()` directly on the dynamics target.

I think we should throw similar to photonics target [here](https://github.com/NVIDIA/cuda-quantum/pull/4624/changes#diff-5fb9cbc6c602222aab9e0d8ebe57d5f2ef93baf16fdb14c9e47a5f4b1ed896a7R206).

@1tnguyen for viz.

sacpis (approved):

runtime/nvqir/cudensitymat/CuDensityMatSim.cpp
Renaud-K: Just above we have.
```
  cudaq::sample_result
  finalizeExecutionContext(const cudaq::sample_policy &policy) override {
    return cudaq::sample_result();
  }
```

Granted, this was added by me, but we probably were not throwing in this case either. Should we throw in both cases? I was also surprised with the name `cuDensityMatSim.cpp`. I thought this had to do with GPU accelerated DM simulations.

sacpis: Yes, I think so. @1tnguyen to confirm.

Renaud-K:
> /ok to test 505ba81230af7580d0fbb43e549d6ee85158b86f

runtime/nvqir/CircuitSimulator.h
```diff
@@ -1017,6 +1024,38 @@ class CircuitSimulatorBase : public CircuitSimulator {
...
+                               "cudaq::spin_op.");
+
+    std::vector<cudaq::ExecutionResult> results;
+    cudaq::spin_op &H = ctx.spin.value();
+    assert(cudaq::spin_op::canonicalize(H) == H);
+
+    if (ctx.canHandleObserve) {
+      auto [exp, data] = measureSpinOp(H);
+      return cudaq::observe_result(exp, H, data);
+    } else {
```
schweitzpgi, line 1042: else after return

Renaud-K: Sorry Eric, only saw this after the PR went through the merge queue. I will apply the changes in the next PR.

PR #4625, opened 2026-05-28
Fix incorrect bitstring qubit ordering on IQM backends (#4621)

### LINKED ISSUE #4621: Qubit ordering problem on IQM backends
### Required prerequisites

### Describe the bug

Qubit ordering on counts returned from sampling IQM backends appears to be incorrect for qvectors of size 5 or greater.

```import cudaq
import os

os.environ['IQM_TOKEN'] = input()

iqm_url = "https://cocos.resonance.meetiqm.com/garnet" 

cudaq.set_target("iqm", url=iqm_url)

shots = 1000    

for qubit_count in range(3, 10):

    @cudaq.kernel
    def circuit(qubit_count: int):
        
        qvector = cudaq.qvector(qubit_count)
        
        x(qvector[2])
        mz(qvector)

    result = cudaq.sample(circuit, qubit_count, shots_count = shots)
    
    counts = dict[Any, Any](result.items())
    most_dominant = max(counts, key=counts.get)
    print(f"Most dominant: {most_dominant} ({counts[most_dominant]} shots)")
```

### Expected behavior

Should output as most dominant
001
0010
00100
001000
0010000
00100000
001000000
0010000000

Not a regression

### Environment

- **CUDA-Q version**: 0.14.2
- **Python version**: 3.11.15
- **C++ compiler**: N/A
- **Operating system**: WSL

### Suggestions

Not a suggestion, but my output:
```
Most dominant: 001 (973 shots)
Most dominant: 0010 (983 shots)
Most dominant: 00010 (972 shots)
Most dominant: 000010 (965 shots)
Most dominant: 0000100 (965 shots)
Most dominant: 00001000 (948 shots)
Most dominant: 000010000 (954 shots)
```

anpaz:
**Summary**

Sampling a CUDA-Q kernel on the IQM target (e.g. Garnet) returned bitstrings whose `1` bits appeared at the wrong positions whenever the kernel used 5 or more qubits. For example, applying `x(qvector[2])` on 5 qubits produced `"01000"` instead of the expected `"00100"`. The result was correct for 3–4 qubits because the qubit-mapping pass happened to pick an identity placement in those small cases.

Fixes #4621.

**Root cause**

The qubit-mapping MLIR pass permutes virtual qubits onto physical qubits and records the inverse permutation in a `mapping_reorder_idx` function attribute. The runtime flow is:

1. `Compiler.cpp` extracts the attribute into `KernelExecution::mapping_reorder_idx`.
2. `Executor::execute` serializes it into the backend config as `reorderIdx.<task_id>`.
3. On the client, `Future::get` calls `serverHelper->initialize(config)`.
4. `ServerHelper::parseConfigForCommonParams(config)` is responsible for picking those entries out of the config and populating `ServerHelper::reorderIdx`.
5. `IQMServerHelper::processResults` then calls `sample_result::reorder(...)` to map physical-qubit-ordered bitstrings back to the user's virtual allocation order.

`IQMServerHelper::initialize` was the only REST helper that never called `parseConfigForCommonParams`, so `reorderIdx` was always empty and the reorder step in `processResults` silently became a no-op. The server's results were correct; the client just never un-permuted them.

**Fix**

Add the missing call to `parseConfigForCommonParams(config)` at the end of `IQMServerHelper::initialize`. While here, replace the stale `// FIXME: VERIFY THIS` comment in `processResults` with an accurate description of the ordering contract.

**Test**

Added a unit-test regression check in IQMTester.cpp:

- `IQMTester.processResultsAppliesReorderIdxFromConfig`
- Drives `IQMServerHelper` directly: seeds `config["reorderIdx.test-job-id"] = "[1, 2, 0]"`, calls `initialize`, feeds a synthetic `ServerMessage` with counts `{"100": 1000}`, and asserts that `processResults` returns the reordered bitstring `"001"`.
- Verified it fails on `main` (`Got "100"`) and passes with the fix.

Full `ctest -R iqm-tests` passes locally.

**Why the IQM mock server didn't catch this earlier**

The Crystal_20 mock used by the existing IQM e2e tests has no two-qubit-gate constraints for the X-only kernels in those tests, so the mapper picks an identity placement and the reorder step happens to be a no-op even on a correct backend. The new test bypasses the mock and exercises the helper directly so it always reproduces the bug.

**Scope check**

Surveyed every other REST helper:

- IonQ, Quantinuum, Braket, OQC, QCI, Infleqtion, Anyon, qBraid, Pasqal, Fermioniq, QuEra — already call `parseConfigForCommonParams`.
- Orca and Scaleway don't, but their targets don't use the qubit-mapping pipeline, so they are not affected.

No other backends require changes.

**Files changed**

- IQMServerHelper.cpp — add the call; refresh comment.
- IQMTester.cpp — add regression test plus required includes.


unittests/backends/iqm/IQMTester.cpp
```diff
@@ -71,6 +73,51 @@ CUDAQ_TEST(IQMTester, executeMultipleMeasuredQubitsProgram) {
...
+  cudaq::ServerMessage response = {
+      {"status", "ready"},
+      {"counts_batch",
+       cudaq::ServerMessage::array(
+           {{{"counts", {{"100", 1000}}}, {"measurement_keys", {"m_0"}}}})}};
+
+  std::string jobIdCopy = jobId;
+  auto result = serverHelper->processResults(response, jobIdCopy);
+
+  // After reordering, the dominant bitstring must be "010".
```
sacpis: nit: need to match the output.

sacpis (approved):
LGTM. Thanks @anpaz.

Renaud-K (approved):
LGTM. Thank you.

PR #4632, opened 2026-05-28
Update docs after LLVM migration

atgeller:


Building.md
```diff
@@ -123,14 +123,9 @@ export LLVM_PROJECTS='clang;lld;mlir;python-bindings'
 ## Building CUDA-Q with a custom LLVM version
 
 CUDA-Q is intended to be built using the LLVM commit that the submodule is set
-to (currently LLVM 16.0.6). If you need to update the LLVM commit, or if you are
+to (currently LLVM 22.1). If you need to update the LLVM commit, or if you are
```
sacpis: nit: 22.1.4

cudaq/include/cudaq/Optimizer/Dialect/CC/CCOps.td
```diff
@@ -1818,8 +1818,8 @@ def cc_NoInlineCallOp : CCOp<"noinline_call",
     [CallOpInterface, SymbolUserOpInterface]> {
   let summary = "Create a call that the inliner will not inline.";
   let description = [{
-    The MLIR inliner in LLVM 16 is too smart and will inline trivial calls.
-    Moreover, there is no mechanism in this version of MLIR to stop the inliner
+    The MLIR inliner in LLVM is too smart and will inline trivial calls.
```
sacpis, line 1821: I think we should put LLVM version as well (Upto Eric and you though). Just would like to specify which version has this ability for inlining trivial calls.

sacpis (approved):
LGTM. Thanks @atgeller.

Building.md
atgeller: Fixed that, and an instance above for MacOS.

cudaq/include/cudaq/Optimizer/Dialect/CC/CCOps.td
atgeller: I'm good either way. @schweitzpgi, any preference?

schweitzpgi: I actually haven't checked if the current version of MLIR has some knobs to control inlining more precisely. I don't know if we have to call out a particular version. As far as I know no version has had the necessary tuning knobs and it may be the case it never happens.

atgeller: I think we're good to leave it as is and we can investigate further later then.

PR #4633, opened 2026-05-29
[Quake] Fold `veq_size` through `cc.if` to activate `ForwardEmptyVeqSizePattern`

### LINKED ISSUE #4631: Mystery crash with cudaq::run in CI
See the crash in [this log](https://github.com/NVIDIA/cuda-quantum/actions/runs/26593612145/job/78359301142?pr=4610).

This is a python test that used cudaq::run, but I've disabled that part of the test as it was crashing for this one particular configuration consistently.

The diff to disable was:
```console
diff --git a/python/tests/mlir/slice.py b/python/tests/mlir/slice.py
index aa86d8f9..30717d0b 100644
--- a/python/tests/mlir/slice.py
+++ b/python/tests/mlir/slice.py
@@ -28,8 +28,8 @@ def test_static_case():

     print(static_empty_slice)

-    result = cudaq.run(static_empty_slice, shots_count=10)
-    print(result)
+    #result = cudaq.run(static_empty_slice, shots_count=10)
+    #print(result)


 def test_another_case():
@@ -47,8 +47,8 @@ def test_another_case():

     print(static_notempty_slice)

-    result = cudaq.run(static_notempty_slice, shots_count=10)
-    print(result)
+    #result = cudaq.run(static_notempty_slice, shots_count=10)
+    #print(result)


 def test_dynamic_case():
@@ -68,8 +68,8 @@ def test_dynamic_case():

     print(dynamic_empty_slice)

-    result = cudaq.run(dynamic_empty_slice, shots_count=10)
-    print(result)
+    #result = cudaq.run(dynamic_empty_slice, shots_count=10)
+    #print(result)


 def test_another_dynamic_case():
@@ -91,8 +91,8 @@ def test_another_dynamic_case():

     print(dynamic_notempty_slice)

-    result = cudaq.run(dynamic_notempty_slice, shots_count=10)
-    print(result)
+    #result = cudaq.run(dynamic_notempty_slice, shots_count=10)
+    #print(result)


 # leave for gdb debugging
@@ -110,7 +110,6 @@ if __name__ == "__main__":
 # CHECK:               %[[CMPI_0:.*]] = arith.cmpi slt, %[[VAL_0]], %[[CONSTANT_0]] : i64
 # CHECK:               cc.condition %[[CMPI_0]](%[[VAL_0]] : i64)
 # CHECK:             } do {
-# CHECK:  [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

 # CHECK-LABEL:   func.func @__nvqpp__mlirgen__static_notempty_slice..0x
 # CHECK:           %[[CONSTANT_0:.*]] = arith.constant 1 : i64
@@ -121,7 +120,6 @@ if __name__ == "__main__":
 # CHECK:               %[[CMPI_0:.*]] = arith.cmpi slt, %[[VAL_0]], %[[CONSTANT_0]] : i64
 # CHECK:               cc.condition %[[CMPI_0]](%[[VAL_0]] : i64)
 # CHECK:             } do {
-# CHECK: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]

 # CHECK-LABEL:   func.func @__nvqpp__mlirgen__dynamic_empty_slice..0x
 # CHECK:           %[[CONSTANT_0:.*]] = arith.constant 0 : i64
@@ -146,7 +144,6 @@ if __name__ == "__main__":
 # CHECK:               %[[CMPI_1:.*]] = arith.cmpi slt, %[[VAL_0]], %[[VEQ_SIZE_0]] : i64
 # CHECK:               cc.condition %[[CMPI_1]](%[[VAL_0]] : i64)
 # CHECK:             } do {
-# CHECK: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

 # CHECK-LABEL:   func.func @__nvqpp__mlirgen__dynamic_notempty_slice..0x
 # CHECK:           %[[CONSTANT_0:.*]] = arith.constant 0 : i64
@@ -185,4 +182,3 @@ if __name__ == "__main__":
 # CHECK:               %[[CMPI_2:.*]] = arith.cmpi slt, %[[VAL_3]], %[[VEQ_SIZE_0]] : i64
 # CHECK:               cc.condition %[[CMPI_2]](%[[VAL_3]] : i64)
 # CHECK:             } do {
-# CHECK: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4]
```

1tnguyen:
Fixes #4631, where Python slice lowering could leave `quake.veq_size` on a `cc.if` result whose else branch yielded `cc.undef`.

The `ForwardEmptyVeqSizePattern` did not fire because it saw the `cc.if` result, not the branch-local `cc.undef`.

- Add a canonicalization pattern that hoists `quake.veq_size` through the `cc.if` by appending the computed size as an extra `cc.if` result. This exposes the else-branch `quake.veq_size(cc.undef)` to the existing empty-veq fold, producing `0` for the empty slice case.

- Re-enable the previously disabled `cudaq.run` slice tests.

Note: I reproduced the segfault on an ARM machine. The issue appears target-sensitive: x86 happened to tolerate the `undef` value, while ARM did not.


cudaq/lib/Optimizer/Dialect/Quake/CanonicalPatterns.inc
```diff
@@ -128,6 +128,111 @@ struct ForwardEmptyVeqSizePattern
   }
 };
 
+// %0 = cc.if(%cond) -> !quake.veq<?> {
```
schweitzpgi: Is the input pattern coming from an explicit place in lowering? Since this is such a special case, I wonder if we can't handle it at the source of the issue.

1tnguyen: Indeed, I believe it came from this if/else op:

https://github.com/NVIDIA/cuda-quantum/blob/2a80ac4a2030fb59aedaae27adaf736976a60a83/python/cudaq/kernel/ast_bridge.py#L4631-L4635

I also pondered the idea of changing it at the bridge side, but I couldn't think of a clean solution to support dynamical slice and be compatible with `ForwardEmptyVeqSizePattern`. Do you have an idea?

schweitzpgi: Yeah, that's sort of gross. Eek!

What if the Python bridge looks at the argument its about to use to `quake.veq_size`, seeing if it is a `cc.if` and does your rewrite (as in this PR) immediately at that point?

That's sort of odd too, though. Another idea would be to put your pattern in a new Python bridge cleanup pass. I like that a bit better as we can add other "cool" patterns there to cleanup squirrelly residue from the front end. :)

schweitzpgi: BTW, `ForwardEmptyVeqSizePattern` was supposed to kick in after arg synth and aggressive const prop, which would be when the `if` condition ought to be a constant.

Were you seeing a case where that wasn't true?

1tnguyen: Yeah, option (1) (Python bridge looks at the argument its about to use to `quake.veq_size`) is indeed quite odd :). There are more than a handful of places in the Python bridge where we add `quake.VeqSizeOp`. 

So, I've rewritten this as a "new Python bridge cleanup pass" as suggested. This makes sense as this is a Python-specific thing.

> ForwardEmptyVeqSizePattern was supposed to kick in after arg synth and aggressive const prop, which would be when the if condition ought to be a constant.

That's true. However, we're supposed to support a truly dynamical value case, not just runtime arguments. For example, one of the test cases is:

```
index = int(mz(q[4]))
x(q[0:index])
```

i.e., the index is converted from quantum measurement results. Hence, the if/else would still survive after arg synth and aggressive const prop.

schweitzpgi (approved):
Thanks, Thien. The new pass looks like it will be handy to undo other code patterns that the Python bridge generates as premature optimizations, etc.

PR #4634, opened 2026-05-29
[testing] Move executable test to the correct place.

schweitzpgi:
Move an executable test out of cudaq/test/ and into targettests/execution/ where the rest of the executable tests live.


cabreraam (approved):
Makes sense. LGTM!

PR #4635, opened 2026-05-29
[core] Remove old testing code from driver

schweitzpgi:
Old testing code leaked into main. Remove it.


cabreraam (approved):
Heh, there's a story here that someone should tell me about someday.

LGTM!

PR #4636, opened 2026-05-29
[python] Use Compiler class within PythonLauncher

lmondada:
This PR rewrites `PythonLauncher::compileModule` to re-use the `Compiler` class already used in `BaseRemoteRESTQPU`. To achieve this without modifying the compilation pipelines effectively run, additional compilation flags are added to `CompileTarget`.

This removes code duplication and unifies the compiler endpoint: any further changes to the JIT compilation pipeline will only have to modify the code in one place -- the `Compiler` class.

Two further consequences of this simplification that will be the subject of further PRs:
- all QPUs share the same logic for `compileModule`: we will thus be able to unify `compileModule` and move it out of the QPU classes,
- the QPUs will no longer depend on the compiler (or on MLIR), which will simplify dependencies and make it possible to remove the `PythonLauncher` plugin system of `runtime/cudaq/platform/default/python/QPU.cpp`


Renaud-K (commented):
This looks good to me. Than you for the unification effort.

Renaud-K (approved):
This looks good to me. Than you for the unification effort.

sacpis:
Seems like we have some lowering quake failures at the remote platform.
https://github.com/NVIDIA/cuda-quantum/actions/runs/26841236004/job/79181608131?pr=4636#step:14:7490

sacpis (commented):
Will review again once we have fix those tests.

lmondada:
> Seems like we have some lowering quake failures at the remote platform

yes good catch, I had introduced an additional flag after discussing with Eric yesterday and missed setting it in one of the `getCompileTarget` overloads. My latest commit should fix that!

runtime/cudaq/platform/default/python/QPU.cpp
sacpis: Does `lower_to_openqasm` runs passes and translation on the incorrect module? I think the pipeline
runPassPipeline
         |
prepareModule
        |
lowerQuakeCodeBuildModule

builds a ModuleOp and clones the functions into it. The original module is left as raw quake. So the OpenQASM passes and translateToOpenQASM here runs against the un-lowered module? Should runPassManager and translateToOPenQASM use compiled_module? Seems like it is correctly being used online 101 and 107.

runtime/internal/compiler/Compiler.cpp
```diff
@@ -166,19 +166,50 @@ void cudaq_internal::compiler::Compiler::applyPipeline(
...
-    contextPtr->disableMultithreading();
-  if (enablePrintMLIREachPass)
-    pm.enableIRPrinting();
   if (failed(cudaq_internal::compiler::runPassManager(pm,
                                                       moduleOp.getOperation())))
     throw std::runtime_error("Remote rest platform Quake lowering failed.");
 }
 
-std::pair<mlir::ModuleOp, mlir::func::FuncOp>
+bool eraseNonCallableArguments(std::span<void *const> &rawArgs,
```
sacpis: Should we make this a static function similar to `hasConditionalsOnMeasureResults`? or move both of these helpers into an anonymous namespace?

runtime/cudaq/platform/qpu.cpp
```diff
@@ -64,6 +64,17 @@ cudaq::QPU::runJITCompiledModule(const CompiledModule &compiled,
...
   ScopedTraceWithContext(cudaq::TIMING_LAUNCH, "QPU::runJITCompiledModule",
                          compiled.getName());
+
+  // Propagate metadata from the compiled artifact to the execution context.
+  if (auto ctx = getExecutionContext()) {
+    ctx->hasConditionalsOnMeasureResults =
+        compiled.getMetadata().hasConditionalsOnMeasureResults;
+
+    if (ctx->name == "resource-count" && compiled.getResources()) {
+      nvqir::resource_counter::prepopulate(std::move(*compiled.getResources()));
```
sacpis: Isn't std::move on a const compiled a silent copy?

sacpis (approved):
Overall LGTM. Thanks @lmondada. Left a few comments.

runtime/cudaq/platform/default/python/QPU.cpp
lmondada: oof, good catch! Thank you!!

runtime/internal/compiler/Compiler.cpp
lmondada: yes for sure, made it static.

runtime/cudaq/platform/qpu.cpp
lmondada: Another good catch, thanks! I had copied it over from `BaseRemoteRESTQPU`, where the move actually does something. Dropped it here.

PR #4639, opened 2026-05-29
python: fix cudaq.to_cupy GPU buffer wrapping

### LINKED ISSUE #4638: python: cudaq.to_cupy() wraps GPU state with a fixed 1024-byte CuPy buffer
### Required prerequisites

### Describe the bug

`cudaq.to_cupy()` in `python/cudaq/runtime/state.py` currently wraps each GPU tensor with:

```python
cp.cuda.UnownedMemory(tensor.data(), 1024, owner=None)
```

That exports a fixed 1024-byte device region regardless of the tensor size, while the returned `cp.ndarray` is created from the full tensor extents. For larger GPU-backed states, the CuPy view can therefore describe more data than the wrapped memory region actually covers.

There is also a lifetime issue here: `owner=None` means the returned CuPy array does not keep the originating `cudaq.State` alive. If the state is released while the CuPy array is still referenced, the CuPy array may hold a dangling device pointer.

This looks distinct from #2017, which fixed dtype selection for fp64 targets.

A simple reproducer is any GPU-backed state that exceeds 1024 bytes.

For example, a 10-qubit fp32 statevector has 1024 amplitudes, which is 8192 bytes of `complex64` data:

```python
import cudaq
import cupy as cp

cudaq.set_target('nvidia')

kernel = cudaq.make_kernel()
q = kernel.qalloc(10)
kernel.h(q[0])
for i in range(9):
    kernel.cx(q[i], q[i + 1])

state = cudaq.get_state(kernel)
arr = cudaq.to_cupy(state)[0]

print(arr.shape)  # (1024,)
print(arr[-1])    # element access extends beyond the 1024-byte wrapped region
```

Likewise, even fp64 statevectors exceed 1024 bytes once they go past 64 amplitudes.

### Expected behavior

`cudaq.to_cupy()` should wrap the full device allocation for each tensor and keep the source state alive for the lifetime of the CuPy view.

There is already a safer pattern elsewhere in the repo that computes the byte size from:

```python
tensor.get_num_elements() * tensor.get_element_size()
```

and passes `owner=state` when constructing `UnownedMemory`.

Not a regression. This looks long-standing: `to_cupy()` was introduced with the fixed `1024` size / `owner=None` behavior, and later changes only touched dtype inference.

### Environment

- **CUDA Quantum version**: `$(git rev-parse HEAD)`
- **Python version**: N/A
- **C++ compiler**: N/A
- **Operating system**: macOS

### Suggestions

If this direction makes sense, I’m happy to send a small fix and regression test. I already contribute in NVIDIA OSS, so happy to take this one too.

fallintoplace:
## Summary
- compute the wrapped CuPy buffer size from each tensor's element count and element size
- keep the source `cudaq.State` alive by passing `owner=state` to `UnownedMemory`
- add a regression test that exercises a GPU-backed state larger than 1024 bytes

## Root cause
`cudaq.to_cupy()` wrapped GPU tensors with a fixed 1024-byte `UnownedMemory` region and `owner=None`, even though the returned `cp.ndarray` used the full tensor extents. That could expose only a partial device buffer and leave the CuPy view detached from the lifetime of the originating state.

Fixes #4638.

## Validation
- `python3 -m py_compile python/cudaq/runtime/state.py python/tests/builder/test_cupy_integration.py`
- `pre-commit run --files python/cudaq/runtime/state.py python/tests/builder/test_cupy_integration.py`


khalatepradnya:
> /ok to test b8c7751

python/tests/builder/test_cupy_integration.py
```diff
@@ -99,6 +99,26 @@ def test_state_vector_to_cupy():
...
+    tensor = state.getTensors()[0]
+    expectedSize = tensor.get_num_elements() * tensor.get_element_size()
+
+    stateInCuPy = cudaq.to_cupy(state)[0]
+
+    assert stateInCuPy.data.mem.size == expectedSize
+    assert np.isclose(stateInCuPy[0], 1. / np.sqrt(2.), atol=1e-3)
+    assert np.isclose(stateInCuPy[-1], 1. / np.sqrt(2.), atol=1e-3)
+    assert np.isclose(stateInCuPy[1], 0., atol=1e-3)
+
```
khalatepradnya, line 123: This test suite does not have a pytest fixture. Please reset target at the end of the test.
`cudaq.reset_target()`

khalatepradnya (approved):
Thanks, @fallintoplace for the contribution!
Overall LGTM, one comment.

python/tests/builder/test_cupy_integration.py
fallintoplace: Thank you. Let me check again.

khalatepradnya:
@fallintoplace - please sign your commit. 
Ref: https://github.com/NVIDIA/cuda-quantum/runs/79348169276

mitchdz:
> /ok to test 18515ba

PR #4641, opened 2026-05-30
[Python] Allow parent-scope reassignment of `measure_handle` vectors

khalatepradnya:
### Summary

Support cross-scope reassignment of `list[measure_handle]` in the Python kernel bridge (the surface-code `mvec = m_new` cross-round pattern), and complete the related `measure_handle` storage migration.

### What changed

- `PyASTBridge.storedAsValue`: back a local `!cc.stdvec<!cc.measure_handle>` with a stack slot (like a scalar handle) so parent-scope reassignment stores through the slot. 
- General lists stay value-backed and are still rejected across scopes.

C++ counterpart: issue #4601.


sacpis (approved):
LGTM. Thanks @khalatepradnya.

anpaz (approved):

PR #4642, opened 2026-05-30
[nvqpp] Copy a returned measure_handle vector to the heap

bmhowe23:
Follow-up to #4603 to fix regression found by CUDA-Q Libraries

A `std::vector<measure_handle>` local is lowered to a descriptor stack slot, so it reaches the return statement in pointer form. The return path promotes it to a value with a load but did not update the type it then inspects, so the branch that copies vector contents to the heap on return was skipped. The function returned a descriptor aliasing a buffer that is freed when the callee returns, double-freed by the caller's binding.

Refresh the inspected type after the load for the measure_handle-vector case so the heap copy is emitted. Other pointer-form returns are unchanged.

Adds an execution test covering return-by-value of a measured handle vector.


sacpis (approved):
LGTM. Thanks @bmhowe23.

bmhowe23:
This ran successfully in the cudaqx repo [here](https://github.com/NVIDIA/cudaqx/actions/runs/26701483543/job/78695369676?pr=573), so merging now.

PR #4558, opened 2026-05-20
bump openssl 3.5.1 -> 3.6.2; curl 8.5.0 -> 8.20.0

mitchdz:
- OpenSSL 3.5.1 → 3.6.2 (latest 3.x; stays off 4.0.0 since aws-sdk-cpp@1.11.454's vendored s2n isn't OpenSSL-4-ready)
- curl 8.5.0 → 8.20.0

PR #4643, opened 2026-05-31
Add private linkage to libdl in cudaq-common

bmhowe23:
This is needed to resolve an issue in:

https://github.com/NVIDIA/cudaqx/actions/runs/26721690916/job/78749797379#step:10:10105

FAILED: lib/libcudaq-common.so
: && /opt/rh/gcc-toolset-12/root/usr/bin/g++ -fPIC -fPIC ... -shared -Wl,-soname,libcudaq-common.so ... ld.lld: error: undefined symbol: dlopen

While I didn't bisect this down to the exact commit that seemed to make a change that induces it, I suspect it was #4558. I was definitely somewhere between 67404cea365c0f78a9319f89ef634afca15011c8 and 5a3f44464c3ebfc8af21402b7c1a5985c1db5015.


runtime/common/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ if(OPENSSL_FOUND)
   if(APPLE)
     target_link_libraries(${LIBRARY_NAME} PRIVATE cpr::cpr ZLIB::ZLIB)
   else()
-    target_link_libraries(${LIBRARY_NAME} PRIVATE cpr::cpr -Wl,--start-group ZLIB::ZLIB)
+    target_link_libraries(${LIBRARY_NAME} PRIVATE cpr::cpr dl -Wl,--start-group ZLIB::ZLIB)
```
sacpis: nit: The current change works just to make it more robust in case if anyone added `--as-needed` or the tool chain ships a static `libdl.a`, let's move dl after the `--start-group` and in `${CMAKE_DL_LIBS` instead of a literal.

```
target_link_libraries(${LIBRARY_NAME} PRIVATE cpr::cpr -Wl,--start-group ZLIB::ZLIB ${CMAKE_DL_LIBS})
```

sacpis (approved):
LGTM. Thanks @bmhowe23.

runtime/common/CMakeLists.txt
bmhowe23: Thanks, @sacpis. I verified that that will work as well [here](https://github.com/NVIDIA/cudaqx/actions/runs/26733941685), so I'll make that change and then turn on auto-merge.

PR #4644, opened 2026-06-01
[python] Move JIT caching out of compileModule

lmondada:
JIT caching was currently done within compilation. I'm moving this out so that compilation doesn't need to touch the execution context.


atgeller (approved):
LGTM

PR #4647, opened 2026-06-01
[ci] Auto-delete stale bot/* branches on the daily cleanup schedule

mitchdz:
The "Delete temporary bot/* branches" job was gated on `github.event_name == 'workflow_dispatch' && inputs.delete_temp_branches`, so it only ran on a manual dispatch and was always skipped on the scheduled cleanup runs. Publishing creates bot/<run_id>/<name> staging branches and deletes them when it finishes, but a failed or cancelled publish leaks the branch, so they accumulated indefinitely.

Let the job also run on the daily (0 0 * * *) schedule, but on scheduled runs only delete branches whose tip commit is older than one day. That clears leaked branches automatically while never touching a branch an in-flight publishing run may still be using. A manual dispatch keeps its existing "delete all bot/* branches" behavior.

Also use ${{ github.repository }} for the ref-delete call instead of the hardcoded nvidia/cuda-quantum, matching the branch-listing call.

### DIFF
--- a/.github/workflows/clean_up.yml
+++ b/.github/workflows/clean_up.yml
@@ -23,25 +23,48 @@ name: Clean up
 jobs:
   bot_branches:
     name: Delete temporary bot/* branches
-    if: github.event_name == 'workflow_dispatch' && inputs.delete_temp_branches
+    # Manual dispatch deletes all bot/* branches; the daily schedule deletes only stale ones.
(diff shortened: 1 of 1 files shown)


1tnguyen (approved):
LGTM 👍

schweitzpgi (approved):

Issue #4626, opened 2026-05-28
Add realtime to the CUDAQ_ENABLE_PROJECTS variable


schweitzpgi:
This subproject should be properly selectable from the top-level.

```console
% cmake ${CUDAQ_TOP_LEVEL_SRC} -DCUDAQ_ENABLE_PROJECTS="cudaq;runtime;python;realtime" ...
```

schweitzpgi:
Depends on #4622

PR #4649, opened 2026-06-02
[CMAKE] Add `realtime` to CUDAQ_ENABLE_PROJECTS and update CI

1tnguyen:
Remove the adhoc realtime integration pipeline.

https://github.com/NVIDIA/cuda-quantum/issues/4626


schweitzpgi:
Do you want to split this into separate PRs?
- add realtime as a subproject to cmake (disabled by default)
- add realtime to the CI

realtime/CMakeLists.txt
```diff
@@ -45,71 +53,102 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
...
+      if (CUDAToolkit_VERSION VERSION_LESS 13.0)
+        # Ampere, Hopper
+        set(CUDA_TARGET_ARCHS  "80;90")
+      else()
+        # Ampere, Hopper, Blackwell
+        set(CUDA_TARGET_ARCHS  "80;90;100")
+      endif()
     endif()
+    CUDA_get_gencode_args(CUDA_gencode_flags ${CUDA_TARGET_ARCHS})
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -shared -std=c++17 ${CUDA_gencode_flags} --compiler-options ${_host_compiler_opts}")
```
mitchdz, line 142: Is there a reason we build with c++17 and not c++20 like cudaq?

1tnguyen: Not that I can think of. I'll update it to C++20 as well to be consistent with CUDA-Q.

cudaq/test/CMakeLists.txt
```diff
@@ -81,14 +81,14 @@ if (CUDA_FOUND AND CUDAQ_ENABLE_REALTIME AND NOT CUDAQ_DISABLE_CPP_FRONTEND AND
         cudaq-device-call-runtime
         cudaq-device-call-runtime-headers
         cudaq::cudaq-realtime-dispatch
-        CUDA::cudart)
+        CUDA::cudart_static)
```
schweitzpgi, line 84: Hmm. I'm wondering why we have realtime tests in the core. The dependency seems backward. That is, realtime should depend on core, so tests for realtime+core ought to simply be in realtime itself.

CMakeLists.txt
```diff
@@ -278,6 +283,9 @@ if(NOT CUDAQ_ENABLE_PROJECTS)
   if(CUDAQ_ENABLE_PYTHON)
     list(APPEND CUDAQ_ENABLE_PROJECTS "python")
   endif()
+  if(CMAKE_CUDA_COMPILER)
```
schweitzpgi, line 286: I think this should be disabled by default even if CUDA is installed on the build system. So in order to enable, one has to take an active step of setting `CUDAQ_ENABLE_PROJECTS="...;realtime"` on the command-line *and* CUDA is installed on the build system.

The `CUDAQ_ENABLE_PYTHON` is legacy cruft that needs to be removed at some point as well. And this shouldn't automatically build the `python` project if there is a python on the build machine, but likewise if `CUDAQ_ENABLE_PROJECTS="...;python"` *and* python is installed on the build system.

1tnguyen: Agreed. I'll fix that in the revised PRs that split build system and CI changes.

cudaq/test/CMakeLists.txt
1tnguyen: Realtime is bit different: it's actually intended to be a standalone library that has no dependency on CUDA-Q at all (can be used without CUDA-Q on real quantum control systems). 

CUDA-Q has a runtime component that depends on realtime to support realtime dispatch for local simulation; hence these realtime+core tests are in CUDA-Q.

1tnguyen:
Closing this. It will be split into more targeted PRs.

PR #4650, opened 2026-06-02
[core] Teach compiler when to combine quantum allocations

### LINKED ISSUE #4407: Quantum memory management within CUDA-Q kernels does not work

schweitzpgi:
When quantum allocations come from entirely different scopes, we track those allocations independently right up until the high-level control flow is erased and we lower to a CFG. Subsequently, if there were quake allocations in those scopes that were erased, we do not want to combine them into a single allocation as that may be an oversubscription to resources. Instead, we want to allocation and deallocate qubits as per the user's algorithm. This is now the default.

The former approach, to combine all qubits into a single allocation, is entirely valid semantically. That approach may still be desirable for certain scenarios, which could be rather simply supported by adding a pass option to LowerToCFG.

Fix #4407.

### DIFF
--- a/cudaq/include/cudaq/Optimizer/Builder/Factory.h
+++ b/cudaq/include/cudaq/Optimizer/Builder/Factory.h
@@ -43,6 +43,8 @@ T convertBitsToBytes(T bits) {
   return (bits + 7) / 8;
 }
 
+constexpr const char disableQubitCombineAttrName[] = "cc.no_qubit_combine";
+
 namespace factory {
 
 constexpr const char targetTripleAttrName[] = "llvm.triple";
--- a/cudaq/include/cudaq/Optimizer/Dialect/CC/CCOps.td
+++ b/cudaq/include/cudaq/Optimizer/Dialect/CC/CCOps.td
@@ -96,7 +96,14 @@ def cc_ScopeOp : CCOp<"scope",
     using BodyBuilderFn =
         llvm::function_ref<void(mlir::OpBuilder &, mlir::Location)>;
 
-    bool hasAllocation(bool quantumAllocs = true);
+    // Determine if the scope has any allocation.
+    bool hasAllocation();
+
+    // Determine if the scope has a classical allocation, ignoring quantum.
+    bool hasClassicalAllocation();
+    
+    // Determine if the scope has a quantum allocation, ignoring classical.
+    bool hasQuantumAllocation();
   }];
 }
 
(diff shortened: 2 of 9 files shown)


1tnguyen (approved):
LGTM 👍

atgeller (approved):
:thumbsup:

PR #4651, opened 2026-06-02
Fix for adaptive QIR

sacpis:
Before this change, a `__qpu__` function that returned a `std::vector<bool>` of size > 1 broke adaptive QIR. The vector was built element-by-element (a store loop) and, after inlining, the frontend's vector copy-ctor idiom was left in the IR. PR #4455 emptied the inliner's per-step canonicalizer to preserve inlined scopes, so this dead heap buffer was no longer folded on the codegen path. Its `store`s survived into adaptive QIR, which bans heap stores, aborting JIT translation.

```
QIR verification error - invalid instruction found:
    store i8 %X, ptr %Y, align 1 (adaptive profile)
```

The fix is to run the existing `erase-vector-copy-ctor` pass right after `return-to-output-log` in `createTargetCodegenPipeline`. The copy-ctor idiom only reaches matchable form after QIR conversion and lower-to-cfg, which run after the inliner's own run of the pass, so the earlier invocation can't catch it.

@bmhowe23 for viz.

### DIFF
--- a/cudaq/test/Transforms/erase_vcc_adaptive_vector_bool.qke
+++ b/cudaq/test/Transforms/erase_vcc_adaptive_vector_bool.qke
@@ -1,0 +1,118 @@
+// ========================================================================== //
+// Copyright (c) 2026 NVIDIA Corporation & Affiliates.                        //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --return-to-output-log --erase-vector-copy-ctor %s | FileCheck %s
+
+module attributes {quake.mangled_name_map = {__nvqpp__mlirgen__function_repro._Z5reprov.run = "__nvqpp__mlirgen__function_repro._Z5reprov.run.entry"}} {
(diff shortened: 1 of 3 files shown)


cudaq/test/Transforms/erase_vcc_adaptive_vector_bool.qke
```diff
@@ -0,0 +1,118 @@
+// ========================================================================== //
+// Copyright (c) 2026 NVIDIA Corporation & Affiliates.                        //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt --return-to-output-log --erase-vector-copy-ctor %s | FileCheck %s
```
khalatepradnya, line 9: Q: Can we write a test to test the whole pipeline? As in the code change is to automatically add the `--erase-vector-copy-ctor` in the pipeline so, you don't need to invoke it explicitly?

schweitzpgi (approved):
This will work as a quick fix. I'll work on the affected passes in the meantime.

cudaq/test/Transforms/erase_vcc_adaptive_vector_bool.qke
sacpis: Added in 6280221.

khalatepradnya (approved):
New tests look good!

PR #4652, opened 2026-06-02
Add CPU-initiated RoCE transport (CpuRoceTransceiver) + HOST_CALL dispatch

cketcham2333:
Adds a low-latency, CPU-initiated RoCEv2 transport that lands data in CPU
memory (vs. the existing GpuRoceTransceiver, which targets GPU memory),
plus the `CUDAQ_DISPATCH_HOST_CALL` path needed to drive it without a GPU.

### DIFF
--- a/realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
+++ b/realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
@@ -52,8 +52,16 @@ typedef enum {
 } cudaq_kernel_type_t;
 
 // Dispatch invocation mode.
-// For CUDAQ_DISPATCH_PATH_HOST only GRAPH_LAUNCH is dispatched; DEVICE_CALL and
-// HOST_CALL table entries are dropped (slot cleared and advanced).
+//
+// On CUDAQ_DISPATCH_PATH_HOST:
+//   - GRAPH_LAUNCH: launched on a graph worker (worker pool, cudaGraphLaunch).
+//   - HOST_CALL:    invoked synchronously inline as a plain C++ function via
+//                   entry->handler.host_fn.  Bypasses the worker pool.  No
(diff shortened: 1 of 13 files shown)


realtime/lib/cpu_transport/roce_transceiver.cpp
```diff
@@ -0,0 +1,1118 @@
...
+  // safety net for the destructor path where close() is the only thing
+  // that runs cleanup.
+  if (impl_->forward_thread.joinable())
+    impl_->forward_thread.join();
+  if (impl_->unified_thread.joinable())
+    impl_->unified_thread.join();
+  if (impl_->rx_thread.joinable())
+    impl_->rx_thread.join();
+  if (impl_->tx_thread.joinable())
+    impl_->tx_thread.join();
```
1tnguyen: These `join()` might be overlapping with those in lines 1031-1038. 
I wonder whether we should move these to a shared helper protected with mutex to guarantee we only call it once.

1tnguyen (approved):
LGTM 👍 Thanks @cketcham2333

realtime/lib/cpu_transport/roce_transceiver.cpp
cketcham2333: Resolved in commit #a5756f1052ea982d13e9c200b4367000a9701855

realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h
bmhowe23, line 1: I suspect this needs docs updates, too. If so, where is that being tracked?

cketcham2333: Agreed that doc updates are needed.  I can do that in a follow up PR.

PR #4654, opened 2026-06-02
[Python] Improved handling of synthesized and lifted arguments

atgeller:
Replaces adhoc logic with `isFullySynthesized` function to determine if a function has had it's arguments synthesized or not (currently `synthesize` is all-or-nothing). Adds logic to prevent caching kernels with lifted callables.

### DIFF
--- a/cudaq/include/cudaq/Optimizer/Builder/Marshal.h
+++ b/cudaq/include/cudaq/Optimizer/Builder/Marshal.h
@@ -28,6 +28,18 @@ inline bool isCodegenArgumentGather(std::size_t kind) {
   return kind == 0 || kind == 2;
 }
 
+/// A kernel is fully synthesized when it has at least one formal argument
+/// and every formal argument has been folded into the body (no uses remain).
+inline bool isFullySynthesized(mlir::func::FuncOp funcOp) {
+  auto args = funcOp.getArguments();
+  if (args.empty())
+    return false;
+  for (mlir::Value arg : args)
+    if (!arg.use_empty())
+      return false;
+  return true;
+}
+
 inline bool isStateType(mlir::Type ty) {
   if (auto ptrTy = dyn_cast<cc::PointerType>(ty))
     return isa<cudaq::quake::StateType>(ptrTy.getElementType());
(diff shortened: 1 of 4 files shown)


sacpis (approved):
LGTM. Thanks @atgeller.

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
```diff
@@ -684,17 +684,15 @@ pyLaunchModule(const std::string &name, ModuleOp mod,
...
     if (!platform.is_simulator() || platform.is_emulated())
       return false;
-    // TODO: would be nice to check for args synthesized via an attribute
-    // rather than a fragile arg size check. This relies on an assumption
-    // that arg sizes are validated so the only case that the arg sizes
-    // don't match is for synthesized kernels.
     auto func = cudaq::getKernelFuncOp(mod, name);
-    mlir::Type resultTy = cudaq::runtime::getReturnType(func);
-    auto hasResult = !!resultTy;
-    size_t numArgs = rawArgs.size() - (hasResult ? 1 : 0);
```
lmondada, line 694: Can you also make the same change in `runtime/internal/compiler/Compiler.cpp`?

lmondada (approved):
Awesome, thanks a lot!

python/runtime/cudaq/platform/py_alt_launch_kernel.cpp
atgeller: It's made in `QPU.cpp` here. It will have to be ported for #4636

cudaq/include/cudaq/Optimizer/Builder/Marshal.h
```diff
@@ -28,6 +28,18 @@ inline bool isCodegenArgumentGather(std::size_t kind) {
   return kind == 0 || kind == 2;
 }
 
+/// A kernel is fully synthesized when it has at least one formal argument
+/// and every formal argument has been folded into the body (no uses remain).
+inline bool isFullySynthesized(mlir::func::FuncOp funcOp) {
```
schweitzpgi, line 33: This seems a bit of an odd place to put this as it isn't used in Marshal.cpp. Consider placing it in Factory.h perhaps?

atgeller: I'll make a follow up PR. I put it there because that's where I remembered various functions used in `py_alt_launch_kernel.cpp` being placed.

PR #4660, opened 2026-06-03
[python] Remove PythonLauncher and move compilation out of QPU

lmondada:
In this PR, we finally pull compilation out of the QPUs (for Python). This means that we can remove the `PythonLauncher` "hack" entirely (a plugin system loaded at runtime, only ever used with a single plugin, whose purpose was to decouple the base QPU class from MLIR).

We are one step closer to (and one step away from) freeing the QPU classes of their MLIR dependency! The last remaining MLIR dependency is within BaseRemoteRESTQPU::launchKernel, which is currently required to JIT compile kernels that were launched from C++.

### DIFF
--- a/runtime/cudaq/platform/default/DefaultQPU.cpp
+++ b/runtime/cudaq/platform/default/DefaultQPU.cpp
@@ -76,20 +76,23 @@ cudaq::DefaultQPU::launchKernel(async_observe_policy &policy,
 
 std::unique_ptr<cudaq::CompileTarget>
 cudaq::DefaultQPU::getCompileTarget(const sample_policy &policy) {
-  // A python-only target suffices, as C++ compilation is done AOT
+  // Currently this is only used for Python kernels, as C++ kernels skip JIT
+  // compilation and call the AOT-generated function directly.
   return getDefaultPythonCompileTarget(policy);
 }
 
 std::unique_ptr<cudaq::CompileTarget>
 cudaq::DefaultQPU::getCompileTarget(const observe_policy &policy) {
-  // A python-only target suffices, as C++ compilation is done AOT
+  // Currently this is only used for Python kernels, as C++ kernels skip JIT
+  // compilation and call the AOT-generated function directly.
   return getDefaultPythonCompileTarget(policy);
 }
 
 std::unique_ptr<cudaq::CompileTarget>
 cudaq::DefaultQPU::getCompileTarget(const other_policies &policy,
                                     ExecutionContext *context) {
-  // A python-only target suffices, as C++ compilation is done AOT
+  // Currently this is only used for Python kernels, as C++ kernels skip JIT
+  // compilation and call the AOT-generated function directly.
   return getDefaultPythonCompileTarget(policy, context);
 }
 
--- a/runtime/cudaq/platform/default/DefaultQPU.cpp
+++ b/runtime/cudaq/platform/default/DefaultQPU.cpp
@@ -10,6 +10,7 @@
 #include "common/ExecutionContext.h"
 #include "common/Timing.h"
 #include "cudaq/algorithms/policies.h"
+#include "cudaq/platform.h"
 #include "cudaq/runtime/logger/logger.h"
 
 cudaq::DefaultQPU::~DefaultQPU() = default;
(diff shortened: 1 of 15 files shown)


runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -227,23 +281,28 @@ quantum_platform::unifiedLaunchModule(const AnyModule &module, KernelArgs args,
...
 
-  return cudaq::policies::withPolicy(ctx->name, [&](auto policy) {
-    using Policy = std::decay_t<decltype(policy)>;
-    if constexpr (std::is_same_v<Policy, observe_policy>) {
-      policy.spin = ctx->spin.value();
-    }
-    return qpu->compileModule(policy, src, args, isEntryPoint);
-  });
+std::unique_ptr<cudaq::CompileTarget>
+quantum_platform::getCompileTarget(const sample_policy &policy,
```
Renaud-K: Can these two be templatized on the policy ?

runtime/cudaq/platform/default/DefaultQPU.cpp
```diff
@@ -73,6 +74,25 @@ cudaq::DefaultQPU::launchKernel(async_observe_policy &policy,
       "DefaultQPU does not support launching the async_observe_policy.");
 }
 
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::DefaultQPU::getCompileTarget(const sample_policy &policy) {
+  // A python-only target suffices, as C++ compilation is done AOT
```
Renaud-K: I am not sure what it means. This is a C++ and Python flow right? It sounds like it is python only.

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
```diff
@@ -134,4 +135,26 @@ class Compiler {
 /// stages. Pass empty strings to skip them.
 std::string getPassPipeline(const cudaq::CompileTarget &target);
 
+/// Compile a source module for the given policy, compile target and
+/// arguments.
+template <typename Policy>
+cudaq::CompiledModule
+compileModule(Policy *policy, std::unique_ptr<cudaq::CompileTarget> &&target,
```
Renaud-K: The unique_ptr should be passed by value. We do not need to combine unique_ptr and r-values to communicate the transfer of ownership.

runtime/cudaq/platform/default/DefaultQPU.cpp
lmondada: For `DefaultQPU` (and derived classes), the execution of C++ kernels always skips JIT compilation (it uses the function pointer compiled AOT). Thus `getCompileTarget` is currently not called in that flow.

This would change once JIT compilation for C++ is done outside of launch.

Could I rephrase the comment in any way to make this clearer?

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
lmondada: ok, will do!

runtime/cudaq/platform/quantum_platform.cpp
lmondada: Moving this to the `quantum_platform.h` header file requires an additional include for `#include "cudaq/platform/qpu.h"`. Let me know if this is a problem, I can revert the change.

runtime/cudaq/platform/default/DefaultQPU.cpp
lmondada: I've changed the phrasing to

```
  // Currently this is only used for Python kernels, as C++ kernels skip JIT
  // compilation and call the AOT-generated function directly.
```

Is that clearer?

Renaud-K: Ok. I was not sure where this was called from. So this is never called in a C++ flow. This makes a bit more sense.

runtime/cudaq/platform/quantum_platform.cpp
Renaud-K: We include `qpu.h` in `launch.h` now. `qpu.h` is visible from user code already. So no, this should not be a problem. And since these are not virtual, there is no need to introduce a `getCompileTargetImpl`.

Renaud-K (approved):
LGTM. Thank you!

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -90,6 +91,59 @@ void quantum_platform::reset_noise(std::size_t qpu_id) {
...
 }
 
+static std::unique_ptr<cudaq::CompileTarget>
+getDefaultPythonCompileTargetImpl() {
+  auto *platform = getQuantumPlatformInternal();
+
+  const bool enablePythonCodegenDump =
+      cudaq::getEnvBool("CUDAQ_PYTHON_CODEGEN_DUMP", false);
+  if (enablePythonCodegenDump) {
+    CUDAQ_WARN("CUDAQ_PYTHON_CODEGEN_DUMP is no longer supported and will be "
```
schweitzpgi, line 101: This isn't really an acceptable change. I want to be able to dump the codegen pipeline in isolation and not have  it gratuitously thrown into a general tidal wave of IR printed twice for every pass.

lmondada: Got it, will happily revert.

Just a note that this flag did not dump the IR just once: it printed it between every pass, but only for the codegen part of compilation. Is that the behaviour you want?

lmondada: See old code here: https://github.com/NVIDIA/cuda-quantum/blob/65f501c27a561b4fb65ab222dff92de974898227/runtime/cudaq/platform/default/python/QPU.cpp#L95

lmondada: Resolved in https://github.com/NVIDIA/cuda-quantum/pull/4766

PR #4661, opened 2026-06-03
[core] Update the erase-vector-copy-ctor pass.

schweitzpgi:
This pass needs a new pattern to be better aligned with the new inlining scope preservation in the compiler. When returning a span, a kernel necessarily muct move the data off the stack. But after inlining a call to such a function, the span can be retained on the calling function's stack trivially and without a copy. This adds a new rewrite pattern to deal with the ScopeOp, etc.

Rename the test to better reflect the regression.


sacpis (approved):
LGTM. Thanks @schweitzpgi.

Issue #3509, opened 2025-10-11
observe has unexpected behavior with an if statement and mz


mmvandieren:
### Required prerequisites

- [x] Consult the [security policy](https://github.com/NVIDIA/cuda-quantum/security/policy). If reporting a security vulnerability, do not report the bug using this form. Use the process described in the policy to report the issue.
- [x] Make sure you've read the [documentation](https://nvidia.github.io/cuda-quantum/latest). Your issue may be addressed there.
- [x] Search the [issue tracker](https://github.com/NVIDIA/cuda-quantum/issues) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] If possible, make a PR with a failing test to give us a starting point to work on!

### Describe the bug

The `mz` gives an error message if it is in an unexecuted if statement when using `observe`. However, if instead the kernel refers to that `mz` command by calling a sub kernel then no error message is flagged.  See the working and broken code examples.

use case:  the if statement was prudent if I wanted to have just one kernel that I needed to sample a subset of qubits, but I also wanted to compute expectation values for.  for instance in a variational quantum random walk with auxiliary qubits

### Steps to reproduce the bug

import cudaq

@cudaq.kernel()
def broken_kernel(with_measure : int):
    q=cudaq.qvector(2)
    if with_measure == 1:
        mz(q)

hamiltonian = cudaq.spin.z(0) + cudaq.spin.z(1)

#the next line breaks the code
result = cudaq.observe(broken_kernel, hamiltonian, 0).expectation()

#but if we nest the mz command into a separate kernel, we don't get an error message
@cudaq.kernel()
def measure(qubits: cudaq.qview):
    mz(qubits)
    
@cudaq.kernel
def working_kernel(with_measure : int):
    q=cudaq.qvector(2)
    if with_measure == 1:
        measure(q)

result = cudaq.observe(working_kernel, hamiltonian, 0).expectation()

### Expected behavior

expected that both versions of the kernel would either both give errors or both work ok (both working ok would be preferred)

### Is this a regression? If it is, put the last known working version (or commit) here.

Not a regression

### Environment

- **CUDA-Q version**: cu12-0.10.0
- **Python version**: 3.10.12
- **C++ compiler**: 
- **Operating system**: Mac OS

mmvandieren:
oops, here's the code properly formatted
```import cudaq

@cudaq.kernel()
def broken_kernel(with_measure : int):
    q=cudaq.qvector(2)
[...]

PR #4662, opened 2026-06-03
Implementing observe_policy end-to-end

Renaud-K:
This is a similar change as #4557 albeit smaller since it reuse the sample_policy infrastructure. 
There are still some duplication between the policy and the execution context. It will take more policies to be implemented to remove fields from the executionContext. 

Maybe the more contentious point is that library mode needs to be handled differently with policies. The policies require the result to be extracted earlier in the flow on the QPU side. Library mode, however, never gets to the QPU side. This needs to be known at the time of the launch to avoid relying on the executeKernelApi. In the case of `evolve`, which uses `observe` in library mode under the hood, an RAII class is used ` details::with_platform_in_library_mode` to temporarily turn on library mode. We can discuss ways of knowing library mode statically so we do not have to use this setting.


sacpis:
Seems like we have a test failure.
https://github.com/NVIDIA/cuda-quantum/actions/runs/26898046741/job/79343843693?pr=4662#step:14:7443

Renaud-K:
Sorry, about the CI failures. I did test these changes thoroughly but I might have added some some seemingly benign last minute changes that have fallen through the cracks. I will take care of those now.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -424,71 +443,35 @@ class BaseRemoteRESTQPU : public QPU {
                     ? executionContext->hasConditionalsOnMeasureResults
                     : false;
 
-            if (hasConditionals && isObserve)
+            if (hasConditionals)
               throw std::runtime_error("error: spin_ops not yet supported with "
                                        "kernels containing conditionals");
```
sacpis: Do we need this condition anymore?

python/cudaq/runtime/observe.py
sacpis: Let's delete this function.

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -579,6 +563,65 @@ class BaseRemoteRESTQPU : public QPU {
...
+
+    executor->setShots(localShots);
+    assert(emulate);
+
+    std::size_t seed = cudaq::get_random_seed();
+    if (seed > 0)
+      cudaq::set_random_seed(seed);
+
+    std::vector<cudaq::ExecutionResult> results;
+    auto reorderIdx = policy.reorderIdx;
```
sacpis: Seems like we are not using `reorderIdx`.

runtime/cudaq/algorithms/observe.h
```diff
@@ -73,41 +72,49 @@ runObservation(KernelFunctor &&k, const cudaq::spin_op &H,
...
-    expectationValue = sum;
+/// @brief Take the input KernelFunctor (a lambda that captures runtime
+/// arguments and invokes the quantum kernel) and invoke the `spin_op`
+/// observation process.
+template <typename KernelFunctor>
+observe_result
+runObservation(KernelFunctor &&k, const cudaq::spin_op &H,
+               quantum_platform &platform, int shots,
+               const std::string &kernelName, std::size_t qpu_id = 0,
+               details::future *futureResult = nullptr,
```
sacpis: Let's drop `futureResult` as it is not being used inside the function.

runtime/cudaq/algorithms/observe.h
```diff
@@ -50,18 +51,16 @@ concept ObserveCallValid =
...
-std::optional<observe_result>
-runObservation(KernelFunctor &&k, const cudaq::spin_op &H,
-               quantum_platform &platform, int shots,
-               const std::string &kernelName, std::size_t qpu_id = 0,
-               details::future *futureResult = nullptr,
-               std::size_t batchIteration = 0, std::size_t totalBatchIters = 0,
-               std::optional<std::size_t> numTrajectories = {}) {
-  ExecutionContext ctx("observe", shots, qpu_id);
+void observePreamble(ExecutionContext &ctx, observe_policy &policy,
+                     KernelFunctor &&, quantum_platform &platform,
```
sacpis: Let's drop KernelFunctor as we are not using it.

sacpis (commented):
Left a few comments with the initial review.

runtime/common/BaseRemoteRESTQPU.h
Renaud-K: I fixed the other changes. They were nice clean-ups. This one highlights the fact that there is no test coverage for this exception. I am going to have to add a test and move the message. Good catch.

schweitzpgi: While we're fixing this can we tweak the message if needed? "Conditionals" just screams vagueness and wrongness to me. It's almost like saying "kernels containing code" as far as how imprecise it is.

All of the following example kernels, except the last one (`qernel4`), are conditionals.
```c++
__qpu__ void qernel1(cudaq::qview<> qs) {
   x<cudaq::ctrl>(qs[0], qs[1]);
   ...
 }
 
 __qpu__ void qernel2(bool flag) {
   if (flag) ...
 }

__qpu__ void qernel3(cudaq::qview<> qs) {
  bool flag = mz(qs[0]);
  if (flag) ....
}

__qpu__ void qernel4(double angle, cudaq::qview<> qs) {
  bool _flag = mz(qs[0]);
  int flag = _flag;
  rz(angle * (flag + 1), qs[1]);
  ...
}
```

I suspect this diagnostic message actually means the cases in `qernel3` and `qernel4` despite the way it reads, which implies `qernel1`, `qernel2` and `qernel3`.

We shouldn't assume users are totally familiar with tribal jargon and shorthand inaccuracies.

Renaud-K:
Running CI experimentally without the nvcc fixes. 
/ok to test cc354d8b687f5193de90d378b9894385a73cfaf7

Renaud-K:
> /ok to test cc354d8b687f5193de90d378b9894385a73cfaf7

runtime/common/BaseRemoteRESTQPU.h
Renaud-K: I have removed this message because it is unreachable.
I added 2 tests. In the python flow we error sooner (for which there is no test either). In the C++ flow, there is an earlier exception throw which had no coverage either.

Renaud-K:
> /ok to test 6c311574c78fbede4ce0829e1059a78c3af71e0d

python/tests/kernel/test_observe_measures_error.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_observe_kernel_with_measures_fails():
+    kernel = cudaq.make_kernel()
+    q = kernel.qalloc(1)
+    kernel.h(q[0])
+    kernel.mz(q[0])
+
+    with pytest.raises(
+            RuntimeError,
+            match=r"kernels passed to observe cannot have measurements specified"
```
khalatepradnya, line 22: I think you fixed at least one issue in https://github.com/NVIDIA/cuda-quantum/issues/3509

Renaud-K: Thank you but I don't think so, I have not fixed anything, just added coverage. This message is still only issued in python.

runtime/cudaq/algorithms/observe/policy.h
```diff
@@ -27,10 +29,30 @@ struct observe_policy {
...
 
+  /// @brief The name of the kernel being executed.
+  std::string kernelName;
+
+  /// @brief The spin operator being observed.
+  spin_op spin;
+
+  /// @brief Flag to indicate that a warning about named measurement registers
+  /// in sampling context has already been emitted.
+  bool warnedNamedMeasurements = false;
```
lmondada: AFAIK this flag is only used in sample. It shouldn't be required here?

runtime/cudaq/platform/qpu.cpp
```diff
@@ -96,24 +112,37 @@ cudaq::QPU::runJITCompiledModule(const CompiledModule &compiled,
...
 std::unique_ptr<cudaq::CompileTarget>
-cudaq::QPU::getCompileTarget(sample_policy &) {
+cudaq::QPU::getCompileTarget(const sample_policy &) {
+  throw std::runtime_error(
+      "no CompileTarget defined for sample_policy this QPU");
+}
+std::unique_ptr<cudaq::CompileTarget>
+cudaq::QPU::getCompileTarget(const observe_policy &) {
   throw std::runtime_error(
       "no CompileTarget defined for sample_policy this QPU");
```
lmondada:
```suggestion
      "no CompileTarget defined for observe_policy this QPU");
```

runtime/cudaq/platform/qpu.h
```diff
@@ -144,28 +140,44 @@ class QPU : public registry::RegisteredType<QPU> {
...
   /// Get the compile target of the QPU
   // Overload for sample policy
   [[nodiscard]] virtual std::unique_ptr<CompileTarget>
-  getCompileTarget(sample_policy &policy);
+  getCompileTarget(const sample_policy &policy);
+  [[nodiscard]] virtual std::unique_ptr<CompileTarget>
+  getCompileTarget(const observe_policy &policy);
   // Overload for currently unsupported policies (to be removed).
   [[nodiscard]] virtual std::unique_ptr<CompileTarget>
   getCompileTarget(ExecutionContext *context);
```
lmondada: Could you mirror the change you made to `compileModule` here, i.e. change the argument to `other_policies`? I suppose within the other_policies overload we can then call `getExecutionContext`.

This will make it easier to write the unified `compileModule(const Policy &p, ...)` function that just forwards the policy to `getCompileTarget`.

runtime/cudaq/platform/quantum_platform.h
```diff
@@ -36,7 +36,8 @@ class LinkedLibraryHolder;
 
 namespace detail {
 class TargetSetter;
-}
+class with_platform_in_library_mode;
```
lmondada, line 39: I am perpetually confused about our naming conventions: given that this is within `detail` and hence not public-facing, shouldn't this use LLVM's CamelCase style?

runtime/cudaq/qis/execution_manager.cpp
```diff
@@ -50,6 +51,12 @@ void ExecutionManager::configureExecutionContext(const sample_policy &policy) {
   nvqir::getCircuitSimulatorInternal()->configureExecutionContext(policy);
 }
 
+void ExecutionManager::configureExecutionContext(const observe_policy &policy) {
+  if (auto *ctx = getExecutionContext())
```
lmondada, line 55: is this override because of the noise model?

runtime/cudaq/qis/execution_manager.h
```diff
@@ -119,18 +119,20 @@ class ExecutionManager {
 
   /// Configure the execution context before an execution.
   void configureExecutionContext(const sample_policy &policy);
+  void configureExecutionContext(const observe_policy &policy);
```
lmondada, line 122: Probably not as part of this PR, but `configureExecutionContext` seems like a wrong name at this point. Maybe `configurePolicy`?

The same comment applies to `finalizeExecutionContext` as well.

lmondada: This thought also made me wonder why the policies are passed by const reference. Presumably they will typically get modified within this configure step? I see you marked some fields as `mutable` in `runtime/cudaq/algorithms/observe/policy.h` for this reason. How are you thinking about this?

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -234,11 +234,16 @@ CompiledModule quantum_platform::compileModule(const SourceModule &src,
...
-  if (ctx && ctx->name == "sample") {
-    sample_policy policy;
+  if (!ctx)
+    return qpu->compileModule(other_policies{}, src, args, isEntryPoint);
+
+  return cudaq::policies::withPolicy(ctx->name, [&](auto policy) {
+    using Policy = std::decay_t<decltype(policy)>;
+    if constexpr (std::is_same_v<Policy, observe_policy>) {
+      policy.spin = ctx->spin.value();
+    }
```
lmondada, line 244: I see, so for compilation we have to construct another policy object from scratch. This is fine, although this means that any mutation of the policy object in `compileModule` will not propagate to the launch. We will need another way to propagate e.g. the `warnedNamedMeasurements` flag.

We also don't get the logic of e.g. `observePreamble` for free, and thus here we need to manually propagate the spin from the ExecutionContext to the policy. Could we expose some factory functions for each of the policy objects, so that withPolicy and `observePreamble` (resp. `samplePreamble`) share the same logic whenever policy objects are needed?

runtime/cudaq/platform/quantum_platform.cpp
```diff
@@ -234,11 +234,16 @@ CompiledModule quantum_platform::compileModule(const SourceModule &src,
   validateQpuId(qpu_id);
   auto &qpu = platformQPUs[qpu_id];
   auto ctx = cudaq::getExecutionContext();
-  if (ctx && ctx->name == "sample") {
-    sample_policy policy;
+  if (!ctx)
+    return qpu->compileModule(other_policies{}, src, args, isEntryPoint);
+
+  return cudaq::policies::withPolicy(ctx->name, [&](auto policy) {
```
lmondada, line 240: Heads up that in #4660 I will need to move this dispatching logic one level up in the call stack, i.e. in `streamlinedCompileModule`. You can leave this as is for now, though.

lmondada (commented):
Thank you, making great strides in the refactor 🔥

I'll base my work for my next PR (#4660) on top of this one.

runtime/cudaq/algorithms/observe/policy.h
Renaud-K: ok.

runtime/cudaq/platform/quantum_platform.h
Renaud-K: The only enforced style is the snake case on the public interface. We do not enforce any other. We say this: https://github.com/Renaud-K/cuda-quantum/blob/observe_policy/CppAPICodingStyle.md?plain=1#L179

I followed this style.

```
  auto with_execution_context(ExecutionContext &ctx, Callable &&f,
                              Args &&...args) {
    // Save the outer execution context (if any) so we can restore it after.
    auto *outerContext = getExecutionContext();

    configureExecutionContext(ctx);
```

But ultimately we will have to work on consistency. Different styles within one class's own interface should not be allowed.

runtime/cudaq/qis/execution_manager.cpp
Renaud-K: Wow, how did that get in there? Good catch. No, should not be needed.

runtime/cudaq/qis/execution_manager.h
Renaud-K: My thinking is that they should be mostly constant. The add-on information should be minimum.

runtime/cudaq/platform/quantum_platform.cpp
Renaud-K: I would much rather expose the issue and make it difficult to create a policy from scratch rather than make it so easy that it becomes the preferred path. The policy has to surface to the top.

runtime/cudaq/platform/qpu.h
Renaud-K: I can do that.

Renaud-K:
> /ok to test 31f5fcab65bcfa896b79bfb0a15df8ebee134878

runtime/cudaq/platform/quantum_platform.h
schweitzpgi: Historically, we do have a style.

The CUDA-Q user API is snake case to be like the STL. Everything else in our code base should follow the LLVM (well, more precisely MLIR's flavor of LLVM) coding guidelines. We do not want "everyone invent their own coding style" on the project.

lmondada: @schweitzpgi the confusion comes from the fact that is not always clear what is and is not in the user API.

I had introduced `with_execution_context` with snake case because it lives within the `quantum_platform.h` file, which is a (very) user-facing header file. Do we expect users to use that directly? Probably not.

@Renaud-K My comment came from the fact that you have hidden the declaration within a `detail` namespace (which I think is the right call btw, arguably `with_execution_context` also belongs there). This puts it firmly outside of the user API, so CamelCase is probably the right call.

Going forward, we might want to hide a lot more declarations within `detail` namespaces and follow the LLVM style guide accordingly. This still doesn't solve the question of which naming convention to adopt for methods within a same class, where only a subset of methods should be public (e.g. the `quantum_platform` class).

runtime/cudaq/qis/execution_manager.h
lmondada: Ok, understood. I agree this is the better design.

However, in this case the policy is the wrong object to pass information (metadata) from compilation to runtime. This is an additional argument to move `sample_policy::warnedNamedMeasurements` out of the policy.

runtime/cudaq/platform/quantum_platform.cpp
lmondada: Ok, same as my response above: I agree this is the nicer design, but means we should not use the policy object as a mule to propagate data from compiler to runtime.

runtime/cudaq/platform/quantum_platform.h
Renaud-K: Say a developer designs a new public header, following the snake_case style, as required. I don't believe that it will be reasonable to ask of this developer to switch to CamelCase as soon as he/she introduces a detail namespace in the header. I don't think that we would have much credibility enforcing such a rule. 

Since we want also CamelCase, then there will unavoidably be areas where the styles are mixed.

runtime/cudaq/qis/execution_manager.h
Renaud-K: I am not against moving `warnedNamedMeasurements` out of the policy. We can do that when we move it out of the execution context.

runtime/cudaq/platform/quantum_platform.h
schweitzpgi: I totally agree that there are grey areas with this. We may want to be a bit more proactive and push `cudaq::detail` namespace artifacts into headers in separate folders (like the `blah_internals` folder(s) that were added to separate public API from internal support code previously. (Thinking out loud.) But, yeah, ultimately some things may not clearly fall in either bucket.

I think that doesn't change the general direction though. We want to choose one of the two styles throughout, and not have every commit use its own custom style on an individual  directory or file basis. :)

Renaud-K: What is the conclusion here? Are we saying that detail namespaces must be CamelCase moving forward? That would be a big decision to make. It would make at odds a lot of our runtime.

lmondada: I mean, one additional inconsistency isn't a deal breaker. Happy to approve as-is.

I continue to believe that our current "rule" isn't tenable, as evidenced by this discussion: if we apply it strictly, we end up with different styles within the same file and even within the same class. That's not nice to work with and impossible to enforce.

Eric's suggestion of moving non-public API (e.g. stuff within `detail` namespaces) to separate files might be a good resolution. To explore further.

lmondada (approved):

PR #4665, opened 2026-06-04
ci: verify installer integrity and re-fetch on corrupt download

mitchdz:
The makeself self-extracting installer is downloaded per validate job via download-artifact, then run with --accept. A transient truncated download surfaces as a cryptic `Verifying archive integrity... Unexpected archive size` failure (the file size no longer matches the makeself header), which intermittently fails the merge queue even though the built artifact is good.

Verify the installer with `makeself --check` right after download and, on failure, re-fetch the artifact (with retry) before continuing. Applied to both installer-run sites: prebuilt_binaries.yml (validate) and publishing.yml (sanity checks).

Fixes issues such as https://github.com/NVIDIA/cuda-quantum/actions/runs/26913098462/job/79403999184 and https://github.com/NVIDIA/cuda-quantum/actions/runs/26913098462/job/79403999184


1tnguyen (approved):
LGTM 👍

schweitzpgi (approved):

PR #4669, opened 2026-06-04
[core] Add dead quantum elimination (DQE) pass

schweitzpgi:
Introduce a dedicated pass to eliminate qubits that are unused from the IR. As unused qubits can increase time and space demands in simulation, we generally want to get rid of them if they aren't used in the kernel.

This pass can be expanded in the future to cover other quake types and cases.

### DIFF
--- a/cudaq/lib/Optimizer/Transforms/DeadQuantumElimination.cpp
+++ b/cudaq/lib/Optimizer/Transforms/DeadQuantumElimination.cpp
@@ -1,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "PassDetails.h"
+#include "cudaq/Optimizer/Transforms/Passes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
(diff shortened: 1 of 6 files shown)


schweitzpgi:
@sacpis Contains the build dependence fix.

cudaq/lib/Optimizer/Transforms/DeadQuantumElimination.cpp
```diff
@@ -0,0 +1,84 @@
...
+};
+
+class WirePattern : public OpRewritePattern<cudaq::quake::NullWireOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(cudaq::quake::NullWireOp nullWire,
+                                PatternRewriter &rewriter) const override {
+    // Wires are linear types. There must be exactly 1 use.
+    auto sink = dyn_cast<cudaq::quake::SinkOp>(*nullWire->getUsers().begin());
```
sacpis, line 60: nit: Should we add a check here for `nullWire->hasOneUse()`?
```
if (!nullWire->hasOneUse())
    return failure();
```

sacpis (approved):
LGTM. Thanks @schweitzpgi.

cudaq/lib/Optimizer/Transforms/DeadQuantumElimination.cpp
schweitzpgi: That property is maintained by the IR itself (see the comment). Re-checking it here is not necessary and inefficient.

sacpis: Thanks @schweitzpgi. Makes sense.

PR #4674, opened 2026-06-05
Fix no-observable evolve paths for intermediate states and collapse noise

huaweil-nv:
## Summary

This fixes several broken no-observable paths in the non-`dynamics` simulator implementation of `cudaq.evolve` and `cudaq.evolve_async`.

The affected paths are triggered when all of the following are true:

- The target uses the non-`dynamics` simulator implementation, for example `qpp-cpu` or `density-matrix-cpu`.
- The caller does not provide `observables`.
- The caller requests intermediate results, provides `collapse_operators`, or uses the async API.

The main user-visible failure was that sync intermediate-state evolution without observables failed before returning states:

```text
TypeError: evolve(): incompatible function arguments
Invoked with types: State, list, bool
```
There were also related no-observable correctness issues: collapse operators could be ignored when no observables were requested, async intermediate-state evolution relied on the binding default for save_intermediate_states, and no-observable result semantics needed to remain None for expectation values.

## Root cause
evolve_single and evolve_single_async have separate early-return branches for the case where observables is empty. Those branches had drifted away from the observable branches.

For sync intermediate-state evolution, the no-observable branch called:
```
cudaq_runtime.evolve(initial_state, kernels, save_intermediate_states)
```
but the vector-kernel runtime binding expects the third positional argument to be params, not save_intermediate_states:
```
evolve(initial_state, kernels, params, observables, shots_count,
       save_intermediate_states)
```
As a result, the boolean was interpreted as schedule parameters and the call failed with:
```
Invoked with types: State, list, bool
```
The same early-return structure also skipped collapse-noise setup. The generated Kraus channels were registered into an internal NoiseModel, but the no-observable branches returned before installing or passing that noise model. This affected both final-state-only evolution and intermediate-state evolution.

The async no-observable intermediate path had another mismatch: it called
```
cudaq_runtime.evolve_async(initial_state, kernels)
```
without passing save_intermediate_states. The Python binding defaults save_intermediate_states to True, so IntermediateResultSave.EXPECTATION_VALUE could incorrectly save all intermediate states unless the flag is passed explicitly.

## How to repro
```
import numpy as np
import cudaq
from cudaq.operators import spin

cudaq.set_target("density-matrix-cpu")

dimensions = {0: 2}
schedule = Schedule(np.linspace(0, 0.1, 3), ["time"])
rho0 = cudaq.State.from_data(
    np.array([[1.0, 0.0], [0.0, 0.0]], dtype=np.complex128))

result = cudaq.evolve(
    hamiltonian,
    dimensions,
    schedule,
    rho0,
    store_intermediate_results=cudaq.IntermediateResultSave.ALL,
)

print(len(result.intermediate_states()))
print(result.expectation_values())
print(result.final_expectation_values())

cudaq.reset_target()
```
Before this fix:
```
TypeError: evolve(): incompatible function arguments
Invoked with types: State, list, bool
```
After this fix:
```
3
None
None
```

## This change
* Passes save_intermediate_states by keyword in sync no-observable intermediate evolution, so it is not interpreted as params.
* Preserves no-observable result semantics: expectation_values() and final_expectation_values() remain None.
* Applies collapse noise even when no observables are requested.
* Uses try/finally around internally installed sync noise models so the internal collapse noise is cleaned up only when this code installed it.
* Passes noise_model, shots_count, and save_intermediate_states explicitly in async no-observable paths.
* Adds sync and async regression coverage for no-observable evolution on density-matrix-cpu.

### DIFF
--- a/python/cudaq/dynamics/evolution.py
+++ b/python/cudaq/dynamics/evolution.py
@@ -319,15 +319,30 @@ def evolve_single(
         kernels = [kernel for kernel in evolution]
         save_intermediate_states = store_intermediate_results == IntermediateResultSave.ALL
         if len(observables) == 0:
-            return cudaq_runtime.evolve(initial_state, kernels,
-                                        save_intermediate_states)
+            if len(collapse_operators) > 0:
+                cudaq_runtime.set_noise(noise)
+                try:
+                    return cudaq_runtime.evolve(
+                        initial_state,
+                        kernels,
+                        save_intermediate_states=save_intermediate_states)
+                finally:
+                    cudaq_runtime.unset_noise()
+            return cudaq_runtime.evolve(
+                initial_state,
+                kernels,
+                save_intermediate_states=save_intermediate_states)
         if len(collapse_operators) > 0:
             cudaq_runtime.set_noise(noise)
-        result = cudaq_runtime.evolve(initial_state, kernels, parameters,
-                                      observable_spinops, shots_count,
-                                      save_intermediate_states)
-        cudaq_runtime.unset_noise()
-        return result
+            try:
+                return cudaq_runtime.evolve(initial_state, kernels, parameters,
+                                            observable_spinops, shots_count,
+                                            save_intermediate_states)
+            finally:
+                cudaq_runtime.unset_noise()
+        return cudaq_runtime.evolve(initial_state, kernels, parameters,
+                                    observable_spinops, shots_count,
+                                    save_intermediate_states)
     else:
         kernel = next(
             _evolution_kernel(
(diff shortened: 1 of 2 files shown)


sacpis (approved):
LGTM. Thanks @huaweil-nv.

sacpis:
> /ok to test 5c72682

huaweil-nv:
> /ok to test 68893faa56fbd57c0fce63ccef6b283de637b035

1tnguyen (approved):
Thanks @huaweil-nv!

1tnguyen:
> /ok to test [99110e2](https://github.com/NVIDIA/cuda-quantum/pull/4674/commits/99110e28fade20709e7c7df874017f6c28bfd155)

PR #4675, opened 2026-06-05
Fix: `--qbraid-machine` routing in C++ target

TheGupta2012:
The yml mapped target argument `machine` to `platform-arg: qpu`, but `QbraidServerHelper::initialize()` reads `config["machine"]`. As a result, `nvq++ --qbraid-machine <device>` was silently dropped and every C++ job defaulted to `qbraid:qbraid:sim:qir-sv`. The Python path `(cudaq.set_target("qbraid", machine=...))` was unaffected because it writes config["machine"] directly, bypassing the YAML mapping.

Set `platform-arg: machine` so the CLI flag, runtime config key, and helper read all agree.


mitchdz:
> /ok to test [6e138fc](https://github.com/NVIDIA/cuda-quantum/pull/4675/commits/6e138fc53a1301a72c98779a0acca99e0e8e3d08)

mitchdz:
@TheGupta2012 Thanks a lot for this, this looks good to me.

We should add a case to `unittests/target_config/TargetConfigTester.cpp` using the public `processRuntimeArgs(config, {…})` and assert the emitted platform-args contain `;machine;<value>` (not `;qpu;`…).

But due to the rapidly approaching release, let's add that test as a follow-up PR and merge this in first.

mitchdz (approved):

PR #4677, opened 2026-06-05
Move warnNamedMeasurement out of ExecutionContext

lmondada:
Currently, a warning gets emitted when using named measurements in a context that does not support it. This behaviour seems to be temporary in "sample" (as a deprecation measure), but permanent for PTSBE. It comes with a QOL feature in the runtime that ensures this warning is only emitted once per kernel launch.

As @Renaud-K and I were moving things out of the global execution context into policy objects, compiler flags etc, we've had to add a lot of logic just to track the status of whether this warning has been emitted yet. This seems overkill.

I suggest a simplification that tracks the `warnedNamedMeasurements` flag outside of the execution context, within a utiliy function called `emitNamedMeasurementsWarning`. This function keeps a static boolean flag to only emit the warning on its first call.

This changes the behaviour a bit: whereas before, the warning would get emitted _once per kernel launch_, it now gets emitted only once in the lifetime of the process. This change seems harmless to me but if anyone feels strongly about this change, please speak up!

### DIFF
--- a/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
+++ b/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
@@ -438,7 +438,7 @@ def test_named_reg_in_sample(capfd):
 
     cudaq.sample(foo).dump()
     captured = capfd.readouterr()
-    assert "WARNING" in captured.err
+    assert "[warning]" in captured.err
 
 
 def test_sample_with_conditional():
--- a/python/tests/kernel/test_kernel_features.py
+++ b/python/tests/kernel/test_kernel_features.py
@@ -2919,7 +2919,7 @@ def test_named_reg_in_sample(capfd):
 
     cudaq.sample(baz)
     captured = capfd.readouterr()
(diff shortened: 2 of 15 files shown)


Renaud-K:
The best way to solve this is to move the compiler warning AOT, so it does not have to also be issued during simulation. 

If this is not easy for some reason, then I would just make the CircuitSimulator more stateful. It is already stateful, literally. It now also holds the intermediate results that were previously on the executionContext as well as the noise. So, as we are moving more things out of the executionContext, I do not see how we can prevent the CircuitSimulator from growing more stateful. I think that it is ok because:
1. It already has state
2. It is not meant to be reentrant
3. It has the hooks for initialization and finalization precisely so its state can be reset, which can in turn, reset the warnedNamedMeasurements flag.

In the worse case, the warning is issued twice per launch. Once during JIT compilation and once during simulation.  But we should not be worrying about carrying over the one warning emission from compilation over to simulation. We only care about the warning not being issued on every iteration of every operation.

lmondada:
A more 'holistic' approach will be needed.

PR #4678, opened 2026-06-05
[unitaryhack] Topology-aware initial placement for the qubit-mapping pass

### LINKED ISSUE #4289: SABRE qubit-mapping pass: topology-aware initial placement to reduce SWAPs on irregular devices

border-b:
Fixes #4289.

The qubit-mapping pass starts its SABRE router from identity placement (`virtual i -> physical i`). On irregular topologies that ties the routed SWAP count to how the physical qubits happen to be numbered: the issue repro shows two isomorphic stars routing differently, `star(5,2)` at 2 swaps versus `star(5,0)` at 1.

My first attempt added a greedy interaction-aware placement, then used a static distance score to pick between that layout and identity. As @taalexander and @Zaneham pointed out, that left the placement, the selection, and the router each optimizing a different model, and the router's dynamic decisions can invalidate the static ones. This revision follows the structure they suggested and chooses the layout by the thing the router actually rewards: the routed SWAP count.

### What changed

The router is now a read-only search with a single writer at the end:

- A `RoutingProblem` (the gate dependency DAG, source users, and virtual-qubit operands) is built once from the IR.
- `identity` and the greedy placement become *seeds* that only propose a starting layout. The static `layoutScore` selection and the silent fallback to identity are gone.
- A `RoutingSearchStrategy` routes each seed in analyze mode and refines it with the paper's forward-backward-forward reverse traversal (Sec. IV-C2). The reverse pass is just a forward route over a transposed `RoutingProblem`, so it reuses the router unchanged.
- The candidate with the fewest routed swaps wins, behind a small comparator so other metrics (depth, ...) can slot in later.
- A `RoutingEmitter` rewrites the IR once, from the winner.

Two independent options replace the old single one:

- `placement = auto | identity | greedy` (default `auto`: try both seeds, keep whichever routes better)
- `search = sabre | none` (default `sabre`: the reverse-traversal refinement, with `none` a single forward route)

`placement=identity search=none` preserves the old single-shot behavior and keeps the existing exact-output tests as the legacy regression net. A LightSABRE-style strategy (random restarts) would later be a new `search` value rather than another rewrite. For now the seeds are deterministic, so the exact-output tests stay reproducible, as agreed in the thread.

### Routing robustness

The SABRE cost function is a heuristic. With reverse-traversal refinement it can occasionally emit a long run of swaps without making any front-layer gate executable. `route()` now has a release valve (inspired by Qiskit/LightSABRE): after a generous no-progress budget, it discards that local episode and forces the closest front-layer gate along a shortest path. On a connected device, any such direct route takes at most the device diameter, so this guarantees progress while leaving normally converging searches alone. The budget is intentionally loose by default (`max(64, 4 * numPhysicalQubits)`) and is covered by a small `path(6)` regression that contrasts a one-swap stall budget against a high budget: the valve route emits 5 swaps, the unrestricted heuristic emits 3, and `CircuitCheck` verifies both mapped outputs.

### Results

Issue repro (`@hub_pattern`), defaults:

| device    | legacy (identity) | default (search) |
|-----------|:-----------------:|:----------------:|
| path(5)   |         1         |        1         |
| star(5,0) |         1         |        1         |
| star(5,2) |         2         |        1         |
| grid(3,3) |         1         |        1         |

`star(5,2)` now matches `star(5,0)`, so the center-index artifact is gone and the mean over the sweep drops from 1.25 to 1.00.

For a larger check I swept quantum-volume-style circuits (n qubits, n layers, random two-qubit pairings, 10 seeds per size) over `path`, `ring`, `grid`, and `star`, and compared against Qiskit's SABRE (Qiskit 2.4.2, sabre layout and routing) on the same circuits and coupling maps. Mean added swaps at 24 qubits:

| topology | identity | default | Qiskit SABRE |
|----------|---------:|--------:|-------------:|
| path     |   1410.3 |  1250.9 |       1091.6 |
| ring     |    911.5 |   809.0 |        732.1 |
| grid     |    396.2 |   357.3 |        300.0 |
| star     |    139.8 |   138.5 |        127.5 |

Through initial placement alone, the default closes 40 to 57 percent of the identity-to-Qiskit swap gap on `path`, `ring`, and `grid`, landing within roughly 10 to 19 percent of Qiskit. The remaining gap is the router-side work, the LightSABRE cost function and random restarts, that this issue lists as out-of-scope follow-ups. On `star` identity is already near-optimal, so there is little to gain.

On two-qubit depth the default is competitive with Qiskit and lower on `path` and `grid`:

| topology | identity | default | Qiskit SABRE |
|----------|---------:|--------:|-------------:|
| path     |    301.8 |   273.6 |        285.3 |
| ring     |    183.8 |   166.7 |        164.2 |
| grid     |    173.0 |   166.6 |        170.0 |
| star     |    427.8 |   426.5 |        415.5 |

The cost is up to two seeds times three routing walks instead of one routing walk, but the absolute pass time stays small. Mean `qubit-mapping` pass time from `--mlir-timing`:

| n  | legacy | default | factor |
|---:|-------:|--------:|-------:|
|  4 | 0.28ms |  0.31ms |   1.1x |
|  8 | 0.37ms |  0.57ms |   1.5x |
| 12 | 0.86ms |  2.01ms |   2.3x |
| 16 | 1.74ms |  5.20ms |   3.0x |
| 20 | 3.38ms | 11.28ms |   3.3x |
| 24 | 6.19ms | 22.33ms |   3.6x |

The full sweep is in the plot, with script attached:

<details>
<summary>benchmark.py</summary>

```python
#!/usr/bin/env python3
"""Routing benchmark for the qubit-mapping SABRE redesign (PR #4678).

Compares the legacy single-shot identity path against the new default search on
quantum-volume-style circuits, sweeping device topologies and sizes. Reports
routing quality (added swaps, two-qubit gate count, two-qubit depth) and the
qubit-mapping pass wall-clock, so the quality gain can be weighed against the
compile-time cost.

Reproduce:

    python benchmark.py --cudaq-opt /path/to/cudaq-opt --out results

Quantum-volume circuits are generated directly as value-semantics Quake: n
qubits, n layers, each layer pairing qubits by a seeded random permutation with
a two-qubit gate per pair. Only the two-qubit structure matters for routing, so
a controlled-x stands in for the Haar-random SU(4) block. Fixed seeds make every
number reproducible.
"""

import argparse
import csv
import os
import re
import statistics
import subprocess
import tempfile
import time
from collections import defaultdict


def gen_qv_qke(n, seed):
    """Return a Quake module for an n-qubit, n-layer QV-style circuit."""
    import random

    rng = random.Random(seed)
    lines = ["quake.wire_set @wires[2147483647]", "func.func @qv() {"]
    cur = {}
    for q in range(n):
        lines.append(f"  %v{q} = quake.borrow_wire @wires[{q}] : !quake.wire")
        cur[q] = f"%v{q}"
    k = 0
    for _ in range(n):
        perm = list(range(n))
        rng.shuffle(perm)
        for i in range(0, n - 1, 2):
            a, b = perm[i], perm[i + 1]
            g = f"%g{k}"
            lines.append(
                f"  {g}:2 = quake.x [{cur[a]}] {cur[b]} : "
                "(!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)")
            cur[a], cur[b] = f"{g}#0", f"{g}#1"
            k += 1
    for q in range(n):
        lines.append(f"  quake.return_wire {cur[q]} : !quake.wire")
    lines.append("  return")
    lines.append("}")
    return "\n".join(lines) + "\n"


# A 2q gate consumes two wires and produces two; everything else carries depth
# through unchanged.
RESULT_RE = re.compile(r"^\s*([%\w:,\s#]+?)\s*=\s*quake\.(\w+)\b(.*?):", re.S)
SSA_RE = re.compile(r"%[\w#]+")


def two_qubit_metrics(ir):
    """Count swaps and two-qubit gates and compute two-qubit depth from IR."""
    depth = defaultdict(int)
    swaps = twoq = 0
    maxdepth = 0
    for line in ir.splitlines():
        m = RESULT_RE.match(line)
        if not m:
            continue
        results_part, op, operands_part = m.group(1), m.group(2), m.group(3)
        operands = SSA_RE.findall(operands_part)
        results = []
        for tok in results_part.split(","):
            tok = tok.strip()
            base = SSA_RE.findall(tok)
            if not base:
                continue
            base = base[0]
            arity = 2 if tok.endswith(":2") else 1
            results += [f"{base}#{i}" for i in range(arity)] if arity == 2 \
                else [base]
        is_2q = op == "swap" or (op == "x" and "[" in operands_part)
        if is_2q:
            twoq += 1
            if op == "swap":
                swaps += 1
            d = 1 + max((depth[o] for o in operands), default=0)
        else:
            d = max((depth[o] for o in operands), default=0)
        for r in results:
            depth[r] = d
        maxdepth = max(maxdepth, d)
    return swaps, twoq, maxdepth


PASS_TIME_RE = re.compile(r"([\d.]+)\s+\(.*?\)\s+(.*Mapping.*)")


def pass_ms(timing_text):
    """Sum the mapping pass wall-clock (ms) from -mlir-timing output. The first
    float on each pass line is its wall time in seconds."""
    total = 0.0
    for line in timing_text.splitlines():
        if "apping" in line:  # MappingPrep, MappingFunc, ...
            nums = re.findall(r"(\d+\.\d+)", line)
            if nums:
                total += float(nums[0]) * 1000.0
    return total


def run_once(cudaq_opt, qke, device, placement, search):
    args = (f"--qubit-mapping=device={device} placement={placement} "
            f"search={search}")
    with tempfile.NamedTemporaryFile("w", suffix=".qke", delete=False) as fh:
        fh.write(qke)
        path = fh.name
    try:
        cmd = [cudaq_opt, args, path, "-mlir-timing"]
        t0 = time.perf_counter()
        try:
            p = subprocess.run(cmd, capture_output=True, text=True, timeout=120)
        except subprocess.TimeoutExpired:
            return None
        wall = (time.perf_counter() - t0) * 1000.0
    finally:
        os.unlink(path)
    if p.returncode != 0:
        return None
    swaps, twoq, depth = two_qubit_metrics(p.stdout)
    ms = pass_ms(p.stderr)
    return dict(swaps=swaps, twoq=twoq, depth=depth,
                pass_ms=round(ms, 4), wall=round(wall, 4),
                ms=round(ms or wall, 4))


CONFIGS = {
    "legacy": dict(placement="identity", search="none"),
    "search": dict(placement="auto", search="sabre"),
}


def devices_for(n):
    w = int(round(n ** 0.5))
    h = (n + w - 1) // w
    return {
        "path": f"path({n})",
        "ring": f"ring({n})",
        "grid": f"grid({w},{h})",
        "star": f"star({n},0)",
    }


def main():
    ap = argparse.ArgumentParser()
    ap.add_argument("--cudaq-opt", required=True)
    ap.add_argument("--out", default="results")
    ap.add_argument("--sizes", default="4,6,8,10,12,14,16")
    ap.add_argument("--seeds", type=int, default=10)
    args = ap.parse_args()

    sizes = [int(s) for s in args.sizes.split(",")]
    os.makedirs(args.out, exist_ok=True)
    rows = []
    for n in sizes:
        for topo, device in devices_for(n).items():
            for seed in range(args.seeds):
                qke = gen_qv_qke(n, seed)
                for cfg, opts in CONFIGS.items():
                    r = run_once(args.cudaq_opt, qke, device, **opts)
                    if r is None:
                        continue
                    rows.append(dict(n=n, topo=topo, seed=seed, config=cfg, **r))
            done = {c: [x for x in rows
                        if x["n"] == n and x["topo"] == topo and x["config"] == c]
                    for c in CONFIGS}
            msg = []
            for c in CONFIGS:
                if done[c]:
                    msg.append(f"{c} swaps~{statistics.mean(x['swaps'] for x in done[c]):.1f}")
            print(f"n={n:>3} {topo:<5} " + "  ".join(msg), flush=True)

    csv_path = os.path.join(args.out, "bench.csv")
    with open(csv_path, "w", newline="") as fh:
        w = csv.DictWriter(fh, fieldnames=["n", "topo", "seed", "config",
                                           "swaps", "twoq", "depth",
                                           "pass_ms", "wall", "ms"])
        w.writeheader()
        w.writerows(rows)
    print(f"wrote {csv_path} ({len(rows)} rows)")

    try:
        make_plot(rows, sizes, os.path.join(args.out, "bench.png"))
    except Exception as e:  # plotting is optional; the CSV is the source of truth
        print(f"[plot skipped: {e}]")


def make_plot(rows, sizes, path):
    import matplotlib
    matplotlib.use("Agg")
    import matplotlib.pyplot as plt

    topos = ["path", "ring", "grid", "star"]
    fig, axes = plt.subplots(3, len(topos), figsize=(4 * len(topos), 9),
                             sharex=True)

    def mean(n, topo, cfg, key):
        vals = [r[key] for r in rows
                if r["n"] == n and r["topo"] == topo and r["config"] == cfg]
        return statistics.mean(vals) if vals else float("nan")

    # Dashed/square for legacy, solid/circle for search, so the two stay
    # readable where they nearly coincide (e.g. star).
    styles = {
        "legacy": dict(color="C0", linestyle="--", marker="s", label="legacy"),
        "search": dict(color="C1", linestyle="-", marker="o", label="search"),
    }

    for j, topo in enumerate(topos):
        for cfg, st in styles.items():
            axes[0][j].plot(sizes, [mean(n, topo, cfg, "swaps") for n in sizes],
                            **st)
            axes[2][j].plot(sizes,
                            [mean(n, topo, cfg, "pass_ms") for n in sizes], **st)
        reduction = []
        for n in sizes:
            leg = mean(n, topo, "legacy", "swaps")
            se = mean(n, topo, "search", "swaps")
            reduction.append((leg - se) / leg * 100.0 if leg else 0.0)
        axes[1][j].plot(sizes, reduction, color="C2", marker="o")
        axes[1][j].axhline(0, color="0.7", lw=0.8)
        axes[0][j].set_title(topo)
        axes[2][j].set_xlabel("qubits")
    axes[0][0].set_ylabel("added swaps (mean)")
    axes[1][0].set_ylabel("swap reduction (%)")
    axes[2][0].set_ylabel("mapping pass ms (mean)")
    axes[0][-1].legend()
    fig.suptitle("QV-style routing: legacy identity vs default search")
    fig.tight_layout()
    fig.savefig(path, dpi=120)
    print(f"wrote {path}")


if __name__ == "__main__":
    main()
```

</details>

image

### Tests

The existing exact-output tests are pinned to `placement=identity search=none` and become the legacy-behavior regression net, with their CHECK bodies untouched. New tests cover the issue repro, the review's late-interaction scenario, a case where the reverse pass beats a single forward pass, the option matrix, invalid values for both options (fatal and warn), measurement/control-flow failure modes, and a small release-valve contrast test with `FileCheck` and `CircuitCheck`.

**AI disclosure:** I used Opus via Claude Code to set up the remote build/test environment, help carry out the restructuring, and write the tests and benchmark harness. The design decisions, final edits, and verification are mine.

### DIFF
--- a/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
@@ -956,6 +956,14 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
            "Decay delta">,
     Option<"roundsDecayReset", "roundsDecayReset", "unsigned", /*default=*/"5",
            "Number of rounds before decay is reset">,
+    Option<"placement", "placement", "std::string",
+           /*default=*/"\"auto\"",
+           "Initial placement strategy: auto (propose identity and greedy "
+           "seeds, select by routed swap count), identity, greedy">,
+    Option<"search", "search", "std::string", /*default=*/"\"sabre\"",
+           "Layout search strategy: sabre (SABRE forward-backward-forward "
+           "reverse-traversal refinement of each seed), none (single forward "
+           "route per seed; legacy mapping is placement=identity search=none)">,
     Option<"nonComposable", "raise-fatal-errors", "bool", /*default=*/"false",
            "Run the pass in a non-composable way, which may cause immediate "
            "internal compiler errors">
--- a/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
@@ -956,6 +956,16 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
            "Decay delta">,
     Option<"roundsDecayReset", "roundsDecayReset", "unsigned", /*default=*/"5",
            "Number of rounds before decay is reset">,
+    Option<"minStallSwapBudget", "minStallSwapBudget", "unsigned",
+           /*default=*/"64",
+           "Release valve floor: minimum number of consecutive swaps that route "
+           "no gate before a stuck front layer is force-routed along a shortest "
+           "path (advanced)">,
+    Option<"stallSwapBudgetPerQubit", "stallSwapBudgetPerQubit", "unsigned",
+           /*default=*/"4",
+           "Release valve scaling: per-device-qubit stall budget. The budget "
+           "used is max(minStallSwapBudget, stallSwapBudgetPerQubit * numQubits) "
+           "(advanced)">,
     Option<"placement", "placement", "std::string",
            /*default=*/"\"auto\"",
            "Initial placement strategy: auto (propose identity and greedy "
--- a/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
@@ -948,24 +948,24 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
     Option<"device", "device", "std::string", /*default=*/"\"-\"",
       "Device topology: path(N), ring(N), star(N), star(N,c), grid(w,h), "
       "file(/path/to/file), bypass">,
-    Option<"extendedLayerSize", "extendedLayerSize", "unsigned",
+    Option<"extendedLayerSize", "extended-layer-size", "unsigned",
            /*default=*/"20", "Extended layer size">,
-    Option<"extendedLayerWeight", "extendedLayerWeight", "float",
+    Option<"extendedLayerWeight", "extended-layer-weight", "float",
            /*default=*/"0.5", "Extended layer weight">,
-    Option<"decayDelta", "decayDelta", "float", /*default=*/"0.5",
+    Option<"decayDelta", "decay-delta", "float", /*default=*/"0.5",
            "Decay delta">,
-    Option<"roundsDecayReset", "roundsDecayReset", "unsigned", /*default=*/"5",
+    Option<"roundsDecayReset", "rounds-decay-reset", "unsigned", /*default=*/"5",
            "Number of rounds before decay is reset">,
-    Option<"minStallSwapBudget", "minStallSwapBudget", "unsigned",
+    Option<"minStallSwapBudget", "min-stall-swap-budget", "unsigned",
            /*default=*/"64",
            "Release valve floor: minimum number of consecutive swaps that route "
            "no gate before a stuck front layer is force-routed along a shortest "
            "path (advanced)">,
-    Option<"stallSwapBudgetPerQubit", "stallSwapBudgetPerQubit", "unsigned",
+    Option<"stallSwapBudgetPerQubit", "stall-swap-budget-per-qubit", "unsigned",
            /*default=*/"4",
            "Release valve scaling: per-device-qubit stall budget. The budget "
-           "used is max(minStallSwapBudget, stallSwapBudgetPerQubit * numQubits) "
-           "(advanced)">,
+           "used is max(min-stall-swap-budget, stall-swap-budget-per-qubit * "
+           "numQubits) (advanced)">,
     Option<"placement", "placement", "std::string",
            /*default=*/"\"auto\"",
            "Initial placement strategy: auto (propose identity and greedy "
--- a/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
+++ b/cudaq/include/cudaq/Optimizer/Transforms/Passes.td
@@ -967,10 +967,10 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
            "used is max(min-stall-swap-budget, stall-swap-budget-per-qubit * "
            "numQubits) (advanced)">,
     Option<"placement", "placement", "std::string",
-           /*default=*/"\"auto\"",
+           /*default=*/"\"identity\"",
            "Initial placement strategy: auto (propose identity and greedy "
            "seeds, select by routed swap count), identity, greedy">,
-    Option<"search", "search", "std::string", /*default=*/"\"sabre\"",
+    Option<"search", "search", "std::string", /*default=*/"\"none\"",
            "Layout search strategy: sabre (SABRE forward-backward-forward "
            "reverse-traversal refinement of each seed), none (single forward "
            "route per seed; legacy mapping is placement=identity search=none)">,
--- a/cudaq/lib/Optimizer/Transforms/Mapping.cpp
+++ b/cudaq/lib/Optimizer/Transforms/Mapping.cpp
@@ -204,8 +204,8 @@ private:
   /// The unplaced virtual qubit most connected to the placed set. Ties break by
   /// total weighted degree, then by lower virtual index for determinism. The
   /// disconnected case (no interaction with the placed set) reduces to highest
-  /// weighted degree, then lower index, since weighted degree is always at least
-  /// the placed weight.
+  /// weighted degree, then lower index, since weighted degree is always at
+  /// least the placed weight.
   unsigned chooseNextVirtual() const {
     unsigned pick = n;
     CandidateScore best;
@@ -219,7 +219,8 @@ private:
         best = score;
       }
     }
-    assert(pick != n && "chooseNextVirtual called with no unplaced user qubits");
+    assert(pick != n &&
+           "chooseNextVirtual called with no unplaced user qubits");
     return pick;
   }
 
@@ -842,8 +843,7 @@ void SabreRouter::forceClosestGate() {
       closest = n;
     }
   }
-  assert(closest.isValid() &&
-         "a stalled front layer must hold a 2-qubit gate");
+  assert(closest.isValid() && "a stalled front layer must hold a 2-qubit gate");
   const RoutingProblem::Node &node = problem[closest];
   cudaq::Device::Path path = device.getShortestPath(
       placement.getPhy(node.qubits[0]), placement.getPhy(node.qubits[1]));
@@ -1670,8 +1670,7 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
           op.emitOpError("is not supported by the mapper");
           signalPassFailure();
         }
-        LLVM_DEBUG(llvm::dbgs()
-                   << "unsupported quantum operation in mapper\n");
+        LLVM_DEBUG(llvm::dbgs() << "unsupported quantum operation in mapper\n");
         return;
       }
     }
@@ -1687,11 +1686,13 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
       return;
     }
 
-    // Measurement deferral is only safe for terminal readout. Reject unsupported
-    // mid-circuit/adaptive uses before mutating IR, even in composable mode.
+    // Measurement deferral is only safe for terminal readout. Reject
+    // unsupported mid-circuit/adaptive uses before mutating IR, even in
+    // composable mode.
     if (Operation *measOp = findNonTerminalMeasuredWireUse(func)) {
-      measOp->emitOpError("unsupported mid-circuit measurement: a measured wire "
-                          "is used by a later operation");
+      measOp->emitOpError(
+          "unsupported mid-circuit measurement: a measured wire "
+          "is used by a later operation");
       signalPassFailure();
       return;
     }
@@ -1704,9 +1705,10 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
     auto measIt = measInfo.find(func);
     assert(measIt != measInfo.end() && "missing measurement analysis for func");
     if (measIt->second.hasConditionalsOnMeasure) {
-      func.emitOpError("unsupported measurement-dependent behavior: "
-                       "measurement-dependent control flow, quantum operations, "
-                       "calls, or resets cannot be preserved by qubit mapping");
+      func.emitOpError(
+          "unsupported measurement-dependent behavior: "
+          "measurement-dependent control flow, quantum operations, "
+          "calls, or resets cannot be preserved by qubit mapping");
       signalPassFailure();
       return;
     }
@@ -1740,9 +1742,9 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
                                                finalQubitWire[i].getLoc(),
                                                resTy, measureOp.getMeasOut());
 
-          wireToVirtualQ.insert({measureOp.getWires()[0],
-                                 requireVirtualQ(wireToVirtualQ,
-                                                 finalQubitWire[i])});
+          wireToVirtualQ.insert(
+              {measureOp.getWires()[0],
+               requireVirtualQ(wireToVirtualQ, finalQubitWire[i])});
 
           userQubitsMeasured.push_back(i);
         }
@@ -1773,9 +1775,9 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
     const unsigned numV = sources.size();
     const unsigned numPhy = deviceInstance->getNumQubits();
 
-    SmallVector<SmallVector<unsigned>> seeds = buildPlacementSeeds(
-        placementStrategy, numV, *deviceInstance, interactions,
-        userVirtualQubits);
+    SmallVector<SmallVector<unsigned>> seeds =
+        buildPlacementSeeds(placementStrategy, numV, *deviceInstance,
+                            interactions, userVirtualQubits);
 
     // Build the routing problem once (it does not depend on the layout), then
     // search over the seeds for the result with the fewest swaps.
@@ -1891,7 +1893,8 @@ struct MappingPipelineOptions
   DECLARE_SUB_OPTION(MappingFuncOptions, extendedLayerWeight,
                      "extended-layer-weight");
   DECLARE_SUB_OPTION(MappingFuncOptions, decayDelta, "decay-delta");
-  DECLARE_SUB_OPTION(MappingFuncOptions, roundsDecayReset, "rounds-decay-reset");
+  DECLARE_SUB_OPTION(MappingFuncOptions, roundsDecayReset,
+                     "rounds-decay-reset");
   DECLARE_SUB_OPTION(MappingFuncOptions, minStallSwapBudget,
                      "min-stall-swap-budget");
   DECLARE_SUB_OPTION(MappingFuncOptions, stallSwapBudgetPerQubit,
--- a/cudaq/lib/Optimizer/Transforms/Mapping.cpp
+++ b/cudaq/lib/Optimizer/Transforms/Mapping.cpp
@@ -9,6 +9,7 @@
 #include "PassDetails.h"
 #include "cudaq/Optimizer/Transforms/Passes.h"
 #include "cudaq/Support/Device.h"
+#include "cudaq/Support/Handle.h"
 #include "cudaq/Support/Placement.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/StringSwitch.h"
@@ -35,24 +36,382 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
 // Placement
 //===----------------------------------------------------------------------===//
 
-void identityPlacement(cudaq::Placement &placement) {
-  for (unsigned i = 0, end = placement.getNumVirtualQubits(); i < end; ++i)
-    placement.map(cudaq::Placement::VirtualQ(i), cudaq::Placement::DeviceQ(i));
+/// Initial-layout strategy selected by the `placement` pass option.
+enum class PlacementStrategy { Auto, Identity, Greedy };
+
+/// Parse the `placement` option string, or nullopt for an unknown value.
+std::optional<PlacementStrategy> parsePlacementStrategy(llvm::StringRef name) {
+  return llvm::StringSwitch<std::optional<PlacementStrategy>>(name)
+      .Case("auto", PlacementStrategy::Auto)
+      .Case("identity", PlacementStrategy::Identity)
+      .Case("greedy", PlacementStrategy::Greedy)
+      .Default(std::nullopt);
+}
+
+/// Builds a deterministic topology-aware initial layout by assigning highly
+/// interacting virtual qubits to central physical qubits first. The greedy
+/// growth and its tie-breaks make the layout a deterministic function of the
+/// interaction counts and the device, so it is reproducible across runs.
+class GreedyInitialPlacer {
+public:
+  GreedyInitialPlacer(const cudaq::Device &device,
+                      ArrayRef<SmallVector<unsigned>> interactions,
+                      ArrayRef<bool> userVirtualQubits)
+      : device(device), interactions(interactions),
+        userVirtualQubits(userVirtualQubits), n(device.getNumQubits()),
+        vrToPhy(n, 0), placedVirtual(n, false), usedPhysical(n, false) {}
+
+  /// Produce the `vrToPhy` seed layout.
+  SmallVector<unsigned> run() {
+    computeDegrees();
+
+    // No two-qubit interactions, so every layout routes identically; return the
+    // identity seed for a deterministic result.
+    if (!hasInteraction) {
+      for (unsigned v = 0; v < n; ++v)
+        vrToPhy[v] = v;
+      return vrToPhy;
+    }
+
+    computeCentrality();
+
+    // Seed the highest-degree virtual qubit onto the most central physical
+    // qubit, then grow the layout around it.
+    place(chooseSeedVirtual(), bestFreePhysical());
+
+    unsigned remaining = 0;
+    for (unsigned u = 0; u < n; ++u)
+      if (userVirtualQubits[u] && !placedVirtual[u])
+        ++remaining;
+    while (remaining > 0) {
+      unsigned v = chooseNextVirtual();
+      place(v, bestPhysicalFor(v));
+      --remaining;
+    }
+
+    assignRemainingVirtuals();
+    return vrToPhy;
+  }
+
+private:
+  /// Weighted degree of each virtual qubit: the sum of its two-qubit
+  /// interaction counts. Also records whether any interaction exists at all.
+  void computeDegrees() {
+    weightedDegree.assign(n, 0);
+    for (unsigned u = 0; u < n; ++u) {
+      for (unsigned v = 0; v < n; ++v)
+        weightedDegree[u] += interactions[u][v];
+      if (weightedDegree[u] > 0)
+        hasInteraction = true;
+    }
+  }
+
+  /// Physical centrality used to break ties: total distance to every other
+  /// qubit, and connectivity degree.
+  void computeCentrality() {
+    using Qubit = cudaq::Device::Qubit;
+    distanceSum.assign(n, 0);
+    physDegree.assign(n, 0);
+    for (unsigned p = 0; p < n; ++p) {
+      for (unsigned q = 0; q < n; ++q)
+        distanceSum[p] += device.getDistance(Qubit(p), Qubit(q));
+      physDegree[p] =
+          static_cast<unsigned>(device.getNeighbours(Qubit(p)).size());
+    }
+  }
+
+  /// More central first: lower distance-sum, then higher degree, then lower
+  /// index.
+  bool physBetter(unsigned a, unsigned b) const {
+    if (distanceSum[a] != distanceSum[b])
+      return distanceSum[a] < distanceSum[b];
+    if (physDegree[a] != physDegree[b])
+      return physDegree[a] > physDegree[b];
+    return a < b;
+  }
+
+  /// The most central physical qubit not yet used.
+  unsigned bestFreePhysical() const {
+    unsigned best = n;
+    for (unsigned p = 0; p < n; ++p)
+      if (!usedPhysical[p] && (best == n || physBetter(p, best)))
+        best = p;
+    return best;
+  }
+
+  /// The highest weighted-degree virtual qubit, breaking ties toward the lower
+  /// index.
+  unsigned chooseSeedVirtual() const {
+    unsigned seed = n;
+    for (unsigned u = 0; u < n; ++u)
+      if (userVirtualQubits[u] &&
+          (seed == n || weightedDegree[u] > weightedDegree[seed]))
+        seed = u;
+    return seed;
+  }
+
+  /// The unplaced virtual qubit most connected to the placed ones, falling back
+  /// to weighted degree for disconnected components.
+  unsigned chooseNextVirtual() const {
+    unsigned pick = n;
+    unsigned pickWeight = 0;
+    for (unsigned u = 0; u < n; ++u) {
+      if (!userVirtualQubits[u] || placedVirtual[u])
+        continue;
+      unsigned placedWeight = 0;
+      for (unsigned v = 0; v < n; ++v)
+        if (placedVirtual[v])
+          placedWeight += interactions[u][v];
+      if (pick == n || placedWeight > pickWeight) {
+        pick = u;
+        pickWeight = placedWeight;
+      }
+    }
+    if (pickWeight == 0) {
+      pick = n;
+      unsigned pickDegree = 0;
+      for (unsigned u = 0; u < n; ++u) {
+        if (!userVirtualQubits[u] || placedVirtual[u])
+          continue;
+        if (pick == n || weightedDegree[u] > pickDegree) {
+          pick = u;
+          pickDegree = weightedDegree[u];
+        }
+      }
+    }
+    return pick;
+  }
+
+  /// The free physical qubit minimizing weighted distance from `v` to its
+  /// placed partners, breaking ties by centrality.
+  unsigned bestPhysicalFor(unsigned v) const {
+    using Qubit = cudaq::Device::Qubit;
+    unsigned bestPhy = n;
+    unsigned bestCost = 0;
+    for (unsigned p = 0; p < n; ++p) {
+      if (usedPhysical[p])
+        continue;
+      unsigned cost = 0;
+      for (unsigned w = 0; w < n; ++w)
+        if (placedVirtual[w] && interactions[v][w] > 0)
+          cost += interactions[v][w] *
+                  device.getDistance(Qubit(p), Qubit(vrToPhy[w]));
+      bool better = bestPhy == n || cost < bestCost ||
+                    (cost == bestCost && physBetter(p, bestPhy));
+      if (better) {
+        bestPhy = p;
+        bestCost = cost;
+      }
+    }
+    return bestPhy;
+  }
+
+  /// Map virtual `v` onto physical `p` and mark both as taken.
+  void place(unsigned v, unsigned p) {
+    vrToPhy[v] = p;
+    placedVirtual[v] = true;
+    usedPhysical[p] = true;
+  }
+
+  /// Assign any still-unplaced virtuals (non-user qubits) to the remaining free
+  /// physicals in order.
+  void assignRemainingVirtuals() {
+    unsigned nextPhy = 0;
+    for (unsigned v = 0; v < n; ++v) {
+      if (placedVirtual[v])
+        continue;
+      while (nextPhy < n && usedPhysical[nextPhy])
+        ++nextPhy;
+      place(v, nextPhy);
+    }
+  }
+
+  const cudaq::Device &device;
+  ArrayRef<SmallVector<unsigned>> interactions;
+  ArrayRef<bool> userVirtualQubits;
+  const unsigned n;
+
+  SmallVector<unsigned> weightedDegree;
+  bool hasInteraction = false;
+  SmallVector<unsigned> distanceSum;
+  SmallVector<unsigned> physDegree;
+
+  SmallVector<bool> placedVirtual;
+  SmallVector<bool> usedPhysical;
+  SmallVector<unsigned> vrToPhy;
+};
+
+/// Greedy initial placement over the circuit interaction graph. Returns a
+/// `vrToPhy` array proposing a starting layout for the router (the greedy
+/// seed).
+SmallVector<unsigned>
+interactionPlacement(const cudaq::Device &device,
+                     ArrayRef<SmallVector<unsigned>> interactions,
+                     ArrayRef<bool> userVirtualQubits) {
+  return GreedyInitialPlacer(device, interactions, userVirtualQubits).run();
 }
 
 //===----------------------------------------------------------------------===//
 // Routing
 //===----------------------------------------------------------------------===//
 
-/// This class encapsulates an quake operation that uses wires with information
-/// about the virtual qubits these wires correspond.
-struct VirtualOp {
-  mlir::Operation *op;
-  SmallVector<cudaq::Placement::VirtualQ, 2> qubits;
+/// The dependency DAG the router walks. It is built once from the IR and never
+/// modified, so a layout can be routed without touching the circuit. Each node
+/// is a routable operation. Its successors are the operations that consume its
+/// result wires.
+struct RoutingProblem {
+  /// A handle to a node in the DAG, i.e. an index into `nodes`.
+  struct NodeRef : cudaq::Handle {
+    using Handle::Handle;
+  };
 
-  VirtualOp(mlir::Operation *op, ArrayRef<cudaq::Placement::VirtualQ> qubits)
-      : op(op), qubits(qubits) {}
+  struct Node {
+    mlir::Operation *op;
+    /// Virtual qubits used by `op`, in quantum-operand order.
+    SmallVector<cudaq::Placement::VirtualQ, 2> qubits;
+    /// Routable users of `op`'s result wires, in use-list order. A user appears
+    /// once per result wire it consumes, so a node becomes ready once it has
+    /// been visited as many times as it has wire operands.
+    SmallVector<NodeRef, 4> successors;
+    bool isMeasure = false;
+    /// A gate (not a measurement, sink, or return). Only these participate in
+    /// the reverse-traversal pass.
+    bool isUnitary = false;
+    /// Two-qubit unitaries are the only nodes that join the extended layer.
+    bool isTwoQ = false;
+  };
+
+  /// Routable operations, in program order.
+  SmallVector<Node> nodes;
+  /// Routable users of the source wires, in source order then use-list order.
+  /// These seed the first front layer.
+  SmallVector<NodeRef> sourceUsers;
+
+  const Node &operator[](NodeRef n) const {
+    assert(n.isValid() && "invalid node handle");
+    return nodes[n.index];
+  }
 };
+
+/// A single routing decision: a gate mapped onto physical qubits, or a swap
+/// inserted between them. The router records these as it walks the circuit and
+/// the emitter replays them to rewrite the IR.
+struct RoutingEvent {
+  enum class Kind { Gate, Swap };
+
+  /// A gate mapped onto the physical qubits `phys`, in operand order.
+  static RoutingEvent gate(mlir::Operation *op,
+                           ArrayRef<cudaq::Placement::DeviceQ> phys) {
+    return RoutingEvent{
+        Kind::Gate, op,
+        SmallVector<cudaq::Placement::DeviceQ, 2>(phys.begin(), phys.end())};
+  }
+  /// A swap inserted between physical qubits `q0` and `q1`.
+  static RoutingEvent swap(cudaq::Placement::DeviceQ q0,
+                           cudaq::Placement::DeviceQ q1) {
+    return RoutingEvent{Kind::Swap, nullptr, {q0, q1}};
+  }
+
+  Kind kind;
+  mlir::Operation *op;
+  SmallVector<cudaq::Placement::DeviceQ, 2> phys;
+};
+
+/// The outcome of routing one layout. The emitter replays `trace` onto the IR.
+/// `swapCount` is the metric used to compare layouts.
+struct RoutingResult {
+  /// Virtual-to-physical layout at the start of the walk, before any swap.
+  SmallVector<unsigned> initialLayout;
+  SmallVector<RoutingEvent> trace;
+  unsigned swapCount = 0;
+};
+
+/// Build the routing problem from `block`. The nodes are the routable
+/// operations that `isSupportedMappingOperation` accepts, other than the source
+/// borrows. Edges and source successors are captured in MLIR use-list order so
+/// the walk visits successors in the same order as the SSA use-def chains.
+RoutingProblem buildRoutingProblem(
+    Block &block, ArrayRef<cudaq::quake::BorrowWireOp> sources,
+    const DenseMap<Value, cudaq::Placement::VirtualQ> &wireToVirtualQ) {
+  RoutingProblem problem;
+  DenseMap<Operation *, RoutingProblem::NodeRef> nodeIndex;
+
+  for (Operation &op : block) {
+    if (isa<cudaq::quake::BorrowWireOp>(op) ||
+        !cudaq::quake::isSupportedMappingOperation(&op))
+      continue;
+    RoutingProblem::Node node;
+    node.op = &op;
+    for (auto wire : cudaq::quake::getQuantumOperands(&op))
+      node.qubits.push_back(wireToVirtualQ.lookup(wire));
+    node.isMeasure = op.hasTrait<cudaq::QuantumMeasure>();
+    node.isUnitary = isa<cudaq::quake::OperatorInterface>(op);
+    // A two-qubit gate the router has to make adjacent: a unitary on two wires,
+    // not a measurement or a sink.
+    node.isTwoQ = node.isUnitary && node.qubits.size() == 2;
+    nodeIndex[&op] = RoutingProblem::NodeRef(problem.nodes.size());
+    problem.nodes.push_back(std::move(node));
+  }
+
+  // Record successor edges in use-list order, keeping the routable users only.
+  // A user is listed once per result wire it consumes, so a node's visit count
+  // reaches its wire-operand count exactly when all of its inputs are ready.
+  auto recordUsers = [&](Operation *producer,
+                         SmallVectorImpl<RoutingProblem::NodeRef> &out) {
+    for (Operation *user : producer->getUsers())
+      if (auto it = nodeIndex.find(user); it != nodeIndex.end())
+        out.push_back(it->second);
+  };
+  for (auto &node : problem.nodes)
+    recordUsers(node.op, node.successors);
+  for (auto borrow : sources)
+    recordUsers(borrow.getOperation(), problem.sourceUsers);
+
+  return problem;
+}
+
+/// Build the transposed problem over the unitary gates only, for the SABRE
+/// reverse-traversal pass. Routing this forward is equivalent to routing the
+/// original circuit in reverse: a node's successors here are its forward
+/// predecessors, and result wires that do not feed another unitary (the
+/// circuit's outputs) seed the walk. Measurements, sinks, and returns are not
+/// unitary nodes, so they drop out, which is the paper's "skip measurements in
+/// the reverse pass". Readiness is unchanged: a unitary has equal operand and
+/// result arity, so its threshold is `qubits.size()` in both directions.
+RoutingProblem buildReverseProblem(const RoutingProblem &forward) {
+  RoutingProblem reverse;
+  SmallVector<RoutingProblem::NodeRef> fwdToRev(forward.nodes.size());
+  for (unsigned i = 0, end = forward.nodes.size(); i < end; ++i) {
+    const RoutingProblem::Node &node = forward.nodes[i];
+    if (!node.isUnitary)
+      continue;
+    fwdToRev[i] = RoutingProblem::NodeRef(reverse.nodes.size());
+    RoutingProblem::Node rev;
+    rev.op = node.op;
+    rev.qubits = node.qubits;
+    rev.isUnitary = true;
+    rev.isTwoQ = node.isTwoQ;
+    reverse.nodes.push_back(std::move(rev));
+  }
+
+  for (unsigned i = 0, end = forward.nodes.size(); i < end; ++i) {
+    const RoutingProblem::Node &node = forward.nodes[i];
+    if (!node.isUnitary)
+      continue;
+    unsigned unitarySuccessors = 0;
+    for (RoutingProblem::NodeRef s : node.successors) {
+      if (!forward[s].isUnitary)
+        continue;
+      ++unitarySuccessors;
+      // Processing the consumer in reverse makes this producer ready.
+      reverse.nodes[fwdToRev[s.index].index].successors.push_back(fwdToRev[i]);
+    }
+    // Each result wire that does not feed a unitary is a reverse-circuit input.
+    for (unsigned k = unitarySuccessors; k < node.qubits.size(); ++k)
+      reverse.sourceUsers.push_back(fwdToRev[i]);
+  }
+  return reverse;
+}
 
 /// The `SabreRouter` class is modified implementation of the following paper:
 /// Li, Gushu, Yufei Ding, and Yuan Xie. "Tackling the qubit mapping problem for
@@ -85,45 +444,46 @@ struct VirtualOp {
 /// measurement mapping until the end, which is required for QIR Base Profile
 /// programs (see the `allowMeasurementMapping` member variable).
 class SabreRouter {
-  using WireMap = DenseMap<Value, cudaq::Placement::VirtualQ>;
   using Swap = std::pair<cudaq::Placement::DeviceQ, cudaq::Placement::DeviceQ>;
+  using NodeRef = RoutingProblem::NodeRef;
 
 public:
-  SabreRouter(const cudaq::Device &device, WireMap &wireMap,
+  SabreRouter(const cudaq::Device &device, const RoutingProblem &problem,
               cudaq::Placement &placement, unsigned extendedLayerSize,
               float extendedLayerWeight, float decayDelta,
               unsigned roundsDecayReset)
-      : device(device), wireToVirtualQ(wireMap), placement(placement),
+      : device(device), problem(problem), placement(placement),
         extendedLayerSize(extendedLayerSize),
         extendedLayerWeight(extendedLayerWeight), decayDelta(decayDelta),
         roundsDecayReset(roundsDecayReset),
-        phyDecay(device.getNumQubits(), 1.0), phyToWire(device.getNumQubits()),
-        allowMeasurementMapping(false) {}
+        phyDecay(device.getNumQubits(), 1.0), allowMeasurementMapping(false) {}
 
-  /// Main entry point into SabreRouter routing algorithm
-  void route(Block &block, ArrayRef<cudaq::quake::BorrowWireOp> sources);
-
-  /// After routing, this contains the final values for all the qubits
-  ArrayRef<Value> getPhyToWire() { return phyToWire; }
+  /// Main entry point into SabreRouter routing algorithm. Walks the DAG without
+  /// modifying the IR and returns the decisions for the emitter to apply.
+  RoutingResult route();
 
 private:
-  void visitUsers(ResultRange::user_range users,
-                  SmallVectorImpl<VirtualOp> &layer,
-                  SmallVectorImpl<Operation *> *incremented = nullptr);
+  /// Visit each node in `successors` and bump its count. A node that has been
+  /// visited once per wire operand joins `layer`, or `measureLayer` if it is a
+  /// deferred measurement. `incremented` records the bumps so a lookahead can
+  /// undo them.
+  void visitSuccessors(ArrayRef<NodeRef> successors,
+                       SmallVectorImpl<NodeRef> &layer,
+                       SmallVectorImpl<NodeRef> *incremented = nullptr);
 
-  LogicalResult mapOperation(VirtualOp &virtOp);
+  LogicalResult mapOperation(NodeRef node);
 
   LogicalResult mapFrontLayer();
 
   void selectExtendedLayer();
 
-  double computeLayerCost(ArrayRef<VirtualOp> layer);
+  double computeLayerCost(ArrayRef<NodeRef> layer);
 
   Swap chooseSwap();
 
 private:
   const cudaq::Device &device;
-  WireMap &wireToVirtualQ;
+  const RoutingProblem &problem;
   cudaq::Placement &placement;
 
   // Parameters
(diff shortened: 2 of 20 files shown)


Zaneham:
Nice repro. Two isomorphic stars routing to different SWAP counts is a clean way to surface the physical-numbering artifact, and scoring the greedy layout against identity and keeping whichever wins is a sensible guard against regressions.

The thing worth raising is that the placement and the selection are optimising slightly different models. `interactionPlacement` builds the layout from an untimed interaction matrix where every two-qubit gate counts the same, while `layoutScore` weights interactions by gate order through the `numI - i` term. Because the router runs off the front layer with a lookahead window, a placement that only knows total interaction weight can seat a pair well by overall count and still badly for the gates the router reaches first. Making the placement itself front-layer aware, rather than only the final selection, is the more principled version, and it matches what the router actually rewards.

Two smaller things. The selection runs on a static distance proxy, which you flag honestly, and the stronger if costlier option is to route both candidate layouts and compare real SWAP counts. And the linear recency weighting reads as a little arbitrary without a sentence tying it to the router's own decay behaviour.

Solid work overall, and the regression-guard instinct is the right one.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -40,6 +40,149 @@ void identityPlacement(cudaq::Placement &placement) {
     placement.map(cudaq::Placement::VirtualQ(i), cudaq::Placement::DeviceQ(i));
 }
 
+/// Greedy initial placement over the circuit interaction graph. Returns a
+/// `vrToPhy` array without mutating a `cudaq::Placement`, so the caller can
+/// score it against identity and apply only the winner.
+SmallVector<unsigned>
+interactionPlacement(const cudaq::Device &device,
```
taalexander: This method is a monolith and should be refactored.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -40,6 +40,149 @@ void identityPlacement(cudaq::Placement &placement) {
...
+                     ArrayRef<SmallVector<unsigned>> interactions,
+                     ArrayRef<bool> userVirtualQubits) {
+  using Qubit = cudaq::Device::Qubit;
+  const unsigned n = device.getNumQubits();
+  SmallVector<unsigned> vrToPhy(n, 0);
+
+  // Logical weighted degree, and whether any two-qubit interaction exists.
```
taalexander: What is meant by logical here? Is this the connectivity degree of a given qubit?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -40,6 +40,149 @@ void identityPlacement(cudaq::Placement &placement) {
...
+    if (logicalDegree[u] > 0)
+      hasInteraction = true;
+  }
+
+  // No two-qubit interactions: fall back to identity.
+  if (!hasInteraction) {
+    for (unsigned v = 0; v < n; ++v)
+      vrToPhy[v] = v;
+    return vrToPhy;
+  }
```
taalexander: Is this replicating the current Cuda-Q behaviour? To me this appears to be a failure mode that should be communicated (no mapping can be found?)?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -40,6 +40,149 @@ void identityPlacement(cudaq::Placement &placement) {
...
+    physDegree[p] =
+        static_cast<unsigned>(device.getNeighbours(Qubit(p)).size());
+  }
+  auto physBetter = [&](unsigned a, unsigned b) {
+    if (distanceSum[a] != distanceSum[b])
+      return distanceSum[a] < distanceSum[b];
+    if (physDegree[a] != physDegree[b])
+      return physDegree[a] > physDegree[b];
+    return a < b;
+  };
```
taalexander: This requires further comments to describe the intent and usage of these code sections (and should be refactored).

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -40,6 +40,149 @@ void identityPlacement(cudaq::Placement &placement) {
...
+  auto bestFreePhysical = [&]() {
+    unsigned best = n;
+    for (unsigned p = 0; p < n; ++p)
+      if (!usedPhysical[p] && (best == n || physBetter(p, best)))
+        best = p;
+    return best;
+  };
+
+  // Seed the highest-degree logical qubit onto the most central physical qubit.
+  unsigned seedVirtual = n;
```
taalexander: This code should be refactored for clarity

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -711,13 +854,42 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
     SmallVector<std::size_t> userQubitsMeasured;
     DenseMap<std::size_t, Value> finalQubitWire;
     Operation *lastSource = nullptr;
+
+    // Validate the strategy before deciding whether to collect interaction
+    // data.
+    std::string placementStrategy = this->placement;
+    if (placementStrategy != "identity" && placementStrategy != "greedy") {
```
taalexander: Can we please create an enum and mapping for placement strategies.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -859,12 +1044,42 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
       }
     }
 
-    // Place
-    cudaq::Placement placement(sources.size(), deviceInstance->getNumQubits());
-    identityPlacement(placement);
+    cudaq::Placement layout(sources.size(), deviceInstance->getNumQubits());
+    if (placementStrategy == "identity") {
```
taalexander: Use enum for determining strategy.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -859,12 +1044,42 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+        return total;
+      };
+
+      ArrayRef<unsigned> chosen =
+          layoutScore(interactionCand) <= layoutScore(identityCand)
+              ? interactionCand
+              : identityCand;
+      for (unsigned v = 0; v < numV; ++v)
+        layout.map(cudaq::Placement::VirtualQ(v),
+                   cudaq::Placement::DeviceQ(chosen[v]));
```
taalexander: Why is this clause so complex and not a callout to a single function implementing the placement? Falling back to identity placement if better seems like a behaviour I would not expect naively if I selected a specific placement strategy. That might be fine for a strategy like "auto" which would select the best strategy (or if no arguments were provided).

taalexander (changes requested):
Hi @border-b, thank you for your contribution. It is a great start but will need some degree of rethinking to be ready for merge. As noted by @Zaneham in his [comment](https://github.com/NVIDIA/cuda-quantum/pull/4678#issuecomment-4648146420) the current placement, layout and router are optimizing different models. I also think we are missing larger demos/benchmarks on some quantum volume like circuits to understand that the PR is having the intended improvement.

The interaction placement builds the candidate from the total two-qubit interaction counts. This treats all interactions equally indepedent of when they happen in the circuit. Later on you use the `layoutScore` to choose between greedy and identity with a distance score. I don't think you should be silently falling back to identity if a different placement strategy was selected.

However, the bigger issue is that `chooseSwap` then does it's selection based on the dynamic front layer. This will invalidate most of the static assumptions and might lead to a situation like:

1. Circuit has many late interactions between `q0` and `q1
2. Greedy placement puts `q0` and `q1` close because of this.
3. The first front-layer gates use `q2-q3` and `q3-q4`.
4. The router adds immediate swaps
5. Swaps have not changed the placement so that original static score is invalidated

The issue is that these models are not working together. The first layout could optimize for pairs that matter often but not until late in the circuit. The second model could choose a layout based on early distances without understand the routing. Then the third model could invalidate all of these static assumptions by inserting early swaps.

I think a good direction would be to not make `greedy` the whole placement system and instead structure this as a small (internal) layout/routing system to which we could add the greedy candidate now and maybe SABRE/LightSABRE style candidates in the future (although one of these as an alternative to `greedy` is acceptable and preferred). 

An ideal approach might look something like:
1. Build a `RoutingProblem` from the IR once. Capture the device, routeable operations, source wires, measurement constraints, and virtual-qubit mapping.
2. Generate starting `LayoutCandidate`s. Keep this as a small helper layer. Identity, greedy, dense, and random layouts should only propose starting layouts. Now `greedy` is just a potential placement seed to kickoff the search.
3. Add a `RoutingSearchStrategy` which should own the search. It should route candidates in analyze mode and then run the SABRE forward/backward refinement. It's output should be the final routing selected for the router to apply.
4. Select the best `RoutingResult`. Choose by the routed SWAP count. Longer term we might want to make this selectable (eg., `swap-count`, `depth`, etc.) or at the choice of the strategy. 
5. Emit the selected result once through a `RoutingEmitter`. Rewrite the IR only after the best routed result has been selected.

I would prefer if the the reverse traversal strategy was used from the original paper as opposed to the greedy solution. In this way something like a LightSABRE extension would just add a new strategy (evaluating many candidate layouts) in the future and not a new pass rewrite.


Please feel free to ask me any clarifying questions you might have :smile:

border-b:
Thanks @taalexander and @Zaneham for reviewing and the thoughtful comments. The point about the placement, selection, and router optimizing different models is well taken. I'm going back through the SABRE paper and the proposed restructuring now, with reverse traversal as the likely direction (I'd considered it early on but went with the smaller change at the time). 

I'll follow up with some questions in the next couple of days.

taalexander:
Thank you @border-b, looking forward to seeing the new and improved version!

border-b:
@taalexander The plan below is essentially the points from your review with the details filled in. I'm writing it out mostly to verify we're on the same page before I start, plus a few questions where I'd rather ask than guess.

1. `SabreRouter::route()` currently rewires each gate the moment it's mappable and inserts every chosen swap immediately, so routing a candidate layout means changing the circuit. We'll split it: the router produces a `RoutingResult` (swap insertions, final mapping, swap count) from a read-only walk, and a `RoutingEmitter` applies the selected result to the IR once at the end. Everything else builds on this split.
2. A `RoutingProblem` gets built once from the IR: device and distance matrix, the gate dependency DAG, qubit sources, measurement constraints, and the wire-to-virtual mapping.
3. Identity and the greedy `interactionPlacement` from this PR become seed generators that only propose starting layouts. The static `layoutScore` selection goes away, and so does the silent fallback to identity.
4. For each seed: route forward in analyze mode, route the two-qubit gate DAG in reverse starting from the resulting final mapping, and take where the qubits land as the refined starting layout. One more forward pass produces that candidate's `RoutingResult`. This is the paper's forward-backward-forward setup; I'd keep the traversal count an internal constant rather than a new pass option.
5. The result with the fewest routed swaps wins and gets emitted once. The comparison stays behind a small interface so depth or other metrics can slot in later, as you suggested.

From the current PR, `interactionPlacement` survives as a seed generator and I think the tests get reworked around the new structure.

**Questions**

1. The paper starts reverse traversal from a random layout, and trying many random seeds is where a LightSABRE-style strategy would get most of its value. But randomness makes compilation nondeterministic and the exact-output tests can't pin results. For this PR should I use deterministic seeds only (identity + greedy) and leave random restarts to the future strategy? Or do you want randomness with a fixed default seed?
2. I plan to make the search the default, since the issue shows identity placement ties output quality to physical qubit numbering. The single-shot identity path stays reachable as an explicit option. Currently `placement=` option covers everything, but seeds and search strategy are now independent choices, so I'd lean toward two options (seeds, strategy) so a future LightSABRE would just be a new strategy value. Let me know if the default flip and the config option split sounds good.
3. You listed measurement constraints as part of `RoutingProblem`, which I'm interpreting as: preserve measurement order, and defer measurement mapping to the end as the router does currently. Both forward passes route the real circuit under those constraints, only the reverse pass skips measurements. Is there anything else that `RoutingProblem` should capture?
4. For the benchmark request, what would you consider sufficient for this PR: a reproducible table in the PR conversation comparing identity vs the new search on QV-style circuits across a few topologies and sizes? Or should I add a checked-in benchmark script?

taalexander:
Hi @border-b. Yes, your understanding of my proposal is correct. For your questions.

> 1. The paper starts reverse traversal from a random layout, and trying many random seeds is where a LightSABRE-style strategy would get most of its value. But randomness makes compilation nondeterministic and the exact-output tests can't pin results. For this PR should I use deterministic seeds only (identity + greedy) and leave random restarts to the future strategy? Or do you want randomness with a fixed default seed?

Yes deterministic seeds is great for now. A follow-up PR with lightSABRE like layout exploration would be great but doesn't need to go into this.

> 2. I plan to make the search the default, since the issue shows identity placement ties output quality to physical qubit numbering. The single-shot identity path stays reachable as an explicit option. Currently placement= option covers everything, but seeds and search strategy are now independent choices, so I'd lean toward two options (seeds, strategy) so a future LightSABRE would just be a new strategy value. Let me know if the default flip and the config option split sounds good.

Yes, the initial placement maybe would have an option to be fixed., eg., the identity or something else but the search by default.


> 3. You listed measurement constraints as part of RoutingProblem, which I'm interpreting as: preserve measurement order, and defer measurement mapping to the end as the router does currently. Both forward passes route the real circuit under those constraints, only the reverse pass skips measurements. Is there anything else that RoutingProblem should capture?

These may or may not exist. If the current pass doesn't use these then do not include them.

> 4. For the benchmark request, what would you consider sufficient for this PR: a reproducible table in the PR conversation comparing identity vs the new search on QV-style circuits across a few topologies and sizes? Or should I add a checked-in benchmark script?

Yes, this seems good. A plot with a script attached sweeping circuit optimization metrics (circuit depth/gates) and wall-clock compilation (ideally for the passes before/after). To ensure circuit mapping is improved with limited wall-clock cost.

border-b:
@taalexander reworked along the structure you proposed, now selecting the layout by routed swap count. 

The QV benchmark you asked for is in the description, swap reduction next to the qubit-mapping pass wall-clock, with the plot attached. Happy to iterate.

taalexander:
Thanks for your changes @border-b. I am reviewing now! Just FYI the hackathon closes June 17th so let's work hard to finish this up and get it to a state where it can be merged :rocket:

border-b:
Hi @taalexander, I've updated the description with some new benchmarks. I added a Qiskit SABRE comparison and two-qubit depth alongside the swap counts.

taalexander:
I ran this on a large VQE benchmark on a square lattice and received some very nice results thanks to the improved layout/routing implementing the proper SABRE algorithm:

| Metric | Current branch (`b558e7b369`) | PR #4678 (`057fc186db`) | Delta |
|---|---:|---:|---:|
| Wall-clock compile time | 476.27 s | 310.32 s | -165.95 s (-34.8%) |
| Routing/mapping stage | 92.65 s | 26.15 s | -66.50 s (-71.8%) |
| Routing share of wall-clock | 19.5% | 8.4% | -11.0 pp |
| `MappingFunc` pass time | 91064.83 ms | 24945.18 ms | -66119.65 ms (-72.6%) |
| `MappingFunc` share of pass-classified time | 43.2% | 22.5% | -20.7 pp |
| Input total gate count | 347680 | 347680 | 0 |
| Input 2Q gate count | 61344 | 61344 | 0 |
| Input depth | 640 | 640 | 0 |
| Output total gate count | 804764 | 471008 | -333756 (-41.5%) |
| Output 2Q gate count | 172596 | 61344 | -111252 (-64.5%) |
| Output depth | 126265 | 962 | -125303 (-99.2%) |

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+}
+
+/// Builds a deterministic topology-aware initial layout by assigning highly
+/// interacting virtual qubits to central physical qubits first. The greedy
+/// growth and its tie-breaks make the layout a deterministic function of the
+/// interaction counts and the device, so it is reproducible across runs.
+class GreedyInitialPlacer {
+public:
+  GreedyInitialPlacer(const cudaq::Device &device,
+                      ArrayRef<SmallVector<unsigned>> interactions,
```
taalexander: Interactions is really a weighted virtual-qubit interaction graph, but the `ArrayRef<SmallVector<unsigned>>` type exposes it as an anonymous dense matrix. Can we either introduce a named type/struct, or pass a sparse list of `{virtual0, virtual1, count}` edges? The current dense matrix is not clear.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+      for (unsigned q = 0; q < n; ++q)
+        distanceSum[p] += device.getDistance(Qubit(p), Qubit(q));
+      physDegree[p] =
+          static_cast<unsigned>(device.getNeighbours(Qubit(p)).size());
+    }
+  }
+
+  /// More central first: lower distance-sum, then higher degree, then lower
+  /// index.
+  bool physBetter(unsigned a, unsigned b) const {
```
taalexander: A better name for the method or comment might be needed?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+    for (unsigned u = 0; u < n; ++u)
+      if (userVirtualQubits[u] &&
+          (seed == n || weightedDegree[u] > weightedDegree[seed]))
+        seed = u;
+    return seed;
+  }
+
+  /// The unplaced virtual qubit most connected to the placed ones, falling back
+  /// to weighted degree for disconnected components.
+  unsigned chooseNextVirtual() const {
```
taalexander, line 209: `chooseNextVirtual()` is doing two selection passes: first maximizing interaction weight to the already placed set, then falling back to total weighted degree when that score is zero. Could this be written as a single candidate-scoring loop with a comparison? That would make the tie-breaking policy clearer and avoid the reset logic. It's hard to read right now.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+          pickDegree = weightedDegree[u];
+        }
+      }
+    }
+    return pick;
+  }
+
+  /// The free physical qubit minimizing weighted distance from `v` to its
+  /// placed partners, breaking ties by centrality.
+  unsigned bestPhysicalFor(unsigned v) const {
```
taalexander, line 231: Several placement helpers scan the full virtual or physical index range and then filter out already placed or used entries. Would it be clearer to maintain explicit `freePhysicals`, `placedVirtuals`, and `unplacedUserVirtuals` lists? That would make the greedy placer's state more direct and avoid repeated dense scans over the full device size.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -778,6 +1416,18 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
           for (const auto &wire : wireOperands)
             userQubitsMeasured.push_back(wireToVirtualQ[wire].index);
 
+        // Record two-qubit interactions for placement.
+        if (collectInteractions &&
+            !isa<cudaq::quake::MeasurementInterface>(op) &&
+            wireOperands.size() == 2) {
+          unsigned v0 = wireToVirtualQ[wireOperands[0]].index;
+          unsigned v1 = wireToVirtualQ[wireOperands[1]].index;
```
taalexander: What happens if the wire is missing or an unsupported operation is supplied? Could this ever default to 0?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander: It also falls back to weightedDegree only when all candidates have zero weight to the placed set. For nonzero ties, it keeps the lower-index candidate even if another candidate has much higher total weighted degree. Is that intentional?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander: What is the behaviour has no interactions with everything that is already placed? Should we instead call `bestFreePhysical`?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+    nodeIndex[&op] = RoutingProblem::NodeRef(problem.nodes.size());
+    problem.nodes.push_back(std::move(node));
+  }
+
+  // Record successor edges in use-list order, keeping the routable users only.
+  // A user is listed once per result wire it consumes, so a node's visit count
+  // reaches its wire-operand count exactly when all of its inputs are ready.
+  auto recordUsers = [&](Operation *producer,
+                         SmallVectorImpl<RoutingProblem::NodeRef> &out) {
+    for (Operation *user : producer->getUsers())
```
taalexander: This seems to rely on successor multiplicity per consumed wire, but with `producer->getUsers()` maybe a two-result op feeding both operands of the same later two-qubit op is collapsed to one?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
 //===----------------------------------------------------------------------===//
 
-/// This class encapsulates an quake operation that uses wires with information
-/// about the virtual qubits these wires correspond.
-struct VirtualOp {
+/// The dependency DAG the router walks. It is built once from the IR and never
+/// modified, so a layout can be routed without touching the circuit. Each node
+/// is a routable operation. Its successors are the operations that consume its
+/// result wires.
+struct RoutingProblem {
```
taalexander, line 332: It seems reasonable to build a separate simple DAG here for evaluating routing candidates.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+  const Node &operator[](NodeRef n) const {
+    assert(n.isValid() && "invalid node handle");
+    return nodes[n.index];
+  }
+};
+
+/// A single routing decision: a gate mapped onto physical qubits, or a swap
+/// inserted between them. The router records these as it walks the circuit and
+/// the emitter replays them to rewrite the IR.
+struct RoutingEvent {
```
taalexander, line 369: This is helpful to build up the solution candidate.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+  SmallVector<unsigned> initialLayout;
+  SmallVector<RoutingEvent> trace;
+  unsigned swapCount = 0;
 };
 
+/// Build the routing problem from `block`. The nodes are the routable
+/// operations that `isSupportedMappingOperation` accepts, other than the source
+/// borrows. Edges and source successors are captured in MLIR use-list order so
+/// the walk visits successors in the same order as the SSA use-def chains.
+RoutingProblem buildRoutingProblem(
```
taalexander, line 423: Blocking scoping makes sense to me. This might require a bit of thought to integrate #4733 with this but I would prioritize this work as is right now.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -35,25 +36,383 @@ constexpr StringRef mappedWireSetName("mapped_wireset");
...
+
+/// Build the transposed problem over the unitary gates only, for the SABRE
+/// reverse-traversal pass. Routing this forward is equivalent to routing the
+/// original circuit in reverse: a node's successors here are its forward
+/// predecessors, and result wires that do not feed another unitary (the
+/// circuit's outputs) seed the walk. Measurements, sinks, and returns are not
+/// unitary nodes, so they drop out, which is the paper's "skip measurements in
+/// the reverse pass". Readiness is unchanged: a unitary has equal operand and
+/// result arity, so its threshold is `qubits.size()` in both directions.
+RoutingProblem buildReverseProblem(const RoutingProblem &forward) {
```
taalexander, line 493: Might be a bit cleaner with a few helpers like `shouldIncludeInReverse` and `makeReverseNode`.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -155,99 +517,74 @@ class SabreRouter {
...
-            measureLayer.emplace_back(user, qubits);
-            measureLayerSet.insert(user);
-          }
-        }
-      }
+    const RoutingProblem::Node &node = problem[s];
+    if (count != node.qubits.size())
+      continue;
+    // Don't process measurements until we're ready.
+    if (allowMeasurementMapping || !node.isMeasure) {
```
taalexander, line 668: What happens if we allow mid circuit measurements and measurement mapping is not allowed? I imagine this should be an error if there are later users of the measurement and an error should be raised? I'm not sure this is the best place this should happen but please take a look.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -349,47 +686,84 @@ SabreRouter::Swap SabreRouter::chooseSwap() {
...
+    result.trace.push_back(RoutingEvent::swap(q0, q1));
+    ++result.swapCount;
   };
 
+  // Release valve: bring the closest front-layer gate together along a shortest
+  // path, ignoring the heuristic. SABRE's decay only softly discourages the
+  // local minima the heuristic can fall into, so a stuck front layer would
+  // otherwise loop forever; forcing the gate guarantees progress. This follows
+  // the release-valve idea from Qiskit/LightSABRE (arXiv:2409.08368).
+  auto forceClosestGate = [&]() {
```
taalexander: This is it's own routing policy after a bade heuristic rewinding and force routing. Could this be extracted out of the route lambda into named helpers?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -349,47 +686,84 @@ SabreRouter::Swap SabreRouter::chooseSwap() {
...
+  // direct-routing cost to explore and recover before the release valve fires.
+  // The floor keeps a usable budget on small devices, where the scaled term
+  // would otherwise be too tight.
+  constexpr unsigned minStallSwapBudget = 64;
+  constexpr unsigned stallSwapBudgetPerQubit = 4;
+  const unsigned stallSwapLimit = std::max(
+      minStallSwapBudget, stallSwapBudgetPerQubit * device.getNumQubits());
   std::size_t numSwapSearches = 0;
+  unsigned swapsSinceRouted = 0;
+  SmallVector<Swap> episodeSwaps;
```
taalexander, line 911: The existing SABRE cost knobs are already parameterized, but several of these are hardcoded? Perhaps these should also be pass arguments or hidden debug options?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -408,15 +782,36 @@ void SabreRouter::route(Block &block,
...
-    if (succeeded(mapFrontLayer()))
+    if (succeeded(mapFrontLayer())) {
+      swapsSinceRouted = 0;
+      episodeSwaps.clear();
       continue;
+    }
 
     LLVM_DEBUG(logger.getOStream() << "\n";);
 
+    if (swapsSinceRouted >= stallSwapLimit) {
```
taalexander, line 938: This release-valve path is doing a fairly important state transition inline. Could this be extracted into a named helper, for example
  `applyReleaseValve()` with a smaller `rewindEpisode()` helper?

taalexander (changes requested):
Phase 1 review.  Will continue from here. This is looking great, mostly refactoring questions but a few potential bugs.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -428,8 +823,205 @@ void SabreRouter::route(Block &block,
...
+/// applies the paper's forward-backward-forward reverse-traversal refinement,
+/// keeping both the unrefined and refined results as candidates so refinement
+/// can never select a worse layout than the seed alone. The candidate with the
+/// fewest routed swaps wins, compared through `isBetter`. No IR is touched
+/// here.
+class RoutingSearchStrategy {
+  /// Reverse-traversal initial-mapping refinement from the SABRE paper (Li et
+  /// al. 2019, Sec. IV-C2): route forward, route backward from the resulting
+  /// layout, then forward again.
+  static constexpr unsigned numTraversals = 3;
```
taalexander: Should this be a Sabre pass parameter like the others?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -428,8 +823,205 @@ void SabreRouter::route(Block &block,
...
+    auto consider = [&](RoutingResult result,
+                        const cudaq::Placement &finalPlace) {
+      if (!haveBest || isBetter(result, best.result)) {
+        best.result = std::move(result);
+        best.finalLayout = finalPlace;
+        haveBest = true;
+      }
+    };
+
+    for (ArrayRef<unsigned> seed : seeds) {
```
taalexander: Could this be extracted into a helper such as routeAndRefineSeed() or made explicit as forward, reverse, forward while numTraversals is fixed at 3? The current parity-based loop is compact, but it makes the candidate-selection policy harder to see, especially that reverse traversals only produce the next seed and are not themselves candidates.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -428,8 +823,205 @@ void SabreRouter::route(Block &block,
...
+    // permutation, so the rewritten identities stay distinct and within the
+    // device range.
+    for (auto borrowWire : sources) {
+      Value wire = borrowWire.getResult();
+      unsigned phy = result.initialLayout[wireToVirtualQ[wire].index];
+      borrowWire.setIdentity(phy);
+      phyToWire[phy] = wire;
+    }
+
+    OpBuilder builder(&block, block.begin());
```
taalexander, line 1134: The emitter creates swaps with the builder's insertion point initialized at the block.begin() and the later on it seems we use `sortTopologically` to place them. Another option to consider would be the emitter managing this insertion point to avoid having to perform an additional topological sort? 

This is just an idea to consider.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -859,15 +1509,55 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
 
-    // Place
-    cudaq::Placement placement(sources.size(), deviceInstance->getNumQubits());
-    identityPlacement(placement);
+    const unsigned numV = sources.size();
+    const unsigned numPhy = deviceInstance->getNumQubits();
+
+    // Generate the seed layouts to try, in deterministic order. Each seed only
+    // proposes a starting vrToPhy. The router decides the rest.
+    SmallVector<SmallVector<unsigned>> seeds;
```
taalexander: Could we somehow refactor this into something like `buildPlacementSeeds()` to make this code a bit easier to read?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -859,15 +1509,55 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+    }
 
-    // Route
-    SabreRouter router(*deviceInstance, wireToVirtualQ, placement,
-                       extendedLayerSize, extendedLayerWeight, decayDelta,
-                       roundsDecayReset);
-    router.route(*blocks.begin(), sources);
+    // Build the routing problem once (it does not depend on the layout), then
+    // search over the seeds for the result with the fewest swaps.
+    RoutingProblem problem =
```
taalexander, line 1784: Excellent this logic is much more straightforward now!

cudaq/test/Transforms/mapping_search_termination.qke
```diff
@@ -0,0 +1,263 @@
+// ========================================================================== //
```
taalexander, line 1: Could we find a more simple circuit that reproduces this (maybe your agent can)? It is quite long for a test.

cudaq/test/Transforms/mapping_search_termination.qke
```diff
@@ -0,0 +1,263 @@
...
+  quake.return_wire %g199#0 : !quake.wire
+  quake.return_wire %g197#1 : !quake.wire
+  quake.return_wire %g199#1 : !quake.wire
+  quake.return_wire %g198#1 : !quake.wire
+  quake.return_wire %g191#1 : !quake.wire
+  return
+}
+
+// CHECK: func.func @qv_path20() attributes
+// CHECK-SAME: mapping_v2p
```
taalexander: This only checks the termination not that the release valve is correct. Could there be a smaller release valve reproducer?

cudaq/test/Transforms/mapping_connectivity.qke
```diff
@@ -0,0 +1,40 @@
+// ========================================================================== //
+// Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                 //
```
taalexander:
```suggestion
// Copyright (c) 2026 NVIDIA Corporation & Affiliates.                 //
```

Same for all new files.

taalexander (changes requested):
Phase 2 review. A few more minor comments. Can you please make sure you that all new file copyrights are for 2026 only. 

Would appreciate a quick turnaround on this so we can get merged 🚀

border-b:
Hi @taalexander, thank you for the thorough and helpful reviews! I've been working on them and you can expect the changes pushed in the next hour or so. Sharing a couple of questions under the specific comments.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Yeah, the release-valve idea is from the LightSABRE paper but the constants here are my heuristic. I'll promote these to pass options so they can be tuned without recompiling.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: I kept this as a documented constant rather than a pass option since it's the structural forward-backward-forward of SABRE's reverse traversal. I could expose it as an advanced option too if you think it would be symmetrical with the others. Keeping this unchanged for the first pass.

taalexander: Fair enough, thank you!

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander: Great, it also may be worth adding a comment on this idea coming from LightSABRE just so the code ties back to the literature :pray:

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Thanks for pointing this out, this looked pretty tricky. I did a bit of digging into this. The mapper currently defers measurements, which is only safe for terminal readout. I added a pre-routing check (in `runOnOperation`) that preserves terminal measurement use but rejects unsupported mid-circuit/adaptive cases, matching the existing AddMetadata measurement dependence check. I also added diagnostic tests for measured wire reuse and store/load-mediated feed-forward.

border-b:
Hi @taalexander, I pushed the follow-up refactor addressing the current review threads. Sorry this took longer than I initially expected. 😅 

I verified the updated mapping tests pass, and the benchmarks match. I’ll be on standby for the next few hours to iterate if needed. Otherwise, I’ll follow up early tomorrow. Thanks again for the detailed reviews, they’ve been really helpful!

taalexander:
Thanks @border-b. Would you mind commenting how you addressed them and resolving the comments as necessary if you believe that you have :)

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
```diff
@@ -956,6 +956,16 @@ def MappingFunc: Pass<"qubit-mapping-func", "mlir::func::FuncOp"> {
...
+    Option<"minStallSwapBudget", "minStallSwapBudget", "unsigned",
+           /*default=*/"64",
+           "Release valve floor: minimum number of consecutive swaps that route "
+           "no gate before a stuck front layer is force-routed along a shortest "
+           "path (advanced)">,
+    Option<"stallSwapBudgetPerQubit", "stallSwapBudgetPerQubit", "unsigned",
+           /*default=*/"4",
+           "Release valve scaling: per-device-qubit stall budget. The budget "
+           "used is max(minStallSwapBudget, stallSwapBudgetPerQubit * numQubits) "
+           "(advanced)">,
```
taalexander: Almost all other cuda-q pass/mlir pass arguments are kebab case. I would change this and all other Mapping pass arguments to be consistent. Eg.,:
- min-stall-swap-budget
- stall-swap-budget-per-qubit
- ...

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -48,18 +51,42 @@ std::optional<PlacementStrategy> parsePlacementStrategy(llvm::StringRef name) {
       .Default(std::nullopt);
 }
 
+/// A symmetric weighted interaction graph over virtual qubits. Each entry
+/// counts the two-qubit gates acting on a pair of virtual qubits. It is stored
+/// as a dense `n x n` matrix so the placer can query any pair in O(1).
+class VirtualInteractionGraph {
```
taalexander: Is this a dense nxn matrix? I would think this could be stored sparsely (at least in the adjacency axis) for O(n).

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: This was not intentional. I rewrote `chooseNextVirtual()` as a single candidate-scoring loop. It now computes the placed-set interaction weight and total weighted degree for each unplaced user virtual qubit, then compares them directly, breaks ties by degree and then lower virtual index for determinism.

cudaq/test/Transforms/mapping_connectivity.qke
border-b: All new test files now use the 2026-only copyright.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -143,61 +179,56 @@ class GreedyInitialPlacer {
...
+    for (unsigned u : unplacedUserVirtuals) {
       unsigned placedWeight = 0;
-      for (unsigned v = 0; v < n; ++v)
-        if (placedVirtual[v])
-          placedWeight += interactions[u][v];
-      if (pick == n || placedWeight > pickWeight) {
+      for (unsigned v : placedVirtuals)
+        placedWeight += interactions.count(u, v);
+      unsigned degree = weightedDegree[u];
+      bool better = pick == n || placedWeight > pickPlacedWeight ||
```
taalexander: This might be simpler with something like:

```cpp
struct CandidateScore {
  unsigned placedWeight = 0;
  unsigned degree = 0;
  unsigned index = 0;

  bool isBetterThan(const CandidateScore &other) const {
    if (placedWeight != other.placedWeight)
      return placedWeight > other.placedWeight;
    if (degree != other.degree)
      return degree > other.degree;
    return index < other.index;
  }
};
```

cudaq/test/Transforms/mapping_search_termination.qke
border-b: Replaced the long termination circuit with a small `path(6)` reproducer. The test runs the same circuit with a one-swap stall budget and with a high budget. On this circuit, the forced route uses five swaps where the unrestricted heuristic uses three. CircuitCheck confirms both outputs are correct mappings.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Extracted the release-valve transition into helpers like: `addSwap()`, `rewindEpisode()`, `forceClosestGate()`, and `applyReleaseValve()`. `route()` now just detects the stall-budget condition and calls `applyReleaseValve(episodeSwaps)`

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Added `minStallSwapBudget` and `stallSwapBudgetPerQubit` as pass options. The router now computes the stall budget as `max(minStallSwapBudget, stallSwapBudgetPerQubit * numPhysicalQubits)`, with defaults `64` and `4` to preserve the loose default behavior.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -237,20 +270,68 @@ class GreedyInitialPlacer {
...
 
+/// Generate the seed layouts to try, in deterministic order. Each seed only
+/// proposes a starting vrToPhy. The router decides the rest. `interactions` is
+/// required for the greedy strategies and ignored for identity.
+SmallVector<SmallVector<unsigned>>
+buildPlacementSeeds(PlacementStrategy strategy, unsigned numV,
+                    const cudaq::Device &device,
+                    const std::optional<VirtualInteractionGraph> &interactions,
+                    ArrayRef<bool> userVirtualQubits) {
+  auto identitySeed = [&]() {
```
taalexander: Nit - these are only used once? Might be able to just put inline.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: I split this into `shouldIncludeInReverse()` and `makeReverseNode()`. The reverse-problem construction now first decides which nodes belong in the reverse walk, then builds the reversed nodes/edges with those helpers.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -481,6 +602,28 @@ class SabreRouter {
...
+  void rewindEpisode(SmallVectorImpl<Swap> &episodeSwaps);
+
+  /// Release valve for a stalled front layer. SABRE's decay only softly
+  /// discourages the local minima the heuristic can fall into, so a stuck
+  /// front layer would otherwise loop forever. Discard the current episode's
+  /// swaps and force the closest gate together so the walk always makes
+  /// progress. The decay state is left as is. It is a soft heuristic and resets
+  /// on its own cycle. This follows the release-valve idea from Qiskit and
+  /// LightSABRE (arXiv:2409.08368).
+  void applyReleaseValve(SmallVectorImpl<Swap> &episodeSwaps);
```
taalexander: Prefer just tying reference to LIghtSABRE.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -686,6 +833,51 @@ SabreRouter::Swap SabreRouter::chooseSwap() {
   return candidates[minIdx];
 }
 
+void SabreRouter::addSwap(cudaq::Placement::DeviceQ q0,
```
taalexander, line 825: Thank you this whole section is much better.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1451,6 +1755,18 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
       return;
     }
 
+    // Measurement deferral is only safe for terminal readout. Reject
+    // unsupported mid-circuit/adaptive uses before mutating IR.
+    if (auto unsupported = findUnsupportedMeasurement(func)) {
```
taalexander: This should probably be an unconditional failure. The mapper cannot preserve this program semantics because it defers measurements to the end, so silently returning in the default path leaves the IR partially prepared and unmapped. Could we emit the diagnostic and signal pass failure for unsupported mid-circuit/adaptive measurement use regardless of `nonComposable`?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: `GreedyInitialPlacer` now maintains explicit placement worklists. `bestFreePhysical()` now iterates only `freePhysicals`, and `chooseNextVirtual()` / `bestPhysicalFor()` iterate `unplacedUserVirtuals` and `placedVirtuals`. `place()` keeps those lists in sync, so the helpers no longer scan the full virtual/physical ranges and filter out used entries.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -914,6 +1048,37 @@ class RoutingSearchStrategy {
...
+    consider(routeSeed(seed, numV, numPhy, finalPlace), finalPlace);
+    if (!refine)
+      return;
+
+    // The remaining traversals alternate reverse and forward. A reverse pass
+    // only refines the current layout into the next seed. The forward pass that
+    // follows routes that seed and is the actual candidate.
+    SmallVector<unsigned> current(seed.begin(), seed.end());
+    cudaq::Placement currentFinal = finalPlace;
+    for (unsigned t = 2; t <= numTraversals; ++t) {
```
taalexander: Since `numTraversals` is fixed at 3, could this be written explicitly as forward, reverse, forward instead of using parity over a loop? The `% 2` logic is correct, but it makes the policy look more general than it is, and a future change to `numTraversals` could silently change candidate-selection semantics.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1186,6 +1356,107 @@ struct MappingPrep : public cudaq::opt::impl::MappingPrepBase<MappingPrep> {
...
+};
+
+/// The first measurement in `func` whose results are used in a way the mapper
+/// cannot support, or nullopt when every measurement is a terminal readout.
+/// Because the mapper defers measurements to the end of the circuit, it only
+/// supports measured wires that flow to a terminal consumer (`return_wire` or
+/// `sink`) and measurement results that do not feed control flow, quantum
+/// dataflow, or calls.
+std::optional<UnsupportedMeasurement>
+findUnsupportedMeasurement(func::FuncOp func) {
```
taalexander: This duplicates a fair amount of the measurement-dependence analysis from `AddMetadata.cpp`, including `seekAllocaFrom()` and  store/load forwarding. Could we reuse `cudaq::quake::detail::QuakeFunctionAnalysis` for the classical feedback part, and keep only the mapper-specific terminal measured-wire check here?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1391,16 +1661,36 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
           return;
         }
 
+        // Get the wire operands and their virtual qubits. If a supported op
+        // consumes an untracked wire, some earlier unsupported operation must
+        // have produced it. Do not let DenseMap default that to virtual qubit 0.
+        auto wireOperands = cudaq::quake::getQuantumOperands(&op);
+        SmallVector<cudaq::Placement::VirtualQ, 2> virtualOperands;
```
taalexander: Might be simplified with a helper ` lookupVirtualOperands` to consolidate the loop.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1411,32 +1701,46 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
 
         // Map the result wires to the appropriate virtual qubits.
-        for (auto &&[wire, newWire] : llvm::zip_equal(
-                 wireOperands, cudaq::quake::getQuantumResults(&op))) {
-          // Don't use wireToVirtualQ[a] = wireToVirtualQ[b]. It will work
-          // *most* of the time but cause memory corruption other times because
-          // DenseMap references can be invalidated upon insertion of new pairs.
-          wireToVirtualQ.insert({newWire, wireToVirtualQ[wire]});
-          finalQubitWire[wireToVirtualQ[wire].index] = newWire;
+        auto wireResults = cudaq::quake::getQuantumResults(&op);
```
taalexander: Another candidate for a helper like `recordQuantumResults` to simplify.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Nice idea. I considered this, but left the emitter as trace replay plus `sortTopologically()` for this PR because changing insertion-point management would be a separate cleanup and would add more risk/time. The current path is covered by the mapping tests, but I agree managing insertion points directly could be a good follow-up cleanup if you prefer.

taalexander (changes requested):
Thank you very much for your changes. We're almost there! A few more comments. If you address these I will get it merged in the next day or two 🚀 and you will be eligible for the bounty still (I've sourced a bit more time). Can you please give me push access to your branch just in case?

border-b:
Thanks @taalexander. Maintainer edits are already enabled on this branch. I’ll revisit this tomorrow morning and address the remaining comments!

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Good point. I originally used dense storage for simple O(1) pair lookups in the greedy placer, but after revisiting the scoring loops, sparse adjacency feels like a better fit here. We only need to iterate over actual interaction neighbors, and cached weighted degrees avoid full scans.  I’m changing this to sparse per-virtual adjacency storage.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Dropped the Qiskit mention so the comment references LightSABRE only

cudaq/include/cudaq/Optimizer/Transforms/Passes.td
border-b: Renamed the Mapping pass arguments to kebab-case, including `min-stall-swap-budget` and `stall-swap-budget-per-qubit`. I updated both the generated pass options in `Passes.td` and the manual `qubit-mapping` pipeline options in `Mapping.cpp`.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Done.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Moved the result-wire mapping into a `recordQuantumResults` helper, including the operand count check.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Pulled the operand-resolution loop into a `lookupVirtualOperands` helper that returns the virtual qubits or nullopt. The scan body now just calls it and bails on nullopt.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Unrolled into explicit forward, reverse, and forward steps instead of the parity loop. I also removed the `numTraversals` constant so the code just directly shows the policy.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Reused `cudaq::quake::detail::QuakeFunctionAnalysis` for the measurement dependent behavior and kept only the mapper specific terminal measured-wire validation in `Mapping.cpp`. This removes the duplicated `seekAllocaFrom` and store-load forwarding.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Removed the local seed lambdas and the single-use `interactionPlacement` wrapper. The seed list is now built directly from the selected placement strategy, with the same order and duplicate-greedy suppression for `auto`.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
border-b: Made unsupported mid-circuit or adaptive measurement use an unconditional pass failure.

taalexander:
> /ok to test dec2106

taalexander:
Thank you for addressing my comments @border-b, the PR is in very good shape now with one minor issue and some CI work to get it passing that I can guide along. I have marked this as `unitaryhack-accepted` and will approve/merge it as soon as I have CI pipelines passing (which requires my approvals to run) and I set up some benchmarks to run internally.

It has been a pleasure working with you on this. This came together as a very nice improvement for the project

taalexander:
> /ok to test b167553

taalexander (approved):
LGTM, thank you very much for your unitaryHACK contribution @border-b. You went above and beyond.

taalexander:
> /ok to test 822d0d1

border-b:
Thank you @taalexander for your patience and for guiding me throughout this PR. I've had an excellent learning experience and really enjoyed working on this over the last couple of weeks! Your feedback was incredibly helpful and made the change much better, and it's been a pleasure working with you on it.

taalexander:
> /ok to test c6ac021

taalexander:
> /ok to test 48f2325

PR #4680, opened 2026-06-05
Add kernel math function support

### LINKED ISSUE #2942: Mathematical functions support
### Required prerequisites

### Describe the feature

Now we only supports acos and sqrt in C++.
https://github.com/NVIDIA/cuda-quantum/blob/df4609e2287020f5bbe88dbaf01b6d16f252231b/targettests/execution/test-3.cpp#L17

This request would extend this to Python and also allow them to work (at least in our simulators).

* [ ] asin, acos, atan
* [ ] sin, cos, tan
* [ ] sqrt
* [ ] exp, log
* [ ] ceiling, floor
* [ ] mod
* [ ] pop count

LubuSeb:
Closes #2942.

## Summary
- add C++ AST lowering for `asin`, `acos`, and `atan`, including float overloads
- add Python AST bridge lowering for real-valued NumPy `tan`, `asin`, `acos`, `atan`, and `log`
- add focused C++ FileCheck and Python kernel tests for the newly covered math functions

## Reproduction / scope
The bounty issue asks for continuous mathematical function support in CUDA-Q kernels. Existing C++ lowering already handled several functions through `visitMathLibFunc`; this change extends that same path for inverse trig functions. Python lowering already mapped selected NumPy calls through the AST bridge; this change adds the missing real-valued operations there while leaving untested complex inverse trig/log/tan support out of scope.

## Minimality
This is limited to the existing lowering tables/switches and targeted tests. It does not refactor the frontend, broaden discrete-control support, or add backend-specific behavior.

## Validation
- `python -m py_compile python/cudaq/kernel/ast_bridge.py python/tests/kernel/test_kernel_float.py`
- `git diff --cached --check`
- Local standalone `<cmath>` syntax smoke check for the float overload names used by the new C++ test

Not run locally: CUDA-Q native lit/runtime tests, because the local environment does not have the CUDA-Q compiler/test toolchain installed.

## AI assistance disclosure
I used AI assistance to inspect the codebase and draft parts of this change; I reviewed the final patch and validation results before submission.


sacpis:
> /ok to test 157f60f

LubuSeb:
Pushed `5f79e6a` to address the Python CI failure from the previous tested head: the tests now use NumPy's current inverse-trig spellings (`arcsin`, `arccos`, `arctan`) while the bridge still accepts the existing `asin`/`acos`/`atan` aliases.

Local validation on the latest head:

```bash
python -m py_compile python/cudaq/kernel/ast_bridge.py python/tests/kernel/test_kernel_float.py
```

The latest head currently only has DCO, so it should be ready for another `/ok to test` when you have a chance.

sacpis:
> /ok to test 5cc9d4a

cudaq/test/AST-Quake/math_functions.cpp
```diff
@@ -0,0 +1,56 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
```
schweitzpgi:
```suggestion
 * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
```

schweitzpgi (approved):
LGTM

schweitzpgi:
> /ok to test a4daeb0

LubuSeb:
Pushed `f8cc1de3c` to address the remaining ARM Python CI failure on the latest tested head.

Root cause from the failed job: `test_cudaq_uccsd1` now allows either 5 or 6 sampled states, but still asserted all six specific states were present. The failed ARM run hit the valid 5-state case and missed `00000011`, so the membership assertion was stricter than the preceding `len(counts) == 6 or len(counts) == 5` guard.

The follow-up keeps the test scoped to the known UCCSD output support: observed states must still be a 5-or-6-sized subset of the six expected bitstrings. No math lowering code changed in this commit.

Local validation available in this Windows checkout:

```bash
python -m py_compile python/tests/kernel/test_kernel_uccsd.py python/cudaq/kernel/ast_bridge.py python/tests/kernel/test_kernel_float.py
git diff --check
```

I could not run the CUDA-Q runtime pytest locally because this checkout does not have a built `cudaq.kernel` runtime installed.

The latest head only has DCO again, so it should need another `/ok to test` for NVIDIA runners.

schweitzpgi:
> /ok to test 19a14bb

schweitzpgi:
> /ok to test aa7d30b0

PR #4682, opened 2026-06-05
[device_call] Add cpu_roce: a CPU-RDMA channel for cc.device_call

cketcham2333:
## Summary
Adds **`cpu_roce`**, a `DeviceCallChannel` that carries `cc.device_call` RPCs
over a pure-CPU RoCEv2 RDMA transport, landing request/response payloads
directly in host memory at microsecond latency. It gives a fully GPU-free
`device_call` path for CPU-based decoders, and its wire format is compatible
with a real HSB-enabled FPGA on the other end.
This builds on `cc.device_call` (#4565) and the merged CPU transport
(`CpuRoceTransceiver` + `CUDAQ_DISPATCH_HOST_CALL`); it is the device_call
runtime channel on top of that transport, plus the transport refinements and
ABI change the channel required.
## What's included
**Runtime channel**
- `CpuRoceChannel` (registered as `cpu_roce`): the device_call caller end. It
  RDMA-Writes requests into the service's rx ring and receives the service's
  Sends as responses. Connection identities (QP/rkey/RoCE-IP) are exchanged via
  a minimal bidirectional TCP rendezvous between the transceiver's `setup()` and
  `connect()` phases — no Hololink/HSB dependency is pulled into the runtime.
  v1 serializes response-bearing dispatch so the service's FIFO rx slot tracks
  the caller slot; fire-and-forget is async with a drain-on-reuse so a late
  (ignored) response can't poison a later call.
**Transport refinements (`CpuRoceTransceiver`)**
- `setup()`/`connect()` split for the two-phase RDMA rendezvous; `set_local_ip()`
  to select the source GID by configured IPv4 on multi-IP ports.
- TX modes renamed to name the verb the transmitter issues (`kRdmaSend` /
  `kRdmaWriteWithImm`) rather than a peer role.
- Wire direction set to FPGA-as-caller (caller Writes requests, service Sends
  responses), mirroring a real FPGA<->bridge decode loop.
- Hardening: SQ backpressure / scaled signal cadence, constructor frame-size
  validation, and improved error reporting.
**HOST_CALL ABI**
- `cudaq_host_rpc_fn_t` is now two-pointer `(const void *rx_slot, void *tx_slot,
  size_t)`: the handler reads the request from the RX slot and writes the
  response straight into the TX slot, eliminating the dispatcher's per-message
  RX->TX copy.
**Test harness (test-only)**
- `cpu_roce_test_daemon`: a service daemon running the libcudaq-realtime host
  dispatcher with `CUDAQ_DISPATCH_HOST_CALL` handlers, standing in for a
  decoder/FPGA.
- `CpuRoceChannelTester` (`CpuRoceDispatchTest`): synchronous, fire-and-forget,
  concurrency, latency, service-restart, and a fire-and-forget poisoning
  regression test (skipped unless loopback topology env vars are set).
- `cpu_roce_device_call_app` + `--app` mode: a real `__qpu__`
  `cudaq::device_call` app, `nvq++ -frealtime-lowering` compiled, that proves the
  compiler-lowered path end-to-end over the channel.
- `cpu_roce_device_call_test.sh` orchestration.
**Docs**
- New "Device Call Channels" and "CPU RoCE Transport" pages; correct the Host
  API docs (HOST_CALL is processed on the host dispatch path; two-pointer ABI).
## Testing
- `cpu_roce` device_call loopback: 6/6 (p50 ~6 µs full round trip), including the
  fire-and-forget poisoning regression.
- `--app` (compiler-lowered `cudaq::device_call`) returns 42 over `cpu_roce`.
- Full `cmake --build` and the existing `test_device_call_dispatch` pass.
## Notes
- The cpu_roce tests and the `hsb_bridge_cpu` harness require a ConnectX NIC (or
  two RoCE ports / a loopback) and are **not run in CI**: the GoogleTest cases
  `GTEST_SKIP` unless the loopback topology env vars are set, and the compiled
  `--app` proof is driven only by the orchestration script.


.github/pre-commit/spelling_allowlist.txt
```diff
@@ -291,6 +304,7 @@ integrators
 interferometer
 interoperational
 intrinsics
+iova
```
schweitzpgi: Is this an English word? AI says it is not.

schweitzpgi (commented):
From the title I was curious what the change was to the core compiler. It doesn't look like there is any change. heh

1tnguyen (approved):
LGTM 👍

.github/pre-commit/spelling_allowlist.txt
cketcham2333: Resolved in commit #465389e4989bd99f2e34251895d77ca77be4e9be

PR #4686, opened 2026-06-06
Reduce unnecessary warning messages about noise

bmhowe23:
There is no need to spam the console or log file with so many warnings about noise not being supported on a simulator when the user is not commanding any noise to be injected.

### DIFF
--- a/runtime/nvqir/CircuitSimulator.h
+++ b/runtime/nvqir/CircuitSimulator.h
@@ -872,6 +872,23 @@ protected:
                                  const std::vector<std::size_t> &controls,
                                  const std::vector<std::size_t> &targets,
                                  const std::vector<double> &params) {
+    // Do nothing if no execution context
+    if (!cudaq::getExecutionContext())
+      return;
+
+    // Do nothing if no noise model
+    auto noiseModel = getNoiseModel();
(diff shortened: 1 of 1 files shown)


sacpis (approved):
LGTM. Thanks @bmhowe23.

Renaud-K:
There needs to be a test for a change of this nature. The test has to check that indeed the warning is only issued once. Without a test these issues are likely to creep back in as we refactor the runtime. 
Thank you.

bmhowe23:
> There needs to be a test for a change of this nature.

Do we have any other tests for warning messages that we could use as an example here? I searched for tests for this warning message, and I couldn't find any.

> The test has to check that indeed the warning is only issued once.

The behavior change was to not throw warnings when users did not specify any noise in their noise model, not to ensure that warnings were only issued once. The early bailouts were consistent with the CUDA-Q simulators that do support noise (e.g. stim and cutensornet).

Renaud-K:
I was not sure what a good test would be so I had to do a bit of research. lit tests are more suited than gtests. We have a few tests that use gtest `internal::GetCapturedStdout()` but that is an API not to use according to Google as it is internal. So targettest/execution seemed like a good fit.  #4701

Issue #4479, opened 2026-05-07
[C++] [bridge] `measure_handle`: definite-assignment analysis


khalatepradnya:
## Background
PR #4409 introduced a static `isBoundHandle` walk in `lib/Frontend/nvqpp/ConvertExpr.cpp` that diagnoses
`bool b = h;` when `h` was default-constructed. The check is intentionally narrow and has two known unsoundness gaps, documented inline as "defer to a follow-up.

## Gap 1: function arguments
```cpp
__qpu__ bool f(cudaq::measure_handle h) { return h; }
```
`isBoundHandle` short-circuits on `BlockArgument` SSA values (if (!load) return true;). The bridge silently accepts the discriminate even though the caller may pass a default-constructed handle. Solving requires either per-callee verification at every call site that the argument is bound, or a runtime sentinel check at every `quake.discriminate`.

## Gap 2: aggregate / array per-element tracking
For `measure_handle hs[N];` and `struct Holder { measure_handle h; };`, any store anywhere into the base alloca is treated as binding regardless of which element / member is loaded. Per-cc.compute_ptr-index reasoning needs SSA-value index resolution, deferred from PR #4409.

PR #4688, opened 2026-06-06
fix(nvqpp): insert quake.discriminate for `for (bool b : mz(reg))`

### LINKED ISSUE #4562: `cuda-quake` mishandles measure results, causing compiler crash
### Required prerequisites

### Describe the bug

`cudaq-quake` fails to emit a `quake.discriminate` to map measure result type to bool when performing range based for loop.

Build from source and run the circuit below 
```C++
struct main_circuit { 
uint64_t operator()() __qpu__ {
	cudaq::qarray<2>reg_1190;

        int64_t packed_result = 0; int bit_offset = 0;	
	
	for (bool b : mz(reg_1190)){
               if (b) packed_result |= (1ULL << bit_offset);
               bit_offset++;
        }

	return packed_result;
}
};
```
and get a compiler crash, stemming from `cudaq-quake`. 
`loc("test.cpp":10:2):error: 'cc.store' op failed to verify that type of value matches element type of pointer`

Line 10 being `for (bool b : mz(reg_1190)){`

The MLIR that contains that `cc.store` is
```c
%18 = "cc.alloca"() <{elementType = i1}> : () -> !cc.ptr<i1>
%19 = "cc.load"(%17) : (!cc.ptr<!cc.measure_handle>) -> !cc.measure_handle
"cc.store"(%19, %18) : (!cc.measure_handle, !cc.ptr<i1>) -> ()
```

The `cc.store` tries map something of type `!cc.measure_handle` to `!cc.ptr<i1>` which is ptr to a bool. 

Running the code but with `auto` instead of `bool` works. The MLIR at the relevant section is
```c
cc.store %9, %8 : !cc.ptr<!cc.measure_handle> loc(#loc6)
%10 = cc.load %8 : !cc.ptr<!cc.measure_handle> loc(#loc8)
%11 = quake.discriminate %10 : (!cc.measure_handle) -> i1 loc(#loc8)
```

`quake.discriminate` added here maps the measure result to the correct type.

### Expected behavior

Compiler should not crash

Not a regression

### Environment

- **CUDA-Q version**: b04dee5177
- **Python version**: 3.12.3
- **C++ compiler**: clang++-22

### Suggestions

_No response_

aryanputta:
### Summary

Fixes #4562.

A range-based `for` loop with a `bool` loop variable over a multi-qubit `mz` result crashes `cudaq-quake` in the verifier:

```
'cc.store' op failed to verify that type of value matches element type of pointer
```

### Root cause

In `QuakeBridgeVisitor::TraverseCXXForRangeStmt` (`lib/Frontend/nvqpp/ConvertStmt.cpp`), `mz(reg)` lowers to `!cc.stdvec<!cc.measure_handle>`, so the container element type is `!cc.measure_handle`, not `i1`. The non-reference / non-const "create a local copy" path loads the element and stores it into the loop variable's slot:

```cpp
Value atOffset = cc::LoadOp::create(builder, loc, addr);   // !cc.measure_handle
if (isBool)                                                // false: eleTy is measure_handle, not i1
  atOffset = cc::CastOp::create(..., builder.getI1Type(), atOffset);
cc::StoreOp::create(builder, loc, atOffset, iterVar);      // store measure_handle into !cc.ptr<i1>
```

For `for (bool b : mz(reg))` the loop variable slot is `!cc.ptr<i1>`, so the store has mismatched value/pointer types and the verifier rejects the module. `for (auto b : mz(reg))` works because the slot is `!cc.ptr<!cc.measure_handle>` and the discriminate is inserted later at the `operator bool()` use site.

### Fix

When the loaded value is a `!cc.measure_handle` and the loop variable slot is `i1`, lower the handle through `quake.discriminate` before the store, mirroring the single sanctioned coercion in `measure_handle::operator bool()` (`ConvertExpr.cpp`). The `auto` (handle-typed) binding is unchanged.

### Test

Adds `test/AST-Quake/range_for_measure.cpp` covering both the `bool` binding (discriminate inserted, stores `i1`) and the `auto` binding (handle retained). Pre-fix, `cudaq-quake` crashes on the `bool` case, so the `CHECK-LABEL` alone fails without the fix.

### DIFF
--- a/cudaq/lib/Frontend/nvqpp/ConvertStmt.cpp
+++ b/cudaq/lib/Frontend/nvqpp/ConvertStmt.cpp
@@ -226,9 +226,27 @@ bool QuakeBridgeVisitor::TraverseCXXForRangeStmt(clang::CXXForRangeStmt *x,
             }
             auto iterVar = popValue();
             Value atOffset = cc::LoadOp::create(builder, loc, addr);
-            if (isBool)
+            if (isBool) {
               atOffset = cc::CastOp::create(builder, loc, builder.getI1Type(),
                                             atOffset);
+            } else if (isa<cc::MeasureHandleType>(atOffset.getType())) {
+              // `for (bool b : mz(reg))` binds an `i1` loop variable to a
+              // container of `!cc.measure_handle`. Mirror
+              // `measure_handle::operator bool()` and lower the handle through
+              // `quake.discriminate` before storing into the `i1` slot.
+              // Without it the `cc.store` value type (`!cc.measure_handle`) and
+              // pointer element type (`i1`) disagree and the verifier rejects
+              // the module. A `measure_handle` loop variable
+              // (`for (auto b : mz(reg))`) keeps the handle and is unaffected.
+              //
+              // TODO(NVIDIA/cuda-quantum#4479): this does not diagnose
+              // discriminating an unbound handle. Definite-assignment analysis
+              // for `measure_handle` is tracked there.
+              if (auto iterPtrTy = dyn_cast<cc::PointerType>(iterVar.getType()))
+                if (iterPtrTy.getElementType() == builder.getI1Type())
+                  atOffset = cudaq::quake::DiscriminateOp::create(
+                      builder, loc, builder.getI1Type(), atOffset);
+            }
             cc::StoreOp::create(builder, loc, atOffset, iterVar);
           }
         }
--- a/cudaq/test/AST-Quake/range_for_measure.cpp
+++ b/cudaq/test/AST-Quake/range_for_measure.cpp
@@ -1,0 +1,56 @@
+/*******************************************************************************
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+// RUN: cudaq-quake %s | FileCheck %s
+
+#include <cudaq.h>
+
+// A range-based for loop with a `bool` loop variable over the result of a
+// multi-qubit `mz` iterates a container of `!cc.measure_handle`. The bridge
+// must lower each per-iteration handle through `quake.discriminate` before
+// storing it into the `i1` slot of the loop variable. Without it the bridge
+// emitted `cc.store %handle, %i1ptr`, whose value/pointer types disagree, and
+// the verifier crashed `cudaq-quake`.
+
+struct range_for_measure_bool {
+  bool operator()() __qpu__ {
+    cudaq::qarray<2> reg;
+    bool any = false;
+    for (bool b : mz(reg)) {
+      any = any || b;
+    }
+    return any;
+  }
+};
+
+// clang-format off
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__range_for_measure_bool()
+// CHECK:           quake.mz %{{.*}} : (!quake.veq<2>) -> !cc.stdvec<!cc.measure_handle>
+// CHECK:           %[[HANDLE:.*]] = cc.load %{{.*}} : !cc.ptr<!cc.measure_handle>
+// CHECK:           %[[BIT:.*]] = quake.discriminate %[[HANDLE]] : (!cc.measure_handle) -> i1
+// CHECK:           cc.store %[[BIT]], %{{.*}} : !cc.ptr<i1>
+// clang-format on
+
+// A `measure_handle` loop variable (`auto`) keeps the handle and must NOT have
+// a discriminate inserted at bind time; the discriminate happens at the
+// `operator bool()` use site instead.
+
+struct range_for_measure_auto {
+  bool operator()() __qpu__ {
+    cudaq::qarray<2> reg;
+    bool any = false;
+    for (auto b : mz(reg)) {
+      any = any || b;
+    }
+    return any;
+  }
+};
+
+// CHECK-LABEL:   func.func @__nvqpp__mlirgen__range_for_measure_auto()
+// CHECK:           quake.mz %{{.*}} : (!quake.veq<2>) -> !cc.stdvec<!cc.measure_handle>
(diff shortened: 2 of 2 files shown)


schweitzpgi (commented):
LGTM. Thanks for the bug fix.

khalatepradnya:
> /ok to test 3ec0720

cudaq/lib/Frontend/nvqpp/ConvertStmt.cpp
```diff
@@ -226,9 +226,23 @@ bool QuakeBridgeVisitor::TraverseCXXForRangeStmt(clang::CXXForRangeStmt *x,
...
+              // container of `!cc.measure_handle`. Mirror
+              // `measure_handle::operator bool()` and lower the handle through
+              // `quake.discriminate` before storing into the `i1` slot.
+              // Without it the `cc.store` value type (`!cc.measure_handle`) and
+              // pointer element type (`i1`) disagree and the verifier rejects
+              // the module. A `measure_handle` loop variable
+              // (`for (auto b : mz(reg))`) keeps the handle and is unaffected.
+              if (auto iterPtrTy = dyn_cast<cc::PointerType>(iterVar.getType()))
+                if (iterPtrTy.getElementType() == builder.getI1Type())
+                  atOffset = cudaq::quake::DiscriminateOp::create(
```
khalatepradnya, line 247: This inserts discriminate without the bound-handle check.

Please add the check similar to `isBoundHandle` - okay to refactor to reuse the logic, and add a negative test like below (or equivalent) in `cudaq/test/AST-error/measure_handle.cpp`
```
struct RangeForUnboundVector {
  void operator()() __qpu__ {
    std::vector<cudaq::measure_handle> handles(1);
    // expected-error@+1{{discriminating an unbound measure_handle}}
    for (bool b : handles)
      (void)b;
  }
};
```

khalatepradnya (commented):
Thank you, @aryanputta for the contribution! This fixes a bug, but the code changes can be polished IMO.

aryanputta:
@khalatepradnya Added the bound-handle check and the negative test...thank you for the review!
**One idea I wanted to float for future cleanup** (happy to do it in a follow-up or here if you prefer):

The `storeDominatesLoad` + `isBoundHandle` pair could move from file-static in `ConvertExpr.cpp` into `ASTBridge.h` as `inline` free functions under `namespace cudaq::detail`. That would:
- Drop the cross-TU forward declaration in `ConvertStmt.cpp`
- Make the helpers naturally available to any future `Convert*.cpp` that needs them
- Keep the header change contained since `ASTBridge.h` is only included by 5 files in `lib/Frontend/nvqpp/`

The required additions to `ASTBridge.h` would be `#include "cudaq/Optimizer/Dialect/Quake/QuakeOps.h"` and `#include "llvm/ADT/SmallPtrSet.h"`. Does that refactor direction sound good to you, or would you prefer a different approach?

khalatepradnya:
@aryanputta : Thanks, for the updates. I will review today.
Can you run the code formatter to fix the CI failure on `Basic content checks`? I find it easiest to run the pre-commit checks as `pre-commit run --all-files --hook-stage pre-push`

khalatepradnya:
> /ok to test 82e7844

aryanputta:
Hi @khalatepradnya, the PR is updated at `9afb8afa`. Two issues from the last CI run are fixed:

All four commits have `Signed-off-by`. Ready for `/ok to test 9afb8afa` when you get a chance.

khalatepradnya:
> Hi @khalatepradnya, the PR is updated at `9afb8afa`. Two issues from the last CI run are fixed:
> 
> All four commits have `Signed-off-by`. Ready for `/ok to test 9afb8afa` when you get a chance.

Thanks, @aryanputta for the updates.

I don't think removing the negative test is the right resolution on its own. The test was doing its job by exposing that the range-for guard is ineffective. `atOffset` is loaded through `cc.stdvec_data`, and `isBoundHandle` returns true whenever it can't peel the pointer back to a scalar `cc.alloca`. Since a range-for over measurement handles is always iterating a `!cc.stdvec`, that fall-through is hit on every reachable path, so the guard never fires and the `reportClangError` branch is dead code.

Two coherent ways forward:

1. Keep the guard → make it real. `isBoundHandle` would need to peel `cc::StdvecDataOp` and decide boundedness from the vector's provenance (produced by mz/mx/my, directly or via a local store, vs. default/sized-constructed). Then the negative test stays.

2. Defer to #4479 → remove the dead code. Revert to the plain `quake.discriminate` insertion and also drop the un-static on `isBoundHandle` and the cross-TU forward declaration, with a `// TODO(#4479)` note. As written, the `isBoundHandle` call in `ConvertStmt.cpp` is misleading: it reads as protecting the range-for vector case it was added for, but it never does.

For a fix scoped to #4562, I'd lean toward the second.

khalatepradnya:
Additional note: If you "Update branch" / align with main, the CI is more likely to not have cache misses and tends to go faster.

aryanputta:
Thanks @khalatepradnya. Went with option 2, scoped to #4562.

What changed at `37fac382`:
- Removed the `isBoundHandle` guard and the `reportClangError` branch from the range-for path. As you noted, `atOffset` comes through `cc.stdvec_data`, so `isBoundHandle` always hit the non-alloca fall-through and returned true, making that branch dead code.
- Reverted the un-`static` on `isBoundHandle` in `ConvertExpr.cpp`, so that file is back to its main state (zero net change).
- Dropped the cross-TU forward declaration and the `llvm/ADT/SmallPtrSet.h` include in `ConvertStmt.cpp`.
- Left a `// TODO(NVIDIA/cuda-quantum#4479)` at the insertion site noting that discriminating an unbound handle is not diagnosed here, deferred to the definite-assignment work.

The range-for body is now just the plain `quake.discriminate` insertion, gated on the loaded element being a `!cc.measure_handle` and the loop-variable slot being `i1`. The `auto` binding is untouched.

Also rebased the branch onto current `main` for the cache hits you mentioned, and squashed to a single commit. clang-format (v22) passes clean on `ConvertStmt.cpp`.

Ready for `/ok to test 37fac382` when you get a chance.

khalatepradnya:
> /ok to test 37fac38

khalatepradnya (approved):
LGTM. Thanks, @aryanputta for the fix!

PR #4690, opened 2026-06-07
test: move unittests/Optimizer to cudaq/unittests (reslove #4684)

### LINKED ISSUE #4684: move unittests/Optimizer/* to cudaq/unittests
We want to link cudaq/test/Unit/ to cudaq/unittests/ exactly like llvm-project does with llvm/test/Unit and llvm/unittests.

zoldyck13:
### Description:
This PR moves the Optimizer unit tests from `unittests/Optimizer/` to the `cudaq/unittests/` directory to align with the project's evolving directory structure.

### Changes:
- Relocated source files using `git mv` to preserve commit history.
- Updated `unittests/CMakeLists.txt` to reflect the new directory structure.

### Issue:
Fixes #4684


cabreraam (changes requested):
Hi @zoldyck13. Thanks for working on this!

I think I need a better understanding of what needs to get done and why.

- Based on issue https://github.com/NVIDIA/cuda-quantum/issues/4684, the goal is to refactor the unit tests from `unittests/Optimizer/` into a new directory, `cudaq/unittests`. Okay, understood.
- But then `unittests/CMakeLists.txt` is supposed to be responsible for building the relocated tests?
  - this doesn't really make sense to me. Why would we use this `CMakeLists.txt` once the tests are relocated? I think the CMake infrastructure needs to be reworked here.
  - even as written, I'm not convinced that CMake would be able to find those relocated tests. I think it would look in a directory called `unittests/cudaq/unittests` instead of `cudaq/unittests`.
  
- if the goal is to link `cudaq/test/Unit` to `cudaq/unittests/` in the same way that [`llvm/test/unit`](https://github.com/llvm/llvm-project/tree/main/llvm/test/Unit) is linked to [`llvm/unittests`](https://github.com/llvm/llvm-project/tree/main/llvm/unittests), then I imagine that there should be some kind of `CMakeLists.txt` definition that creates a target like `check-cudaq-unit` that depends on all of the tests in `cudaq/unittests`. Based on the way `llvm` does it, they take advantage of some LLVM CMake functions, with  `add_lit_testsuite` and `add_llvm_unittest` seeming like the most important ones.
  - Note that the tests in `cudaq/unittests` are not `lit` tests though, so we wouldn't be able to use those already implemented functions  

IIUC, it seems like the goal is to create a CMake target that can be driven like `cmake --build build --target check-cudaq-unit`. 
- This would drive the tests in `cudaq/unittests`. This would _not_ run the tests in `unittests`.
- This would require some CMake plumbing that looks similar to the CMake files in `llvm-project/llvm/test/Unit` and `llvm-project/llvm/unittests` 

I think that moving the tests to `cudaq/unittests` seems like a great first step.

@schweitzpgi If I'm misrepresenting your thoughts, please feel free to chime in!

PR #4697, opened 2026-06-08
Add angle-producing math functions to kernels with docs and tests

### LINKED ISSUE #2942: Mathematical functions support

LakshikkaNathan:
Closes #2942

## Summary
Completes the continuous, angle-producing math functions from the acceptance plan in #2942 (`sin, cos, tan, asin, acos, atan, sqrt, exp, log`) for C++ and Python: it adds the functions that were not yet supported, and documents and tests the full set.

- **C++** (`lib/Frontend/nvqpp/ConvertExpr.cpp`): lowering for `std::asin`, `std::acos`, `std::atan`. The other angle functions were already supported.
- **Python** (`python/cudaq/kernel/ast_bridge.py`): adds `np.tan`, `np.asin`, `np.acos`, `np.atan`, `np.log` to the supported NumPy functions and lowers them to the matching `math` dialect ops (real-valued; complex inputs raise).

## Against the acceptance plan
1. C++ supported functions documented in the kernel specification (`docs/sphinx/specification/cudaq/kernels.rst`).
2. Python supported functions documented in the same place.
3. Per-function simulator value tests for all nine functions (`test_kernel_float.py`), plus a FileCheck lowering test (`test/AST-Quake/math_functions.cpp`) covering double and float.
4. An executed gate-parameter test: `ry(2*asin(sqrt(p)))` prepares a state with P(|1>) = p, verified by sampling.
5. Continuous angle functions kept separate from discrete control functions.

## Testing
Built from source in the `cuda-quantum-dev` container. The new tests pass, the existing `test_kernel_float.py` suite still passes, and the FileCheck test passes via `cudaq-quake | cudaq-opt | FileCheck`.

## Note
I'm aware of #4680, which addresses the same issue. This PR takes a similar core approach and additionally includes the documentation and the executed gate-parameter test from the acceptance plan.

## AI usage
I used Claude as a coding assistant to help draft this change. I built CUDA-Q from source, ran and verified the tests, and reviewed the full patch before submitting.


sacpis:
> /ok to test 43a36fb

LakshikkaNathan:
Thanks @sacpis. Pushed a yapf formatting fix in the latest commit (3ffd148). Could you re-run on the latest commit when you get a chance? Thanks

sacpis:
> /ok to test 845b40b

cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp
```diff
@@ -1399,6 +1399,21 @@ bool QuakeBridgeVisitor::visitMathLibFunc(clang::CallExpr *x,
...
+
+  // Handle std::acos
+  if ((isInNamespace(func, "std") || isNotInANamespace(func)) &&
+      (funcName == "acos" || funcName == "acosf"))
+    return floatOperator(math::AcosOp{}, "acos");
+
+  // Handle std::atan
+  if ((isInNamespace(func, "std") || isNotInANamespace(func)) &&
+      (funcName == "atan" || funcName == "atanf"))
+    return floatOperator(math::AtanOp{}, "atan");
```
sacpis, line 1476: Should we also include ` arcsin`, `arccos`, `arctan`?

python/tests/kernel/test_kernel_float.py
```diff
@@ -310,3 +310,75 @@ def check(c: any):
...
+
+    @cudaq.kernel
+    def f_tan() -> np.float64:
+        return np.tan(np.float64(0.5))
+
+    assert is_close(np.tan(0.5), f_tan())
+
+    @cudaq.kernel
+    def f_asin() -> np.float64:
+        return np.asin(np.float64(0.5))
```
sacpis: Should we use `arcsin` here as we have compared the result with `arcsin` on line 340?

python/tests/kernel/test_kernel_float.py
```diff
@@ -310,3 +310,75 @@ def check(c: any):
...
+
+    assert is_close(np.arcsin(0.5), f_asin())
+
+    @cudaq.kernel
+    def f_acos() -> np.float64:
+        return np.acos(np.float64(0.5))
```
sacpis: same here with `arccos`.

python/tests/kernel/test_kernel_float.py
```diff
@@ -310,3 +310,75 @@ def check(c: any):
...
+
+    assert is_close(np.arccos(0.5), f_acos())
+
+    @cudaq.kernel
+    def f_atan() -> np.float64:
+        return np.atan(np.float64(0.5))
```
sacpis: same here with `arctan`.

sacpis (commented):
Overall LGTM. Thanks @LakshikkaNathan. Left a few comments. Once we have those changes, will take a look at it again before approving.

cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp
LakshikkaNathan: Thanks for the suggestion! C++ <cmath> only provides asin/acos/atan (and the f-suffixed float versions), so there's no arcsin/arccos/arctan to map on the C++ side. Since those are NumPy's spellings, I've added support for them in the Python frontend (ast_bridge.py) instead, where those names are valid.

python/tests/kernel/test_kernel_float.py
LakshikkaNathan: Thanks! Switched the kernel to np.arcsin to match the host comparison.

python/tests/kernel/test_kernel_float.py
LakshikkaNathan: Done, switched to np.arccos to match the comparison.

python/tests/kernel/test_kernel_float.py
LakshikkaNathan: Done, switched to np.arctan to match the comparison.

LakshikkaNathan:
Thanks for the review, @sacpis! Addressed all the comments in `41c15fd5`, added `arcsin`/`arccos`/`arctan` support in `ast_bridge.py` and switched the test kernels to match. Ready for another look whenever you have time.

sacpis:
> /ok to test a53ca21

python/tests/kernel/test_kernel_float.py
```diff
@@ -310,3 +310,75 @@ def check(c: any):
     check([np.float32(np.pi / 2), 0])
     check([1, 0])
     check([np.pi / 2, 0, True])
+
+
+def test_math_functions():
```
sacpis: Let's add tests for `np.float32` as well.

LakshikkaNathan: Done, added `test_math_functions_float32` and renamed the existing one to `test_math_functions_float64`.

LakshikkaNathan:
Added the `np.float32` tests in `84e441c`. Ready for another look, thanks @sacpis!

sacpis:
> /ok to test 84e441c

LakshikkaNathan:
Thanks for running the tests. The failing `test_quantinuum_observe` doesn't appear to be related to this PR's changes, since its ansatz uses only `x`, `ry`, and `x.ctrl`, none of the math functions added here. The tests added in this PR pass. Could you let me know if this is a known failure on `main`? Happy to rebase once it's resolved.

sacpis:
> /ok to test 4ecc8bd

sacpis:
> /ok to test ca74a0f

PR #4699, opened 2026-06-08
[nvq++] Clean up QPU launches and remove MLIR header deps

lmondada:
My last PR on compiler unification (for now)! This came out nicely in the end, I'm very satisifed! Thanks @Renaud-K for suggesting this direction.

## Previous status

When launching from Python, kernels were first JIT compiled and then passed to `launchKernel` or `unifiedLaunchModule` (depending on policy) as `CompiledModule`. This is the desired execution flow, as the launch endpoint does not need to perform any compilation (other than codegen, if required).

When launching from C++, kernels were not JIT compiled before launch. They were passed as-is to `launchKernel`/`unifiedLaunchModule` as `SourceModule`s. If the backend required JIT compilation, the launch endpoint would then launch compilation before proceeding to execution. This is the "old" execution flow that we are moving away from, as it couples QPU launch logic and compilation, making caching impossible, etc.

## This PR

Moves compilation out of QPU::launchKernel, also for C++. The JIT compilation now happens in `runtime/cudaq/algorithms/launch.h`. This means that the binaries produced by `nvq++` **must link in the `cudaq-mlir-runtime`** for the compiler symbols to resolve. By default, `nvq++` was already linking in this library, so the default behaviour of `nvq++` remains unchanged.

There is one small breaking change. It concerns the `--disable-mlir-links` option flag of `nvq++`, which is used to exclude `cudaq-mlir-runtime` from the list of linked libraries. Setting this flag when compiling for targets that require JIT compilation (e.g. remote devices and emulation mode) will now cause kernel executions to fail at launch time. More specifically, `RemoteRESTQPU::launchKernel` will throw an `std::runtime_error` with message:
```
QPU does not support launching a CompiledModule without MLIR artifacts.
```

This worked previously because QPUs that required JIT compilation took care themselves of linking in `cudaq-mlir-runtime` (and ignored `--disable-mlir-links`). The new behaviour is achieved by introducing a `CUDAQ_DISABLE_JIT_COMPILER` preprocessor flag, which is turned on when `--disable-mlir-links` is set and produces dummy implementations for the missing compiler symbols.

## Clean ups

These changes complete the compiler unification and allow the following further clean ups:
- QPUs no longer need to link in `cudaq-mlir-runtime-headers`.
- `cudaq-mlir-runtime-headers` library is removed.
- `QPU::launchKernel` endpoints now take a `CompiledModule` instead of `AnyModule`.

Finally, the only "leftover"s that are yet to be cleaned up are
-  `unifiedLaunchModule`, which still accepts an `AnyModule` and still performs JIT compilation when required: this is used for policies that have not been migrated to using policy objects. These launches are not yet using the `executeKernelApi` hook of `launch.h` and thus cannot be compiled before launch without introducing a dependency on the compiler in `quantum_platform.cpp`. This will go away once all launch policies have been migrated (WIP).
- relying on `nvq++` to link in the compiler library is not ideal (and forces us to put the compilation step in a public-facing header file). We'd be better served by a dedicated C++ launch library that would take care of resolving symbols and hide away non-public details.
- The `executeKernelApi` is the only place where C++ compilation can happen (it must be in a header file, must have access to the policy and must have access to the SourceModule). This means that both JIT compilation and launch must happen in that hook, which is different from how Python does it, where `executeKernelApi` is purely for launch. This is another consequence of not having a C++ launch library that can compile before launching.


runtime/cudaq/algorithms/launch.h
```diff
@@ -51,7 +74,23 @@ auto launch(Policy &policy, std::size_t qpu_id, ExecutionContext &ctx,
   auto &qpu = platform.getQPU(qpu_id);
   ctx.executeKernelApi = [&qpu, &result, &policy](const AnyModule &module,
                                                   const KernelArgs &args) {
-    result = qpu.launchKernel(policy, module, args);
+    CompiledModule compiled;
+    if (const auto *source = std::get_if<SourceModule>(&module)) {
```
taalexander, line 66: The intent of this logic is to handle AOT/JIT complilation?

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
```diff
@@ -7,76 +7,98 @@
...
-  return completeLaunchKernel(policy, kernelName, std::move(codes));
+
+  if (module.getMlirArtifacts().empty())
+    throw std::runtime_error("QPU does not support launching a "
+                             "CompiledModule without MLIR artifacts.");
+
+  cudaq_internal::compiler::Compiler compiler(getCompileTarget(policy));
+  auto codes = compiler.emitKernelExecutions(module);
+
+  return completeLaunchKernel(policy, module.getName(), std::move(codes));
```
taalexander: Nit - It seems like the sequence below is repeated ~3 times. Could it use a helper method?

runtime/cudaq/CMakeLists.txt
```diff
@@ -66,7 +66,6 @@ if (CUDA_FOUND)
       cudaq-common
       cudaq-nlopt
       cudaq-ensmallen
-      cudaq-mlir-runtime-headers # Public until compilation is moved out of the QPUs
```
taalexander, line 70: 🎉

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
```diff
@@ -143,17 +137,32 @@ compileModule(const Policy &policy,
...
+  }
+
   const auto &kernelName = src.getName();
-  auto modulePtr = src.getMlirOpaqueModulePtr();
-  assert(modulePtr && "Compiler::compileModule requires an MLIR artifact");
+  auto mlirArt = src.getMlir();
+  if (!mlirArt.has_value()) {
+    auto srcOwned = src;
+    CompiledModuleHelper::loadMlirArtifacts(srcOwned);
+    mlirArt = srcOwned.getMlir();
```
sacpis: Why are we doing deep copy just to call `loadMlirArtifacts`? Should we just make `loadMlirArtifacts` return the `MlirArtifact`?

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
```diff
@@ -143,17 +137,32 @@ compileModule(const Policy &policy,
               std::unique_ptr<cudaq::CompileTarget> target,
               const cudaq::SourceModule &src, cudaq::KernelArgs args,
               bool isEntryPoint = true) {
+  if (!target->overrideAOTCompilation && src.getFunctionPtr()) {
+    // We are allowed to use the AOT-compiled module as-is, so nothing to do.
+    CUDAQ_INFO("No JIT compilation required. Using AOT-compiled module as-is.");
+    return cudaq::CompiledModule{src};
```
sacpis: Should we do `cudaq::CompiledModule{std::move(src)}`?

runtime/cudaq/algorithms/launch.h
```diff
@@ -17,13 +18,35 @@
...
+
+// If JIT compilation is disabled, make compilation a no-op. QPUs may throw an
+// error if they expect a JIT-compiled module.
+#ifdef CUDAQ_DISABLE_JIT_COMPILER
+template <typename Policy>
+cudaq::CompiledModule cudaq_internal::compiler::compileModule(
+    const Policy &policy, std::unique_ptr<cudaq::CompileTarget> target,
+    const cudaq::SourceModule &src, cudaq::KernelArgs args, bool isEntryPoint) {
+  CUDAQ_INFO("JIT compilation is disabled. Compilation is a no-op.");
+  return cudaq::CompiledModule{src};
```
sacpis: Should we do `cudaq::CompiledModule{std::move(src)}`?

sacpis (approved):
Overall LGTM. Thanks @lmondada. Nicely written comments and PR description.

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
```diff
@@ -7,76 +7,98 @@
...
-  auto [kernelName, codes] =
-      compileKernelExecutions(policy.inner, module, args);
-  return completeLaunchKernel(policy, kernelName, std::move(codes));
+
+  if (module.getMlirArtifacts().empty())
+    throw std::runtime_error("QPU does not support launching a "
+                             "CompiledModule without MLIR artifacts.");
+
+  cudaq_internal::compiler::Compiler compiler(getCompileTarget(policy.inner));
+  auto codes = compiler.emitKernelExecutions(module);
```
Renaud-K: So codegen is still happening in the QPUs?

runtime/cudaq/CMakeLists.txt
```diff
@@ -94,7 +93,6 @@ else()
...
```
Renaud-K, line 98: The goal was to move the dependency from public back to private. 
But I don't think that it can be removed in the general case. We might be lucky that for now mlir dependencies go in templatized code that defer compilation and link. But this may not always be the case.

unittests/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ SET(CMAKE_SHARED_LINKER_FLAGS "")
 # survives as an internal implementation detail.
 add_compile_definitions(CUDAQ_LIBRARY_MODE)
 
+# Disable JIT compilation as it is not needed (and we would otherwise have to
+# link in the compiler library for no reason)
+add_compile_definitions(CUDAQ_DISABLE_JIT_COMPILER)
```
Renaud-K: I am rather uncomfortable with these tests needing special settings. They should be working as-is.

runtime/internal/compiler/Compiler.cpp
```diff
@@ -706,3 +708,24 @@ mlir::ModuleOp cudaq_internal::compiler::Compiler::lowerQuakeCodeBuildModule(
...
+// from `cudaq::detail::launch` (see runtime/cudaq/algorithms/launch.h).
+template cudaq::CompiledModule
+cudaq_internal::compiler::compileModule<cudaq::sample_policy>(
+    const cudaq::sample_policy &, std::unique_ptr<cudaq::CompileTarget>,
+    const cudaq::SourceModule &, cudaq::KernelArgs, bool);
+template cudaq::CompiledModule
+cudaq_internal::compiler::compileModule<cudaq::observe_policy>(
+    const cudaq::observe_policy &, std::unique_ptr<cudaq::CompileTarget>,
+    const cudaq::SourceModule &, cudaq::KernelArgs, bool);
+template cudaq::CompiledModule cudaq_internal::compiler::compileModule<
```
Renaud-K: The explicit instantiation itself is acceptable since the code is in the cuda-mlir-runtime-library. 

However, do we really need to instantiate the `cudaq::async_policy_wrapper`?

runtime/cudaq/algorithms/launch.h
lmondada: Rather, it's to handle Python vs C++. If I could have had it my way, JIT compilation would always happen before getting to this point. This is possible in Python, thus when coming from Python, the module is already a `CompiledModule`. For C++, there is no earlier point where the JIT compilation can be inserted, so it has to happen here. This is as early as it can be.

We'll be able to clean this up once/if we introduce a dedicated C++ library that links in the compiler and can be used to launch kernels. I will add this point in the PR description, thanks for pointing it out :)

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
lmondada: Yes. My reasoning for this (early on) was that because the output of codegen is hardware-dependent, there will be many custom formats. I didn't want to add all these formats to `CompiledModule`, so anything "beyond" MLIR would have to happen in the launch code.

Happy to revisit this decision: we could add support for the most common codegen emission types and move that into compilation.

runtime/cudaq/CMakeLists.txt
lmondada: > But I don't think that it can be removed in the general case. We might be lucky that for now mlir dependencies go in templatized code that defer compilation and link. But this may not always be the case.

Can you point me to where you see compiler headers being included? When I grep for `cudaq_internal/compiler`, I only find results either in QPUs, in tests, in `runtime/internal/compiler` or in `runtime/cudaq/builder/kernel_builder.cpp`. None of these make it into the `cudaq` library.

Besides, if I do try to include a header from `cudaq_internal/compiler` within a `.cpp` file of the `cudaq` library (e.g. `runtime/cudaq/cudaq.cpp`), I get an `#include` error, confirming that the compiler headers are not in the include paths of `cudaq`.

unittests/CMakeLists.txt
lmondada: This could be removed if we added `cudaq-mlir-runtime` as a dependency to all the tests (which `nvq++` would do). I preferred this approach as it would have slowed down test compilation for nothing. Happy to add the dependency everywhere if you prefer.

runtime/cudaq/CMakeLists.txt
Renaud-K: I meant that for future work, I am not sure that we will always be able to rely on C++ templates and forward declarations to defer the link. I was not going make it a hard requirement that  `cudaq-mlir-runtime-headers` ought to be removed altogether. The fact that we need forward declarations shows that the independence is only an illusion. All we really need is the ability to defer the link.

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
Renaud-K: Based on these PR titles, I was lead to believe that yes, this would be moved out. I got a bit suspicious when `getCompileTarget` became a virtual method on the QPU. :) I was not sure where this was all heading. 

But this is part of bigger discussions around where we want to go with the QPU classes. What will their role be as more and more code is moved out. There has already been great progress made. The fact that `AnyModule` is now always `CompiledModule` is a very good step for this PR.

unittests/CMakeLists.txt
Renaud-K: It is a shared library. It only needs to be loaded in memory once. So the cost is very low compared to parsing `cudaq.h` which has to be done for each test.  

We are also working on removing 
```
add_compile_definitions(CUDAQ_LIBRARY_MODE)
```
So, this feels like the need to set another flag to be able to compile host code with the C++ compiler.

Renaud-K: I had no failures with removing this setting.

runtime/cudaq/platform/default/rest/RemoteRESTQPU.cpp
lmondada: Good suggestion, done.

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
lmondada: You're so right, this makes it much simpler. Thanks for pointing that out.

runtime/internal/compiler/include/cudaq_internal/compiler/Compiler.h
lmondada: I don't think this performance gain justifies this. The source module is passed in as const reference and I'd rather keep it this way and not invalidate it.

runtime/cudaq/algorithms/launch.h
lmondada: same comment as above

runtime/internal/compiler/Compiler.cpp
lmondada: with #4743 we don't even need the policy at all. I will remove the template and the specializations.

lmondada: btw the policy-free version is more "correct": the compilation should only depend on the compilation target, not the launch policy.

unittests/CMakeLists.txt
lmondada: If I remove it, I get the expected error:

```
/usr/bin/ld: unittests/nvqpp/CMakeFiles/test_runtime_qpp.dir/integration/adjoint_tester.cpp.o: undefined reference to symbol '_ZN14cudaq_internal8compiler13compileModuleESt10unique_ptrIN5cudaq13CompileTargetESt14default_deleteIS3_EERKNS2_12SourceModuleENS2_10KernelArgsEb'
/usr/bin/ld: /workspaces/cuda-quantum/build/lib/libcudaq-mlir-runtime.so: error adding symbols: DSO missing from command line
```

Once you have migrated to using `nvq++`, you can remove this flag in any case.

lmondada: I've added commit [992ccc9](https://github.com/NVIDIA/cuda-quantum/pull/4699/commits/992ccc958e4ecb2e808485facf0cb24eb88af534) which removes the `CUDAQ_DISABLE_JIT_COMPILER` flag from the tests and adds `cudaq-mlir-runtime` dependencies everywhere. See if this is what you had in mind or if you prefer the original.

Either way, this would go away once `nvq++` is used.

runtime/cudaq/CMakeLists.txt
lmondada: as agreed, I've restored the ` cudaq-mlir-runtime-headers` (possibly for future use), but I am keeping it out of the `cudaq` dependencies.

lmondada:
Ready to merge, just waiting for https://github.com/NVIDIA/cuda-quantum/pull/4743 to go first.

PR #4703, opened 2026-06-09
[python] Fix dynamics target evolve_async regression

huaweil-nv:
## Summary

- Add the missing Python binding for the callable `evolve_async` overload used by the `dynamics` target.
- Propagate callable failures through the async evolution future so `AsyncEvolveResult.get()` raises instead of terminating the worker.
- Add regression coverage for successful dynamics async evolution, intermediate result modes, observable expectation values, and async error propagation.

## How To Reproduce

This is a regression. The sample below works with CUDA-Q `0.13.0`, but fails with CUDA-Q `0.14.0`.

```python
import cupy as cp
import cudaq
from cudaq import operators, Schedule

cudaq.set_target("dynamics")

steps = [0.0, 0.1]
schedule = Schedule(steps, ["t"])
hamiltonian = operators.number(0)
dimensions = {0: 2}

state = cp.zeros(2, dtype=cp.complex128)
state[1] = 1.0
initial_state = cudaq.State.from_data(state)

result = cudaq.evolve_async(
    hamiltonian,
    dimensions,
    schedule,
    initial_state,
).get()

print(result.final_state())
```

The dynamics target dispatches cudaq.evolve_async() by passing a Python callable to cudaq_runtime.evolve_async():
```
cudaq_runtime.evolve_async(lambda: evolve_dynamics(...))
```
The runtime bindings already covered the kernel-based async overloads, but did not expose the callable overload to Python. This caused dynamics target async evolution to fail with an incompatible function arguments error.

This PR adds a callable overload binding and updates the callable async implementation to catch exceptions from the queued callable and store them on the promise. With this behavior, failures in evolve_dynamics() are reported when users call .get(), matching normal future semantics.

### DIFF
--- a/runtime/cudaq/algorithms/evolve_internal.h
+++ b/runtime/cudaq/algorithms/evolve_internal.h
@@ -109,13 +109,19 @@ evolve_async(state initial_state, QuantumKernel &&kernel,
       [p = std::move(promise), func = std::forward<QuantumKernel>(kernel),
        initial_state, observables, noise_model, shots_count,
        &platform]() mutable {
-        if (noise_model.has_value())
-          platform.set_noise(&noise_model.value());
-        with_platform_in_library_mode libraryMode(platform);
-        auto result = evolve(initial_state, func, observables, shots_count);
-        if (noise_model.has_value())
-          platform.set_noise(nullptr);
-        p.set_value(std::move(result));
+        try {
+          if (noise_model.has_value())
+            platform.set_noise(&noise_model.value());
+          with_platform_in_library_mode libraryMode(platform);
+          auto result = evolve(initial_state, func, observables, shots_count);
+          if (noise_model.has_value())
+            platform.set_noise(nullptr);
+          p.set_value(std::move(result));
+        } catch (...) {
+          if (noise_model.has_value())
+            platform.set_noise(nullptr);
+          p.set_exception(std::current_exception());
+        }
       });
 
   platform.enqueueAsyncTask(qpu_id, wrapped);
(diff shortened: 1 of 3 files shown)


1tnguyen (approved):
LGTM 👍

1tnguyen:
> /ok to test 3a29086

runtime/cudaq/algorithms/evolve_internal.h
```diff
@@ -164,7 +165,11 @@ evolve_async(std::function<evolve_result()> evolveFunctor,
 
   QuantumTask wrapped = detail::make_copyable_function(
       [p = std::move(promise), evolveFunctor]() mutable {
-        p.set_value(evolveFunctor());
+        try {
+          p.set_value(evolveFunctor());
+        } catch (...) {
+          p.set_exception(std::current_exception());
+        }
```
sacpis, line 184: nit: Should we apply this try/catch block in the other 2 overloads?

sacpis (approved):
LGTM. Thanks @huaweil-nv.

schweitzpgi (commented):
Nice. Thanks!

runtime/cudaq/algorithms/evolve_internal.h
huaweil-nv: updated. The other two `evolve_async` overloads now also catch worker exceptions and propagate them through the returned future.

huaweil-nv:
> /ok to test 3ccb768661aeb3244f47b75c674437724dc6e769

huaweil-nv:
> /ok to test 805ae31235d6f9644ad86b63964336f4708f945f

PR #4663, opened 2026-06-03
[nfc] Unify internal-use namespaces to `detail`

khalatepradnya:
### Summary

Collapses the five namespaces CUDA-Q used for the "non-public implementation detail" namespace into the single `detail` namespace required by [`CppAPICodingStyle.md`](https://github.com/NVIDIA/cuda-quantum/blob/9f1bb0105642195ee54c075923328aa052afa733/CppAPICodingStyle.md).

### Motivation

`detail`, `details`, `internal`, `__internal__`, and `__internal` were all used for the same role — at one point three of them coexist in a single file (`runtime/cudaq.h`). The `__internal__`/`__internal` forms are also non-conforming: an identifier containing a double underscore is reserved to the implementation, so this doubles as a conformance fix.

### What changed

- `details` / `internal` / `__internal__` / `__internal` → `detail`, within their existing parents (`cudaq::detail`, `cudaq_fmt::detail`,   `cudaq::opt::detail`).
- Intentionally untouched: third-party `spdlog::details`, `fmt::detail`, and `testing::internal`; the `details` helper local to `NVQIR.cpp`; and the `cudaq_internal::<module>` API tier (a separate, already-consistent concept).

PR #4708, opened 2026-06-09
[bugfix] Namespace name change dropped information

schweitzpgi:
on the floor.


schweitzpgi:
#4663

khalatepradnya (approved):
Thanks, @schweitzpgi for the fix!

PR #4710, opened 2026-06-10
[python] Canonicalize InitialState in simulator evolve_async

huaweil-nv:
## Summary

- Reuse the simulator initial-state canonicalization logic for `evolve_async`.
- Fix `evolve_async` on simulator targets when `initial_state` is an `InitialState` enum.
- Add regression coverage for `InitialState.ZERO` and `InitialState.UNIFORM` on `density-matrix-cpu` and `qpp-cpu`.

## How To Reproduce

`evolve_async` advertises `InitialStateArgT` for `initial_state`, but the simulator async path did not canonicalize `InitialState` enum values before calling the runtime binding.

```python
import cudaq
from cudaq import operators, Schedule
from cudaq.dynamics import InitialState

cudaq.set_target("qpp-cpu")

hamiltonian = operators.number(0)
dimensions = {0: 2}
schedule = Schedule([0.0, 0.1], ["t"])

result = cudaq.evolve_async(
    hamiltonian,
    dimensions,
    schedule,
    InitialState.ZERO,
).get()

print(result.final_state())
```
Before this fix, the call fails with:
```
TypeError: evolve_async(): incompatible function arguments
Invoked with types: InitialStateType, PyKernel
```
The synchronous cudaq.evolve path already converted InitialState.ZERO / InitialState.UNIFORM into a concrete State, so this was a simulator async API consistency bug.

### DIFF
--- a/python/cudaq/dynamics/evolution.py
+++ b/python/cudaq/dynamics/evolution.py
@@ -71,6 +71,28 @@ def _canonicalize_operator_to_dimensions(
     return system_operator
 
 
+def _canonicalize_initial_state(initial_state: InitialStateArgT,
+                                num_qubits: int) -> InitialStateArgT:
+    if not isinstance(initial_state, InitialState):
+        return initial_state
+
+    state_size = 2**num_qubits
+    if initial_state == InitialState.ZERO:
+        state_data = numpy.zeros(state_size, dtype=numpy.complex128)
+        state_data[0] = 1.0
+    elif initial_state == InitialState.UNIFORM:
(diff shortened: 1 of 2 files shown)


1tnguyen (approved):
LGTM 👍

1tnguyen:
> /ok to test [0b13fa2](https://github.com/NVIDIA/cuda-quantum/pull/4710/commits/0b13fa2b96fb4ede8be2d2823b801ebbccf2bfff)

PR #4714, opened 2026-06-10
[testing] Create the CMake infrastructure for `cudaq/unittests`

cabreraam:
## Motivation

## Goal

## Implementation
This PR will be the first of two.
This PR defines the directory and the CMake infrastructure for `cudaq/unittests`.
In particular, we take inspiration from `add_llvm_unittest` when implementing `add_cudaq_unittest`.

## Notes


sacpis:
@cabreraam Wondering if we can get all changes in 1 PR? That will be easy to build and test locally.

cabreraam:
> @cabreraam Wondering if we can get all changes in 1 PR? That will be easy to build and test locally.

A valid question! Here are my thoughts:

I just posted the next PR (#4722). I like keeping them modular in this way because it disambiguates what is necessary to make the change and then the next PR implements the change. To me, it feels like one PR is the hammer and one PR is a nail for the hammer.

PR #4722 runs the resulting CI as I've based it on top of this current PR.

If after seeing both and you think it best to combine them, I will close the second PR and just add the changes from that PR here.

cudaq/unittests/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+# as arguments, as well as optional arguments for:
+# - sources
+# - includes
+# - linked libraries
+# - compile options
+# - compile definitions 
+# for the test. 
+# The test will be added to the `CudaQUnitTests` target, and a custom target will 
+# be created for it that we can run with `make check-<test_label>`.
+function(add_cudaq_unittest test_name test_label)
```
sacpis: Wondering if `test_label` is being used? I think here `optimizer-unit-tests`, `analysis-passes`, `quake-synth` are passed which are silently ignored. I think we should either use the label or drop the parameter.

sacpis (approved):
LGTM. Thanks @cabreraam.

cudaq/unittests/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+add_custom_target(CudaQUnitTests)
```
schweitzpgi, line 9: I was kind of expecting this to look like `tpls/llvm/llvm/unittests/CMakeLists.txt`. Not sure that it does though?

cabreraam: Yes, that is the goal. I believe it does that.

- `add_custom_target(CudaQUnitTests)` is analagous to [`add_custom_target(UnitTests)` in `llvm`](https://github.com/llvm/llvm-project/blob/main/llvm/unittests/CMakeLists.txt#L1) representing an aggregate build target for all of the tests in `cudaq/unittest`
- I chose to implement `add_cudaq_unittest` which models `add_llvm_unittest`. The main difference here is that I'm not wrapping `add_unittest` and instead just building the test executable as part of `add_cudaq_unittest`. Otherwise, functionally, they achieve the same thing: build the test executable and add the test executable's build target to an aggregate target (`CudaQUnitTest` for us, `UnitTests` in `llvm/unittests`).
- What's missing from this PR is the adding of all of the relevant `cudaq/unittest` subdirectories to be built. The follow-on PR (#4722) addresses this and adds each [subdirectory within `cudaq/unittest`](https://github.com/cabreraam/cuda-quantum/blob/5f973e483565144cf8b7a2a05cd208a405f7507c/cudaq/unittests/CMakeLists.txt#L104) to be [built using `add_cudaq_unittest`](https://github.com/cabreraam/cuda-quantum/blob/5f973e483565144cf8b7a2a05cd208a405f7507c/cudaq/unittests/Optimizer/CMakeLists.txt#L9-L66).

cudaq/unittests/CMakeLists.txt
cabreraam: Good catch! I reworked some things and forgot to remove `test_label`. It's no longer being used!

schweitzpgi (approved):

PR #4715, opened 2026-06-10
Handle adjoint declarations gracefully

### LINKED ISSUE #4268: `apply-op-specialization` crashes instead of erroring when adjointing a forward-declared kernel

fallintoplace:
Fixes #4268.

This adds an explicit diagnostic when apply-op-specialization is asked to create an adjoint variant for a declaration-only kernel. That avoids walking into an empty function body and crashing before the user gets a useful compiler error.

Validation:
- git diff --check
- Not run locally: cudaq-opt/llvm-lit are not available in this checkout and LLVM_INSTALL_PREFIX is unset.


atgeller:
> /ok to test a0c0180

cudaq/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -545,6 +545,13 @@ class ApplySpecializationPass
     auto &funcBody = func.getBody();
 
     // Check our restrictions.
+    if (func.isDeclaration()) {
+      func.emitError() << "cannot make adjoint of kernel '" << func.getName()
+                       << "' because its body is unavailable";
+      signalPassFailure();
```
schweitzpgi: I don't think we want the pass itself to fail here. The kernel to create the adjoint form from may not be available early, but may become available at JIT time. This would crash the compiler either way.

schweitzpgi (changes requested):

schweitzpgi (approved):

schweitzpgi:
> /ok to test 7b33779

schweitzpgi:
Closing as duplicate of #4752

schweitzpgi (changes requested):
Looks like a duplicate. There are about 5 of these.

PR #4716, opened 2026-06-10
Fix heap corruption from `list[bool]` kernel arguments

mitchdz:
On local simulators a list[bool] is marshaled as a bit-packed
  std::vector<bool>, but argument synthesis read every vector arg as a
  {begin, end, capacity} pointer triple. That layout is wrong for
  std::vector<bool> (its members are _Bit_iterators, not raw pointers), so it
  walked off the buffer and corrupted the heap, or silently read an empty
  vector when the padding was zero. This showed up as the flaky
  test_list_subscript segfault.

  Read i1 vector args through the std::vector<bool> API when the target is a
  local simulator (gated by CompileTarget::isLocalSimulator). Remote and
  emulated targets pack std::vector<char> and are unchanged. Covers sample,
  observe, run, get_state, translate/get_qir, and lifted closure args.

  Regression from https://github.com/NVIDIA/cuda-quantum/pull/4265 (argsCreator packing).


schweitzpgi:
I think the bug is in the Python code. Arguments are Python objects in Python and we are free to marshal them (correctly) any way we want as C++ proxies.

The ArgumentConversion algorithm is correctly assuming that all `vector` types are the triple pointer encoding. This is a universal assumption on the device.

Unfortunately, Python has decided to incorrectly encode a `list[bool]` as a `vector<bool>` which is just plain wrong. It must be encoded as a `vector<char>`.

Renaud-K:
I would tend to agree with Eric. We are having to touch a lot of API just for vector<bool> support. Any chance this can be handled up the stack?

schweitzpgi (approved):
Thanks for the cleanup!

PR #4720, opened 2026-06-11
[CI] Run Python test suite on nvidia GPU simulator in merge queue

mitchdz:
Per-PR CI runs python/tests only on qpp-cpu, so regressions that fail only on the nvidia (GPU) simulator slip through to publishing.

Add a gpu_validation job to python_wheels.yml that reuses the wheel from build_wheel and runs python/tests (minus backends/contrib) under CUDAQ_DEFAULT_SIMULATOR=nvidia on an A100 runner. Gated to a single config (amd64/py3.11/cu12.6) and, via python_wheels, to merge_group, schedule, and workflow_dispatch only (not per-PR push).

### DIFF
--- a/runtime/test/test_argument_conversion.cpp
+++ b/runtime/test/test_argument_conversion.cpp
@@ -152,7 +152,8 @@ void dumpSubstitutionModules(ArgumentConverter &con) {
 
 void doSimpleTest(mlir::MLIRContext *ctx, const std::string &typeName,
                   std::vector<void *> args,
-                  const std::string &additionalCode = "") {
+                  const std::string &additionalCode = "",
+                  bool boolVecBitPacked = false) {
   std::string code = additionalCode + R"#(
 func.func private @callee(%0: )#" +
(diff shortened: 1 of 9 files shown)


1tnguyen (approved):
👍

runtime/test/test_argument_conversion.cpp
```diff
@@ -400,6 +401,22 @@ void test_vectors(mlir::MLIRContext *ctx) {
  // CHECK:         }
   // clang-format on
 
+  {
+    // Real bit-packed `std::vector<bool>`, as the local-simulator launch path
+    // passes it. Reading this as a `{begin, end, capacity}` triple corrupts the
+    // heap; `boolVecBitPacked` selects the correct reader.
+    std::vector<bool> x = {true, false, true, true};
```
schweitzpgi, line 408: Why was this added? Was this failing somewhere other than this test?

I see, in the comment above. Where is this local-simulator launch path? The bug is there.

PR #4721, opened 2026-06-11
[fix] Skip compilation in the absence of BackendConfig

lmondada:
I suggest this fix over #4718, it correctly fixes my omission from #4636. It preserves the old behavior that if no optimization pipeline is defined, none is run.

I'm writing some tests that I will push in a minute. EDIT: done.

Having had to go again through the details of this logic, this could definitely use a good cleanup. I'll create a jira task for this.

### DIFF
--- a/cudaq/lib/Target/CompileTarget.cpp
+++ b/cudaq/lib/Target/CompileTarget.cpp
@@ -48,8 +48,10 @@ cudaq::CompileTarget::CompileTarget(
     config::TargetConfig targetConfig,
     std::map<std::string, std::string> runtimeConfig, bool emulate_)
     : emulate(emulate_) {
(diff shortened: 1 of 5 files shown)


Renaud-K:
Was it tested on a GPU?

lmondada:
> Was it tested on a GPU?

Yes, on an AWS Ubuntu instance with GPU. I reproduced the failures and confirmed that this fixes it.

On the GPU instance I wasn't able yet to run the `nvidia-mqpu` tests -- I will install the target and confirm that it works -- should be unrelated though.

EDIT: I confirm mqpu works

sacpis (commented):
LGTM. Thanks @lmondada.

sacpis (approved):

Renaud-K (approved):

PR #4722, opened 2026-06-11
[testing] use new `CMake cudaq/unittests` infra to refactor `unittests/Optimizer` into `cudaq/unittests/Optimizer` and create `check-cudaq-unit` build target

cabreraam:
> [!NOTE]
> This is the follow on PR to #4714 

## Motivation

## Goal

## Implementation
This PR is the second of two.
This PR:
- refactors the `unittests/Optimizer` tests to  `cudaq/unittests` using the infra from #4714 
- creates a new `check-cudaq-unit` target in a similar way to `llvm/test/Unit` 

## Notes


unittests/Optimizer/CMakeLists.txt
```diff
@@ -1,86 +0,0 @@
...
-target_include_directories(OptimizerUnitTests
-  PRIVATE ${CMAKE_SOURCE_DIR}/runtime
-  PRIVATE ${CMAKE_SOURCE_DIR}/cudaq/lib/Optimizer/Transforms
-)
-
-# MLIR/LLVM is built without RTTI, so we must disable it here too
-# to avoid linker errors (undefined typeinfo for base classes)
-target_compile_options(OptimizerUnitTests PRIVATE -fno-rtti)
-
-gtest_discover_tests(OptimizerUnitTests DISCOVERY_TIMEOUT 120)
```
sacpis, line 41: Seems like we are removing these tests which used to be run with ctest. I see in `cudaq/test/Unit/CMakeLists.txt` that we have added a test `ctest-cudaq-unit` which runs with target as `check-cudaq-unit`. Are we expecting these tests to be executed with `run_tests.sh` script?

sacpis (dismissed):
Overall LGTM. Thanks @cabreraam. Should we add the following to `run_tests.sh`?

```
ctest --output-on-failure --test-dir "$build_dir" -R ctest-cudaq-unit
```

unittests/Optimizer/CMakeLists.txt
cabreraam: That's a great point. I had not considered how this would affect `run_script.sh`, though @schweitzpgi did warn me that I might run into issues with our scripts :smile:.

The solution I have now is to modify the regex https://github.com/cabreraam/cuda-quantum/blob/661a30425397e758d1e3d16eb7a3ccf4aae7001b/scripts/run_tests.sh#L84 to not exclude `ctest-cudaq-unit` (which I created [here](https://github.com/cabreraam/cuda-quantum/blob/661a30425397e758d1e3d16eb7a3ccf4aae7001b/cudaq/test/Unit/CMakeLists.txt#L30)). Now, it specifically matches `ctest-cudaq`, `ctest-targettests`, and `pycudaq-mlir`. So now, during that `ctest` phase, `ctest-cudaq-unit` (which wraps `check-cudaq-unit`) will run. I thought about creating a new phase specifically for `check-cudaq-unit`, but the way that the commands are written, it's in phases of
- `ctest`
- `lit`
- `lit`
- `lit`
- `pytest`
 
so I wanted to just maintain that structure of running all of the CUDA-Q `ctest` tests together.

cabreraam: Just validated with
```
ctest --test-dir build -N -E "^(ctest-cudaq|ctest-targettests|pycudaq-mlir)$"
```

and observed in the output:

```
  Test #1844: ctest-cudaq-unit
```

sacpis (approved):
LGTM. Thanks @cabreraam.

PR #4724, opened 2026-06-11
Follow-up to PR #4688: diagnose iterating an unbound `measure_handle` vector

khalatepradnya:
### Summary

* Follow-up to PR #4688

* Diagnose `for (bool b : v)` over a `std::vector<measure_handle>` whose handles were never measured. The loop discriminates each element, so an unbound vector lowers `quake.discriminate` onto a `cc.undef` and reads a result that never existed.


cabreraam (approved):
Okay, I've given it a pretty thorough look. LGTM!

PR #4726, opened 2026-06-11
Allow non-quantum calls in AssignWireIndices

atgeller:


1tnguyen (approved):
LGTM 👍 Thanks @atgeller

cabreraam (approved):
LGTM! thanks @atgeller

cudaq/lib/Optimizer/Transforms/WiresToWiresets.cpp
```diff
@@ -75,11 +75,21 @@ struct AssignWireIndicesPass
...
-    // TODO: someday we may want to allow calls to non-quantum functions
+    // Only bail out if there are calls to quantum kernels; non-quantum calls
+    // are allowed through.
     if (cudaq::opt::hasCallOp(func)) {
-      func.emitRemark(
-          "AssignWireIndicesPass function has calls, pass will not be run.");
-      return;
+      auto module = func->getParentOfType<ModuleOp>();
+      bool hasQuantumCall = false;
+      func.walk([&](func::CallOp call) {
```
schweitzpgi, line 83: Calling a function with wires is supported now with the `quake.call_by_ref`. We should be using that.

atgeller: Added handling for that (anti-support?), and apply as well just in case.

PR #4728, opened 2026-06-12
Fix ORCA backend integration test

Omar-ORCA:
ORCA backend integration is failing with an "Failed to set CUDA device" error.
Error occurs when host machine does not have a GPU.

### DIFF
--- a/runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
+++ b/runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
@@ -70,10 +70,13 @@ public:
   bool supports_task_distribution() const override { return true; }
 
   void beginExecution() override {
-    // Set the current CUDA device for this thread based on the QPU ID in the
-    // execution context.
+    // Only set the CUDA device when GPU-backed QPUs are in use. When a
+    // non-GPU platform (e.g. ORCA) replaces the default QPUs via
+    // setTargetBackend.
     auto qid = cudaq::getCurrentQpuId();
-    cudaq::setCudaDevice(qid);
+    int nDevices = cudaq::getCudaDeviceCount();
+    if (nDevices > 0)
+      cudaq::setCudaDevice(qid);
     // Base implementation of beginExecution will be called after this.
     cudaq::quantum_platform::beginExecution();
   }
--- a/runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
+++ b/runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
@@ -70,9 +70,9 @@ public:
   bool supports_task_distribution() const override { return true; }
 
   void beginExecution() override {
-    // Only set the CUDA device when GPU-backed QPUs are in use. When a
-    // non-GPU platform (e.g. ORCA) replaces the default QPUs via
-    // setTargetBackend.
+    // Only set the CUDA device when GPU-backed QPUs are active.
+    // Non-GPU platforms (e.g. ORCA) that replace the default QPUs
+    // via setTargetBackend do not require a CUDA device assignment.
     auto qid = cudaq::getCurrentQpuId();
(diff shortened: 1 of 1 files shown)


runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
```diff
@@ -70,10 +70,13 @@ class MultiQPUQuantumPlatform : public cudaq::quantum_platform {
   bool supports_task_distribution() const override { return true; }
 
   void beginExecution() override {
-    // Set the current CUDA device for this thread based on the QPU ID in the
-    // execution context.
+    // Only set the CUDA device when GPU-backed QPUs are in use. When a
+    // non-GPU platform (e.g. ORCA) replaces the default QPUs via
```
sacpis: nit: incomplete sentence?

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
```diff
@@ -70,10 +70,13 @@ class MultiQPUQuantumPlatform : public cudaq::quantum_platform {
...
+    // setTargetBackend.
     auto qid = cudaq::getCurrentQpuId();
-    cudaq::setCudaDevice(qid);
+    int nDevices = cudaq::getCudaDeviceCount();
+    if (nDevices > 0)
+      cudaq::setCudaDevice(qid);
```
sacpis, line 79: We already have this check in the `MultiQPUQuantumPlatform` constructor. Can we just throw an error in the else part there?

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
Omar-ORCA: Sentence is complete.

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
Omar-ORCA: The orca backend uses mqpu as the platform-library but does not requires gpus. So this is not an error in this case.
The constructor is being skipped as `nDevices == 0`.

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
sacpis: Sorry, I should have mentioned about the sentence starting with "When...", Is it a continuation of the previous sentence? If yes, then I am a bit confused with the fullstop.

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
sacpis: Thanks @Omar-ORCA.

>> So this is not an error in this case.

Yes, definitely. It is not an error. Just thought of checking it in one place.

>> The constructor is being skipped as nDevices == 0.

As the Orca backend does not require gpus, should we remove this check from the constructor?

Omar-ORCA: I'm not sure if it's being used by other mqpu backends like custatevec.

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
Omar-ORCA: Sorry about that. It has a better description now.

sacpis:
> /ok to test 59faeff

runtime/cudaq/platform/mqpu/MultiQPUPlatform.cpp
sacpis: For now we can keep this check. Thanks @Omar-ORCA.

sacpis (approved):
LGTM. Thanks @Omar-ORCA.

PR #4729, opened 2026-06-12
Enable HOST_CALL entries in device_call host_dispatch channel

vedika-saravanan:
## Description

This updates the `device_call` `host_dispatch` channel so QPU/kernel-initiated realtime device calls can invoke `CUDAQ_DISPATCH_HOST_CALL` entries, not only `CUDAQ_DISPATCH_GRAPH_LAUNCH` entries.

## Changes

- Allow `host_dispatch` tables to contain `CUDAQ_DISPATCH_HOST_CALL` handlers.
- Require a pinned mailbox only when graph-launch entries are present.
- Allow the realtime host dispatcher thread to start for HOST_CALL-only tables.
- Add a unit test covering device-call frame dispatch through a HOST_CALL handler.
- Update comments/docs to clarify the host-dispatch boundary.

## Notes

It does not add CPU RoCE, shared-memory, or network-specific behavior to `cudaq-realtime`; transport-specific behavior remains behind the existing channel/transport interfaces.


runtime/internal/device_call/HostDispatchChannel.cpp
```diff
@@ -37,18 +37,27 @@ class HostDispatchChannel : public DeviceCallChannel {
...
+          "table entry");
+
+    // Graph-launch workers require a pinned, mapped mailbox. HOST_CALL
+    // handlers execute inline on the host dispatcher thread and do not use it.
+    if (hasGraphLaunchEntry() && !mailbox)
+      throw DeviceCallError(
+          DeviceCallStatus::InvalidArgument,
+          "host device_call graph-launch entries require a pinned, mapped "
+          "mailbox "
+          "(cudaHostAlloc with cudaHostAllocMapped) sized for the "
```
1tnguyen: nit: the format of this is a bit weird (mailbox on its own line)

1tnguyen (approved):
LGTM 👍 Thanks @vedika-saravanan

runtime/internal/device_call/HostDispatchChannel.cpp
vedika-saravanan: fixed in 1e499962f0367e821a229e7eed3a2442a79a0dd7

PR #4731, opened 2026-06-12
Fix multi-qubit noise broadcast in Stim MSM mode

bmhowe23:
~~_Note: it's debatable whether we should pursue merging this, as this bug only seems to affect MSM mode, which could go away now that the new DEM mode is being used, but I am posting anyway._~~ Edit: as long as MSM mode is here, it should be correct.

Related: https://github.com/NVIDIA/cudaqx/issues/606

A noise channel attached to a multi-qubit gate is broadcast independently across each group of `num_targets` qubits in the sampling path (via Stim's multi-target semantics) and the DEM path. The MSM path, which injects errors into `x_table`/`z_table` by hand, only iterated `num_targets` and indexed `qubits[0]`, so for a single-qubit-tagged channel applied to >1 qubit (e.g. the QEC library's `two_qubit_depolarization` / `two_qubit_bitflip`) every qubit past the first was silently dropped from the MSM and any MSM-derived DEM, while the sampled syndromes still contained that noise.

Make the MSM branch mirror the broadcast: enumerate each group of `num_targets` qubits, giving each group its own error id, and update the `msm_size` pass's mechanism count to match. Throw if the qubit count is not a positive multiple of `num_targets`.

Add Stim MSM regression tests for the two-qubit depolarization and bit-flip broadcast cases.


1tnguyen (approved):
LGTM 👍

PR #4732, opened 2026-06-12
Harden the verifier for custom ops

khalatepradnya:
This PR adds stronger validation for `quake.custom_unitary_constant` matrix operands and aligns the NVQIR runtime guard with the actual 4^N matrix-size overflow boundary.

Note that the Python frontend already had most of these checks in place, but they escaped C++ frontend.


1tnguyen (approved):
LGTM 👍

PR #4733, opened 2026-06-13
Support trivial control flow in Mapping

atgeller:
Adds support for `IfOp`s with single qubit operations only for Mapping.


cudaq/lib/Optimizer/Transforms/Mapping.cpp
1tnguyen, line 2018: Looks like we collect `measureOrder` here (regardless of the scope) for 

```
// Ensure that the original measurement ordering is still honored by moving
// the measurements to the end (in their original order). Note that we must
// move the users of those measurements to the end as well.
```

For safety, we might also reject `MeasurementInterface` in the `Reject any cc::IfOp branch ...` validation above.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -349,6 +382,60 @@ SabreRouter::Swap SabreRouter::chooseSwap() {
   return candidates[minIdx];
 }
 
+void SabreRouter::rewireBranchRegion(
```
taalexander: Could we leave a comment on what this is doing here/policy is taken.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -651,6 +738,33 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
       return;
     }
 
+    // Reject any cc::IfOp branch that contains a 2-qubit quantum gate.
+    // Single-qubit gates never need routing; 2-qubit gates inside branches
+    // would require placement decisions that cross the branch boundary.
+    auto branchCheckResult = func.walk([&](cudaq::cc::IfOp ifOp) {
```
taalexander: I think this is a limitation of the new pass contract that is maybe a bit implicit (That only single-qubit operations are supported within if/else ops for routing). Maybe pulling this into a helper like `containsUnsupportedBranchQuantumOp()` would make this more explicit?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -711,84 +825,127 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+          }
+          // Map IfOp wire results to the same virtual qubits as the linearArgs.
+          unsigned linearIdx = 0;
+          for (auto result : ifOp->getResults()) {
+            if (isa<cudaq::quake::WireType>(result.getType())) {
+              auto virt = wireToVirtualQ[linearArgs[linearIdx++]];
+              wireToVirtualQ.insert({result, virt});
+              finalQubitWire[virt.index] = result;
+            }
+          }
```
taalexander: Could we refactor this into it's own logic? (Ideally for most cases in the if/else if chain).

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -711,84 +825,127 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+                "reference semantics");
             signalPassFailure();
           }
-          LLVM_DEBUG(llvm::dbgs() << "operator with >2 qubits not expected");
+          LLVM_DEBUG(llvm::dbgs()
+                     << "quantum reference semantics not expected");
+          analysisOk = false;
           return;
-        }
+        } else if (auto ifOp = dyn_cast<cudaq::cc::IfOp>(op)) {
```
taalexander: Right now does this assume the if/else if results are positionally corresponding to the input wire args? What happens if the results are a classical bit? Could this break the mapping? 

Perhaps this should be factored into a helper method as well.

taalexander (changes requested):
Hey Adam this is a nice start! Let's merge #4678 first before this and it might require a bit of rethinking to integrate nicely.

Next we should think about what is required for arbitrary control flow and subroutines. The first approach we would likely want is a block-level contract where each block is entered with a known logical-to-physical boundary mapping and exits with a known mapping, with the top-level mapper choosing those boundary qubits statically for some heuristic mapping score that works reasonably well given we can't know how what the complete control-flow is at runtime. This PR can (and should) stay scoped to single-qubit `cc.if` regions, but documenting that as a temporary subset would make the path to general control flow clearer. Happy to talk more about this next week!

I believe we're missing a test or two in the PR as well 🙏.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
atgeller: I did some refactoring on top of the new SABRE implementation that handles this better.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
atgeller: The refactoring on top of the new SABRE implementation no longer makes this assumption.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
atgeller: Good point. Added a check.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
atgeller: I believe this function is no longer relevant.

atgeller:
@taalexander @1tnguyen I've reworked this PR on top of #4678, and based on the plan towards broader control flow support in mapping.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -501,18 +526,29 @@ RoutingProblem buildReverseProblem(const RoutingProblem &forward) {
     reverse.nodes.push_back(makeReverseNode(node));
   }
 
+  // Splice out IfOps transitively so unitary gates on either side of an IfOp
+  // still form a dependency edge in the reverse problem.
+  std::function<void(RoutingProblem::NodeRef, RoutingProblem::NodeRef,
```
taalexander, line 531: Thanks for looking at and resolving this.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1131,43 +1175,79 @@ class RoutingEmitter {
...
-        auto q0 = ev.phys[0];
-        auto q1 = ev.phys[1];
-        auto swap = cudaq::quake::SwapOp::create(
-            builder, builder.getUnknownLoc(), TypeRange{wireType, wireType},
-            false, ValueRange{}, ValueRange{},
-            ValueRange{phyToWire[q0.index], phyToWire[q1.index]},
-            DenseBoolArrayAttr{});
-        phyToWire[q0.index] = swap.getResult(0);
-        phyToWire[q1.index] = swap.getResult(1);
+    std::function<void(Block &)> emitBlock;
```
taalexander: Might be cleaner if this is refactored out to a separate class method.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1601,106 +1727,159 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+                  analysisOk = false;
+                  return;
+                }
+              }
+            } else if (auto ifOp = dyn_cast<cudaq::cc::IfOp>(op)) {
+              auto linearArgs = ifOp.getLinearArgs();
+              for (Region *region : ifOp.getRegions()) {
+                if (region->empty())
+                  continue;
+                assert(region->hasOneBlock());
```
taalexander: Why only one block? I thought the IfOp has multiple regions?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1601,106 +1727,159 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
-                         << '\n';
-            func.emitError("The mapper requires value semantics.");
-            signalPassFailure();
+    // Recursive analysis: walk `b` and all IfOp branch regions, populating
+    // wireToVirtualQ, finalQubitWire, sources, and interaction data.
+    // parentOp is the enclosing cc::IfOp when recursing into a branch; null at
+    // the top level. cc.continue uses it to set (first branch) or verify (else
+    // branch) the parent's wire result VQs directly in wireToVirtualQ.
+    bool analysisOk = true;
+    std::function<void(Block &, bool, Operation *)> analyzeBlock =
```
taalexander: Perhaps this should also be a method instead of a lambda for readability?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1781,20 +1960,94 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
         search.run(seeds, numV, numPhy);
-    RoutingResult &best = selection.result;
     cudaq::Placement &bestLayout = selection.finalLayout;
 
+    // For each block (outer and all branch blocks), store its RoutingResult in
+    // blockResults. Routes branch blocks using the placement at the point each
+    // cc::IfOp is reached as the seed. Recurses for nested IfOps.
+    DenseMap<Block *, RoutingResult> blockResults;
+    std::function<void(Block &, RoutingResult)> buildBlockResults;
+    buildBlockResults = [&](Block &blk, RoutingResult flat) {
```
taalexander: Could this be a function/method?

cudaq/test/Transforms/mapping_if.qke
```diff
@@ -0,0 +1,238 @@
+// ========================================================================== //
```
taalexander, line 1: Should some of these tests be checking `mapping_v2p`? I find it a bit hard to see the mappings that were found. Ideally we sanity check them against one of the other greedy/sabre mapping tests with trivial ifops inserted at somepoint to make sure it's otherwise idempotent on mapping?

taalexander (changes requested):
Looks a lot better. Mostly a few minor changes suggested. I think we can get this in today.

I was wondering if the intent is to support `cc.br/cc.cond_br`? I think these are trivial to support right now but are likely being disallowed by your one region check (that I commented on).

cudaq/lib/Optimizer/Transforms/Mapping.cpp
atgeller: It does has multiple regions (then/else), which is handled. As a starter, I'm assuming only one block per region (basically that the program has not yet been lowered to CFG) as a simplification.

cudaq/test/Transforms/mapping_if.qke
atgeller: I've added CHECK lines, and a couple tests to verify that adding the if doesn't change anything.

taalexander (approved):
LGTM, thank you for changes!

PR #4734, opened 2026-06-13
Moving future nvq++ tests under nvqpp directory

Renaud-K:
This is the first step towards converting tests to using nvq++. All the tests which have a __qpu__ kernel are moved under a new nvqpp directory. 

The idea is that in this directory tree the CXX compiler will be set to nvq++, in a follow-up PR,  with
```
set(CMAKE_CXX_COMPILER "${CMAKE_BINARY_DIR}/bin/nvq++")
```

The cmake files will be updated to use the nvq++ command line options to set the target and nvq++ will be used as the linker. So the cmake files are expected to change significantly. I decided it would be better to do this transition in 2 steps. 

This PR only moves files. Nothing is changing yet with respect to removing library mode. The compiler is still the C++ compiler throughout.


Renaud-K:
> /ok to test adbe552df8d8083f0fea9ed29eb6351831a179d3

1tnguyen (approved):
LGTM 👍

sacpis (approved):
LGTM. Thanks @Renaud-K.

PR #4736, opened 2026-06-13
[testing] use new CMake cudaq/unittests infra to refactor unittests/Optimizer into cudaq/unittests/Optimizer and create check-cudaq-unit build target

cabreraam:
> [!NOTE]
> This is exactly the same PR as #4722. It was erroneously closed. 

---

> [!NOTE]
> This is the follow on PR to #4714 

## Motivation

## Goal

## Implementation
This PR is the second of two.
This PR:
- refactors the `unittests/Optimizer` tests to  `cudaq/unittests` using the infra from #4714 
- creates a new `check-cudaq-unit` target in a similar way to `llvm/test/Unit` 

## Notes


sacpis (approved):
LGTM. Thanks @cabreraam.

PR #4742, opened 2026-06-16
Add target profiling traces to ast bridge

taalexander:
Some useful traces as part of my compiler profiling work to the ast bridge identifying some hot spots here.


sacpis (approved):
LGTM. Thanks @taalexander.

PR #4743, opened 2026-06-16
Removing warnedNamedMeasurements from ExecutionContext

Renaud-K:
We do not really need the flag on the executionContext or the policy. In the worse case we issue the warning once during JIT compilation and once during simulation but I don't think that we should be trying too hard to carry the information over from one to the next just so the warning is only issue once. I feel that it is unnecessary churn. 

The simulator already holds state, so it keeps track of whether the warning was issued or not. What we really need is only to make sure that it is not issued for every operation of every iteration. The test has been improved to check for that.

### DIFF
--- a/runtime/common/BaseRemoteRESTQPU.h
+++ b/runtime/common/BaseRemoteRESTQPU.h
@@ -299,10 +299,6 @@ public:
 
       compiled = compiler.runPassPipeline(kernelName, moduleOp, args, true,
                                           std::move(context));
-      if constexpr (std::is_same_v<Policy, sample_policy>) {
-        if (compiler.hasWarnedNamedMeasurements())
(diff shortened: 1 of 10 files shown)


runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -493,7 +488,6 @@ class BaseRemoteRESTQPU : public QPU {
     for (std::size_t i = 0; i < codes.size(); i++) {
       cudaq::ExecutionContext context("sample", localShots);
       // Avoid emitting the warning again during execution
```
lmondada: Also remove the comment:

```suggestion
```

lmondada (approved):
Thanks Renaud, music to my ears!

PR #4744, opened 2026-06-16
Support `reset` op in qubit mapping

1tnguyen:
Reset op is equivalent to a quantum operation (`OperatorInterface`): one wire in, one wire out. However, we didn't put it in the `isSupportedMappingOperation`, hence the mapping failed for any programs that have `reset`.

This PR fixed that and added a regression test.


atgeller (approved):
LGTM, thanks @1tnguyen!

PR #4746, opened 2026-06-16
Reuse the cuStateVec handle across kernel executions

ikkoham:
## Summary
`CuStateVecCircuitSimulator` calls `custatevecCreate` on every state-vector  allocation and `custatevecDestroy` in `deallocateStateImpl`, i.e. it creates+destroys the cuStateVec handle on **every kernel execution**. The handle is a device-level context (independent of any state vector), so it can be created once per device and reused. This adds `ensureHandle()` which creates the handle lazily and recreates it only when the active CUDA device changes, drops the per-deallocation destroy, and destroys the handle once in the destructor.

  ## Benchmark

  | call | before | after |
  |------|-------:|------:|
  | `get_state` | 2.53 ms | 0.16 ms (~15×) |
  | `observe` (shots=0) | 2.60 ms | 0.25 ms (~10×) |
  | `sample` (1k shots) | 2.53 ms | 0.53 ms (~4.7×) |
  | `run` (100 shots) | 201 ms | 10 ms (~20×) |

These benchmarks do not include the first execution. (warm)

<details>
<summary>Benchmark kernels</summary>

```python
import time, cudaq
from cudaq import spin
NQ, THETA, LAYERS = 5, 0.5, 1

# for observe and get_state
@cudaq.kernel
def ansatz(n: int, theta: float, layers: int):
    q = cudaq.qvector(n)
    for _ in range(layers):
        h(q[0])
        for i in range(n - 1):
            x.ctrl(q[i], q[i + 1])
        ry(theta, q[0])

# for sample
@cudaq.kernel
def ansatz_measured(n: int, theta: float, layers: int):
    q = cudaq.qvector(n)
    for _ in range(layers):
        h(q[0])
        for i in range(n - 1):
            x.ctrl(q[i], q[i + 1])
        ry(theta, q[0])
    mz(q)

# for run
@cudaq.kernel
def ansatz_returning(n: int, theta: float, layers: int) -> bool:
    q = cudaq.qvector(n)
    for _ in range(layers):
        h(q[0])
        for i in range(n - 1):
            x.ctrl(q[i], q[i + 1])
        ry(theta, q[0])
    return mz(q[0])
```
</details>

The cuStateVec handle create+destroy round-trip is ~1–2 ms (warm), paid per launch before this change. `run` benefits most because it re-executes the kernel per shot, paying N handle create/destroys per call.


ikkoham:
> /ok to test [7588733](https://github.com/NVIDIA/cuda-quantum/pull/4746/commits/7588733a1aae0d115e7e0e7bc4bc6108dad59d15)

1tnguyen (approved):
LGTM 👍 Thanks @ikkoham

khalatepradnya:
> /ok to test 687b217

PR #4748, opened 2026-06-16
[CI] Suppress unwanted ghcr deployment records in CI

mitchdz:
The ghcr-ci environment only gates CI secrets/access and is not a real deployment, but using it implicitly created a deployment record (and the "deployed to ghcr-ci" message on PRs). Set deployment: false for ghcr-ci in the shared dev_environment and docker_images workflows so the record is never created, while ghcr-deployment and other environments are unaffected.


1tnguyen (approved):
LGTM 👍

PR #4693, opened 2026-06-07
[unitaryhack] Support unitary synthesis for 3+ qubit custom operations (QSD)

thedaemon-wizard:
# [custom op] Support unitary synthesis for 3+ qubit operations (QSD)

Closes #2242

## Summary

The `unitary-synthesis` optimization pass
(`lib/Optimizer/Transforms/UnitarySynthesis.cpp`) decomposes the matrix of a
custom operation (`cudaq.register_operation` / `CUDAQ_REGISTER_OPERATION`) into a
sequence of native gates. Until now it only handled **1-qubit (ZYZ)** and
**2-qubit (KAK)** operations; any custom operation acting on **3 or more qubits**
(e.g. a Toffoli) failed to legalize `quake.custom_unitary_constant` when targeting
hardware backends such as Quantinuum.

This PR adds a general **recursive Quantum Shannon Decomposition (QSD)** so that
a `2^n x 2^n` custom unitary can be synthesized for **3 ≤ n ≤ 5** (matrix
dimension 8–32), including multi-controlled operations such as a 4-qubit `cccx`
and 5-qubit `c4x`. The existing `TwoQubitOpKAK` (`n = 2`) and `OneQubitOpZYZ`
(`n = 1`) decomposers are reused as the recursion base cases; the 1-qubit base
case's Y-angle extraction is made numerically robust (`atan2` of magnitudes
instead of `acos`/`asin`) so the near-degenerate sub-blocks that arise deep in
the recursion are handled correctly. Per review, the range is capped at 5 qubits
(the `4^n` CNOT growth makes 6q+ impractical); larger power-of-two dimensions are
left unchanged in a **composable** way (the pass emits a warning / `LLVM_DEBUG`
note and does not modify the IR — no `emitError`, no `signalPassFailure`). To
handle the advertised 3–5 qubit range **predictably**, the always-on
reconstruction self-checks use a tolerance (`RECON_TOL = 1e-5`) that is *separate*
from, and looser than, the tight input-unitarity contract (`TOL = 1e-7`). The
split is calibrated to measured behavior, not chosen arbitrarily: a correctly
synthesized unitary reconstructs to ~1e-10 when well-conditioned, while rare
near-degenerate sub-blocks push the residual up to ~8e-7 even though the emitted
circuit still reproduces the target to <4e-7 end-to-end; a genuinely wrong
decomposition reconstructs to O(0.1). `RECON_TOL` therefore sits ~1 order above
[...]

PR #4751, opened 2026-06-18
[custom op] Drive unitary-synthesis with the dialect-conversion framework

### LINKED ISSUE #2242: [custom op] Support unitary synthesis for 3+ qubit operations
### Required prerequisites

### Describe the feature

Relates to issue #1475 

#### Description
CUDA-Q implements a `unitary-synthesis` pass defined [here](lib/Optimizer/Transforms/UnitarySynthesis.cpp).
As of October 1, 2024, this pass supports decomposition of 1-qubit and 2-qubit custom operations.

Extend this pass by implementing 3-qubit operation synthesis. 

#### Steps:
 - Choose an appropriate algorithm. Following references can be considered:
   - Cosine-sine decomposition: https://arxiv.org/pdf/quant-ph/0404089
   - Quantum Shannon decomposition: https://arxiv.org/pdf/quant-ph/0406176
 - Extend the `Decomposer` class with new class which implements the 3-qubit decomposition.
 - Override `decompose` method which encapsulates the logic for decomposition, and
 - Override `emitDecomposedFuncOp` method which translates the result into native gate set.
 - Update caller switch case [here](https://github.com/NVIDIA/cuda-quantum/blob/main/lib/Optimizer/Transforms/UnitarySynthesis.cpp#L464).
    For example,
 ```
case 8: {
        auto csd = ThreeQubitOpCSD(unitary);   // rename according to the chosen algorithm
        csd.emitDecomposedFuncOp(customOp, rewriter, funcName);
      } break;
 ```
- Can reuse the `OneQubitOpZYZ ` defined [here](https://github.com/NVIDIA/cuda-quantum/blob/main/lib/Optimizer/Transforms/UnitarySynthesis.cpp#L73) for any 1-q decomposition.
- Following tests should now pass:
  - Python: [here](https://github.com/NVIDIA/cuda-quantum/blob/main/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py#L290)
     - Remove `with pytest.raises(RuntimeError):` and directly invoke `sample` on line#306.
     - To run, `python3 -m pytest -rP ../python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py::test_3q_unitary_synthesis`
  - C++: [here](https://github.com/NVIDIA/cuda-quantum/blob/main/targettests/execution/custom_operation_toffoli.cpp#L10)
     - To run, `nvq++ --target quantinuum --emulate targettests/execution/custom_operation_toffoli.cpp && ./a.out`
- Additional test(s) with random 8x8 unitary matrices should be added.

thedaemon-wizard:
# [custom op] Drive unitary-synthesis with the dialect-conversion framework

Follow-up to #4693 (recursive Quantum Shannon Decomposition for 3–5 qubit custom
operations, which closed #2242). As discussed there
(https://github.com/NVIDIA/cuda-quantum/pull/4693#issuecomment-4703748770), the
conversion-rewriter improvement was deferred to a separate PR — this is that PR.

## Summary

The `unitary-synthesis` pass
(`lib/Optimizer/Transforms/UnitarySynthesis.cpp`) is reworked from a **greedy** rewrite
driver (`applyPatternsGreedily` + `OpRewritePattern`) to the **dialect-conversion**
framework (`ConversionTarget` + `applyPartialConversion` + `OpConversionPattern`).

This is a **driver-only** change: the decomposition math and the emitted gate sequences
are unchanged, so every existing `UnitarySynthesis` test passes as-is.

## Motivation

With a conversion target, the "do not commit a rewrite unless the legality constraints are
met" guarantee is expressed by the framework rather than by hand. Previously the pass
enforced composability by *validating the whole decomposition before emitting* and
returning `failure()` to leave the IR untouched. With dialect conversion, a
`quake.custom_unitary_constant` is simply marked **illegal** (must be rewritten) exactly
when a decomposed replacement kernel can be produced for it; everything else stays
**legal** and is left untouched. The pass remains composable — unsupported dimensions,
non-unitary matrices, and the rare reconstruction miss are left unchanged with a warning /
`LLVM_DEBUG` note, never a `signalPassFailure` or a crash.

## What changed

- **Driver:** `runOnOperation` now builds a `ConversionTarget` and runs a single
  module-level `applyPartialConversion`. `CustomUnitaryPattern` is an
  `OpConversionPattern<quake::CustomUnitaryConstantOp>` whose body is a pure structural
  replacement to `quake.apply`.
- **Legality predicate:** a custom op is illegal iff a decomposed kernel exists for its
  generator; this is what enforces composability.
- **Kernel pre-generation:** the decomposed replacement kernels are generated once per
  generator with a plain `OpBuilder`, walking the custom ops in reverse so the kernels keep
  their original module order independent of the conversion driver's traversal order. Each
  generator matrix is classified/decomposed at most once via a small cache.
- **API:** `emitDecomposedFuncOp` / the gray-code `emitMux` now take an `OpBuilder &` so the
  kernels are built outside the conversion process.

The design intentionally does **not** rely on conversion rollback: kernels are
pre-generated and a custom op is only rewritten once it is known-legalizable, so there is
nothing to undo. This matches the direction of MLIR's One-Shot Dialect Conversion driver,
where `ConversionConfig::allowPatternRollback` now defaults to `false`
(https://github.com/llvm/llvm-project/pull/151865).

## Verification

Built `cudaq-opt` against LLVM and ran every `test/Transforms/UnitarySynthesis/*.qke`
with its RUN lines:

- **All FileCheck + CircuitCheck RUN lines pass (19/19 across the 11 tests)** — the 8
  pre-existing ZYZ/KAK tests, the 3-qubit `random_unitary-5`, and the CircuitCheck-only
  4-qubit/5-qubit `random_unitary-6/-7`. Because the emitted IR is unchanged, every
  FileCheck still matches and CircuitCheck still confirms unitary equivalence.
- **Composability:** feeding a power-of-two `dim = 64` (6-qubit) op and a non-unitary
  supported-dimension op through `cudaq-opt --unitary-synthesis` leaves the
  `quake.custom_unitary_constant` unchanged with `returncode 0` (the dim-64 case prints the
  `LLVM_DEBUG` cap note; the non-unitary case prints the "matrix must be unitary" warning).
  No `signalPassFailure`, no crash.
- `clang-format` clean.

## AI usage disclosure

This change was prepared with AI assistance (Claude) for code refactoring, test execution,
and drafting; all changes were reviewed and verified locally by the author.


khalatepradnya:
> /ok to test 9e69867

cudaq/lib/Optimizer/Transforms/UnitarySynthesis.cpp
```diff
@@ -922,18 +930,89 @@ class UnitarySynthesisPass
...
     }
+
+    /// Use the dialect-conversion framework to replace each synthesizable
+    /// custom operation with a `quake.apply` of its decomposed kernel. A custom
+    /// op is *illegal* (must be rewritten) exactly when such a kernel exists;
+    /// everything else is *legal* and left untouched, so the pass stays
+    /// composable (no `signalPassFailure`, no crash) -- the rewrite is never
+    /// committed unless the legality constraint is met.
+    ConversionTarget target(*ctx);
+    target.markUnknownOpDynamicallyLegal([](Operation *) { return true; });
```
schweitzpgi: This legality is extremely open ended. What is it catching? Maybe we can get away with just making the Quake dialect legal or even just `quake.apply`?

schweitzpgi:
Thanks for updating this. Much appreciated.

cudaq/lib/Optimizer/Transforms/UnitarySynthesis.cpp
thedaemon-wizard: Hi, @schweitzpgi 
Good catch — narrowed it. You're right that the catch-all was doing more than needed: a
`applyPartialConversion` already leaves every op that isn't explicitly marked illegal
unconverted, so the only legality the target actually needs is

```cpp
target.addLegalOp<cudaq::quake::ApplyOp>();          // the single op this pass produces
target.addDynamicallyLegalOp<cudaq::quake::CustomUnitaryConstantOp>(...);  // illegal iff synthesizable
```

I went with your "even just `quake.apply`" suggestion: marking `quake.apply` legal is required
because it's the op the pattern *produces* during conversion (without it the driver can't
legalize the replacement and reports the original op as "failed to legalize"). Everything else
— the pre-generated kernel `func`s and their bodies, `cc.global`, unrelated ops — is left
untouched by the partial conversion, so no blanket is needed. Unsupported / non-unitary /
reconstruction-miss custom ops still simply remain (composable, no `signalPassFailure`).

Verified: all UnitarySynthesis FileCheck + CircuitCheck tests still pass and the composability
cases (dim > 32, non-unitary) are unchanged. Pushed in the latest commit.

khalatepradnya:
> /ok to test b4526c4

cudaq/lib/Optimizer/Transforms/UnitarySynthesis.cpp
schweitzpgi: Thanks for the update.

schweitzpgi (approved):
LGTM. Reduced the code size nicely.

PR #4752, opened 2026-06-21
fix(opt): diagnose adjoint of forward-declared kernel instead of crashing

### LINKED ISSUE #4268: `apply-op-specialization` crashes instead of erroring when adjointing a forward-declared kernel

aryanputta:
## Description

Fixes #4268.

Adjointing a forward-declared `__qpu__` kernel could crash the optimizer while specializing an apply operation. The crash came from trying to clone an empty callee body and then dereferencing the cloned region's entry block.

## Root cause

Forward-declared kernels are represented as body-less functions. `apply-op-specialization` may still see a `quake.apply <adj>` targeting that function, but the pass cannot assume full-program visibility: the body may be supplied later, for example at JIT time.

The older behavior tried to specialize immediately and could produce either:

- an assertion while reading the empty body, or
- an invalid `func.call` to a specialized symbol that was never created.

## Fix

The PR now follows the maintainer-requested split:

- `apply-op-specialization` skips body-less callees and leaves the original `quake.apply` in place instead of failing the pass.
- `ApplyOpPattern` now checks that the computed specialized symbol exists before replacing a `quake.apply` with `func.call`.
- `convert-to-qir-api` diagnoses any `quake.apply` that survives to codegen, because at that point the apply is unambiguously unlowerable.

The diagnostic is:

```text
could not generate the specialized form of kernel 'test'; its body was not available (it may be only forward-declared)
```

## Tests

- `cudaq/test/Transforms/adjoint-forward-declared.qke`: verifies `apply-op-specialization` leaves the unresolved adjoint apply in place instead of crashing or emitting an invalid call.
- `cudaq/test/Transforms/apply-forward-declared-trap.qke`: verifies `convert-to-qir-api` emits the clear diagnostic for a lingering apply.

I have read the CONTRIBUTING document and signed off all commits (DCO).


cudaq/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
```diff
@@ -544,6 +544,18 @@ class ApplySpecializationPass
     auto loc = func.getLoc();
     auto &funcBody = func.getBody();
 
+    // A forward-declared kernel has no body to invert. Cloning its empty
```
schweitzpgi: I thought I had commented on this earlier?

We want passes to be composable. A pass cannot assume that it has full program information whenever it runs, which would be the case with this change. It is entirely possible that the kernel declaration _will_ be supplied as a kernel definition at a later point of the compiler pipeline and thus this failure would be premature and deeply incorrect.

The diagnostic ought to be emitted properly at codegen time when the kernel is "fully processed" and any lingering ApplyOp would be clearly incorrect as we cannot lower it at codegen time.

schweitzpgi (changes requested):

cudaq/lib/Optimizer/Transforms/ApplyOpSpecialization.cpp
schweitzpgi: That said, yes, we don't want to attempt to convert the body of a kernel that has an empty body. Yikes!

aryanputta:
Thanks @schweitzpgi, that makes sense, and you are right that failing in the pass is premature since the body may be supplied later (e.g. at JIT time). Reworked along those lines:

- `apply-op-specialization` no longer emits a diagnostic or fails. It simply skips specializing any kernel whose body is empty and leaves the `quake.apply` in place. The guard sits in `step1`, so it covers the adjoint, control, and adjoint-control paths (all three clone the body and read the entry block). The pass stays composable.
- The diagnostic moved to codegen, in `ApplyOpTrap` (`ConvertToQIRAPI.cpp`). By that point the kernel is fully processed, so a lingering `quake.apply` is unambiguously unlowerable. It now emits a clear error naming the callee before falling back to the trap/poison replacement.

Tests:
- `adjoint-forward-declared.qke`: confirms the pass leaves the apply in place without crashing or failing.
- `apply-forward-declared-trap.qke`: confirms `convert-to-qir-api` emits the diagnostic.

Rebased onto `main`. I do not have an LLVM/MLIR build on hand to run lit locally, so a CI run to confirm the two FileCheck/verify-diagnostics tests would be appreciated whenever you can approve the workflow.

schweitzpgi:
> /ok to test 18d235b

schweitzpgi (approved):
Thank you for the quick work!

aryanputta:
I checked the failed CI run and it was for `18d235b`. The failure was `Transforms/adjoint-forward-declared.qke` still rewriting the unresolved adjoint apply to `func.call @test.adj`.

The current head `b248fe9` adds the missing guard in `ApplyOpPattern`: if the computed variant symbol does not exist, the pass leaves the `quake.apply` intact instead of emitting an invalid `func.call`. This is intended to address that exact failure.

Could you please approve CI for `b248fe9` when convenient?

atgeller:
> /ok to test b248fe9

atgeller:
Looks like the check line in `adjoint-forward-declared.qke` has an extra space:
```
// CHECK: quake.apply <adj> @test
<stdin>:4:2: note: possible intended match here
#12 275.6  quake.apply<adj> @test %arg0 : (!quake.ref) -> () {operandSegmentSizes = array<i32: 0, 0, 1>}
```

schweitzpgi:
> /ok to test 1b37e2b

schweitzpgi:
> /ok to test b8db356

aryanputta:
Thanks for the approval and the CI run. The failure was mine: the FileCheck pattern had a space (`apply <adj>`) but the printer emits `apply<adj>`, so it never matched. Fixed the check to match the printed form and pushed (`ae9a4d5`). Could you re-run when you get a chance?

schweitzpgi:
> /ok to test ae9a4d5

PR #4754, opened 2026-06-22
External backends Plugin Framework

### LINKED ISSUE #4619: Enable External Backend Integration
### Required prerequisites

### Summary

We want CUDA-Q to support backend integrations that are developed, built, and distributed outside the main CUDA-Q repository. This would allow hardware providers, research projects, and compiler/runtime extensions to integrate with CUDA-Q without requiring their target configuration, runtime libraries, or backend-specific assets to be copied into the CUDA-Q installation tree.

### Motivation

Today, adding a new backend effectively requires treating it as part of the CUDA-Q distribution. That makes experimentation and third-party backend development harder than necessary, especially for backends that evolve independently or depend on additional packages, architecture descriptions, compiler extensions, or runtime components.

External backend integration would make CUDA-Q more extensible by allowing backend packages to provide their own target metadata, runtime support, optional architecture descriptions, and backend-specific execution behavior while still appearing to users as normal CUDA-Q targets.

### Proposed Functionality

CUDA-Q should define a supported backend package model where an external package can:

- Advertise one or more CUDA-Q targets.
- Provide the target configuration needed by CUDA-Q.
- Provide any backend-specific runtime library required to execute or submit jobs.
- Register any compiler/runtime extensions needed by that backend.
- Be discovered by CUDA-Q at runtime without modifying the CUDA-Q install tree.

From the user’s perspective, an external backend should behave like a built-in target:

```python
import cudaq

cudaq.set_target("my-backend", ...)
result = cudaq.sample(kernel, shots_count=1000)
```

The integration should be based on a stable backend contract rather than one-off special cases. A backend package should be able to provide its configuration, runtime assets, and optional compiler hooks in a self-contained way. CUDA-Q should discover these packages, make their targets available through the normal target-selection APIs, and route compilation/execution through the backend-provided capabilities.

### Non-Goals

This item does not propose upstreaming any specific backend implementation.

### Success Criteria

- External backends can be installed separately from CUDA-Q.
- CUDA-Q can discover those backends automatically or through a documented configuration mechanism.
- External targets appear through the normal CUDA-Q target APIs.
- Backend packages can provide their own target configuration and runtime assets.

anpaz:
Implements the external QPU plugin mechanism proposed in #4619, enabling third parties to ship CUDA-Q backend targets as independently installable packages — without modifying the cuda-quantum source tree.

After this change, `cudaq.set_target("my-qpu")` and `nvq++ --target=my-qpu` work for targets provided by external packages, with the same user experience as in-tree targets.

### Core runtime changes

- **`cudaq.register_backend_path(path)`** — new public Python API (bound to `LinkedLibraryHolder::registerBackendPath`) that registers an external plugin directory with the runtime
- **`pluginLibDir` field + `pluginYamlPath()` accessor** on `RuntimeTarget` — replaces the old `serverHelperLibDir` with a properly named field and an explicit YAML-path reconstructor used by `BaseRemoteRESTQPU`
- **`plugin-libraries` YAML field** — lists shared libraries to `dlopen` when a target is activated; supports both QPU registration and MLIR pass/dialect registration via `cudaqGetPluginInfo`
- **`%PLUGIN_ROOT%` token substitution** — expanded in all YAML values at parse time so plugins can reference their own bundled data files portably
- **`parseTargetConfig()` / `loadTargetConfig()`** — new helpers in `TargetConfigYaml.cpp` that unify YAML parsing with substitution; adopted by all platform code

### `nvq++` and CLI tooling

- **3-level plugin discovery in `nvq++`** — resolves `--target=<name>` against: in-tree → user scope (`$CUDAQ_PLUGIN_ROOT` / `$XDG_DATA_HOME/.../cudaq/plugins`) → system scope (`${install_dir}/plugins/`)
- **`cudaq-install-plugin`** — new CLI tool to symlink/copy a plugin into the user or system scope; supports `--list`, `--uninstall`, `--copy`, `--system`
- **`CUDAQ_PLUGIN_ROOT`** environment variable — overrides the user-scope plugin directory for both `nvq++` discovery and `cudaq-install-plugin`

### Python plugin discovery

- **`cudaq.backends` entry-point group** — at `import cudaq`, all entry points in this group are loaded; each calls `register_backend_path` to make its targets visible
- **`cudaq.plugins` module** — helper for `--install-nvqpp` post-install hooks (symlinks site-packages into the nvq++ plugin scope)
- Failures are reported via `warnings.warn` with entry-point name and traceback (never silently swallowed)

### Build system

- **`CUDAQ_EXTERNAL_PROJECTS`** CMake variable — mirrors LLVM's pattern; external plugins are added via `add_subdirectory` after in-tree targets, with outputs under `${CMAKE_BINARY_DIR}/external/<name>/`

### Reference plugins (plugins)

| Plugin | Shape | Demonstrates |
|--------|-------|-------------|
| `mock_rest` | REST | `ServerHelper` subclass, auto-loaded via the existing serverhelper convention |

## Testing

- **C++ unit tests** (target_config) — 13 tests covering `registerBackendPath`, `pluginLibDir`, `pluginYamlPath`, `plugin-libraries` dlopen, and `%PLUGIN_ROOT%` substitution
- **Python tests** (test_register_backend_path.py) — 4 tests covering the API, error handling, and entry-point-based auto-discovery
- **Lit tests** — `external_plugin_target.cpp` (nvq++ 3-level discovery + precedence + rpath), `install_plugin.cpp` (full cudaq-install-plugin lifecycle), `list_simulators.config` (filter correctness)
- **Plugin check targets** — `ninja check-mock-rest check-trace-qpu` exercises end-to-end build → register → compile → run for both plugin shapes

## Plugin package layout

```
my_qpu_pkg/
├── targets/my-qpu.yml
├── lib/libcudaq-qpu-my-qpu.so
└── data/              # optional auxiliary files
```

## Closes

Closes #4619


docs/sphinx/examples/plugins/mock_rest/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+# ============================================================================ #
```
schweitzpgi, line 1: Is this part of the build or is it just here as an example template?

.github/workflows/docker_images.yml
```diff
@@ -884,7 +884,7 @@ jobs:
 
   validation:
     name: Validation
-    needs: [metadata, cudaq_image]
+    needs: [metadata, cudaqdev_image, cudaq_image]
```
sacpis, line 915: nit: would be nice to name it as `cudaq_dev_image`.

cudaq/lib/Target/TargetConfig.cpp
```diff
@@ -7,7 +7,43 @@
...
+#include <tuple>
+
+namespace {
+struct NumericVersion {
+  std::uint64_t Major = 0;
+  std::uint64_t Minor = 0;
+  std::uint64_t Patch = 0;
+};
+
+std::optional<NumericVersion> parseNumericVersion(std::string_view value) {
```
sacpis, line 22: If I am reading this function correctly, then it does require version to be in the form of Major.Minor.Patch. But if you issue `cudaq.getVersion()` it returns a version having text as `proto` and/or release tags are also non-numeric. 

Can we please confirm it this works for all cases?

runtime/common/BaseRemoteRESTQPU.h
```diff
@@ -201,12 +201,16 @@ class BaseRemoteRESTQPU : public QPU {
...
     /// from there we can get the URL/PORT and the required MLIR pass pipeline.
     std::string fileName = mutableBackend + std::string(".yml");
-    auto configFilePath = platformPath / fileName;
+    std::filesystem::path configFilePath;
+    auto ymlPathIter = backendConfig.find("__yml_path");
+    if (ymlPathIter != backendConfig.end()) {
+      configFilePath = ymlPathIter->second;
+      backendConfig.erase(ymlPathIter);
+    } else {
+      configFilePath = platformPath / fileName;
```
sacpis: Isn't `__ymp_path` injected only in the Python setTarget path? So a binary built with `nvq++` with the target as the plugin won't have the `__ymp_path` at runtime. That will make it fall back to `plugin.yml` in the /targets dir. I don't think that exists for a user or system scope plugin. In this case, `loadTargetConfig` will read an empty stream and return an empty config without throwing an error.

sacpis (approved):
Overall LGTM. Thanks @anpaz. Left few comments.

lmondada:
Writing this down here for future reference:

Until we have moved the ServerHelper header file and its dependencies into a clearly marked 'public' folder, tracking breaking changes and versioning will not be feasible. We don't check/enforce any API backwards compatibility in our internal C++ APIs and we should expect that PRs will break these APIs without even realizing. 

In other words, we will realistically have to bump the version and break plugins with every CUDAQ release.

docs/sphinx/examples/plugins/mock_rest/CMakeLists.txt
anpaz: The later. This is indeed just an example template to show how to create one of this backends and it is not built as part of the cudaq build. 
I specifically use CUDAQ_EXTERNAL_PROJECTS to set up anyone in a path that is easy to incorporate later into our core branch.
It is still built as part of CI, but as a validation step inside the cudaq_dev docker image.

.github/workflows/docker_images.yml
anpaz: Agree. Out of scope, though.

cudaq/lib/Target/TargetConfig.cpp
anpaz: We do have a unit test covering the current version form, 0.9.0-rc2-developer (see checksExternalTargetVersionCompatibility in unittests/target_config/TargetConfigTester.cpp). The parser intentionally extracts the leading numeric Major.Minor.Patch and ignores a trailing prerelease/build suffix. 
However, this does not cover all historical tag formats: versions such as proto-0.8.0 or 0.15-rc3 would be rejected because they do not begin with a complete three-component numeric version. So the current implementation works for semver-like versions with suffixes, but not every tag form present in the repository. But that should be fine as this is the format we use for our releases.

runtime/common/BaseRemoteRESTQPU.h
anpaz: Thanks for catching this. Indeed we were missing this for the native scenarios. It has been fixed.

cudaq/include/cudaq/Target/TargetConfig.h
```diff
@@ -118,6 +120,8 @@ struct BackendEndConfigEntry {
   std::vector<std::string> CompilerFlags;
   /// Extra libraries to be linked in if any
   std::vector<std::string> LinkLibs;
+  /// List of plugin shared libraries to dlopen at runtime (B3)
```
1tnguyen: nit: what does `B3` mean here?

1tnguyen (approved):
There seems to be a CI failure (empty version string in Debug build). Otherwise, LGTM 👍

cudaq/include/cudaq/Target/TargetConfig.h
anpaz: Thanks for catching this up. B3 was the step in the planning I put together for the implementation, lol.

tomers-qedma:
Cross-link from Qedma: for the `mock_rest` / external-plugin packaging model in this PR, we need an extension for **Fermioniq/QESEM-style server-side `observe`** (full `spin_op` on the wire, not Pauli-split sample jobs + client recombination). Stock `remote_rest` + external `ServerHelper` alone isn't enough for that contract.

Opened **#4926** with preferred options:
- **A:** allow external custom QPU plugins (preferred)
- **B:** `observe-mode: server-side` + observe hooks on `ServerHelper` / `remote_rest`
- **C:** document in-tree-only for observe backends until A/B exist

Happy to prototype a `mock_observe_rest` reference plugin or sync on payload shape if useful.

unittests/target_config/CMakeLists.txt
```diff
@@ -7,11 +7,62 @@
...
+  target_compile_definitions(test_target_config PRIVATE CUDAQ_ENABLE_PYTHON)
+  target_compile_definitions(test_target_config
+    PRIVATE
+      "CUDAQ_DLOPEN_SENTINEL_PLUGIN_PATH=\"$<TARGET_FILE:cudaq-dlopen-sentinel-plugin>\""
+      "CUDAQ_DLOPEN_SENTINEL_PLUGIN_FILENAME=\"$<TARGET_FILE_NAME:cudaq-dlopen-sentinel-plugin>\""
+  )
+  add_dependencies(test_target_config cudaq-dlopen-sentinel-plugin)
+endif()
+
+if(CUDAQ_ENABLE_PYTHON)
```
mitchdz: This whole section is duplicated from above, probably from a merge conflict resolution?

anpaz: yes, thanks for catching.

PR #4756, opened 2026-06-22
Added `stim::ErrorAnalyzer::circuit_to_detector_error_model` flags to `dem_from_kernel`

### LINKED ISSUE #4741: Add `decompose_errors` flag to `dem_from_kernel`
### Required prerequisites

### Describe the feature

For integrating detectors/logical observables into CUDA-Q Libraries, we need access to error-decomposed (i.e. graphlike) DEMs. The current path called by `dem_from_kernel` has hardcoded `decompose_errors=false` in `StimCircuitSimulator.cpp:generateDem`; since the actual decomposition is ultimately relegated to stim, we can get this functionality by just wiring up a flag.

eliotheinrich:
Closes #4741

## Changes
For integrating detectors and logical observables into CUDA-Q Libraries, callers need access to error-decomposed (graph-like) detector error models (DEMs). `dem_from_kernel` currently routes through `StimCircuitSimulator::generateDem`, which hardcodes `decompose_errors=false` when invoking Stim's `ErrorAnalyzer`. Stim already supports decomposition; this change wires a user-facing flag through the existing DEM path. While adding this functionality, we can also add the other flags accepted by the stim error analyzer via a `dem_options` struct which is passed and parsed as a dict from Python. 

- Add optional `dem_options` parameter to C++ and Python binding of `dem_from_kernel`, preserving previous hardcoded defaults.
- Store the options as a field of `ExecutionContext` as `opts` and pass it to `stim::ErrorAnalyzer::circuit_to_detector_error_model` in `StimCircuitSimulator::generateDem`.
- Thread the flags through `runDemFromKernel`, the Python binding (`py_dem.cpp`), and the public header overloads in `dem.h`.

## Testing

- **Python:** Added six new Python tests covering new flags `decompose_errors`, `fold_loops`, `allow_gauge_detectors`, `approximate_disjoint_errors_threshold`, `ignore_decomposition_failures`, and `block_decomposition_from_introducing_remnant_edges`, as well as handling invalid kwargs.
- **C++ target test:** `dem_from_kernel` target test adds a `CORRELATED_XX` case exercising raw vs. decomposed DEM output via FileCheck.


runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -212,10 +212,12 @@ class StimCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
   /// @brief Compute the detector error model from the accumulated
   /// `recordedCircuit` and return it as `.dem` text.
   std::string generateDem() override {
+    auto *ctx = getExecutionContext();
+    const bool decompose_errors = ctx ? ctx->dem_decompose_errors : false;
     stim::DetectorErrorModel dem =
         stim::ErrorAnalyzer::circuit_to_detector_error_model(
             recordedCircuit,
-            /*decompose_errors=*/false,
```
bmhowe23, line 218: As long as we're opening the hood here, I'd recommend that we add support for all the underlying options, not just the one needed for this one particular case.

khalatepradnya: +1
In C++ we can use `dem_options` , dictionary-like param with all the necessary options. Okay to use named parameters in Python.

runtime/common/ExecutionContext.h
```diff
@@ -162,6 +162,11 @@ class ExecutionContext {
 
   /// @brief Slot for the detector error model, as `.dem` text.
   std::string dem_text;
+
+  /// @brief When true, hyper-edge error mechanisms are decomposed into pairs of
+  /// two-detector edges by the Stim ErrorAnalyzer before the `.dem` text is
+  /// returned by `dem_from_kernel`. Defaults to false.
+  bool dem_decompose_errors = false;
```
khalatepradnya: I know we have precedence for adding options for algorithmic primitives in the `ExecutionContext`, but in the light of refactor, there might be a better way. 
Perhaps @Renaud-K / @lmondada can provide pointers - okay to do this temporarily?

runtime/nvqir/stim/StimCircuitSimulator.cpp
```diff
@@ -212,10 +212,12 @@ class StimCircuitSimulator : public nvqir::CircuitSimulatorBase<double> {
...
+            /*decompose_errors=*/decompose_errors,
```
khalatepradnya: Q: Does this flag have connection with `ignore_decomposition_failures` on line # 224?

eliotheinrich: Yes; if `decompose_errors=false`, then `ignore_decomposition_failures` has no effect. If `decompose_errors=true`, then `ignore_decomposition_failures` determines if an un-decomposable error should throw an exception or be ignored. See this new test: https://github.com/NVIDIA/cuda-quantum/pull/4756/changes/41381cdcc3d7348503d330299d011e601898016e#diff-e872e8ece9924ca0c37c2ee9854c903e19750be3cd009e332d8818192df8fb1eR297-R330.

runtime/nvqir/stim/StimCircuitSimulator.cpp
eliotheinrich: Sounds good! Added in 41381cd.

khalatepradnya:
@eliotheinrich - Please look into the Python test failure:
```FAILED python/tests/builder/test_NoiseModel.py::test_depolarization_channel[stim]```
It is consistent and across multiple test envs.

File: python/tests/analysis/test_dem_from_kernel.py [hunk 4]
```diff
@@ -246,5 +246,159 @@ def kernel():
...
+
+    # Without decompose_errors the raw hyperedge is returned fine.
+    dem_raw = cudaq.dem_from_kernel(kernel, noise_model=noise)
+    assert "D0 D1 D2" in dem_raw
+
+    # decompose_errors=True on an odd hyperedge raises.
+    with pytest.raises(Exception):
+        cudaq.dem_from_kernel(kernel, noise_model=noise, decompose_errors=True)
+
+    # ignore_decomposition_failures=True keeps the undecomposable edge as-is
```
bmhowe23, line 351: This comment seems to contradict line 300. Does it silently drop the bad edge or keep it as-is?

eliotheinrich: Yes, that's a typo. It keeps the bad edge as-is. I've clarified this in [e1b56c3](https://github.com/NVIDIA/cuda-quantum/pull/4756/commits/e1b56c369afc839432767c27c89a2dbceadab867).

runtime/common/ExecutionContext.h
```diff
@@ -158,6 +179,9 @@ class ExecutionContext {
 
   /// @brief Slot for the detector error model, as `.dem` text.
   std::string dem_text;
+
+  /// @brief Options forwarded to the Stim ErrorAnalyzer when generating a DEM.
+  dem_options opts;
```
khalatepradnya: nit: Perhaps we should name this variable as `dem_opts` for clarity.

runtime/common/ExecutionContext.h
```diff
@@ -24,6 +24,27 @@ class SimulationState;
...
 class KernelArgs;
 
+/// @brief Options forwarded to
+/// `stim::ErrorAnalyzer::circuit_to_detector_error_model` when generating a
+/// Detector Error Model (DEM) from a kernel.
+struct dem_options {
+  /// Decompose hyper-edge error mechanisms into pairs of two-detector edges.
+  bool decompose_errors = false;
+  /// Fold loop bodies in the circuit for a more compact DEM.
+  bool fold_loops = false;
```
khalatepradnya, line 34: We don't need this yet do we, since CUDA-Q always feeds in unrolled circuit... Okay to keep but no-op for now.

runtime/common/ExecutionContext.h
```diff
@@ -24,6 +24,27 @@ class SimulationState;
...
+  /// Threshold (in [0,1]) for approximating disjoint-error products.
+  /// Set to 0 to disable approximation.
+  double approximate_disjoint_errors_threshold = 0.0;
+  /// Silently skip error mechanisms that cannot be decomposed instead of
+  /// raising an exception.
+  bool ignore_decomposition_failures = false;
+  /// Prevent the decomposer from introducing remnant edges that would otherwise
+  /// be needed to satisfy the decomposition.
+  bool block_decomposition_from_introducing_remnant_edges = false;
+};
```
khalatepradnya, line 46: Note - the docs need to be updated to match these changes, okay to do a follow up PR.

khalatepradnya (commented):
Thanks, @eliotheinrich , for the contribution! This is looking good (with the understanding that adding fields to `ExecutionContext` is temporary workaround and will be refactored in the future).

I think you are missing updates to the remote targets, please update `BaseRemoteRESTQPU.h` [around line 320](https://github.com/NVIDIA/cuda-quantum/blob/1b23e94d78c9ecff94cdb1f996e5fdf128ca1ca0/runtime/common/BaseRemoteRESTQPU.h#L320) so the options are carried forward. Then add a simple C++ and Python emulation test to any of the targets (for e.g. `ionq` or `quantinuum`).

eliotheinrich:
> @eliotheinrich - Please look into the Python test failure: `FAILED python/tests/builder/test_NoiseModel.py::test_depolarization_channel[stim]` It is consistent and across multiple test envs.

It looks like this was caused by some stale state that wasn't cleaned up between concurrent tests; it should be fixed now:
https://github.com/NVIDIA/cuda-quantum/pull/4756/changes/e1b56c369afc839432767c27c89a2dbceadab867#diff-b832960a8965a4f3fab59c6e66cb716c4eea1d0d932539d7b90e2a0cdc07ec43L112-R120

runtime/cudaq/platform/quantum_platform.h
```diff
@@ -107,18 +107,17 @@ class quantum_platform {
...
-          [this, &ctx] {
-            finalizeExecutionContext(ctx);
-            endExecution();
-          },
-          [&outerContext] {
-            detail::resetExecutionContext();
-            if (outerContext)
-              detail::setExecutionContext(outerContext);
-          });
+      detail::try_finally([this, &ctx] { finalizeExecutionContext(ctx); },
```
lmondada: This here is quite hairy. We've had to fiddle with this order of execution/clean-ups quite a bit, as every target behaves differently (and does different things on each hook).

I would be more comfortable if this change could be avoided... otherwise we'd have to test all QPUs thoroughly and check all memory deallocations very carefully. My worry is whether it is always safe to call endExecution if finalizeExecutionContext failed.

runtime/common/ExecutionContext.h
```diff
@@ -158,6 +179,9 @@ class ExecutionContext {
 
   /// @brief Slot for the detector error model, as `.dem` text.
   std::string dem_text;
+
+  /// @brief Options forwarded to the Stim ErrorAnalyzer when generating a DEM.
+  dem_options dem_opts;
```
lmondada, line 184: I think I can give the green light to this :) This would be moved into a dem-specific launch policy object, but that does not exist yet. So the ExecutionContext is the correct place for the time being. Thanks for checking with me!

runtime/common/ExecutionContext.h
lmondada: yes all good, thanks for checking!

runtime/cudaq/platform/quantum_platform.h
eliotheinrich: I see; the issue was happening because one of the new tests intentionally throws, and it seems that the StimCircuitSimulator was not properly cleaned up between tests. I moved the cleanup to the StimCircuitSimulator (see 79b13f7) ; would this work?

runtime/common/ExecutionContext.h
eliotheinrich: Sounds good; we can follow up with the new policy later. Thanks!

bmhowe23: > I think I can give the green light to this :) This would be moved into a dem-specific launch policy object, but that does not exist yet. So the ExecutionContext is the correct place for the time being. Thanks for checking with me!

@lmondada would you mind taking a look at https://github.com/NVIDIA/cuda-quantum/pull/4759 as well and let us know if the same decision applies there, too?

runtime/cudaq/platform/quantum_platform.h
lmondada: Yes, that's a reasonable fix. Thank you!

lmondada:
Happy for this to get merged as far as I'm concerned :)

khalatepradnya (approved):

mitchdz (approved):

PR #4757, opened 2026-06-22
Add realtime device_call function ID lowering test

vedika-saravanan:
## Description

This PR adds a CUDA-Q transform test that verifies realtime `cc.device_call` lowering produces function IDs using the expected hash scheme.

The test intentionally uses generic `device_call_*` symbols rather than CUDA-QX/QEC RPC names. CUDA-Q should own the compiler-side lowering/hash contract, while downstream libraries such as CUDA-QX/QEC own their specific RPC handler names.

Covered cases:
- `device_call_no_result`
- `device_call_with_result`
- `device_call_reset`

This locks down the compiler-side function-id behavior used by realtime host-dispatch consumers without coupling CUDA-Q tests to CUDA-QX/QEC naming.

## Testing

- Added `cudaq/test/Transforms/device_call_realtime_function_ids.qke`
- Covered by the existing `llvm-lit (build/cudaq/test)` suite in CI


bmhowe23:
We should get @schweitzpgi to review regarding placement for these types of tests. 

Also, while the test names in the PR are specific to CUDA-Q QEC right now, I want to confirm/ask that we have the liberty to change them in CUDA-Q QEC without updating CUDA-Q. In other words, the contract is the type of hash used, not the specific string name.

vedika-saravanan:
> We should get @schweitzpgi to review regarding placement for these types of tests.
> 
> Also, while the test names in the PR are specific to CUDA-Q QEC right now, I want to confirm/ask that we have the liberty to change them in CUDA-Q QEC without updating CUDA-Q. In other words, the contract is the type of hash used, not the specific string name.

Agreed. I updated the PR to avoid CUDA-QX/QEC-specific RPC names. The test now uses generic `device_call_*` symbols and verifies the realtime lowering/function-id hash behavior directly.

So the intended CUDA-Q contract here is the hash scheme used during realtime `device_call` lowering, not ownership of any downstream CUDA-QX/QEC RPC string names. CUDA-QX/QEC should be free to rename its handlers as long as its QPU-side declarations and host-side service stay in sync.

cudaq/test/Transforms/device_call_realtime_function_ids.qke
```diff
@@ -0,0 +1,40 @@
...
+// Copyright (c) 2026 NVIDIA Corporation & Affiliates.                        //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
+// RUN: cudaq-opt -pass-pipeline='builtin.module(canonicalize,distributed-device-call{realtime=1},cse,canonicalize)' %s | FileCheck %s
+
+// CHECK-LABEL:   func.func @noResultKernel
```
schweitzpgi: nit: We normally put the CHECK lines after the code by convention.

nit: Let's wrap any lines that clang-format will "fix" accidentally with
```c++
// clang-format off
...
// clang-format on
```
just to be on the safe side.

schweitzpgi (commented):
LGTM. A couple of nits.

cudaq/test/Transforms/device_call_realtime_function_ids.qke
vedika-saravanan: thanks Eric. It's fixed in [edc063a](https://github.com/NVIDIA/cuda-quantum/pull/4757/commits/edc063a516f8d2c96bb8fddcbefb8c8630e1f18d)

PR #4759, opened 2026-06-22
Add optional m2d and m2o sparse matrix output to dem_from_kernel

bmhowe23:
Add m2d and m2o sparse matrix outputs to cudaq.dem_from_kernel

## Summary

- **Measurement matrix output** - `dem_from_kernel` accepts a new `return_measurement_matrices=True` keyword argument. When set, the function returns a 3-tuple `(dem_text, m2d, m2o)` instead of just the DEM string.
  - `m2d` is a `scipy.sparse.csr_matrix` of shape `(num_detectors, num_measurements)`: entry `[d, m] == 1` means measurement `m` contributes to detector `d`.
  - `m2o` is a `scipy.sparse.csr_matrix` of shape `(num_observables, num_measurements)`: entry `[k, m] == 1` means measurement `m` contributes to observable `k`.
  - Both matrices are computed in a single pass over the flattened Stim circuit alongside DEM generation; no additional circuit execution is required.
  - The default (`return_measurement_matrices=False`) is unchanged - the function still returns a plain string.

## Example usage

```python
@cudaq.kernel
def kernel(n_rounds: int):
    q = cudaq.qubit()
    m = mz(q)
    for _ in range(n_rounds):
        m_new = mz(q)
        cudaq.detector(m_new, m)
        m = m_new
    cudaq.logical_observable(m)

# Plain DEM string (unchanged default)
dem_text = cudaq.dem_from_kernel(kernel, 2)

# DEM string with Stim analyzer option
dem_decomposed = cudaq.dem_from_kernel(kernel, 2, decompose_errors=True)

# DEM string + measurement matrices
dem_text, m2d, m2o = cudaq.dem_from_kernel(kernel, 2, return_measurement_matrices=True)

# All combined
dem_text, m2d, m2o = cudaq.dem_from_kernel(kernel, 2,
                                           decompose_errors=True,
                                           return_measurement_matrices=True)

### CODE REVIEW

**Review by schweitzpgi — COMMENTED** [review]
LGTM

File: python/cudaq/runtime/dem.py [hunk 1]
```diff
@@ -22,7 +22,7 @@ def _detail_check_conditionals_on_measure(kernel):
 
 
 @trace.traced
-def dem_from_kernel(kernel, *args, noise_model=None):
+def dem_from_kernel(kernel, *args, noise_model=None, return_m2d=False):
```
eliotheinrich: Can we make the new flag available in the C++ API for `dem_from_kernel` as well?

bmhowe23: Good point. Updated in 8a7841a7

python/cudaq/runtime/dem.py
```diff
@@ -49,5 +60,26 @@ def dem_from_kernel(kernel, *args, noise_model=None):
...
+        row_idx = [r for r, ms in enumerate(rows) for _ in ms]
+        col_idx = [m for ms in rows for m in ms]
+        return sp.csr_matrix(
+            (np.ones(len(row_idx), dtype=np.uint8), (row_idx, col_idx)),
+            shape=(len(rows), num_cols),
+        )
+
+    m2d = _make_csr(det_rows, num_measurements)
+    m2o = _make_csr(obs_rows, num_measurements)
+    return dem_text, m2d, m2o
```
khalatepradnya, line 118: Can you help me understand the purpose of this change? It seems to be post-processing of the result, so why does it need to be part of the API? Can't it be done ad-hoc in the application, or is it more of a convenience / common use case?

bmhowe23: It's not just a simple post-processing step; it required changes to `runtime/nvqir/stim/StimCircuitSimulator.cpp` in order to populate this data. Basically, decoders will need a way to construct detectors from regular circuit measurements, and this is the only way to do that.

This information will be passed to the decoder as a matrix called `D_sparse` (and `O_sparse` for observables). You can search for https://nvidia.github.io/cudaqx/examples_rst/qec/realtime_decoding.html to see how we do it now: it uses `qec.generate_timelike_sparse_detector_matrix`, but that is very fragile because it *assumes* that the kernel had a very specific structure, so it doesn't generalize well.

bmhowe23: Just to elaborate a bit more, cudaq.sample still returns measurements, so we need a way to convert those measurements into detectors; the m2d matrix (aka D_sparse matrix) gives us all the information we need to do that.

khalatepradnya: Thanks, Ben for the clarification!

runtime/common/ExecutionContext.h
```diff
@@ -158,6 +158,20 @@ class ExecutionContext {
...
+  /// @brief Sparse m2d/m2o matrix data; populated when `compute_m2` is true.
+  /// See `cudaq::M2DSparseMatrix` and `cudaq::M2OSparseMatrix` for the
+  /// public-facing types.
+  struct M2Data {
+    std::size_t num_measurements = 0;
+    std::vector<std::vector<std::size_t>>
+        det_rows; // det_rows[d] = measurement indices for detector d
+    std::vector<std::vector<std::size_t>>
+        obs_rows; // obs_rows[k] = measurement indices for observable k
+  } m2;
```
lmondada: Yes, the same exemption applies here too :) Would you be able to merge this into the `struct dem_options` of #4756?

bmhowe23: Yes, I'll look at doing that right now. Of course, that means #4756 would have to be merged first, which is ok, but I would ideally like to merge today if possible since our regularly scheduled weekly merge from CUDA-Q main occurs over the weekend.

bmhowe23: Mostly addressed in fb8fb1f3dc5e5892a3161a382ddac983caa56e96. However, since `dem_options` is described as a struct containing input options and is often used in a `const` manner, I decided to only move the input DEM option (named `compute_m2` above) to that struct leave the data fields (previously called `M2Data`, later commits renamed to `MeasurementMatrices`) in the regular ExecutionContext. This makes for a cleaner separation of what is truly an input configuration vs what is data being generated by the kernel. Of course, this will be modified once the policy refactor completes at a later date.

runtime/common/ExecutionContext.h
```diff
@@ -44,6 +44,10 @@ struct dem_options {
   /// Prevent the decomposer from introducing remnant edges that would otherwise
   /// be needed to satisfy the decomposition.
   bool block_decomposition_from_introducing_remnant_edges = false;
+
+  /// When true, the DEM execution also populates
+  /// `ExecutionContext::measurement_matrices`.
+  bool compute_measurement_matrices = false;
```
khalatepradnya: Q: Any reason why this field is named differently than the Python option `return_measurement_matrices`?

python/cudaq/runtime/dem.py
```diff
@@ -79,6 +93,26 @@ def dem_from_kernel(kernel, *args, noise_model=None, **dem_kwargs):
...
+
+    dem_text, num_measurements, det_rows, obs_rows = result
+
+    def _make_csr(rows, num_cols):
+        row_idx = [r for r, ms in enumerate(rows) for _ in ms]
+        col_idx = [m for ms in rows for m in ms]
+        return sp.csr_matrix(
+            (np.ones(len(row_idx), dtype=np.uint8), (row_idx, col_idx)),
+            shape=(len(rows), num_cols),
+        )
```
khalatepradnya, line 114: This came up with an agentic review so take it with a grain / pinch of salt since I am not 100% sure that it represents a "valid" use case.

`return_measurement_matrices` does not canonicalize detector/observable rows over GF(2). For a valid parity declaration like `detector(m, m)`, the measurement should cancel out, but the PR currently returns `m2d.toarray() == [[2]]` in Python and `row[0]: 0 0` in C++. The matrix contract says binary entries, so `computeM2IntoContext` should toggle duplicate measurement indices modulo 2 before storing each row, with a regression test for duplicate handles.

```
@cudaq.kernel
def duplicate_detector_kernel():
    q = cudaq.qubit()
    m = mz(q)
    cudaq.detector(m, m)
```

khalatepradnya (commented):
Overall LGTM.
One nit, one potential issue.
Also, we will have to update the docs, but that can be a follow-up.

runtime/common/ExecutionContext.h
bmhowe23: Yes, I named them differently because the Python kwarg affects the return signature, so I wanted to it to have "return" in the name, but the ExecutionContext doesn't "return" anything; rather, it is telling the simulator to just compute it and save the results in the context's measurement matrices, regardless of whether or not they are returned. I thought I'd be making it more precise that way, but I don't feel too strongly about it. If you'd like, just say the word, and I'll make them all be "return_measurement_matrices".

khalatepradnya: I think it will be better to have them aligned like all other options.

python/cudaq/runtime/dem.py
bmhowe23: Addressed in a01f53d52fb5d9a30cda4cf125483c217da9c824

runtime/common/ExecutionContext.h
bmhowe23: OK, will do. Stand by.

bmhowe23: Addressed in eef8be53ccf8c132401f1582aa8150d9a3650cfa

khalatepradnya (approved):
LGTM 👍
Thanks @bmhowe23 for all the revisions!

PR #4762, opened 2026-06-23
[qpu] Add default fallback for getCompileTarget

lmondada:
A tiny change to address feedback from @anpaz.

The new QPU APIs require a dedicated `launchKernel` overload for each supported policy. The verbosity is by design and lets QPUs opt-in to each policy, rather than defaulting to all-policies-are-supported (the previous behaviour).

This API choice was mirrored for `getCompileTarget` -- so for each policy, two overloads (one for `launchKernel` and one for `getCompileTarget`) were required. This is unnecessarily verbose (after all, an unsupported launch will not compile, so there's no need to impose the same rigidity on getCompileTarget).

I've changed `getCompileTarget` to fall back to the implementation of `other_policies`. This solution is not ideal, as the goal is eventually to remove that overload -- but good enough for the time being. We can revisit once we remove `other_policies` whether to introduce a separate fallback `getCompileTarget`


1tnguyen (approved):
LGTM 👍

PR #4764, opened 2026-06-24
Add a pass to annotate quake kernel with the number of required qubits

1tnguyen:
We have this logic for QIR lowering. For generic Quake code (in value semantics at a similar stage of lowering), we want to be able to annotate the number of qubits as well via a pass (w/o going through the QIR lowering).

Hence, pull the helper in `WireSetToProfileQIRPass` into a common location so that we can make it a standalone pass as well as using it in that QIR pass as-is.


1tnguyen:
No longer needed.

PR #4766, opened 2026-06-24
Support dumping IR between passes during specialization only

lmondada:
#4636 removed support for the `CUDAQ_PYTHON_CODEGEN_DUMP` environment variable in favour of `CUDAQ_MLIR_PRINT_EACH_PASS`. The reasoning was that
- the flag was needlessly Python specific (it is also useful for C++ JIT compilation)
- it overlapped with functionality of `CUDAQ_MLIR_PRINT_EACH_PASS`
- its name was confusing as "codegen" referred to kernel specialization, rather than codegen in the sense of lowering to target-specific output format

However, in doing so, we lost the option to dump IR between MLIR passes during specialization only.

This PR introduces support for `CUDAQ_MLIR_PRINT_EACH_PASS=specialize` (on top of the already accepted ON/OFF settings) to turn on IR dumping for that phase of compilation only. Setting this env variable reproduces the old behaviour of `CUDAQ_PYTHON_CODEGEN_DUMP=1`.

drive-by: IR printing instrumentation was often registered twice (once in `Compiler.cpp` and once in `runPassManager`), causing duplicate outputs. I've removed this duplication where it was easy.

@schweitzpgi is this a good solution for you?

### DIFF
--- a/runtime/common/Environment.cpp
+++ b/runtime/common/Environment.cpp
@@ -10,32 +10,33 @@
 #include <algorithm>
 #include <string>
 
-namespace cudaq {
+static std::string toLower(const std::string &str) {
+  std::string tmp(str);
+  std::transform(tmp.begin(), tmp.end(), tmp.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+  return tmp;
+}
+
+static bool isTruthy(const std::string &str) {
+  return str == "1" || str == "on" || str == "true" || str[0] == 'y';
+}
 
 /// @brief Helper function to get boolean environment variable
-bool getEnvBool(const char *envName, bool defaultVal = false) {
+bool cudaq::getEnvBool(const char *envName, bool defaultVal = false) {
   if (auto envVal = std::getenv(envName)) {
-    std::string tmp(envVal);
-    std::transform(tmp.begin(), tmp.end(), tmp.begin(),
-                   [](unsigned char c) { return std::tolower(c); });
-    return (tmp == "1" || tmp == "on" || tmp == "true" || tmp == "y" ||
-            tmp == "yes");
+    return isTruthy(toLower(envVal));
   }
   return defaultVal;
 }
 
-PrintEachPassMode getEnvPrintEachPassMode(const char *envName) {
+cudaq::PrintEachPassMode cudaq::getEnvPrintEachPassMode(const char *envName) {
   if (auto envVal = std::getenv(envName)) {
-    std::string tmp(envVal);
-    std::transform(tmp.begin(), tmp.end(), tmp.begin(),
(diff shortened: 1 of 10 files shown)


schweitzpgi:
This seems like a good compromise. Thanks, Luca.

runtime/common/Environment.cpp
```diff
@@ -24,4 +24,18 @@ bool getEnvBool(const char *envName, bool defaultVal = false) {
...
 }
 
+PrintEachPassMode getEnvPrintEachPassMode(const char *envName) {
+  if (auto envVal = std::getenv(envName)) {
+    std::string tmp(envVal);
+    std::transform(tmp.begin(), tmp.end(), tmp.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+    if (tmp == "specialize")
+      return PrintEachPassMode::Specialize;
+    if (tmp == "1" || tmp == "on" || tmp == "true" || tmp == "y" ||
```
schweitzpgi: maybe
```c++
tmp[0] == 'y' || tmp[0] == 't'
```
I see you `tolower`ed above, good.

runtime/common/Environment.h
```diff
@@ -13,4 +13,11 @@ namespace cudaq {
 /// @brief Helper function to get boolean environment variable
 bool getEnvBool(const char *envName, bool defaultVal);
 
+/// Valid options for `CUDAQ_MLIR_PRINT_EACH_PASS` environment variable.
+enum class PrintEachPassMode { None = 0, Specialize = 1, All = 2 };
```
schweitzpgi: nit: Maybe `All` ought to be `1` to be consistent with the previous usage. Then we can add special modes as successor values.

nit: Should `Specialize` be more detailed? Something like `ArgSynthToQIR`?

schweitzpgi (commented):
This looks great, Luca. Thanks for adding back this filter. It's very helpful for debugging.

A couple of minor nit comments.

runtime/common/Environment.cpp
lmondada: Agree this could be improved. It was copied from above. I moved it into a `isTruthy` function.

runtime/common/Environment.h
lmondada: Interesting idea. I thought about it but I would prefer to keep the current ordering of the variants, so that `<` on the integer value corresponds to increasing IR dumping. I don't think this should be too confusing.

Yes, good idea. I've renamed `Specialize` to `ArgSynthesis`. It's not specific to lowering to QIR, it could also be used for any other JIT compilation for remote QPUs.

schweitzpgi (approved):
+1

PR #4769, opened 2026-06-24
[nfc] Restore 80 column formatting.

schweitzpgi:


taalexander (approved):

taalexander:
Approving although FWIW I think 80 cols is a little too few in the era of ultrawide monitors. I prefer >=100.

schweitzpgi:
> Approving although FWIW I think 80 cols is a little too few in the era of ultrawide monitors. I prefer >=100.

https://llvm.org/docs/CodingStandards.html#source-code-width

We use this standard so we don't have to debate these sorts of things. :)

cabreraam (approved):

mitchdz (approved):

mitchdz:
For source code files maybe we can enforce this in pre-check?

PR #4770, opened 2026-06-24
[realtime] Split graph launch dispatch from the host ring loop

boschmitt:
Factor the GRAPH_LAUNCH worker machinery into a standalone engine and have the host ring loop drive it through a narrower internal interface. This separates graph offload state from transport polling so the realtime dispatcher can evolve toward unified ring and data-plane support without carrying the old monolithic host-dispatch API.


realtime/include/cudaq/realtime/daemon/dispatcher/cpu_relax.h
```diff
@@ -0,0 +1,25 @@
+/*******************************************************************************
```
1tnguyen: nit: I think we should have `*-*- C++ -*-*` annotation for `.h` files.

realtime/lib/daemon/dispatcher/host_dispatcher.cu
```diff
@@ -126,155 +101,67 @@ static void finish_slot_and_advance(const cudaq_host_dispatch_loop_ctx_t *ctx,
...
-  const size_t num_slots = ctx->config.num_slots;
+  const size_t num_slots = config->num_slots;
   uint64_t packets_dispatched = 0;
   const bool use_function_table =
-      (ctx->function_table.entries != nullptr && ctx->function_table.count > 0);
+      (table->entries != nullptr && table->count > 0);
 
-  while (as_atomic_int(ctx->shutdown_flag)
-             ->load(cuda::std::memory_order_acquire) == 0) {
+  while (*shutdown_flag == 0) {
```
1tnguyen: I think the original approach (using `atomic`) is safer. Could we keep it?

realtime/lib/daemon/dispatcher/graph_launch_engine.cu
```diff
@@ -0,0 +1,296 @@
...
+    set_status(CUDAQ_ERR_INVALID_ARG);
+    return nullptr;
+  }
+
+  // HOST_CALL-only table: no GRAPH_LAUNCH workers, so no engine.  Not an error.
+  const size_t num_workers = count_graph_launch_workers(table);
+  if (num_workers == 0) {
+    set_status(CUDAQ_OK);
+    return nullptr;
+  }
```
1tnguyen, line 201: Since we will do bit shift with `num_workers`, perhaps we should add a validation for `num_workers < 64` here as well.

1tnguyen (approved):
LGTM 👍 Thanks @boschmitt

bmhowe23:
The removal of host_dispatcher.h is a breaking change, so I'm adding that tag. If we want to keep that change, I feel like we should just advance the version number to 0.16.0 for our next release rather than trying to mark it as 0.15.1, which would imply no breaking changes.

Edit: this isn't the only breaking change we've pulled in, so I'm not saying it's just this one. It just might be the one that pushes us over the edge though.

PR #4771, opened 2026-06-25
[python] Serialize MLIR pass manager runs for async compilation

khalatepradnya:
### Summary

- Serialize Python MLIR pass-manager execution to protect shared Python MLIR contexts during async compilation.
- Add a regression test that concurrently compiles a Python kernel from multiple host threads.
- Prevents `MLIRContext::appendDialectRegistry` assertions seen in `sample_async` / mqpu `observe` paths.

### Root Cause

Python async workers clone modules but share the originating `MLIRContext`. Since compilation releases the GIL, multiple worker threads can enter `PassManager::run()` concurrently. MLIR may append dependent dialect registries during pass-manager setup, which is not allowed while the same context is in a multi-threaded execution region.


1tnguyen (approved):
LGTM 👍

mitchdz (approved):

PR #4777, opened 2026-06-25
[macOS] Surface simulator kernel exceptions across the macOS JIT boundary

mitchdz:
On macOS arm64, a C++ exception thrown by the simulator inside a
JIT-compiled kernel frame cannot be unwound by the system unwinder and
calls `std::terminate`, crashing the process instead of raising a catchable
error (e.g. Stim rejecting a non-Clifford gate, or dem_from_kernel on a
non-Clifford kernel).

This defers such exceptions onto the ExecutionContext while a kernel frame
is executing and re-throws them from the C++ launcher frame above the JIT
boundary, so callers observe the same exception as on platforms where
direct unwinding works. The deferral is scoped to the kernel frame via an
RAII guard (`ExecutionContext::inKernelLaunch`); outside it - e.g. gate
application during sample/observe finalization - exceptions throw directly,
preserving existing behavior on all platforms.

This was merged into the release branch for 0.15 to unblock the publishing piepline, which is fully green now - https://github.com/NVIDIA/cuda-quantum/actions/runs/28160514270

### DIFF
--- a/.github/pre-commit/spelling_allowlist_cxx.txt
+++ b/.github/pre-commit/spelling_allowlist_cxx.txt
@@ -3,3 +3,4 @@ gcc
 nullary
 pre
 typedef
+unwinder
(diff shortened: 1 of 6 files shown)


1tnguyen (approved):
LGTM 👍

PR #4779, opened 2026-06-25
[ci] read-only ccache on PRs, larger dev-env runners

mitchdz:
PR runs spend ~7 min extracting and pushing the ccache tarball through the build context and GHCR. Skip those steps on PRs (push to pull-request/*); only the merge queue writes the cache now. Reads are unchanged and the cache stays warm from merge-queue runs.

Also move the Debug and Python dev-env jobs to cpu16 runners.


taalexander (approved):

1tnguyen (approved):
LGTM 👍

PR #4782, opened 2026-06-25
[ci] add multi-GPU MPI validation on 2-GPU runners

mitchdz:
Publishing only exercised nvidia-mgpu/MPI on a single GPU (mpiexec -np 4, oversubscribed).

Add a job that runs the distributed examples with -np 2 (one rank per GPU) on real 2-GPU runners (nv-gpu-amd64-rtxpro6000-2gpu / nv-gpu-arm64-l4-2gpu), using the published multi-arch image.

It checks that two GPUs are visible and gates the release via clean_up.


1tnguyen (approved):
💯

PR #4783, opened 2026-06-25
realtime: self-relaunching device-graph scheduler for unbounded device-side dispatch

cketcham2333:
## Summary
Extends the existing device-side graph-launch dispatch path so a single
persistent scheduler graph can trigger an **unbounded** number of follow-up
graphs, bypassing CUDA's limit of **120 fire-and-forget graph launches per
parent graph**.
A `DEVICE_CALL` handler can now return a sentinel asking the dispatcher to fire
a configured follow-up graph; the scheduler then tail-relaunches itself, which
resets the per-parent launch budget every cycle. The mechanism is
workload-agnostic — any consumer that needs to trigger one follow-up graph per
incoming request over a long-lived session benefits.
## Background / problem
The device-graph dispatch path already lets a `DEVICE_CALL` handler launch child
graphs from device code. However, a parent graph may issue at most 120
fire-and-forget device-side launches in its lifetime. For a long-running session
that triggers one follow-up graph per request, that caps the total number of
triggered launches a single dispatch graph can ever drive.
## What this PR does
- **`cudaq_realtime.h`**
  - Add the `CUDAQ_DISPATCH_STATUS_TRIGGER_GRAPH` status sentinel a `DEVICE_CALL`
    handler can return to request a follow-up graph launch.
  - Add a `triggered_graph_exec` parameter to
    `cudaq_create_dispatch_graph_regular` (the graph to fire on that sentinel).
  - Mark the three graph-dispatch entry points `CUDAQ_REALTIME_DISPATCH_API`
    (default visibility) so a consumer `.so` can resolve them after the dispatch
    archive is absorbed.
  - Widen the graph-API `#if` guard to include `CUDART_VERSION` so host TUs that
    include only `<cuda_runtime.h>` still see the API.
- **`dispatch_kernel.cu`**
  - When a handler returns `CUDAQ_DISPATCH_STATUS_TRIGGER_GRAPH`: fire
    `triggered_graph_exec` fire-and-forget, ACK the request normally, break the
    drain loop, and tail-self-relaunch via `cudaGetCurrentGraphExec()` +
    `cudaStreamGraphTailLaunch`. The tail launch is ordered **after** the
    triggered graph, so a subsequent request only runs once the triggered graph
    is done (≤1 triggered graph outstanding; no race).
  - Persist the RX ring cursor across relaunches in a `__device__` global
    (`g_graph_dispatch_cursor`): load at kernel entry, store (release fence)
    before the tail relaunch, reset in `cudaq_create_dispatch_graph_regular`. A
    relaunch resumes at the next slot instead of rescanning from slot 0 (strict
    FIFO consumption).
- **`test_dispatch_kernel.cu`**
  - New `DeviceGraphSelfRelaunchTest.Exceeds120`: drives 200 triggered launches
    (> 120) to prove the budget actually resets across self-relaunch.
  - Update the existing `cudaq_create_dispatch_graph_regular` call for the new
    signature.
## Backward compatibility
Opt-in and additive. `triggered_graph_exec` defaults to "none"; a graph created
without it behaves exactly as before, and handlers that never return
`TRIGGER_GRAPH` are unaffected. No changes to existing dispatch semantics.
## Testing
- New `DeviceGraphSelfRelaunchTest.Exceeds120` passes (200 > 120 triggered
  launches, no launch-cap error/hang).
- Existing `test_dispatch_kernel` suite still passes (no regression).
- Exercised end-to-end by a downstream CUDA-QX realtime consumer that triggers a
  follow-up graph per request over the Hololink RoCE ring: well over 120
  triggered launches in a single session, validated on a real FPGA and an
  emulator.
## Notes for reviewers
The symbol-visibility (`CUDAQ_REALTIME_DISPATCH_API`) and header-guard
(`CUDART_VERSION`) tweaks exist so a downstream host-compiled translation unit
can link these graph entry points after absorbing the dispatch static archive;
they have no effect on device-side or in-tree callers.


realtime/lib/daemon/dispatcher/dispatch_kernel.cu
```diff
@@ -391,10 +404,22 @@ __global__ void dispatch_kernel_with_graph(
...
-  std::size_t current_slot = 0;
+  // Resume the RX-ring cursor from the persistent global so a tail
+  // self-relaunch continues at the next slot instead of rescanning from 0.
+  // Reset to 0 by cudaq_create_dispatch_graph_regular for the first launch.
+  std::size_t current_slot = g_graph_dispatch_cursor;
+
+  // Set when a DEVICE_CALL handler returns CUDAQ_DISPATCH_STATUS_TRIGGER_GRAPH:
+  // we fired triggered_graph_exec fire-and-forget and must tail-self-relaunch
+  // (which resets the device-graph fire-and-forget budget) before continuing.
+  __shared__ bool s_relaunch;
```
1tnguyen, line 419: Since we're using `__shared__` memory here, perhaps we should add a check that `num_blocks == 1`  in `cudaq_create_dispatch_graph_regular` if `triggered_graph_exec` is provided.

1tnguyen (approved):
LGTM 👍 

Just have one validation suggestion.

realtime/lib/daemon/dispatcher/dispatch_kernel.cu
cketcham2333: Done.  Added the check + a negative test in commit SHA 5a1689f554e483e85f8939eeb518bd20c3c73bd9

bmhowe23:
I tagged this is a breaking change since it broke the CUDA-Q QEC build.

PR #4784, opened 2026-06-26
Implement conditional loop unrolling for value-semantics pipelines

1tnguyen:
Adds an opt-in `unroll-only-wire-blocking-loops` option to `cc-loop-unroll` (default `false` so existing pipelines unchanged). 

This allows a value-semantics / `wireset` lowering pipeline unroll only the loops affecting qubit (quantum data type) indexing.
A loop touches quantum data by a dynamic index or slice needs unrolling to make qubit index static (can be translated to a static `wire`). Everything else: loops over classical data (including measurement data), purely repeating outer loops, etc., is left rolled. 

Majority of the changeset is for test coverage:

(1) Low-level quake tests for basic checks of the new unroll mode.

(2) Semi-integrated lit tests to emulate the JIT pipeline on various loops: nested, transitive, etc.

(3) Fully integrated tests with the fake quake target: mimic a backend that expects value-semantics IR (verify the incoming IR), also checking the expected number of loops in the IR.

### DIFF
--- a/cudaq/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
+++ b/cudaq/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
@@ -29,16 +29,111 @@ unrollLoopByValue(cudaq::cc::LoopOp loop,
   return *resultOpt;
 }
 
-static bool exceedsThresholdValue(cudaq::cc::LoopOp loop,
-                                  std::size_t threshold) {
+static bool exceedsThresholdValue(cudaq::cc::LoopOp loop, std::size_t threshold,
+                                  bool emitWarning = true) {
   auto components = cudaq::opt::getLoopComponents(loop);
   if (components->hasAlwaysTrueCondition()) {
-    loop->emitWarning("Loop condition is always true. This loop is not "
-                      "supported in a kernel.");
+    if (emitWarning)
+      loop->emitWarning("Loop condition is always true. This loop is not "
+                        "supported in a kernel.");
     return true;
   }
   auto upperBound = unrollLoopByValue(loop, *components);
   return upperBound >= threshold;
+}
+
+static bool loopCanBeUnrolled(cudaq::cc::LoopOp loop, bool allowBreak,
+                              std::size_t threshold,
+                              bool emitWarning = true) {
+  if (!allowBreak && !cudaq::opt::isaCountedLoop(loop))
+    return false;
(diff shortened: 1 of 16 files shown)


cudaq/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
```diff
@@ -29,18 +29,113 @@ unrollLoopByValue(cudaq::cc::LoopOp loop,
...
+/// (non-constant) index or slice and therefore blocks conversion to value
+/// (wire) semantics: it contains a `quake.extract_ref` with a dynamic index or
+/// a `quake.subveq` with a dynamic bound in its own body. The walk does NOT
+/// descend into nested `cc.loop` ops -- a nested loop is classified (and
+/// unrolled) on its own, and once it is unrolled its accesses become
+/// constant-index, at which point this loop no longer blocks. This lets a
+/// purely-repeating outer loop stay rolled while inner index-dependent loops
+/// are unrolled. A loop with only constant-index references/slices (or only
+/// `!quake.wire` values) does not block: mem2reg can thread those qubits as
+/// loop iter-args.
```
atgeller, line 93: I'm a little confused about how a loop such as
```
for (int i = 0; i < N; i++)
  for (int j = 0; j < M; j++)
    x(qubits[i*M + j]);
```
gets handled here. Is it that the inner loop is first unrolled, and then the outer loop is later unrolled by the same pass since the dependency is still detected?

atgeller (commented):
Thanks @1tnguyen! I just have one question about the forms of loops handled by these changes.

cudaq/lib/Optimizer/Transforms/LoopUnrollPatterns.inc
1tnguyen: Indeed, I believe the above theory is correct (`the inner loop is first unrolled, and then the outer loop is later unrolled by the same pass`). This is handled by this loop:

https://github.com/NVIDIA/cuda-quantum/blob/3344a8a997b93bba04e89a524386cc45c0cb3176/cudaq/lib/Optimizer/Transforms/LoopUnroll.cpp#L53-L60

I also added the above test case ([b712f4d](https://github.com/NVIDIA/cuda-quantum/pull/4784/commits/b712f4dc97d8cb1f514ad990ac43729fe18f3f20)) as well.

atgeller (approved):
LGTM 👍

PR #4785, opened 2026-06-26
Accept np.floor in Python kernels

### LINKED ISSUE #4780: np.floor is not accepted in Python kernels although np.ceil works and C++ lowers std::floor
Hi CUDA-Q team,

Thank you for landing the kernel math-function work in #4680. I checked current `main` after #2942 was closed, and the accepted continuous set looks healthy from what I can see. In particular, I verified the C++ path executes `asin`, `acos`, `atan`, `tan`, `log`, `sqrt`, and `exp` end-to-end on `qpp-cpu` with the expected deterministic results.

This report comes from my local work on an unsubmitted PR attempt for #2942 (mathematical-function support in kernels) during UnitaryHack 2026. I was not able to submit that PR because of the UnitaryHack 2026 per-participant PR limit, so I continued checking the current `main` locally and found one small remaining Python/C++ coverage gap that seems worth tracking: `np.floor` is still not accepted in Python kernels, even though `np.ceil` works and the C++ frontend already lowers `std::floor`.

This does not look like a regression from #4680. It appears to be a pre-existing Python AST-bridge coverage gap that became more visible now that the surrounding math support has been filled in.

## Environment Checked

- `NVIDIA/cuda-quantum` `main` at `0d318de58ce3db090eed1d514f3bed20a08c51df`, fetched on 2026-06-25.
- Math-function support from #4680 is present at `f40d19c`.
- C++ execution checks were run on `qpp-cpu` with a local source build whose `cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp` math-lowering code has no relevant diff from `origin/main`.

## Source-Level Mismatch

On current `main`, the Python AST bridge accepts `ceil` but not `floor`:

```bash
git fetch origin main
git show origin/main:python/cudaq/kernel/ast_bridge.py | sed -n '857,862p'
git show origin/main:python/cudaq/kernel/ast_bridge.py | sed -n '3692,3711p'
```

The supported NumPy list contains:

```python
'sin', 'cos', 'tan', 'asin', 'acos', 'atan', 'arcsin', 'arccos',
'arctan', 'sqrt', 'ceil', 'exp', 'log'
```

and the dispatch block has a `ceil` handler followed by:

```python
self.emitFatalError(
    f"unsupported NumPy call ({node.func.attr})", node)
```

There is no corresponding `floor` handler.

The C++ frontend does lower both `ceil` and `floor`:

```bash
git show origin/main:cudaq/lib/Frontend/nvqpp/ConvertExpr.cpp | sed -n '1505,1514p'
```

which shows the adjacent `std::ceil` and `std::floor` handling through `math::CeilOp` and `math::FloorOp`.

## Python Reproducer

With a Python build from current `main`:

```python
# floor_demo.py
import numpy as np
import cudaq

cudaq.set_target("qpp-cpu")

@cudaq.kernel
def uses_ceil() -> float:
    return np.ceil(2.1)

@cudaq.kernel
def uses_floor() -> float:
    return np.floor(2.7)

print("ceil :", uses_ceil())
print("floor:", uses_floor())
```

Run:

```bash
python floor_demo.py
```

Expected, for parity with `np.ceil` and the C++ frontend:

```text
ceil : 3.0
floor: 2.0
```

Actual behavior from the current `ast_bridge.py` dispatch path:

```text
ceil : 3.0
...
cudaq.kernel.ast_bridge.CompilerError: floor_demo.py:<line>: error: unsupported NumPy call (floor)
     (offending source -> np.floor(2.7))
```

## C++ Contrast

The same operation works through the C++ frontend today. This uses a runtime argument so the call is lowered and executed, not just constant-folded by Clang:

```cpp
// floor.cpp
#include <cudaq.h>
#include <cmath>

struct floor_k {
  void operator()(double x) __qpu__ {
    cudaq::qubit q;
    ry(M_PI * std::floor(x), q);
    mz(q);
  }
};

int main() {
  cudaq::sample(floor_k{}, 1.0).dump();
}
```

Run:

```bash
nvq++ -std=c++20 floor.cpp -o floor
./floor
```

Expected and observed:

```text
{ 1:1000 }
```

This indicates `std::floor` lowers and executes correctly on the simulator.

## Minor Related Consistency Notes

While checking this, I also noticed two related C++-only functions:

- `std::abs` is lowered in C++, but Python kernels reject both `np.abs(...)` (`unsupported NumPy call (abs)`) and builtin `abs(...)` (`unhandled function call - abs`).
- `std::round` is lowered in C++, but Python kernels reject both `np.round(...)` and builtin `round(...)`. This one may need an explicit semantics decision because Python/NumPy and C++ differ on half-way cases such as `2.5`, so I would treat it as a separate, lower-priority consistency question rather than the same straightforward parity issue as `floor`.

I am listing these only because they share the same C++/Python coverage shape. The main issue here is `np.floor`, which has no obvious semantic ambiguity, pairs naturally with the already-supported `np.ceil`, appears in the original #2942 checklist, and is already implemented on the C++ side.

## Summary

- The merged continuous math-function set from #4680 looks good from these checks.
- `np.floor` remains unsupported in Python kernels on current `main`.
- C++ already lowers and executes `std::floor`, and Python already accepts `np.ceil`.
- `abs` and `round` have similar C++-only coverage, but `round` may need a separate semantics decision.

Thanks again for the work on the math-function support.

aryanputta:
Fixes #4780.

## Problem
The Python AST bridge accepts `np.ceil` but not `np.floor`, even though the C++ frontend already lowers `std::floor`. Calling `np.floor` inside a `@cudaq.kernel` hits the `unsupported NumPy call` fatal error. This is a pre-existing Python/C++ coverage gap (not a regression from the math-function work in #4680), surfaced in #4780.

## Fix
Mirror the existing `ceil` path exactly in `python/cudaq/kernel/ast_bridge.py`:
- add `floor` to `__isSupportedNumpyFunction`;
- add a `floor` dispatch handler that rejects complex inputs (same as `ceil`) and emits `math.FloorOp`.

`math.FloorOp` is the MLIR Math dialect analog of `math.CeilOp` (both are defined in the dialect), so this needs no C++ changes.

## Tests
Extended `python/tests/kernel/test_kernel_float.py` with `np.floor` cases for both `np.float64` and `np.float32`, mirroring the existing `np.ceil` checks.

## Validation
- `python -m py_compile` passes on both changed files.
- Confirmed `FloorOp` and `CeilOp` are both defined in the MLIR Math dialect (`mlir/Dialect/Math/IR/MathOps.h.inc`), so `math.FloorOp` resolves the same way `math.CeilOp` does.
- Full kernel execution requires a source build; the change is a line-for-line mirror of the verified `ceil` path, so CI on this PR exercises the new tests end to end.

### DIFF
--- a/python/cudaq/kernel/ast_bridge.py
+++ b/python/cudaq/kernel/ast_bridge.py
@@ -857,7 +857,7 @@ class PyASTBridge(ast.NodeVisitor):
     def __isSupportedNumpyFunction(self, id):
         return id in [
             'sin', 'cos', 'tan', 'asin', 'acos', 'atan', 'arcsin', 'arccos',
-            'arctan', 'sqrt', 'ceil', 'exp', 'log'
+            'arctan', 'sqrt', 'ceil', 'floor', 'exp', 'log'
         ]
 
     def __isSupportedVectorFunction(self, id):
@@ -3704,6 +3704,14 @@ class PyASTBridge(ast.NodeVisitor):
                                 f"supported for complex numbers", node)
                             return
                         self.pushValue(math.CeilOp(value).result)
+                        return
+                    if node.func.attr == 'floor':
+                        if ComplexType.isinstance(value.type):
+                            self.emitFatalError(
+                                f"numpy call ({node.func.attr}) is not "
+                                f"supported for complex numbers", node)
+                            return
+                        self.pushValue(math.FloorOp(value).result)
                         return
 
                     self.emitFatalError(
--- a/python/tests/kernel/test_kernel_float.py
+++ b/python/tests/kernel/test_kernel_float.py
@@ -183,6 +183,14 @@ def test_float64_use():
     t = np.ceil(np.float64(np.pi / 2 + 1))
     assert is_close(t, float_np_use())
 
+    # Use a float inside np in a kernel (floor)
+    @cudaq.kernel
+    def float_np_use() -> np.float64:
+        return np.floor(np.float64(np.pi / 2 + 1))
+
+    t = np.floor(np.float64(np.pi / 2 + 1))
+    assert is_close(t, float_np_use())
+
     # Use a float inside np in a kernel (exp)
     @cudaq.kernel
     def float_np_use() -> np.float64:
@@ -310,6 +318,14 @@ def test_float32_use():
         return np.ceil(np.float32(np.pi / 2 + 1))
 
     t = np.ceil(np.float32(np.pi / 2 + 1))
+    assert is_close(t, float_np_use())
+
+    # Use a float inside np in a kernel (floor)
+    @cudaq.kernel
+    def float_np_use() -> np.float32:
+        return np.floor(np.float32(np.pi / 2 + 1))
+
+    t = np.floor(np.float32(np.pi / 2 + 1))
     assert is_close(t, float_np_use())
 
     # Use a float inside np in a kernel (exp)


schweitzpgi (commented):
LGTM

schweitzpgi:
> /ok to test 243d65e

schweitzpgi (approved):

schweitzpgi:
> /ok to test 5b2e591

PR #4786, opened 2026-06-26
Bump openssl, zlib, and curl to patched versions

mitchdz:
Update the vendored dependency versions pinned in the prerequisite build scripts to pull in upstream security fixes:

  - openssl 3.6.2 -> 3.6.3
  - zlib    1.3.1 -> 1.3.2
  - curl    8.20.0 -> 8.21.0

Both install_prerequisites.sh and bootstrap_prerequisites.sh define these pins; the tarball URLs are derived from the version variables, so no other changes are required.


1tnguyen (approved):
👍

PR #4787, opened 2026-06-26
[braket] Fix the basis gate set for decomposition so that unsupported gates such as `sdg` and `tdg` are not emitted

khalatepradnya:
### Summary

* Follow-up to PR https://github.com/NVIDIA/cuda-quantum/pull/4211

* Fix a `braket` regression where kernels using adjoint S/T gates, or controlled gates that decompose through adjoint S/T, could emit `sdg` / `tdg` into the OpenQASM payload. The Braket service rejects those gate names for the default simulator path.

```
# test_adjoint_modifier: kernel uses s.adj(q) and t.adj(q)  -> sdg / tdg
>   cudaq.sample(single_qubit_gates, shots_count=100).dump()
E   RuntimeError: [line 11] uses a gate: sdg which is either not supported by the device,
    not defined via a defcal, or not used in the right scope
    (e.g., when a device does not support using native gates outside of verbatim)

# test_control_modifier: kernel uses s.ctrl / t.ctrl  -> tdg
E   RuntimeError: [line 10] uses a gate: tdg which is either not supported by the device, ...
```

* This removes native `s,t` from the Braket decomposition basis so CUDA-Q lowers S/T and S†/T† through supported rotation gates. 

* It also removes the retired TN1 simulator from the live Braket backend test matrix since it reports to be `RETIRED`.
```
RuntimeError: Device is not available because it is retired. Please try another device.
```

### DIFF
--- a/cudaq/test/Translate/OpenQASM/braket-adjoint-s-t-basis.qke
+++ b/cudaq/test/Translate/OpenQASM/braket-adjoint-s-t-basis.qke
@@ -1,0 +1,29 @@
+// ========================================================================== //
+// Copyright (c) 2026 NVIDIA Corporation & Affiliates.                        //
+// All rights reserved.                                                       //
+//                                                                            //
+// This source code and the accompanying materials are made available under   //
+// the terms of the Apache License 2.0 which accompanies this distribution.   //
+// ========================================================================== //
+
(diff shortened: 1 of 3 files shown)


mitchdz (approved):

lmondada:
Thank you @khalatepradnya!!

PR #4792, opened 2026-06-29
Format test dirs and sync run_clang_format.sh exclusions with .pre-commit-config.yaml

bmhowe23:
Read the excluded directories dynamically from the clang-format hook in .pre-commit-config.yaml so the script and pre-commit stay in sync automatically. Also adds *.hpp to the file pattern list.

Additionally, it formats the cudaq/test directory and targetttests directory.


scripts/run_clang_format.sh
```diff
@@ -35,7 +35,16 @@ clang_format_executable=${clang_format_executable:-clang-format}
...
+# Exclusions are read dynamically from the clang-format hook in .pre-commit-config.yaml
+mapfile -t excludes < <(python3 -c "
+import re
+text = open('.pre-commit-config.yaml').read()
+m = re.search(r\"id: clang-format.*?exclude: '\\^\\((.+?)\\)/'\", text, re.DOTALL)
+if m:
+    for d in m.group(1).split('|'):
+        print(':!:' + d + '/*')
+")
+git ls-files -- '*.cpp' '*.h' '*.hpp' "${excludes[@]}" | xargs $clang_format_executable -i
```
schweitzpgi, line 47: For consistency with [the LLVM coding conventions](https://llvm.org/docs/CodingStandards.html), we shouldn't have `.hpp` files in our code base.

schweitzpgi (commented):
Skimmed over this and LGTM.

scripts/run_clang_format.sh
bmhowe23: It looks like `realtime/include/cudaq/realtime/cpu_transport/roce_transceiver.hpp` is one file that has snuck through. (It is part of cudaq-relatime, not the main CUDA-Q.)

The [.pre-commit-config.yaml file](https://github.com/NVIDIA/cuda-quantum/blob/4410ea29e90766220c34ae8d7015287a175d409e/.pre-commit-config.yaml#L37) references hpp, so I'm just trying to be consistent with it.

schweitzpgi: Right. Let's just file a defect against the .hpp file. It's not really anything with this PR per se.

schweitzpgi (approved):

PR #4793, opened 2026-06-29
[Python] Add validations for kernel decorator's `from_json`

1tnguyen:
This would produce more targeted error messages rather than waiting for latent errors, e.g., invalid data access. 

Fixes: nvbugs/6382827


bettinaheim (approved):

sacpis (approved):
LGTM. Thanks @1tnguyen.

PR #4796, opened 2026-06-29
Fixing heap OOB write in AnyonServerHelper::processResults

sacpis:
`AnyonServerHelper::processResults` sizes the bitstrings vector from the first register's shot count. Then writes every register's result vector into it with `bitstrings[i++] += bit`. Here there is no check that later registers are not longer. A response from Anyon's server with a short first register and a longer later register drives the index go past the end.

The helper (extracted it with the purpose of finding similar issues in other providers response) now throws a runtime error when a register's length differs from the shot count.


runtime/cudaq/platform/default/rest/helpers/anyon/AnyonHelper.h
```diff
@@ -0,0 +1,40 @@
+/*******************************************************************************
```
1tnguyen: nit: `-*- C++ -*-*`

1tnguyen (approved):
LGTM 👍

unittests/nvqpp/backends/anyon/AnyonUtilTester.cpp
```diff
@@ -0,0 +1,39 @@
...
+  EXPECT_EQ(bitstrings[0], "01");
+  EXPECT_EQ(bitstrings[1], "11");
+  EXPECT_EQ(bitstrings[2], "00");
+}
+
+CUDAQ_TEST(AnyonUtilTester, checkRejectsUnequalRegisterLengths) {
+  nlohmann::json results;
+  results["r0"] = std::vector<std::string>{"0"};
+  results["r1"] = std::vector<std::string>(64, "1");
+  EXPECT_THROW(
```
khalatepradnya: `EXPECT_ANY_THROW` for the installer tests?

sacpis: Fixed in 59d7009.

runtime/cudaq/platform/default/rest/helpers/anyon/AnyonHelper.h
sacpis: Fixed in fbc19ed.

PR #4797, opened 2026-06-29
Add validations to state accessor API

1tnguyen:
- Add index bound checks for `state` data accessor API.

- For photonics, as it supports qudits, add a number of levels check as well (basis state exceeds the qudit dimension).

- Add unit tests catching those invalid cases that we should handle.

### DIFF
--- a/python/runtime/cudaq/algorithms/py_state.cpp
+++ b/python/runtime/cudaq/algorithms/py_state.cpp
@@ -16,6 +16,7 @@
 #include "cudaq/algorithms/get_state.h"
 #include "cudaq/runtime/logger/logger.h"
 #include "mlir/Bindings/Python/NanobindAdaptors.h"
+#include <limits>
 #include <nanobind/ndarray.h>
 
 using namespace cudaq;
@@ -649,8 +650,18 @@ void cudaq::bindPyState(nanobind::module_ &mod, LinkedLibraryHolder &holder) {
           "__getitem__",
           [](state &s, int idx) {
             // Support Pythonic negative index
-            if (idx < 0)
+            if (idx < 0) {
+              // Reject negative indices when the existing signed-int shift
+              // cannot represent the logical qubit dimension safely.
+              if (s.get_num_qubits() >=
+                  static_cast<std::size_t>(std::numeric_limits<int>::digits))
+                throw std::out_of_range("State index out of bounds.");
               idx += (1 << s.get_num_qubits());
+              // An index that remains negative was below the Pythonic lower
+              // bound and must not be converted to an unsigned C++ index.
+              if (idx < 0)
+                throw std::out_of_range("State index out of bounds.");
+            }
             return s[idx];
           },
           R"#(Return the `index`-th element of the state vector.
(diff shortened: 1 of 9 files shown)


unittests/nvqpp/integration/get_state_tester.cpp
```diff
@@ -47,6 +56,12 @@ CUDAQ_TEST(GetStateTester, checkSimple) {
   EXPECT_NEAR(0.0, state.amplitude({0, 1}).real(), 1e-3);
   EXPECT_NEAR(1. / std::sqrt(2.), state.amplitude({1, 1}).real(), 1e-3);
 
+  // state[4]: index equals logical dimension.
+  EXPECT_THROW(state[4], std::out_of_range);
```
khalatepradnya: `EXPECT_ANY_THROW` for the installer tests?

1tnguyen: Good point. Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4797/changes/d90a39112f8a837d924d234c432fbbfce4fdd99c

runtime/cudaq/qis/managers/photonics/PhotonicsExecutionManager.cpp
```diff
@@ -47,6 +47,13 @@ struct PhotonicsState : public cudaq::SimulationState {
           "[photonics] getAmplitude with an invalid number of bits in the "
           "basis state: expected {}, provided {}.",
           getNumQubits(), basisState.size()));
+    // Each basis-state digit indexes one qudit and must be within that qudit's
+    // number of levels before it is converted to a flat state-vector index.
+    if (std::any_of(basisState.begin(), basisState.end(), [&](int digit) {
```
sacpis, line 53: For `std::any_of`, should we include `<algorithm>` in this file? Seems like it is relying on transitive includes, which is fragile.

python/runtime/cudaq/algorithms/py_state.cpp
```diff
@@ -649,8 +650,18 @@ void cudaq::bindPyState(nanobind::module_ &mod, LinkedLibraryHolder &holder) {
...
+              // cannot represent the logical qubit dimension safely.
+              if (s.get_num_qubits() >=
+                  static_cast<std::size_t>(std::numeric_limits<int>::digits))
+                throw std::out_of_range("State index out of bounds.");
               idx += (1 << s.get_num_qubits());
+              // An index that remains negative was below the Pythonic lower
+              // bound and must not be converted to an unsigned C++ index.
+              if (idx < 0)
+                throw std::out_of_range("State index out of bounds.");
+            }
```
sacpis: Seems like most of the code here is repeated in line 686-697? Should we create a lambda instead?

sacpis (commented):
LGTM. Thanks @1tnguyen.

runtime/cudaq/qis/managers/photonics/PhotonicsExecutionManager.cpp
1tnguyen: Fixed in https://github.com/NVIDIA/cuda-quantum/pull/4797/changes/64a22b854d9bcfc343aedfc0b3085fe4d2d5f00f

python/runtime/cudaq/algorithms/py_state.cpp
1tnguyen: Good point! Refactor it into a lambda in  https://github.com/NVIDIA/cuda-quantum/pull/4797/changes/64a22b854d9bcfc343aedfc0b3085fe4d2d5f00f

mitchdz (approved):

PR #4798, opened 2026-06-29
Fix amplitude encoding `contrib` module to be compatible with GPU simulators

1tnguyen:
Fixes to make the `encoding` C++ module compatible with GPU simulators, which support single-precision/multiple tensors.

- The detection of precision via `get_runtime_target` doesn't work with non-`nvq++` compiled code, including some unit tests.

- Use whole `state` data to determine the expected size rather than tensor 0, as there might be multiple tensors in the tensor network states.


runtime/cudaq/algorithms/encoding.cpp
```diff
@@ -58,8 +61,7 @@ prepareAmplitudeVector(std::span<const std::complex<double>> data,
 }
 
 std::vector<std::complex<double>> stateToAmplitudeVector(const state &data) {
-  const auto tensor = data.get_tensor(0);
-  const std::size_t numElements = tensor.get_num_elements();
+  const std::size_t numElements = 1ULL << data.get_num_qubits();
   if (numElements == 0)
     throw std::invalid_argument("amplitude_encode: input must be non-empty.");
```
sacpis: Seems like this is unreachable now. As 1ULL < n is never 0 for any realistic qubit count.

sacpis (approved):
Overall LGTM. Thanks @1tnguyen.

runtime/cudaq/algorithms/encoding.cpp
1tnguyen: Good catch! Remove in https://github.com/NVIDIA/cuda-quantum/pull/4798/changes/d67d527f2bc122799eeba7e0a76659fa948baba8

schweitzpgi (approved):

PR #4802, opened 2026-06-30
[quera] Fix heap out-of-bounds read in QuEraServerHelper::processResults

mitchdz:
processResults converts QuEra-on-Braket measurement results by iterating over preSequence while indexing postSequence at the same position:

    for (size_t i = 0; i < pre.size(); ++i)
      state_idx[i] = pre[i] * (1 + post[i]);

preSequence and postSequence are deserialized as two independent vectors from the cloud response (S3 results.json), and the loop never checks that they have the same length. A response whose postSequence is shorter than preSequence reads past the end of the postSequence heap buffer. Because this data crosses an untrusted boundary, a malformed or truncated server payload can trigger an out-of-bounds read and crash the client.

Add a length-equality guard before the conversion and throw a descriptive runtime_error when the two sequences differ in size. This is consistent with how the surrounding code already rejects malformed results (the adjacent .value() calls throw on missing fields) and is a no-op for well-formed responses, where preSequence and postSequence always match.

Add a regression test that drives the real processResults with a mismatched payload and confirms it throws rather than reading out of bounds. The QuEra unit test now links cudaq-serverhelper-quera so it can call the helper directly.


1tnguyen (approved):
LGTM 👍

schweitzpgi (approved):

unittests/nvqpp/backends/quera/JsonPayloadTester.cpp
```diff
@@ -461,3 +462,16 @@ CUDAQ_TEST(QueraTester, checkProcessResult) {
...
+// than preSequence must not cause a heap out-of-bounds read. processResults
+// should reject it with an exception instead.
+CUDAQ_TEST(QueraTester, checkProcessResultLengthMismatch) {
+  const std::string mismatched = R"(
+    {"measurements":[{"shotMetadata":{"shotStatus":"Success"},
+     "shotResult":{"preSequence":[0,0],"postSequence":[0]}}]})";
+  cudaq::ServerMessage resultsJson = nlohmann::json::parse(mismatched);
+  cudaq::QuEraServerHelper helper;
+  std::string jobId = "";
+  EXPECT_THROW(helper.processResults(resultsJson, jobId), std::runtime_error);
```
khalatepradnya:
```suggestion
  EXPECT_ANY_THROW(helper.processResults(resultsJson, jobId));
```

PR #4804, opened 2026-06-30
Capping gzipDecompress output

sacpis:
`cudaq::qio::gzipDecompress` inflated an untrusted zlib stream into an ever-growing `std::string` with no cap on output size.

Adding a maxOutputSize parameter to gzipDecompress (default 512 MB) and a check inside the inflate loop that throws std::runtime_error before the output exceeds the cap.

### DIFF
--- a/runtime/cudaq/platform/default/rest/helpers/scaleway/qio/Compression.h
+++ b/runtime/cudaq/platform/default/rest/helpers/scaleway/qio/Compression.h
@@ -15,7 +15,7 @@ inline constexpr std::size_t defaultMaxDecompressedSize = 512UL * 1024 * 1024;
 
 std::string gzipCompress(const std::string &input);
 
-/// Decompress a zlib stream, which produces `maxOutputSize` bytes.
+/// Decompress a `zlib` stream, which produces `maxOutputSize` bytes.
 std::string
 gzipDecompress(const std::string &input,
                std::size_t maxOutputSize = defaultMaxDecompressedSize);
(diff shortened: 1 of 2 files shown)


runtime/cudaq/platform/default/rest/helpers/scaleway/qio/Compression.h
```diff
@@ -6,9 +6,17 @@
...
 #include <string>
 
 namespace cudaq::qio {
+/// Default upper bound on the number of bytes `gzipDecompress` will produce.
+inline constexpr std::size_t defaultMaxDecompressedSize = 512UL * 1024 * 1024;
+
 std::string gzipCompress(const std::string &input);
-std::string gzipDecompress(const std::string &input);
+
+/// Decompress a `zlib` stream, which produces `maxOutputSize` bytes.
```
1tnguyen: nit: I think the comment "which produces `maxOutputSize` bytes" is not accurate. Perhaps, less than `maxOutputSize` bytes?

1tnguyen (approved):
LGTM 👍

runtime/cudaq/platform/default/rest/helpers/scaleway/qio/Compression.h
sacpis: Missed the word `atmost`.

sacpis: Fixed in 07e08b7.

mitchdz (approved):

PR #4806, opened 2026-06-30
[Realtime] Refactor `DeviceCallService` interface 

1tnguyen:
As downstream libraries (e.g., CUDA-QX) may extend the `DeviceCallService` interface, we move it from `runtime/internal/device_call` to `runtime/cudaq/realtime/device_call_service.h`. 

Note: this is just for the base abstract interface.

While doing this, we also trim down its API surface: remove a bunch of hooks (which are only suitable for an internal interface) so the service owns everything.


vedika-saravanan (approved):
Thanks, Thien. LGTM

PR #4807, opened 2026-06-30
Fail publishing if installer simulator libraries are missing

mitchdz:
The Installer build job is marked allow_failure, so a failed MGMN SV simulator build (e.g. a base-image pull failure) let the deploy stage upload only the build-info file, silently shipping an installer release without the closed-source simulator.

Before uploading cuda-quantum-assets, require that the libnvqir-*.so plugins for this pipeline's architecture were produced, and fail the job otherwise.


1tnguyen (dismissed):
👍

PR #4809, opened 2026-06-30
[build] Upgrade to CMake 4

lmondada:
As discussed in standup, I needed the [improvements on `WHOLE_ARCHIVE`](https://cmake.org/cmake/help/latest/variable/CMAKE_LINK_LIBRARY_FEATURE_ATTRIBUTES.html#example) handling introduced in cmake 3.30.

I've made all the upgrades to our CI and dev environments to CMake 4.0. However, we're keeping support for the time being for CMake >= 3.30, to avoid breaking downstream.

The CI and dev images pull the CMake binaries from the project directly rather than their packaged versions. This can be temporary:
- The latest Ubuntu LTS 26.04 ships with version 4.2, so we'll be golden as soon as we do the upgrade on our CI. 
- On the other hand, the most recent version of RHEL still ships with 3.31


schweitzpgi:
I use CMake 4.3.2 myself.

FWIW, it spews a ton more warnings. :)

docker/test/installer/runtime_dependencies.sh
```diff
@@ -126,10 +137,22 @@ elif [ "$pkg_manager" == "zypper" ]; then
...
 else
     echo "Installation via $pkg_manager is not yet implemented." >&2
     (return 0 2>/dev/null) && return 1 || exit 1
 fi
 
+## [CMake via Kitware installer]
+if [ -n "${INSTALL_CMAKE_VERSION}" ]; then
+    wget -q "https://github.com/Kitware/CMake/releases/download/v${INSTALL_CMAKE_VERSION}/cmake-${INSTALL_CMAKE_VERSION}-linux-$(uname -m).sh" \
+        -O cmake-install.sh
+    bash cmake-install.sh --skip-licence --exclude-subdir --prefix=/usr/local
```
mitchdz: is license misspelled here?

.github/workflows/realtime_ci.yml
```diff
@@ -85,10 +85,20 @@ jobs:
           # Install cmake for building a test example that uses the installed CUDA Quantum Realtime
           # Install cmake depending on distro
           if [[ "${{ matrix.distro }}" == "ubi9" ]]; then
-            retry_dnf install -y cmake
+            # cmake shipped with ubi9 is too old (3.20), so use the Kitware .sh installer.
+            retry_dnf install -y wget
+            wget -q https://github.com/Kitware/CMake/releases/download/v4.0.7/cmake-4.0.7-linux-$(uname -m).sh -O cmake-install.sh
+            bash cmake-install.sh --skip-licence --exclude-subdir --prefix=/usr/local
```
mitchdz: is license misspelled here?

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-cmake_minimum_required(VERSION 3.22 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.30 FATAL_ERROR)
```
mitchdz, line 9: How do we know this is the correct floor when we actually install 4.0?

scripts/install_prerequisites.sh
```diff
@@ -593,6 +593,7 @@ if [ -n "$AWS_INSTALL_PREFIX" ] && [ -z "$(echo $exclude_prereq | grep aws)" ];
     cmake -G Ninja .. \
       -DCMAKE_INSTALL_PREFIX="${AWS_INSTALL_PREFIX}" \
       -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
```
mitchdz, line 596: Adding the cmake minimum version of 3.5 seems to contradict the docs saying cudaq is compatible with 3.30

lmondada: This is CMake versioning weirdness (that I also learnt about working on this). The minimum CMake version you support affects two things
- on the one hand, if you raise the minimum version, you get to use CMake's new features (as you'd expect)
- on the other hand, the minimum version determines the behaviour of CMake. All versions of CMake will adjust their behavior based on `cmake_minimum_required` to act in a backwards compatible way. In other words, any version of CMake bakes in multiple behaviors for each policy, and which one is picked is based on the declared `cmake_minimum_required`.

Here's the thing. With CMake 4.0, all the backwards compatibility below 3.5 was removed. This means that any project that advertises support for CMake <3.5 will fail to configure with CMake 4 (because it would require knowing how to emulate the old behavior, which was removed). So here, because the AWS SDK advertises support for old CMake, we need to pass `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to override the declared `cmake_minimum_required`. We are essentially telling CMake 4 to act as if it was CMake 3.5, which is the oldest CMake behaviour it can emulate.

CMakeLists.txt
lmondada: I want to bump this to 4 as soon as possible, however I'm leaving in support for 3.30 at the moment so that CUDAQX can keep using CMake 3.30 for the time being.

I agree that in principle, the installed CMake should be the minimum supported, however I want to avoid going through the process of upgrading CI and everyone's dev environment twice. I've tested with 3.30 and I think this is reasonable as a temporary solution (and tbh this is already how it was: we were advertising support for 3.20 but were using 3.28).

docker/test/installer/runtime_dependencies.sh
lmondada: this is UK vs US spelling 😅 you're right though, I've doubled checked and it should be `--skip-license`. Will fix!

mitchdz (approved):
Thanks for answering all my questions @lmondada ! I learned some fun things about CMake. This all looks good to me, let's :shipit:

PR #4811, opened 2026-06-30
[Build] Remove `minizip` remnant from `scripts/bootstrap_prerequisites.sh`

1tnguyen:
We don't need `minizip` anymore (was only used for unzipping NVQC responses). 

It was removed in `scripts/install_prerequisites.sh`. However, it was not synced with the latter `scripts/bootstrap_prerequisites.sh`.

This PR fixes that.


schweitzpgi (approved):
Thanks for cleaning this up, Thien.

PR #4814, opened 2026-06-30
Validating nQubits in IonQ processResults

sacpis:
Validating nQubits in IonQ processResults


runtime/cudaq/platform/default/rest/helpers/ionq/IonQServerHelper.cpp
```diff
@@ -390,6 +390,9 @@ IonQServerHelper::processResults(ServerMessage &postJobResponse,
         "ServerMessage doesn't tell us how many qubits there were");
 
   auto nQubits = jobs[0].at("qubits").get<int>();
+  if (nQubits < 0 || nQubits > 64)
```
khalatepradnya: Q: Curious to know whether `nQubits == 0` is valid?

sacpis: I don't think 0 qubits makes sense here. Others can chime in.

1tnguyen: I also don't think zero qubits makes a lot of sense for remote hardware backends (i.e., `<= 0` condition).
If we want to be sure, we can submit an empty job, perhaps?

sacpis: Used the following empty kernels

```
struct empty {
  void operator()() __qpu__ {
    // cudaq::qubit q;
    // mz(q);
  }
};

int main() {
  auto counts = cudaq::sample(empty{});
  counts.dump();

  return 0;
}
```
which gives output as
```
{ :1000 }
```

and for this one

```
struct empty {
  void operator()() __qpu__ {
    cudaq::qubit q;
    mz(q);
  }
};

int main() {
  auto counts = cudaq::sample(empty{});
  counts.dump();

  return 0;
}
```
it shows

```
{ 
  __global__ : { 0:1000 }
   r00000 : { 0:1000 }
}
```

Will update the condition to `<=0` as @khalatepradnya suggested so that we don't send this job to the server as it does not makes sense.

schweitzpgi (commented):
LGTM, modulo the 0 case as has already been mentioned.

runtime/cudaq/platform/default/rest/helpers/ionq/IonQServerHelper.cpp
sacpis: Fixed in 04b88ea.

khalatepradnya (approved):

PR #4816, opened 2026-07-01
[Realtime] Use bit-packed representation for bool vectors

1tnguyen:
Updates `realtime` device-call lowering to use the canonical bit-packed representation for `std::vector<bool>` and measurement vectors.

- Emit `CUDAQ_TYPE_BIT_PACKED` for boolean and measurement arrays.

- Pack input values into bytes during device-call lowering and unpack bit-packed output values into boolean vectors.

- Add native bit-packed handler tests (both input and output boolean vectors) with multiple vector sizes.

Code reorg: move device-call lowering to `realtime` into a separate pass, not injecting into the `distributed-device-call` direct lowering pass. Hence, all `realtime` protocol requirements/conventions are captured in a single pass. Use intrinsics for the packing/unpacking procedures.


cudaq/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -115,6 +115,11 @@ static std::uint64_t realtimeArrayElementSize(Type ty) {
   return realtimePrimitiveSize(ty);
 }
 
+static bool isRealtimeBitPackedType(Type type) {
+  const auto integerType = dyn_cast<IntegerType>(type);
+  return integerType && integerType.getWidth() == 1;
```
schweitzpgi: I think this test is already written somewhere
```suggestion
  if (auto intTy = type.dyn_cast<mlir::IntegerType>())
    return intTy.getWidth() == 1;
  return false;
```

cudaq/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
```diff
@@ -233,8 +238,19 @@ static Value writeLengthPrefix(OpBuilder &builder, Location loc, Value cursor,
       builder, loc, cursor, i64Constant(builder, loc, sizeof(std::uint64_t)));
 }
 
+/// Compute the number of payload bytes required for `length` logical array
```
schweitzpgi: Is realtime using `std::vector<bool>`?

You should not roll your own code for that. Different STL implementations use different encodings, so whatever you do will be wrong if someone uses a different STL.

schweitzpgi (changes requested):
All of the issues related to `std::vector<bool>` portability have been solved for kernel launches from C++. Device calls should be reusing the same code and methodology (albeit in reverse) and not rolling something new.

Perhaps `realtime` has its own unique `realtime::vector<bool>` type though? That would be strange indeed.

bmhowe23:
> All of the issues related to `std::vector<bool>` portability have been solved for kernel launches from C++. Device calls should be reusing the same code and methodology (albeit in reverse) and not rolling something new.
> 
> Perhaps `realtime` has its own unique `realtime::vector<bool>` type though? That would be strange indeed.

cudaq-realtime has a type description for bit-packed types [here](https://github.com/NVIDIA/cuda-quantum/blob/bfe5266dd09ca6d8b2bb99d2d57ee084b89458aa/realtime/include/cudaq/realtime/daemon/dispatcher/cudaq_realtime.h#L103). We need `std::vector<bool>` to be able to lower to this format so that we can send reasonably packed data over the wire to the decoder (i.e. not have 7 out of every 8 bis be a wasted 0).

cudaq/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: I've simplified the code to just use `isInteger(1)` directly instead.

cudaq/lib/Optimizer/Transforms/DistributedDeviceCall.cpp
1tnguyen: Yeah, `realtime` is expected to send `vector<bool>` as `TYPE_BIT_PACKED` (LSB)

1tnguyen:
I've revised the code structure to pull all the `realtime`-specific lowering, which requires knowledge of the `realtime` message protocol to a dedicated pass to encapsulate the knowledge of this protocol. I also added comments about the consideration for byte-ordering encoding (for other types, not bit-packed array). Currently, we don't handle any byte-ordering conversion.

cudaq/test/NVQPP/Inputs/device_call_realtime_bitpacked_native_shim.cu
```diff
@@ -0,0 +1,149 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "DeviceCallLibrary.h"
+
```
schweitzpgi:
```suggestion
```
nit

cudaq/test/NVQPP/Inputs/DeviceCallLibrary.h
```diff
@@ -242,37 +342,88 @@ struct DeviceCallWrapper<Fn> {
         if constexpr (is_output_array_pointer<Arg>::value) {
           // Output array (non-const pointer): elements are written into the TX
           // result buffer rather than read from the RX argument buffer.
-          if (byRefResultLen > maxResultLen ||
-              elementCount > (maxResultLen - byRefResultLen) / sizeof(Element))
+          if (byRefResultLen > maxResultLen)
+            return -1; // result buffer too small
```
schweitzpgi: May want to consider a return value of type `std::optional<integral>` to avoid magic values.

cudaq/test/NVQPP/device_call_realtime_array.cpp
```diff
@@ -30,6 +30,7 @@ extern "C" int countTrueBits(const std::vector<bool> &bits, std::uint64_t bias);
 extern "C" int countTrueMeasures(const std::vector<cudaq::measure_result> &bits,
                                  std::uint64_t bias);
 extern "C" int sumIntVector(const std::vector<int> &values, int bias);
+extern "C" int sumByteVector(const std::vector<std::uint8_t> &values, int bias);
```
schweitzpgi: May want to consider block style here to avoid all this redundant text.
```c++
extern "C" {
... function decls ...
}
```

schweitzpgi (approved):
Thanks for the updates, Thien. Looks like everything we discussed last week.

cudaq/test/NVQPP/device_call_realtime_array.cpp
1tnguyen: Absolutely, fixed in [2f2f454](https://github.com/NVIDIA/cuda-quantum/pull/4816/commits/2f2f45493ee7476b68e665481ec2680e8fde4340)

cudaq/test/NVQPP/Inputs/DeviceCallLibrary.h
1tnguyen: Indeed. For this test library, those internal helpers only return pass/fail (`true`/`false`) so I updated them accordingly and only return a status code for `realtime` API requirement ([2f2f454](https://github.com/NVIDIA/cuda-quantum/pull/4816/commits/2f2f45493ee7476b68e665481ec2680e8fde4340))

PR #4818, opened 2026-07-01
Qubit ordering problem on IQM backends  #4621

iqm-bhoffmann:
For the IQM backend correct if needed the order of the qubits returned in the counts bitstrings. This includes also cases in which the list of measured qubits has gaps.

To know in which measurement a bit in the bitstring represents the corresponding "measurement_keys"  information needs to be interpreted. This is now used to order the bitstrings physical qubit order as expected by CUDA-Q.

Testing:
The IQM mock server did not return a correct list of measurement keys with the results. Because of this the reordering of bitstring results in the IQM backend did not work. The list of measurement keys is now derived from the given circuit and returned together with the counts results. Implemented the GET V1 measurement_counts API needed for the reordering code in the IQMServerHelper because it is the only returning correct measurement keys. Both changes are necessary to pass the ctests for IQM.

Tickets:
[SW-2741]
#4621


sacpis:
> /ok to test 2ed4b56

unittests/nvqpp/backends/iqm/IQMTester.cpp
```diff
@@ -96,10 +96,8 @@ CUDAQ_TEST(IQMTester, processResultsAppliesReorderIdxFromConfig) {
   serverHelper->initialize(config);
 
   cudaq::ServerMessage response = {
-      {"status", "ready"},
-      {"counts_batch",
-       cudaq::ServerMessage::array(
-           {{{"counts", {{"100", 1000}}}, {"measurement_keys", {"m_0"}}}})}};
+      {{"counts", {{"100", 1000}}},
+       {"measurement_keys", {"m_QB0, m_QB1, m_QB2"}}}};
```
sacpis: Should `measurement_keys` be `"m_QB0", "m_QB1", "m_QB2"` in order to check the ordering of qubits? I think with the current value, the reorder flag would stay false.

runtime/cudaq/platform/default/rest/helpers/iqm/IQMServerHelper.cpp
```diff
@@ -270,34 +275,112 @@ bool IQMServerHelper::jobIsDone(ServerMessage &getJobResponse) {
   auto jobStatus = getJobResponse["status"].get<std::string>();
   std::unordered_set<std::string> terminalStatuses = {"ready", "failed",
                                                       "aborted"};
-  return terminalStatuses.find(jobStatus) != terminalStatuses.end();
+  bool done = terminalStatuses.find(jobStatus) != terminalStatuses.end();
+
+  if (done) {
+    // if the job failed exit with an exception
+    auto jobStatus = getJobResponse["status"].get<std::string>();
```
sacpis: Wondering if we can use the jobStatus from line 275?

iqm-bhoffmann: Good catch. Yes the status can of course be reused.
This is an artifact coming from the last refactoring which nobody noticed. Thank you. 👍

unittests/nvqpp/backends/iqm/IQMTester.cpp
iqm-bhoffmann: Well, this is the test for the issue #4625 which had to be adapted because the expected response had changed. The issue #4625 fixes a bug of incorrect usage of the framework.
The issue #4621 is a problem created by the IQM servers not returning the measurements in always physical order. This reordering could be tested with the same code but is not addressed here.

zohimchandani:
@sacpis any ideas when this will be merged please? 

@orlov-a for vis

sacpis:
> /ok to test fff5926

sacpis:
> @sacpis any ideas when this will be merged please?
> 
> @orlov-a for vis

@zohimchandani Will merge it today if everything looks good.

sacpis (approved):
LGTM. Thanks @iqm-bhoffmann.

PR #4822, opened 2026-07-01
Validate cusvsim libraries are bundled in published artifacts

mitchdz:
The nvidia target and its options (fp32/fp64 and mgpu) are backed by the cusvsim GPU simulator libraries (libnvqir-cusvsim-* and libnvqir-nvidia-mgpu-*). If a packaging regression drops these, installation validation passes silently: the affected backends are simply skipped when their library is absent rather than failing.

The existing check parsed the validation script's stdout for an "fp32,mgpu" test line, which only confirmed a test was attempted (not that the library is present) and is brittle as the nvidia target options evolve. Replace it with an explicit library-presence check.

- Add scripts/validate_cusvsim_libraries.sh, which verifies the required libraries exist under a given install/search root.
- Run it in the installer, image, and wheel validation jobs so publishing fails when the cusvsim libraries are missing from an artifact:
    - installer and image: require the full single-device + mgpu set
    - wheel (pip only): require the single-device libraries the wheel exercises (mgpu needs MPI to run and is not asserted here)
- Remove the previous stdout-parsing checks in the installer and image jobs.


1tnguyen (approved):
LGTM 👍

PR #4824, opened 2026-07-01
Support disconnected device topologies in the qubit mapper

### LINKED ISSUE #4681: Invalid qubit mapping on disconnected topology
### Required prerequisites

### Describe the bug

When using qubit mapping pass with a topology that contains isolated qubits, it outputs invalid results.

- Device topology:  10 qubits, only edge {5,6}:
```
Number of nodes: 10

Number of edges: 1

0 --> {}
1 --> {}
2 --> {}
3 --> {}
4 --> {}
5 --> {}
6 --> {5}
7 --> {}
8 --> {}
9 --> {}
```

- Quake code 

```
quake.wire_set @wires[2147483647]

func.func @bell() attributes {"cudaq-entrypoint", "cudaq-kernel"} {
  %0 = quake.null_wire
  %1 = quake.null_wire
  %2 = quake.h %0 : (!quake.wire) -> !quake.wire
  %3:2 = quake.x [%2] %1 : (!quake.wire, !quake.wire) -> (!quake.wire, !quake.wire)
  %m0, %4 = quake.mz %3#0 : (!quake.wire) -> (!quake.measure, !quake.wire)
  %m1, %5 = quake.mz %3#1 : (!quake.wire) -> (!quake.measure, !quake.wire)
  quake.sink %4 : !quake.wire
  quake.sink %5 : !quake.wire
  return
}
```

- Command:
 
```
cudaq-opt --assign-wire-indices '--qubit-mapping=device=file(disconnected.txt)' bell.mlir
```

The output still maps the Bell pair to physical qubits 0 and 1, even though they are not connected.

### Expected behavior

- We shouldn't allow an invalid mapping to succeed (i.e., gates between disconnected qubits)

- The mapping pass can find the connected qubit region to map the qubits to.

Not a regression

### Environment

- **CUDA-Q version**:  `main`
- **Python version**:  3.11
- **C++ compiler**: gcc-12

### Suggestions

_No response_

khalatepradnya:
### Summary

* Address issue #4681 
* Built using logic from PR #4753 

* Teaches the qubit mapper to model and respect disconnected device coupling graphs. Previously a device whose connectivity graph had more than one connected component could silently produce an *invalid* mapping (a two-qubit gate routed across qubits with no physical path between them). 
* The mapper now represents unreachable qubit pairs explicitly, rejects circuits it cannot map with a precise diagnostic, and only emits mappings that satisfy the device's physical adjacency constraints.


cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -137,8 +138,13 @@ class GreedyInitialPlacer {
...
     physDegree.assign(n, 0);
     for (unsigned p = 0; p < n; ++p) {
-      for (unsigned q = 0; q < n; ++q)
-        distanceSum[p] += device.getDistance(Qubit(p), Qubit(q));
+      for (unsigned q = 0; q < n; ++q) {
+        unsigned distance = device.getDistance(Qubit(p), Qubit(q));
+        // Any finite shortest path has fewer than `n` edges, so `n` keeps
+        // unreachable qubits less central without risking sentinel overflow.
+        distanceSum[p] +=
+            distance == cudaq::Device::unreachableDistance ? n : distance;
```
taalexander: I wonder how this plays with disjoint islands (see my top-level comment)

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -287,13 +305,162 @@ class GreedyInitialPlacer {
...
+    components.push_back(std::move(component));
+  }
+  return components;
+}
+
+/// Physical islands of mutually reachable qubits. A two-qubit gate can never be
+/// routed (even with swaps) across different islands, so placement and
+/// rejection both reason in terms of these.
+SmallVector<SmallVector<unsigned>>
+computeDeviceComponents(const cudaq::Device &device) {
```
taalexander: Is `DeviceComponents` the right word for this? `ConnectedQubitSets` might be more clear?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -287,13 +305,162 @@ class GreedyInitialPlacer {
...
+        for (const auto &edge : interactions.neighbors(current))
+          visit(edge.first);
+      });
+}
+
+/// On a disconnected device the plain greedy seed can strand an interacting
+/// group across islands (un-routable). This offers an alternative that keeps
+/// each group inside one island via best-fit-decreasing bin-packing;
+/// `std::nullopt` means no such packing was found, so `auto` falls back to its
+/// other seeds.
```
taalexander: Can this comment explain how it works/it is expected this is going to be used as a comment. I think I understand but the comment doesn't do a great job explaining.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -287,13 +305,162 @@ class GreedyInitialPlacer {
...
+      if (deviceQubits.size() < virtualComponent.size())
+        continue;
+      if (bestComponent == availableDeviceQubits.size() ||
+          deviceQubits.size() < availableDeviceQubits[bestComponent].size() ||
+          (deviceQubits.size() == availableDeviceQubits[bestComponent].size() &&
+           deviceQubits.front() < availableDeviceQubits[bestComponent].front()))
+        bestComponent = componentId;
+    }
+    if (bestComponent == availableDeviceQubits.size())
+      return std::nullopt;
```
taalexander: If `if (bestComponent == availableDeviceQubits.size() ||` can I possibly return an invalid size and not reach the std::nullopt clause below resulting in an invalid placement?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -287,13 +305,162 @@ class GreedyInitialPlacer {
...
+
+    auto &deviceQubits = availableDeviceQubits[bestComponent];
```
taalexander: A comment here would be helpful (and with the next clause below as well).

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -1702,14 +1932,38 @@ struct MappingFunc : public cudaq::opt::impl::MappingFuncBase<MappingFunc> {
...
+        func.emitError("cannot place two-qubit interaction between virtual "
+                       "qubits " +
+                       std::to_string(identityBlockedInteraction->first) +
+                       " and " +
+                       std::to_string(identityBlockedInteraction->second) +
+                       " on disconnected device topology");
+      } else {
+        func.emitError("could not find a routable initial layout for "
+                       "disconnected device topology");
       }
```
taalexander, line 2294: Could this be something like?

```cpp
auto shouldDiscardSeed = [&](const auto &seed) {
  const auto blocked =
      findUnroutableInteraction(seed, *deviceInstance, interactions);
  if (!blocked)
    return false;

  if (placementStrategy == PlacementStrategy::Identity &&
      !identityBlockedInteraction)
    identityBlockedInteraction = blocked;

  return true;
};

std::erase_if(seeds, shouldDiscardSeed);
```

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander, line 464: Does this actually do bin-packing or just a decreasing order best-fit algorithm? This might need another look (I think it goes hand in hand with my question about different topology tests in the main comment).

taalexander (changes requested):
I think we might be missing a test case for mapping a circuit into two disjoint sets? Eg., 0-2 connected and 3-5 connected. What happens if we have a circuit with 6 qubits that cleanly splits onto the two disjoint islands? Will it be mapped? What about if we have three islands, eg., 0-1, 2-3 and 4-5?

Should there be a `Device` focused unit-test for some of the above?

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: In isolation, may be, but see other functions like `computeVirtualComponents` and the template these both call -`computeComponents`, and "components" may read better. I can change if you feel strongly against this.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: Sure, this reads better, replaced.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: Added clarifying comments.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: Updated in https://github.com/NVIDIA/cuda-quantum/commit/ffbbf13b2d66c03fe0a43b7d15e6846885546745

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: decreasing order best-fit algorithm

khalatepradnya: I couldn't find such an example (but I tried only small examples), if you have something in mind, I can add it to the test suite.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: How do you mean? See the `*.txt` files for the device topology and how they are invoked across tests.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander: I looked again and I think this is safe as the first clause is true on the first island  (doesn't evaluate the later clauses which have invalid idices) and that is valid.  This then immediately changes bestComponent so the second clause is no longer valid. It's only in the case where componentId is equal to the device qubits where this comes up which is expected.

cudaq/lib/Optimizer/Transforms/Mapping.cpp
taalexander: What I was getting at is that it seems to me that device-island assignment is a constraint rather than an optional placement strategy and should be applied outside of whether greedy or identity is selected. Could greedy first assign interaction components to islands, then apply its current heuristic within each island? auto could then compare identity with one or more component-aware greedy seeds. This would make the options’ behavior easier to reason about and avoid greedy rejecting a layout solely because the device is disconnected.

khalatepradnya: Addressed in 5b0e109

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -372,84 +408,106 @@ computeVirtualComponents(const VirtualInteractionGraph &interactions,
...
+/// greedy walk places every virtual qubit over the whole device. On a
+/// multi-island (disconnected) device, each interacting virtual component is
+/// packed onto one island first, then the same greedy heuristic runs again
+/// restricted to that island's qubits.
+std::optional<SmallVector<unsigned>>
+buildGreedySeed(unsigned numV, const cudaq::Device &device,
+                const VirtualInteractionGraph &interactions,
+                ArrayRef<bool> userVirtualQubits) {
+  SmallVector<SmallVector<unsigned>> deviceComponents =
+      computeDeviceComponents(device);
```
taalexander: Nit - should this be `computeDeviceIslands` now? (same with other mentions of components?)

cudaq/lib/Optimizer/Transforms/Mapping.cpp
```diff
@@ -372,84 +408,106 @@ computeVirtualComponents(const VirtualInteractionGraph &interactions,
...
+  SmallVector<unsigned> virtualQubits;
+};
+
+/// Assign each interacting virtual component wholly to one device island.
+/// Components are packed largest-first into the smallest remaining capacity
+/// that fits. This best-fit-decreasing heuristic is deterministic but not
+/// complete: it can fail even when a different packing exists.
+std::optional<SmallVector<IslandPlan>>
+buildBestFitIslandPlan(SmallVector<SmallVector<unsigned>> deviceComponents,
+                       SmallVector<SmallVector<unsigned>> virtualComponents) {
```
taalexander: I've been thinking about the packing algorithm more and trying to understand the case where it fails (and is expected to). Let's say there are available islands of size `{5, 4, 4}` and a requested islands of `{3, 3, 2, 2}` to be packed. `{3+2, 3, 2+2}` should be valid packings but I think this approach will assign {3, 3, 2}` and then fail on the remaining 2?

General bin packing is NP-complete and the BFD is completely valid. The other aspect of the packing right now is we're just looking for an island that fits the requested qubits. Not how well that island ultimately supports the qubits. This is an optimization to consider at a further point in time that might be worth noting as a comment (or leaving an open issue for).

taalexander (approved):
LGTM, thanks @khalatepradnya!

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: Yes. Renamed in [efc0c2d](https://github.com/NVIDIA/cuda-quantum/pull/4824/commits/efc0c2dda2960f5af125d077e696dfd0ac782503)

cudaq/lib/Optimizer/Transforms/Mapping.cpp
khalatepradnya: Good points. `{5,4,4}` with `{3,3,2,2}` succeeds under the current best-fit logic (added another test). 
A representative miss is `{5,4}` with `{3,2,2,2}`. I’ll track topology-aware island selection separately.

PR #4825, opened 2026-07-01
[testing][Python] Fix backend `observe` tests

khalatepradnya:
### Summary

Fix intermittent Python backend observe-test failures by propagating explicit random seeds across the execution boundaries the tests use.

- Seed Quantinuum and IonQ mock servers inside their multiprocessing child processes.
- Capture and apply the caller seed in the local `observe_async` worker thread.

### Root cause

The tests seeded their parent Python process, but mock servers run in `forkserver` children and local async observation runs on a worker thread.
CUDA-Q random-seed and simulator state are thread-local, so those execution contexts started with the default nondeterministic seed.

### DIFF
--- a/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
+++ b/python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
@@ -126,7 +126,8 @@ def test_quantinuum_exp_pauli():
 
     # Run the observe task on quantinuum synchronously
     res = cudaq.observe(ansatz, hamiltonian, .59 * -0.5, shots_count=100000)
-    assert assert_close(-1.7, res.expectation())
+    sync_expectation = res.expectation()
+    assert assert_close(-1.7, sync_expectation)
 
     # Launch it asynchronously, enters the job into the queue
     future = cudaq.observe_async(ansatz,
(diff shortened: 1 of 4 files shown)


python/tests/backends/test_Quantinuum_LocalEmulation_kernel.py
```diff
@@ -126,7 +126,8 @@ def ansatz(theta: float):
 
     # Run the observe task on quantinuum synchronously
     res = cudaq.observe(ansatz, hamiltonian, .59 * -0.5, shots_count=100000)
-    assert assert_close(-1.7, res.expectation())
+    sync_expectation = res.expectation()
+    assert assert_close(-1.7, sync_expectation)
```
sacpis, line 130: Is this needed? Sorry didn't follow this change.

khalatepradnya: Not needed. I ran the tests 25+ times and also compared the values from the sync and async executions, `sync_expectation == async_expectation`, to see the effect of the seed. I dropped that comparison but kept this, doesn't hurt... can remove it if seems confusing.

bmhowe23 (approved):

Issue #4791, opened 2026-06-28
Quantum embeddings: amplitude_encode accepts nested Python sequences, plus smaller follow-ups


spital:
Hi CUDA-Q team, thank you for adding the quantum embedding helpers on `main` and resolving #2982. I had been working locally on the same feature area during unitaryHACK 2026, but I was not able to submit that work because I had reached the unitaryHACK 2026 PR limit. After testing the current `main` branch locally, I found one silent correctness issue and a few smaller robustness / docs follow-ups that seem worth tracking.

I checked the items below against `main` at commit `3344a8a99` (`3344a8a997b93bba04e89a524386cc45c0cb3176`, 2026-06-27). I did not find a duplicate issue for these in `NVIDIA/cuda-quantum`.

To reproduce from the same revision:

```bash
git clone --filter=blob:none https://github.com/NVIDIA/cuda-quantum.git
cd cuda-quantum
git checkout 3344a8a99   # or use current main
git rev-parse --short HEAD
```

## Summary

| # | Area | Item | Severity |
|---|------|------|----------|
| 1 | Python correctness | `amplitude_encode` rejects 2-D NumPy arrays but silently flattens equivalent nested Python `list` / `tuple` inputs | High |
| 2 | C++ portability | `runtime/cudaq/builder/kernels.h` uses `std::optional` without directly including `<optional>` | Low |
| 3 | Docs | Unicode math glyphs are used inside Doxygen / Sphinx math expressions | Low |
| 4 | C++ robustness | `amplitude_encode(const state&)` reads `get_tensor(0)` without validating tensor count / rank in that helper | Low-Med |
| 5 | C++ robustness | `nextPowerOfTwo` can overflow into a non-terminating loop for pathological sizes | Low |
| 6 | Python API consistency | Builder-mode `angular_encode` rejects NumPy angle arrays, while `amplitude_encode` accepts NumPy arrays | Low / maybe intentional |

## 1. `amplitude_encode` silently flattens nested Python sequences

`amplitude_encode` documents that input must be one-dimensional, and it correctly rejects a 2-D NumPy array. The equivalent nested Python `list` or `tuple`, however, takes the list/tuple branch and is flattened with `.ravel()` before any `ndim` check.

Source check:

```bash
grep -n -A1 'isinstance(data, (list, tuple))' python/cudaq/contrib/encoding.py
grep -n -A2 'isinstance(data, np.ndarray)' python/cudaq/contrib/encoding.py
[...]

PR #4826, opened 2026-07-02
fix(python): reject 2-D list/tuple in amplitude_encode

aryanputta:
## Summary

`cudaq.contrib.amplitude_encode` raises `ValueError("expected a 1D vector")` for a 2-D NumPy array, but an **equivalent nested Python list/tuple is silently flattened** by `np.asarray(data, dtype=dtype).ravel()` in `_as_amplitude_vector`, producing a different-length state with no error.

```python
cudaq.contrib.amplitude_encode(np.eye(2))          # ValueError (correct)
cudaq.contrib.amplitude_encode([[0.5, 0.5],
                                [0.5, 0.5]])       # silently encodes a 4-vector (wrong)
```

This addresses **item 1 (High severity, Python correctness)** of #4791.

## Fix

Validate `ndim` in the `list`/`tuple` branch of `_as_amplitude_vector`, mirroring the existing NumPy-array path, so nested sequences raise the same `ValueError`. 1-D lists/tuples and empty lists are unaffected.

## Test

Adds `test_amplitude_encode_2d_list_rejected` (nested list and nested tuple), mirroring the existing `test_amplitude_encode_2d_rejected` array case.

Scope is limited to item 1; the C++/docs follow-ups in #4791 are left for separate PRs.

Credit to @spital for the detailed report.

_Note: verified the input-handling logic standalone with NumPy; the full CUDA-Q Python suite requires a built runtime._

### DIFF
--- a/python/tests/contrib/test_amplitude_encode.py
+++ b/python/tests/contrib/test_amplitude_encode.py
@@ -99,3 +99,12 @@ def test_amplitude_encode_zero_vector():
 def test_amplitude_encode_2d_rejected():
     with pytest.raises(ValueError, match="1D vector"):
         cudaq.contrib.amplitude_encode(np.eye(2), pad=0)
+
+
+def test_amplitude_encode_2d_list_rejected():
+    # A nested list/tuple must be rejected like the equivalent 2-D array,
+    # rather than being silently flattened into a 1-D vector.
+    with pytest.raises(ValueError, match="1D vector"):
+        cudaq.contrib.amplitude_encode([[0.5, 0.5], [0.5, 0.5]], pad=0)
+    with pytest.raises(ValueError, match="1D vector"):
+        cudaq.contrib.amplitude_encode(((0.5, 0.5), (0.5, 0.5)), pad=0)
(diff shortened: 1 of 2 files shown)


python/tests/contrib/test_amplitude_encode.py
```diff
@@ -99,3 +99,12 @@ def test_amplitude_encode_zero_vector():
...
         cudaq.contrib.amplitude_encode(np.eye(2), pad=0)
+
+
+def test_amplitude_encode_2d_list_rejected():
+    # A nested list/tuple must be rejected like the equivalent 2-D array,
+    # rather than being silently flattened into a 1-D vector.
+    with pytest.raises(ValueError, match="1D vector"):
+        cudaq.contrib.amplitude_encode([[0.5, 0.5], [0.5, 0.5]], pad=0)
+    with pytest.raises(ValueError, match="1D vector"):
+        cudaq.contrib.amplitude_encode(((0.5, 0.5), (0.5, 0.5)), pad=0)
```
sacpis: Are these the same tests?

aryanputta: Good catch. They were not really different tests: a nested list and a nested tuple both convert to the same 2-D array through np.asarray, so both assertions were exercising the exact same rejection path. I collapsed them into a single parametrized test that runs over both container types, so the list/tuple coverage stays without duplicating the assertion.

PR #4829, opened 2026-07-02
Preserve adjoints in quake simplification

taalexander:
This change fixes incorrect adjoint handling in quake-simplify. Previously, inverse pairs such as $T$, $T^\dagger$ and $S$, $S^\dagger$ were rewritten to $S$ and $Z$ instead of cancelling, while $T^\dagger$ and  $T^\dagger$ incorrectly produced $S$ instead of $S\dagger$. These errors change relative phase and are observable circuit differences. Likewise for `ReduceYSX` under control the optimization is  not valid as the global phase is observable, ie., it is $YSX \rightarrow e^{i\pi} S$ (@cabreraam , fyi). 

I am considering how to build a grindable validation framework for optimization passes and I'd fold global phase into this in the future.

The updated rewrites account for and preserve adjoint attributes. I've also added regressions covering controlled and uncontrolled cases.

### DIFF
--- a/cudaq/lib/Optimizer/Transforms/QuakeSimplify.cpp
+++ b/cudaq/lib/Optimizer/Transforms/QuakeSimplify.cpp
@@ -297,7 +297,8 @@ public:
   }
 };
 
-// Z = SS
+// Z = SS = S<adj>S<adj>
+// I = SS<adj> = S<adj>S
 class DoubleSOp : public OpRewritePattern<cudaq::quake::SOp> {
 public:
   using OpRewritePattern::OpRewritePattern;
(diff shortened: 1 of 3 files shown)


cabreraam (approved):
Nice, thanks for working on this @taalexander; LGTM!

Just a couple nits: is it possible to add a couple comments [here](https://github.com/taalexander/cuda-quantum/blob/5e1e50564e4d6adca463b17f18cd38b744da6ff2/cudaq/lib/Optimizer/Transforms/QuakeSimplify.cpp#L300) and [here](https://github.com/taalexander/cuda-quantum/blob/5e1e50564e4d6adca463b17f18cd38b744da6ff2/cudaq/lib/Optimizer/Transforms/QuakeSimplify.cpp#L386) just saying that a double op might also be a gate and it's adjoint? Just so readers know that those cases are also now handled? Just something like

```
I = SS_adj
```

and 

```
I = TT_adj
```

---

I'm not quite sure what "grindable validation framework" means, but I can appreciate that the transform for $YSX \rightarrow e^{i\pi}S$ should be held off until global phase tracking is handled.

PR #4820, opened 2026-07-01
[build] Continued improvements to modularity.

schweitzpgi:
Make inclusion of nlopt at the user's discretion. Default ON.

PR #4810, opened 2026-06-30
[build] Add a better error message if OpenSSL is MIA.

schweitzpgi:

PR #4831, opened 2026-07-02
[build] Fix export namespace loss for cudaq-nlopt/cudaq-ensmallen deps

bmhowe23:
CMake does not rewrite target names to their namespaced export form when they are embedded inside a generator expression, only when they appear as plain target_link_libraries() entries. Since #4810 and #4820 made these two optional dependencies conditional via $<$<BOOL:...>:cudaq-nlopt|cudaq-ensmallen>, the installed CUDAQTargets.cmake now references the bare, unqualified names instead of cudaq::cudaq-nlopt / cudaq::cudaq-ensmallen. That these are the names consumers actually get is confirmed by cmake/modules/CUDAQNloptConfig.cmake and CUDAQEnsmallenConfig.cmake, whose `if(NOT TARGET cudaq::cudaq-nlopt)` and `if(NOT TARGET cudaq::cudaq-ensmallen)` guards are exactly what find_package(CUDAQNlopt)/find_package(CUDAQEnsmallen) are expected to produce. Downstream consumers that import cudaq via find_package(CUDAQ) (e.g. cudaqx) fail to link with "cannot find -lcudaq-ensmallen", even though the library and its CMake package are installed correctly.

Fix by aliasing each optimizer library under the cudaq:: namespace at the point it's defined, and referencing that alias in the generator expressions, so the namespaced name is correct from the start and needs no export-time rewriting.


sacpis (approved):
LGTM. Thanks @bmhowe23.

PR #4799, opened 2026-06-29
Harden QIR output record parsing

khalatepradnya:
* Validate metadata, container indices, and tuple layout bounds before accessing result storage. 
* Avoid misaligned array-buffer accesses.
* Add focused regression coverage for malformed output records.

PR #4832, opened 2026-07-02
More hardening for the QIR output-record parsing

khalatepradnya:
### Summary

* Follow-up to https://github.com/NVIDIA/cuda-quantum/pull/4799

- Hardens `RecordLogParser` against malformed output records.
-- Validates framing, schema-specific arity, and complete numeric tokens.
-- Accepts signed exponent, negative, and case-insensitive non-finite floating-point values.
-- Adds parser and end-to-end regression coverage.


cabreraam (approved):
I needed to get an understanding of the log schema and what the current limitations were, but now that I'm (sort of?) up to speed, LGTM! Thanks for making the QIR output record checking more rigorous!

khalatepradnya:
I see macOS failures since ` error: 'from_chars' is unavailable: introduced in macOS 26.0`, will address them later.

PR #4835, opened 2026-07-03
Fix cudaq.evolve with shared Hamiltonian and batched initial states

huaweil-nv:
## Summary
This PR fixes a `cudaq.evolve` batching bug on non-`dynamics` simulator targets when users provide a single Hamiltonian with multiple initial states.

The affected usage pattern is:
```
cudaq.evolve(
    hamiltonian,
    dimensions,
    schedule,
    [initial_state_0, initial_state_1],
)
```

Before this change, the non-dynamics path treated the call as batched because initial_state was a sequence, but then incorrectly zipped:
```
zip(hamiltonian, initial_state, collapse_operators)
```
When hamiltonian was a single operator and collapse_operators used the default empty list, the zip produced no items and cudaq.evolve silently returned an empty list instead of one result per initial state.

## Changes
* Reuse the same Hamiltonian and collapse operators when initial_state is a sequence but hamiltonian is not.
* Preserve the existing batched-Hamiltonian behavior when hamiltonian is a sequence.
* Reset the schedule before each per-state evolve_single call so each batched evolution starts from the full schedule.
* Add a regression test covering both qpp-cpu and density-matrix-cpu.


1tnguyen (approved):
LGTM 👍 Thanks @huaweil-nv

huaweil-nv:
> /ok to test 56e5a6693c13217f9c111ccd1c9a15ef6ed81f20

1tnguyen:
> /ok to test [9b11e38](https://github.com/NVIDIA/cuda-quantum/pull/4835/commits/9b11e3872f60dfedb4dd95ca44a37ac3da2d3ec9)

